PR #3, opened 2018-07-11
dev: Add a Dockerfile for development

olix0r:
When working on the proxy, it's important to be able to build a Docker
image that can be tested in the context of the existing linkerd2
project.

This change adds a `make docker` target that produces a docker image,
optionally tagged via the `DOCKER_TAG` environment variable.

This is intended to be used for development--especially on non-Linux
OSes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,12 @@ ifndef TEST_FLAKEY
 endif
 CARGO_TEST = $(CARGO) test --frozen $(RELEASE) $(TEST_FLAGS)
 
+DOCKER = docker
+DOCKER_BUILD = docker build
+ifdef DOCKER_TAG
+	DOCKER_BUILD = docker build -t $(DOCKER_TAG)
+endif
+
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -29,6 +29,12 @@ ifndef TEST_FLAKEY
 endif
 CARGO_TEST = $(CARGO) test --frozen $(RELEASE) $(TEST_FLAGS)
 
+DOCKER = docker
+DOCKER_BUILD = docker build
+ifdef DOCKER_TAG
+	DOCKER_BUILD = docker build -t $(DOCKER_TAG)
```
hawkw: Previously, in the `conduit` repository, there was a `bin/root-tag` script that could be used to determine what the tag would be if it's not overridden. It might be nice if the makefile had similar functionality eventually, but that may be out of scope for this change...

hawkw:
(Also, I've checked out this branch and manually verified that `make docker` works correctly)

PR #5, opened 2018-07-12
Add labels for `tls::ReasonForNoIdentity`

hawkw:
Fixes linkerd/linkerd2#1276.

Currently, metrics with the `tls="no_identity"` label are duplicated.
This is because that label is generated from the `tls_status` label on
the `TransportLabels` struct, which is either `Some(())` or a
`ReasonForNoTls`. `ReasonForNoTls` has a
variant`ReasonForNoTls::NoIdentity`, which contains a
`ReasonForNoIdentity`, but when we format that variant as a label, we
always just produce the string "no_identity", regardless of the value of
the `ReasonForNoIdentity`. 

However, label types are _also_ used as hash map keys into the map that
stores the metrics scopes, so although two instances of
`ReasonForNoTls::NoIdentity` with different `ReasonForNoIdentity`s
produce the same formatted label output, they aren't _equal_, since that
field differs, so they correspond to different metrics.

This branch resolves this issue, by adding an additional label to these
metrics, based on the `ReasonForNoIdentity`. Now, the separate lines in
the metrics output that correspond to each `ReasonForNoIdentity` have a
label differentiating them from each other.

Note that the the `NotImplementedForTap` and `NotImplementedForMetrics`
reasons will never show up in metrics labels, currently, since we don't gather 
metrics from the tap and metrics servers at the moment.

### DIFF
--- a/src/telemetry/metrics/labels/mod.rs
+++ b/src/telemetry/metrics/labels/mod.rs
@@ -426,7 +426,7 @@ impl fmt::Display for TlsStatus {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self.0 {
             Conditional::None(tls::ReasonForNoTls::NoIdentity(why)) =>
-                write!(f, "tls=\"no_identity\",no_identity=\"{}\"", why),
+                write!(f, "tls=\"no_identity\",tls_error_reason=\"{}\"", why),
             status => write!(f, "tls=\"{}\"", status),
         }
     }
--- a/src/telemetry/metrics/labels/mod.rs
+++ b/src/telemetry/metrics/labels/mod.rs
@@ -9,6 +9,7 @@ use std::{
 use http;
 
 use ctx;
+use conditional::Conditional;
 use telemetry::event;
 use transport::tls;
 
@@ -423,7 +424,11 @@ impl fmt::Display for TransportCloseLabels {
 
 impl fmt::Display for TlsStatus {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "tls=\"{}\"", self.0)
+        match self.0 {
+            Conditional::None(tls::ReasonForNoTls::NoIdentity(why)) =>
+                write!(f, "tls=\"no_identity\",no_identity=\"{}\"", why),
+            status => write!(f, "tls=\"{}\"", status),
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/telemetry/metrics/labels/mod.rs
```diff
@@ -423,7 +424,11 @@ impl fmt::Display for TransportCloseLabels {
 
 impl fmt::Display for TlsStatus {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "tls=\"{}\"", self.0)
+        match self.0 {
+            Conditional::None(tls::ReasonForNoTls::NoIdentity(why)) =>
+                write!(f, "tls=\"no_identity\",no_identity=\"{}\"", why),
```
olix0r: I at least suggest prefixing this as `tls_no_identity` so that these labels logically group together

Futhermore, do we expect that this may more generally useful as a `tls_error_reason`, etc?

briansmith: RE `tls_error_reason`, note that these reasons aren't necessarily explaining *error* conditions.

src/telemetry/metrics/labels/mod.rs
```diff
@@ -488,6 +493,27 @@ impl fmt::Display for TlsConfigLabels {
...
+            tls::ReasonForNoIdentity::NotHttp => f.pad("not_http"),
+            tls::ReasonForNoIdentity::NoAuthorityInHttpRequest =>
+                f.pad("no_authority_in_http_request"),
+            tls::ReasonForNoIdentity::NotProvidedByServiceDiscovery =>
+                f.pad("not_provided_by_service_discovery"),
+            tls::ReasonForNoIdentity::Loopback => f.pad("loopback"),
+            tls::ReasonForNoIdentity::NotConfigured => f.pad("not_configured"),
+            // Note: the following reasons will never show up in metrics labels
+            // currently, since we don't gather metrics from the tap and metrics
+            // servers.
```
briansmith: I suggest removing this comment because it will quickly become out of date and we'll forget to remove it then. It makes sense to be in the commit message though.

src/telemetry/metrics/labels/mod.rs
hawkw: @briansmith Yeah, I was a little iffy about the word "reason" in there, but I think we may eventually want to use this label for actual error conditions as well, and I couldn't think of a better name that wasn't way too verbose --- any suggestions?

briansmith: The problem isn't with the word "reason" but rather with the word "error". Not using TLS isn't an error, necessarily.

hawkw:
@briansmith and @olix0r, what do you think about changing the label to `no_tls_reason` (instead of `no_error_reason`) before merging this (based on https://github.com/linkerd/linkerd2-proxy/pull/5#discussion_r202138822)?

briansmith:
> @briansmith and @olix0r, what do you think about changing the label to no_tls_reason (instead of no_error_reason)

That seems better to me. I just think "error" is too misleading.

PR #6, opened 2018-07-12
Replace references to _Conduit_

olix0r:
There are various comments, examples, and documentation that refers to
Conduit. This change replaces or removes these refernces.

CONTRIBUTING.md has been updated to refer to GOVERNANCE/MAINTAINERS.

### DIFF
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -63,8 +63,8 @@ impl fmt::Display for ResolveOneCtx {
 
 impl Resolver {
 
-    /// Construct a new `Resolver` from the system configuration and Conduit's
-    /// environment variables.
+    /// Construct a new `Resolver` from the system configuration and the
+    /// proxy's environment variables.
     ///
     /// # Returns
     ///
--- a/src/telemetry/metrics/labels/mod.rs
+++ b/src/telemetry/metrics/labels/mod.rs
@@ -54,7 +54,7 @@ pub struct RequestLabels {
     direction: Direction,
 
     // Additional labels identifying the destination service of an outbound
-    // request, provided by the Conduit control plane's service discovery.
+    // request, provided by the control plane's service discovery.
     outbound_labels: Option<DstLabels>,
 
     /// The value of the `:authority` (HTTP/2) or `Host` (HTTP/1.1) header of
--- a/src/transparency/h1.rs
+++ b/src/transparency/h1.rs
@@ -62,10 +62,10 @@ fn set_authority(uri: &mut http::Uri, auth: Authority) {
 
     // If this was an origin-form target (path only),
     // then we can't *only* set the authority, as that's
-    // an illegal target (such as `conduit.io/docs`).
+    // an illegal target (such as `linkerd.io/docs`).
     //
     // But don't set a scheme if this was authority-form (CONNECT),
-    // since that would change it's meaning (like `https://conduit.io`).
+    // since that would change it's meaning (like `https://linkerd.io`).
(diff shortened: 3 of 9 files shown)


src/transparency/h1.rs
```diff
@@ -62,10 +62,10 @@ fn set_authority(uri: &mut http::Uri, auth: Authority) {
 
     // If this was an origin-form target (path only),
     // then we can't *only* set the authority, as that's
-    // an illegal target (such as `conduit.io/docs`).
+    // an illegal target (such as `linkerd.io/docs`).
     //
     // But don't set a scheme if this was authority-form (CONNECT),
-    // since that would change it's meaning (like `https://conduit.io`).
+    // since that would change it's meaning (like `https://linkerd.io`).
```
klingerf: Not your change, but mind fixing the typo here -- `it's` => `its`?

src/dns.rs
```diff
@@ -63,8 +63,8 @@ impl fmt::Display for ResolveOneCtx {
 
 impl Resolver {
 
-    /// Construct a new `Resolver` from the system configuration and Conduit's
-    /// environment variables.
+    /// Construct a new `Resolver` from the system configuration and the
+    /// proxy's environment variables.
```
briansmith: I would just remove "the proxy's" too. After all, who else are we going to get environment variables from, other than what's available in our own process?

src/telemetry/metrics/labels/mod.rs
```diff
@@ -54,7 +54,7 @@ pub struct RequestLabels {
     direction: Direction,
 
     // Additional labels identifying the destination service of an outbound
-    // request, provided by the Conduit control plane's service discovery.
+    // request, provided by the control plane's service discovery.
```
briansmith: I would remove "the control plane's" here because whether or not there's a control plane is irrelevant in this module.

src/transparency/h1.rs
```diff
@@ -62,10 +62,10 @@ fn set_authority(uri: &mut http::Uri, auth: Authority) {
 
     // If this was an origin-form target (path only),
     // then we can't *only* set the authority, as that's
-    // an illegal target (such as `conduit.io/docs`).
+    // an illegal target (such as `linkerd.io/docs`).
```
briansmith: Let's use `example.com` or `example.org` for example domains where the actual domain name doesn't matter (here and elsewhere throughout this PR). If we'd done that before we'd not need to do this now, and also using `example.com`/`example.org` makes it clearer to readers that the actual domain doesn't matter.

tests/telemetry.rs
```diff
@@ -510,7 +510,7 @@ mod outbound_dst_labels {
     // Ignore this test on CI, as it may fail due to the reduced concurrency
     // on CI containers causing the proxy to see both label updates from
     // the mock controller before the first request has finished.
-    // See https://github.com/runconduit/conduit/issues/751
+    // See https://github.com/linkerd/linkerd2/issues/751
```
briansmith: Continuing to use links is OK but maybe we should just shorten these to "issue #X" instead.

olix0r: because the repos are disparate, i think i'll shorten it to the github syntax linkerd/linkerd2#751?

briansmith: > because the repos are disparate, i think i'll shorten it to the github syntax linkerd/linkerd2#751?

I think the links would be better than that. The advantage of what I proposed is to remove the branding from the references. If you don't think that's important then I'd just leave them as links.

PR #8, opened 2018-07-12
remove busy loop from destination background future during shutdown

seanmonstar:
When the proxy is shutting down, once there are no more outbound
connections, the sender side of the resolver channel is dropped. In the
admin background thread, when the destination background future is
notified of the closure, instead of shutting down itself, it just busy
loops. Now, after seeing shutdown, the background future ends as well.

While examining this, I noticed all the background futures are joined
togther into a single `Future` before being spawned on a dedicated
current_thread executor. Join in this case is inefficient, since *every*
single time *one* of the futures is ready, they are *all* polled again.
Since we have an executor handy, it's better to allow it to manage each
of the futures individually.

------

I can't prove that this is the cause for https://github.com/linkerd/linkerd2/issues/1231, but this was the first obvious thing I could find.

### DIFF
--- a/src/control/destination/background.rs
+++ b/src/control/destination/background.rs
@@ -11,7 +11,7 @@ use bytes::Bytes;
 use futures::{
     future,
     sync::mpsc,
-    Async, Future, Stream,
+    Async, Future, Poll, Stream,
 };
 use h2;
 use http;
@@ -135,9 +135,7 @@ pub(super) fn task(
     );
 
     future::poll_fn(move || {
-        disco.poll_rpc(&mut client);
-
-        Ok(Async::NotReady)
+        disco.poll_rpc(&mut client)
     })
 }
 
@@ -163,22 +161,25 @@ where
         }
     }
 
-   fn poll_rpc(&mut self, client: &mut Option<T>) {
+   fn poll_rpc(&mut self, client: &mut Option<T>) -> Poll<(), ()> {
         // This loop is make sure any streams that were found disconnected
         // in `poll_destinations` while the `rpc` service is ready should
         // be reconnected now, otherwise the task would just sleep...
         loop {
-            self.poll_resolve_requests(client);
+            if let Async::Ready(()) = self.poll_resolve_requests(client) {
+                // request_rx has closed, meaning the main thread is terminating.
+                return Ok(Async::Ready(()));
+            }
             self.retain_active_destinations();
             self.poll_destinations();
 
             if self.reconnects.is_empty() || !self.rpc_ready {
-                break;
+                return Ok(Async::NotReady);
             }
         }
     }
 
-    fn poll_resolve_requests(&mut self, client: &mut Option<T>) {
+    fn poll_resolve_requests(&mut self, client: &mut Option<T>) -> Async<()> {
         loop {
             if let Some(client) = client {
                 // if rpc service isn't ready, not much we can do...
@@ -188,12 +189,12 @@ where
                     },
                     Ok(Async::NotReady) => {
                         self.rpc_ready = false;
-                        break;
+                        return Async::NotReady;
                     },
                     Err(err) => {
                         warn!("Destination.Get poll_ready error: {:?}", err);
                         self.rpc_ready = false;
-                        break;
+                        return Async::NotReady;
                     },
                 }
 
@@ -255,10 +256,10 @@ where
                     }
                 },
                 Ok(Async::Ready(None)) => {
-                    trace!("Discover tx is dropped, shutdown?");
-                    return;
+                    trace!("Discover tx is dropped, shutdown");
+                    return Async::Ready(());
                 },
-                Ok(Async::NotReady) => break,
+                Ok(Async::NotReady) => return Async::NotReady,
                 Err(_) => unreachable!("unbounded receiver doesn't error"),
             }
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -333,18 +333,15 @@ where
 
                     let metrics_server = telemetry.serve_metrics(metrics_listener);
 
-                    let fut = ::logging::admin().bg("resolver").future(resolver_bg)
-                        .join5(
-                            ::logging::admin().bg("telemetry").future(telemetry),
-                            tap.map_err(|_| {}),
-                            metrics_server.map_err(|_| {}),
-                            ::logging::admin().bg("dns-resolver").future(dns_bg),
-                        )
-                        // There's no `Future::join6` combinator...
-                        .join(::logging::admin().bg("tls-config").future(tls_cfg_bg))
-                        .map(|_| {});
+                    rt.spawn(::logging::admin().bg("resolver").future(resolver_bg));
+                    rt.spawn(::logging::admin().bg("telemetry").future(telemetry));
+                    // tap is already wrapped in a logging Future.
+                    rt.spawn(tap);
+                    // metrics_server is already wrapped in a logging Future.
+                    rt.spawn(metrics_server);
+                    rt.spawn(::logging::admin().bg("dns-resolver").future(dns_bg));
 
-                    rt.spawn(Box::new(fut));
+                    rt.spawn(::logging::admin().bg("tls-config").future(tls_cfg_bg));
 
                     let shutdown = admin_shutdown_signal.then(|_| Ok::<(), ()>(()));
                     rt.block_on(shutdown).expect("admin");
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -333,18 +333,15 @@ where
...
-                            tap.map_err(|_| {}),
-                            metrics_server.map_err(|_| {}),
-                            ::logging::admin().bg("dns-resolver").future(dns_bg),
-                        )
-                        // There's no `Future::join6` combinator...
-                        .join(::logging::admin().bg("tls-config").future(tls_cfg_bg))
-                        .map(|_| {});
-
-                    rt.spawn(Box::new(fut));
+                    rt.spawn(::logging::admin().bg("resolver").future(resolver_bg));
```
hawkw: 👍 This is much cleaner, too!

src/control/destination/background.rs
```diff
@@ -163,22 +161,25 @@ where
         }
     }
 
-   fn poll_rpc(&mut self, client: &mut Option<T>) {
+   fn poll_rpc(&mut self, client: &mut Option<T>) -> Poll<(), ()> {
```
hawkw: I'd been thinking about making this change anyway, just for clarity's sake. 👍

src/lib.rs
```diff
@@ -333,18 +333,15 @@ where
...
+                    rt.spawn(::logging::admin().bg("telemetry").future(telemetry));
+                    // tap is already wrapped in a logging Future.
```
briansmith: Would it be problematic to make this consistent, so that either all of the wrapping in logging futures is done here, or it is all done when they are constructed? I'd slightly prefer the wrapping to be done here, but it seems like a good idea to at least be consistent or explain *why* we're not consistent.

briansmith:
> While examining this, I noticed all the background futures are joined
> togther into a single Future before being spawned on a dedicated
> current_thread executor. Join in this case is inefficient, since every
> single time one of the futures is ready, they are all polled again.
> Since we have an executor handy, it's better to allow it to manage each
> of the futures individually.

Not a huge deal, but it would be better in the future if we split these kinds of changes into their own PR when practical.

src/lib.rs
seanmonstar: Peeking around, it seems those 2 (tap and metrics_server) are wrapping themselves after grabbing some relevant values from other arguments (such as the `local_addr`), and *then* making a clone of the logging context to be used for all internally spawned futures, so they share the context.

I just added these comments because *I* was also confused why it looked like some were wrapped and some were forgotten.

seanmonstar:
> Not a huge deal, but it would be better in the future if we split these kinds of changes into their own PR when practical.

Yea, I could have gone either way this. It turned out it was the first thing I fixed, as I was also worried that the big joined future meant a lot of wasted polls, which could also increase CPU usage. Then I looked into each one, to see if they busy looped.

klingerf:
As far as I can tell, this change has introduced a regression in the linekrd2 integration tests. Specifically, when installing linkerd2 without TLS enabled, the web pod is no longer able to serve HTTP requests from the dashboard. The integration test failure looks like this:

```
--- FAIL: TestDashboard (61.51s)
  install_test.go:161: Unexpected error: Get http://127.0.0.1:52237/api/v1/namespaces/kltest/services/web:http/proxy/api/version: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

In the web pod's linkerd-proxy log, I see:

```
2018-07-13T17:17:22.288270536Z ERR! proxy={server=out listen=127.0.0.1:4140 remote=172.17.0.9:55282} linkerd2_proxy turning operation timed out after 10s into 500
2018-07-13T17:17:22.290554495Z ERR! proxy={server=in listen=0.0.0.0:4143 remote=172.17.0.1:56144} linkerd2_proxy turning Error caused by underlying HTTP/2 error: protocol error: unexpected internal error encountered into 500
```

If I roll the version of the linkerd2-proxy that I'm using back to 3c48ba7, the test passes. Once I introduce 4bee7b0, however, the test fails.

Interestingly enough, if I install linkerd2 with `--tls optional`, then the dashboard works as expected. So the bug appears to only impact installations that do not use TLS.

hawkw:
@klingerf I don't suppose you have any more proxy logs from the failing test?

klingerf:
@hawkw Here you go:

```
$ kubectl -n kltest logs web-554d5db5c-tbzzf linkerd-proxy --timestamps
2018-07-13T18:29:12.305595895Z INFO linkerd2_proxy using controller at Some(HostAndPort { host: DnsName(DnsName(DNSName("proxy-api.kltest.svc.cluster.local"))), port: 8086 })
2018-07-13T18:29:12.305654103Z INFO linkerd2_proxy routing on V4(127.0.0.1:4140)
2018-07-13T18:29:12.30565915Z INFO linkerd2_proxy proxying on V4(0.0.0.0:4143) to None
2018-07-13T18:29:12.305662345Z INFO linkerd2_proxy serving Prometheus metrics on V4(0.0.0.0:4191)
2018-07-13T18:29:12.31245236Z INFO linkerd2_proxy protocol detection disabled for inbound ports {25, 3306}
2018-07-13T18:29:12.312507263Z INFO linkerd2_proxy protocol detection disabled for outbound ports {25, 3306}
2018-07-13T18:29:12.443255624Z INFO admin={bg=resolver} linkerd2_proxy::transport::connect DNS resolved DnsName(DnsName(DNSName("proxy-api.kltest.svc.cluster.local"))) to 10.108.53.152
2018-07-13T18:29:15.42286033Z ERR! admin={bg=resolver} linkerd2_proxy::control::util controller error: Error attempting to establish underlying session layer: operation timed out after 3s
2018-07-13T18:30:00.175927154Z ERR! proxy={server=out listen=127.0.0.1:4140 remote=172.17.0.9:51556} linkerd2_proxy turning operation timed out after 10s into 500
2018-07-13T18:30:00.177140101Z ERR! proxy={server=in listen=0.0.0.0:4143 remote=172.17.0.1:47130} linkerd2_proxy turning Error caused by underlying HTTP/2 error: protocol error: unexpected internal error encountered into 500
2018-07-13T18:30:11.160580759Z ERR! proxy={server=out listen=127.0.0.1:4140 remote=172.17.0.9:51556} linkerd2_proxy turning operation timed out after 10s into 500
2018-07-13T18:30:11.161441109Z ERR! proxy={server=in listen=0.0.0.0:4143 remote=172.17.0.1:47270} linkerd2_proxy turning Error caused by underlying HTTP/2 error: protocol error: unexpected internal error encountered into 500
2018-07-13T18:30:21.162606831Z ERR! proxy={server=out listen=127.0.0.1:4140 remote=172.17.0.9:51556} linkerd2_proxy turning operation timed out after 10s into 500
2018-07-13T18:30:21.163332894Z ERR! proxy={server=in listen=0.0.0.0:4143 remote=172.17.0.1:47386} linkerd2_proxy turning Error caused by underlying HTTP/2 error: protocol error: unexpected internal error encountered into 500
2018-07-13T18:30:31.166428078Z ERR! proxy={server=out listen=127.0.0.1:4140 remote=172.17.0.9:51556} linkerd2_proxy turning operation timed out after 10s into 500
2018-07-13T18:30:31.171456621Z ERR! proxy={server=in listen=0.0.0.0:4143 remote=172.17.0.1:47502} linkerd2_proxy turning Error caused by underlying HTTP/2 error: protocol error: unexpected internal error encountered into 500
2018-07-13T18:30:41.173926844Z ERR! proxy={server=out listen=127.0.0.1:4140 remote=172.17.0.9:51556} linkerd2_proxy turning operation timed out after 10s into 500
2018-07-13T18:30:41.176118471Z ERR! proxy={server=in listen=0.0.0.0:4143 remote=172.17.0.1:47618} linkerd2_proxy turning Error caused by underlying HTTP/2 error: protocol error: unexpected internal error encountered into 500
```

seanmonstar:
Oh noes! Is it possible to get logs at the trace level? Those include some lines about when this background service does work and when it shuts down.

briansmith:
We should back this out until the problem is fixed. We shouldn't leave it in the broken state while we figure out what's wrong.

seanmonstar:
Hm, that trace-log seems to show the destination service is working, and I only see one error log, about an HTTP/1 request where the supposedly the response wasn't complete.

seanmonstar:
@briansmith I'd definitely agree, except that looking at the logs, it doesn't actually *look* like it's related to this change...

seanmonstar:
Re-scanning the log file, I see that the incomplete error message is from logs of another proxy, noticing a socket closing before getting a response. It's sufficiently confusing that I'll prepare a revert, just in case.

seanmonstar:
Actually, I have this suspicion now: I wonder if the destination background future sometimes returns without having polled something to register its task, but because it was joined with the other futures, when *they* were ready, the destination task would get polled again anyways...

PR #10, opened 2018-07-13
spawn individual admin tasks instead of joining all

seanmonstar:
This was originally in #8, but I've separated it out since it is conceptually a different thing to fix.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -333,18 +333,15 @@ where
 
                     let metrics_server = telemetry.serve_metrics(metrics_listener);
 
-                    let fut = ::logging::admin().bg("resolver").future(resolver_bg)
-                        .join5(
-                            ::logging::admin().bg("telemetry").future(telemetry),
-                            tap.map_err(|_| {}),
-                            metrics_server.map_err(|_| {}),
-                            ::logging::admin().bg("dns-resolver").future(dns_bg),
-                        )
-                        // There's no `Future::join6` combinator...
-                        .join(::logging::admin().bg("tls-config").future(tls_cfg_bg))
-                        .map(|_| {});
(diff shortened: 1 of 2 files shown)


olix0r:
@seanmonstar Is this a functional change? or just an aesthetic change?

hawkw:
@olix0r I _believe_ this results in at least some performance improvements. As @seanmonstar commented in the #8 PR description:

> While examining this, I noticed all the background futures are joined
togther into a single `Future` before being spawned on a dedicated
current_thread executor. Join in this case is inefficient, since every
single time one of the futures is ready, they are all polled again.
Since we have an executor handy, it's better to allow it to manage each
of the futures individually.

seanmonstar:
It is functional. The runtime knows which future needs to be woken up when notified, but when it's just a single large `join5`, any of the 5 futures needing a wake up means the runtime only knows it as 1 future, and wakes the whole `join5` up.

However, I'm starting to suspect that this behavior was actually patching over that the destination task might not have been registering for wakeups, but since the other tasks were, it'd get polled anyways...

olix0r:
> However, I'm starting to suspect that this behavior was actually patching over that the destination task might not have been registering for wakeups, but since the other tasks were, it'd get polled anyways...

Meaning that we probably should hold off on merging this until we've answered this question? Is that correct?

hawkw:
I'm testing this branch to see whether or not it introduces the same problem as #8.

hawkw:
After a preliminary test, it appears that the dashboard doesn't load correctly on a proxy built from this branch, either.

seanmonstar:
Rebased now that #13 was merged, this should no longer break us.

PR #13, opened 2018-07-13
fix control client Backoff to poll its timer when backing off

seanmonstar:
The `Backoff` service wrapper is used for the controller client service
so that if the proxy can't find the controller (there is a connection
error), it doesn't keep trying in a tight loop, but instead waits a
couple seconds before trying again, presuming that the control plane
was rebooting.

When "backing off", a timer would be set, but it wasn't polled, so the
task was never registered to wake up after the delay. This turns out to
not have been a problem in practice, since the background destination
task was joined with other tasks that were constantly waking up,
allowing it to try again anyways.

To add tests for this, a new `ENV_CONTROL_BACKOFF_DELAY` config value
has been added, so that the tests don't have to wait the default 5
seconds.

Signed-off-by: Sean McArthur <sean@buoyant.io>

### DIFF
--- a/src/control/util.rs
+++ b/src/control/util.rs
@@ -21,13 +21,30 @@ pub(super) struct Backoff<S> {
     wait_dur: Duration,
 }
 
-impl<S> Backoff<S> {
+impl<S> Backoff<S>
+where
+    S: Service,
+{
     pub(super) fn new(inner: S, wait_dur: Duration) -> Self {
         Backoff {
             inner,
             timer: Delay::new(Instant::now() + wait_dur),
             waiting: false,
             wait_dur,
+        }
+    }
+
+    fn poll_timer(&mut self) -> Poll<(), S::Error> {
+        debug_assert!(self.waiting, "poll_timer expects to be waiting");
+
+        match self.timer.poll().expect("timer shouldn't error") {
+            Async::Ready(()) => {
+                self.waiting = false;
+                Ok(Async::Ready(()))
+            }
+            Async::NotReady => {
+                Ok(Async::NotReady)
+            }
         }
     }
 }
@@ -44,11 +61,7 @@ where
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
         if self.waiting {
-            if self.timer.poll().unwrap().is_not_ready() {
-                return Ok(Async::NotReady);
-            }
-
-            self.waiting = false;
+            try_ready!(self.poll_timer());
         }
 
         match self.inner.poll_ready() {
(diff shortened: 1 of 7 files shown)


hawkw:
🙌

src/control/util.rs
```diff
@@ -30,6 +33,20 @@ impl<S> Backoff<S> {
             wait_dur,
         }
     }
+
+    fn poll_timer(&mut self) -> Poll<(), S::Error> {
```
hawkw: It looks like this never returns an error, so I wonder if this is really the appropriate return type( (instead of, say, `Async<()>`)? In that case, we wouldn't need the `S: Service` bound on this impl.

On the other hand, this lets us use `try_ready!` in the `Service` impl for `Backoff` below, so maybe it's the right thing. 🤷‍♀️

tests/discovery.rs
```diff
@@ -153,6 +153,47 @@ macro_rules! generate_tests {
             assert_eq!(client.get("/"), "hello");
             assert_eq!(client.get("/bye"), "bye");
         }
+
+        #[test]
+        fn outbound_error_reconnects_after_backoff() {
```
hawkw: Thanks for adding a test for this (and I appreciate the amount of comments in the test, as well).

However, I wonder if we might also want to have a more granular unit test for just the backoff behaviour as well? Since the failure to notify the backoff future is masked by the way the various control futures are `join`ed and are thus all polled, the backoff future still gets polled when it's inside of `Proxy::run`, maybe we would also want to test the backoff future in isolation.

I'm imagining something along the lines of:

```rust

struct MockService {
    polls: usize,
    succeed_after: usize,
}

impl MockService {
    fn new(succeed_after: usize) -> Self {
          Self { polls: 0, succeed_after, }
    }
}

impl Service for MockService {
    type Request = ();
    type Response = ();
    type Error = ();
    type Future = FutureResult<(), ()>;

        self.polls += 1;
        if self.polls > self.succeed_after {
            Ok(Async::Ready(()))
        } else {
            Err(())
        }
    }

    fn call(&mut self, req: Self::Request) -> Self::Future {
        if self.polls > self.succeed_after {
            future::ok(())
        } else {
            future::err(())
        }
    }
}

#[test]
fn test_backoff() {
    let mock = MockService::new(2);
    let mut backoff = Backoff::new(mock, Duration::from_millis(5)); // or whatever.
    let mut rt = tokio::runtime::current_thread::Runtime::new()
        .unwrap();
    let call = rt.block_on_for(backoff.call(), Duration::from_millis(15));
    assert!(call.is_ok());
    assert_eq!(mock.polls, 3);
}

```

PR #15, opened 2018-07-16
remove busy loop from destination background future when shutdown

seanmonstar:
When the proxy is shutting down, once there are no more outbound
connections, the sender side of the resolver channel is dropped. In the
admin background thread, when the destination background future is
notified of the closure, instead of shutting down itself, it just busy
loops. Now, after seeing shutdown, the background future ends as well.

With #13 merged, and #10 being a separate change, this shouldn't break the proxy again. Related to https://github.com/linkerd/linkerd2/issues/1231

### DIFF
--- a/src/control/destination/background.rs
+++ b/src/control/destination/background.rs
@@ -11,7 +11,7 @@ use bytes::Bytes;
 use futures::{
     future,
     sync::mpsc,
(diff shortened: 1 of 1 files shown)


olix0r:
retroactive shipit ;)

PR #18, opened 2018-07-17
Improve error messages in logs

hawkw:
Currently, the messages that the proxy logs on errors are often not 
very useful. For example, when an error occurred that caused the proxy
to return HTTP 500, we log a message like this:

```
ERR! proxy={server=in listen=0.0.0.0:4143 remote=127.0.0.1:57416} linkerd2_proxy turning Error caused by underlying HTTP/2 error: protocol error: unexpected internal error encountered into 500
```

Note that:
+ Regardless of what the error actually was, the current log message
  *always* says protocol error: unexpected internal error encountered,
  which is both fairly unclear *and* often not actually the case.
+ Regardless of whether the error was encountered by an HTTP/1 or 
  HTTP/2 client, the error message always includes the string
  "underlying HTTP/2 error". This is probably fairly confusing for 
  users who are, say, only proxying HTTP/1 traffic.

This branch fixes several related issues around the clarity of the
proxy's error messages:

+ A couple cases in the `transparency` module that returned
  `io::Error::from(io::ErrorKind::Other)` have been replaced with
  more descriptive errors that propagate the underlying error. This
  necessitated adding bounds on some error types.
+ Introduced a new `transparency::client::Error` enum that can be 
  either a `h2::Error` or a `hyper::Error`, depending on whether
  the client is HTTP/1 or HTTP/2, and proxies its' `std::error::Error`
  impl to the wrapped error types. This way, we don't return a
  `tower_h2::client::Error` (with format output that includes the 
  string "HTTP/2") from everything, and discard significantly less
  error information.

<details>
<summary>Examples</summary>
As an example, here's what happens when I hit the proxy's inbound 
router with no inner service running.

Before this change:
```
➜ LINKERD2_PROXY_POD_NAMESPACE=foo LINKERD2_PROXY_POD_NAME=bar LINKERD2_PROXY_LOG=info,linkerd2_proxy=debug LINKERD2_PROXY_PRIVATE_FORWARD=tcp://127.0.0.1:8888 cargo run
   Compiling linkerd2-proxy v0.1.0 (file:///Users/eliza/Code/linkerd2-proxy)
    Finished dev [unoptimized + debuginfo] target(s) in 28.95s
DBUG linkerd2_proxy::task creating single-threaded proxy
INFO linkerd2_proxy using controller at None
INFO linkerd2_proxy routing on V4(127.0.0.1:4140)
INFO linkerd2_proxy proxying on V4(0.0.0.0:4143) to Some(Addr(V4(127.0.0.1:8888)))
INFO linkerd2_proxy serving Prometheus metrics on V4(127.0.0.1:4191)
INFO linkerd2_proxy protocol detection disabled for inbound ports {25, 3306}
INFO linkerd2_proxy protocol detection disabled for outbound ports {25, 3306}
INFO linkerd2_proxy::telemetry::metrics procinfo not supported on this operating system
DBUG proxy={server=in listen=0.0.0.0:4143} linkerd2_proxy::transport::addr_info no support for SO_ORIGINAL_DST
DBUG proxy={server=in listen=0.0.0.0:4143} linkerd2_proxy::telemetry::sensor server connection open
DBUG proxy={server=in listen=0.0.0.0:4143 remote=127.0.0.1:57416} linkerd2_proxy::inbound building inbound Http1 { host: Authority(localhost:4143), was_absolute_form: false } client to 127.0.0.1:8888
DBUG proxy={server=in listen=0.0.0.0:4143 remote=127.0.0.1:57416} linkerd2_proxy::bind bind_stack: endpoint=Endpoint { address: V4(127.0.0.1:8888), metadata: Metadata { dst_labels: None, tls_identity: None(NotProvidedByServiceDiscovery) } }, protocol=Http1 { host: Authority(localhost:4143), was_absolute_form: false }
DBUG proxy={server=in listen=0.0.0.0:4143 remote=127.0.0.1:57416} linkerd2_proxy::bind rebinding endpoint stack for Endpoint { address: V4(127.0.0.1:8888), metadata: Metadata { dst_labels: None, tls_identity: None(NotProvidedByServiceDiscovery) } }:Http1 { host: Authority(localhost:4143), was_absolute_form: false } on TLS config change
DBUG proxy={server=in listen=0.0.0.0:4143 remote=127.0.0.1:57416} linkerd2_proxy::bind bind_inner_stack endpoint=Endpoint { address: V4(127.0.0.1:8888), metadata: Metadata { dst_labels: None, tls_identity: None(NotProvidedByServiceDiscovery) } }, protocol=Http1 { host: Authority(localhost:4143), was_absolute_form: false }
DBUG proxy={client=in dst=local remote=127.0.0.1:8888} linkerd2_proxy::transparency::client client request: method=GET uri=http://localhost:4143/ headers={"host": "localhost:4143", "user-agent": "HTTPie/0.9.9", "accept-encoding": "gzip, deflate", "accept": "*/*"}
DBUG proxy={server=in listen=0.0.0.0:4143 remote=127.0.0.1:57416} linkerd2_proxy::transparency::client http/1 client error: an error occurred trying to connect: other os error
ERR! proxy={server=in listen=0.0.0.0:4143 remote=127.0.0.1:57416} linkerd2_proxy turning Error caused by underlying HTTP/2 error: protocol error: unexpected internal error encountered into 500
```

After this change:
```
➜ LINKERD2_PROXY_POD_NAMESPACE=foo LINKERD2_PROXY_POD_NAME=bar LINKERD2_PROXY_LOG=info,linkerd2_proxy=debug LINKERD2_PROXY_PRIVATE_FORWARD=tcp://127.0.0.1:8888 cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.28s
DBUG linkerd2_proxy::task creating single-threaded proxy
INFO linkerd2_proxy using controller at None
INFO linkerd2_proxy routing on V4(127.0.0.1:4140)
INFO linkerd2_proxy proxying on V4(0.0.0.0:4143) to Some(Addr(V4(127.0.0.1:8888)))
INFO linkerd2_proxy serving Prometheus metrics on V4(127.0.0.1:4191)
INFO linkerd2_proxy protocol detection disabled for inbound ports {25, 3306}
INFO linkerd2_proxy protocol detection disabled for outbound ports {25, 3306}
INFO linkerd2_proxy::telemetry::metrics procinfo not supported on this operating system
DBUG proxy={server=in listen=0.0.0.0:4143} linkerd2_proxy::transport::addr_info no support for SO_ORIGINAL_DST
DBUG proxy={server=in listen=0.0.0.0:4143} linkerd2_proxy::telemetry::sensor server connection open
DBUG proxy={server=in listen=0.0.0.0:4143 remote=127.0.0.1:57406} linkerd2_proxy::inbound building inbound Http1 { host: Authority(localhost:4143), was_absolute_form: false } client to 127.0.0.1:8888
DBUG proxy={server=in listen=0.0.0.0:4143 remote=127.0.0.1:57406} linkerd2_proxy::bind bind_stack: endpoint=Endpoint { address: V4(127.0.0.1:8888), metadata: Metadata { dst_labels: None, tls_identity: None(NotProvidedByServiceDiscovery) } }, protocol=Http1 { host: Authority(localhost:4143), was_absolute_form: false }
DBUG proxy={server=in listen=0.0.0.0:4143 remote=127.0.0.1:57406} linkerd2_proxy::bind rebinding endpoint stack for Endpoint { address: V4(127.0.0.1:8888), metadata: Metadata { dst_labels: None, tls_identity: None(NotProvidedByServiceDiscovery) } }:Http1 { host: Authority(localhost:4143), was_absolute_form: false } on TLS config change
DBUG proxy={server=in listen=0.0.0.0:4143 remote=127.0.0.1:57406} linkerd2_proxy::bind bind_inner_stack endpoint=Endpoint { address: V4(127.0.0.1:8888), metadata: Metadata { dst_labels: None, tls_identity: None(NotProvidedByServiceDiscovery) } }, protocol=Http1 { host: Authority(localhost:4143), was_absolute_form: false }
DBUG proxy={client=in dst=local remote=127.0.0.1:8888} linkerd2_proxy::transparency::client client request: method=GET uri=http://localhost:4143/ headers={"host": "localhost:4143", "user-agent": "HTTPie/0.9.9", "accept-encoding": "gzip, deflate", "accept": "*/*"}
DBUG proxy={server=in listen=0.0.0.0:4143 remote=127.0.0.1:57406} linkerd2_proxy::transparency::client http/1 client error: an error occurred trying to connect: Connection refused (os error 61); cause: Connection refused (os error 61);
ERR! proxy={server=in listen=0.0.0.0:4143 remote=127.0.0.1:57406} linkerd2_proxy turning an error occurred trying to connect: Connection refused (os error 61) into 500
^X^CINFO linkerd2_proxy::signal received SIGINT, starting shutdown
DBUG linkerd2_proxy shutdown signaled
DBUG linkerd2_proxy shutdown complete
```

Note that after the changes in this branch, we actually output that the
error was caused by "Connection refused (os error 61)", rather than
an "unexpected internal error", and the string "HTTP/2" is no longer
present in the log output for an HTTP/1 request.

</details>

### DIFF
--- a/src/transparency/client.rs
+++ b/src/transparency/client.rs
@@ -91,6 +91,7 @@ impl<C, E, B> Client<C, E, B>
 where
     C: Connect + Clone + Send + Sync + 'static,
     C::Future: Send + 'static,
+    <C::Future as Future>::Error: ::std::error::Error + Send + Sync,
     C::Connected: Send,
     E: Executor + Clone,
     E: future::Executor<Box<Future<Item = (), Error = ()> + Send + 'static>> + Send + Sync + 'static,
@@ -130,6 +131,7 @@ impl<C, E, B> NewService for Client<C, E, B>
 where
     C: Connect + Clone + Send + Sync + 'static,
     C::Future: Send + 'static,
+    <C::Future as Future>::Error: ::std::error::Error + Send + Sync,
     C::Connected: Send,
     E: Executor + Clone,
     E: future::Executor<Box<Future<Item = (), Error = ()> + Send + 'static>> + Send + Sync + 'static,
@@ -192,6 +194,7 @@ where
     C: Connect + Send + Sync + 'static,
     C::Connected: Send,
     C::Future: Send + 'static,
+    <C::Future as Future>::Error: ::std::error::Error + Send + Sync,
     E: Executor + Clone,
     E: future::Executor<Box<Future<Item = (), Error = ()> + Send + 'static>> + Send + Sync + 'static,
     B: tower_h2::Body + Send + 'static,
@@ -263,7 +266,13 @@ impl Future for ClientServiceFuture {
                     },
                     Ok(Async::NotReady) => Ok(Async::NotReady),
                     Err(e) => {
-                        debug!("http/1 client error: {}", e);
+                        match e.cause2() {
+                            Some(cause) => debug!(
+                                "http/1 client error: {}; cause: {}; ",
+                                e, cause
+                            ),
+                            None => debug!("http/1 client error: {}", e),
+                        };
                         Err(h2::Reason::INTERNAL_ERROR.into())
                     }
                 }
--- a/src/transparency/client.rs
+++ b/src/transparency/client.rs
@@ -14,8 +14,16 @@ use telemetry::sensor::http::RequestBody;
 use super::glue::{BodyPayload, HttpBody, HyperConnect};
 use super::upgrade::{HttpConnect, Http11Upgrade};
 
+use std::{self, fmt};
+
 type HyperClient<C, B> =
     hyper::Client<HyperConnect<C>, BodyPayload<RequestBody<B>>>;
+
+#[derive(Debug)]
+pub enum Error {
+    Http1(hyper::Error),
+    Http2(tower_h2::client::Error),
+}
 
 /// A `NewService` that can speak either HTTP/1 or HTTP/2.
 pub struct Client<C, E, B>
@@ -140,7 +148,7 @@ where
 {
     type Request = bind::HttpRequest<B>;
     type Response = http::Response<HttpBody>;
-    type Error = tower_h2::client::Error;
+    type Error = Error;
     type InitError = tower_h2::client::ConnectError<C::Error>;
     type Service = ClientService<C, E, B>;
     type Future = ClientNewServiceFuture<C, E, B>;
@@ -202,13 +210,13 @@ where
 {
     type Request = bind::HttpRequest<B>;
     type Response = http::Response<HttpBody>;
-    type Error = tower_h2::client::Error;
+    type Error = Error;
     type Future = ClientServiceFuture;
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
         match self.inner {
             ClientServiceInner::Http1(_) => Ok(Async::Ready(())),
-            ClientServiceInner::Http2(ref mut h2) => h2.poll_ready(),
+            ClientServiceInner::Http2(ref mut h2) => h2.poll_ready().map_err(Error::from),
         }
     }
 
@@ -248,7 +256,7 @@ pub enum ClientServiceFuture {
 
 impl Future for ClientServiceFuture {
     type Item = http::Response<HttpBody>;
-    type Error = tower_h2::client::Error;
+    type Error = Error;
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         match self {
@@ -273,7 +281,7 @@ impl Future for ClientServiceFuture {
                             ),
                             None => debug!("http/1 client error: {}", e),
                         };
-                        Err(h2::Reason::INTERNAL_ERROR.into())
+                        Err(e.into())
                     }
                 }
             },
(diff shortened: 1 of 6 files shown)


src/transparency/glue.rs
```diff
@@ -303,15 +306,15 @@ where
...
+    F::Error: Error + fmt::Debug + Send + Sync + 'static,
 {
     type Item = http::Response<BodyPayload<B>>;
     type Error = h2::Error;
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         let mut res = try_ready!(self.inner.poll().map_err(|e| {
-            debug!("h2 error: {:?}", e);
-            h2::Error::from(io::Error::from(io::ErrorKind::Other))
+            debug!("h2 error: {}", e);
```
seanmonstar: This log message looks funny... it's not actually an h2 error, right?

hawkw: Ah, you're right! Will fix.

src/transparency/client.rs
```diff
@@ -14,9 +14,17 @@ use telemetry::sensor::http::RequestBody;
...
 
+use std::{self, fmt};
+
 type HyperClient<C, B> =
     hyper::Client<HyperConnect<C>, BodyPayload<RequestBody<B>>>;
 
+#[derive(Debug)]
+pub enum Error {
+    Http1(hyper::Error),
+    Http2(tower_h2::client::Error),
```
briansmith: This makes it seem like non-version-specific errors will be reported two different ways depending what the underyling version is. I suspect there are very few errors that are really protocol-version-specific, and I also suspect that the protocol-version-specific errors are mostly H2-specific and not very interesting to end users, so it seems like it would be good to find a way to unify these errors in the future.

src/transparency/glue.rs
```diff
@@ -303,15 +306,15 @@ where
...
 {
     type Item = http::Response<BodyPayload<B>>;
     type Error = h2::Error;
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         let mut res = try_ready!(self.inner.poll().map_err(|e| {
-            debug!("h2 error: {:?}", e);
-            h2::Error::from(io::Error::from(io::ErrorKind::Other))
+            debug!("HTTP/1 error: {}", e);
+            h2::Error::from(io::Error::new(io::ErrorKind::Other, e))
```
briansmith: Do we need a TODO or something about clarifying this? It's strange to construct an `h2::Error` for a HTTP/1 error.

src/transparency/client.rs
seanmonstar: Yea, we definitely should! A problem currently is that metrics (really, tap) requires that we send an HTTP2 Error Code if there is any error at all, which doesn't make sense if the request were HTTP/1 (for reference, in that case, I believe we just report `INTERNAL_ERROR`...).

hawkw: > This makes it seem like non-version-specific errors will be reported two different ways depending what the underyling version is

That's a good point. I think ideally, we'd want to handle cases where the error was caused by, say, an `io::Error` and map that to a separate variant, and restrict the `Http1`/`Http2` variants to errors specific to those protocols. For now, though, I think I'll just add a comment saying that these refer to errors in the HTTP/1 or HTTP/2 _client_ rather than errors specific to those protocols.

src/transparency/glue.rs
hawkw: Actually, I just checked, and it turns out there's no longer any reason for this to be an `h2::Error` --- the future's error type can now be changed to `S::Error`, and this conversion can be removed.

src/transparency/client.rs
hawkw: > A problem currently is that metrics (really, tap) requires that we send an HTTP2 Error Code if there is any error at all, which doesn't make sense if the request were HTTP/1 (for reference, in that case, I believe we just report INTERNAL_ERROR...)

I'd like to fix this, but my understanding is that it isn't currently prioritized.

PR #21, opened 2018-07-18
Configure listen ports' TLS when constructing them.

briansmith:
The way TLS is done for a bound port is fixed based on its role and whatever
the TLS settings are, so it makes sense to configure the TLS aspects of the
bound port during construction. This will also make writing tests easier.

Signed-off-by: Brian Smith <brian@briansmith.org>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -119,6 +119,7 @@ pub use watch_service::WatchService;
 
 pub struct Main<G> {
     config: config::Config,
+    tls_config_watch: tls::ConfigWatch,
 
     control_listener: BoundPort,
     inbound_listener: BoundPort,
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -276,15 +294,8 @@ where
...
                 config.inbound_router_max_idle_age,
             );
-            let tls_settings = config.tls_settings.as_ref().map(|settings| {
-                tls::ConnectionConfig {
-                    identity: settings.pod_identity.clone(),
-                    config: tls_server_config
-                }
-            });
             serve(
                 inbound_listener,
```
hawkw: Seeing a branch that _reduces_ the number of parameters passed to this function rather than increasing it is refreshing! :)

PR #25, opened 2018-07-21
fixed some typos in comments and Dockerfile

MarkusJais:
Signed-off-by: Markus Jais <markusjais@googlemail.com>

no code changes, just some fixed typos in comments and Dockerfile

### DIFF
--- a/src/control/destination/endpoint.rs
+++ b/src/control/destination/endpoint.rs
@@ -7,7 +7,7 @@ use conditional::Conditional;
 
 /// An individual traffic target.
 ///
-/// Equality, Ordering, and hashability is determined soley by the Endpoint's address.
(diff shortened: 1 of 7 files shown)


src/control/destination/endpoint.rs
```diff
@@ -7,7 +7,7 @@ use conditional::Conditional;
 
 /// An individual traffic target.
 ///
-/// Equality, Ordering, and hashability is determined soley by the Endpoint's address.
+/// Equality, Ordering, and hashability is determined solely by the Endpoint's address.
```
hawkw: I believe this comment is no longer the case, and should actually probably just be removed entirely --- the `Eq`/`PartialEq` implementations are derived now (so they also consider the endpoint's metadata), and the `Hash` and `Ord` implementations have been removed entirely.

PR #26, opened 2018-07-23
update to hyper 0.12.7 to fix a keep-alive bug

seanmonstar:
Specifically proxied bodies would make use of an optimization in hyper
that resulted in the connection not knowing (but it did know! just
didn't tell itself...) that the body was finished, and so the connection
was closed. 0.12.7 includes the fix in hyper.

As part of this upgrade, the keep-alive tests have been adjusted to send
a small body, since the empty body was not triggering this case.

I believe this fully fixes https://github.com/linkerd/linkerd2/issues/1353

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -340,7 +340,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "hyper"
-version = "0.12.6"
+version = "0.12.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "bytes 0.4.7 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


hawkw:
> As part of this upgrade, the keep-alive tests have been adjusted to send
a small body, since the empty body was not triggering this case

Should we have tests with an empty body _as well_, in case of future regressions that only occur when the body is empty, or does this seem too unlikely to be worth worrying about?

seanmonstar:
> Should we have tests with an empty body as well?

:shrug: We could. In the hyper PR, the test suite was updated to run those tests with the matrix of (no body, with body, no proxy, proxy).

briansmith:
> Should we have tests with an empty body as well, in case of future regressions that only occur when the body is empty, or does this seem too unlikely to be worth worrying about?

I agree that it seems very useful to test this with an empty body too. Pretty much any HTTP-level thing we do should be tested with an empty body.

It's great that there's a hyper-level test but, especially since we have previously discussed refactoring things significantly to unify the hyper and tower-h2 logic, having linkerd2-level tests seems prudent, esp. since the "extra" test was already written.

seanmonstar:
I've updated that test to run each request against two routes, one returning a small body, and one returning no body.

olix0r:
thanks for hunting this down, @seanmonstar !  I'm 👍 when @briansmith is.

PR #27, opened 2018-07-24
Separate max destination resolutions from route cache capacity

hawkw:
Currently, the proxy places a limit on the number of active routes
in the route cache. This limit defaults to 100 routes, and is intended
to prevent the proxy from requesting more than 100 lookups from the 
Destination service. 

However, in some cases, such as Prometheus scraping a large number of
pods, the proxy hits this limit even though none of those requests 
actually result in requests to service discovery (since Prometheus 
scrapes pods by their IP addresses). 

This branch implements @briansmith's suggestion in 
https://github.com/linkerd/linkerd2/issues/1322#issuecomment-407161829.
It splits the router capacity limit to two separate, configurable 
limits, one that sets an upper bound on the number of concurrently 
active destination lookups, and one that limits the capacity of the
router cache.

I've done some preliminary testing using the `lifecycle` tests, where a
single Prometheus instance is configured to scrape a very large number 
of proxies. In these tests, neither limit is reached.


hawkw:
I ran a fairly large-scale test (`linkerd-lifecycle` with >1500 pods), and the Prometheus proxy no longer hits its' router capacity. The scrape success rate isn't at 100%, but this is partially due "connection refused" and "no route to host" errors, which I believe mainly result fromthe lifecycle test creating and destroying pods during the scrape. After running all night, the Prometheus instance itself has been restarted four times and becomes fairly slow, but it no longer seems like the proxy is the limiting factor.

tests/discovery.rs
```diff
@@ -35,6 +74,7 @@ macro_rules! generate_tests {
             let client = $make_client(proxy.outbound, "disco.test.svc.cluster.local");
 
             assert_eq!(client.get("/recon"), "nect");
+
```
olix0r: a wild space appears

tests/discovery.rs
```diff
@@ -21,6 +21,45 @@ macro_rules! generate_tests {
...
+            let client = $make_client(proxy.outbound, "disco10.test.svc.cluster.local");
+            println!("trying 10th destination....");
+            let mut req = client.request_builder("/");
+            let rsp = client.request(req.method("GET"));
+            // The request should fail.
+            assert_eq!(rsp.status(), http::StatusCode::INTERNAL_SERVER_ERROR);
+
+            // TODO: The controller should also assert the proxy doesn't make requests
+            // when it has hit the maximum number of resolutions.
```
olix0r: is it possible to test that capacity is made available as routes are replaced in the cache?

hawkw: Ah, good point. Will see about that!

hawkw:
I don't think this branch is correct as-is. After seeing @olix0r's comment https://github.com/linkerd/linkerd2-proxy/pull/27#discussion_r205202508, I thought about this a bit & realized that resolutions for inactive routes are only dropped when the _routes_ that created those resolutions are evicted from the router cache, so resolution capacity won't be made available until the route cache evicts inactive routes. If we want to move forwards with this, we'd probably need some separate eviction mechanism for resolutions.

olix0r:
@hawkw 

>  resolutions for inactive routes are only dropped when the routes that created those resolutions are evicted from the router cache, so resolution capacity won't be made available until the route cache evicts inactive routes. 

this sounds like _correct_ behavior to me. can you elaborate on what you think the bug is?

briansmith:
> I don't think this branch is correct as-is. After seeing @olix0r's comment #27 (comment), I thought about this a bit & realized that resolutions for inactive routes are only dropped when the routes that created those resolutions are evicted from the router cache, so resolution capacity won't be made available until the route cache evicts inactive routes.

If the route cache isn't evicting inactive routes based on LRU then isn't that a bug? What does LRU if not the router cache?

> If we want to move forwards with this, we'd probably need some separate eviction mechanism for resolutions.

We'd need both this and that so IMO it is fine to carry on with this.

Earlier you said tested this and verified that it fixes the Prometheus problem, which is the most urgent problem that we're trying to fix now.  I don't understand how to reconcile the most recent comment with that.

hawkw:
Okay, I discussed this with @olix0r a bit further, and I don't think my concern necessarily makes this branch invalid. 

I was concerned that if the maximum number of resolutions was lower than the route capacity, we could enter into a state where we don't evict inactive routes, even if there are inactive routes to evict. In this state, we could never reach the route capacity and start evicting if we don't get any incoming requests that wouldn't go through the Destination service.

However, after discussing this with @olix0r, we determined that in most cases, we shouldn't set the route capacity higher than the maximum number of resolutions; by default, they should be the same. This change would just serve as a defensive measure against overloading the Destination service when we _do_ set a higher route capacity as a special case. 

So, the correct solution to the Prometheus problem is to merge this branch, but leave the default route capacity at 100, rather than increasing it to 10,000. Then, we can merge linkerd/linkerd2#1358, which increases the route capacity for the Prometheus proxy specifically, as a special case. This change in the proxy will just guard against a situation where proxies whose route capacity is set above the default can't overload the Destination service if they _do_ end up receiving a lot of requests that go through service discovery.

hawkw:
@briansmith 
> If the router capacity is higher and we reach the destination service capacity then any new resolutions will fail which seems like it would be the same as setting the router capacity to that value, basically the same as now. 

I don't think that's the case. If we reach the router capacity, and any of those routes are inactive, then the least recently used inactive route will be evicted, and the request will succeed. On the other hand, if we reach the destination service capacity but still have remaining router capacity, the resolution will fail, failing the request.

> BTW, my understanding is that the limit in the discovery module is supposed to be the total number of destination service queries and not the total number of resolutions, since the destination service queries are really the limiting factor, IIUC.

Ah, that's a good point, I think I'll have to make some additional changes so that DNS resolutions aren't also capped.

briansmith:
> On the other hand, if we reach the destination service capacity but still have remaining router capacity, the resolution will fail, failing the request.

(In the future) if the resolution fails because it is at capacity then that should be backpressure for the router and the router should drop a route that would ease the pressure on the destination service. (This means that the results the discovery module give to the router need to indicate whether or not dropping the subscription would help.)

hawkw:
@olix0r and @briansmith, I've made some additional changes so that the Destination query limit no longer also applies to DNS queries. I've also added a test verifying that this is the case.

Also, I've changed the default outbound router capacity back to 100 routes, as we discussed above.

src/control/destination/background.rs
```diff
@@ -69,16 +70,28 @@ type ClientService = AddOrigin<Backoff<LogErrors<Reconnect<
...
 struct Background<T: HttpService<ResponseBody = RecvBody>> {
-    dns_resolver: dns::Resolver,
-    namespaces: Namespaces,
-    destinations: HashMap<DnsNameAndPort, DestinationSet<T>>,
-    /// A queue of authorities that need to be reconnected.
-    reconnects: VecDeque<DnsNameAndPort>,
+    cfg: QueryCfg,
+    dsts: DestinationCache<T>,
+    /// A receiver of new watch requests.
+    request_rx: mpsc::UnboundedReceiver<ResolveRequest>,
```
olix0r: Please try to avoid unnecessary formatting changes like moving variables around. It makes the diff a bit more complicated than it needs to be.

hawkw: Are you referring to how a lot of the fields in this struct were split into two separate structs? In this case, that was actually a functional change --- it was necessary for this code to compile due to borrow checker reasons.

hawkw:
@olix0r I agree that this module ought to be split apart. I avoided doing so in this branch because I felt that it would likely make the diff harder to read, and was intending to do so in a follow-up. Is that okay, or would you prefer for me to go ahead and do it now?

olix0r:
@hawkw I think we should try to break this into some smaller PRs. Let's take some time tomorrow afternoon to pair on this to find some smaller easier-to-ship scopes for this change so that we can move forward with some more confidence.

hawkw:
@olix0r Okay, I have some thoughts on what parts of this change can be easily factored out. I'd be happy to pair with you, or just try to prepare some smaller branches.

PR #28, opened 2018-07-24
Add initial tests for client and server connection handling w.r.t. TLS.

briansmith:
Add a simple framework for TLS connection handling and some initial tests
that use it.

An explicit effort has been made to keep the test configuration as close to the
production configuration as possible; e.g. we use regular TCP sockets instead of some
mock TCP sockets. This matters less now, but will matter more later, if/when we
implement more low-level TLS-over-TCP optimizations.

Rename `ConnectionConfig::identity` to `ConnectionConfig::server_identity` to make
it clearer that it is always the identity of the server, regardless of which role
the `ConnectionConfig` is being used in.

Signed-off-by: Brian Smith <brian@briansmith.org>

### DIFF
--- a/src/transport/connection.rs
+++ b/src/transport/connection.rs
@@ -147,7 +147,7 @@ impl BoundPort {
             Fut: IntoFuture<Item = T, Error = std::io::Error> + Send + 'static,
             <Fut as IntoFuture>::Future: Send,
     {
-        self.listen_and_fold_(std::u64::MAX, initial, f)
+        self.listen_and_fold_inner(std::u64::MAX, initial, f)
     }
 
     #[cfg(test)]
@@ -163,10 +163,10 @@ impl BoundPort {
             Fut: IntoFuture<Item = T, Error = std::io::Error> + Send + 'static,
             <Fut as IntoFuture>::Future: Send,
     {
-        self.listen_and_fold_(connection_limit, initial, f)
+        self.listen_and_fold_inner(connection_limit, initial, f)
     }
 
-    fn listen_and_fold_<T, F, Fut>(
+    fn listen_and_fold_inner<T, F, Fut>(
         self,
         connection_limit: u64,
         initial: T,
--- a/src/transport/connection.rs
+++ b/src/transport/connection.rs
@@ -141,6 +141,37 @@ impl BoundPort {
         initial: T,
         f: F)
         -> impl Future<Item = (), Error = io::Error> + Send + 'static
+        where
+            F: Fn(T, (Connection, SocketAddr)) -> Fut + Send + 'static,
+            T: Send + 'static,
+            Fut: IntoFuture<Item = T, Error = std::io::Error> + Send + 'static,
+            <Fut as IntoFuture>::Future: Send,
+    {
+        self.listen_and_fold_(std::u64::MAX, initial, f)
+    }
+
+    #[cfg(test)]
+    pub fn listen_and_fold_n<T, F, Fut>(
+        self,
+        connection_limit: u64,
+        initial: T,
+        f: F)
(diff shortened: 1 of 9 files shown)


src/transport/connection_tests.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    // The server's connection will succeed with the TLS client hello passed
+    // through, because the SNI doesn't match its identity.
+    assert_eq!(client_result.is_tls(), false);
+    assert!(client_result.result.is_err());
+    assert_eq!(server_result.is_tls(), false);
+    assert_eq!(&server_result.result.unwrap()[..], START_OF_TLS);
+}
+
+#[allow(dead_code)]
```
olix0r: is this necessary? or a leftover from development?

src/transport/connection.rs
```diff
@@ -141,6 +141,37 @@ impl BoundPort {
         initial: T,
         f: F)
         -> impl Future<Item = (), Error = io::Error> + Send + 'static
+        where
+            F: Fn(T, (Connection, SocketAddr)) -> Fut + Send + 'static,
+            T: Send + 'static,
+            Fut: IntoFuture<Item = T, Error = std::io::Error> + Send + 'static,
+            <Fut as IntoFuture>::Future: Send,
+    {
+        self.listen_and_fold_(std::u64::MAX, initial, f)
```
hawkw: I'm slightly uncomfortable with the use of `u64::MAX` as a connection limit. I don't really expect we'll hit  2E64 - 1 connections in practice, but I think passing the limit as an `Option` might be better...

src/transport/tls/config.rs
```diff
@@ -76,14 +76,22 @@ impl std::fmt::Debug for ClientConfig {
...
+
 pub type ClientConfigWatch = Watch<Conditional<ClientConfig, ReasonForNoTls>>;
 pub type ServerConfigWatch = Watch<ServerConfig>;
 
 /// The configuration in effect for a client (`ClientConfig`) or server
 /// (`ServerConfig`) TLS connection.
 #[derive(Clone, Debug)]
 pub struct ConnectionConfig<C> where C: Clone {
-    pub identity: Identity,
+    pub server_identity: Identity,
```
hawkw: 👍 this is definitely clearer!

src/transport/tls/config.rs
```diff
@@ -506,6 +521,39 @@ mod test_util {
                 private_key: dir.join(self.private_key),
             }
         }
+
+        // Returns a `ConnectionConfig<ClientConfigWatch>` preloaded with a
+        // valid client TLS configuration.
+        pub fn client(&self, server_identity: Identity) -> (ConnectionConfig<ClientConfigWatch>)
```
hawkw: Do the parentheses here indicate that this returned inside of a 1-element tuple? Is there a reason for this?

src/transport/connection.rs
hawkw: I _think_ something like

```rust
let stream = match connection_limit {
    Some(limit) => stream::Either::A(listener.incoming().take(limit)),
    None => stream::Either::B(listener.incoming()),
};
stream.and_then(...
```
ought to work?

Though you're right that we're probably not going to hit that limit.

briansmith: > I think something like

I spent some time trying it. Maybe there is some way to get it to work but I don't think it would be better than just using `usize::MAX` as the limit considering the extra complexity (in fact i didn't get something that can even build yet). I think we should just go with the original way.

hawkw: Fair enough!

briansmith: OK, I figured out what I was doing wrong: Using the wrong `Either`. Now I have it working both ways, but I still think the original way is clearer so I'm going to keep it that way.

PR #29, opened 2018-07-25
Add TLS status to logging contexts

hawkw:
This branch adds the connection TLS status to the proxy's logging
contexts, if it is known. I added this as I found it useful while 
debugging a different issue, but figured it ought to be added.

### DIFF
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -321,7 +321,7 @@ impl fmt::Display for Server {
             write!(f, " remote={}", remote)?;
         }
         if let Some(tls_status) = self.tls_status {
-            write!(f, " tls_status={}", tls_status)?;
+            write!(f, " tls={}", tls_status)?;
         }
         write!(f, "}}")
     }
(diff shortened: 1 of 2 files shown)


src/logging.rs
```diff
@@ -292,6 +298,13 @@ impl Server {
         }
     }
 
+    pub fn with_tls(self, tls: TlsStatus) -> Self {
```
briansmith: Please name this `with_tls_status`. I think the name `with_tls` is misleading.

src/logging.rs
```diff
@@ -337,6 +353,13 @@ impl<C: fmt::Display, D: fmt::Display> Client<C, D> {
...
```
briansmith: Ditto.

(It seems wrong that we have to implement this twice.)

src/logging.rs
hawkw: Done in 710b54ac7922793eeba7939ab7ca1c03b0822f0f. You're right that the old name was potentially confusing.

src/logging.rs
hawkw: > (It seems wrong that we have to implement this twice.)

Yeah, I think it might be worth factoring out the common code between the `Client` and `Server` contexts somehow, but I feel like that runs the risk of making it more complex overall?

briansmith: > Yeah, I think it might be worth factoring out the common code between the Client and Server contexts somehow, but I feel like that runs the risk of making it more complex overall?

I think there are probably relatively simple ways of doing it, but hard to say without trying it. It doesn't need to be done in this PR, for sure.

seanmonstar:
Closing as inactive.

PR #30, opened 2018-07-26
Remove the telemetry system's event channel

olix0r:
The proxy's telemetry system is implemented with a channel: the proxy thread generates events and the control thread consumes these events to record metrics and satisfy Tap observations. This design was intended to minimize latency overhead in the data path.

However, this design leads to _substantial_ CPU overhead: the control thread's work scales with the proxy thread's work, leading to resource contention in busy, resource-limited deployments. This design also has other drawbacks in terms of allocation & makes it difficult to implement planned features like payload-aware Tapping.

This change removes the event channel so that all telemetry is recorded instantaneously in the data path, setting up for further simplifications so that, eventually, the metrics registry properly uses service lifetimes to support eviction.

This change has a potentially negative side effect: metrics scrapes obtain the same lock that the data path uses to write metrics so, if the metrics server gets heavy traffic, it can directly impact proxy latency.  These effects will be ameliorated by future changes that reduce the need for the Mutex in the proxy thread.


olix0r:
@siggy has tested this against master and observed notable reductions in CPU utilization

hawkw:
> [the prior] design was intended to minimize latency overhead in the data path.

Does this branch have a noticeable latency impact? If so, do we think it's worth trading decreased CPU utilization in the control thread for an increase in latency in the data path?

hawkw:
Also, I'm pleasantly surprised by how (relatively) small this change ended up being! :tada:

src/telemetry/sensor/mod.rs
```diff
@@ -36,27 +41,29 @@ impl Handle {
...
-            // We may want to capture timestamps here instead of on the consumer-side...  That
-            // level of precision doesn't necessarily seem worth it yet.
-
+        if let Some(inner) = self.0.as_mut() {
             let ev = mk();
             trace!("event: {:?}", ev);
 
-            if tx.lossy_send(ev).is_err() {
-                debug!("dropped event");
+            if let Ok(mut taps) = inner.taps.lock() {
```
seanmonstar: Taking this lock for every event is a little worrisome. It seems benchmarks weren't hurt, though they also don't run while a user is tapping, right?

src/telemetry/sensor/mod.rs
```diff
@@ -36,27 +41,29 @@ impl Handle {
...
+                taps.inspect(&ev);
             }
+
+            inner.metrics.record_event(&ev);
```
seanmonstar: This also results in taking a mutex lock, right?

src/telemetry/sensor/mod.rs
olix0r: the lock is acquired regardless of whether taps are active, but it will be held for the full tap inspection -- so yes, tap can impact latency. We can likely change this to use a ReadWriteLock pretty easily, and, ultimately i want to do away with events entirely so that lock acquisition occurs at most once per request

src/telemetry/sensor/mod.rs
olix0r: yes, for now. I'd like to do away with the event model entirely so that the mutex is only needed when provisioning a route and not per-event or even per-request.

src/telemetry/sensor/mod.rs
seanmonstar: Right, but since no taps are active, there's never contention for the lock in those benchmarks. Eh, it's *probably* fine.

olix0r: aiui there's only contention while the tap request is being registered, not throughout the full lifetime of the  tap observation -- basically a map insert https://github.com/linkerd/linkerd2-proxy/blob/7788f60e0ef92b51f69828be689683db22f964b9/src/control/observe.rs#L75-L87

briansmith:
> This design also has other drawbacks in terms of allocation & makes it difficult to implement planned features like payload-aware Tapping.

I'm a bit skeptical that this is going to help with implementing the full tap functionality. Please make an issue for payload-aware tap and describe how you're expecting it to be implemented.

briansmith:
> This change has a potentially negative side effect: metrics scrapes obtain the same lock that the data path uses to write metrics so, if the metrics server gets heavy traffic, it can directly impact proxy latency. These effects will be ameliorated by future changes that reduce the need for the Mutex in the proxy thread.

Here's how I expected metrics collection to work: Every I/O thread (currently just the main thread) collects metrics and stores them internally. Occasionally (e.g. every time metrics are requested, or on a regular interval) the metrics server posts a task to each I/O thread that makes a copy of that thread's metrics and sends the copy to the metrics server. Note that this can be done without any mutexes at all. Is this the design you're eventually aiming for, or something else?

olix0r:
> The metrics server is never going to get heavy traffic

When only prometheus is in the mix, yes.  In my experience, people will build some janky scripts/uis that hammer the admin ports... I don't think we should optimize for this, but we should generally aim to minimize the impact of rendering stats

olix0r:
> Here's how I expected metrics collection to work: Every I/O thread (currently just the main thread) collects metrics and stores them internally. Occasionally (e.g. every time metrics are requested, or on a regular interval) the metrics server posts a task to each I/O thread that makes a copy of that thread's metrics and sends the copy to the metrics server. Note that this can be done without any mutexes at all. Is this the design you're eventually aiming for, or something else?

This is similar to what I expect.

I expect the controller to be responsible for building services (routes, endpoints) with the needed policy instrumented in that service, so I don't think that we need to have the control thread post a _task_ to the i/o thread, rather, I think the controller can maintain weak references to the metrics structures owned by the i/o threads... when rendering metrics, the controller attempts to dereference the stats structure to render atomic counters.

There are likely some more details to be worked out here (through prototypes, probably), but I think this change is a good first step no matter the details of the outcome, as long as we agree that we want to move away from the current Event-based system

PR #32, opened 2018-07-30
Add orig-proto which uses HTTP2 between proxies

seanmonstar:
When the destination service returns a hint that an endpoint is another
proxy, eligible HTTP1 requests are translated into HTTP2 and sent over
an HTTP2 connection. The original protocol details are encoded in a
header, `l5d-orig-proto`. When a proxy receives an inbound HTTP2
request with this header, the request is translated back into its HTTP/1
representation before being passed to the internal service.

Ref https://github.com/linkerd/linkerd2/issues/1360

### DIFF
--- a/src/bind.rs
+++ b/src/bind.rs
@@ -14,7 +14,7 @@ use control;
 use control::destination::Endpoint;
 use ctx;
 use telemetry::{self, sensor};
-use transparency::{self, HttpBody, h1};
+use transparency::{self, HttpBody, h1, orig_proto};
 use transport;
 use tls;
 use ctx::transport::TlsStatus;
@@ -93,6 +93,13 @@ where
 pub enum Protocol {
     Http1 {
         host: Host,
+        /// Whether the request wants to use HTTP/1.1's Upgrade mechanism.
+        ///
+        /// Since these cannot be translated into orig-proto, it must be
+        /// tracked here so as to allow those with `is_h1_upgrade: true` to
+        /// use an explicitly HTTP/1 service, instead of one that might
+        /// utilize orig-proto.
+        is_h1_upgrade: bool,
         /// Whether or not the request URI was in absolute form.
         ///
         /// This is used to configure Hyper's behaviour at the connection
@@ -134,7 +141,7 @@ pub type Service<B> = BoundService<B>;
 
 pub type Stack<B> = WatchService<tls::ConditionalClientConfig, RebindTls<B>>;
 
-type StackInner<B> = Reconnect<NormalizeUri<NewHttp<B>>>;
+type StackInner<B> = Reconnect<orig_proto::Upgrade<NormalizeUri<NewHttp<B>>>>;
 
 pub type NewHttp<B> = sensor::NewHttp<Client<B>, B, HttpBody>;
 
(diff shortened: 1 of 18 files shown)


src/bind.rs
```diff
@@ -93,6 +93,7 @@ where
 pub enum Protocol {
     Http1 {
         host: Host,
+        is_upgrade: bool,
```
olix0r: Probably warrants a comment.  Specifically, not immediately clear if this means that we _should_ upgrade or if it was already upgraded and we should downgrade... perhaps a more descriptive name would clear this up?

seanmonstar: Sure, I changed the name to `is_h1_upgrade`, and added comments on it and also where it's used in `bind_service`.

olix0r: looks like there's a reference to the old name in the tests in inbound.rs

src/control/destination/background/destination_set.rs
```diff
@@ -302,7 +303,23 @@ fn pb_to_addr_meta(
         }
     };
 
-    let meta = Metadata::new(DstLabels::new(labels.into_iter()), tls_identity);
+    let mut proto_hint = ProtocolHint::Unknown;
+
+    if let Some(hint) = pb.protocol_hint {
+        if let Some(proto) = hint.protocol {
+            match proto {
+                Protocol::H2(..) => {
```
olix0r: tioli i think you could combine this to

```rust
if let Some(Protocol::H2(..)) = hint.protocol {
```

src/control/destination/endpoint.rs
```diff
@@ -36,6 +36,13 @@ impl Endpoint {
         self.metadata.dst_labels()
     }
 
+    pub fn can_speak_h2(&self) -> bool {
```
olix0r: This might be a slightly misleading name --- the proxy can communicate with h2 endpoints that are not other proxies (grpc services for instance).  Is this `prefers_h2()` clearer?

src/transparency/orig_proto.rs
```diff
@@ -0,0 +1,244 @@
...
+            *req.version_mut() = http::Version::HTTP_2;
+            downgrade_response = true;
+        }
+
+        let fut = self.inner.call(req);
+
+        if downgrade_response {
+            fut.map(|mut res| {
+                debug_assert_eq!(res.version(), http::Version::HTTP_2);
+                let version = if let Some(orig_proto) = res.headers().get(L5D_ORIG_PROTO) {
```
olix0r: Just to confirm... it's necessary to use a response header because the request and response may advertise different HTTP versions.

src/transparency/orig_proto.rs
```diff
@@ -0,0 +1,244 @@
...
+        let s = self.inner.new_service();
+        // This weird dance is so that the closure doesn't have to
+        // capture `self` and can just be a `fn` (so the `Map`)
+        // can be returned unboxed.
+        if self.upgrade_h1 {
+            s.map(|inner| Upgrade::new(inner, true))
+        } else {
+            s.map(|inner| Upgrade::new(inner, false))
+        }
+    }
```
olix0r: this is equivalent to the following, isn't it?
```rust
  let upgrade_h1 = self.upgrade_h1;
  s.map(|inner| Upgrade::new(inner, upgrade_h1))
}
```

tioli, seems like a slightly less weird dance to just copy the bool imo

src/control/destination/background/destination_set.rs
seanmonstar: Yea, I went back and forth, and ultimately felt being exhaustive on the `Protocol` match was a better idea, since it makes us consider if this needs updating if more hints are added.

src/transparency/orig_proto.rs
seanmonstar: That way makes the closure capture `upgrade_h1`, and so it can no longer be cast to just a `fn`, and thus is unnameable.

src/transparency/orig_proto.rs
seanmonstar: Yep exactly. The request may say its `HTTP/1.1`, but if the response comes back as `HTTP/1.0`, we should be sure to pass that along (and vice versa).

src/control/destination/endpoint.rs
seanmonstar: I've been finding it confusing to just say that it prefers or can use h2, when really, we want to know if it can use h2 *and* translate with our orig-proto headers. So, I've changed this method to `can_use_orig_proto`.

PR #34, opened 2018-08-01
Refactor `control::destination::background` into smaller modules

hawkw:
This branch is purely a refactor and should result in no functional 
changes.

The `control::destination::background` module has become quite large,
making the code difficult to read and review changes to. This branch
separates out the `DestinationSet` type and the destination service
client code into their own modules inside of `background`. Furthermore,
it rolls the `control::utils` module into the `client` submodule of
`background`, as that code is only used in the `client` module.

I think there's some additional work that can be done to make this code
clearer beyond simply splitting apart some of these large files, and I
intend to do some refactoring in additional follow-up branches.


src/control/destination/background/client.rs
```diff
@@ -1,16 +1,46 @@
...
+use tower_service::Service;
+use tower_reconnect::{Reconnect, Error as ReconnectError};
+use conditional::Conditional;
+use dns;
+use timeout::{Timeout, TimeoutError};
+use transport::{tls, HostAndPort, LookupAddressAndConnect};
+use watch_service::Rebind;
+
+/// Type of the client service stack used to make destination requests.
+pub(super) type ClientService = AddOrigin<Backoff<LogErrors<Reconnect<
```
hawkw: If this was defined as
```rust
pub(super) struct ClientService(AddOrigin<Backoff<LogErrors<Reconnect<
        tower_h2::client::Connect<
            Timeout<LookupAddressAndConnect>,
            ::logging::ContextualExecutor<
                ::logging::Client<
                    &'static str,
                    HostAndPort
                >
            >,
            BoxBody,
        >
    >>>>)
```
instead of a type alias, then the `AddOrigin`, `Backoff`, and `LogErrors` types wouldn't have to be `pub(super)`. I'm considering making that change in a subsequent branch.

src/control/destination/background/destination_set.rs
```diff
@@ -0,0 +1,352 @@
...
+use dns::{self, IpAddrListFuture};
+use telemetry::metrics::DstLabels;
+use transport::{tls, DnsNameAndPort};
+use conditional::Conditional;
+
+use super::{DestinationServiceQuery, UpdateRx};
+
+/// Holds the state of a single resolution.
+pub(super) struct DestinationSet<T: HttpService<ResponseBody = RecvBody>> {
+    pub addrs: Exists<Cache<SocketAddr, Metadata>>,
```
hawkw: I'd like to make some of the fields on `DestinationSet` not be `pub`, and have `Background` interact with them using methods with more descriptive names. Planning on making that change in a follow-up branch.

hawkw:
@olix0r 
> The deeper this module goes, the less i like the word `background` as a module name.

Strongly agree. I can't come up with anything better off the top of my head, though.

PR #35, opened 2018-08-01
refactor: Make `Background::query_destination_service_if_relevant` into a method

hawkw:
This is strictly a refactor which should make no functional changes.

Currently, the function used to construct new Destination service
queries (`Background::query_destination_service_if_relevant`) is a
function rather than a method on `Background`, although it takes as an
argument a field from `Background`. This is because in some cases, it is
called where `self.destinations` is borrowed mutably, preventing `self`
from being borrowed immutably.

Right now, this means that one additional field has to be passed
explicitly. However, in order to add the limit on active Destination
service queries, it was necessary to add two additional fields to
`Background` that have to be passed to this function. Since these
arguments should always come from fields on `Background`, it would be
preferable for this to be a method.

This branch breaks out some of the fields on
`control::destination::Background` into their own structs:
`DestinationCache`, which holds the map of DNS names to
`DestinationSet`s and the deque of DNS names that need reconnects; and
`QueryCfg`, which holds the configuration necessary to create a new
Destination service query (currently just the `Namespaces` config). This
allows us to have separate borrows on the `DestinationCache` and
`QueryCfg`. 

When I make the additional changes necessary to add the limit on active 
destination queries, the two additional fields necessary can be added to 
`QueryCfg`, rather than having to explicitly pass them into 
`query_destination_service_if_relevant` every time it's called.

### DIFF
--- a/src/control/destination/background/mod.rs
+++ b/src/control/destination/background/mod.rs
@@ -51,7 +51,7 @@ type UpdateRx<T> = Receiver<PbUpdate, T>;
 /// provided authority to a set of addresses, and ensures that resolution updates are
 /// propagated to all requesters.
 struct Background<T: HttpService<ResponseBody = RecvBody>> {
-    cfg: QueryCfg,
+    config: Config,
     dns_resolver: dns::Resolver,
     dsts: DestinationCache<T>,
     /// The Destination.Get RPC client service.
@@ -63,6 +63,7 @@ struct Background<T: HttpService<ResponseBody = RecvBody>> {
 
 /// Holds the currently active `DestinationSet`s and a list of any destinations
 /// which require reconnects.
+#[derive(Default)]
 struct DestinationCache<T: HttpService<ResponseBody = RecvBody>> {
     destinations: HashMap<DnsNameAndPort, DestinationSet<T>>,
     /// A queue of authorities that need to be reconnected.
@@ -71,7 +72,7 @@ struct DestinationCache<T: HttpService<ResponseBody = RecvBody>> {
 
 /// The configurationn necessary to create a new Destination service
 /// query.
-struct QueryCfg {
+struct Config {
     namespaces: Namespaces,
 }
 
@@ -88,8 +89,8 @@ pub(super) fn task(
     // Build up the Controller Client Stack
     let mut client = host_and_port.map(|host_and_port| {
         let (identity, watch) = match controller_tls {
-            Conditional::Some(cfg) =>
-                (Conditional::Some(cfg.server_identity), cfg.config),
+            Conditional::Some(config) =>
+                (Conditional::Some(config.server_identity), config.config),
             Conditional::None(reason) => {
                 // If there's no connection config, then construct a new
                 // `Watch` that never updates to construct the `WatchService`.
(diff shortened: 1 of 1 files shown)


hawkw:
Note that this is based on a change I originally introduced in #27, but factored out into its own branch.

src/control/destination/background/mod.rs
```diff
@@ -51,18 +51,30 @@ type UpdateRx<T> = Receiver<PbUpdate, T>;
...
+/// which require reconnects.
+struct DestinationCache<T: HttpService<ResponseBody = RecvBody>> {
+    destinations: HashMap<DnsNameAndPort, DestinationSet<T>>,
+    /// A queue of authorities that need to be reconnected.
+    reconnects: VecDeque<DnsNameAndPort>,
+}
+
+/// The configurationn necessary to create a new Destination service
+/// query.
+struct QueryCfg {
```
hawkw: I'm not particularly fond of the names "DestinationCache" and "QueryCfg" --- these are simply the best names I could think of that weren't already being used for something else in this module. 

If anyone can think of more descriptive names, I'd welcome suggestions!

src/control/destination/background/mod.rs
```diff
@@ -333,4 +347,33 @@ where
...
+}
+
+// ===== impl DestinationCache =====
+
+impl<T> DestinationCache<T>
+where
+    T: HttpService<RequestBody = BoxBody, ResponseBody = RecvBody>,
+    T::Error: fmt::Debug,
+{
+    fn new() -> Self {
```
olix0r: tioli derive default; no need for this constructor, really?

src/control/destination/background/mod.rs
```diff
@@ -51,18 +51,30 @@ type UpdateRx<T> = Receiver<PbUpdate, T>;
 /// provided authority to a set of addresses, and ensures that resolution updates are
 /// propagated to all requesters.
 struct Background<T: HttpService<ResponseBody = RecvBody>> {
+    cfg: QueryCfg,
```
olix0r: nit: What do you think about making this `config: Config`. There isn't really any ambiguity within this file, and we get to avoid _cfg_ (which imo isn't a significant win over _config_, whereas _dst_ is big win over _destination_)

hawkw: That seems reasonable to me.

I added the word `Query` because I thought it was worth indicating that this is used to construct destination service queries. I had initially considered `QueryBuilder`, but that made it sound too much like the builder pattern...

src/control/destination/background/mod.rs
hawkw: If I recall correctly, I originally tried to derive `Default` for this, but can't because `T` doesn't implement `Default`. It shouldn't be _necessary_ for it to, since the default HashMap should be empty, but I think that's still required to derive it.

hawkw: Huh, nevermind, I'm wrong. Looks like it can derive `Default`. Will change that, thanks!

olix0r: you could manually impl Default but it's probably not worth it. sgtm.

hawkw: Yeah, I thought this was better than adding another `impl` block + import just to get `Default`, since it isn't a member of anything that itself requires `Default`...

PR #36, opened 2018-08-01
Limit concurrent Destination service queries

hawkw:
Required for linkerd/linkerd2#1322.

Currently, the proxy places a limit on the number of active routes
in the route cache. This limit defaults to 100 routes, and is intended
to prevent the proxy from requesting more than 100 lookups from the 
Destination service. 

However, in some cases, such as Prometheus scraping a large number of
pods, the proxy hits this limit even though none of those requests 
actually result in requests to service discovery (since Prometheus 
scrapes pods by their IP addresses). 

This branch implements @briansmith's suggestion in 
https://github.com/linkerd/linkerd2/issues/1322#issuecomment-407161829.
It splits the router capacity limit to two separate, configurable 
limits, one that sets an upper bound on the number of concurrently 
active destination lookups, and one that limits the capacity of the
router cache.

I've done some preliminary testing using the `lifecycle` tests, where a
single Prometheus instance is configured to scrape a very large number 
of proxies. In these tests, neither limit is reached. Furthermore, I've added
integration tests in `tests/discovery` to exercise the destination service 
query limit. These tests ensure that query capacity is released when inactive
routes which create queries are evicted from the router cache, and that the
limit does _not_ effect DNS queries.

This branch obsoletes and closes #27, which contained an earlier version of
these changes.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -54,6 +54,10 @@ pub struct Config {
 
     pub outbound_router_max_idle_age: Duration,
 
+    /// The maximum number of queries to the Destination service which may be
+    /// active concurrently.
+    pub max_dst_queries: usize,
+
     pub tls_settings: Conditional<tls::CommonSettings, tls::ReasonForNoTls>,
 
     /// The path to "/etc/resolv.conf"
@@ -179,6 +183,14 @@ pub const ENV_OUTBOUND_ROUTER_CAPACITY: &str = "LINKERD2_PROXY_OUTBOUND_ROUTER_C
 pub const ENV_INBOUND_ROUTER_MAX_IDLE_AGE: &str = "LINKERD2_PROXY_INBOUND_ROUTER_MAX_IDLE_AGE";
 pub const ENV_OUTBOUND_ROUTER_MAX_IDLE_AGE: &str = "LINKERD2_PROXY_OUTBOUND_ROUTER_MAX_IDLE_AGE";
 
+
+/// Limits the maximum number of outbound Destination service queries.
+///
+/// Routes which do not result in service discovery lookups will not be capped
+/// by this limit. This will have no effect if it is greater than the total
+/// router capacity (as configured by `ENV_OUTBOUND_ROUTER_CAPACITY`).
+pub const ENV_MAX_DESTINATION_QUERIES: &str = "LINKERD2_PROXY_MAX_DESTINATION_QUERIES";
+
 // These *disable* our protocol detection for connections whose SO_ORIGINAL_DST
 // has a port in the provided list.
 pub const ENV_INBOUND_PORTS_DISABLE_PROTOCOL_DETECTION: &str = "LINKERD2_PROXY_INBOUND_PORTS_DISABLE_PROTOCOL_DETECTION";
@@ -227,6 +239,8 @@ const DEFAULT_OUTBOUND_ROUTER_CAPACITY: usize = 100;
 const DEFAULT_INBOUND_ROUTER_MAX_IDLE_AGE:  Duration = Duration::from_secs(60);
 const DEFAULT_OUTBOUND_ROUTER_MAX_IDLE_AGE: Duration = Duration::from_secs(60);
 
+const DEFAULT_MAX_DESTINATION_QUERIES: usize = 100;
+
 // By default, we keep a list of known assigned ports of server-first protocols.
 //
 // https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.txt
@@ -271,6 +285,7 @@ impl<'a> TryFrom<&'a Strings> for Config {
         let outbound_router_capacity = parse(strings, ENV_OUTBOUND_ROUTER_CAPACITY, parse_number);
         let inbound_router_max_idle_age = parse(strings, ENV_INBOUND_ROUTER_MAX_IDLE_AGE, parse_duration);
         let outbound_router_max_idle_age = parse(strings, ENV_OUTBOUND_ROUTER_MAX_IDLE_AGE, parse_duration);
+        let max_dst_queries = parse(strings, ENV_MAX_DESTINATION_QUERIES, parse_number);
         let tls_trust_anchors = parse(strings, ENV_TLS_TRUST_ANCHORS, parse_path);
         let tls_end_entity_cert = parse(strings, ENV_TLS_CERT, parse_path);
         let tls_private_key = parse(strings, ENV_TLS_PRIVATE_KEY, parse_path);
@@ -405,6 +420,9 @@ impl<'a> TryFrom<&'a Strings> for Config {
                 .unwrap_or(DEFAULT_INBOUND_ROUTER_MAX_IDLE_AGE),
             outbound_router_max_idle_age: outbound_router_max_idle_age?
                 .unwrap_or(DEFAULT_OUTBOUND_ROUTER_MAX_IDLE_AGE),
+
+            max_dst_queries: max_dst_queries?
+                .unwrap_or(DEFAULT_MAX_DESTINATION_QUERIES),
 
             tls_settings,
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -56,7 +56,7 @@ pub struct Config {
 
     /// The maximum number of queries to the Destination service which may be
     /// active concurrently.
-    pub max_dst_queries: usize,
+    pub destination_concurrency_limit: usize,
 
     pub tls_settings: Conditional<tls::CommonSettings, tls::ReasonForNoTls>,
 
@@ -189,7 +189,8 @@ pub const ENV_OUTBOUND_ROUTER_MAX_IDLE_AGE: &str = "LINKERD2_PROXY_OUTBOUND_ROUT
 /// Routes which do not result in service discovery lookups will not be capped
 /// by this limit. This will have no effect if it is greater than the total
 /// router capacity (as configured by `ENV_OUTBOUND_ROUTER_CAPACITY`).
-pub const ENV_MAX_DESTINATION_QUERIES: &str = "LINKERD2_PROXY_MAX_DESTINATION_QUERIES";
+pub const ENV_DESTINATION_CLIENT_CONCURRENCY_LIMIT: &str =
+    "LINKERD2_PROXY_DESTINATION_CLIENT_CONCURRENCY_LIMIT";
 
 // These *disable* our protocol detection for connections whose SO_ORIGINAL_DST
 // has a port in the provided list.
@@ -239,7 +240,7 @@ const DEFAULT_OUTBOUND_ROUTER_CAPACITY: usize = 100;
 const DEFAULT_INBOUND_ROUTER_MAX_IDLE_AGE:  Duration = Duration::from_secs(60);
 const DEFAULT_OUTBOUND_ROUTER_MAX_IDLE_AGE: Duration = Duration::from_secs(60);
 
-const DEFAULT_MAX_DESTINATION_QUERIES: usize = 100;
+const DEFAULT_DESTINATION_CLIENT_CONCURRENCY_LIMIT: usize = 100;
 
 // By default, we keep a list of known assigned ports of server-first protocols.
 //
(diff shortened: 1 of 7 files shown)


src/control/destination/background/mod.rs
```diff
@@ -317,15 +353,42 @@ where
...
+    ///   for querying the Destination service and there is sufficient capacity
+    ///   to initiate a new query.
     fn query_destination_service_if_relevant<T>(
         &self,
-        client: &mut T,
+        client: Option<&mut T>,
         auth: &DnsNameAndPort,
         connect_or_reconnect: &str,
-    ) -> Option<DestinationServiceQuery<T>>
+    ) -> (Option<DestinationServiceQuery<T>>, bool)
```
hawkw: One obvious criticism is that it seems more idiomatic to return a `Result<Option<DestinationServiceQuery<T>>, TooManyErrors>` here. That was my original feeling as well. 

However, when I did so, it ended up being fairly repetitive, as every time this function was called, we had to unpack the result into an `Option` and a `bool` _anyway_. Just having this return `Option<...>, bool)` obviates the need for repeating that code.

src/config.rs
```diff
@@ -54,6 +54,10 @@ pub struct Config {
 
     pub outbound_router_max_idle_age: Duration,
 
+    /// The maximum number of queries to the Destination service which may be
+    /// active concurrently.
+    pub max_dst_queries: usize,
```
olix0r: WDYT about naming this `destination_client_concurrency_limit`. I think this leaves less room for interpretation

src/config.rs
```diff
@@ -179,6 +183,14 @@ pub const ENV_OUTBOUND_ROUTER_CAPACITY: &str = "LINKERD2_PROXY_OUTBOUND_ROUTER_C
 pub const ENV_INBOUND_ROUTER_MAX_IDLE_AGE: &str = "LINKERD2_PROXY_INBOUND_ROUTER_MAX_IDLE_AGE";
 pub const ENV_OUTBOUND_ROUTER_MAX_IDLE_AGE: &str = "LINKERD2_PROXY_OUTBOUND_ROUTER_MAX_IDLE_AGE";
 
+
+/// Limits the maximum number of outbound Destination service queries.
+///
+/// Routes which do not result in service discovery lookups will not be capped
+/// by this limit. This will have no effect if it is greater than the total
+/// router capacity (as configured by `ENV_OUTBOUND_ROUTER_CAPACITY`).
+pub const ENV_MAX_DESTINATION_QUERIES: &str = "LINKERD2_PROXY_MAX_DESTINATION_QUERIES";
```
olix0r: Similarly, _DESTINATION_CONCURRENCY_LIMIT

src/control/destination/background/mod.rs
```diff
@@ -194,12 +198,20 @@ where
                     let config = &self.config;
                     let dsts = &mut self.dsts;
 
+                    // If the requested authority currently needs more
+                    // query capacity to query the destination service, go
+                    // ahead and try to free up capacity (by dropping any
+                    // inactive DestinationSets).
+                    if dsts.needs_query_for(&resolve.authority) {
```
hawkw: Logically, it seems like we should do this only inside of the pattern match, after we know the entry for the given authority is occupied. However, `dsts.destinations.entry(...)` borrows `dsts` mutably, so we can't call `dsts.retain_active()` there. So, we have to do this here, instead.

The overhead of just `retain_active`-ing every time is hopefully reduced by pre-checking that it would actually be _necessary_.

src/control/destination/background/mod.rs
```diff
@@ -74,6 +75,8 @@ struct DestinationCache<T: HttpService<ResponseBody = RecvBody>> {
 /// query.
 struct Config {
     namespaces: Namespaces,
+    active_queries: Arc<()>,
+    max_queries: usize,
```
olix0r: similarly, max_queries makes it sound like it could be the number of queries before which the client must be torn down, or something else. How about `concurrency_limit`?

src/config.rs
hawkw: Yeah, that sounds good to me. I'm not terribly attached to the current name; picked it mainly because i wanted _something_ better than the _original_ working name, `OUTBOUND_MAX_RESOLUTIONS`, which was obviously wrong

hawkw:
@olix0r changed the names as requested. :)

src/control/destination/background/destination_set.rs
```diff
@@ -35,6 +35,13 @@ pub(super) struct DestinationSet<T: HttpService<ResponseBody = RecvBody>> {
     pub query: Option<DestinationServiceQuery<T>>,
     pub dns_query: Option<IpAddrListFuture>,
     pub responders: Vec<Responder>,
+
+    /// This flag is set if `query` is `None` because we wanted to make a
+    /// query for this destination, but were at the query capacity limit.
+    ///
+    /// Otherwise, if `query` is `None` and this is _not_ set, then the
+    /// authority does not require a Destination query.
```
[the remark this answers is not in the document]
olix0r: This has the added benefit of simplifying return types in NewQuery

src/control/destination/background/mod.rs
```diff
@@ -74,6 +75,8 @@ struct DestinationCache<T: HttpService<ResponseBody = RecvBody>> {
 /// query.
 struct Config {
     namespaces: Namespaces,
+    active_queries: Arc<()>,
```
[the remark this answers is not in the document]
hawkw: Yeah, this is why I wasn't ever really a big fan of `Config`...`NewQuery` seems fine to me; I was also considering `Queryer`, but I'm not sure if that's even a _word_...

src/control/destination/background/destination_set.rs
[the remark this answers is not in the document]
hawkw: Yeah, that might be better. I think there's at least one case where we `take()` the `query` field, so that would need to be rewritten, but I definitely like your suggestion more than the `Option` + `bool` situation...

src/control/destination/background/mod.rs
```diff
@@ -194,12 +198,20 @@ where
...
+                    // ahead and try to free up capacity (by dropping any
+                    // inactive DestinationSets).
+                    if dsts.needs_query_for(&resolve.authority) {
+                        trace!("--> no query capacity, try retain_active...", );
+                        dsts.retain_active();
+                    };
+
                     match dsts.destinations.entry(resolve.authority) {
                         Entry::Occupied(mut occ) => {
-                            let set = occ.get_mut();
```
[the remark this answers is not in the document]
hawkw: This was necessary because of the borrow checker. `occ.get_mut()` creates a mutable borrow on `occ`, and because it's assigned to the variable `set`, that borrow lasts for the whole match arm. This was previously fine, but now we also need to get the `key` from the entry to pass to `query_destination_service_if_relevant`, and we can't call `occ.key()` while it's already borrowed by `get_mut()`:
```
error[E0502]: cannot borrow `occ` as immutable because it is also borrowed as mutable
   --> src/control/destination/background/mod.rs:233:41
    |
212 |                             let set = occ.get_mut();
    |                                       --- mutable borrow occurs here
...
233 |                                         occ.key(),
    |                                         ^^^ immutable borrow occurs here
...
241 |                         },
    |                         - mutable borrow ends here

```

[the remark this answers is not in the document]
hawkw: An alternative to this would be something like this:
```rust
Entry::Occupied(mut occ) => {
    let authority = occ.key().clone();
    let set = occ.get_mut()
    // ...
```
However, _always_ cloning the authority regardless of whether or not we'll need to pass it to `query_destination_service_if_relevant` seems...less than ideal.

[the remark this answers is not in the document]
olix0r: cool, thanks

src/control/destination/background/destination_set.rs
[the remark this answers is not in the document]
olix0r: could also do something along the lines of

```rust
enum LimitedCapacity<T> { Available(T), Unavailable }
```

and then return `LimitedCapacity<Option<...>>`.  not sure if that's better or worse

but, i think encoding this in the type system one way or the other will be easier to think about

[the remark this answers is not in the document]
hawkw: You were right, the enum did in fact end up being the right way to represent this. Changed in e7f79dc066a8d87404966ac044a23196cf4273fe.

src/control/destination/background/mod.rs
```diff
@@ -484,3 +484,36 @@ where
...
+impl<T: HttpService<ResponseBody = RecvBody>> DestinationServiceQuery<T> {
+
+    pub fn is_active(&self) -> bool {
+        match self {
+            DestinationServiceQuery::Active(_) => true,
+            _ => false,
+        }
+    }
+
+    pub fn wants_query(&self) -> bool {
```
olix0r: I think this would be slightly clearer if it referred to capacity explicitly.  "wants query" makes me wonder why it wants a query, or has any desires at all.  `needs_query_capacity()` ?

src/control/destination/background/mod.rs
```diff
@@ -208,16 +233,30 @@ where
                                 },
                                 Exists::No | Exists::Unknown => (),
                             }
-                            set.responders.push(resolve.responder);
+
+                            // If the destinationSet was previously unable to
+                            // make a query due to insufficient capacity, try
+                            // to open a new query if possible.
+                            if occ.get().wants_query() {
```
olix0r: for instance, i think we could basically remove this comment by renaming the fn to be more descriptive

src/control/destination/background/mod.rs
```diff
@@ -314,18 +357,45 @@ where
...
+    ///
+    /// - `(None, true)` if the given authority _should_ query the
+    ///   Destination service, but the query limit has already been reached.
+    ///   In this case, this function should be called again when there is
+    ///   capacity for additional queries.
+    /// - `(None, false)` if the authority is not suitable for querying the
+    ///   Destination service.
+    /// - `(Some(DestinationServiceQuery), true)` if the authority is suitable
+    ///   for querying the Destination service and there is sufficient capacity
+    ///   to initiate a new query.
```
olix0r: docs are wrong now

src/control/destination/background/mod.rs
```diff
@@ -369,6 +464,16 @@ where
         }
     }
 
+    /// Returns true if `auth` is currently known to need a Destination
+    /// service query, but was unable to query previously due to the query
+    /// limit being reached.
+    fn needs_query_for(&self, auth: &DnsNameAndPort) -> bool {
+        self.destinations
+            .get(auth)
+            .map(DestinationSet::wants_query)
```
olix0r: similarly: too clever for me

src/control/remote_stream.rs
```diff
@@ -25,7 +28,10 @@ pub enum Remote<M, S: HttpService> {
 /// Streaming gRPC endpoints return a `ResponseFuture` whose item is a `Response<Stream>`.
 /// A `Receiver` holds the state of that RPC call, exposing a `Stream` that drives both
 /// the gRPC response and its streaming body.
-pub struct Receiver<M, S: HttpService>(Rx<M, S>);
+pub struct Receiver<M, S: HttpService> {
+    rx: Rx<M, S>,
+    _active: Weak<()>,
```
olix0r: can you add a comment about how this is used by this type

src/control/destination/background/mod.rs
```diff
@@ -470,7 +470,7 @@ where
     fn needs_query_for(&self, auth: &DnsNameAndPort) -> bool {
         self.destinations
             .get(auth)
-            .map(|dst| dst.wants_query)
+            .map(DestinationSet::wants_query)
```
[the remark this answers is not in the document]
hawkw: `wants_query` is now a function, so this line had to be changed anyway, but I can replace the point-free invocation if you prefer. Note that we use this syntax elsewhere in the codebase, so if this is considered bad style, this ought to be documented.

hawkw:
@olix0r I believe 238575d makes all the changes you requested?

PR #39, opened 2018-08-02
rfc: Granularize locks on metrics registry

olix0r:
I'd like some feedback on this branch in general. It may be helpful to review each commit individually.

This is an incremental step towards the ultimate goal of removing the `Event` type entirely. In this eventual world, the registry is likely to hold weak references, while the instrumented sensors include the actual metrics types.  More details need to be fleshed out here.

This change renames `metrics::Record` to `metrics::Registry` and, more importantly, changes the memory layout: instead of `Record` holding an `Arc<Mutex<Root>>`, locking has been moved within `Root` around each scope.

I've run the `record` benchmark and can't observe any noticeable change.

There's definitely more to figure out with regard to my ultimate goals, so I want some input on whether this change is Good in isolation, or whether we should hold off on merging until there's more motivation for this.


hawkw:
> I've run the record benchmark and can't observe any noticeable change.

I'd be curious to see if there's any observable latency or CPU utilization differences in the proxy-challenge test results....

src/telemetry/metrics/mod.rs
```diff
@@ -209,58 +211,58 @@ impl Root {
...
 }
 
 impl fmt::Display for Root {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        self.requests.fmt(f)?;
-        self.responses.fmt(f)?;
-        self.transports.fmt(f)?;
-        self.transport_closes.fmt(f)?;
-        self.tls_config.fmt(f)?;
+        self.requests.lock().expect("lock").fmt(f)?;
```
hawkw: I wonder if the `self.$THING.lock().expect("lock")` could be made less repetitive...maybe a macro?

src/telemetry/metrics/mod.rs
```diff
@@ -209,58 +211,58 @@ impl Root {
         }
     }
 
-    fn request(&mut self, labels: RequestLabels) -> &mut http::RequestMetrics {
-        self.requests.scopes.entry(labels)
+    fn request<F: FnOnce(&mut http::RequestMetrics)>(&self, labels: RequestLabels, f: F) {
+        f(self.requests.lock().expect("lock").scopes.entry(labels)
```
hawkw: Rather than taking a closure, I think these functions could alternatively just return a `MutexGuard<WhateverMetrics>`? I'm not sure if this would be significantly better or not.

src/telemetry/metrics/mod.rs
```diff
@@ -200,7 +200,9 @@ impl Root {
 
         let process_metrics = process::Sensor::new()
             .map_err(|e| info!("{}", e))
-            .ok();
+            .ok()
+            .map(Mutex::new)
+            .map(Arc::new);
```
hawkw: I thought point-free syntax was [considered "too clever"](https://github.com/linkerd/linkerd2-proxy/pull/36#discussion_r207399016)? 😛

src/telemetry/metrics/mod.rs
olix0r: As this is incremental, I haven't been very focused on polishing any of this, as it will all change in followups

src/telemetry/metrics/mod.rs
olix0r: in that case, we were using it on something that is effectively a method.  it's weird for me to see `.map(MyType::method)` as it seems to indicate that we're passing something to it, and while we are, technically, that's not really how i think about structures/calling things.  In this case, neither Mutex nor Arc take a `&self`, so I view it differently... the cleverness is al about the `&self` being passable to the thing

hawkw: Okay, that's reasonable!

src/telemetry/metrics/mod.rs
hawkw: Yeah, I figured. Definitely not a blocker.

src/telemetry/metrics/mod.rs
```diff
@@ -106,17 +106,17 @@ struct Metric<'a, M: FmtMetric> {
...
 
 /// The root scope for all runtime metrics.
-#[derive(Debug, Default)]
+#[derive(Clone, Debug, Default)]
 struct Root {
-    requests: http::RequestScopes,
-    responses: http::ResponseScopes,
-    transports: transport::OpenScopes,
-    transport_closes: transport::CloseScopes,
+    requests: Arc<Mutex<http::RequestScopes>>,
```
seanmonstar: I'm still reviewing, but is there a use case for having an `Arc` per field, instead of `Arc<Root>`?

olix0r: probably not in this state. I'm trying to drive the Arcs down to the leaves so that the sensor actually holds the Arc<Scope> and the registry holds a Weak reference to it (so the registry doesnt need to attempt to manage metric lifetimes itself)

PR #41, opened 2018-08-06
Fully encapsulate process metrics in `mod process`

olix0r:
The `process` module exposes a `Sensor` type that is different from
other types called `Sensor`. Most `Sensor` types instrument other
types with telemetry. The `process::Sensor` type, on the other hand,
is used to read system metrics from the `/proc` filesystem, returning
a metrics summary.

Furthermore, `telemetry::metrics::Root` owns the process start time
metric.

In the interest of making the telemetry system more modular, this moves
all process-related telemetry concerns into the `process` module.
Instead of exposing a `Sensor` that produces metrics, a single public
`Process` type implements `fmt::Display` directly.

This removes process-related concerns from `telemetry/metrics/mod.rs` to
setup further refactoring along these lines.

### DIFF
--- a/src/telemetry/metrics/process.rs
+++ b/src/telemetry/metrics/process.rs
@@ -1,65 +1,52 @@
 use std::fmt;
-use super::{Counter, Gauge, Metric};
+use std::time::UNIX_EPOCH;
 
-pub use self::imp::Sensor;
+use ctx;
+use super::{Gauge, Metric};
 
-#[derive(Copy, Clone)]
-pub struct ProcessMetrics {
-    cpu_seconds_total: Counter,
-    open_fds: Gauge,
-    max_fds: Option<Gauge>,
-    virtual_memory_bytes: Gauge,
-    resident_memory_bytes: Gauge,
+use self::system::System;
+
+#[derive(Debug, Default)]
+pub struct Process {
+    start_time: Gauge,
+    system: Option<System>,
 }
 
-impl ProcessMetrics {
+impl Process {
     metrics! {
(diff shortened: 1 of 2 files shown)


src/telemetry/metrics/process.rs
```diff
@@ -69,91 +56,128 @@ mod imp {
...
-            if !f?.file_type()?.is_dir() {
-                open += 1;
+            let clock_ticks = stat.utime as u64 + stat.stime as u64;
+            Self::process_cpu_seconds_total.fmt_help(f)?;
+            Self::process_cpu_seconds_total.fmt_metric(
+                f,
+                Counter::from(clock_ticks / self.clock_ticks_per_sec),
+            )?;
+
+            if let Ok(open_fds) = Self::open_fds(stat.pid) {
```
seanmonstar: Is the error case not a big deal, and this it's ignored? Or should it be logged?

src/telemetry/metrics/process.rs
```diff
@@ -69,91 +63,141 @@ mod imp {
...
+                    error!("error getting {}: {:?}", name, error);
+                    Err(error)
+                },
+                val => Ok(val as u64),
+            }
         }
     }
 
+    impl fmt::Display for System {
+        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
hawkw: Eventually, I think it would be good to sample these metrics on a fixed duration and store their values, rather than always computing them on-demand. Of course, the previous implementation _also_ did this, so I'm fine with this change.

PR #42, opened 2018-08-06
fix orig-proto translation with HTTP1 chunked bodies

seanmonstar:
The `transfer-encoding` header was not being removed before translating to HTTP2, which was causing protocol errors. This removes them. hyper will automatically add them when translating back to HTTP/1.1 (and won't add if it was HTTP/1.0). Tests added to ensure this behavior.

Closes https://github.com/linkerd/linkerd2/issues/1404

### DIFF
--- a/src/transparency/orig_proto.rs
+++ b/src/transparency/orig_proto.rs
@@ -1,6 +1,6 @@
 use futures::{future, Future, Poll};
 use http;
-use http::header::HeaderValue;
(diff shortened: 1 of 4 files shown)


seanmonstar:
As mentioned, hyper will automatically add the chunked encoding header if there is a body but no header and it's HTTP/1.1. It also knows to not do so for HTTP/1.0. We have unit tests for these cases already.

olix0r:
@seanmonstar thanks for confirming

PR #43, opened 2018-08-06
strip other connection-level headers from HTTP/1

seanmonstar:
Closes https://github.com/linkerd/linkerd2/issues/1407

### DIFF
--- a/src/transparency/h1.rs
+++ b/src/transparency/h1.rs
@@ -4,7 +4,7 @@ use std::sync::Arc;
 
 use bytes::BytesMut;
 use http;
-use http::header::{HOST, UPGRADE};
+use http::header::{CONNECTION, HOST, UPGRADE};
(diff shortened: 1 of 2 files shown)


src/transparency/h1.rs
```diff
@@ -86,26 +86,26 @@ fn set_authority(uri: &mut http::Uri, auth: Authority) {
...
-
-    // A `Connection` header may have a comma-separated list of
-    // names of other headers that are meant for only this specific connection.
-    //
-    // Iterate these names and remove them as headers.
-    for name in conn_header.split(',') {
-        let name = name.trim();
-        headers.remove(name);
+    if let Some(val) = headers.remove(CONNECTION) {
+        if let Ok(conn_header) = val.to_str() {
```
hawkw: This is much clearer than the previous implementation, IMO 👍

PR #44, opened 2018-08-06
refactor: Make `DestinationSet` aware of its own authority

hawkw:
This branch makes some changes to the `control::destination::background`
module in order to (hopefully) make this code more understandable. It 
should not cause a functional change.

The primary change here is to make each `DestinationSet` aware of its
own authority, so that code which was previously in methods on the
`Background` struct can be moved to methods on `DestinationSet` which
`Background` invokes. This allows more of `DestinationSet`'s fields to
be made private, and puts a bit more of an API boundary between 
`Background` and `DestinationSet`.

Since the authorities are used as keys in the map of `DestinationSet`s,
and are also sometimes put in the reconnect queue, they are now 
reference counted, so that they can be shared with the `DestinationSet`
without having to clone the string representing those authorities. The
reconnect queue can hold a `Weak` reference, since once the 
`DestinationSet` has been dropped (by `DestinationCache::retain_active`),
the authority is no longer needed and can also be dropped.

In addition, I've added a few more comments on some methods and types.

### DIFF
--- a/src/control/destination/background/destination_set.rs
+++ b/src/control/destination/background/destination_set.rs
@@ -33,8 +33,8 @@ use super::{ActiveQuery, DestinationServiceQuery, UpdateRx};
 pub(super) struct DestinationSet<T: HttpService<ResponseBody = RecvBody>> {
     pub addrs: Exists<Cache<SocketAddr, Metadata>>,
     pub query: DestinationServiceQuery<T>,
-    pub dns_query: Option<IpAddrListFuture>,
-    pub responders: Vec<Responder>,
+    dns_query: Option<IpAddrListFuture>,
+    responders: Vec<Responder>,
 }
 
 // ===== impl DestinationSet =====
@@ -44,6 +44,18 @@ where
     T: HttpService<RequestBody = BoxBody, ResponseBody = RecvBody>,
     T::Error: fmt::Debug,
 {
+    pub(super) fn new(
+        responder: Responder,
+        query: DestinationServiceQuery<T>,
+    ) -> Self {
+        Self {
+            addrs: Exists::Unknown,
+            query,
+            dns_query: None,
+            responders: vec![responder],
+        }
+    }
+
     pub(super) fn reset_dns_query(
         &mut self,
         dns_resolver: &dns::Resolver,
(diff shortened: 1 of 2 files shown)


hawkw:
@seanmonstar 
> It does mean though that this will likely mean more allocations...

Yeah...I wondered about moving the `Arc` all the way down to `tls::DnsName` so we could reuse the allocation from the underlying `String`, but I wasn't sure if that was worth the effort or not.

src/control/destination/background/mod.rs
```diff
@@ -67,9 +67,9 @@ struct Background<T: HttpService<ResponseBody = RecvBody>> {
 /// which require reconnects.
 #[derive(Default)]
 struct DestinationCache<T: HttpService<ResponseBody = RecvBody>> {
-    destinations: HashMap<DnsNameAndPort, DestinationSet<T>>,
+    destinations: HashMap<Arc<DnsNameAndPort>, DestinationSet<T>>,
     /// A queue of authorities that need to be reconnected.
-    reconnects: VecDeque<DnsNameAndPort>,
+    reconnects: VecDeque<Weak<DnsNameAndPort>>,
```
olix0r: Is it really necessary to make this change -- putting the names on the heap -- to solve the problem of passing the names around?

src/control/destination/background/mod.rs
```diff
@@ -288,16 +287,17 @@ where
...
-                        "reconnect",
-                    );
-                return !set.needs_query_capacity();
+        let dsts = &mut self.dsts;
+        while let Some(auth) = dsts.reconnects.pop_front() {
+            if let Some(set) = auth
+                // If the `DestinationSet` corresponding to this authority
+                // has already been dropped, the `Weak` reference to it in
+                // `reconnects` no longer points to a present value. That's
+                // fine, as we no longer need to reconnect that authority.
```
olix0r: nit: prefer to avoid having comments within a statement -- it's jarring.  Also, I think this comment could be pithier:

```
// Reconnect iff the referenced authority still exists in `Destinations`
```

src/control/destination/background/destination_set.rs
```diff
@@ -31,10 +32,11 @@ use super::{ActiveQuery, DestinationServiceQuery, UpdateRx};
 
 /// Holds the state of a single resolution.
 pub(super) struct DestinationSet<T: HttpService<ResponseBody = RecvBody>> {
+    auth: Arc<DnsNameAndPort>,
```
olix0r: So, the existence of the destinationset can hold the reference that reconnects holds active, even when it's not in the destinations map?  I'd much prefer to avoid using shared pointers in this review if at all possible. There are subtleties like this that lead to complexity

seanmonstar:
I agree that the introduction of Arcs adds a different complexity, and makes an extra allocation (and indirection) per destination, and at first I didn't like it. While I was writing up that comment, I did feel like it had at least cleaned up the noise of passing the authority as an argument to all methods, and started to wonder if that made up for it. Reading it again with fresh eyes, it does seem the arcs don't outweigh the reduced argument complexity.

hawkw:
I'm not particularly attached to this change, so, if the consensus is that no longer passing authorities around is not worth adding `Arc`s, I'd be fine with closing this branch. 

That said, I'm not sure if there's any practical way to avoid having to pass the authorities around _without_  using some form of shared pointer. We could move the authorities to the `DestinationSet`s only,and stop storing them in a hash map, but having to do a linear search through a vector of `DestinationSet`s seems like it would have worse performance implications...

olix0r:
@hawkw i think my base question is: if we just do this refactor and clone the hostnames for now, is that so bad?  I think we should either try to change the Arc'ing first, or try to change the factoring, but not both at the same time, unless it's really required.

PR #46, opened 2018-08-08
Share a common TlsConfigReload metrics object

olix0r:
The current implementation of TLS config reload telemetry has several
layers: a sensor emits events into a dispatcher that updates a metrics
object.

This can be simplified by sharing a common TlsConfigReload struct
directly from the metrics registry into the config module.

### DIFF
--- a/src/telemetry/metrics/mod.rs
+++ b/src/telemetry/metrics/mod.rs
@@ -111,7 +111,7 @@ struct Root {
     responses: http::ResponseScopes,
     transports: transport::OpenScopes,
     transport_closes: transport::CloseScopes,
-    tls_config_reload: TlsConfigReload,
+    tls_config_reload: Arc<TlsConfigReload>,
     process: process::Process,
 }
 
@@ -136,9 +136,14 @@ struct Stamped<T> {
 /// is a Hyper service which can be used to create the server for the
 /// scrape endpoint, while the `Record` side can receive updates to the
 /// metrics by calling `record_event`.
-pub fn new(process: &Arc<ctx::Process>, idle_retain: Duration) -> (Record, Serve){
-    let metrics = Arc::new(Mutex::new(Root::new(process)));
-    (Record::new(&metrics), Serve::new(&metrics, idle_retain))
+pub fn new(process: &Arc<ctx::Process>, idle_retain: Duration)
+    -> (Record, Arc<TlsConfigReload>, Serve)
+{
+    let root = Root::new(process);
+    let tls_config_reload = root.tls_config_reload().clone();
+
+    let metrics = Arc::new(Mutex::new(root));
+    (Record::new(&metrics), tls_config_reload, Serve::new(&metrics, idle_retain))
 }
 
 // ===== impl Metric =====
@@ -205,8 +210,8 @@ impl Root {
             .stamped()
     }
 
-    fn tls_config(&mut self) -> &mut TlsConfigReload {
-        &mut self.tls_config_reload
+    fn tls_config_reload(&self) -> &Arc<TlsConfigReload> {
+        &self.tls_config_reload
     }
 
     fn retain_since(&mut self, epoch: Instant) {
--- a/src/telemetry/metrics/tls_config_reload.rs
+++ b/src/telemetry/metrics/tls_config_reload.rs
@@ -1,4 +1,4 @@
-use std::{fmt, path::PathBuf, time::{SystemTime, UNIX_EPOCH}};
+use std::{fmt, path::PathBuf, sync::RwLock, time::{SystemTime, UNIX_EPOCH}};
 
 use telemetry::metrics::{Counter, Gauge, Metric, Scopes, labels::Errno};
 use transport::tls;
@@ -14,7 +14,10 @@ metrics! {
 }
 
 #[derive(Debug, Default)]
-pub struct TlsConfigReload {
+pub struct TlsConfigReload(RwLock<Inner>);
+
+#[derive(Debug, Default)]
+struct Inner {
     last_reload: Option<Gauge>,
     by_status: Scopes<Status, Counter>,
 }
@@ -31,21 +34,24 @@ enum Status {
 // ===== impl TlsConfigReload =====
 
 impl TlsConfigReload {
-    pub fn success(&mut self, when: &SystemTime) {
-        let t = when
+    pub fn reloaded(&self) {
+        let t = SystemTime::now()
             .duration_since(UNIX_EPOCH)
             .expect("times must be after UNIX epoch")
             .as_secs();
-        self.last_reload = Some(t.into());
 
-        self.by_status.scopes
+        let mut inner = self.0.write().expect("lock");
+        inner.last_reload = Some(t.into());
+
+        inner.by_status.scopes
             .entry(Status::Reloaded)
             .or_insert_with(|| Counter::default())
             .incr()
     }
 
-    pub fn error(&mut self, e: tls::ConfigError) {
-        self.by_status.scopes
+    pub fn failed(&self, e: tls::ConfigError) {
+        let mut inner = self.0.write().expect("lock");
+        inner.by_status.scopes
             .entry(e.into())
             .or_insert_with(|| Counter::default())
             .incr()
@@ -54,12 +60,14 @@ impl TlsConfigReload {
 
 impl fmt::Display for TlsConfigReload {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        if !self.by_status.scopes.is_empty() {
+        let inner = self.0.read().expect("lock");
+
+        if !inner.by_status.scopes.is_empty() {
             tls_config_reload_total.fmt_help(f)?;
-            tls_config_reload_total.fmt_scopes(f, &self.by_status, |s| &s)?;
+            tls_config_reload_total.fmt_scopes(f, &inner.by_status, |s| &s)?;
         }
 
-        if let Some(timestamp) = self.last_reload {
+        if let Some(timestamp) = inner.last_reload {
             tls_config_last_reload_seconds.fmt_help(f)?;
             tls_config_last_reload_seconds.fmt_metric(f, timestamp)?;
         }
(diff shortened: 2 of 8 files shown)


olix0r:
Based on #45

src/telemetry/metrics/tls_config_reload.rs
```diff
@@ -14,7 +14,10 @@ metrics! {
 }
 
 #[derive(Debug, Default)]
-pub struct TlsConfigReload {
+pub struct TlsConfigReload(RwLock<Inner>);
```
hawkw: TIOLI: It seems a little weird to me to put the `RwLock` inside this struct and the `Arc` outside of it, when seeing the type `Arc<TlsConfigReload>` earlier in this review, I thought, "huh, it's immutable?". This is fairly minor, but it seems to me that either both the sharing and mutability be encapsulated inside this type (and we just pass around an opaque `TlsConfigReload`), or move both of them outside the type (and pass around an `Arc<RwLock<TlsConfigReload>>`). I'm in favour of the first personally.

src/telemetry/metrics/tls_config_reload.rs
```diff
@@ -54,12 +60,14 @@ impl TlsConfigReload {
 
 impl fmt::Display for TlsConfigReload {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        if !self.by_status.scopes.is_empty() {
+        let inner = self.0.read().expect("lock");
```
hawkw: Nit: IMO, `.expect("lock")` is un-descriptive enough that it may as well just be `unwrap`...

src/telemetry/metrics/tls_config_reload.rs
```diff
@@ -31,21 +34,24 @@ enum Status {
...
             .incr()
     }
 
-    pub fn error(&mut self, e: tls::ConfigError) {
-        self.by_status.scopes
+    pub fn failed(&self, e: tls::ConfigError) {
+        let mut inner = self.0.write().expect("lock");
+        inner.by_status.scopes
             .entry(e.into())
             .or_insert_with(|| Counter::default())
```
hawkw: Nit: I know this code wasn't added in this branch, but this could just be `.or_default`.

src/telemetry/metrics/mod.rs
```diff
@@ -136,9 +136,14 @@ struct Stamped<T> {
 /// is a Hyper service which can be used to create the server for the
 /// scrape endpoint, while the `Record` side can receive updates to the
 /// metrics by calling `record_event`.
```
hawkw: I think this comment should be updated to add that it also returns the `TlsConfigReload`?

src/telemetry/mod.rs
```diff
@@ -16,7 +16,7 @@ pub fn new(
     metrics_retain_idle: Duration,
     taps: &Arc<Mutex<tap::Taps>>,
 ) -> (Sensors, metrics::Serve) {
-    let (metrics_record, metrics_serve) = metrics::new(process, metrics_retain_idle);
-    let s = Sensors::new(metrics_record, taps);
+    let (metrics_record, tls_config_reload, metrics_serve) = metrics::new(process, metrics_retain_idle);
+    let s = Sensors::new(metrics_record, tls_config_reload, taps);
```
hawkw: TIOLI: I wonder if it's really necessary for `sensors` to own the `TlsConfigReload`? It seems like it's only used once, to get a reference to the `sensors`' `TlsConfigReload` field and clone it, and then the sensors no longer need it. Maybe `telemetry::new` ought to just return `(Sensors, Arc<TlsConfigReload>, metrics::Serve)` instead?

olix0r: this will all change. this seemed like a fine interim state that makes the review small

src/telemetry/metrics/mod.rs
olix0r: This is all going to change in the near future. I'll probably remove the comment entirely

src/telemetry/metrics/tls_config_reload.rs
olix0r: The intent was that that we shouldn't dictate how the users of this hold the reference. If they want to hold a weak reference, let them.  Nor should the lock be exposed to users, as it's very much an implementation detail of how metrics are recorded.  The `Arc` explicitly says "this thing is shard", which is what I'm trying to convey.

That said, I think I may want to move the Arc for another reason.

seanmonstar: So, is it by design that the public interface of this type can mutate itself through an immutable reference?

olix0r: Basically, yes. I'm going to split this into two halves so the contract is clearer. The point is that this thing could be backed by an mpsc or a rwlock or some other shared datastructure that the writer can't assume exclusive ownership of.

src/telemetry/metrics/tls_config_reload.rs
seanmonstar: Agreed.

On a tangent, it kind of seems like we wish `Mutex` and `RwLock` didn't poison, it might be worth a new PR at some point cleaning up all the `expect`s on our locks somehow...

olix0r: I've avoided this altogether... will post momentarily

olix0r:
I've updated this as follow:

```
    Split tls_config_reload into read and write halves
    
    Rather than exposing a single, shared, `Arc<TlsConfigReload>` type, this
    change splits the concerns between a `Sensor` type and a `Fmt` type.
    
    The `Sensor` type holds a strong reference to the inner metrics
    structure and acquires a write lock on updates.
    
    The `Fmt` type holds a weak reference to the inner metrics structure so
    that the metrics server can print the metrics state, as long as it's
    actually held in memory.  If the `Sensor` is dropped (for instance,
    because TLS has been administratively disabled), then no metrics will be
    formatted by `Fmt`.
```

src/telemetry/metrics/mod.rs
olix0r: the return type is no longer changed, so no change is necessary yet

src/telemetry/metrics/tls_config_reload.rs
```diff
@@ -73,40 +115,41 @@ impl fmt::Display for TlsConfigReload {
...
+                ref path,
+                errno: None,
+            } => write!(
+                f,
+                "status=\"io_error\",path=\"{}\",errno=\"UNKNOWN\"",
+                path.display(),
+            ),
+            Status::InvalidPrivateKey => f.pad("status=\"invalid_private_key\""),
+            Status::InvalidEndEntityCert => f.pad("status=\"invalid_end_entity_cert\""),
+            Status::InvalidTrustAnchors => f.pad("status=\"invalid_trust_anchors\""),
```
olix0r: egh, looks like rustfmt was run here. I can back this out if it's distracting

hawkw:
@olix0r 
>    The `Sensor` type holds a strong reference to the inner metrics
>    structure and acquires a write lock on updates.
>   
>    The `Fmt` type holds a weak reference to the inner metrics structure so
>    that the metrics server can print the metrics state, as long as it's
>    actually held in memory.  If the `Sensor` is dropped (for instance,
>    because TLS has been administratively disabled), then no metrics will be
>    formatted by `Fmt`.

I'm (at least tentatively) in favour of this pattern.

src/telemetry/metrics/tls_config_reload.rs
```diff
@@ -13,8 +18,22 @@ metrics! {
...
+    let inner = Arc::new(RwLock::new(Inner::default()));
+    let fmt = Fmt(Arc::downgrade(&inner));
+    (Sensor(inner), fmt)
+}
+
+/// Supports recording TLS config reload metrics.
+///
+/// When this type is dropped, its metrics may no longer be formatted for prometheus.
+#[derive(Debug)]
+pub struct Sensor(Arc<RwLock<Inner>>);
```
hawkw: Out of interest: do we expect the metrics server to receive enough concurrent requests that it's worth using a `RwLock` rather than a `Mutex`?

olix0r: We don't expect the metrics server to get many requests.

If you're referring to the mutex around `Root`, it's going to be removed in a subsequent PR.

olix0r: Or is your point that RwLock adds overhead over a Mutex?  Does it?

src/telemetry/metrics/tls_config_reload.rs
olix0r: i don't think our version of indexmap has that

src/telemetry/metrics/tls_config_reload.rs
hawkw: > Or is your point that RwLock adds overhead over a Mutex? Does it?
It's somewhat platform-dependent, but at least on Linux, I believe `Mutex` is likely to be faster than `RwLock` when there's a small number of both readers and writers. You may find this interesting: https://www.slideshare.net/mitsunorikomatsu/performance-comparison-of-mutex-rwlock-and-atomic-types-in-rust.

hawkw: That said, I do think that `RwLock` does express intent more clearly, so 🤷‍♀️

src/telemetry/metrics/tls_config_reload.rs
hawkw: Nevermind, it looks like `std::collections::HashMap` has an `.or_default()` method but `IndexMap` doesn't.

src/telemetry/metrics/tls_config_reload.rs
```diff
@@ -25,41 +47,61 @@ enum Status {
...
     }
 }
 
-impl fmt::Display for TlsConfigReload {
+// ===== impl Fmt =====
+
+impl fmt::Display for Fmt {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        if !self.by_status.scopes.is_empty() {
+        let lock = match self.0.upgrade() {
```
hawkw: TIOLI: 
```rust
let lock = match self.0
     .upgrade()
     .and_then(|lock| lock.lock().ok())
{
    None => return Ok(()),
    Some(inner) => inner
}
```

olix0r: i'm pretty sure this doesn't work because `lock's` lifetime is too short

hawkw: Oh, yeah, that's probably the case. Nevermind.

PR #49, opened 2018-08-09
Stop tracking transport metrics for eviction

olix0r:
Metric eviction was introduced to protect against the situation where,
over time, the proxy addresses an effectively unbounded number of
endpoints. Metrics with endpoint-specific dimensions must be removed
over time to prevent a form of memory leak.

Eviction is implemented for transport stats. However, transport stats do
not yet contain per-endpoint dimensions. The eviction logic is
superfluous because transport metrics are bounded and small.

In preparation of changes to transport telemetry, this disables tracking
of transport metrics for eviction. This logic will be restored when we
support per-endpoint transport metrics.

### DIFF
--- a/src/telemetry/metrics/mod.rs
+++ b/src/telemetry/metrics/mod.rs
@@ -197,21 +197,17 @@ impl Root {
 
     fn transport(&mut self, labels: TransportLabels) -> &mut transport::OpenMetrics {
         self.transports.scopes.entry(labels)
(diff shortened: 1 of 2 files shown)


olix0r:
@seanmonstar we don't have any address metadata in transport stats currently. I filed linkerd/linkerd2#1430 to track adding this metadata (at which point we will need an eviction strategy)

PR #52, opened 2018-08-09
Move Errno into telemetry

olix0r:
Following on #50, this change moves the `Errno` types into `telemetry`
so that metrics implementation can be moved out of the `metrics` module.


src/telemetry/errno.rs
```diff
@@ -169,3 +169,21 @@ mk_err_enum! {
         _   => UNKNOWN_ERRNO
     }
 }
+
+#[cfg(target_os="windows")]
```
hawkw: This will conflict with my currently open branch [eliza/fix-win](https://github.com/linkerd/linkerd2-proxy/tree/eliza/fix-win), which fixes issue linkerd/linkerd2#1396. I'd like to merge that change first and rebase this onto it, as the code here currently won't compile.

My branch is currently waiting on the reporter to verify the fix, but I can go ahead and open a PR for it so we can get that merged first? I'm pretty sure it does fix the issue, which is a pretty simple syntax error.

hawkw: Opened #54 to get the Windows fix in.

olix0r:
@hawkw ptal. i've changed Errno slightly so that the platform-dependent type isn't exposed outside of the errno module.

PR #56, opened 2018-08-10
Make `Metric::fmt_scopes` generic over an Iterator

olix0r:
The `Scopes` type should'nt be critical to metrics formatting.

This change makes the `Metric` type unaware of `Scopes` so that other
labeling mechanisms can be used if appropriate.

Follows #55.


olix0r:
ugh, merged this into the wrong branch

PR #57, opened 2018-08-10
Consolidate labels into metrics::transport

olix0r:
Various transport-specific labels are defined in the common
`metrics::labels` module.

In preparation to refactor transport metrics into Sensor and Report
halves, this change moves all transport-specific labels into the
transport metrics module.

### DIFF
--- a/src/telemetry/metrics/labels.rs
+++ b/src/telemetry/metrics/labels.rs
@@ -71,7 +71,7 @@ pub struct TlsStatus(ctx::transport::TlsStatus);
 
 impl RequestLabels {
     pub fn new(req: &ctx::http::Request) -> Self {
-        let direction = req.server.proxy.as_ref().into();
+        let direction = Direction::from_context(req.server.proxy.as_ref());
 
         let outbound_labels = req.dst_labels().cloned();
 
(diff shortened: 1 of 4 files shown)


src/telemetry/metrics/labels.rs
```diff
@@ -244,11 +214,11 @@ impl fmt::Display for Classification {
 
 // ===== impl Direction =====
 
-impl Direction {
-    fn from_context(context: &ctx::Proxy) -> Self {
-        match context {
-            &ctx::Proxy::Inbound(_) => Direction::Inbound,
-            &ctx::Proxy::Outbound(_) => Direction::Outbound,
+impl<'a> From<&'a ctx::Proxy> for Direction {
```
hawkw: FWIW, I had originally written this as a `From` conversion when I first added the `Direction` enum, but @briansmith asked me to change it: https://github.com/linkerd/linkerd2/pull/661#discussion_r179904788. Personally, I like it better as a `From` conversion, but I figured I ought to put that out there...

src/telemetry/metrics/transport.rs
```diff
@@ -23,6 +24,36 @@ pub(super) struct OpenMetrics {
...
+    /// Was the transport secured with TLS?
+    tls_status: TlsStatus,
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub enum Peer { Src, Dst }
+
+/// Labels describing the end of a TCP connection
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub struct TransportCloseLabels {
```
hawkw: Same comment here as the one on `TransportLabels` above.

src/telemetry/metrics/transport.rs
```diff
@@ -23,6 +24,36 @@ pub(super) struct OpenMetrics {
 
 pub(super) type CloseScopes = Scopes<TransportCloseLabels, CloseMetrics>;
 
+/// Labels describing a TCP connection
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub struct TransportLabels {
```
hawkw: TIOLI: Now that this lives in the `Transport` module, perhaps the name ought to be `transport::Labels`? Consumers in other modules can still `use transport::{Labels as TransportLabels}`.

olix0r: this is all going to change a lot more

olix0r: this is all going to change a lot more

PR #58, opened 2018-08-10
Remove `record` benches

olix0r:
As we begin to remove event-based metrics telemetry, the `record` bench
test becomes an obstacle that needs to be updated at every step.

Given that we have already removed the event channel and plan to remove
the `Record` type entirely, it's suitable to remove the bench test.

While this is done, the `test-benches` Make target and rust-nightly
travis stage are obsoleted.


hawkw:
This seems reasonable, given that the code these benchmarks test is being removed. However, I do think that as we continue to rewrite this subsystem, it would be good if we can find some way to measure its performance in a way that we can compare to the previous implementation. I'm not sure if that's something we can do easily, but I figured it was worth bringing up.

PR #61, opened 2018-08-13
Add names to timeout error messages

hawkw:
The proxy contains several timeouts which can fail operations.
Currently, when formatting the error types produced by any of these
timeouts in log messages, the timeout error returns "operation timed out
after $DURATION". Since these errors are typically logged in a different
scope from the one where the timeout was created, it's often not
immediately clear _what_ timed out. An experienced user can often
determine what timed out based on the duration of the timeout, but for
new users, the logs are not very informative.

This branch adds an optional `name` field to the `Timeout` and
`TimeoutError` types. If a name is provided, the `TimeoutError` will
format as "$NAME timed out after $DURATION" instead. Otherwise, it will
continue to use the original behaviour. It also names all the current
uses of `Timeout` in the proxy.

Implementation Details 
----------------------

Note that the `name` field on `Timeout` and `TimeoutError` is an `&str`
with a generic lifetime. In practice, since the returned timeout futures
are part of future chains which must be valid for the `'static`
lifetime, this must be an `&'static str`. This means that the name may
not be dynamically formatted, so we can't have messages such as
"building outbound $PROTOCOL client to $DESTINATION timed out".

I considered the alternative approach of making the `name` field a
generic type that implements `Display`, so that it could also be a
`String` or an `Arc<String>` or similar. However, I wasn't sure if the
performance impact of doing so would be worth the more specific error
messages --- even using only string constants as names makes the logged
messages much more descriptive than they were previously. I'd be happy
to change this if we think adding more information to these messages
would be worth the performance impact of heap allocating strings here.

Examples
--------

Before:
```
➜ cargo test --test discovery -- http2::outbound_times_out
   Compiling linkerd2-proxy v0.1.0 (file:///Users/eliza/Code/linkerd2-proxy)
    Finished dev [unoptimized + debuginfo] target(s) in 29.86s
     Running target/debug/deps/discovery-88fd1ee2f3bc7947

running 1 test
ERROR 2018-08-13T17:59:06Z: linkerd2_proxy: turning operation timed out after 100ms into 500
test http2::outbound_times_out ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 32 filtered out
```

After:
```
➜ cargo test --test discovery -- http2::outbound_times_out
    Finished dev [unoptimized + debuginfo] target(s) in 0.25s
     Running target/debug/deps/discovery-88fd1ee2f3bc7947

running 1 test
ERROR 2018-08-13T17:57:59Z: linkerd2_proxy: turning binding outbound client timed out after 100ms into 500
test http2::outbound_times_out ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 32 filtered out
```

### DIFF
--- a/src/transparency/tcp.rs
+++ b/src/transparency/tcp.rs
@@ -71,7 +71,8 @@ impl Proxy {
         let c = Timeout::new(
             transport::Connect::new(orig_dst, tls),
             self.connect_timeout,
-        );
+        )
+        .named("TCP connection");
         let connect = self.sensors.connect(c, &client_ctx);
 
         future::Either::A(connect.connect()
--- a/src/transparency/tcp.rs
+++ b/src/transparency/tcp.rs
@@ -71,8 +71,7 @@ impl Proxy {
         let c = Timeout::new(
             transport::Connect::new(orig_dst, tls),
             self.connect_timeout,
-        )
-        .named("TCP connection");
+        ).named("TCP connect");
         let connect = self.sensors.connect(c, &client_ctx);
 
         future::Either::A(connect.connect()
--- a/src/control/destination/background/client.rs
+++ b/src/control/destination/background/client.rs
@@ -27,7 +27,7 @@ use watch_service::Rebind;
 /// Type of the client service stack used to make destination requests.
 pub(super) struct ClientService(AddOrigin<Backoff<LogErrors<Reconnect<
         tower_h2::client::Connect<
-            Timeout<'static, LookupAddressAndConnect>,
+            Timeout<LookupAddressAndConnect>,
             ::logging::ContextualExecutor<
                 ::logging::Client<
                     &'static str,
@@ -72,7 +72,6 @@ type LogError = ReconnectError<
     tower_h2::client::Error,
     tower_h2::client::ConnectError<
         TimeoutError<
-            'static,
             io::Error
         >
     >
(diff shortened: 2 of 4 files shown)


hawkw:
@seanmonstar 
> Is the lifetime really needed? Seems like it littered all over the code, and we only have use for static strings...

I was thinking it was better to be generic, but you're right, I don't think we'll ever use a string with a non-static lifetime. I'll change it.

src/transparency/tcp.rs
```diff
@@ -71,7 +71,8 @@ impl Proxy {
         let c = Timeout::new(
             transport::Connect::new(orig_dst, tls),
             self.connect_timeout,
-        );
+        )
+        .named("TCP connection");
```
olix0r: please call this "TCP connect" so that it's less ambiguous.

also the indentation seems pretty weird here.

hawkw: I called it "TCP connection" because "TCP connection timed out" seemed more grammatically correct than "TCP connect timed out". WDYT of "opening TCP connection" so that which end of  the connection this is seems clearer?

And, yeah, I'll fix the indentation, sorry about that.

hawkw: @olix0r I changed both the name and the indentation in cc16f125d72ba58ce3b1360b6133822b63ca4878 :)

seanmonstar:
Closing as inactive.

PR #65, opened 2018-08-14
Narrow transport metric's API

olix0r:
The `telemetry::metrics::transport` module exposes much of its
implementation details to callers, which makes it difficult to make
changes to how the module is structured. In preparation for further
refactors, this change narrows the module's public API:

All labels and scopes types have been made private. A single, public
`Transports` type has been introduce to describe the entire public
interface of the module. This has been crafted to be free of `event`
types and to have minimal external dependencies.

A new `transport::Eos` type has been introduced to replace the
overloaded `labels::Classification` type -- this type was initially
introduced to model _HTTP response_ classification, but it was
reused for transports. This is an undesirable coupling that will have to
be broken when we start to adddress HTTP response classification
properly.

### DIFF
--- a/src/telemetry/metrics/transport.rs
+++ b/src/telemetry/metrics/transport.rs
@@ -3,30 +3,49 @@ use std::time::Duration;
 
 use ctx;
 use super::{
-    labels::{Classification, Direction, TlsStatus},
+    labels::{Direction, TlsStatus},
     latency,
     Counter,
     Gauge,
     Histogram,
     Scopes,
 };
-use telemetry::{event, Errno};
+use telemetry::Errno;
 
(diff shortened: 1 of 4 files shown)


src/telemetry/metrics/transport.rs
```diff
@@ -3,30 +3,49 @@ use std::time::Duration;
...
+
+    tcp_close_total: Counter { "Total count of closed connections" },
+    tcp_connection_duration_ms: Histogram<latency::Ms> { "Connection lifetimes" }
+}
+
+/// Holds all transport stats.
+///
+/// Implements `fmt::Display` to render prometheus-formatted metrics for all transports.
+#[derive(Debug, Default)]
+pub struct Transports {
```
hawkw: Had been wondering about doing this for a while. 👍

PR #66, opened 2018-08-14
Store transport metrics hierarchically

olix0r:
Previously, transport metrics have been stored in two "scopes": an
"open" scope, storing metrics by transport class, and a "close" scope
storing metrics by transport and end-of-stream classes.

Instead of storing these as parallel maps, this changes transport
metrics to be stored hierarchically so that the end-of-stream metrics
are stored _within_ the transport metrics type.

This will make it possible to share the transport metrics structure
directly with the transport wrapper so that a global lock need only be
taken at connect-time. Subsequent updates will then be scoped to the
shared transport structure.

This helps to setup linkerd/linkerd2#1430, among other improvements.

Furthermore, the `Scopes` type is no longer used for transport metrics,
since it's much easier to use the full affordances of `indexmap`,
especially now that it isn't part of an API.

### DIFF
--- a/src/telemetry/metrics/transport.rs
+++ b/src/telemetry/metrics/transport.rs
@@ -52,6 +52,9 @@ struct Metrics {
 }
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+enum Peer { Src, Dst }
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub enum Eos {
     Clean,
     Error {
@@ -65,9 +68,6 @@ struct EosMetrics {
     close_total: Counter,
     connection_duration: Histogram<latency::Ms>,
 }
-
-#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
-pub enum Peer { Src, Dst }
 
 struct FmtEos<'a>(&'a Key, &'a Eos);
 
--- a/src/telemetry/metrics/transport.rs
+++ b/src/telemetry/metrics/transport.rs
@@ -1,3 +1,4 @@
+use indexmap::IndexMap;
 use std::fmt;
 use std::time::Duration;
 
@@ -8,7 +9,6 @@ use super::{
     Counter,
     Gauge,
     Histogram,
-    Scopes,
 };
 use telemetry::Errno;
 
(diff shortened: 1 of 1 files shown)


src/telemetry/metrics/transport.rs
```diff
@@ -27,32 +27,28 @@ metrics! {
...
-    closes: CloseScopes,
+    metrics: IndexMap<Key, Metrics>,
 }
 
-type OpenScopes = Scopes<TransportLabels, OpenMetrics>;
+/// Describes the dimensions across which transport metrics are aggregated.
+///
+/// Implements `fmt::Display` to render a comma-separated list of key-value pairs.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+struct Key {
```
hawkw: TIOLI: I might've called this "labels"...

src/telemetry/metrics/transport.rs
```diff
@@ -97,116 +88,107 @@ impl Transports {
         rx: u64,
         tx: u64,
     ) {
-        let key = TransportLabels::new(ctx);
+        let key = Key::new(ctx);
+
+        let metrics = self.metrics.entry(key).or_insert_with(|| Default::default());
```
hawkw: If we're recording a close, and an open didn't exist for a given key, that's probably a bug, right? It seems like this should at least log.

src/telemetry/metrics/transport.rs
```diff
@@ -27,32 +27,28 @@ metrics! {
...
+}
 
+/// Holds all of the metrics for a class of transport.
 #[derive(Debug, Default)]
-struct OpenMetrics {
+struct Metrics {
     open_total: Counter,
     open_connections: Gauge,
     write_bytes_total: Counter,
     read_bytes_total: Counter,
```
hawkw: I believe the read/written bytes stats used to be labeled with the class of EOS as well? Not sure if this matters, but it is technically a change in behaviour that _could_ be breaking.

olix0r: they were not.

src/telemetry/metrics/transport.rs
olix0r: this api is going to change

src/telemetry/metrics/transport.rs
olix0r: at the end of this, i expect that there will be almost nothing in the telemetry module that talks about labels.  This should be a prometheus-detail

PR #67, opened 2018-08-15
rfc: Introduce the FmtPrometheus and FmtLabels traits

olix0r:
There are many different types of things that implement `fmt::Display`
in the metrics library: lists of key-value labels, individial label keys
or values, groups of metrics output, etc. While modifying metrics code,
it can be very easy to use something that's not a label in the context
of a label.

### DIFF
--- a/src/telemetry/metrics/labels.rs
+++ b/src/telemetry/metrics/labels.rs
@@ -9,6 +9,7 @@ use http;
 
 use ctx;
 use conditional::Conditional;
+use super::prom::FmtLabels;
 use transport::tls;
 
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
@@ -23,7 +24,7 @@ pub struct RequestLabels {
 
     /// The value of the `:authority` (HTTP/2) or `Host` (HTTP/1.1) header of
     /// the request.
-    authority: Option<http::uri::Authority>,
+    authority: Authority,
 
     /// Whether or not the request was made over TLS.
     tls_status: TlsStatus,
@@ -35,11 +36,11 @@ pub struct ResponseLabels {
     request_labels: RequestLabels,
 
     /// The HTTP status code of the response.
-    status_code: u16,
+    status_code: StatusCode,
 
     /// The value of the grpc-status trailer. Only applicable to response
     /// metrics for gRPC responses.
-    grpc_status_code: Option<u32>,
+    grpc_status: Option<GrpcStatus>,
 
     /// Was the response a success or failure?
     classification: Classification,
@@ -66,6 +67,15 @@ pub struct DstLabels {
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub struct TlsStatus(ctx::transport::TlsStatus);
 
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+struct Authority(Option<http::uri::Authority>);
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+struct StatusCode(u16);
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+struct GrpcStatus(u32);
+
 // ===== impl RequestLabels =====
 
 impl RequestLabels {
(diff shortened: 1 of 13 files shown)


olix0r:
https://github.com/linkerd/linkerd2-proxy/pull/68 shows why this should be useful -- we only want/need to store a `ctx::Proxy` type in our data structures. The label-formatting details should be decoupled from the structural representation of the metrics store.

src/telemetry/metrics/labels.rs
```diff
@@ -147,21 +156,25 @@ impl ResponseLabels {
...
-        )?;
-
-        if let Some(ref status) = self.grpc_status_code {
-            // leading comma added between the status code label and the
-            // gRPC status code labels, if there is a gRPC status code.
-            write!(f, ",grpc_status_code=\"{}\"", status)?;
-        }
+impl FmtLabels for ResponseLabels {
+    fn fmt_labels(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let l = (((&self.request_labels,
```
hawkw: Personally, I find this expression rather hard to parse...

src/telemetry/metrics/prom.rs
```diff
@@ -0,0 +1,126 @@
+use std::fmt;
+use std::marker::PhantomData;
+
+/// Writes a block of metrics in prometheus-formatted output.
+pub trait FmtPrometheus {
```
hawkw: I feel like the `FmtPrometheus` and `FmtMetric` trait names aren't super clear --- the comments help a bit, but just seeing the trait names, it's not immediately clear which does what. It might be good if that could be made clearer...

src/telemetry/metrics/serve.rs
```diff
@@ -180,3 +182,9 @@ impl Error for ServeError {
         }
     }
 }
+
+impl<F: FmtPrometheus> fmt::Display for PromWriter<F> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
hawkw: This layer of indirection seemed unnecessary to me until I realized you couldn't `fmt::Display::fmt(metrics, &mut writer)` because `writer` is `io::Write` but not a `Formatter`. Maybe there ought to be a comment here somewhere?

src/telemetry/metrics/prom.rs
```diff
@@ -0,0 +1,126 @@
...
+
+// ===== impl FmtLabels =====
+
+impl<'a, A: FmtLabels + 'a> FmtLabels for &'a A {
+    fn fmt_labels(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        (*self).fmt_labels(f)
+    }
+}
+
+impl<A: FmtLabels, B: FmtLabels> FmtLabels for (A, B) {
```
hawkw: The helper methods on tuples are cute. I wonder if it would be simpler to just define one on `IntoIterator<Item=&'a FmtLabels>`, though? Then, the case of `Option`al labels could be handled with `filter_map`.

olix0r: we need to know that there is at least one valid label, otherwise all of the comma insertion breaks.  This is why we dont provide impls for `Option<A>` or `(Option<A>, Option<B>)`...

are you suggesting that we support `(A, IntoIterator<Item=FmtLabels>)` instead of the Option? would that work?

hawkw: > we need to know that there is at least one valid label, otherwise all of the comma insertion breaks. 

I think we can do something like this:
```rust 
let mut iter = i.into_iter() // iterator of Option<FmtLabels>
    .filter_map(|i| i); // filter out any nones.

if let Some(label) = iter.next() {
    write!(f, "{}", label)?;

    for label in iter {
        write!(f, ",{}", label)?;
    }
}
```

regardless, this seems fine as-is!

olix0r: I think this has two problems:

1. `FmtLabels` must not represent an empty set of labels. Either that, or the `FmtLabels` type would need to expose a `is_empty()`. This is important for the purposes of composition.  If I have an iterator of `FmtLabels` that i want to combine, how do i know if any one of those items is itself empty?
2. AIUI, we would almost always need to use `Box<FmtLabels>`, because we are trying to handle diverse types.

PR #68, opened 2018-08-15
Delete ctx::Process

olix0r:
In order to start dismantling the monolithic ctx structures, this change
removes the root `ctx::Process` type. This simplifies `ctx::Proxy` such
that is copyable and need not be `Arc`ed.

`telemetry::metrics::labels::Direction` has been updated to decorate
`ctx::Proxy` instead of modeling the same variants directly as an enum.

### DIFF
--- a/src/control/pb.rs
+++ b/src/control/pb.rs
@@ -193,10 +193,10 @@ impl ctx::transport::Server {
     }
 
     fn direction(&self) -> tap::tap_event::ProxyDirection {
-        if self.proxy.is_outbound() {
-            tap::tap_event::ProxyDirection::Outbound
-        } else {
-            tap::tap_event::ProxyDirection::Inbound
+        use ctx::Proxy::*;
+        match self.proxy {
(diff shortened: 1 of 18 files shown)


src/control/pb.rs
```diff
@@ -193,10 +193,10 @@ impl ctx::transport::Server {
     }
 
     fn direction(&self) -> tap::tap_event::ProxyDirection {
-        if self.proxy.is_outbound() {
-            tap::tap_event::ProxyDirection::Outbound
-        } else {
-            tap::tap_event::ProxyDirection::Inbound
+        use ctx::Proxy::*;
```
seanmonstar: Not needed to change. I've moved away from this pattern (glob importing an enum's variants and then matching on them unqualified), since typoing a variant name isn't a compiler error, just a possible lint warning. That's because if the pattern isn't an actual variant, then it matches as if that were an identifier you wanted to use in the match arm. :shrug:

PR #69, opened 2018-08-16
travis: Avoid caching proxy binaries

olix0r:
There's no reason that we should ever need to cache proxy binaries
between CI runs.

However, for each pull request build we appear to cache a ~125M
binary in target/debug/linkerd2-proxy.

Cache interactions seem to account for much of our CI times, so any
reduction here may be helpful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,8 @@ jobs:
     - if: type = pull_request
       script: make test
       rust: stable
+      before_cache:
+        - rm -rf target/debug/linkerd2-proxy
 
     # On master, package an artifact and run tests in release mode before
     # publishing the artifact to build.l5d.io/linkerd2-proxy.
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,8 +13,11 @@ jobs:
   include:
     # On PRs and branches, quickly run tests in debug mode.
     - if: type = pull_request
-      script: make test
+      script:
+        - make test
       rust: stable
+      before_cache:
+        - rm -rf target/debug/linkerd2-proxy target/release
```
seanmonstar: Is the culprit actually just the `cache: cargo` line above?

olix0r: we _want_ `cache: cargo`, since it enables caching of intermediate artifacts and dependencies and such. There are just parts of the `target` directory that aren't worth caching.

PR #71, opened 2018-08-16
Make telemetry::metrics::labels::DstLabels private

olix0r:
The `DstLabels` type has a large and undefined scope. Over time, it's
become critical in a variety of contexts, even outside of the scope of
prometheus endpoint labels.

This change makes `DstLabels` private to the
`telemetry::metrics::labels` module. This more clearly separates the
concerns of prometheus labeling from discovery, etc.

Now, destination metadata includes an index map of arbitrary metadata
labels.

Previously, dst label strings would be formatted eagerly at
service-discovery-time. This change moves this string formatting into
the data path(!). I think this is an acceptable tradeoff temporarily,
while we work to stop constructing RequestLabels in the data path
altogether.

### DIFF
--- a/src/control/destination/mod.rs
+++ b/src/control/destination/mod.rs
@@ -24,10 +24,10 @@
 //! - We need some means to limit the number of endpoints that can be returned for a
 //!   single resolution so that `control::Cache` is not effectively unbounded.
 
+use indexmap::IndexMap;
 use std::net::SocketAddr;
 use std::sync::{Arc, Weak};
 use std::time::Duration;
-use tls;
 
 use futures::{
     sync::mpsc,
@@ -41,7 +41,7 @@ use tower_discover::{Change, Discover};
 use tower_service::Service;
 
 use dns;
-use telemetry::DstLabels;
+use tls;
 use transport::{DnsNameAndPort, HostAndPort};
 
 pub mod background;
@@ -93,8 +93,8 @@ pub struct Resolution<B> {
 /// Metadata describing an endpoint.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct Metadata {
-    /// A set of Prometheus metric labels describing the destination.
-    dst_labels: Option<DstLabels>,
+    /// Arbitrary endpoint labels. Useful for telemetry, especially.
+    labels: IndexMap<String, String>,
 
     /// A hint from the controller about what protocol (HTTP1, HTTP2, etc) the
     /// destination understands.
@@ -257,8 +257,8 @@ impl Responder {
 impl Metadata {
     /// Construct a Metadata struct representing an endpoint with no metadata.
     pub fn no_metadata() -> Self {
-        Metadata {
-            dst_labels: None,
+        Self {
+            labels: IndexMap::default(),
             protocol_hint: ProtocolHint::Unknown,
             // If we have no metadata on an endpoint, assume it does not support TLS.
             tls_identity:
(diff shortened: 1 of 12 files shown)


src/telemetry/metrics/labels.rs
```diff
@@ -79,7 +74,7 @@ impl RequestLabels {
     pub fn new(req: &ctx::http::Request) -> Self {
         let direction = Direction::new(req.server.proxy);
 
-        let outbound_labels = req.dst_labels().cloned();
+        let outbound_labels = DstLabels::new(req.labels());
```
hawkw: > Previously, dst label strings would be formatted eagerly at
> service-discovery-time. This change moves this string formatting into
> the data path(!)

Why not do the string formatting at metrics format time instead? The `RequestLabels` could have an `Arc<IndexMap<...>>` and format it on the control thread when the metrics are served. This would hurt the metrics scrape performance a bit, but it seems to me that minimizing work in the data path is more important?

If we're concerned about doing the same work multiple times, we could also use interior mutability to have `DstLabels` memoize the string once we've formatted it once...this would admittedly add a bit of complexity, but it could at least be contained inside the label type.

src/ctx/mod.rs
```diff
@@ -49,18 +48,16 @@ pub mod test_util {
...
+        labels: IndexMap<String, String>,
         tls: ctx::transport::TlsStatus,
-    ) -> Arc<ctx::transport::Client>
-    where
-        L: IntoIterator<Item=(S, S)>,
-        S: fmt::Display,
-    {
-        let meta = destination::Metadata::new(DstLabels::new(labels),
+    ) -> Arc<ctx::transport::Client> {
+        let meta = destination::Metadata::new(
```
hawkw: Thanks for reformatting this --- I find it much more readable now. :)

src/control/pb.rs
```diff
@@ -204,10 +204,8 @@ impl ctx::transport::Client {
     fn dst_meta(&self) -> tap::tap_event::EndpointMeta {
         let mut meta = tap::tap_event::EndpointMeta::default();
 
-        if let Some(ref d) = self.dst_labels() {
-            for (k, v) in d.as_map() {
-                meta.labels.insert(k.clone(), v.clone());
-            }
+        for (k, v) in self.labels() {
+            meta.labels.insert(k.clone(), v.clone());
```
hawkw: TIOLI: I think this could just be
```rust
meta.labels.extend(self.labels().cloned());
```

src/control/destination/mod.rs
```diff
@@ -93,8 +93,8 @@ pub struct Resolution<B> {
 /// Metadata describing an endpoint.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct Metadata {
-    /// A set of Prometheus metric labels describing the destination.
-    dst_labels: Option<DstLabels>,
+    /// Arbitrary endpoint labels. Useful for telemetry, especially.
```
hawkw: This is definitely a nit, sorry, but I find "Useful for telemetry, especially" a bit awkward --- if the labels are basically treated as opaque by the proxy and can contain arbitrary labels, aren't they pretty much useful _only_ for telemetry?

olix0r: The `Metadata` type doesn't dictate how labels are used. They will be put into telemetry. However, there's nothing that prevents middlewares from peaking into the labels to make other decisions.

src/telemetry/metrics/labels.rs
olix0r: We'll get there. Currently, `DstLabels` needs to impl `Hash`, but `IndexMap` doesn't impl `Hash`.  I'm working on breaking these requirements so that we can format the strings at Fmt-time.

olix0r:
I agree the string copying is not ideal. I think, practically, we're talking about the equivalent a few extra formatted log statements. It'll probably impact the long tail, but it's probably a reasonable tradeoff, if only temporarily.

olix0r:
@briansmith that's right.  This has already (largely) been done for transport metrics, though it's in a somewhat incremental state of refactoring. I'm hoping that we get to an end state where there's an `http` module that has _all_ of the http-specific concerns, a `transport` module that has _all_ of the transport-specific concerns, etc.

I very much want to kill off the central `labels` module (I think most of what's left there is HTTP-specific, which is next on my hit list after the transport refactor moves a little further), the `ctx` module, and any other module where we have cross-cutting concerns intermingled.

I should also point out https://github.com/linkerd/linkerd2-proxy/pull/67, which will let us start to more clearly separate the metrics-rendering concerns from caching/structural concerns.

PR #72, opened 2018-08-17
Remove the labels::Direction type

olix0r:
Following #67 and #68, the `labels::Direction` type can be obsoleted with
an `impl FmtLabels for ctx::Proxy`.


src/telemetry/metrics/labels.rs
```diff
@@ -203,17 +196,9 @@ impl FmtLabels for Classification {
...
-// ===== impl Direction =====
-
-impl Direction {
-    pub fn new(ctx: ctx::Proxy) -> Self {
-        Direction(ctx)
-    }
-}
-
-impl FmtLabels for Direction {
+impl FmtLabels for ctx::Proxy {
```
seanmonstar: Hm, this is actually a case where I kind of feel like the newtype provides value, and prevents future bugs if `ctx::Proxy` is used elsewhere, maybe a different set of labels.

hawkw: +1 to @seanmonstar's comment. I'm not sure if I find this change to be clearer than the previous code.

olix0r: `ctx::Proxy` is basically only used for labels and for tap.  I'd either prefer to have a newtype that's local to the transport module and one that's local to the http module, however this makes it easy to change the label for one and not the other, which is undesirable.

What I'm primarily trying to do is avoid having any labeling-specific types used anywhere outside of label formatting.  It's complicated and confusing to have a bunch of redundant types. I'd prefer to remove the Direction type and only introduce a wrapper type if we actually have a need to have redundant representations. I'm betting against ever needing it, though.

src/telemetry/mod.rs
```diff
@@ -19,13 +19,15 @@ mod errno;
 pub mod event;
 mod metrics;
 mod process;
+mod proxy_label;
```
hawkw: I'm not sure if this needs it's own module...

olix0r:
Is c0da52a preferable? This feels like needless boilerplate to me, but if others feel this is indeed much better, i'm okay with this compromise.

src/telemetry/mod.rs
olix0r: What are you suggesting?

I prefer that mod.rs include only types/fns that are core to the module. Otherwise they become dumping grounds for lots of random unrelated stuff.

seanmonstar:
If you feel strongly that the extra newtype doesn't pull it's weight, I'm not set on it. I'm basically -0 on removing it, which is just mildly against, but I don't think either way is a big deal.

hawkw:
@olix0r I was going to approve this before c0da52a, after seeing your rationale...like @seanmonstar, I was only mildly skeptical of the original change.

olix0r:
Ok, in the interest of being minimal, i'm going to revert the most recent change and proceed, though add some commentary. Happy to revisit this when the newtype buys us something

PR #73, opened 2018-08-17
Remove the labels::TlsStatus type

olix0r:
Following #67 and #68, the `labels::TlsStatus` type can be removed in
favor of extending underlying `ctx::transport::TlsStatus` type to
implement `FmtLabels`.

No functional changes.

### DIFF
--- a/src/telemetry/metrics/labels.rs
+++ b/src/telemetry/metrics/labels.rs
@@ -25,7 +25,7 @@ pub struct RequestLabels {
     authority: Authority,
 
     /// Whether or not the request was made over TLS.
-    tls_status: TlsStatus,
+    tls_status: ctx::transport::TlsStatus,
 }
 
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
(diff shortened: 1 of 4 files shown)


src/telemetry/metrics/labels.rs
```diff
@@ -254,43 +251,12 @@ impl FmtLabels for DstLabels {
 
 // ===== impl TlsStatus =====
 
-impl FmtLabels for TlsStatus {
+impl FmtLabels for ctx::transport::TlsStatus {
     fn fmt_labels(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match self.0 {
+        match *self {
```
hawkw: I'm not sure if the dereference here is strictly necessary after Rust 1.26's [match ergonomics](https://github.com/rust-lang/rfcs/blob/master/text/2005-match-ergonomics.md) changes?

src/transport/tls/config.rs
```diff
@@ -154,6 +155,24 @@ impl From<ReasonForNoIdentity> for ReasonForNoTls {
     }
 }
 
+impl fmt::Display for ReasonForNoIdentity {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
hawkw: Moving this formatting to `ReasonForNoIdentity` seems like a good thing 👍

PR #74, opened 2018-08-17
Split transport metrics into read and write halves

olix0r:
In anticipation of removing Transport-related Event types, we want to
separate the concerns of recording transport metrics updates from
reporting them to the metrics endpoint.

The transport module has been split into `Registry` and `Report` types:
`Registry` is responsible for recording updates, and `Report` is
responsible for rendering metrics.

This has the benefit of removing transport metrics from the global metrics
lock.

### DIFF
--- a/src/telemetry/metrics/record.rs
+++ b/src/telemetry/metrics/record.rs
@@ -12,13 +12,14 @@ use super::transport;
 #[derive(Clone, Debug)]
 pub struct Record {
     metrics: Arc<Mutex<Root>>,
+    transports: transport::Registry,
 }
 
 // ===== impl Record =====
 
 impl Record {
-    pub(super) fn new(metrics: &Arc<Mutex<Root>>) -> Self {
-        Self { metrics: metrics.clone() }
+    pub(super) fn new(metrics: &Arc<Mutex<Root>>, transports: transport::Registry) -> Self {
+        Self { metrics: metrics.clone(), transports }
     }
 
     #[inline]
@@ -67,9 +68,7 @@ impl Record {
             },
 
             Event::TransportOpen(ref ctx) => {
-                self.update(|metrics| {
-                    metrics.transports().open(ctx);
-                })
+                self.transports.open(ctx);
             },
 
             Event::TransportClose(ref ctx, ref close) => {
@@ -80,10 +79,8 @@ impl Record {
                         errno: close.errno.map(|e| e.into())
                     }
                 };
-                self.update(|metrics| {
-                    metrics.transports()
-                        .close(ctx, eos, close.duration, close.rx_bytes, close.tx_bytes);
-                })
+                self.transports
+                    .close(ctx, eos, close.duration, close.rx_bytes, close.tx_bytes);
             },
         };
     }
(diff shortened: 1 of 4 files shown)


src/telemetry/transport/mod.rs
```diff
@@ -1,16 +1,23 @@
 use indexmap::IndexMap;
 use std::fmt;
+use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 use ctx;
-use super::{
+use telemetry::Errno;
+use telemetry::metrics::{
+<<<<<<< HEAD
```
seanmonstar: ohai merge conflicts...

src/telemetry/transport/mod.rs
```diff
@@ -86,91 +123,93 @@ impl Transports {
...
+        let key = Key::new(ctx);
+        let metrics = inner.get_or_default(key);
         metrics.open_connections.decr();
         metrics.read_bytes_total += rx;
         metrics.write_bytes_total += tx;
 
-        let class = metrics.by_eos.entry(eos).or_insert_with(|| Default::default());
+        let class = metrics.by_eos
+            .entry(eos)
+            .or_insert_with(|| EosMetrics::default());
```
hawkw: Code golf, TIOLI: this is functionally equivalent to
```rust
    .or_insert_with(EosMetrics::default);
```

src/telemetry/metrics/record.rs
```diff
@@ -12,13 +12,14 @@ use super::transport;
...
+    transports: transport::Registry,
 }
 
 // ===== impl Record =====
 
 impl Record {
-    pub(super) fn new(metrics: &Arc<Mutex<Root>>) -> Self {
-        Self { metrics: metrics.clone() }
+    pub(super) fn new(metrics: &Arc<Mutex<Root>>, transports: transport::Registry) -> Self {
+        Self { metrics: metrics.clone(), transports }
```
hawkw: Code style nit: now that two fields are being initialized here, this probably ought to be:
```rust
    Self { 
        metrics: metrics.clone(), 
        transports,
    }
```

src/telemetry/mod.rs
```diff
@@ -33,8 +34,10 @@ pub fn new(
     metrics_retain_idle: Duration,
     taps: &Arc<Mutex<tap::Taps>>,
 ) -> (Sensors, tls_config_reload::Sensor, ServeMetrics) {
+    let process = process::Report::new(start_time);
```
hawkw: Was moving the construction of the `process::Report` here necessary for this change? I like the new factoring and I'm fine with landing it in this PR, but I just wanted to make sure that there isn't a semantic difference I'm missing...

src/telemetry/metrics/record.rs
hawkw: Ah, I see that this [changes back again](https://github.com/linkerd/linkerd2-proxy/pull/76/files#diff-d59b7d0e0356005002b7a6173a16b0c2R20) in your next PR, not that big a deal then.

src/telemetry/mod.rs
olix0r: it's not strictly necessary, but now `metrics::new` takes several `Report` instances rather than constructing any reports internally.  This will fit well with subsequent changes.

hawkw: Okay, that's fine with me --- just wanted to make sure I hadn't overlooked something.

PR #76, opened 2018-08-20
Record transport telemetry without Events

olix0r:
Previousy, transport telemetry was recorded by emitting Events from an
IO instance to an aggegator. This requires that each update take a
global telemetry lock, and is an impediment to richer telemetry.

This change removes the transport event types so that the Event and
Record types are left only to represent HTTP telemetry. Now, the
transport's IO type holds a reference to a shared `Metrics` structure.
As the transport is used, metric values are updated immediately.

A lock on the transport _registry_ is taken whenever a new transport is
opened/accepted and when metrics are reported. Each transport class's
metrics are now shared & locked independently, so it's possible for a
transport to update its metrics while the registry is being manipulated.

This has one functional change: the `tcp_read_bytes_total` and
`tcp_write_bytes_total` counters are now updated instantaneously.
Previously these values were only incremented on transport close, which
is misleading, especially for long-lived connections.

With this change, all transport-related telemetry logic lives in
`telemetry::transport`.

Follows #74.


src/inbound.rs
```diff
@@ -114,7 +114,7 @@ mod tests {
     use tls;
 
     fn new_inbound(default: Option<net::SocketAddr>, ctx: ctx::Proxy) -> Inbound<()> {
-        let bind = Bind::new(tls::ClientConfig::no_tls()).with_ctx(ctx);
+        let bind = Bind::new(Default::default(), tls::ClientConfig::no_tls()).with_ctx(ctx);
```
hawkw: Nit: I think`telemetry::transport::Registry::default()` might be clearer here...

src/telemetry/transport/io.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// If the transport operation results in a non-recoverable error, record a
+    /// transport closure.
+    fn sense_err<F, U>(&mut self, op: F) -> io::Result<U>
+    where
+        F: FnOnce(&mut T) -> io::Result<U>,
+    {
+        match op(&mut self.io) {
+            Ok(v) => Ok(v),
+            Err(e) => {
+                if e.kind() != io::ErrorKind::WouldBlock {
```
hawkw: TIOLI: What do you think of
```rust
match op(&mut self.io) {
    Ok(v) => Ok(v),
    Err(e) if e.kind() != io::ErrorKind::WouldBlock => {
        let eos = Eos::Error { errno: e.raw_os_error().map(|e| e.into()) };
        self.sensor.record_close(eos);
        Err(e)
    },
    e => e,
```

src/telemetry/transport/mod.rs
```diff
@@ -32,24 +39,26 @@ pub fn new() -> (Registry, Report) {
...
     proxy: ctx::Proxy,
     peer: Peer,
     tls_status: ctx::transport::TlsStatus,
 }
 
-/// Holds all of the metrics for a class of transport.
+/// Stores a class of transport's metrics.
+///
+/// TODO We should probaby use AtomicUsize for most of these counters so that
+/// simple increments don't require a lock. Especially for read|write_bytes_total.
```
hawkw: 👍 Strongly agree.

src/telemetry/transport/mod.rs
```diff
@@ -194,63 +252,141 @@ impl FmtMetrics for Report {
...
+            if let Ok(mut m) = m.lock() {
+                m.read_bytes_total += sz as u64;
+            }
+        }
+    }
+
+    pub fn record_write(&mut self, sz: usize) {
+        if let Some(ref m) = self.metrics {
+            if let Ok(mut m) = m.lock() {
+                m.write_bytes_total += sz as u64;
```
hawkw: The fact that we're now able to update these stats on every read is great --- I wonder what potential new UI features this could unlock?

src/telemetry/transport/mod.rs
```diff
@@ -194,63 +252,141 @@ impl FmtMetrics for Report {
...
+        if let Some(ref m) = self.metrics {
+            if let Ok(mut m) = m.lock() {
+                m.read_bytes_total += sz as u64;
+            }
+        }
+    }
+
+    pub fn record_write(&mut self, sz: usize) {
+        if let Some(ref m) = self.metrics {
+            if let Ok(mut m) = m.lock() {
```
hawkw: This seems like the place where we'd benefit the most from being able to increment counters without locking; it would sure be nice to not have to take a lock on every read/write.

src/telemetry/transport/io.rs
olix0r: that seems fine, but i don't think it's really better/worse than what's there now.

hawkw: i felt like it made the control flow _slightly_ easier to follow, but 🤷‍♀️

PR #79, opened 2018-08-21
Consolidate HTTP telemetry in telemetry::http

olix0r:
In preparation for further simplifications to HTTP telemetry, this
change consolidates all HTTP-specific logic under the `telemetry::http`
module.

Specifically, the following modules and types have been moved:
- `telemetry::event`;
- `telemetry::metrics::Stamped`;
- `telemetry::metrics::labels`;
- `telemetry::metrics::record`;
- `telemetry::sensors`; and
- `telemetry::sensors::http`.

This change takes pains to avoid changing any implementation details, so
some types and methods have been made public temporarily while the
interface boundaries are not well defined. This will be fixed in a
subsequent change.


src/telemetry/http/mod.rs
```diff
@@ -1,17 +1,30 @@
...
+mod record;
+mod sensors;
+pub mod service;
+
+pub use self::labels::{RequestLabels, ResponseLabels};
+pub use self::record::Record;
+pub use self::sensors::Sensors;
+
+#[derive(Debug)]
+pub(super) struct Stamped<T> {
```
hawkw: Out of curiosity: I'm taking this to mean we don't intend to start expiring non-HTTP metrics any time soon?

src/telemetry/metrics/mod.rs
```diff
@@ -84,9 +68,9 @@ pub fn new(
     process: process::Report,
     transport_report: transport::Report,
     tls: tls_config_reload::Report
-) -> (Record, Serve) {
+) -> (http::Record, Serve) {
     let metrics = Arc::new(Mutex::new(Root::new(process, transport_report, tls)));
-    (Record::new(&metrics), Serve::new(&metrics, idle_retain))
+    (http::Record::new(&metrics), Serve::new(&metrics, idle_retain))
```
hawkw: I'm guessing this will eventually be changed to follow the pattern used by the other metrics record/report types?

olix0r: yeah, though it may take a few more PRs to get there

PR #81, opened 2018-08-22
Move metrics eviction into telemetry::http

olix0r:
The metrics server is responsible for evicting unused metrics, which
seems like an unnecessary coupling with the storage implementation.

This change moves this logic into `telemetry::http` so that the
eviction strategy is specific to the implementation.

Now that the metrics structure is shared internally to `http`,
`Report`'s implementation of `FmtMetrics` can evict expired metrics.


src/telemetry/mod.rs
```diff
@@ -34,7 +34,7 @@ pub fn new(
     taps: &Arc<Mutex<tap::Taps>>,
 ) -> (Sensors, transport::Registry, tls_config_reload::Sensor, ServeMetrics) {
     let process = process::Report::new(start_time);
-    let (http_sensors, http_report) = http::new(taps);
+    let (http_sensors, http_report) = http::new(metrics_retain_idle, taps);
```
hawkw: TIOLI: Maybe the argument ought to be called `http_retain_idle` now to indicate that it only effects the HTTP metrics?

PR #82, opened 2018-08-22
Make telemetry::metrics application-agnostic

olix0r:
Now that transport details have been separated into modules, the
`metrics::Root` type makes more sense as a `telemetry::Report` type.
With this change, the `telemetry::metrics` module holds only the
abstract structural details of metrics reporting.

To this end:
- `metrics::Root` is now `telemetry::Report`
- `metrics::Serve` is now generic over `FmtMetrics`. It's only an
  implementation detail that the `telemetry::Report` type is used.
- all _Report_ types now implement `Clone` so that the main report
  can be cloned for each connection (i.e. from prometheus).


src/telemetry/process.rs
```diff
@@ -11,7 +11,7 @@ metrics! {
     }
 }
 
-#[derive(Debug, Default)]
+#[derive(Clone, Debug, Default)]
```
hawkw: I noticed that the other `Report` types implement `Clone` because they're internally backed by an `Arc`, while this one actually clones the `Report` struct. This isn't a problem currently, since the system stats are collected at formatting time, but it seems like it could potentially be an issue if that behaviour changes in the future?

olix0r: Right. I think the contract is that Reports have to impl Clone. If they have they state internally, they're responsible for making it shared and internally mutable.

PR #83, opened 2018-08-22
Move metrics server into control module

olix0r:
The metrics Service implementation that renders prometheus metrics can
be used independently of any specific listener.

This change moves the binding and listening details into the `control`
module, as this seems like the best umbrella for the specifics of
serving things to the control plane.


olix0r:
There are two motivators here:

1. if we ever want to do any sort of TLS/policy/anything on the metrics server, this would all be controller logic, not metrics formatting.
2. after this, `telemetry::metrics` should have no dependencies on proxy code, so it can be split off into a subcrate

PR #86, opened 2018-08-27
Move `control::Bind` to `svc::NewClient`

olix0r:
The `control::destination` exposes an important trait, `Bind`, that
abstracts the logic of instantiating a new service for an individual
endpoint (i.e., in a load balancer).

This interface is not specific to our service discovery implementation,
and can easily be used to model other types of client factory.

In the spirit of consolidating our HTTP-specific logic, and making the
core APIs of the proxy more visible, this change renames the `Bind`
trait to `NewClient`, simplifies the trait to have fewer type
parameters, and documents this new generalized API.

### DIFF
--- a/src/svc/mod.rs
+++ b/src/svc/mod.rs
@@ -3,23 +3,26 @@
 //! This module contains utilities for proxying request-response streams. This
 //! module borrows (and re-exports) from `tower`.
 //!
+//! ## Clients
+//!
+//! A client is a `Service` through which the proxy may dispatch requests.
+//!
+//! In the proxy, there are currently two types of clients:
+//!
+//! - As the proxy routes requests to an outbound `Destination`, a client
+//!   service is resolves the destination to and load balances requests
+//!   over its endpoints.
+//!
+//! - As an outbound load balancer dispatches a request to an endpoint, or as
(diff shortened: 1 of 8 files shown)


src/svc/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+    /// `Async::Ready`. When `Service::poll_ready` returns an error, the
+    /// client must be discarded.
+    type Client: Service;
+
+    /// Creates a client
+    ///
+    /// If the provided `Target` is valid, immediately return a `Client` that may
+    /// become ready lazily, i.e. as the target is resolved and connections are
+    /// established.
+    fn new_client(&mut self, t: &Self::Target) -> Result<Self::Client, Self::Error>;
```
hawkw: Is there an expected use-case for the change to taking `self` by mut?

hawkw:
Looks like your Travis build flaked out due to one of the TCP tests --- I restarted it.

src/svc/mod.rs
olix0r: It seems likely to me that implementations do mutable things. Specifically, implementors are likely to register metrics scopes, service discovery lookups, etc. I could go either way on this detail (since NewService takes `&self`). `&mut self` seemed the most conservative decision.

PR #87, opened 2018-08-31
Remove redundant reconnect logic

olix0r:
`bind::BoundService` wraps a `Reconnect` service and handles its Connect
errors. However, `BoundService` exposes `Reconnect`'s Error type to
callers even though these errors can never be returned.

Furthermore, `Reconnect` is allowed be polled after returning an error,
triggering the inner service to be rebuilt. We needlessly duplicate this
logic in `BoundService`.

Before splitting this file up into smaller chunks, let's update
`BoundService` to more narrowly adhere to `Reconnect`s API:

- Only the inner error type is returned. `unreachable!` assertions
  have been made where error variants cannot be returned.
- Do not "rebind" the stack explicitly. Instead, let `Reconnect` do
  this.
- Now BoundService::call may panic if invoked before poll_ready. It's a
  programming error, since `Reconnect` requires that `poll_ready` be
  called first.

### DIFF
--- a/src/bind.rs
+++ b/src/bind.rs
@@ -63,6 +63,14 @@ where
     protocol: Protocol,
 }
 
+pub struct ResponseFuture<B>
+where
+    B: tower_h2::Body + Send + 'static,
+    <B::Data as ::bytes::IntoBuf>::Buf: Send,
+{
+    inner: <Stack<B> as tower::Service>::Future,
+}
+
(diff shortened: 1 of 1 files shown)


olix0r:
I've tested this by running the proxy locally with nghttp. The same debouncing and reconnect logic is observed as before this change.

PR #88, opened 2018-09-08
Reorder endpoint-specific HTTP middlewares

olix0r:
Currently, the layered service implementations that comprise the HTTP
stack are a mix of `Service` and `NewService` types. In the
endpoint-specific stack. `transparency::Client` is the only layer that
actually needs to be a `NewService` if it is wrapped immediately with a
`Reconnect`.

This allows us to remove several `NewService` implementations.

This extracts a new `svc::Reconnect` middleware from `bind`, handling
connection error logging and hiding `tower_reconnect::Error` from outer
layers.

Furthermore, two HTTP/1-specific middlewares have been moved outside of
the TLS rebinding layer, since they are not dependent on TLS
configuration.

Finally, `bind`'s type aliases have been simplified, removing the
`HttpRequest` and `HttpResponse` aliases. By removing these, and
removing `transparency::Client`'s dependency on the telemetry body
types, it should be easier to change type signatures going forward.

### DIFF
--- a/src/bind.rs
+++ b/src/bind.rs
@@ -2,22 +2,40 @@ use std::error::Error;
 use std::fmt;
 use std::marker::PhantomData;
 use std::net::SocketAddr;
+use std::sync::Arc;
 
-use futures::{Async, Future, Poll, future, task};
+use futures::Poll;
 use http::{self, uri};
 use tower_service as tower;
 use tower_h2;
-use tower_reconnect::{Reconnect, Error as ReconnectError};
 
 use control::destination::Endpoint;
 use ctx;
-use svc::NewClient;
+use svc::{NewClient, Reconnect};
 use telemetry;
 use transparency::{self, HttpBody, h1, orig_proto};
 use transport;
 use tls;
 use ctx::transport::TlsStatus;
 use watch_service::{WatchService, Rebind};
+
+/// An HTTP `Service` that is created for each `Endpoint` and `Protocol`.
+pub type Stack<B> = orig_proto::Upgrade<NormalizeUri<WatchTls<B>>>;
+
+type WatchTls<B> = WatchService<tls::ConditionalClientConfig, RebindTls<B>>;
+
+/// An HTTP `Service` that is created for each `Endpoint`, `Protocol`, and client
+/// TLS configuration.
+pub type TlsStack<B> = telemetry::http::service::Http<HttpService<B>, B, HttpBody>;
+
+type HttpService<B> = Reconnect<
+    Arc<ctx::transport::Client>,
+    transparency::Client<
+        transport::metrics::Connect<transport::Connect>,
+        ::logging::ClientExecutor<&'static str, SocketAddr>,
+        telemetry::http::service::RequestBody<B>,
+    >
+>;
 
 /// Binds a `Service` from a `SocketAddr`.
 ///
@@ -55,20 +73,8 @@ where
 {
     bind: Bind<ctx::Proxy, B>,
     binding: Binding<B>,
-    /// Prevents logging repeated connect errors.
-    ///
-    /// Set back to false after a connect succeeds, to log about future errors.
-    debounce_connect_error_log: bool,
     endpoint: Endpoint,
     protocol: Protocol,
-}
-
-pub struct ResponseFuture<B>
-where
-    B: tower_h2::Body + Send + 'static,
-    <B::Data as ::bytes::IntoBuf>::Buf: Send,
-{
-    inner: <Stack<B> as tower::Service>::Future,
 }
 
 /// A type of service binding.
@@ -83,7 +89,7 @@ where
     B: tower_h2::Body + Send + 'static,
     <B::Data as ::bytes::IntoBuf>::Buf: Send,
 {
-    Bound(WatchService<tls::ConditionalClientConfig, RebindTls<B>>),
+    Bound(Stack<B>),
     BindsPerRequest {
         // When `poll_ready` is called, the _next_ service to be used may be bound
         // ahead-of-time. This stack is used only to serve the next request to this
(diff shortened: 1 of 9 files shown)


src/bind.rs
```diff
@@ -55,22 +73,10 @@ where
 {
     bind: Bind<ctx::Proxy, B>,
     binding: Binding<B>,
-    /// Prevents logging repeated connect errors.
-    ///
-    /// Set back to false after a connect succeeds, to log about future errors.
-    debounce_connect_error_log: bool,
```
hawkw: Moving this to the reconnect service definitely seems like a good idea to me! 👍

src/bind.rs
```diff
@@ -145,24 +151,6 @@ pub struct RebindTls<B> {
...
 
-pub type Service<B> = BoundService<B>;
-
-pub type Stack<B> = WatchService<tls::ConditionalClientConfig, RebindTls<B>>;
-
-type ReconnectStack<B> = Reconnect<NewHttp<B>>;
-
-pub type NewHttp<B> = orig_proto::Upgrade<NormalizeUri<telemetry::http::service::NewHttp<Client<B>, B, HttpBody>>>;
-
-pub type HttpResponse = http::Response<telemetry::http::service::ResponseBody<HttpBody>>;
```
hawkw: Getting rid of the `HttpRequest`/`HttpResponse` aliases is great --- my experience has been that they make writing code in this module (and code that interacts with this module) more confusing.

In particular, it wasn't immediately clear that the meaning of the aliases was that the response had been wrapped with telemetry sensors, so I suspect newcomers wouldn't have immediately understood when to take a `http::Request` and when to take a `bind::HttpRequest`...

src/bind.rs
```diff
@@ -232,27 +220,24 @@ where
...
-    /// + URI normalization
-    /// + HTTP sensors
+    /// A reconnecting HTTP client is established with the given endpont,
+    /// protocol, and TLS configuration.
     ///
-    /// When the TLS client configuration is invalidated, this function will
-    /// be called again to bind a new stack.
-    fn bind_reconnect_stack(
+    /// This client is instrumented with metrics.
+    fn bind_with_tls(
```
hawkw: TIOLI: I think it would be ideal if the name of this function (and the returned stack type) indicated that this function is responsible for setting up both TLS and reconnects. But, on the other hand, I think `bind_with_tls_and_reconnects` is much too wordy, so I'm not sure if there's a better option here...

src/bind.rs
```diff
@@ -273,54 +258,45 @@ where
...
-
-        // Automatically perform reconnects if the connection fails.
-        //
-        // TODO: Add some sort of backoff logic.
-        Reconnect::new(upgrade_orig_proto)
-    }
-
-    /// Binds the endpoint stack used to construct a bound service.
+        // TODO: Add some sort of backoff logic between reconnects.
+        self.sensors.http(
```
hawkw: Nit/TIOLI: Personally, I find this much harder to read than the way it was done previously --- I think something like
```rust
let client = transparency::Client::new(protocol, connect, log.executor());
let reconnect = Reconnect::new(client_ctx.clone(), client);
self.sensors.http(client_ctx.clone(), reconnect)
```
This doesn't require me to track the parentheses in my head, and is actually fewer lines...

src/bind.rs
```diff
@@ -273,54 +258,45 @@ where
...
+    /// with prior knowledge that the endpoint supports HTTP/2).
     ///
-    /// This function will itself be called again by `BoundService` if the
-    /// service binds per request, or if the initial connection to the
-    /// endpoint fails.
+    /// As `tls_client_config` updates, `bind_with_tls` is called to rebuild the
+    /// client with the appropriate TLS configuraiton.
     fn bind_stack(&self, ep: &Endpoint, protocol: &Protocol) -> Stack<B> {
         debug!("bind_stack: endpoint={:?}, protocol={:?}", ep, protocol);
-        // TODO: Since `BindsPerRequest` bindings are only used for a
```
hawkw: Was this TODO comment resolved?

src/bind.rs
```diff
@@ -385,58 +360,18 @@ where
...
-    type Error = <Self::Service as tower::Service>::Error;
-    type Service = NormalizeUri<S::Service>;
-    type InitError = S::InitError;
-    type Future = future::Map<
-        S::Future,
-        fn(S::Service) -> NormalizeUri<S::Service>
-    >;
-    fn new_service(&self) -> Self::Future {
-        let s = self.inner.new_service();
-        // This weird dance is so that the closure doesn't have to
```
hawkw: Glad this code could finally be gotten rid of :)

src/bind.rs
```diff
@@ -485,81 +420,17 @@ where
...
             }
-        };
-
-        // If there was a connect error, don't terminate this BoundService
-        // completely. Instead, simply clean up the inner service, prepare to
-        // make a new one, and tell our caller that we could maybe be ready
-        // if they call `poll_ready` again.
-        //
-        // If they *don't* call `poll_ready` again, that's ok, we won't ever
-        // try to connect again.
```
hawkw: It never made sense to me that this code was in the `NormalizeUri` service --- moving it to the new reconnect service is great.

src/svc/reconnect.rs
```diff
@@ -0,0 +1,122 @@
...
+    inner: <tower_reconnect::Reconnect<N> as Service>::Future,
+}
+
+// ===== impl Reconnect =====
+
+impl<T, N> Reconnect<T, N>
+where
+    T: fmt::Debug,
+    N: NewService,
+    N::InitError: fmt::Display,
```
hawkw: TIOLI: Perhaps this bound should be on the struct as well, since the struct has the `T: fmt::Debug` and `N: NewService` bounds already?

src/bind.rs
olix0r: There's a shorter todo below

src/bind.rs
olix0r: i'm trying to allude to how this might evolve to a pattern more like the following:

``
sensors
    .and_then(reconnect)
    .and_then(client)
    .make_client(&client_ctx)
```

This will change a lot in the near future.

PR #90, opened 2018-09-11
Introduce a "server stack"

olix0r:
Previously, `proxy::Server` was generic over a `NewService` (constructed
in `lib.rs`) that instruments error handling around the router and metrics. In
preparation of adding a metrics module into the server stack (that is configured
by the source connection), `Server` should be changed to instantaneously build
clients with a `MakeClient<Arc<ctx::transport::Server>>`.

In order to do this, the following changes were made:

1. The `svc::NewClient` type was changed to `svc::MakeClient<Target>`. The
    naming change ("New" to "Make") is intended to differentiate the type from
    `NewService`, which is asynchronous and does not accept a `Target` argument.
2. The `proxy::h2_router` module was split from `lib.rs` and `map_err.rs`.  `MapErr`
    tried to be generic, though we only used it in once place. Now, the `h2_router::Make`
    type supports cloning routers and handling their errors.
3. The `TimestampRequestOpen` middleware was split into its own file and given a
    `MakeClient` implementation.
4. The `svc::Layer` trait has been introduced to support layering middlewares like
    `TimestampRequestOpen`. This is analogous to Finagle's `Stack.Module` type.


src/svc/layer.rs
```diff
@@ -0,0 +1,49 @@
+use std::marker::PhantomData;
+
+/// A stackable element.
+///
+/// Given a `Next`-typed inner value, produces a `Bound`-typed value.
+/// This is especially useful for composable types like `MakeClient`s.
+pub trait Layer<Next> {
```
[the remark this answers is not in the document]
olix0r: No, it's not really a config in this case.  The general idea is that we can use this to compose middlewares -- so that both the `Next` and `Bound` types implement `MakeClient`, for instance (see https://github.com/linkerd/linkerd2-proxy/blob/132be6bc31e7953cc55d7d2fa30d73e9bcbc51d1/src/telemetry/http/timestamp_request_open.rs#L64-L90).  `Layer` lets us implement a `MakeClient` in terms of another MakeClient`.

I agree that this isn't 100% obvious right now, but, from experience, I think this will be useful as we expand its usage.

src/lib.rs
```diff
@@ -433,39 +429,16 @@ where
...
-            }
-        });
 
-        // Install the request open timestamp module at the very top
-        // of the stack, in order to take the timestamp as close as
-        // possible to the beginning of the request's lifetime.
-        telemetry::http::service::TimestampRequestOpen::new(map_err)
-    }));
+
+    // Install the request open timestamp module at the very top of the
```
hawkw: Nit: Since we're using "layer" for what Finagle calls "module", maybe s/module/layer in the comments?

src/proxy/server.rs
```diff
@@ -27,44 +27,40 @@ use super::tcp;
...
-    B: tower_h2::Body + 'static,
-    S::Error: Error + Send + Sync + 'static,
-    S::InitError: Send + fmt::Debug,
-    B: tower_h2::Body + Send + Default + 'static,
+        Response = http::Response<B>,
+    >,
+    M::Client: Send + 'static,
+    <M::Client as Service>::Error: error::Error + Send + Sync + 'static,
+    <M::Client as Service>::Future: Send + 'static,
+    B: tower_h2::Body + Default + Send + 'static,
```
hawkw: Whoa, does this change actually _reduce_ the number of bounds on `Server`? Nice!

src/proxy/server.rs
```diff
@@ -153,69 +145,73 @@ where
             return log.future(Either::B(fut));
         }
 
-        // try to sniff protocol
+        let detect_protocol = io.peek()
```
hawkw: Breaking this out is great!

src/svc/layer.rs
```diff
@@ -0,0 +1,49 @@
...
+/// Combines two `Layers` into one layer.
+#[derive(Debug, Clone)]
+pub struct AndThen<Next, Outer, Inner>
+where
+    Outer: Layer<Inner::Bound>,
+    Inner: Layer<Next>,
+{
+    outer: Outer,
+    inner: Inner,
+    _p: PhantomData<Next>,
```
hawkw: TIOLI: perhaps this should be `PhantomData<fn() -> Next>` or something, so that `AndThen` being `Send` or `Sync` isn't dependent on `Next` being `Send` or `Sync`?

src/svc/layer.rs
```diff
@@ -33,7 +33,8 @@ where
 {
     outer: Outer,
     inner: Inner,
-    _p: PhantomData<Next>,
+    // `AndThen` should be Send/Sync independently of `Next`.
+    _p: PhantomData<fn() -> Next>,
```
hawkw: 👍

PR #91, opened 2018-09-12
Split proxy infrastructure from `bind`

olix0r:
The `bind` module includes core types that are used throughout the proxy
module, like `bind::Protocol`. Furthermore, this "protocol" is not really
related to the protocol detection that occurs in `proxy`.

In order to simplify `bind` and set up for further refactoring (adopting
`MakeClient`), this change reorganizes the module structure as follows:

- `bind::Protocol` is now `proxy::http::Dialect`. This change is made to
   distinguish this concept from protocol detection and, furthermore, to
   separate this logic from endpoint-construction implemented in `bind`.

- HTTP proxying utilities have been moved into the `proxy::http` module.

- `bind::NormalizeUri` is now `proxy::http::NormalizeUri`

- `proxy::h2_router` is now `proxy::http::router`

### DIFF
--- a/src/proxy/http/dialect.rs
+++ b/src/proxy/http/dialect.rs
@@ -1,0 +1,115 @@
+use http;
+
+use proxy::http::h1;
+
+/// Describes the required HTTP "dialect".
+///
+/// This marks whether to use HTTP/2 or HTTP/1.x for a request. In
+/// the case of HTTP/1.x requests, it also stores a "host" key to ensure
+/// that each host receives its own connection.
+///
+/// TODO HTTP2 should indicate whether the requester supports PUSH_PROMISE.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum Dialect {
+    Http1 {
+        host: Host,
+        /// Whether the request wants to use HTTP/1.1's Upgrade mechanism.
+        ///
+        /// Since these cannot be translated into orig-proto, it must be
+        /// tracked here so as to allow those with `is_h1_upgrade: true` to
+        /// use an explicitly HTTP/1 service, instead of one that might
+        /// utilize orig-proto.
+        is_h1_upgrade: bool,
+        /// Whether or not the request URI was in absolute form.
+        ///
+        /// This is used to configure Hyper's behaviour at the connection
+        /// level, so it's necessary that requests with and without
+        /// absolute URIs be bound to separate service stacks. It is also
+        /// used to determine what URI normalization will be necessary.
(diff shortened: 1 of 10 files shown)


src/proxy/http/dialect.rs
```diff
@@ -0,0 +1,115 @@
...
+
+/// Describes the required HTTP "dialect".
+///
+/// This marks whether to use HTTP/2 or HTTP/1.x for a request. In
+/// the case of HTTP/1.x requests, it also stores a "host" key to ensure
+/// that each host receives its own connection.
+///
+/// TODO HTTP2 should indicate whether the requester supports PUSH_PROMISE.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum Dialect {
```
hawkw: TIOLI: perhaps calling this `http::Version` would be more descriptive? IDK...

olix0r: I avoided `Version` since that overloads the terminology from `http::Version` -- which is related but not really what we're trying to describe here: the Dialect is effectively a client configuration. We might construct two HTTP/1 stacks to a single endpoint that differ only in terms of whether we need absolute_uri normalization.  They share an `http::Version`, but are two distinct HTTP "dialects".

hawkw: Okay, fair enough!

src/proxy/http/dialect.rs
```diff
@@ -0,0 +1,115 @@
+use http;
+
+use proxy::http::h1;
+
+/// Describes the required HTTP "dialect".
+///
+/// This marks whether to use HTTP/2 or HTTP/1.x for a request. In
+/// the case of HTTP/1.x requests, it also stores a "host" key to ensure
+/// that each host receives its own connection.
```
briansmith: Besides the "host" key other attributes are stored, so this documentation is misleading by omission.

It seems like this type is used as part of the key that determines what connection to route a request to, based on the HTTP version and other version-related properties of the request. I think the documentation should say something like that.

olix0r:
After sleeping on this, I also don't like the word _Dialect_ and don't think it should be used here. It's really something like a `ClientConfig`, because it literally what configures a `Client`...  I think I'm going to try to back out the Protocol change for now and do that in a followup

olix0r:
https://github.com/linkerd/linkerd2-proxy/pull/93

src/proxy/http/dialect.rs
hawkw: @briansmith That's more or less how I feel about it as well...

PR #92, opened 2018-09-13
Add `Server` ctx to requests via a `Layer`

olix0r:
`proxy::Server` is currently responsible for constructing a
`ctx::transport::Server`, building client stacks, and, additionally,
ensuring that the ctx object is added as an extension to all requests.

This change factors out extension-setting from `proxy::Server`. this
may facilitate adding similar metadata into other stacks, and also
serves to illustrate the usefulness of `svc::Layer` (where the layers
are composed in `lib.rs`).


olix0r:
This will be resubmitted in another PR.

PR #93, opened 2018-09-13
Move HTTP-specific `proxy` modules into proxy::http

olix0r:
To prepare to move http-specific proxying logic from bind.rs into
`proxy`, let's carve out a `proxy::http` module for HTTP-specific behavior.


hawkw:
Based on the [build failure](https://travis-ci.org/linkerd/linkerd2-proxy/builds/428267662?utm_source=github_status&utm_medium=notification) it looks like there's an import of `proxy::h2_router` in `Inbound` that needs to be pointed at `proxy::http::router`?

src/proxy/http/router.rs
```diff
@@ -22,7 +22,7 @@ where
     router: Router<R>,
 }
 
-pub struct H2Router<R>
+pub struct RouterService<R>
```
hawkw: TIOLI: why not just `Service`, so it can be referred to as `router::Service`?

src/proxy/http/glue.rs
```diff
@@ -35,13 +35,13 @@ pub enum HttpBody {
 
 /// Glue for `tower_h2::Body`s to be used in hyper.
 #[derive(Debug, Default)]
-pub(super) struct BodyPayload<B> {
+pub struct BodyPayload<B> {
```
hawkw: Do these need to be `pub`, or can they just be `pub(in proxy::http)` or something?

olix0r: they can probably be `pub(in proxy)` (i wasn't aware of the _in_ syntax...)

PR #94, opened 2018-09-13
Move bind::NormalizeUri to its own module in proxy::http

olix0r:
Split proxy infrastructure from `bind`

In order to simplify `bind.rs`, this change moves `NormalizeUri` into
its own module, `proxy::http::normalize_uri`.

This will later become part of the "client stack" built from
`proxy::http`.


olix0r:
Originally part of #91

PR #99, opened 2018-10-02
Deprecate PRIVATE/PUBLIC config variables

olix0r:
There are a few crufty remaining references to the `private` and
`public` proxy interfaces. `private` referred to the pod-local side of
the proxy; and `public` the external side.

These terms have been replaced by `inbound` and `outbound` proxies.

The "private forward address" is now the "inbound forward address".
The "private listener" is now the "outbound listener".
The "public listener" is now the "inbound listener".

This change adds new environment variables to support configuration:
- `LINKERD2_PROXY_INBOUND_CONNECT_TIMEOUT`
- `LINKERD2_PROXY_INBOUND_FORWARD`
- `LINKERD2_PROXY_INBOUND_LISTENER`
- `LINKERD2_PROXY_OUTBOUND_CONNECT_TIMEOUT`
- `LINKERD2_PROXY_OUTBOUND_LISTENER`

The old configuration variables have not been removed. Instead, a
warning is logged when such a variable is used to configure the proxy.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -10,9 +10,9 @@ use http;
 use indexmap::IndexSet;
 use trust_dns_resolver::config::ResolverOpts;
 
-use transport::{Host, HostAndPort, HostAndPortError, tls};
-use convert::TryFrom;
 use conditional::Conditional;
+use convert::TryFrom;
+use transport::{Host, HostAndPort, HostAndPortError, tls};
 
 // TODO:
 //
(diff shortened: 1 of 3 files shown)


olix0r:
```
#; golly ~/b/l2-proxy  (ver/config-naming ✔)
:;
LINKERD2_PROXY_LOG=linkerd2_proxy=trace,linkerd2_stack=debug \
LINKERD2_PROXY_POD_NAMESPACE=ns \
LINKERD2_PROXY_PRIVATE_FORWARD=tcp://127.0.0.1:4191 \
LINKERD2_PROXY_CONTROLLER_NAMESPACE=linkerd \
cargo run

    Finished dev [unoptimized + debuginfo] target(s) in 0.15s
WARN linkerd2_proxy::config LINKERD2_PROXY_PRIVATE_FORWARD has been deprecated; use LINKERD2_PROXY_INBOUND_FORWARD
```

briansmith:
> This change adds new environment variables to support configuration:
> 
> * `LINKERD2_PROXY_INBOUND_CONNECT_TIMEOUT`
> * `LINKERD2_PROXY_INBOUND_FORWARD`
> * `LINKERD2_PROXY_INBOUND_LISTENER`
> * `LINKERD2_PROXY_OUTBOUND_CONNECT_TIMEOUT`
> * `LINKERD2_PROXY_OUTBOUND_LISTENER`

I think we should not be bulk-adding these settings. Instead, each setting should be added with a justification for why it is being added. Some of these timeout settings were things we used to have and we explicitly chose to remove because (at that time, at least) we thought timeouts were the wrong thing to do there. If it is too hard to detangle the addition of the new settings from the renaming then at least let's amend the PR description with the justifications for each of the new settings.

olix0r:
@briansmith 

> Some of these timeout settings were things we used to have and we explicitly chose to remove

I don't follow. This change doesn't add any settings. It creates proper alternatives for all of the existing settings that are incorrectly named.

PR #100, opened 2018-10-05
refactor: Structure the proxy in terms of `Stack`

olix0r:
As the proxy's functionality has grown, the HTTP routing functionality
has become complex. Module boundaries have become ill-defined, which
leads to tight coupling--especially around the `ctx` metadata types and
`Service` type signatures.

This change introduces a `Stack` type (and subcrate) that is used as the
base building block for proxy functionality. The `proxy` module now
exposes generic components--stack layers--that are configured and
instantiated in the `app::main` module.

This change reorganizes the repo as follows:
- Several auxiliary crates have been split out from the `src/` directory
  into `lib/fs-watch`, `lib/stack` and `lib/task`.
- All logic specific to configuring and running the linkerd2 sidecar
  proxy has been moved into `src/app`. The `Main` type has been moved
  from `src/lib.rs` to `src/app/main.rs`.
- The `src/proxy` has reusable, generic components useful for building
  proxies in terms of `Stack`s.

The logic contained in `lib/bind.rs`, pertaining to per-endpoint service
behavior, has almost entirely been moved into `app::main`.

`control::destination` has changed so that it is not responsible for
building services. (It used to take a clone of `Bind` and use it to
create per-endpoint services). Instead, the destination service
implements the new `proxy::Resolve` trait, which produces an infinite
`Resolution` stream for each lookup. This allows the `proxy::balance`
module to be generic over the servie discovery source.

Furthermore, the `router::Recognize` API has changed to only expose a
`recgonize()` method and not a `bind_service()` method. The
`bind_service` logic is now modeled as a `Stack`.

The `telemetry::http` module has been replaced by a
`proxy::http::metrics` module that is generic over its metadata types
and does not rely on the old telemetry event system. These events are
now a local implementation detail of the `tap` module.

There are no user-facing changes in the proxy's behavior.

---

This is, indeed, a very large change. Much larger than I'd normally like to
submit. So large that it's probably not feasible to review this as a straight
diff. However, it's important that we move forward with this in the short
term to unblock linkerd/linkerd2#1418.

I recommend checking out this branch and getting some hands-on experience
before commenting on it. I'm happy to do a live walk-through of the change if
others would find that useful. I'm especially looking for feedback on what
needs clarification/explanation in order for others to be effective in the
new Stack system.

This change has been tested fairly extensively: the repo's tests pass; the
linkerd2 repo's integration tests pass; and our internal performance tests do
not indicate any regressions.

Some metrics-related tests were removed as part of this refactor. It should
be much easier to write tests around the `proxy::http::metrics` module now,
though. I have manually verified that HTTP metrics expire as before.


lib/fs-watch/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "linkerd2-fs-watch"
+version = "0.1.0"
+authors = ["Oliver Gould <ver@buoyant.io>"]
```
hawkw: _Technically_, this should be me 😛

olix0r: Good catch! This was cargo --new ;)

src/app/config.rs
```diff
@@ -10,13 +10,13 @@ use http;
...
-use conditional::Conditional;
-use convert::TryFrom;
 use transport::{Host, HostAndPort, HostAndPortError, tls};
+use convert::TryFrom;
+use Conditional;
 
 // TODO:
 //
-// * Make struct fields private.
+// * Stack struct fields private.
```
seanmonstar: Was this a simple replacement of `s/Make/Stack`? :D

lib/fs-watch/Cargo.toml
```diff
@@ -1,7 +1,10 @@
 [package]
 name = "linkerd2-fs-watch"
 version = "0.1.0"
-authors = ["Oliver Gould <ver@buoyant.io>"]
+authors = [
+    "Eliza Weisman <eliza@buoyant.io>",
```
hawkw: Probably out of scope for this branch, but I think it it _might_ be a good idea to add people to the code owners file for some of the crates that are being factored out, so they're automatically added to PRs touching the code they're most familiar with?

olix0r: sounds reasonable but not in scope

src/app/destination.rs
```diff
@@ -0,0 +1,123 @@
...
+    http::{h1, Settings},
+    Source,
+};
+use transport::{DnsNameAndPort, Host, HostAndPort};
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Destination {
+    pub name_or_addr: NameOrAddr,
+    pub settings: Settings,
+    _p: (),
```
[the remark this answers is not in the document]
olix0r: This is a pretty common idiom to my understanding (for example https://github.com/hyperium/http/blob/master/src/request.rs#L181). It ensures that the type can't be constructed outside of this module and ensures that the type is not used in an exhaustive match situation.

lib/timeout/src/lib.rs
```diff
@@ -1,12 +1,11 @@
 extern crate futures;
 extern crate tokio_connect;
 extern crate tokio_timer;
-extern crate tower_service;
+extern crate tower_service as svc;
```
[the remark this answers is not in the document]
olix0r: This is just for consistency and namespacing.  Having stamped out many of these modules, it's way easier to understand if each type is named after its role.  E.g. `proxy::reconnect::{Layer, Stack, Service}` rather than `proxy::reconnect::{ReconnectLayer, ReconnectStack, ReconnectService}`, etc.  I find it way easier to keep track of what's going on if we always `use svc` and explicitly reference `svc::Service`, `svc::Stack`, etc

src/app/main.rs
```diff
@@ -0,0 +1,545 @@
...
+            // Establishes connections to remote peers.
+            let connect = transport_metrics
+                .connect("outbound")
+                .and_then(proxy::timeout::Layer::new(config.outbound_connect_timeout))
+                .bind(connect::Stack::new());
+
+            // As HTTP requests are accepted, we add some request extensions
+            // including metadata about the request's origin.
+            let source_layer =
+                timestamp_request_open::Layer::new().and_then(insert_target::Layer::new());
```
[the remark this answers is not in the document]
olix0r: Yeah, i think the free function approach is probably the best option.

src/ctx/transport.rs
```diff
@@ -33,36 +29,9 @@ pub struct Client {
     pub proxy: ctx::Proxy,
     pub remote: SocketAddr,
     pub metadata: destination::Metadata,
-    pub tls_status: TlsStatus,
```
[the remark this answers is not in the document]
olix0r: At this point, the `ctx` modules are _only_ used by tap. We'll be able to do away with them entirely in a followup.

src/app/destination.rs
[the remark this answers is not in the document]
hawkw: Ah, yeah, that makes sense, and I had _figured_ it was something like that. I think it might be marginally clearer if it were called "_priv" or "_private", but I don't have a problem with the current naming.

src/app/main.rs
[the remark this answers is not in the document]
hawkw: Yeah, I took a quick pass at at blocking out what a trait that bundles together a stack and layer type would look like, and I think it may be more trouble than it's worth.

lib/stack/src/layer.rs
```diff
@@ -31,22 +32,6 @@ pub trait Layer<T, U, M: super::Stack<U>> {
             _p: PhantomData,
         }
     }
-
-    /// Conditionally compose this `Layer with another.
-    fn and_when<P, N, L>(self, predicate: P, inner: L)
```
hawkw: Since this method (and `Optional`) weren't being used, I think removing them was a good call. But it's good to know that they're possible if we end up needing them.

lib/stack/src/layer.rs
```diff
@@ -1,14 +1,15 @@
...
-
 /// A `Layer` wraps `M`-typed `Stack<U>` to produce a `Stack<T>`
 ///
-/// Some `Stack` types are dependendent on an generic over an inner `Stack`. For
+/// Some `Stack` types are dependendent on and generic over an inner `Stack`. For
 /// example, a load balancer may implement `Stack<Authority>` and be
 /// configured with a `Stack<SocketAddr>` that is used to build a service for
-/// each enpdoint. Such a load balancer would provide an `impl<M: Stack<SocketAddr>>
-/// Layer<Authority, SocketAddr, M > for BalanceLayer<M>`.
+/// each enpdoint. Such a load balancer would provide an implemntation like:
```
hawkw: Typo: "implemntation"

lib/stack/src/layer.rs
```diff
@@ -39,20 +39,16 @@ pub trait Layer<T, U, M: super::Stack<U>> {
...
-    Outer: Layer<T, U, Inner::Stack>,
-    Inner: Layer<U, V, M>,
-    M: super::Stack<V>,
+pub struct AndThen<U, Outer, Inner>
 {
     outer: Outer,
     inner: Inner,
-    // `AndThen` should be Send/Sync independently of `M`.
-    _p: PhantomData<fn() -> (T, U, V, M)>,
+    // `AndThen` should be Send/Sync independently of U M`.
```
hawkw: Nit: I think the `m` in this comment should be removed (that's my fault).

PR #101, opened 2018-10-08
Decouple reconnect & connect layers from proxy::http::client

olix0r:
Previously, the `client` module was responsible for instrument reconnects. Now, the reconnect module becomes its own stack layer that composes over NewService stacks.

Additionally, the `proxy::http::client` module can now layer over an underlying Connect stack.

---
This sets up the controller client to become a Stack.

### DIFF
--- a/src/proxy/reconnect.rs
+++ b/src/proxy/reconnect.rs
@@ -1,15 +1,26 @@
-use std::fmt;
-
 use futures::{task, Async, Future, Poll};
+use std::fmt;
+use std::marker::PhantomData;
 use tower_reconnect;
 
 use svc;
+
+#[derive(Debug)]
+pub struct Layer<T, M> {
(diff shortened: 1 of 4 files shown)


src/proxy/reconnect.rs
```diff
@@ -1,15 +1,26 @@
...
-
 use futures::{task, Async, Future, Poll};
+use std::fmt;
+use std::marker::PhantomData;
 use tower_reconnect;
 
 use svc;
 
+#[derive(Debug)]
+pub struct Layer<T, M> {
```
hawkw: Wasn't sure if it was necessary for all the types in this module to be generic over both `T` and `M`, but it seems like it is.

olix0r: Yeah, I tried and failed to remove it :man_shrugging:

PR #102, opened 2018-10-10
Adopt latest non-breaking version of proxy dependencies

hawkw:
Depends on #103.

This branch bumps all the proxy's main dependencies to the latest
version that doesn't include a breaking change. The latest
`tokio-rustls` is a breaking change that may require more code changes,
so I'm leaving that for a future PR. 

The only changes to the proxy codebase itself were fixes for deprecation
warnings introduced by the dependency upgrade.


olix0r:
Let's only update trustdns for now and update the other groups of dependencies separately so we have more time to find regressions.

hawkw:
@olix0r I've opened #103 with just the minimal changes necessary to depend on `trust-dns-resolver` from crates.io. I'll rebase this branch on top of #103 and we can test it after the edge release.

seanmonstar:
Closing as inactive.

PR #104, opened 2018-10-14
Make a generic Watch stack from outbound::tls_config

olix0r:
The TLS-configuration-watching logic in `app::outbound::tls_config` need
not be specific to the outbound types, or even TLS configuration.

Instead, this change extends the `watch` stack module with a Stack type
that can satisfy the TLS use case independently of the concrete types at
play.

`app::outbound` now need only implement:

    svc::watch::WithUpdate<tls::ClientConnectionConfig> for Endpoint

### DIFF
--- a/lib/stack/src/watch.rs
+++ b/lib/stack/src/watch.rs
@@ -1,15 +1,35 @@
 extern crate futures_watch;
 
+use self::futures_watch::Watch;
 use futures::{future::MapErr, Async, Future, Poll, Stream};
 use std::{error, fmt};
 
 use svc;
 
+/// Implemented by targets that can be updated by a `Watch<U>`
+pub trait WithUpdate<U> {
+    type Updated;
+
+    fn with_update(&self, update: &U) -> Self::Updated;
+}
+
+#[derive(Debug)]
+pub struct Layer<U> {
+    watch: Watch<U>,
+}
+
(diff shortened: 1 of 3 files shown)


lib/stack/src/watch.rs
```diff
@@ -66,6 +140,51 @@ where
     }
 }
 
+impl<U, M> Service<CloneUpdate, U, M>
+where
+    U: Clone,
+    M: super::Stack<U>,
+    M::Value: svc::Service,
+{
+    pub fn try(watch: Watch<U>, stack: M) -> Result<Self, M::Error> {
```
hawkw: Might be nice if there was a doc comment explaining what this is for...

lib/stack/src/watch.rs
```diff
@@ -66,6 +140,51 @@ where
...
+    T: WithUpdate<U> + Clone,
+    M: super::Stack<T::Updated> + Clone,
+    M::Value: svc::Service + Clone,
+{
+    fn clone(&self) -> Self {
+        Self {
+            inner: self.inner.clone(),
+            watch: self.watch.clone(),
+            stack: self.stack.clone(),
+            target: self.target.clone(),
```
hawkw: This seems like it should be derive-able?

PR #105, opened 2018-10-14
Introduce an optional fixed backoff between reconnects

olix0r:
The control client implements a backoff service that dampens reconnect
attempts to the control plane by waiting a fixed period of time after a
failure.

Furthermore, the control client logs errors each time a reconnect
attempt fails.

This change moves backoff logic from
`control::destination::background::client` to `proxy::reconnect`.

Because the reconnect module handles connection errors uniformly, muting
repeated errors, it also has enough context to know when a backoff
should be applied -- when the underlying NewService cannot produce a
Service.

If polling the inner service fails once the Service has been
established, we do _not_ want to apply a backoff, since this may
just be the result of a connection being terminated, a process being
restarted, etc.

### DIFF
--- a/src/control/destination/background/client.rs
+++ b/src/control/destination/background/client.rs
@@ -1,30 +1,25 @@
-use std::{
-    fmt,
-    io,
-    time::{Duration, Instant},
-};
+use std::time::Duration;
 
 use bytes::Bytes;
-use futures::{Async, Future, Poll};
+use futures::Poll;
 use h2;
 use http;
-use tokio::timer::Delay;
 
 use tower_h2::{self, BoxBody, RecvBody};
 use tower_add_origin::AddOrigin;
-use tower_reconnect::{
-    Reconnect,
-    Error as ReconnectError,
-    ResponseFuture as ReconnectFuture,
-};
 use Conditional;
 use dns;
+use proxy;
 use svc;
-use timeout::{Timeout, Error as TimeoutError};
+use timeout::Timeout;
 use transport::{tls, HostAndPort, LookupAddressAndConnect};
 
 /// Type of the client service stack used to make destination requests.
-pub(super) struct ClientService(AddOrigin<Backoff<LogErrors<Reconnect<
+pub(super) struct ClientService(Service);
+
+type Service = AddOrigin<
+    proxy::reconnect::Service<
+        HostAndPort,
         tower_h2::client::Connect<
             Timeout<LookupAddressAndConnect>,
             ::logging::ContextualExecutor<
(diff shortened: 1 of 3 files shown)


src/proxy/reconnect.rs
```diff
@@ -125,19 +179,35 @@ where
...
+                        Ok(Async::NotReady) => return Ok(Async::NotReady),
+                        Ok(Async::Ready(())) => {},
+                        Err(e) => {
+                            error!("timer failed; continuing without backoff: {}", e);
+                        }
+                    }
+                }
+            }
+        };
+        self.active_backoff = None;
```
seanmonstar: This could go up in the match arm of `Backoff::Fixed(_)`, so it doesn't need to write if it's `None`. Something something caches...

src/proxy/reconnect.rs
```diff
@@ -149,6 +219,15 @@ where
                     debug!("connect error to {:?}: {}", self.target, err);
                 }
 
+                // Set a backoff if appropriate.
+                //
+                // This future need not be polled immediately because the
+                // task is notified below.
+                self.active_backoff = match self.backoff {
+                    Backoff::None => None,
+                    Backoff::Fixed(ref wait) => Some(Delay::new(clock::now() + *wait)),
```
seanmonstar: Since `Duration` is `Copy`, this doesn't need to be a `ref` and de-reference, just `BackoffFixed(wait)`.

src/control/destination/background/client.rs
```diff
@@ -46,54 +42,24 @@ pub(super) struct BindClient {
...
-                    HostAndPort,
-                >
-            >, BoxBody>>;
+    type Error = <Service as svc::Service>::Error;
+    type Future = <Service as svc::Service>::Future;
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
-        self.0.poll_ready()
+        // If an error occurs on established connection, log it; otherwise the `reconnect` module will handle re-establishing a client.
+        loop {
```
hawkw: For a minute I thought this might be better expressed with a `while let` loop, but realized that wouldn't bind the `Ok` case as easily, so nevermind.

src/proxy/reconnect.rs
```diff
@@ -125,19 +179,35 @@ where
     type Future = ResponseFuture<N>;
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        match  self.backoff {
```
hawkw: Nit: looks like there's two spaces here?

src/proxy/reconnect.rs
```diff
@@ -187,8 +266,97 @@ impl<N: svc::NewService> Future for ResponseFuture<N> {
...
         self.inner.poll().map_err(|e| match e {
-            tower_reconnect::Error::Inner(err) => err,
+            Error::Inner(err) => err,
             _ => unreachable!("response future must fail with inner error"),
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
```
hawkw: Thanks for adding tests! 👍

PR #106, opened 2018-10-14
Make `resolve` a stack module

olix0r:
The `proxy::http::balance` module uses the `proxy::resolve::Resolve`
trait to implement a `Discover`.

This coupling between the balance and resolve modules prevents
integrating the destination profile API such that there is a per-route,
per-endpoint stack.

This change makes the `balance` stack generic over a stack that produces
a `Discover`. The `resolve` module now implements a stack that produces
a `Discover` and is generic over a per-endpoint stack.

### DIFF
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -20,9 +20,7 @@ use logging;
 use metrics;
 use proxy::{
     self, buffer,
-    http::{
-        balance, client, insert_target, metrics::timestamp_request_open, normalize_uri, router,
-    },
+    http::{client, insert_target, metrics::timestamp_request_open, normalize_uri, router},
     limit, reconnect, timeout,
 };
 use svc::{self, Layer as _Layer, Stack as _Stack};
@@ -223,7 +221,11 @@ where
         let (drain_tx, drain_rx) = drain::channel();
 
         let outbound = {
-            use super::outbound;
+            use super::outbound::{discovery::Resolve, orig_proto_upgrade, tls_config, Recognize};
+            use proxy::{
+                http::{balance, metrics},
+                resolve,
+            };
 
             let http_metrics = http_metrics.clone();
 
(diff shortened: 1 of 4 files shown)


src/app/main.rs
```diff
@@ -252,31 +254,25 @@ where
...
-                    resolver,
-                )))
-                .and_then(outbound::orig_proto_upgrade::Layer::new())
-                .and_then(outbound::tls_config::Layer::new(tls_client_config))
-                .and_then(proxy::http::metrics::Layer::new(
-                    http_metrics,
-                    classify::Classify,
-                ))
+                .and_then(balance::layer())
+                .and_then(resolve::layer(Resolve::new(resolver)))
```
hawkw: Very minor, but: 👍 on changing these to free functions! definitely more readable!

src/lib.rs
```diff
@@ -2,6 +2,9 @@
 #![cfg_attr(feature = "cargo-clippy", allow(new_without_default_derive))]
 #![deny(warnings)]
 
+// Stack type inference requires a deeper recursion limit
+#![recursion_limit="128"]
```
hawkw: I wonder if breaking up the router stack into a series of separate `let` bindings would help with this?

src/proxy/http/balance.rs
```diff
@@ -2,69 +2,40 @@ extern crate tower_balance;
...
-/// build a service for each endpoint.
-pub struct Service<R: Resolution, M: svc::Stack<R::Endpoint>> {
-    resolution: R,
-    make: M,
 }
 
 // === impl Layer ===
 
-impl<T, R, M, A, B> Layer<T, R, M>
-where
```
hawkw: awesome to see all these type parameters just go away! :D

src/lib.rs
olix0r: this will be removed in a followup

PR #111, opened 2018-10-27
refactor: Build stacks from bottom-to-top

olix0r:
Previously, stacks were built with `Layer::and_then`. This pattern
severely impacts compile-times as stack complexity grows.

In order to ameliorate this, `app::main` has been changed to build
stacks from the "bottom" (endpoint client) to "top" (serverside
connection) by _push_-ing Layers onto a concrete stack, i.e. and not
composing layers for an abstract stack.

While doing this, we take the oppportunity to remove a ton of
now-unnecessary `PhantomData`. A new, dedicated `phantom_data` stack
module can be used to aid type inference as needed.

Other stack utilities like `map_target` and `map_err` have been
introduced to assist this transition.

Furthermore, all instances of `Layer::new` have been changed to a free
`fn layer` to improve readability.

This change sets up two upcoming changes: a stack-oriented `controller`
client and, subsequently, service-profile-based routing. There are no functional
changes.

### DIFF
--- a/lib/stack/src/layer.rs
+++ b/lib/stack/src/layer.rs
@@ -10,15 +10,15 @@ use std::marker::PhantomData;
 /// ```ignore
 /// impl<M: Stack<SocketAddr>> Layer<Authority, SocketAddr, M> for BalanceLayer<M> { ... }
 /// ```
-pub trait Layer<T, U, M: super::Stack<U>> {
+pub trait Layer<T, U, S: super::Stack<U>> {
     type Value;
     type Error;
     type Stack: super::Stack<T, Value = Self::Value, Error = Self::Error>;
 
-    /// Produce a `Stack` value from a `M` value.
-    fn bind(&self, next: M) -> Self::Stack;
+    /// Produces a `Stack` value from a `M` value.
+    fn bind(&self, next: S) -> Self::Stack;
 
-    /// Compose this `Layer` with another.
+    /// Produces a new Layer with this layer wrapping the provided inner layer.
     fn and_then<V, N, L>(self, inner: L)
         -> AndThen<U, Self, L>
     where
@@ -31,6 +31,42 @@ pub trait Layer<T, U, M: super::Stack<U>> {
             inner,
             _p: PhantomData,
         }
+    }
+
+    /// Produces a new Layer with another layer wrapping this one.
+    fn push<R, L>(self, outer: L)
+        -> AndThen<T, L, Self>
+    where
+        L: Layer<R, T, Self::Stack>,
+        Self: Sized,
+    {
+        AndThen {
+            outer,
+            inner: self,
+            _p: PhantomData,
+        }
+    }
+
+    /// Wraps this layer such that stack errors are modified by `map_err`.
+    fn map_err<M>(self, map_err: M)
+        -> AndThen<T, super::map_err::Layer<M>, Self>
+    where
+        Self: Sized,
+        M: super::map_err::MapErr<Self::Error>,
+        super::map_err::Layer<M>: Layer<T, T, Self::Stack>,
+    {
+        super::map_err::layer(map_err).and_then(self)
+    }
+}
+
+/// The identity layer.
+impl<T, M: super::Stack<T>> Layer<T, T, M> for () {
+    type Value = M::Value;
+    type Error = M::Error;
+    type Stack = M;
+
+    fn bind(&self, inner: M) -> M {
+        inner
     }
 }
 
--- a/lib/stack/src/map_err.rs
+++ b/lib/stack/src/map_err.rs
@@ -1,0 +1,72 @@
+pub fn layer<E, M>(map_err: M) -> Layer<M>
+where
+    M: MapErr<E>,
+{
+    Layer(map_err)
+}
+
+pub(super) fn stack<T, S, M>(inner: S, map_err: M) -> Stack<S, M>
+where
+    S: super::Stack<T>,
+    M: MapErr<S::Error>,
+{
+    Stack {
+        inner,
+        map_err,
+    }
+}
+
+pub trait MapErr<Input> {
+    type Output;
+
+    fn map_err(&self, e: Input) -> Self::Output;
+}
+
+#[derive(Clone, Debug)]
+pub struct Layer<M>(M);
+
+#[derive(Clone, Debug)]
+pub struct Stack<S, M> {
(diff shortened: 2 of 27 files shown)


lib/stack/src/layer.rs
```diff
@@ -32,6 +32,42 @@ pub trait Layer<T, U, M: super::Stack<U>> {
             _p: PhantomData,
         }
     }
+
+    /// Produces a new Layer with another layer wrapping this one.
+    fn push<R, L>(self, outer: L)
```
hawkw: 👍 This also makes me feel better about the "stack" naming ;)

lib/stack/src/map_err.rs
```diff
@@ -0,0 +1,72 @@
...
+    S: super::Stack<T>,
+    M: MapErr<S::Error>,
+{
+    Stack {
+        inner,
+        map_err,
+    }
+}
+
+pub trait MapErr<Input> {
```
hawkw: I'm assuming that adding this trait, rather than just using `Fn(Input) -> Output` directly, is to allow structs to implement a stateful `MapErr`? It might be good to have a comment explaining that...

lib/stack/src/phantom_data.rs
```diff
@@ -0,0 +1,39 @@
...
+    Layer(PhantomData)
+}
+
+#[derive(Clone, Debug)]
+pub struct Layer<T, M>(PhantomData<fn() -> (T, M)>);
+
+#[derive(Clone, Debug)]
+pub struct Stack<T, M> {
+    inner: M,
+    _p: PhantomData<fn() -> (T, M)>,
```
hawkw: Nit: this should only need a `PhantomData(fn() -> T)`; `M` is bound by the `inner` field.

Also, TIOLI: adding a `where M: super::Stack<T>` bound to the definition of this struct would mean you don't need the `PhantomData` at all.

lib/stack/src/watch.rs
```diff
@@ -14,14 +14,16 @@ pub trait WithUpdate<U> {
 }
 
 #[derive(Debug)]
-pub struct Layer<U> {
+pub struct Layer<T: WithUpdate<U>, U, M> {
     watch: Watch<U>,
+    _p: ::std::marker::PhantomData<(T, M)>,
```
hawkw: Nit: this should probably be `PhantomData<fn() -> (T, M)>` in case `T` or `M` are not `Send`/`Sync`?

lib/stack/src/watch.rs
```diff
@@ -14,14 +14,16 @@ pub trait WithUpdate<U> {
...
 }
 
 #[derive(Debug)]
-pub struct Stack<U, M> {
+pub struct Stack<T: WithUpdate<U>, U, M> {
     watch: Watch<U>,
     inner: M,
+    _p: ::std::marker::PhantomData<T>,
```
hawkw: Nit: again, this should probably be `fn() -> T`?

lib/stack/src/watch.rs
hawkw: Alternatively, TIOLI, I think `Layer` could just be
```rust
pub struct Layer<U> {
    watch: Watch<U>
}
```
and the `T` and `M` types could only be present on the functions that need the `T: WithUpdate<U>` bound...

src/proxy/http/router.rs
```diff
@@ -25,17 +25,14 @@ pub struct Config {
 /// A `Rec`-typed `Recognize` instance is used to produce a target for each
 /// `Req`-typed request. If the router doesn't already have a `Service` for this
 /// target, it uses a `Stk`-typed `Service` stack.
-#[derive(Debug)]
-pub struct Layer<Req, Rec, Stk> {
+#[derive(Clone, Debug)]
+pub struct Layer<Req, Rec: Recognize<Req>> {
     recognize: Rec,
-    _p: PhantomData<fn() -> (Req, Stk)>,
+    _p: PhantomData<fn() -> Req>,
```
hawkw: I _think_ this phantomdata is unnecessary because of the `Rec: Recognize<Req>` bound (but I could be wrong).

lib/stack/src/layer.rs
```diff
@@ -10,15 +10,15 @@ use std::marker::PhantomData;
...
     type Error;
     type Stack: super::Stack<T, Value = Self::Value, Error = Self::Error>;
 
-    /// Produce a `Stack` value from a `M` value.
-    fn bind(&self, next: M) -> Self::Stack;
+    /// Produces a `Stack` value from a `M` value.
+    fn bind(&self, next: S) -> Self::Stack;
 
-    /// Compose this `Layer` with another.
+    /// Produces a new Layer with this layer wrapping the provided inner layer.
```
hawkw: This is minor, but I think it would be good to have a doc comment here explaining the cases where `and_then` can make compile times slow/when you should use `push` instead?

lib/stack/src/map_err.rs
olix0r: It's really so that it's possible to have a named type. I think it's an anti-pattern to expose `Fn` generics in public interfaces.

lib/stack/src/phantom_data.rs
olix0r: > Also, TIOLI: adding a where M: super::Stack<T> bound to the definition of this struct would mean you don't need the PhantomData at all.

This is [not true](https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=c9874cf1ef78b0b7ec4895ca5252a569).

lib/stack/src/watch.rs
olix0r: it doesn't work out that way in practice. PhantomData is necessary.

src/proxy/http/router.rs
olix0r: phanomdata is necessary.

lib/stack/src/phantom_data.rs
hawkw: Oh, color me wrong, then! Sorry!

lib/stack/src/layer.rs
olix0r: I honestly don't understand it well enough to have a cogent explanation.

hawkw: Okay, that's fair. I think even a "This sometimes makes type inference slow, use `push` instead" would be sufficient, though...

PR #112, opened 2018-10-29
Prepare HTTP metrics for per-route classification

olix0r:
In order to support Service Profiles, the proxy will add a new scope of
HTTP metrics prefixed with `route_`, i.e. so that the proxy exposes
`request_total` and `route_request_total` independently.

Furthermore, the proxy must be able to use different
response-classification logic for each route, and this classification
logic should apply to both metrics scopes.

This alters the `proxy::http::metrics` module so that:

1. HTTP metrics may be scoped with a prefix (as the stack is described).

2. The HTTP metrics layer now discovers the classifier by trying to
   extract it from each request's extensions or fall back to a `Default`
   implementation. Only a default implementation is used presently.

3. It was too easy to use the `Classify` trait API incorrectly.
   Non-default classify implementation could cause a runtime panic!
   The API has been changed so that the type system ensures correct
   usage.

4. The HTTP classifier must be configurable per-request. In order to do
   this, we expect a higher stack layer will add response classifiers to
   request extensions when appropriate (i.e., in a follow-up).

Finally, the `telemetry::Report` type requires updating every time a new
set of metrics is added. We don't need a struct to represent this.
`FmtMetrics::and_then` has been added as a combinator so that a fixed
type is not necessary.


src/app/main.rs
```diff
@@ -175,18 +176,16 @@ where
...
-        let report = telemetry::Report::new(
-            http_report,
-            transport_report,
-            tls_config_report,
-            telemetry::process::Report::new(start_time),
-        );
+        let report = http_report
+            .and_then(transport_report)
+            .and_then(tls_config_report)
+            .and_then(telemetry::process::Report::new(start_time));
```
hawkw: This is smart --- I dig it! 👍

src/proxy/http/classify.rs
```diff
@@ -24,28 +30,30 @@ pub trait ClassifyResponse {
...
+    /// Produce a stream classifier for this response.
     ///
     /// If this is enough data to classify a response, a classification may be
-    /// returned. Implementations should expect that `end` or `error` may be
-    /// called even when a class is returned.
-    ///
-    /// This is expected to be called only once.
-    fn start<B>(&mut self, headers: &http::Response<B>) -> Option<Self::Class>;
+    /// returned.
+    fn start<B>(self, headers: &http::Response<B>) -> (Self::ClassifyEos, Option<Self::Class>);
```
hawkw: TIOLI: Maybe instead of a tuple, this should return a struct that more clearly encodes the states that the return type represents? Something like
```rust
pub struct ResponseStart<C: ClassifyResponse> {
    classify_eos: C::ClassifyEos,
    class_at_start: Option<C::Class>
}
```

On the other hand, maybe that just complicates things... 🤷‍♀️

src/proxy/http/metrics/report.rs
```diff
@@ -126,3 +142,49 @@ where
...
+impl Scope {
+    fn prefixed(prefix: &'static str) -> Self {
+        if prefix.is_empty() {
+            return Self::default();
+        }
+
+        Self {
+            request_total_key: format!("{}_request_total", prefix),
+            response_total_key: format!("{}_response_total", prefix),
+            response_latency_ms_key: format!("{}_response_latency_ms", prefix),
```
hawkw: This seems like it'll result in a lot of duplication of the strings "_request_total", "_response_total", and "_response_latency_ms"... if metrics memory usage is an issue, it seems like it should be possible to have this struct only store the prefix and interpolate it with those strings when we format the scope key? 

This is probably only worth bothering with if it turns out to be an issue, but I figured I'd ask about it.

src/proxy/http/metrics/report.rs
```diff
@@ -126,3 +142,49 @@ where
...
+        Self {
+            request_total_key: "request_total".to_owned(),
+            response_total_key: "response_total".to_owned(),
+            response_latency_ms_key: "response_latency_ms".to_owned(),
+        }
+    }
+}
+
+impl Scope {
+    fn prefixed(prefix: &'static str) -> Self {
```
hawkw: Since the prefix is just passed to a `format!` invocation, I don't think it needs to be `'static`. `&str` should be sufficient (or even `T: fmt::Display`).

src/proxy/http/metrics/report.rs
olix0r: > This seems like it'll result in a lot of duplication of the strings

Why? Do you anticipate there being many instances of these?

hawkw: If this type is constructed on a per-route basis, it seems like there's the potential for that, in the case of a proxy which talks to a lot of services that each have a lot of routes...but the scale may never be great enough for it to matter.

olix0r: This is the _report_ type, which means that this is only constructed... once ;)

https://github.com/linkerd/linkerd2-proxy/blob/4e0a1f0100d1a87e7658e8ffe6906290c370f8f5/src/app/main.rs#L176-L179

src/proxy/http/classify.rs
olix0r: Yeah, I think it's OK as-is, for now. We can see how we like it and change it later if desirable.

src/proxy/http/metrics/report.rs
olix0r: Yeah, good point. Let's change this if we ever want to use anything besides a static str ;)

olix0r:
https://github.com/linkerd/linkerd2-proxy/pull/113 follows on this with a bufix and classification unit tests

PR #113, opened 2018-10-29
Use the grpc-status response header for classification

olix0r:
When a gRPC service fails a request eagerly, before it begins sending a
response, a `grpc-status` header is simply added to the initial response
header (rather than added to trailers).

This change ensures that classification honors these status codes.

Fixes linkerd/linkerd2#1819.


src/app/classify.rs
```diff
@@ -87,6 +86,135 @@ impl classify::ClassifyEos for ClassifyEos {
...
+
+    #[test]
+    fn grpc_response_trailer_ok() {
+        let rsp = Response::builder().status(StatusCode::OK).body(()).unwrap();
+        let crsp = ClassifyResponse {};
+        let (ceos, class) = crsp.start(&rsp);
+        assert_eq!(class, None);
+
+        let mut trailers = HeaderMap::new();
+        trailers.insert("grpc-status", "0".parse().unwrap());
```
seanmonstar: PS, numbers can be converted via `0.into()`.

PR #114, opened 2018-10-30
replace proxy::http usage of tower-h2 with hyper

seanmonstar:
This shrinks much of the `proxy::http::glue` module, which before, was around abstracting over both `hyper` and `tower-h2`. This changes it so hyper's HTTP2 support is used instead of tower-h2. Therefore, the `glue` module is able to get much smaller, now mostly just being glue for `hyper` + `tower` (as `tower` stabilizes some, hyper will eventually provide support itself, and the glue can go away altogether).

**One functional change caused is that HTTP2 requests have lost the ability to wait for a reconnect.** The reason for the change is due to how `hyper::Client` cannot implement `NewService` the same way that `tower_h2::Client` could. hyper's `Client` cannot pre-emptively prepare a connection without a `Request`. This "loss" makes it function the same as HTTP1 has the whole time. With the eventual introduction of general retries, this functionality should be re-gained, but even better, working for both HTTP/1 *and* HTTP2.

Due to this change, I'm hesitant to merge this right before announcing Linkerd 2.1. Ideally, this change could be tested internally for some time, through development and integration tests.


olix0r:
Now that https://github.com/linkerd/linkerd2-proxy/pull/115 is on master, which reuses our http modules with a grpc client, this unfortunately becomes a more difficult change.

seanmonstar:
After a rebase of master, it still compiles! Starting up the tests now...

seanmonstar:
And.... the tests still pass...

seanmonstar:
Rebased again, and still seems to be working!

seanmonstar:
To add metrics or balancing (before the overhaul of tower-grpc to not depend directly on tower-h2, which is turning out to be really entrenched), we could add some `proxy::http::payload_glue::layer()`, I believe?

olix0r:
This needs to be rebased with the tower update, but I think we can move forward with this after that. I suspect the `app::controll::box_request_body` trait may need to be tweaked to do more, but seems straightforward now that body types have been made tower-grpc specific.

seanmonstar:
I realized also that I can probably keep the h2 reconnect functionality by specifically using `hyper::client::conn` in the h2 case. That allows managing a single connection, and since we don't otherwise need the pool behavior for h2.

src/app/control.rs
```diff
@@ -369,87 +369,107 @@ pub mod client {
...
-            ::logging::ClientExecutor<&'static str, Addr>,
-            BoxBody,
+            B,
         >;
         type Error = C::Error;
 
         fn make(&self, target: &Target) -> Result<Self::Value, Self::Error> {
             let c = self.connect.make(&target.connect)?;
-            let h2 = target.builder.clone();
+            let _h2 = target.builder.clone();
```
olix0r: should this line be removed? how can we configure window/stream concurrency?

src/proxy/http/client.rs
```diff
@@ -290,12 +233,7 @@ where
                 }
             },
             Settings::Http2 => {
-                let mut h2_builder = h2::client::Builder::default();
-                // h2 currently doesn't handle PUSH_PROMISE that well, so we just
-                // disable it for now.
-                h2_builder.enable_push(false);
-                let h2 = tower_h2::client::Connect::new(connect, h2_builder, BoxExecutor::new(executor));
-
+                let h2 = h2::Connect::new(connect, executor);
```
olix0r: similarly, this is probably fine for now. but we'll need to be able to configure h2 clients at some point.

src/proxy/http/h2.rs
```diff
@@ -0,0 +1,164 @@
...
+                        .execute(conn.map_err(|err| debug!("http2 conn error: {}", err)))
+                        .unwrap();
+
+                    return Ok(Connection { tx }.into());
+                }
+            };
+
+            let _e = &self.executor;
+            let handshake = conn::Builder::new()
+                .http2_only(true)
```
olix0r: does this disable push-promise by default?

src/proxy/server.rs
```diff
@@ -287,8 +284,8 @@ where
                 (p, io)
             });
 
-        let h1 = self.h1.clone();
-        let h2_settings = self.h2_settings.clone();
+        let http = self.http.clone();
+        let _h2_settings = self.h2_settings.clone();
```
olix0r: ?

src/proxy/http/h2.rs
seanmonstar: Yes: https://github.com/hyperium/hyper/blob/master/src/proto/h2/client.rs#L43

src/app/control.rs
seanmonstar: Ah yea, it should be removed. hyper can add builder options to configure HTTP2 things as we need them. We currently don't configure anything (besides disabling push, which hyper also does).

seanmonstar:
Updated again.

seanmonstar:
It would seem then that we should adjust CI so it would have caught that, for next time!

PR #115, opened 2018-10-31
refactor: Use a stack-based controller client

olix0r:
The controller's client is instantiated in the
`control::destination::background` module and is tightly coupled to its
use for address resolution.

In order to share this client across different modules---and to bring it
into line with the rest of the proxy's modular layout---the controller
client is now configured and instantiated in `app::main`. The
`app::control` module includes additional stack modules needed to
configure this client.

Our dependency on tower-buffer has been updated so that buffered
services may be cloned.

The `proxy::reconnect` module has been extended to support a
configurable fixed reconnect backoff; and this backoff delay has been
made configurable via the environment.


src/app/main.rs
```diff
@@ -190,172 +198,209 @@ where
...
+                .push(svc::stack::phantom_data::layer())
+                .push(control::add_origin::Layer::new())
+                .push(buffer::layer())
+                .push(limit::layer(config.destination_concurrency_limit));
 
-        let (drain_tx, drain_rx) = drain::channel();
+            // Because the control client is buffered, we need to be able to
+            // spawn a task on an executor. An executor is only available in the
+            // context of a running task, however, we the stack is instantiated lazily.
+            future::poll_fn(move || {
```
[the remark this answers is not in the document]
seanmonstar: I don't actually understand why or what this is doing... Why does this need to be in a `poll_fn`?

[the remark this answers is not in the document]
olix0r: @seanmonstar `tower-buffer` Services need an executor...

https://github.com/tower-rs/tower/blob/b95c8d103056d9876b80b856b5f76754bf0f7b85/tower-buffer/src/lib.rs#L88-L95

I didn't spend a lot of time thinking about a cleaner way to do this, so sticking it all in a future::poll_fn was the quickest hack i could think of.  (Also, I'm not intimately familiar with the various Executor types and how they work...)

We configure the buffer with a `logging::context_executor`:

https://github.com/linkerd/linkerd2-proxy/blob/f97239baf0b1d5f7f1cb93a895c7ecadf19b004b/src/proxy/buffer.rs#L61-L65

Without doing this within the context of a task, this `make` fails.

src/app/config.rs
```diff
@@ -77,6 +77,8 @@ pub struct Config {
     /// a new connection.
     pub control_backoff_delay: Duration,
 
+    pub control_connect_timeout: Duration,
```
hawkw: Nit: It seems like most of the other fields here have doc comments, can we get one here as well>

src/app/control.rs
```diff
@@ -0,0 +1,419 @@
...
+            }
+        }
+    }
+
+    // === impl Error ===
+
+    impl<S: fmt::Display, I: fmt::Display> fmt::Display for Error<S, I> {
+        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+            match self {
+                Error::Dns(_) => write!(f, "dns error"),
```
hawkw: Is there a reason we don't try and format the specific DNS error?

src/app/main.rs
```diff
@@ -172,12 +172,20 @@ where
             config.outbound_ports_disable_protocol_detection,
         );
 
+        let (drain_tx, drain_rx) = drain::channel();
+
+        let (dns_resolver, dns_bg) = dns::Resolver::from_system_config_and_env(&config)
+            .unwrap_or_else(|e| {
+                // TODO: Stack DNS configuration infallible.
```
hawkw: Nit: this comment makes sense but is kind of grammatically weird.

src/app/main.rs
```diff
@@ -190,172 +198,209 @@ where
...
+        };
 
-        let outbound = {
-            use super::outbound::{discovery::Resolve, orig_proto_upgrade, Endpoint, Recognize};
-            use proxy::{
-                http::{balance, metrics},
-                resolve,
+        let (resolver_bg_tx, resolver_bg_rx) = futures::sync::oneshot::channel();
+
+        // Build the outbound and inbound proxies using the controller client.
```
hawkw: I like the way these futures are composed --- this is very readable.

src/app/main.rs
```diff
@@ -190,172 +198,209 @@ where
...
+        let main_fut = controller_fut.and_then(move |controller| {
+            let (resolver, resolver_bg) = control::destination::new(
+                controller.clone(),
+                dns_resolver.clone(),
+                config.namespaces.clone(),
+                config.destination_concurrency_limit,
+            );
+            resolver_bg_tx
```
hawkw: Is there a reason `main_fut` can't just `tokio::spawn(resolver_bg)`, rather than having to send it on the channel? This seems unnecessarily complex at a glance.

src/control/destination/mod.rs
```diff
@@ -24,30 +24,32 @@
...
+    future,
     sync::mpsc,
     Async,
     Future,
     Poll,
     Stream
 };
+use indexmap::IndexMap;
+use std::fmt;
+use std::sync::{Arc, Weak};
```
hawkw: It looks like some imports were moved around, was this a `rustfmt` change?

src/app/main.rs
olix0r: The intent is so that resolver_bg runs on the other core.

src/control/destination/mod.rs
olix0r: probably, sort of. i've tried to make our imports more uniform as i touch files, since they're all over the place.

src/app/main.rs
hawkw: Ah, nevermind; this does seem like the best way to do that. I forgot we were still running the separate admin runtime.

src/app/control.rs
```diff
@@ -315,7 +315,8 @@ pub mod resolve {
     impl<S: fmt::Display, I: fmt::Display> fmt::Display for Error<S, I> {
         fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
             match self {
-                Error::Dns(_) => write!(f, "dns error"),
+                Error::Dns(dns::Error::NoAddressesFound) => write!(f, "no addresses found"),
+                Error::Dns(dns::Error::ResolutionFailed(e)) => fmt::Display::fmt(&e, f),
```
hawkw: 👍

src/app/main.rs
```diff
@@ -190,172 +198,210 @@ where
...
-            )
+                .push(limit::layer(config.destination_concurrency_limit));
+
+            // Because the control client is buffered, we need to be able to
+            // spawn a task on an executor when `make` is called. This is done
+            // lazily so that a default executor is available to spawn the
+            // background buffering task.
+            future::lazy(move || match control_config {
+                None => future::ok(None),
+                Some(config) => match stack.make(&config) {
```
hawkw: Nit: Since this now returns `Result<Option<...>, ...>` rather than `Option<...>`, I think this match can be reduced to something like 
```suggestion
                Some(config) => stack
                    .make(&config)
                    .map(Some)
                    .map_err(|e| { error!("failed to build controller: {}", e); })
```

hawkw: (Also, IIRC, it's not strictly necessary to use `future::ok`/`future::err` here; `future::lazy` takes a closure returning `IntoFuture`, which `Ok` and `Err` implement)

PR #116, opened 2018-10-31
Only use the classification label for response_total

olix0r:
As described in https://github.com/linkerd/linkerd2/issues/1832, our eager-classification is too complicated.

This changes the `classification` label to only be used with the `response_total` label.

The following changes have been made:
1. response_latency metrics only include a status_code and not a classification.
2. response_total metrics include classification labels.
3. transport metrics no longer expose a `classification` label (since it's misleading). now the `errno` label is set to be empty when there is no error.
4. gRPC classification only applies when the request's content type starts with `application/grpc+`

The `proxy::http::classify` APIs have been changed so that classifiers cannot return a classification before the classifier is fully consumed.


olix0r:
related grafana dashboard fix: https://github.com/linkerd/linkerd2/pull/1839

src/app/classify.rs
```diff
@@ -3,22 +3,36 @@ use http;
 
 use proxy::http::classify;
 
-#[derive(Clone, Debug, Default)]
-pub struct Classify {}
+pub use proxy::http::classify::insert;
 
 #[derive(Clone, Debug, Default)]
-pub struct ClassifyResponse {}
+pub struct Request {}
```
hawkw: Nit: This can also be declared as just
```suggestion
pub struct Request;
```
and then you can construct it by just saying `Request`, rather than having to say `Request {}`.

TIOLI.

src/app/inbound.rs
```diff
@@ -38,6 +40,14 @@ impl ShouldStackPerRequest for Endpoint {
     }
 }
 
+impl CanClassify for Endpoint {
+    type Classify = super::classify::Request;
```
hawkw: Nit/TIOLI: personally, I find referring to these types as `super::foo` a little unclear --- I'd prefer if we used the full path for these. Not a big deal though.

src/proxy/http/classify.rs
```diff
@@ -57,3 +54,88 @@ pub trait ClassifyEos {
...
 }
+
+// Used for stack targets that can produce a `Classify` implementation.
+pub trait CanClassify {
+    type Classify: Classify;
+
+    fn classify(&self) -> Self::Classify;
+}
+
+pub mod insert {
```
hawkw: Nit: it's not immediately clear to me why this needs its own module --- why not just `http::classify::layer`?

src/proxy/http/metrics/mod.rs
```diff
@@ -79,8 +87,19 @@ where
...
+            by_status: IndexMap::default(),
+        }
+    }
+}
+
+impl<C> Default for StatusMetrics<C>
+where
+    C: Hash + Eq,
+{
+    fn default() -> Self {
```
hawkw: Nit: This `Default` impl seems like it should be able to be derived. If there's a reason it can't be, it would be good to have a comment, I think.

src/proxy/http/metrics/report.rs
```diff
@@ -120,6 +119,28 @@ where
...
+        metric: Metric<M>,
+        get_metric: F,
+    ) -> fmt::Result
+    where
+        M: FmtMetric,
+        F: Fn(&StatusMetrics<C>) -> &M,
+    {
+        for (tgt, tm) in &self.by_target {
+            if let Ok(tm) = tm.lock() {
+                for (status, m) in &tm.by_status {
```
hawkw: TIOLI: if you bind this like
```suggestion
                for (&status, ref m) in &tm.by_status {
```
then you don't need to dereference later:
```rust
                    let labels = (tgt, Status(status));
                    get_metric(m).fmt_metric_labeled(f, metric.name, labels)?;
```

src/proxy/http/metrics/service.rs
```diff
@@ -363,39 +357,58 @@ where
...
-        let class_metrics = match class {
-            Some(c) => metrics
-                .by_class
-                .entry(c)
-                .or_insert_with(|| ClassMetrics::default()),
-            None => &mut metrics.unclassified,
+        let status_metrics = metrics
+            .by_status
+            .entry(self.status)
+            .or_insert_with(|| StatusMetrics::default());
```
hawkw: Nit: [`or_default`](https://docs.rs/indexmap/1.0.2/indexmap/map/enum.Entry.html#method.or_default)?

src/proxy/http/metrics/service.rs
```diff
@@ -413,16 +426,11 @@ where
...
-        if self.first_byte_at.is_none() {
-            self.first_byte_at = Some(clock::now());
-        }
-
-        if let c @ Some(_) = self.class_at_first_byte.take() {
-            self.record_class(c);
-            self.classify = None;
+        if !self.latency_recorded {
+            self.record_latency();
+            self.latency_recorded = true;
```
hawkw: Nit/TIOLI: Since `record_latency` takes `&mut self`, it seems reasonable to assume that _it_ would be responsible for setting this flag...

src/proxy/http/metrics/service.rs
olix0r: still doesn't exist on indexmap

src/proxy/http/metrics/mod.rs
olix0r: because C does not have to impl default

src/app/classify.rs
olix0r: In the next PR this gets more stuff

src/proxy/http/metrics/service.rs
hawkw: the link above is to the `IndexMap` docs...maybe we're not on a new enough version?

src/app/classify.rs
hawkw: ah okay, carry on then :)

PR #117, opened 2018-11-01
Introduce outbound route metrics

olix0r:
The Destination Profile API---provided by linkerd2-proxy-api v0.1.3--
allows the proxy to discovery route information for an HTTP service. As
the proxy processes outbound requests, in addition to doing address
resolution through the Destination service, the proxy may also discover
profiles including route patterns and labels.

When the proxy has route information for a destination, it applies the
RequestMatch for each route to find the first-matching route. The
route's labels are used to expose `route_`-prefixed HTTP metrics (and
each label is prefixed with `rt_`).

Furthermore, if a route includes ResponseMatches, they are used to
perform classification (i.e. for the `response_total` and
`route_response_total` metrics).

A new `proxy::http::profiles` module implements a router that consumes
routes from an infinite stream of route lists.

The `app::profiles` module implements a client that continually and
repeatedly tries to establish a watch for the destination's routes (with
some backoff).

Route discovery does not _block_ routing; that is, the first request to
a destination will likely be processed before the route information is
retrieved from the controller (i.e. on the default route). Route
configuration is applied in a best-effort fashion.


olix0r:
This needs more automated testing, but it has been tested manually quite a bit.

src/app/main.rs
```diff
@@ -290,22 +299,34 @@ where
...
+                    KubernetesNormalizer::new(config.namespaces.pod.clone()),
+                );
+
+                let dst_route_stack = endpoint_stack
                     .push(resolve::layer(Resolve::new(resolver)))
                     .push(balance::layer())
                     .push(buffer::layer())
+                    .push(profiles::router::layer(
+                        profiles_client,
+                        svc::stack::phantom_data::layer()
```
hawkw: Just curious --- is the `phantom_data` layer here needed because of type inference issues, or is it just used to push the metrics and classification layers to something? I'd like a better understanding of when/where that layer needs to be used...

src/app/profiles.rs
```diff
@@ -0,0 +1,235 @@
...
+                    Ok(Async::NotReady) => return Ok(Async::NotReady),
+                    Err(_) | Ok(Async::Ready(())) => State::Disconnected,
+                },
+            };
+        }
+    }
+}
+
+impl<T: HttpService> Rx<T> {
+    fn convert_routes(
```
hawkw: Nit/TIOLI: It's a little strange to me that these functions are all defined on `Rx`, since none of them actually involve an `Rx` or any of it's type parameters. If they're only intended to be used within the future impl, perhaps they should just be free functions defined inside `poll`? Not a big deal though.

src/proxy/http/profiles.rs
```diff
@@ -0,0 +1,389 @@
...
+
+pub type Routes = Vec<(RequestMatch, Route)>;
+
+pub trait GetRoutes {
+    type Stream: Stream<Item = Routes, Error = Error>;
+
+    fn get_routes(&self, dst: &DnsNameAndPort) -> Option<Self::Stream>;
+}
+
+pub trait WithRoute {
```
hawkw: Would love some doc comments on these traits explaining what they're for and how they're intended to be used...

src/proxy/http/profiles.rs
```diff
@@ -0,0 +1,389 @@
...
+// === impl RequestMatch ===
+
+impl RequestMatch {
+    fn is_match<B>(&self, req: &http::Request<B>) -> bool {
+        match self {
+            RequestMatch::Method(ref method) => req.method() == *method,
+            RequestMatch::Path(ref re) => re.is_match(req.uri().path()),
+            RequestMatch::Not(ref m) => !m.is_match(req),
+            RequestMatch::All(ref matches) => {
+                for ref m in matches {
```
hawkw: TIOLI: I think this can just be:
```suggestion
                matches.iter().all(RequestMatch::is_match),
```

src/proxy/http/profiles.rs
```diff
@@ -0,0 +1,389 @@
...
+                    if !m.is_match(req) {
+                        return false;
+                    }
+                }
+                true
+            }
+            RequestMatch::Any(ref matches) => {
+                for ref m in matches {
```
hawkw: ...and this is just
```suggestion
                matches.iter().any(RequestMatch::is_match),
```

src/proxy/http/profiles.rs
```diff
@@ -0,0 +1,389 @@
...
+
+impl ResponseMatch {
+    fn is_match<B>(&self, req: &http::Response<B>) -> bool {
+        match self {
+            ResponseMatch::Status { ref min, ref max } => {
+                *min <= req.status() && req.status() <= *max
+            }
+            ResponseMatch::Not(ref m) => !m.is_match(req),
+            ResponseMatch::All(ref matches) => {
+                for ref m in matches {
```
hawkw: Again, TIOLI, but these for-loops would be terser as `Iterator::all`/`Iterator::any`...

src/app/main.rs
olix0r: me too. afaict this is needed for inference.

PR #119, opened 2018-11-05
Remove settings checks from orig-proto module

olix0r:
When constructing the `outbound::orig_proto_upgrade` module, we check
the endpoint's `settings` to determine the `Upgrade` middleware should
be applied. The `Upgrade` middleware performs a similar, and partially
redundant set of checks on each request.

linkerd/linkerd2#1798 is simplif if we can separate destination-routing
from client settings. Specifically, if the `settings` field will be
removed from `Destination`s and `Endpoint`s, we won't a discovery &
routing stack for each combination of Settings. Instead, a routing layer
will be installed around the client, beneath the profile route stack,
that is responsible for dispatching requests to the appropriate client.

This change does NOT change user-facing behavior. An additional
`h1::wants_upgrade` check is now perormed within the `Upgrade` service
so that HTTP-Upgrade requests can pass through this stack.

### DIFF
--- a/src/proxy/http/orig_proto.rs
+++ b/src/proxy/http/orig_proto.rs
@@ -48,66 +48,58 @@ where
     }
 
     fn call(&mut self, mut req: Self::Request) -> Self::Future {
-        let mut downgrade_response = false;
+        if req.version() == http::Version::HTTP_2 || h1::wants_upgrade(&req) {
+            // Just passing through...
+            return self.inner.call(req).map(|res| res)
+        }
 
-        if req.version() != http::Version::HTTP_2 {
-            debug!("upgrading {:?} to HTTP2 with orig-proto", req.version());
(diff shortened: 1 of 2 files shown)


src/proxy/http/orig_proto.rs
```diff
@@ -48,66 +48,58 @@ where
...
-            );
-
-            // transfer-encoding is illegal in HTTP2
-            req.headers_mut().remove(TRANSFER_ENCODING);
-
-            *req.version_mut() = http::Version::HTTP_2;
-            downgrade_response = true;
+        if req.version() == http::Version::HTTP_2 || h1::wants_upgrade(&req) {
+            // Just passing through...
+            return self.inner.call(req).map(|res| res)
```
olix0r: heads up @seanmonstar  -- i moved the non-upgrade logic into an early return so that the control flow is a bit easier (for me) to follow

PR #120, opened 2018-11-07
Unify Name/Host/Addr types under Addr

olix0r:
Currently, the proxy uses a variety of types to represent the logical
destination of a request. Outbound destinations use a `NameAddr` type
which may be either a `DnsNameAndPort` or a `SocketAddr`. Other parts of
the code used a `HostAndPort` enum that always contained a port and also
contained a `Host` which could either be a `dns::Name` or a `IpAddr`.
Furthermore, we coerce these types into a `http::uri::Authority` in many
cases.

All of these types represent the same thing; and it's not clear when/why
it's appropriate to use a given variant.

In order to simplify the situtation, a new `host_port` module has been
introduced with `Addr` and `NameAddr` types. A `Addr` may
contain either a `NameAddr` or a `SocketAddr`.

The `Host` value has been removed from the `Settings::Http1` type,
replaced by a boolean, as it's redundant information stored elsewhere in
the route key.

There is one small change in behavior: The `authority` metrics label is
now omitted only for requests that include an `:authority` or `Host`
with a _name_ (i.e. and not an IP address).

### DIFF
--- a/src/host_port.rs
+++ b/src/host_port.rs
@@ -1,0 +1,170 @@
+use http;
+use std::fmt;
+use std::net::{IpAddr, SocketAddr};
+use std::str::FromStr;
+
+use convert::TryFrom;
+use dns;
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub enum HostPort {
+    Name(NamePort),
+    Addr(SocketAddr),
+}
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct NamePort {
+    name: dns::Name,
+    port: u16,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum Error {
+    /// The host is not a valid DNS name or IP address.
+    InvalidHost,
+
+    /// The port is missing.
+    MissingPort,
+}
+
+// === impl HostAndPort ===
+
+impl HostPort {
+    pub fn new(host: &str, port: u16) -> Result<Self, Error> {
+        match IpAddr::from_str(host) {
+            Ok(ip) => Ok(HostPort::Addr((ip, port).into())),
+            Err(_) => NamePort::new(host, port).map(HostPort::Name),
+        }
+    }
+
+    pub fn from_authority_and_default_port(
+        a: &http::uri::Authority,
+        default_port: u16,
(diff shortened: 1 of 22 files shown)


src/proxy/http/settings.rs
```diff
@@ -10,95 +8,81 @@ use super::h1;
...
 
+        let was_absolute_form = super::h1::is_absolute_form(req.uri());
         let is_h1_upgrade = super::h1::wants_upgrade(req);
-
         Settings::Http1 {
-            host,
-            is_h1_upgrade,
             was_absolute_form,
+            is_h1_upgrade,
+            stack_per_request: !has_name,
```
hawkw: TIOLI: I think the logic here could be simplified so that we have to do fewer negations. Consider something like:
```rust
        let stack_per_request = req
            .uri()
            .authority_part()
            .map(|_| false)
            .or_else(|| {
                req.headers()
                    .get(HOST)
                    .and_then(|h| h.to_str().ok())
                    .map(|h| h.is_empty())
            })
            .unwrap_or(true);

         // ...

        Settings::Http1 {
            // ...
            stack_per_request,
        }
```

olix0r: I had this initially but I found it confusing.  I’ll see if I can land on something that reduces negations but is still readable (probably naming the temp Val `missing_host` or something)

src/host_port.rs
```diff
@@ -0,0 +1,170 @@
+use http;
+use std::fmt;
+use std::net::{IpAddr, SocketAddr};
+use std::str::FromStr;
+
+use convert::TryFrom;
+use dns;
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub enum HostPort {
```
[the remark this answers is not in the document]
hawkw: I was _this_ suggesting `Addr`...

src/proxy/http/settings.rs
hawkw: Okay, that's also fair. I just found the number of negations made it hard to think through. 

Maybe more intermediate `let`s (and/or some comments) would help?

olix0r:
@hawkw wdyt about the `Addr`/`NameAddr` rename?

PR #121, opened 2018-11-09
Disable influxdb port detection

tsupertramp:
InfluxDB does not work with linkerd proxy.

### DIFF
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -255,6 +255,7 @@ const DEFAULT_DESTINATION_CLIENT_CONCURRENCY_LIMIT: usize = 100;
 const DEFAULT_PORTS_DISABLE_PROTOCOL_DETECTION: &[u16] = &[
     25,   // SMTP
     3306, // MySQL
+    8086, // InfluxDB
 ];
 
 // ===== impl Config =====


olix0r:
Unfortunately, 8086 is a very common, unregistered port. For instance, we serve the controller's proxy api on 8086.  I'm not sure that this is a good candidate for default configuratoin...

tsupertramp:
You are right. Sorry did not check that.
Thought it was an IANA assigned port.

What about adding influxdb protocol just to the docs so there is one more example or even create a page just with protocols which are unsupported? - No need for autodection. Just some hint that it won't work.

tsupertramp:
You are right. Sorry did not check that.
Thought it was an IANA assigned port.

What about adding influxdb protocol just to the docs so there is one more example or even create a page just with protocols which are unsupported? - No need for autodection. Just some hint that it won't work.

olix0r:
@tsupertramp yeah, i definitely think we should it add it to the docs! Would you like to take a crack at that change? I believe that would go here: https://github.com/linkerd/website/blob/master/linkerd.io/content/2/adding-your-service/_index.md#server-speaks-first-protocols

tsupertramp:
I am closing this one in the meantime. Thanks!

PR #127, opened 2018-11-13
Convert several Stack unit errors into Never

seanmonstar:
Since this stack pieces will never error, we can mark their
`Error`s with a type that can "never" be created. When seeing an `Error
= ()`, it can either mean the error never happens, or that the detailed
error is dealt with elsewhere and only a unit is passed on. When seeing
`Error = Never`, it is clearer that the error case never happens.
Besides helping humans, LLVM can also remove the error branchs entirely.

### DIFF
--- a/lib/stack/src/lib.rs
+++ b/lib/stack/src/lib.rs
@@ -1,6 +1,7 @@
 extern crate futures;
 #[macro_use]
 extern crate log;
+extern crate linkerd2_never as never;
 extern crate tower_service as svc;
 
 pub mod either;
@@ -53,7 +54,7 @@ pub trait Stack<T> {
 
 /// Implements `Stack<T>` for any `T` by cloning a `V`-typed value.
 pub mod shared {
-    use std::{error, fmt};
+    use never::Never;
 
     pub fn stack<V: Clone>(v: V) -> Stack<V> {
         Stack(v)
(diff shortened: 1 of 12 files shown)


lib/stack/src/lib.rs
```diff
@@ -63,22 +63,22 @@ pub mod shared {
     pub struct Stack<V: Clone>(V);
 
     #[derive(Debug)]
-    pub enum Error {}
+    pub enum Never {}
 
     impl<T, V: Clone> super::Stack<T> for Stack<V> {
         type Value = V;
-        type Error = Error;
+        type Error = Never;
```
hawkw: TIOLI: It might be worth adding a comment that this should be replaced with `!` when it becomes stable?

src/never.rs
```diff
@@ -0,0 +1,15 @@
+use std::{error::Error, fmt};
+
+/// A type representing a value that can never materialize.
+///
+/// This would be `!`, but it isn't stable yet.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum Never {}
```
hawkw: TIOLI: Maybe we ought to put this in a crate in `lib` so that `lib/stack` and the main binary can use the same definition of the never type? May not be worth it...

lib/stack/src/lib.rs
olix0r: :-1: i don't find these comments to be useful

src/never.rs
olix0r: that seems like a good idea to me

PR #132, opened 2018-11-16
Update all tower pieces to use Service<Request>

seanmonstar:


hawkw:
It looks like CI is failing because `lib/router` hasn't been updated?

seanmonstar:
Weird... Is it even used in the proxy? The proxy builds...

seanmonstar:
I've restarted CI several times now, seems intent on exploding due to disk space...

PR #135, opened 2018-11-16
Disable debug symbols for tests

olix0r:
Our test artifacts end up consuming several GB of disk space, largely
due to debug symbols. This can prevent CI from passing, as CI hosts only
have about 9G of real estate.

By disabling debug symbols, we reduce artifact size by >90% (and total
target directory size from 14G to 4G).

Thanks to @hawkw for spotting this.


iamwwc:
If we set debug=false, How do you debug program bug? Does unit test can cover it?

PR #136, opened 2018-11-16
Ensure l5d-orig-proto header is removed before returning responses

seanmonstar:
I noticed a pod was seeing this response header, it should be removed before the pod gets it.


olix0r:
nice find, thanks

PR #139, opened 2018-11-17
metrics/counter: wrap values over 2^53

lucab:
This implements Prometheus reset semantics for counters, in order to
preserve precision when deriving rate of increase.
Wrapping is based on the fact that Prometheus models counters as `f64`
(52-bits mantissa), thus integer values over 2^53 are not guaranteed to
be correctly exposed.

### DIFF
--- a/lib/metrics/src/counter.rs
+++ b/lib/metrics/src/counter.rs
@@ -1,70 +1,81 @@
 use std::fmt::{self, Display};
-use std::num::Wrapping;
 use std::ops;
 
-use super::prom::{FmtMetric, FmtLabels};
+use super::prom::{FmtLabels, FmtMetric};
 
-/// A Prometheus counter is represented by a `Wrapping` unsigned 64-bit int.
+/// A Prometheus counter is represented by a `Wrapping` unsigned 52-bit integer.
 ///
-/// Counters always explicitly wrap on overflows rather than panicking in
-/// debug builds. Prometheus' [`rate()`] and [`irate()`] queries handle breaks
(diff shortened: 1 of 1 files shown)


wmorgan:
@lucab nice!

lucab:
@olix0r late followup to this. Derived metrics are now precise but this is discarding the actual counter magnitude when the counter reaches ~8 Pebi-units.

For comparison, Go Prometheus library tries to tackle this in a [peculiar way](https://github.com/prometheus/client_golang/blob/v0.9.1/prometheus/counter.go#L69-L116), which is still going to loose precision at exposition time due to the `uint64` -> `float64` conversion.

Speaking with some other monitoring folks in the office today, it looks like it would make sense for the counter to store two values internally (current wrapping-integer and number of wraps), and serialize two metrics at exposition time (normal counter metric plus a `_magnitude` one). What are your thoughts on that?

olix0r:
@lucab That sounds reasonable, though I'm not sure if the storage overhead is worth it. Specifically, it would be unfortunate to have to export a `_magnitude` for each bucket in a histogram, as this would effectively double to cost of exporting a histogram for a marginal operational benefit --- if my napkin math is right, a counter incremented once per millisecond would not wrap for ~285,000 years.

lucab:
No indeed. My bad, I didn't see histograms reuse this exposition path, let's leave it as is for the now then. This doesn't concern much the single-increments counters, but more the "quantity adding" ones (data-transfer counters and similar metering).

PR #140, opened 2018-11-20
Profile Retries

seanmonstar:
This is a WIP that adds request retries if a service profile includes retry instructions, based on https://github.com/linkerd/linkerd2-proxy-api/pull/16.

- [x] Retries requests if a response class from a service profile matches and `is_failure`.
- [x] Uses `retry_budget` per service profile.
- [x] ~~Uses `retry_timeout` per `Route`.~~
- [x] ~~Use `default_retry_timeout` from the service profile if a route doesn't include a `retry_timeout`.~~
- [x] Figure out what metrics should look like.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ sudo: false
 
 language: rust
 cache: cargo
+env:
+  - RUSTFLAGS="-C debuginfo=0"
 
 branches:
   only:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,7 @@ tower-buffer          = { git = "https://github.com/tower-rs/tower" }
 tower-discover        = { git = "https://github.com/tower-rs/tower" }
 tower-in-flight-limit = { git = "https://github.com/tower-rs/tower" }
 tower-reconnect       = { git = "https://github.com/tower-rs/tower" }
+tower-retry           = { git = "https://github.com/tower-rs/tower" }
 tower-service         = { git = "https://github.com/tower-rs/tower" }
 tower-util            = { git = "https://github.com/tower-rs/tower" }
 tower-http            = { git = "https://github.com/tower-rs/tower-http" }
--- a/src/app/classify.rs
+++ b/src/app/classify.rs
@@ -188,6 +188,19 @@ fn grpc_class(headers: &http::HeaderMap) -> Option<Class> {
         })
 }
 
+// === impl Class ===
+
+impl Class {
+    pub(super) fn is_failure(&self) -> bool {
+        match self {
+            Class::Default(SuccessOrFailure::Failure) |
+            Class::Grpc(SuccessOrFailure::Failure, _) |
+            Class::Stream(SuccessOrFailure::Failure, _) => true,
+            _ => false,
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use http::{HeaderMap, Response, StatusCode};
--- a/src/proxy/http/metrics/report.rs
+++ b/src/proxy/http/metrics/report.rs
@@ -7,7 +7,7 @@ use tokio_timer::clock;
 
 use metrics::{latency, Counter, FmtLabels, FmtMetric, FmtMetrics, Histogram, Metric};
 
-use super::{ClassMetrics, Metrics, Registry, StatusMetrics};
+use super::{ClassMetrics, RequestMetrics, Registry, RetrySkipped, StatusMetrics};
 
 /// Reports HTTP metrics for prometheus.
 #[derive(Clone, Debug)]
@@ -28,6 +28,7 @@ struct Scope {
     request_total_key: String,
     response_total_key: String,
     response_latency_ms_key: String,
+    retry_skipped_total_key: String,
 }
 
 // ===== impl Report =====
@@ -89,6 +90,9 @@ where
         self.scope.response_total().fmt_help(f)?;
         registry.fmt_by_class(f, self.scope.response_total(), |s| &s.total)?;
 
+        self.scope.retry_skipped_total().fmt_help(f)?;
+        registry.fmt_by_retry(f, self.scope.retry_skipped_total())?;
+
         Ok(())
     }
 }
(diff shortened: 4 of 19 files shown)


seanmonstar:
For starters, I've changed the `route_*` metrics to be `route_effective`, which records metrics from *outside* the retry layer, and added another set of metrics prefixed by `route_actual`, which records all the retried requests and responses.

From these metrics, we can answer the effective vs actual success rate.

@ver In order to report about retryable requests that weren't retried due to timeout or budget, the `retry` stack needs some way of holding a recorder. Did you have something in mind for when *other* stacks need some metrics thing, so we don't duplicate what is in `metrics/service.rs` everywhere else?

seanmonstar:
@olix0r In order to report about retryable requests that weren't retried due to timeout or budget, the `retry` stack needs some way of holding a recorder. Did you have something in mind for when *other* stacks need some metrics thing, so we don't duplicate what is in `metrics/service.rs` everywhere else?

seanmonstar:
> I've changed the `route_*` metrics to be `route_effective_*` [...]

I've revert this change, the main route metrics just have the `route_` prefix. There is still a `route_actual_*` set of metrics, which can be used to compare what extra requests were made via retries. There are also new `retry_skipped_total{skipped="reason"}` metrics.

These end up looking like this:

```
route_response_total{direction="outbound",dst="dest:80",status_code="200",classification="success"} 5
route_response_total{direction="outbound",dst="dest:80",status_code="533",classification="failure"} 1


route_actual_response_total{direction="outbound",dst="dest:80",status_code="200",classification="success"} 5
route_actual_response_total{direction="outbound",dst="dest:80",status_code="533",classification="failure"} 6

route_actual_retry_skipped_total{direction="outbound",dst="dest:80",skipped="budget"} 1
```

.travis.yml
```diff
@@ -4,6 +4,8 @@ sudo: false
 
 language: rust
 cache: cargo
+env:
+  - RUSTFLAGS="-C incremental=false"
```
olix0r: wdyt about passing `-C debuginfo=0` as well? It seems unlikely this is at all useful to us in CI; and it reduces the artifact size from ~2G to ~30M.

Perhaps incremental can be enabled if we turn off debug info?

olix0r: On my Linux laptop, from a clean build:

```
:; make test >/dev/null 2>&1 && du -sh target
6.6G    target
:; env RUSTFLAGS="-C debuginfo=0" make test >/dev/null 2>&1 && du -sh target
1.2G    target
```

seanmonstar: Ah, I thought debuginfo had already been disabled in CI, and I was still destroying it...

src/proxy/http/retry.rs
```diff
@@ -0,0 +1,227 @@
...
+
+            if let Some(ext) = self.extensions().get::<::proxy::server::Source>() {
+                clone.extensions_mut().insert(ext.clone());
+            }
+            if let Some(ext) = self.extensions().get::<::app::classify::Response>() {
+                clone.extensions_mut().insert(ext.clone());
+            }
+            if let Some(ext) = self.extensions().get::<FirstRequestStartedAt>() {
+                clone.extensions_mut().insert(ext.clone());
+            }
```
olix0r: It's unfortunate that these extensions are hardocded here. I don't think the `app` types, especially, should be relied on anywhere outside of `app`.  Perhaps we should hang this trait off of the Target type, as is done for tap::Inspect?

seanmonstar:
Filed https://github.com/carllerche/h2/pull/334 to fix the HTTP2 connection window leak for retries.

olix0r:
I've run run a 6000+ requests (enough to trigger the issue previously) through the repro setup using a proxy patched with the h2 fix. I can no longer reproduce the starvation issue!

seanmonstar:
Updated to the newest proposal in https://github.com/linkerd/linkerd2-proxy-api/pull/16, which removed "retry timeouts".

Cargo.toml
```diff
@@ -33,7 +33,7 @@ linkerd2-stack     = { path = "lib/stack" }
 linkerd2-task      = { path = "lib/task" }
 linkerd2-timeout   = { path = "lib/timeout" }
 
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", tag = "v0.1.4", version = "0.1.4" }
+linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", branch = "alex/retries" } #tag = "v0.1.4", version = "0.1.4" }
```
olix0r: This should probably be changed to pin the master SHA until a new release is cut. That is, unless @adleong wants to cut an 0.1.5 in the current state.

src/app/classify.rs
```diff
@@ -171,6 +171,18 @@ impl classify::ClassifyEos for Eos {
     fn error(self, err: &h2::Error) -> Self::Class {
         Class::Stream(SuccessOrFailure::Failure, format!("{}", err))
     }
+
+    fn no_body(&self) -> Option<Self::Class> {
```
olix0r: How is this different from `eos(None)`?  Why does this not consume `self`?

seanmonstar: This is meant to try to classify before the body is streamed.

src/proxy/http/metrics/service.rs
```diff
@@ -262,16 +263,20 @@ where
...
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-        let rsp = try_ready!(self.inner.poll());
+        let mut rsp = try_ready!(self.inner.poll());
 
         let classify = self.classify.take().map(|c| c.start(&rsp));
 
+        if let Some(class) = classify.as_ref().and_then(|c| c.no_body()) {
+            rsp.extensions_mut().insert(class);
+        }
```
olix0r: OK, if I understand this properly, the retry layer depends on the metric layer setting this classification?

Can these things be decoupled?  This is subtle, and probably fragile.

olix0r: Specifically, if a response fails mid-stream, we'd mark it as a failure in metrics, though it wouldn't be retryable. What are the disadvantages of duplicating the classification logic?  I imagine the CPU overhead is infinitesimal...  Should the `classify` trait perhaps be specific to metrics? I imagine a retry-specific classification API would be a lot simpler...

seanmonstar: This part was added so that a "default" classifier could be used, if the route doesn't include any response classifiers. An alternative is to duplicate the default classifier stuff in retry as well.

olix0r: I don't think this is a blocker, just talking through it a bit...

I'm fine with leaving this as-is if we add a little more documentation that covers:
- the role of no_body in the classify API
- the soft extension-coupling between metrics and retries

src/proxy/http/metrics/report.rs
```diff
@@ -89,6 +90,9 @@ where
         self.scope.response_total().fmt_help(f)?;
         registry.fmt_by_class(f, self.scope.response_total(), |s| &s.total)?;
 
+        self.scope.retry_skipped_total().fmt_help(f)?;
+        registry.fmt_by_retry(f, self.scope.retry_skipped_total())?;
```
olix0r: As this is, we'll report these metrics for all http scopes, which is a bit confusing.  Perhaps the retry-specific stats logic should just be embedded directly in the retries module? something to the effect of having `retries::new() -> (Layer, Report)` ?

We would probably end up with keys like
- `route_effective_...` -- http metrics
- `route_retry_(count|skipped|budget)` -- retry metrics 
- `route_absolute_...` -- http metrics

seanmonstar:
> I'm pretty skeptical about the changes to classify

What changes are you referring to? The only change to `classify` in this PR is adding an `is_failure` method to `Class`, it's otherwise left alone...

PR #142, opened 2018-11-27
Apply tapping logic only when taps are active

olix0r:
Previously, as the proxy processed requests, it would:

1. Obtain the taps mutex ~4x per request to determine whether taps are active.
2. Construct an "event" ~4x per request, regardless of whether any taps were
   active.

Furthermore, this relied on fragile caching logic, where the grpc server
manages individual stream states in a Map to determine when all streams have
been completed. And, beyond the complexity of caching, this approach makes it
difficult to expand Tap functionality (for instance, to support tapping of
payloads).

This change entirely rewrites the proxy's Tap logic to (1) prevent the need
to acquire muteces in the request path, (2) only produce events as needed to
satisfy tap requests, and (3) provide clear (private) API boundaries between
the Tap server and Stack, completely hiding gRPC details from the tap service.

The `tap::service` module now provides a middleware that is generic over a
way to discover Taps; and the `tap::grpc` module (previously,
`control::observe`), implements a gRPC service that advertises Taps such that
their lifetimes are managed properly, leveraging RAII instead of hand-rolled
map-based caching.

There is one user-facing change: tap stream IDs are now calculated relative to
the tap server. The base id is assigned from the count of tap requests that have
been made to the proxy; and the stream ID corresponds to an integer on [0, limit).


olix0r:
This review is probably less-useful as a diff. I recommend checking out the branch and reading through the `tap` module, especially.

I'm especially looking for aspects of this that need more explanation, so please ask questions!.

PR #147, opened 2018-11-30
Expose route labels via tap

olix0r:
Route labels are not queryable by tap, nor are they exposed to in tap
events.

This change adds route label matching and adds route labels to events.

Depends on linkerd/linkerd2-proxy-api#17

### DIFF
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -341,6 +341,7 @@ where
                 // extension into each request so that all lower metrics
                 // implementations can use the route-specific configuration.
                 let dst_route_layer = phantom_data::layer()
+                    .push(insert_target::layer())
                     .push(metrics::layer::<_, classify::Response>(route_http_metrics))
                     .push(classify::layer());
 
@@ -495,6 +496,7 @@ where
                 // extension into each request so that all lower metrics
                 // implementations can use the route-specific configuration.
                 let dst_route_stack = phantom_data::layer()
+                    .push(insert_target::layer())
                     .push(http_metrics::layer::<_, classify::Response>(
                         route_http_metrics,
                     ))
--- a/src/app/outbound.rs
+++ b/src/app/outbound.rs
@@ -1,5 +1,6 @@
 use indexmap::IndexMap;
 use std::{fmt, net};
+use std::sync::Arc;
 
 use control::destination::{Metadata, ProtocolHint};
 use proxy::http::settings;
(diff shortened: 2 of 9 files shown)


src/app/main.rs
```diff
@@ -495,6 +496,7 @@ where
                 // extension into each request so that all lower metrics
                 // implementations can use the route-specific configuration.
                 let dst_route_stack = phantom_data::layer()
+                    .push(insert_target::layer())
```
seanmonstar: It's hard to keep straight, so what is the target at this stage?

src/app/outbound.rs
```diff
@@ -77,6 +78,10 @@ impl tap::Inspect for Endpoint {
         self.metadata.tls_status()
     }
 
+    fn route_labels<B>(&self, req: &http::Request<B>) -> Option<Arc<IndexMap<String, String>>> {
+        req.extensions().get::<super::dst::Route>().map(|r| r.labels().clone())
```
seanmonstar: Does this need to clone the `Arc`? It looks like this could probably just return `&IndexMap`.

olix0r: to return &IndexMap, we'd need to change the signature to 
```
    fn route_labels<'r, B>(&self, req: &'r http::Request<B>) -> Option<&'r IndexMap<String, String>>
```
which feels like exposing an unfortunate implementation detail in the interface

olix0r: I think that's because the request stores the Arc'd Indexmap.. I couldn't get it to work, at least.

src/app/main.rs
olix0r: The comment directly above this says: _A per-`dst::Route` layer..._

Would you prefer to explicit type annotations?

src/tap/grpc/server.rs
```diff
@@ -544,6 +544,11 @@ fn base_event<B, I: Inspect>(req: &http::Request<B>, inspect: &I) -> api::TapEve
             m.labels.insert("tls".to_owned(), tls);
             m
         }),
+        route_meta: inspect.route_labels(req).map(|labels| {
+            let mut m = api::tap_event::RouteMeta::default();
+            m.labels.extend(labels.as_ref().clone());
```
seanmonstar: I believe this will double clone the map. What about `m.labels.as_ref().iter()`?

olix0r: I _think_ the clone is necessary.  I get:

```
   --> src/tap/grpc/server.rs:549:22
    |
549 |             m.labels.extend(labels.as_ref().iter());
    |                      ^^^^^^ expected reference, found struct `std::string::String`
    |
    = note: expected type `(&std::string::String, &std::string::String)`
               found type `(std::string::String, std::string::String)`
```

So it could be `.iter().map(|(k, v)| (k.clone(), v.clone()))` ? which avoid copying the indexmap structure? Is that worth it?

PR #148, opened 2018-11-30
RFC sketch out using stacked tap server

olix0r:
This is a sketch at adding metrics to the tap grpc server.

This doesn't compile because of the hyper::Payload / tower_h2::Body dichotomy:

```
    |                              ^^^^^^^^^ the trait `hyper::body::Payload` is not implemented for `tower_h2::RecvBody`
```

@seanmonstar do you have any suggestions about the best way to resolve this?


olix0r:
@seanmonstar i tried to merge master onto this but was still struggling to get the body type signatures to be happy.  Would you like to take this branch over?

seanmonstar:
I'll take a look :P

seanmonstar:
@olix0r Here it is. I admit, I hit my head for several hours on type inference thanks to `stack::shared(whatever).push` not being able to infer `T`. I finally understood what the `phantom_data::layer()` was for.

src/app/main.rs
```diff
@@ -196,6 +195,22 @@ where
...
+        let (tap_http_metrics, tap_http_report) =
+            http_metrics::new::<TapLabels, Class>(config.metrics_retain_idle);
+
+
+        let tap_http_server = shared::stack(::api::tap::server::TapServer::new(tap_grpc))
+            .push(phantom_data::layer())
+            .push(proxy::grpc::server::layer())
+            .push(http_metrics::layer::<_, classify::Response>(
+                tap_http_metrics,
+            ))
```
seanmonstar: Actually, without a `classify::layer()`, would this ever report metrics?

seanmonstar: Or... does the `classify::layer` provide the labels, and we've already done so with `TapLabels`?

olix0r: https://github.com/linkerd/linkerd2-proxy/blob/2f2050537dd8caeaf5704e66a3fd7be89f2455a3/src/proxy/http/metrics/service.rs#L250

A classify layer is only necessary when it needs configuration. When using the default logic, the metrics layer uses a default.

seanmonstar:
@olix0r So what do you think of the changes here?

PR #149, opened 2018-12-03
Replace tower-h2 tap service with hyper

seanmonstar:
This untangles some of the HTTP/gRPC glue, providing services/stacks
that have more specific focuses. The `HyperServerSvc` now *only*
converts to a `tower::Service`, and the HTTP/1.1 and Upgrade pieces were
moved to a specific `proxy::http::upgrade::Service`.

Several stack modules were added to `proxy::grpc`, which can map request
and response bodies into `Payload`, or into `grpc::Body`, as needed.


seanmonstar:
@olix0r this should make it easier to turn the Tap server into a proper `Stack`.

src/app/main.rs
```diff
@@ -730,29 +728,41 @@ fn serve_tap<N, B>(
...
         // TODO: serve over TLS.
         bound_port
-            .listen_and_fold(server, move |mut server, (session, remote)| {
+            .listen_and_fold(new_service, move |mut new_service, (session, remote)| {
                 let log = log.clone().with_remote(remote);
-                let serve = server.serve(session).map_err(|_| ());
+                let log_clone = log.clone();
+                let serve = new_service
+                    .make_service(())
+                    .map_err(|_e| unimplemented!("serve_tap MakeError"))
```
seanmonstar: Er, woops, this snuck in.

PR #157, opened 2018-12-08
Fix test-flakey target

jonrichards:


seanmonstar:
Thanks for the PR! The change looks clear. Was something not working?

jonrichards:
> Thanks for the PR! The change looks clear. Was something not working?

The documentation says you should be able to run `make test-flakey`, but that wouldn't work (unless I was misunderstanding something - which is entirely possible).

PR #158, opened 2018-12-11
Make TLS accept logic compatible with disabled protocol detection

hawkw:
This branch changes the proxy's accept logic so that the proxy will no
longer attempt to terminate TLS on ports which are configured to skip
protocol detection. This means that a Linkerd deployment with 
`--tls optional` will no longer break server-speaks-first protocols like
 MySQL (although that traffic will not be encrypted). 

Since it's necessary to get the connection's original destination to
determine if it's on a port which should skip protocol detection, I've
moved the SO_ORIGINAL_DST call down the stack from `Server` to
`BoundPort`. However, to avoid making an additional unnecessary syscall,
the original destination is propagated to the server, along with the
information about whether or not protocol detection is enabled. This is
the approach described in
https://github.com/linkerd/linkerd2/issues/1270#issuecomment-406124236.

I've also written a new integration test for server-speaks-first
protocols with TLS enabled. This test is essentially the same as the
existing `transparency::tcp_server_first` test, but with TLS enabled for
the test proxy. I've confirmed that this fails against master.
Furthermore, I've validated this change by deploying the `booksapp` demo
with MySQL with TLS enabled, which [previously didn't work](https://github.com/linkerd/linkerd2/issues/1648#issuecomment-432867702).

Closes linkerd/linkerd2#1270

### DIFF
--- a/src/proxy/server.rs
+++ b/src/proxy/server.rs
@@ -249,9 +249,6 @@ where
             Err(never) => match never {},
         };
 
-        // We are using the port from the connection's SO_ORIGINAL_DST to
-        // determine whether to skip protocol detection, not any port that
-        // would be found after doing discovery.
         if disable_protocol_detection {
             trace!("protocol detection disabled for {:?}", orig_dst);
             let fwd = tcp::forward(io, &self.connect, &source);
--- a/src/proxy/server.rs
+++ b/src/proxy/server.rs
@@ -229,7 +229,7 @@ where
         -> impl Future<Item=(), Error=()>
     {
         let orig_dst = connection.original_dst_addr();
-        let disable_protocol_detection = connection.disable_protocol_detection();
+        let disable_protocol_detection = connection.is_protocol_detection_disabled();
 
         let log = self.log.clone()
             .with_remote(remote_addr);
@@ -241,7 +241,6 @@ where
             tls_status: connection.tls_status(),
             _p: (),
         };
-
 
         let io = match self.accept.make(&source) {
             Ok(accept) => accept.accept(connection),
--- a/src/proxy/server.rs
+++ b/src/proxy/server.rs
@@ -229,7 +229,7 @@ where
         -> impl Future<Item=(), Error=()>
     {
         let orig_dst = connection.original_dst_addr();
-        let disable_protocol_detection = connection.is_protocol_detection_disabled();
+        let disable_protocol_detection = !connection.should_detect_protocol();
 
         let log = self.log.clone()
             .with_remote(remote_addr);
--- a/src/proxy/server.rs
+++ b/src/proxy/server.rs
@@ -1,7 +1,6 @@
 use futures::{future::Either, Future};
 use http;
 use hyper;
-use indexmap::IndexSet;
 use std::{error, fmt};
 use std::net::SocketAddr;
 
@@ -9,7 +8,7 @@ use Conditional;
 use drain;
 use never::Never;
 use svc::{Stack, Service};
-use transport::{connect, tls, Connection, GetOriginalDst, Peek};
+use transport::{connect, tls, Connection, Peek};
 use proxy::http::glue::{HttpBody, HyperServerSvc};
 use proxy::protocol::Protocol;
 use proxy::tcp;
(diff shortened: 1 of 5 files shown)


hawkw:
@klingerf if you're interested in validating this change, it fixes the issues you described in https://github.com/linkerd/linkerd2/issues/1648#issuecomment-432867702 :)

hawkw:
As a side note, this change basically means that all the SO_ORIGINAL_DST code is encapsulated inside the `transport` module.

src/proxy/server.rs
```diff
@@ -251,21 +242,13 @@ where
             _p: (),
         };
 
+
```
dadjeibaah: Probably the first valuable review comment I can make with this rust code. Extra whitespace 😂

src/transport/connection.rs
```diff
@@ -14,11 +14,14 @@ use tokio::{
 
 use Conditional;
 use transport::{AddrInfo, BoxedIo, GetOriginalDst, tls};
+use indexmap::IndexSet;
 
-pub struct BoundPort {
-    inner: std::net::TcpListener,
+pub struct BoundPort<G = ()> {
+    inner: Option<std::net::TcpListener>,
```
dadjeibaah: Out of curiousity, why does this need to be wrapped inside an option?

hawkw: Yeah, so that's because we need to [move](https://doc.rust-lang.org/book/ch04-01-what-is-ownership.html#ways-variables-and-data-interact-move) the socket into [`tokio::net::TcpListener::from_std`](https://tokio-rs.github.io/tokio/tokio/net/tcp/struct.TcpListener.html#method.from_std) (which takes a `std::net::TcpListener` _by value_) here:
https://github.com/linkerd/linkerd2-proxy/blob/8a9e01ec503dbeb938e51860c3562c592c08a72e/src/transport/connection.rs#L230
but, we now call the new `new_conn` method on a reference to the struct (`&self`) in the closure passed to `Stream::and_then` here:
https://github.com/linkerd/linkerd2-proxy/blob/8a9e01ec503dbeb938e51860c3562c592c08a72e/src/transport/connection.rs#L239-L248

Since the listener can no longer be moved out of the struct by deconstructing it, as the struct still needs to be alive when the closure is invoked, we now use [`Option::take`](https://doc.rust-lang.org/std/option/enum.Option.html#method.take), which moves the value out of the `Option`, replacing it with `None`. We can use this to move the `TcpListener` out of the `BoundPort` struct without dropping the `BoundPort`:
https://github.com/linkerd/linkerd2-proxy/blob/8a9e01ec503dbeb938e51860c3562c592c08a72e/src/transport/connection.rs#L222-L223
since `Option::take` takes the receiver by a _mutable reference_ (`&mut self`). 

This is a fairly common pattern in Rust when dealing with similar situations.

src/transport/connection.rs
```diff
@@ -412,6 +518,10 @@ impl Connection {
     pub fn tls_status(&self) -> tls::Status {
         self.tls_status
     }
+
+    pub fn disable_protocol_detection(&self) -> bool {
```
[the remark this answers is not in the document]
hawkw: I had originally called this `should_disable_protocol_detection` but that seemed a little too wordy; will rename it.

hawkw:
I'm trying to figure out why the CI build for 38782e30327d010a5e9f61accf9e504abd09d0f7 is failing. It seems unrelated to the change (and builds on my machine):

```
error: overflow representing the type `*const alloc::sync::ArcInner<support::linkerd2_proxy::proxy::http::glue::HyperConnect<support::linkerd2_proxy::transport::metrics::Connect<linkerd2_timeout::Timeout<support::linkerd2_proxy::transport::connect::Target>>>>`
```

The really curious thing is that 38782e30327d010a5e9f61accf9e504abd09d0f7 only changes formatting and names, and the previous commit (with functionally equivalent code) passed CI...

hawkw:
It seems like the CI error was caused by the names of certain fields on `Connection` causing some internal storage in the compiler to overflow when a `Connection` is deeply nested in another type. What's interesting about this is that apparently the size of that storage depends on the environment where the code is being compiled, because the crash occurs on CI but I can't trigger it on my machine. 

I've fixed the issue by shortening some of the field names on `Connection`. @seanmonstar, since you previously had feedback on naming, is the new naming (as of ed31188dcf1023ed994a55423c17284f60755669) still acceptable?

PR #162, opened 2018-12-13
Workaround for Rust 1.31 crash with verbose mode enabled

hawkw:
This branch replaces the `export CARGO_VERBOSE=1` on CI release-mode
builds with the `travis_wait` script. Verbose mode was previously being
set to prevent long release-mode builds from timing out. However, there
appears to be a bug in `rustc` 1.31.0, which causes the compiler to
crash when building the proxy with verbose mode enabled. Hopefully, this
will fix the build on master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,11 +22,11 @@ jobs:
     # build.l5d.io/linkerd2-proxy/latest.txt is updated to reference the latest
     # uploaded binary.
     - if: type != pull_request
-      # Verbose is enabled so that the build doesn't timeout.
-      install: export CARGO_RELEASE=1 CARGO_VERBOSE=1
+      install: export CARGO_RELEASE=1
       script:
-        - make clean-package package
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -22,11 +22,11 @@ jobs:
...
     # uploaded binary.
     - if: type != pull_request
-      # Verbose is enabled so that the build doesn't timeout.
-      install: export CARGO_RELEASE=1 CARGO_VERBOSE=1
+      install: export CARGO_RELEASE=1
       script:
-        - make clean-package package
-        - make test
+        # travis-wait is used so that the build doesn't timeout.
+        - travis_wait 30 make clean-package package
```
seanmonstar: 30 is what, minutes?

hawkw: @seanmonstar yup. I figured that ought to be plenty?

PR #165, opened 2018-12-19
 Add Route timeouts

seanmonstar:
If a service profile specifies a timeout for a given route, this applies
a timeout waiting for responses on that route.

Timeouts show up in proxy metrics like this:

```
route_response_total{direction="outbound",dst="localhost:80",status_code="504",classification="failure",error="timeout"} 1
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ linkerd2-stack     = { path = "lib/stack" }
 linkerd2-task      = { path = "lib/task" }
 linkerd2-timeout   = { path = "lib/timeout" }
 
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", rev = "1ad7018", version = "0.1.4" }
+linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", tag = "v0.1.5" }
 
 bytes = "0.4"
 env_logger = { version = "0.5", default-features = false }
@@ -48,7 +48,7 @@ log = "0.4.1"
 indexmap = "1.0.0"
 prost = "0.4.0"
 prost-types = "0.4.0"
-rand = "0.5.1"
+rand = "0.6"
 try-lock = "0.2"
 
 # for config parsing
@@ -88,10 +88,10 @@ procinfo = "0.4.2"
 
 [dev-dependencies]
 net2 = "0.2"
-quickcheck = { version = "0.6", default-features = false }
+quickcheck = { version = "0.8", default-features = false }
 linkerd2-metrics = { path = "./lib/metrics", features = ["test_util"] }
 linkerd2-task    = { path = "lib/task", features = ["test_util"] }
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", rev = "1ad7018", version = "0.1.4", features = ["arbitrary"] }
+linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], tag = "v0.1.5" }
 flate2 = { version = "1.0.1", default-features = false, features = ["rust_backend"] }
 # `tokio-io` is needed for TCP tests, because `tokio::io` doesn't re-export
 # the `read` function.
@@ -99,5 +99,8 @@ tokio-io = "0.1.6"
 
 # Debug symbols end up chewing up several GB of disk space, so better to just
 # disable them.
+
+[profile.dev]
+debug = false
 [profile.test]
 debug = false
--- a/src/app/classify.rs
+++ b/src/app/classify.rs
@@ -1,8 +1,11 @@
+use std::borrow::Cow;
+
 use h2;
 use http;
 
 pub use proxy::http::metrics::classify::{self, layer, CanClassify};
 use proxy::http::profiles;
+use proxy::http::timeout;
 
 #[derive(Clone, Debug)]
 pub enum Request {
@@ -22,6 +25,7 @@ pub enum Eos {
     Default(http::StatusCode),
     Grpc(GrpcEos),
     Profile(Class),
+    Error(&'static str),
 }
 
 #[derive(Clone, Debug)]
@@ -34,7 +38,7 @@ pub enum GrpcEos {
 pub enum Class {
     Default(SuccessOrFailure),
     Grpc(SuccessOrFailure, u32),
-    Stream(SuccessOrFailure, String),
+    Stream(SuccessOrFailure, Cow<'static, str>),
 }
 
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
@@ -124,6 +128,10 @@ impl classify::ClassifyResponse for Response {
     type ClassifyEos = Eos;
 
     fn start<B>(self, rsp: &http::Response<B>) -> Eos {
+        if rsp.extensions().get::<timeout::ProxyTimedOut>().is_some() {
+            return Eos::Error("timeout");
+        }
+
         match self {
             Response::Default => grpc_class(rsp.headers())
                 .map(|c| Eos::Grpc(GrpcEos::NoBody(c)))
@@ -142,7 +150,7 @@ impl classify::ClassifyResponse for Response {
     }
 
     fn error(self, err: &h2::Error) -> Self::Class {
-        Class::Stream(SuccessOrFailure::Failure, format!("{}", err))
+        Class::Stream(SuccessOrFailure::Failure, h2_error(err).into())
     }
 }
 
@@ -165,11 +173,12 @@ impl classify::ClassifyEos for Eos {
                 .and_then(grpc_class)
                 .unwrap_or_else(|| Class::Grpc(SuccessOrFailure::Failure, 0)),
             Eos::Profile(class) => class,
+            Eos::Error(msg) => Class::Stream(SuccessOrFailure::Failure, msg.into())
         }
     }
 
     fn error(self, err: &h2::Error) -> Self::Class {
-        Class::Stream(SuccessOrFailure::Failure, format!("{}", err))
+        Class::Stream(SuccessOrFailure::Failure, h2_error(err).into())
     }
 }
 
@@ -186,6 +195,16 @@ fn grpc_class(headers: &http::HeaderMap) -> Option<Class> {
             };
             Class::Grpc(ok, grpc_status)
         })
+}
+
+fn h2_error(err: &h2::Error) -> String {
+    if let Some(reason) = err.reason() {
+        // This should output the error code in the same format as the spec,
+        // for example: PROTOCOL_ERROR
+        format!("h2({:?})", reason)
+    } else {
+        format!("h2({})", err)
+    }
 }
 
 // === impl Class ===
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -24,7 +24,7 @@ use proxy::{
         client, insert_target, metrics as http_metrics, normalize_uri, profiles, router, settings,
         strip_header,
     },
-    limit, reconnect, timeout,
+    limit, reconnect,
 };
 use svc::{
     self, shared,
@@ -252,7 +252,7 @@ where
                 .push(control::client::layer())
                 .push(control::resolve::layer(dns_resolver.clone()))
                 .push(reconnect::layer().with_fixed_backoff(config.control_backoff_delay))
-                .push(proxy::timeout::layer(config.control_connect_timeout))
+                .push(svc::timeout::layer(config.control_connect_timeout))
                 .push(http_metrics::layer::<_, classify::Response>(
                     ctl_http_metrics,
                 ))
@@ -315,7 +315,7 @@ where
                 // Establishes connections to remote peers (for both TCP
                 // forwarding and HTTP proxying).
                 let connect = connect::Stack::new()
-                    .push(proxy::timeout::layer(config.outbound_connect_timeout))
+                    .push(svc::timeout::layer(config.outbound_connect_timeout))
                     .push(transport_metrics.connect("outbound"));
 
                 // Instantiates an HTTP client for for a `client::Config`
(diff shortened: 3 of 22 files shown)


src/proxy/http/timeout.rs
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+
+    fn call(&mut self, req: Request<B1>) -> Self::Future {
+        self.0.call(req).or_else(|err| match err {
+            TimeoutError::Error(err) => Err(err),
+            TimeoutError::Timeout(dur) => {
+                debug!("request timed out after {:?}", dur);
+                let mut res = Response::default();
+                *res.status_mut() = StatusCode::GATEWAY_TIMEOUT;
```
seanmonstar: Things like finagle return a 503, but usually finagle is used as *part of* the actual service. Since we're a proxy, it might be more appropriate to return 504, which is described as a timeout encountered inside a proxy. I could be convinced either way.

seanmonstar:
Rebased on master after merging retries support. I think the build may fail since the `alex/timeouts` branch is missing the recent `tower-h2` patch.

olix0r:
Also, it appears that the Cargo.lock is not up-to-date...

seanmonstar:
> Do you have any thoughts about the way to highlight these errors in metrics?

Do we want a separate metric for them? Or just an additional label? Do we want to surface them by default, or just provide the ability to, such as by adding a custom response header that Route Response Classifiers could look for?

src/app/main.rs
```diff
@@ -345,13 +345,19 @@ where
...
+                // 2. A timeout is optionally enabled if the target `dst::Route`
+                //    specifies a timeout. This goes before `retry` to cap
+                //    retries.
+                // 3. Retries are optionally enabled depending on if the route
+                //    is retryable.
                 let dst_route_layer = phantom_data::layer()
                     .push(insert_target::layer())
                     .push(metrics::layer::<_, classify::Response>(retry_http_metrics.clone()))
                     .push(retry::layer(retry_http_metrics))
+                    .push(proxy::http::timeout::layer())
```
seanmonstar: Looking at the metrics output, since this layer resolves the error into an HTTP response directly, it gets counted in the `route_http_metrics` as normal response with `status_code="504"`. I wonder if that is misleading, suggesting that the other service actually returned a 504. If so, I can put this *after* the `route_http_metrics` layer.

olix0r: I haven't thought about this exhaustively, but I think that there isn't any other good way to determine that this proxy returned the 5xx (as opposed to another). In that case, we probably need to reorder this with the metrics layer, and verify that the metrics layer has a reasonable way to format the error in an unambiguous way.

As to whether this should be a label or a new metric...

On one hand, it's convention to use `response_total - request_total` to indicate the number of in-flight requests. So this would suggest using a label.

I suspect this should fit into the metrics classify api, and the classifier type would have to be aware of the types of errors it can receive so that it can format them appropriately?

adleong: I would expect `route_actual_response_total` to reflect the actual responses from the service and `route_response_total` to reflect the responses that the proxy sends back to the client.  Therefore, in the case where the request to the service times out, I would expect `route_actual_response_total` to be incremented with a label like `{error=timeout}` or something like that and `route_response_total` to be incremented with `status_code=504`.  

Does that make sense?  We can tell that the service itself never returned a 504 because `route_actual_response_total{status_code=504}` will be zero.

seanmonstar: Latest commit adds this to the metrics (text copied straight from the unit test):

```
route_response_total{direction="outbound",dst="profiles.test.svc.cluster.local:80",status_code="504",classification="failure",error="timeout"} 1
```

olix0r:
Several tests appear to hang on this branch:

```
:; cargo test -- http1::
    Finished dev [unoptimized + debuginfo] target(s) in 1.04s                                                                               
     Running target/debug/deps/linkerd2_proxy-5ddb558e37dea604

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 60 filtered out

     Running target/debug/deps/linkerd2_proxy-1c44db4914b1100b

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running target/debug/deps/discovery-e9c29a5a89d924cc

running 20 tests
test http1::absolute_uris::outbound_reconnects_if_controller_stream_ends ... ok
test http1::outbound_asks_controller_without_orig_dst ... ok
test http1::outbound_asks_controller_api ... ok
test http1::absolute_uris::outbound_asks_controller_api ... ok
test http1::absolute_uris::outbound_asks_controller_without_orig_dst ... ok
test http1::absolute_uris::outbound_error_reconnects_after_backoff ... ok
test http1::outbound_error_reconnects_after_backoff ... ok
test http1::outbound_reconnects_if_controller_stream_ends ... ok
test http1::absolute_uris::outbound_destinations_reset_on_reconnect_followed_by_add_none ... test http1::absolute_uris::outbound_destinations_reset_on_reconnect_followed_by_add_none has been running for over 60 seconds
test http1::absolute_uris::outbound_dest_concurrency_limit ... test http1::absolute_uris::outbound_dest_concurrency_limit has been running for over 60 seconds
test http1::absolute_uris::outbound_dest_limit_does_not_limit_dns ... test http1::absolute_uris::outbound_dest_limit_does_not_limit_dns has been running for over 60 seconds
test http1::absolute_uris::outbound_destinations_reset_on_reconnect_followed_by_no_endpoints_exists ... test http1::absolute_uris::outbound_destinations_reset_on_reconnect_followed_by_no_endpoints_exists has been running for over 60 seconds
test http1::absolute_uris::outbound_destinations_reset_on_reconnect_followed_by_remove_none ... test http1::absolute_uris::outbound_destinations_reset_on_reconnect_followed_by_remove_none has been running for over 60 seconds
test http1::absolute_uris::outbound_times_out ... test http1::absolute_uris::outbound_times_out has been running for over 60 seconds
test http1::outbound_dest_limit_does_not_limit_dns ... test http1::outbound_dest_limit_does_not_limit_dns has been running for over 60 seconds
test http1::outbound_dest_concurrency_limit ... test http1::outbound_dest_concurrency_limit has been running for over 60 seconds
test http1::outbound_destinations_reset_on_reconnect_followed_by_add_none ... test http1::outbound_destinations_reset_on_reconnect_followed_by_add_none has been running for over 60 seconds
test http1::outbound_destinations_reset_on_reconnect_followed_by_no_endpoints_exists ... test http1::outbound_destinations_reset_on_reconnect_followed_by_no_endpoints_exists has been running for over 60 seconds
test http1::outbound_destinations_reset_on_reconnect_followed_by_remove_none ... test http1::outbound_destinations_reset_on_reconnect_followed_by_remove_none has been running for over 60 seconds
test http1::outbound_times_out ... test http1::outbound_times_out has been running for over 60 seconds
```

seanmonstar:
Looks like gRPC errors. My guess is something about my reference to the linkerd2-proxy-api is out of sync with this repo...

seanmonstar:
Yea, so the tests are checking that if we don't get `destination::Update`s from the controller after a certain amount of time, the request should have been failed. By removing the top `BIND_TIMEOUT`, these requests just sit in the buffer waiting for `balance` to have a service ready...

We can either:

- Decide to just wait, and allow the pod to determine how long it wants to wait, in which case I'd remove many of these tests.
- Figure out some sort of "service factory timeout" that could be applied to balance or buffer or similar.

olix0r:
@seanmonstar For now, i'm in favor of removing this timeout. We don't want to stall forever when we're doing service discovery, but we should be conditioned to fail open rather than erroring hard as we do today.

seanmonstar:
@olix0r to clarify, you suggest just removing these tests?

I believe adding some sort of `timeout::ready::layer` would be pretty straight forward, that wouldn't put a timeout on the whole request, just on how long the underlying `poll_ready` has to become ready before failing.

seanmonstar:
@olix0r those tests are specifically checking that we *don't* hang if for various reasons we aren't able to determine where to route to.

For reference, I've already written a `timeout::ready::layer`, which fits in between the `balance` and `buffer` layers, which *only* times out if resolving takes too long. Looks like tests are passing again. Or I can just scrap it.

seanmonstar:
@olix0r Newest commit re-introduces the bind timeout, but it *only* applies to the time a request will wait in the buffer until the balance layer is ready. Once it's dispatched through (`call`), the timeout no longer applies. Yay or nay?

olix0r:
I'm still seeing test failures here

```
failures:
    http1::absolute_uris::outbound_dest_concurrency_limit
    http1::outbound_dest_concurrency_limit
    http2::outbound_dest_concurrency_limit
```

seanmonstar:
Hrm, it's passing locally for me. Fun!

olix0r:
Tested on both OSX and Linux (at 4bbd0c3) and I reliably get:
```
:; cargo test --test discovery outbound_dest_concurrency_limit
    Finished dev [unoptimized + debuginfo] target(s) in 2.71s                                                                                                                 
     Running target/debug/deps/discovery-bca328795f37f3fc

running 3 tests
test http2::outbound_dest_concurrency_limit ... FAILED
test http1::outbound_dest_concurrency_limit ... FAILED
test http1::absolute_uris::outbound_dest_concurrency_limit ... FAILED

failures:

---- http2::outbound_dest_concurrency_limit stdout ----
Http2 server running; addr=127.0.0.1:63302
proxy running; control=127.0.0.1:63308, inbound=127.0.0.1:63310, outbound=127.0.0.1:63313 (SO_ORIGINAL_DST=127.0.0.1:63302), metrics=127.0.0.1:63316
trying 1th destination...
trying 2th destination...
trying 3th destination...
trying 4th destination...
trying 5th destination...
trying 6th destination...
trying 7th destination...
trying 8th destination...
trying 9th destination...
trying 10th destination...
trying 11th destination...
trying 127.0.0.1:63302 directly...
trying 11th destination again...
thread 'http2::outbound_dest_concurrency_limit' panicked at 'client.get("/") expects 2xx, got "500 Internal Server Error"', tests/support/client.rs:75:9
note: Run with `RUST_BACKTRACE=1` for a backtrace.
server Listening dropped; addr=127.0.0.1:63302

---- http1::outbound_dest_concurrency_limit stdout ----
Http1 server running; addr=127.0.0.1:63300
proxy running; control=127.0.0.1:63306, inbound=127.0.0.1:63309, outbound=127.0.0.1:63312 (SO_ORIGINAL_DST=127.0.0.1:63300), metrics=127.0.0.1:63315
trying 1th destination...
trying 2th destination...
trying 3th destination...
trying 4th destination...
trying 5th destination...
trying 6th destination...
trying 7th destination...
trying 8th destination...
trying 9th destination...
trying 10th destination...
trying 11th destination...
trying 127.0.0.1:63300 directly...
trying 11th destination again...
thread 'http1::outbound_dest_concurrency_limit' panicked at 'client.get("/") expects 2xx, got "500 Internal Server Error"', tests/support/client.rs:75:9
server Listening dropped; addr=127.0.0.1:63300

---- http1::absolute_uris::outbound_dest_concurrency_limit stdout ----
Http1 server running; addr=127.0.0.1:63301
proxy running; control=127.0.0.1:63307, inbound=127.0.0.1:63311, outbound=127.0.0.1:63314 (SO_ORIGINAL_DST=127.0.0.1:63301), metrics=127.0.0.1:63317
trying 1th destination...
trying 2th destination...
trying 3th destination...
trying 4th destination...
trying 5th destination...
trying 6th destination...
trying 7th destination...
trying 8th destination...
trying 9th destination...
trying 10th destination...
trying 11th destination...
trying 127.0.0.1:63301 directly...
trying 11th destination again...
thread 'http1::absolute_uris::outbound_dest_concurrency_limit' panicked at 'client.get("/") expects 2xx, got "500 Internal Server Error"', tests/support/client.rs:75:9
server Listening dropped; addr=127.0.0.1:63301


failures:
    http1::absolute_uris::outbound_dest_concurrency_limit
    http1::outbound_dest_concurrency_limit
    http2::outbound_dest_concurrency_limit

test result: FAILED. 0 passed; 3 failed; 0 ignored; 0 measured; 30 filtered out

error: test failed, to rerun pass '--test discovery'
```

seanmonstar:
I'm still shocked it passes locally, as I've found why it *definitely* shouldn't. The `ready` timeout was causing the `buffer` to be closed for all later requests. So I'm going to pull that back out, and just see what I can do for the discovery tests.

seanmonstar:
Many of the discovery tests were exercising things like not exceeding our gRPC Destination Service stream count, or that reconnects make the proxy "forget" old addresses, and in both cases, the request will wait for a capacity to free up, or an update from the controller. Thus, they were hitting our `BIND_TIMEOUT`.

I've changed *those* tests to instead expect a client side timeout, asserting that the request doesn't proceed past the our `balance` step, and timing out the request directly in the unit test.

There was one test that was specifically testing the `BIND_TIMEOUT`, that a destination with no updates would eventually timeout. I've slapped `#[ignore]` on that, since I figure we will eventually want to add in some sort of destination-acquisition timeout.

Cargo.toml
```diff
@@ -33,7 +33,7 @@ linkerd2-stack     = { path = "lib/stack" }
 linkerd2-task      = { path = "lib/task" }
 linkerd2-timeout   = { path = "lib/timeout" }
 
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", rev = "1ad7018", version = "0.1.4" }
+linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", branch = "sean/timeouts" } #tag = "v0.1.4", version = "0.1.4" }
```
seanmonstar: Should point back at master (or a tag) before merging.

src/app/classify.rs
```diff
@@ -124,6 +128,10 @@ impl classify::ClassifyResponse for Response {
     type ClassifyEos = Eos;
 
     fn start<B>(self, rsp: &http::Response<B>) -> Eos {
+        if rsp.extensions().get::<timeout::ProxyTimedOut>().is_some() {
+            return Eos::Error("timeout");
+        }
```
olix0r: This seems fine, though you might also consider doing this slightly differently: Instead of making the classify module enumerate all of the types of errors it may observfe, we could instead have modules add a `classify::ProxyError(&'static str)` marker onto responses -- so this would be able to work with arbitrary error types in the stack, not just timeouts.

This then becomes something to the effect of:
```rust
        if let Some(ProxyError(msg)) = rsp.extensions().get::<ProxyError>() {
            return Eos::Error(msg);
        }
````

seanmonstar: That'd seem to have `app` types present in `proxy` modules... Unless you mean just a general `proxy::http::classify::ProxyError` thingy?

seanmonstar:
Oh noes, seems I was bitten by #170 not yet being merged, and thus `quickcheck` versions mismatched...

src/app/classify.rs
olix0r: ah, yeah... good point.

seanmonstar:
Updated, green, just needs final r+.

PR #169, opened 2019-01-05
Update trust-dns-resolver to fix UDP socket leak

hawkw:
An upstream bug in the `trust-dns-proto` library can cause
`trust-dns-resolver` to leak UDP sockets when DNS queries time out. This
issue appears to be the cause of the memory leak described in
linkerd/linkerd2#2012.

This branch updates the `trust-dns` dependency to pick up the change in
bluejekyll/trust-dns#635, which fixes the UDP socket leak.

I confirmed that the socket leak was fixed by modifying the proxy to
hard-code a 0-second DNS timeout, sending requests to the proxy's
outbound listener, and using

``` lsof -p $(pgrep linkerd2-proxy) ```

to count the number of open UDP sockets. On master, every request to a
different DNS name that times out leaves behind an additional open UDP
socket, which show up in `lsof`, while on this branch, only TCP sockets
remain open after the request ends.

In addition, I'm running a test in GCP to watch the memory and file
descriptor use of the proxy over a long period of time. This is still in
progress, but given the above, I strongly believe this branch fixes the
leak.

Fixes linkerd/linkerd2#2012.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,6 +30,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "autocfg"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "backtrace"
 version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Cargo.lock [hunk 1]
```diff
@@ -934,11 +939,87 @@ dependencies = [
  "winapi 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
+[[package]]
+name = "rand"
+version = "0.6.3"
```
seanmonstar: Rand 0.4, 0.5, and 0.6... 😭

hawkw: Oh boy, I hadn't noticed. It would be nice if we could do something about that...

hawkw: Looks like the `rand` 0.4 dependency is pulled in by `tempdir`, which is a dependency of `prost-build`. The `tempdir` crate is deprecated in favour of `tempfile`, but we use it in our tests as well (to pull in fewer dependencies).

I'll look at a patch to `prost-build` replacing `tempdir` with `tempfile`, and hopefully we can get rid of `rand` 0.4

hawkw: @seanmonstar if we merge both this branch and #170, we will depend exclusively on the latest `rand` (0.6.3). 

Sorry it took a bit, the dependency update was a surprisingly large amount of work.

PR #170, opened 2019-01-08
Update dependencies to remove duplicate `rand` versions

hawkw:
This branch updates the proxy's dependencies to eliminate multiple
incompatible versions of the `rand` crate. 

Previously we depended on both `rand` 0.4.3 and 0.5.1. After this
branch, we depend on `rand` 0.6.3 (the latest) and 0.5.1. However, PR
#169 will also update `trust-dns-resolver` (the only crate depending on
`rand` 0.5.1) to a version that depends on `rand` 0.6.3, so once both
this branch and #169 are merged, we will depend only on `rand` 0.6.3.

Note that this is a fairly large change to `Cargo.toml` --- it was
necessary to update many of the proxy's dependencies in order to
consolidate on one `rand` version. Additionally, I had to push branches
of some of those dependencies in order to update their `rand`
dependency, so it's currently necessary to patch some of the proxy's
dependencies. When those branches merge upstream, the patches can be
removed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,3 +104,6 @@ tokio-io = "0.1.6"
 debug = false
 [profile.test]
 debug = false
+
+[patch.'https://github.com/tower-rs/tower-grpc']
+prost-build = { git = "https://github.com/hawkw/prost", branch = "s/tempdir/tempfile"}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ log = "0.4.1"
 indexmap = "1.0.0"
 prost = "0.4.0"
 prost-types = "0.4.0"
-rand = "0.6"
+rand = "0.6.3"
 try-lock = "0.2"
 
 # for config parsing
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,7 +83,7 @@ untrusted = "0.6"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 libc = "0.2"
-inotify = { git = "https://github.com/inotify-rs/inotify" }
+inotify = "0.6"
 procinfo = "0.4.2"
 
 [dev-dependencies]
@@ -108,3 +108,4 @@ debug = false
 
 [patch.crates-io]
 prost-build = { git = "https://github.com/hawkw/prost", branch = "s/tempdir/tempfile"}
+inotify = { git = "https://github.com/hawkw/inotify", branch = "eliza/more-flexible-event-stream", rev = "a2d73d61ba61e993b71053e06b9dfe6bc848ef61" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,7 +83,7 @@ untrusted = "0.6"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 libc = "0.2"
-inotify = "0.6"
+inotify = { git = "https://github.com/hawkw/inotify", rev = "a2d73d61ba61e993b71053e06b9dfe6bc848ef61" }
 procinfo = "0.4.2"
 
 [dev-dependencies]
@@ -110,5 +110,4 @@ debug = false
 prost = { git = "https://github.com/hawkw/prost", branch = "s/tempdir/tempfile"}
 prost-derive = { git = "https://github.com/hawkw/prost", branch = "s/tempdir/tempfile"}
 prost-build = { git = "https://github.com/hawkw/prost", branch = "s/tempdir/tempfile"}
-prost-types = { git = "https://github.com/hawkw/prost", branch = "s/tempdir/tempfile"}
 inotify = { git = "https://github.com/hawkw/inotify", rev = "a2d73d61ba61e993b71053e06b9dfe6bc848ef61" }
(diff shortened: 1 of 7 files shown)


olix0r:
@hawkw It looks like this is blocked on a variety of PRs to other repos. Are there open PRs for tower, etc? Would you mind including links so these related PRs are easily discoverable?

hawkw:
@olix0r I've opened the upstream PRs:
- [ ] danburkert/prost#143
- [ ] tower-rs/tower#142
- [ ] linkerd/linkerd2-proxy-api#21

olix0r:
Cargo.lock is out-of-date.

olix0r:
```
$ make test
cargo fetch --locked
    Updating git repository `https://github.com/hawkw/prost`
    Updating git repository `https://github.com/linkerd/linkerd2-proxy-api`
error: failed to load source for a dependency on `linkerd2-proxy-api`
Caused by:
  Unable to update https://github.com/linkerd/linkerd2-proxy-api?branch=eliza/bump-quickcheck#bf88617c
```

olix0r:
It would be good to move this issue forward. It looks like prost hasn't taken the update yet, but also the maintainer seems not to have had much public activity recently, so perhaps it's best to roll forward with a patched dependency.

hawkw:
Sorry, I'll fix the lockfile & we can try to move forward on this.

hawkw:
@olix0r interesting; I'm trying to un-break the lockfile, but it looks like a clean `make test` of this branch on my machine builds fine. Pushed a new empty commit to see what happens.

hawkw:
Oh, *now* I see what the problem is:
screen shot 2019-01-23 at 3 02 43 pm
but don't worry, I fixed it:
screen shot 2019-01-23 at 3 02 48 pm

And of course it was building locally because I had a checkout of the deleted branch in my cargo cache. 

I've restarted the build and hopefully this should succeed.

hawkw:
Now that #165 has merged we can just depend on the head version of proxy-api I think?

hawkw:
@olix0r rebased this against master, _should_ build fine now.

lib/fs-watch/src/lib.rs
```diff
@@ -193,8 +193,12 @@ pub mod inotify {
 
     impl WatchStream {
         pub fn new(paths: Vec<PathBuf>) -> Result<Self, io::Error> {
+            // XXX: this is unfortunately now necessary due to
+            // https://github.com/inotify-rs/inotify/issues/120
+            static mut BUF: [u8; 32] = [0; 32];
+
             let mut inotify = Inotify::init()?;
-            let stream = inotify.event_stream();
+            let stream = inotify.event_stream(unsafe { &mut BUF });
```
olix0r: I understand that this is just cargo-culted from their example, but I don't understand why, in either case, this `unsafe` is okay.

hawkw: I...don't actually think this is safe --- AFAICT, multiple calls to `WatchStream::new` will cause them to share one buffer. This won't hurt us too bad since we don't make multiple `WatchStream`s currently, but I don't like the potential gotcha.

I kind of really hate that API in `inotify` and am a bit mad that they hacked up my code (which was _fine_) to basically require you to do this.

Let me see if there's a better workaround.

hawkw: As of 9d9884ed8ffdb23962f70ed6b5983d52cb62b7bf, every `WatchStream` gets its own buffer on the heap and all is right in the world once again.

hawkw: Oh derp never mind, the only reason that branch built for me locally is because  this code is only built on Linux. hmm, I am not sure if there is a reasonable safe solution for this...

Cargo.toml
```diff
@@ -83,7 +83,7 @@ untrusted = "0.6"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 libc = "0.2"
-inotify = { git = "https://github.com/inotify-rs/inotify" }
+inotify = { git = "https://github.com/hawkw/inotify", branch = "eliza/fix-bad-api" }
```
olix0r: Would you mind renaming this branch to something less judgey? I'd prefer to avoid even the appearance of picking on the inotify project, especially since they've very amenable to feedback in the past.

olix0r: Or, should we just depend on a revision so that the build doesn't depend on the branch's existence?

hawkw: we should probably depend on a revision, but I'll rename the branch regardless. fwiw, the `inotify` maintainer also seems to think this API is bad, but yeah, I can see how it comes off as too judge-y.

hawkw:
@olix0r agh, wiill do!

hawkw:
@olix0r Okay, I think this should be non-conflicty and ready to merge now as of 77742bc780a5a901e92ce26322199dc4c8f3cb82.

I think while doing all the merges, I inadvertently ended up having two versions of `prost` in the tree, but I fixed that along the way.

hawkw:
@olix0r alright, _finally_ got this to build --- I forgot that the `inotify` dependency is Linux-only, so I spent a while mystified about why it was building fine on my machine. Should be merge-able now (for real this time).

PR #173, opened 2019-01-17
Make outbound router honor `l5d-dst-override` header

kleimkuhler:
# Problem
It is valuable to inject an ingress controller with LD2. As ingress, the host
header tends to be the external host and most controllers just route directly
back to the IP address of an endpoint. Unfortunately, this makes it so that
Linkerd cannot discover the destination (as that works off the host header).

# Solution
The outbound router now recognizes request destinations by preferrentially
choosing the `l5d-dst-override` header if it present. This header takes
precedent over the Host header, original dest, etc.

In addition, both the inbound and outbound routers strip this header. The
outbound router strips it after it has used it to determine the authority.
This is done with the `strip_header` layer that was added to the stack.

# Validation
There are integration tests for this behavior in the discovery tests. There
are tests for http1 and http2. The tests test the following scenarios:
- the outbound router routes based on the header when it is present
- the override header applies to only requests with the override header
- the inbound router never honors the override header
- both routers strip the header before forwarding the request

Fixes linkerd/linkerd2#1998

Co-Authored-By: Eliza Weisman <eliza@buoyant.io>

### DIFF
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -423,13 +423,18 @@ where
                     .push(limit::layer(MAX_IN_FLIGHT))
                     .push(strip_header::layer(super::DST_OVERRIDE_HEADER))
                     .push(router::layer(|req: &http::Request<_>| {
-                        let addr = super::http_request_l5d_override_dst_addr(req)
-                            .or_else(|_| super::http_request_authority_addr(req))
+                        let addr = super::http_request_authority_addr(req)
                             .or_else(|_| super::http_request_host_addr(req))
                             .or_else(|_| super::http_request_orig_dst_addr(req))
                             .ok();
                         debug!("outbound addr={:?}", addr);
-                        addr
+                        super::http_request_l5d_override_dst_addr(req)
+                            .ok()
+                            .map(|override_addr| {
+                                debug!("outbound dst-override={:?}", override_addr);
+                                override_addr
+                            })
+                            .or(addr)
                     }))
                     .make(&router::Config::new("out addr", capacity, max_idle_age))
                     .map(shared::stack)
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -432,13 +432,11 @@ where
                         super::http_request_l5d_override_dst_addr(req)
                             .ok()
                             .map(|override_addr| {
-                                if log_enabled!(log::Level::Debug) {
-                                    debug!(
-                                        "outbound addr={:?}; dst-override={:?}",
-                                        addr(),
-                                        override_addr
-                                    );
-                                }
+                                debug!(
+                                    "outbound addr={:?}; dst-override={:?}",
+                                    addr(),
+                                    override_addr
+                                );
                                 override_addr
                             })
                             .or_else(|| {
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -423,24 +423,17 @@ where
                     .push(limit::layer(MAX_IN_FLIGHT))
                     .push(strip_header::layer(super::DST_OVERRIDE_HEADER))
                     .push(router::layer(|req: &http::Request<_>| {
-                        let addr = || {
-                            super::http_request_authority_addr(req)
-                                .or_else(|_| super::http_request_host_addr(req))
-                                .or_else(|_| super::http_request_orig_dst_addr(req))
-                                .ok()
-                        };
                         super::http_request_l5d_override_dst_addr(req)
                             .ok()
                             .map(|override_addr| {
-                                debug!(
-                                    "outbound addr={:?}; dst-override={:?}",
-                                    addr(),
-                                    override_addr
-                                );
+                                debug!("outbound addr={:?}; dst-override", override_addr);
                                 override_addr
                             })
                             .or_else(|| {
-                                let addr = addr();
+                                let addr = super::http_request_authority_addr(req)
+                                    .or_else(|_| super::http_request_host_addr(req))
+                                    .or_else(|_| super::http_request_orig_dst_addr(req))
+                                    .ok();
                                 debug!("outbound addr={:?}", addr);
                                 addr
                             })
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -22,6 +22,7 @@ use proxy::{
     self, buffer,
     http::{
         client, insert_target, metrics as http_metrics, normalize_uri, profiles, router, settings,
+        strip_header,
     },
     limit, reconnect, timeout,
 };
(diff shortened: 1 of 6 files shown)


src/proxy/http/h1.rs
```diff
@@ -49,17 +49,7 @@ pub fn set_origin_form(uri: &mut Uri) {
...
-             host.to_str().ok()
-                .and_then(|s| {
-                    if s.is_empty() {
-                        None
-                    } else {
-                        s.parse::<Authority>().ok()
-                    }
-                })
-        })
+    super::authority_from_header(req, HOST)
```
hawkw: This was generalized into the `authority_from_header` function in `proxy::http`, as we re-used it for parsing the override header.

src/app/main.rs
```diff
@@ -420,13 +421,20 @@ where
...
                     .push(limit::layer(MAX_IN_FLIGHT))
+                    .push(strip_header::layer(super::DST_OVERRIDE_HEADER))
                     .push(router::layer(|req: &http::Request<_>| {
                         let addr = super::http_request_authority_addr(req)
                             .or_else(|_| super::http_request_host_addr(req))
                             .or_else(|_| super::http_request_orig_dst_addr(req))
                             .ok();
                         debug!("outbound addr={:?}", addr);
-                        addr
+                        super::http_request_l5d_override_dst_addr(req)
```
seanmonstar: If the override always takes precedent, perhaps we should check it first, before the other checks?

hawkw: @seanmonstar that was the original implementation (see 2493c85): https://github.com/linkerd/linkerd2-proxy/blob/2493c85d1c091afea2e44129fa8377d7fedd5517/src/app/main.rs#L423-L431
I thought that for debugging purposes, it would be valuable to log the non-overridden addr _as well_ as the overridden one (similar to how the inbound router logs both the value of the `l5d-canonical-dst` header _and_ the final dst addr), so I encouraged @kleimkuhler to rewrite this closure so that we always determine the pre-override addr before checking for the override.

I think this can be changed back if you don't think logging both is necessary.

seanmonstar: The logging *could* be useful. In that case, though, it can be structured so that the extra parsing only occurs if actually logging at `debug` level. That way, if no one is looking, why pay extra.

hawkw: Okay, in that case, I propose something like:
```rust
 .push(router::layer(|req: &http::Request<_>| { 
     let addr = || {
         super::http_request_authority_addr(req)) 
            .or_else(|_| super::http_request_host_addr(req)) 
            .or_else(|_| super::http_request_orig_dst_addr(req)) 
            .ok(); 
     };
    super::http_request_l5d_override_dst_addr(req) 
        .map(|override_addr| {
            if log_enabled!(log::Level::Debug) {
                debug!("outbound addr={:?}; override_dst={:?}", addr(), override_addr); 
            }
            override_addr
        })
        .or_else(|_| {
            let addr = addr();
            debug!("outbound addr={:?}", addr);
            addr
        })
 })
```
This way, the pre-override addr is only computed if we're actually going to log it.

Unfortunately this does make the code a tad more complex...perhaps some of this could be factored out into functions or something.

kleimkuhler: Would we still want to keep the log line for `dst-override` outside of `log_enabled!`?

src/app/main.rs
```diff
@@ -423,18 +423,29 @@ where
...
+                            super::http_request_authority_addr(req)
+                                .or_else(|_| super::http_request_host_addr(req))
+                                .or_else(|_| super::http_request_orig_dst_addr(req))
+                                .ok()
+                        };
                         super::http_request_l5d_override_dst_addr(req)
                             .ok()
                             .map(|override_addr| {
-                                debug!("outbound dst-override={:?}", override_addr);
+                                if log_enabled!(log::Level::Debug) {
```
hawkw: @kleimkuhler turns out I mislead you here a bit --- I just checked the `log` crate's macros, and the formatting _is_ already guarded by a similar `if` statement. So I think it _is_ okay to remove the `if` and just say
```rust 
debug!(
    "outbound addr={:?}; dst-override={:?}",
    addr(),
    override_addr
);
```

My bad, sorry about that.

kleimkuhler: Oh cool sounds good! Thanks for actually checking; it probably would have been good for me to do the same instead of just asking.

hawkw: Hah, it's definitely my fault! :P

src/app/main.rs
```diff
@@ -420,13 +421,29 @@ where
...
+                                    "outbound addr={:?}; dst-override={:?}",
+                                    addr(),
+                                    override_addr
+                                );
+                                override_addr
+                            })
+                            .or_else(|| {
+                                let addr = addr();
+                                debug!("outbound addr={:?}", addr);
+                                addr
```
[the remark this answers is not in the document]
kleimkuhler: There are two places where we may need to compute `addr`. The debug line above also needs the same logic, but neither of the cases is guaranteed to happen.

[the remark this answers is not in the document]
hawkw: Yeah, the intention behind the closure was to avoid repeating the same code for logging the pre-override addr for debugging and for actually getting the addr when the override header is not set.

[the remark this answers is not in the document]
olix0r: We should not log the addr that we're not using.  Let's make `outbound addr=...` always reference the address that's being used for routing, and not contextual.

We can either do that by having a line like `outbound addr=...; dst-override`; or we can have a log line "outbound dst-override=..."; and always log the addr being used with `outbound addr=...`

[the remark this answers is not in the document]
hawkw: @olix0r Okay!   thought logging what the addr _would_ have been if it were not overridden would be useful for debugging purposes, but not logging it simplifies things significantly.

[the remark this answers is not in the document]
olix0r: Yeah, let's optimize for simplicity and ensure that we (1) log the dst addr being used and (2) indicate whether it was or was not from an override.

src/app/main.rs
```diff
@@ -420,13 +421,22 @@ where
...
+                            .ok()
+                            .map(|override_addr| {
+                                debug!("outbound addr={:?}; dst-override", override_addr);
+                                override_addr
+                            })
+                            .or_else(|| {
+                                let addr = super::http_request_authority_addr(req)
+                                    .or_else(|_| super::http_request_host_addr(req))
+                                    .or_else(|_| super::http_request_orig_dst_addr(req))
+                                    .ok();
```
olix0r: nit: I think you only have to have an `.ok()` once in this callchain (also one in L427)

kleimkuhler: Yep I can change it. It will change how it get's printed in `debug!` if that is okay - `outbound addr=Ok(...)` instead of `outbound addr=Some(...)`

PR #176, opened 2019-01-23
Remove flakiness from 'profiles' tests

seanmonstar:
By polling /metrics looking for the label provided in the service
profile, we can prevent the unit test from proceeding too early.

Signed-off-by: Sean McArthur <sean@buoyant.io>


tests/profiles.rs
```diff
@@ -85,9 +85,12 @@ macro_rules! profile_test {
         dst_tx.send_addr(srv.addr);
 
         let profile_tx = ctrl.profile_tx(host);
-        let routes = vec![$(
-            $route,
-        ),+];
+        let routes = vec![
+            controller::route()
```
hawkw: Maybe a comment here explaining what the test route is for?

PR #180, opened 2019-01-28
Add l5d-server-id header for outbound responses

seanmonstar:
If we know the meshed TLS identity of an endpoint, and a connection successfully uses TLS to that endpoint, then an `l5d-server-id` header is included in the response headers sent back to the underlying service.

Closes https://github.com/linkerd/linkerd2/issues/2124


src/app/outbound.rs
```diff
@@ -263,3 +263,162 @@ pub mod orig_proto_upgrade {
...
+                        Ok(value) => {
+                            debug!("l5d-server-id enabled for {:?}", endpoint);
+                            return Ok(svc::Either::A(Service {
+                                inner: svc,
+                                value,
+                                _marker: PhantomData,
+                            }));
+                        },
+                        Err(_err) => {
+                            warn!("l5d-server-id identity is invalid header: {:?}", endpoint);
```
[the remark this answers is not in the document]
seanmonstar: It's trying to say that the header would be invalid, not that the identity itself is...

And the error is just [`InvalidHeaderValue`](https://docs.rs/http/0.1.*/http/header/struct.InvalidHeaderValue.html), which won't say anything extra.

src/app/outbound.rs
```diff
@@ -263,3 +263,162 @@ pub mod orig_proto_upgrade {
...
+                        }
+                    },
+                    Conditional::None(_why) => {
+                        warn!("endpoint tls is some, but no identity: {:?}", endpoint);
```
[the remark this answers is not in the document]
seanmonstar: Yes, this should be a bug. I can promote it to `error!` level...

src/app/outbound.rs
```diff
@@ -263,3 +263,162 @@ pub mod orig_proto_upgrade {
...
+
+        fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+            self.inner.poll_ready()
+        }
+
+        fn call(&mut self, req: Req) -> Self::Future {
+            let fut = self.inner.call(req);
+
+            ResponseFuture {
+                inner: fut,
```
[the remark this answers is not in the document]
seanmonstar: I dunno, it just seemed clearer to call it a future, since both the service and future have `inner` fields...

src/proxy/http/glue.rs
```diff
@@ -220,6 +226,11 @@ where
         let transport = try_ready!(self.inner.poll());
         let connected = hyper_connect::Connected::new()
             .proxy(self.absolute_form);
+        let connected = if let Conditional::Some(()) = transport.tls_status() {
+            connected.extra(ClientUsedTls(()))
```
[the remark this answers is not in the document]
seanmonstar: If we have a use later, we can change this. For now, this prevents an allocation if TLS is not used...

src/proxy/http/strip_header.rs
```diff
@@ -1,91 +1,180 @@
...
 
 /// Wraps HTTP `Service` `Stack<T>`s so that a given header is removed from a
-/// request.
-#[derive(Debug, Clone)]
-pub struct Layer<H> {
+/// request or response.
+#[derive(Clone, Debug)]
+pub struct Layer<H, R> {
     header: H,
+    _req_or_res: PhantomData<fn(R)>,
```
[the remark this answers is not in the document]
seanmonstar: It prevents needing to duplicate the `Layer` and `Stack` boilerplate.

src/proxy/http/glue.rs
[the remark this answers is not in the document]
hawkw: Ah, yeah, that's a fair point 👍

src/app/outbound.rs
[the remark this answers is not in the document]
hawkw: what about something like this?
```suggestion
                            warn!("l5d-server-id identity header is invalid: {:?}", endpoint);
```

src/app/outbound.rs
```diff
@@ -361,8 +361,13 @@ pub mod server_id {
                             warn!("l5d-server-id identity header is invalid: {:?}", endpoint);
                         }
                     },
-                    Conditional::None(_why) => {
-                        warn!("endpoint tls is some, but no identity: {:?}", endpoint);
+                    Conditional::None(why) => {
+                        error!("endpoint tls is some, but no identity ({:?}): {:?}", why, endpoint);
+
+                        // This is a bug, so panic in tests!
+                        if cfg!(debug_assertions) {
```
hawkw: Nice --- I wonder if there are other places where we might want to do this behavior. Maybe eventually we ought to add a `bug!` macro or something.

olix0r: There is/have been places where we've done `debug_assert!(false, "here's a bug you hit")`

src/transport/tls/mod.rs
```diff
@@ -73,3 +73,7 @@ impl fmt::Display for Status {
     }
 }
 
+/// Fetch the `tls::Status` of this type.
+pub trait TlsStatus {
```
olix0r: nit/tioli: should this be called `HasStatus`? This might be less ambiguous as `tls::HasStatus` and `tls::Status`, since (I think?) we sometimes import tls::Status as TlsStatus...

not a blocker

src/app/main.rs
```diff
@@ -337,6 +337,7 @@ where
                 let endpoint_stack = client_stack
                     .push(buffer::layer(MAX_IN_FLIGHT))
                     .push(settings::router::layer::<Endpoint, _>())
+                    .push(server_id::layer())
```
olix0r: I'd suggest that we explicitly clear the header "before" setting it in this layer, so there's no potential that the remote could have sent it to forge the proxy's value. Unless there's something else preventing this?

src/transport/tls/mod.rs
seanmonstar: I'm more used to traits being "things", typically named after their methods... And many of our `Has*` traits have a method that returns an `Option<T>`, making me feel like `HasStatus` implies the status is optional... 🤷‍♂️

src/app/main.rs
seanmonstar: Done.

src/transport/tls/mod.rs
olix0r: although, `tls::Status` is _effectively_ an optional, since it's Conditional<T> -- not a big deal either way

seanmonstar: That's a good point, I had forgotten. I'll make the change.

PR #183, opened 2019-01-31
Set proxy_id in calls to Get and GetProfile

adleong:
Pick up API changes to the proxy-api where Get and GetProfiles now accept a proxy_id parameter.  We set this parameter to be equal to the proxy's TLS pod identity, which is configured through an environment variable.  Sending this value to the proxy-api allows it to return more tailored results, for example, by looking for service profiles in the proxy's namespace.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ linkerd2-stack     = { path = "lib/stack" }
 linkerd2-task      = { path = "lib/task" }
 linkerd2-timeout   = { path = "lib/timeout" }
 
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", tag = "v0.1.5" }
+linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", rev = "36f54ad" }
 
 bytes = "0.4"
 env_logger = { version = "0.5", default-features = false }
@@ -91,7 +91,7 @@ net2 = "0.2"
 quickcheck = { version = "0.8", default-features = false }
 linkerd2-metrics = { path = "./lib/metrics", features = ["test_util"] }
 linkerd2-task    = { path = "lib/task", features = ["test_util"] }
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], tag = "v0.1.5" }
+linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], rev = "36f54ad" }
 flate2 = { version = "1.0.1", default-features = false, features = ["rust_backend"] }
 # `tokio-io` is needed for TCP tests, because `tokio::io` doesn't re-export
 # the `read` function.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ linkerd2-stack     = { path = "lib/stack" }
 linkerd2-task      = { path = "lib/task" }
 linkerd2-timeout   = { path = "lib/timeout" }
 
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", rev = "36f54ad" }
+linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", tag = "v0.1.6" }
 
 bytes = "0.4"
 env_logger = { version = "0.5", default-features = false }
@@ -91,7 +91,7 @@ net2 = "0.2"
 quickcheck = { version = "0.8", default-features = false }
 linkerd2-metrics = { path = "./lib/metrics", features = ["test_util"] }
 linkerd2-task    = { path = "lib/task", features = ["test_util"] }
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], rev = "36f54ad" }
+linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], tag = "v0.1.6" }
 flate2 = { version = "1.0.1", default-features = false, features = ["rust_backend"] }
 # `tokio-io` is needed for TCP tests, because `tokio::io` doesn't re-export
 # the `read` function.
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -549,7 +549,7 @@ impl TestEnv {
         let mut values: HashMap<&'static str, String> = Default::default();
         values.insert(ENV_PROXY_ID, "foo.deployment.default.linkerd-managed.linkerd.svc.cluster.local".into());
         Self {
-            values: values,
+            values,
         }
     }
 
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -93,6 +93,8 @@ pub struct Config {
 
     pub namespaces: Namespaces,
 
+    pub proxy_id: String,
+
     /// Optional minimum TTL for DNS lookups.
     pub dns_min_ttl: Option<Duration>,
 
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -91,7 +91,7 @@ net2 = "0.2"
 quickcheck = { version = "0.8", default-features = false }
 linkerd2-metrics = { path = "./lib/metrics", features = ["test_util"] }
 linkerd2-task    = { path = "lib/task", features = ["test_util"] }
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], tag = "v0.1.5" }
+linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], rev = "36f54ad" }
```
seanmonstar: We should probably get a new tag and update here before merging.

src/app/main.rs
```diff
@@ -280,6 +280,11 @@ where
         // This channel is used to move the task.
         let (resolver_bg_tx, resolver_bg_rx) = futures::sync::oneshot::channel();
 
+        let proxy_id = match config.tls_settings {
+            Conditional::Some(ref settings) => settings.pod_identity.as_ref().to_string(),
+            Conditional::None(_) => String::new(),
+        };
```
olix0r: I think the main question i have for this PR is whether the `proxy_id` should be explicitly tie dto the TLS settings or whether it should be provided as its own configuration, independently of the tls settings (though, both values would be the same).

The downside of the current approach is that an ID will only be set when there are TLS settings for the proxy. We expect this to become the default behavior in the ~2.3 time frame, but we may want a way to provide this configuration when TLS is disabled?

adleong: My plan for this was to make a controller change where LINKERD2_PROXY_TLS_POD_IDENTITY would always be set, even when TLS is disabled.

hawkw:
It looks like the CI build for this branch is currently failing because
https://github.com/linkerd/linkerd2-proxy/blob/d3e7f8bb69c0f2ed58bcbedc8acecf3079344e9b/tests/support/controller.rs#L62-L65
needs to be updated to include the new `proxy_id` field.

src/app/main.rs
olix0r: I don't think that this will work as you hope. `tls_settings` is only set to Some when _all_ of the TLS settings have been properly configured:

https://github.com/linkerd/linkerd2-proxy/blob/3832b49a65b9feda168cf4b96f2ed541d6fc2434/src/app/config.rs#L379-L439

To make this work, you should modify the Config type to have a `proxy_id` field which is set despite the state of the other TLS configuration.

adleong: ack, ok, thanks

src/app/config.rs
```diff
@@ -546,8 +546,10 @@ impl Strings for Env {
 
 impl TestEnv {
     pub fn new() -> Self {
+        let mut values: HashMap<&'static str, String> = Default::default();
```
hawkw: I don't think this type annotation is strictly necessary.

src/app/config.rs
```diff
@@ -540,8 +546,10 @@ impl Strings for Env {
...
+        values.insert(ENV_PROXY_ID, "foo.deployment.default.linkerd-managed.linkerd.svc.cluster.local".into());
         Self {
-            values: Default::default(),
+            values: values,
```
hawkw: Nit: this can actually just be 
```suggestion
            values,
```
Rust has a nice shorthand for initializing a field from a name that's in scope.

src/control/destination/background/mod.rs
```diff
@@ -318,12 +320,13 @@ where
...
 impl NewQuery {
 
-    fn new(namespaces: Namespaces, suffixes: Vec<dns::Suffix>, concurrency_limit: usize) -> Self {
+    fn new(namespaces: Namespaces, suffixes: Vec<dns::Suffix>, concurrency_limit: usize, proxy_id: String) -> Self {
         Self {
             namespaces,
             suffixes,
             concurrency_limit,
             active_query_handle: Arc::new(()),
+            proxy_id,
```
hawkw: Since this is cloned into every request, I wonder if it ought to be reference-counted?

src/app/config.rs
adleong: it seems like it is

```
error[E0282]: type annotations needed
   --> src/app/config.rs:550:9
    |
549 |         let mut values = Default::default();
    |             ---------- consider giving `values` a type
550 |         values.insert(ENV_PROXY_ID, "foo.deployment.default.linkerd-managed.linkerd.svc.cluster.local".into());
    |         ^^^^^^ cannot infer type
    |
    = note: type must be known at this point
```

seanmonstar: It can be `let mut values = HashMap::new()`. It can infer key and value types.

src/app/config.rs
```diff
@@ -409,6 +412,9 @@ impl<'a> TryFrom<&'a Strings> for Config {
             tls_controller: controller_namespace?,
         };
 
+        let proxy_id_template = strings.get(ENV_PROXY_ID)?.unwrap_or(String::new());
```
olix0r: tioli... `.unwrap_or_default()`

src/app/config.rs
```diff
@@ -582,8 +590,10 @@ impl Strings for Env {
 
 impl TestEnv {
     pub fn new() -> Self {
+        let mut values = HashMap::new();
+        values.insert(ENV_PROXY_ID, "foo.deployment.default.linkerd-managed.linkerd.svc.cluster.local".into());
```
olix0r: i'd rather not hardcode this string structure into the proxy, even tests, if possible. The structure will change (soon) and we'll be left with vestigial names. Can we just use an obviously example name like "id.test.example.com" or something?   Or can we just omit it altogether?

tests/support/controller.rs
```diff
@@ -62,6 +62,7 @@ impl Controller {
         let dst = pb::GetDestination {
             scheme: "k8s".into(),
             path,
+            proxy_id: "foo.deployment.default.linkerd-managed.linkerd.svc.cluster.local".into(),
```
olix0r: likewise, prefer to use unrealistic ids for tests

tests/support/controller.rs
```diff
@@ -97,6 +98,7 @@ impl Controller {
...
```
olix0r: likewise, prefer to use unrealistic ids for tests

src/app/config.rs
```diff
@@ -409,6 +412,9 @@ impl<'a> TryFrom<&'a Strings> for Config {
             tls_controller: controller_namespace?,
         };
 
+        let proxy_id_template = strings.get(ENV_PROXY_ID)?.unwrap_or(String::new());
+        let proxy_id = proxy_id_template.replace(VAR_POD_NAMESPACE, &namespaces.pod);
```
olix0r: wait... what is this doing? why are we doing any special templating in the proxy?

olix0r: Specifically, all of this templating can be done via the kubernetes manifest:

```yaml
---
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  labels:
    app: env
  name: env
spec:
  selector:
    matchLabels:
      app: env
  template:
    metadata:
      labels:
        app: env
    spec:
      containers:
      - name: env
        command: ["sh", "-c"]
        args:
        - while true; do echo "id=${ID}" ; sleep 10 ; done
        image: debian:jessie
        env:
        - name: NS
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
        - name: NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
        - name: ID
          value: "$(NAME).$(NS).suffix"
```

olix0r:
Let's roll forward with this, since it's at least internally consistent, and replace this soon after.

PR #184, opened 2019-01-31
Add l5d-client-id on inbound requests if meshed TLS

seanmonstar:
If meshed TLS is used on an inbound connection, the TLS Identity of client is added as a header `l5d-client-id` to the request, so that the internal service can see it was encrypted.

- Client certificates were already automatically used if a proxy was configured with TLS.
- The server verifier was changed to *require* a client certificates.
- We assume the first DNS name in the certificate is the TLS identity assigned by the controller's CA.
  - In order to get the DNS names from the certificate, a fork of webpki is used, while waiting for the pull request to land.
- The `l5d-client-id` header is stripped on both inbound and outbound requests at the beginning, so *only* the inbound proxy will expose this header.

Closes https://github.com/linkerd/linkerd2/issues/2125
Closes https://github.com/linkerd/linkerd2/issues/2126


seanmonstar:
I wasn't sure about format or labels, so I just arbitrarily decided on this:

- In metrics, `tls="true"` is still present, and `client_id="..."` or `server_id="..."` is also included.
- In tap events, the `{"tls": "true"}` is still set, and `"client_id": "..."` or `"server_id": "..."` are also included.
- In both cases, if TLS is not enabled, the reason why is still included, and the `server_id`/`client_id` is left off, so to not state the reason twice.

olix0r:
@seanmonstar those sound like some good arbitrary decisions to me ;)  i'll review & test in the AM

src/app/inbound.rs
```diff
@@ -254,6 +254,137 @@ pub mod rewrite_loopback_addr {
...
+    }
+
+    impl<M, B> svc::Stack<Source> for Stack<M, B>
+    where
+        M: svc::Stack<Source>,
+    {
+        type Value = svc::Either<Service<M::Value, B>, M::Value>;
+        type Error = M::Error;
+
+        fn make(&self, source: &Source) -> Result<Self::Value, Self::Error> {
```
hawkw: The similarity between this and the `server_id` `Stack` kind of makes me wonder if it would be worth writing a generic `add_header` layer that takes a  header name, `fn(T) -> Option<HeaderValue>` (where `T` is the target type), and a request or response marker (like the strip-header layer). 

Not a blocker here, but might be handy if we end up writing a lot more similar code...

olix0r:
After some testing, it seems that the introduction of client authentication breaks our TLS detection logic:

```
linkerd-controller-856bc5d794-lmvjj linkerd-proxy TRCE proxy={server=in listen=0.0.0.0:4143} linkerd2_proxy::transport::tls::conditional_accept match_client_hello: failed to parse up to SNI
linkerd-controller-856bc5d794-lmvjj linkerd-proxy TRCE proxy={server=in listen=0.0.0.0:4143} linkerd2_proxy::transport::connection passing through accepted connection without TLS
```

the buffering/parsing logic wlil have to be made more robust.

olix0r:
This issue may not actually be caused by buffering. It appears that SNI is failing to match due to an injection bug.

olix0r:
The issue i saw is fixed by https://github.com/linkerd/linkerd2/commit/4798ad3f44f8e6386c7639b0ff7632586d8f04aa.  I'm going to run a few more tests off of this branch (after merging master) and prepare it to merge.

src/proxy/server.rs
```diff
@@ -70,7 +70,7 @@ pub struct Source {
     pub remote: SocketAddr,
     pub local: SocketAddr,
     pub orig_dst: Option<SocketAddr>,
-    pub tls_status: tls::Status,
+    pub tls_peer: tls::ConditionalIdentity,
```
olix0r: If I understand correctly, this means that we can't describe the situation where TLS is established and a client ID is not provided. This case is important, especially for the service that has to bootstrap certificates -- proxies will have to be able to connect to this service before they have an identity.

I think we'll need to make the ConditionalIdentity type that is more like:

```rust
type ClientIdentity = Conditional<Identity, ReasonForNoClientIdentity>;
type ServerStatus = Conditional<ClientIdentity, ReasonForNoTls>;
```

olix0r: After taking a stab at this, it's a little complex; and, since much of the internals of this are going to have to change in an upcoming release, it's probably best to merge this as-is and address these concerns later.

seanmonstar: I may be missing some context, but as I was reading through the TLS code, it appeared to me that TLS couldn't be established without an identity... 🤷‍♂️

PR #186, opened 2019-02-02
Introduce TCP keepalive configuration

olix0r:
In some network environments, peers may silently drop connections such that the proxy cannot detect that the peer's socket has been closed.

The [TCP keepalive socket options][tcp-keepalive] configures the kernel to actively probe connections to ensure connectivity and prevent idle timeouts.

This change adds stack modules that attempt to configure accept and connect sockets' TCP keepalive socket options. There are four new environment configurations the proxy supports:

    - `LINKERD2_PROXY_INBOUND_ACCEPT_KEEPALIVE`
    - `LINKERD2_PROXY_OUTBOUND_ACCEPT_KEEPALIVE`
    - `LINKERD2_PROXY_INBOUND_CONNECT_KEEPALIVE`
    - `LINKERD2_PROXY_OUTBOUND_CONNECT_KEEPALIVE`

When an environment variable is unset, no keepalive is set on the corresponding sockets. Otherwise, its value is parsed as a duration. OSes may or may not understand subsecond values.

It is recommended to only set the inbound-accept and outbound-connect keepalive values, as keepalives shouldn'tbe necessary on localhost.

Addresses linkerd/linkerd2#1949 linkerd/linkerd2#2182

[tcp-keepalive]: http://www.tldp.org/HOWTO/TCP-Keepalive-HOWTO/overview.html

---

Furthermore, adds trace logging to the TCP forwarder and fixes docker compilation by updating the stable rust version in Dockerfile.

### DIFF
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -683,10 +683,10 @@ fn serve<A, C, R, B, G>(
 where
     A: svc::Stack<proxy::server::Source, Error = Never> + Send + Clone + 'static,
     A::Value: proxy::Accept<Connection>,
-    <A::Value as proxy::Accept<Connection>>::Io: Send + transport::Peek + 'static,
+    <A::Value as proxy::Accept<Connection>>::Io: fmt::Debug + Send + transport::Peek + 'static,
     C: svc::Stack<connect::Target, Error = Never> + Send + Clone + 'static,
     C::Value: connect::Connect + Send,
-    <C::Value as connect::Connect>::Connected: Send + 'static,
+    <C::Value as connect::Connect>::Connected: fmt::Debug + Send + 'static,
     <C::Value as connect::Connect>::Future: Send + 'static,
     <C::Value as connect::Connect>::Error: fmt::Debug + 'static,
     R: svc::Stack<proxy::server::Source, Error = Never> + Send + Clone + 'static,
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -34,7 +34,7 @@ use svc::{
 use tap;
 use task;
 use telemetry;
-use transport::{self, connect, tls, BoundPort, Connection, GetOriginalDst};
+use transport::{self, connect, keepalive, tls, BoundPort, Connection, GetOriginalDst};
 use {Addr, Conditional};
 
 use super::config::Config;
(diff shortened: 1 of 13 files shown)


olix0r:
This can be tested via https://github.com/linkerd/linkerd2/pull/2193:

I've published images with both the controller/cli and proxy changes to `git-0783c26c`, but you need to make sure to use an updated CLI:

```
:; cd ~/b/linkerd2
:; git fetch origin ver/proxy-keepalive-default
:; git co ver/proxy-keepalive-default
:; rm -rf target/cli
:; bin/linkerd install -v git-0783c26c | kubectl apply -f -
...
:; bin/linkerd inject -v git-0783c26c ... | kubectl apply -f -
...
```

src/app/main.rs
```diff
@@ -363,7 +368,9 @@ where
                 //    is retryable.
                 let dst_route_layer = phantom_data::layer()
                     .push(insert_target::layer())
-                    .push(metrics::layer::<_, classify::Response>(retry_http_metrics.clone()))
+                    .push(metrics::layer::<_, classify::Response>(
+                        retry_http_metrics.clone(),
```
hawkw: looks like there were a lot of unrelated formatting changes here...not super important, though.

src/transport/keepalive.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    impl<T, M> svc::Stack<T> for Stack<M>
+    where
+        M: svc::Stack<T>,
+    {
+        type Value = Accept<M::Value>;
+        type Error = M::Error;
+
+        fn make(&self, target: &T) -> Result<Self::Value, Self::Error> {
```
hawkw: TIOLI: it seems like we could probably reduce some duplicate code (the `Layer` and `Stack` implementations) if this was implemented similarly to the `strip_header` layer, using a marker to indicate whether we're building an `Accept` or a `Connect:
https://github.com/linkerd/linkerd2-proxy/blob/d3e7f8bb69c0f2ed58bcbedc8acecf3079344e9b/src/proxy/http/strip_header.rs#L33-L83

src/app/main.rs
olix0r: this file is now in the hands of rustfmt

src/transport/keepalive.rs
olix0r: I find that approach a bit harder to work with, personally.  The overhead of having two similarly-shaped types is preferable, in my opinion, to having to reason about how a single type changes behavior in various situations. As it is, it's pretty clear and unambiguous, if slightly more verbose.

PR #188, opened 2019-02-08
Minor cleanup in the config tests

adleong:


tests/support/controller.rs
```diff
@@ -62,7 +62,7 @@ impl Controller {
         let dst = pb::GetDestination {
             scheme: "k8s".into(),
             path,
-            proxy_id: "foo.deployment.default.linkerd-managed.linkerd.svc.cluster.local".into(),
+            proxy_id: String::new(),
```
hawkw: TIOLI:
```suggestion
            ..Default::default()
```
to use the default initializer for all other fields. This means the code will still compile if the proto adds fields later.

tests/support/controller.rs
```diff
@@ -98,7 +98,7 @@ impl Controller {
...
```
hawkw: Same as above,
```suggestion
            ..Default::default()
```

tests/support/controller.rs
adleong: I'd be inclined to leave it.  I'm not yet familiar with rust style or idioms but `String::new()` feels more explicit to me than `Default::default()`.  And I feel like compilation failing when we add proto fields is a good thing as it forces us to decide what we want to do about those fields.

hawkw: That's fair. `String::new` is definitely more idiomatic in general, I've just been bitten in the past by proto changes breaking a bunch of tests when those fields aren't actually _required_ by the test...

olix0r: i think the `default` approach is fine for protobuf structs, especially in tests. but i don't feel strongly.

PR #189, opened 2019-02-09
Log canonicalization warnings on only the first error

olix0r:
When a canonicalization task fails to resolve a name, our logging is not
particularly clear about the current state of the stack. Specifically,
it's difficult to know whether the stack has resolved the name
successfully before.

With this change, canonicalization failures are logged (at warning, not
error) only when the task has not previously resolved a name.
Subsequent errors are now logged at the debug level (instead of
warning).

### DIFF
--- a/src/proxy/canonicalize.rs
+++ b/src/proxy/canonicalize.rs
@@ -44,11 +44,25 @@ pub struct Service<M: svc::Stack<Addr>> {
 
 struct Task {
     original: NameAddr,
-    resolved: Option<NameAddr>,
+    resolved: Cache,
     resolver: dns::Resolver,
     state: State,
     timeout: Duration,
     tx: mpsc::Sender<NameAddr>,
+}
+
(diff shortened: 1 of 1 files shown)


src/proxy/canonicalize.rs
```diff
@@ -167,30 +178,41 @@ impl Future for Task {
...
-                                // Pretend the original name was resolved so
-                                // that we don't re-publish on subsequent errors.
-                                self.resolved = Some(self.original.clone());
+                                // There's now no need to re-publish the
+                                // original name on subsequent failures.
+                                self.resolved = Cache::Unresolved;
                             } else {
-                                warn!("failed to refine {}: {}, falling back", self.original.name(), e);
+                                debug!(
+                                    "failed to refresh {}: {}; cache={:?}",
```
hawkw: :+1: for "failed to refresh", that definitely makes it clearer that we were just trying to update a previously resolved name.

PR #190, opened 2019-02-11
Update metrics quickcheck to 0.8, and hyper to 0.12.24

seanmonstar:


hawkw:
While we're bumping dependencies, we should be able to pull `inotify 0.7` from crates.io (instead of patching it to my fork), and `prost` from origin rather than my fork, since my changes have merged upstream.

PR #192, opened 2019-02-14
Fix compilation warnings in fs-watch

olix0r:
fs-watch contains compilation warnings due to an unused macro and a
public enum that is not constructed within the crate. This resolves
these warnings.

### DIFF
--- a/lib/fs-watch/src/either_stream.rs
+++ b/lib/fs-watch/src/either_stream.rs
@@ -5,6 +5,7 @@ use futures::{Poll, Stream};
 /// Combines two different `Stream`s yielding the same item and error
 /// types into a single type.
 #[derive(Clone, Debug)]
+#[allow(dead_code)]
 pub enum EitherStream<A, B> {
     A(A),
     B(B),
--- a/lib/fs-watch/src/either_stream.rs
+++ b/lib/fs-watch/src/either_stream.rs
@@ -5,7 +5,6 @@ use futures::{Poll, Stream};
 /// Combines two different `Stream`s yielding the same item and error
 /// types into a single type.
 #[derive(Clone, Debug)]
-#[allow(dead_code)]
 pub enum EitherStream<A, B> {
     A(A),
     B(B),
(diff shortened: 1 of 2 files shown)


lib/fs-watch/src/either_stream.rs
```diff
@@ -5,6 +5,7 @@ use futures::{Poll, Stream};
 /// Combines two different `Stream`s yielding the same item and error
 /// types into a single type.
 #[derive(Clone, Debug)]
+#[allow(dead_code)]
```
seanmonstar: It's not dead code on Linux. Perhaps better would be to put the same cfg on the `mod either_stream` in `lib.rs`.

lib/fs-watch/src/lib.rs
```diff
@@ -1,6 +1,5 @@
 #[macro_use]
 extern crate log;
-#[macro_use]
```
seanmonstar: I suspect this macro isn't used on Mac, but it is on Linux.

lib/fs-watch/src/either_stream.rs
hawkw: Alternatively, we could add `#[cfg(target_os="linux")]` to the module definition.

lib/fs-watch/src/lib.rs
hawkw: Yeah, that's correct.

lib/fs-watch/src/either_stream.rs
seanmonstar: Sorry, I meant `#[cfg(target_os="linux")] mod either_stream`, just was on mobile and didn't want to type the symbols.

lib/fs-watch/src/lib.rs
```diff
@@ -1,7 +1,10 @@
 #[macro_use]
 extern crate log;
+#[cfg(target_os = "linux")]
 #[macro_use]
 extern crate futures;
+#[cfg(not(target_os = "linux"))]
```
seanmonstar: I believe this can use `#[cfg_attr(target_os = "linux", macro_use)]`.

Alternatively, with newer Rust version, we can just use `futures::try_ready!` at the invocation point, I believe.

PR #193, opened 2019-02-14
 make: Optionally include debug symbols with builds

olix0r:
Per linkerd/linkerd2#2199, we want to be able to include the proxy's debug symbols in auxiliary diagnostic containers. In order to do this, the process that produces the proxy's binaries also should be responsible for producing its debug symbols.

This change adds a new environment variable, `CARGO_DEBUG`, which the Makefile uses to instrument the build to produce debug symbols. If `objcopy` is available on the local system (i.e. on Linux), then the symbols are copied from the binary into a .obj file, and the executable is stripped from 3.3G down to 9.9M before packaging.
    
This increases the size of the compressed package artifact by 76% (from 5.0M to 8.8M). This may have similar impacts on build time. This may have similar impacts on build time. To that end, CARGO_DEBUG is not enabled in CI yet.
    
CI has been changed, however, to use build stages. Now, master will undergo the same test stage as PRs, though packages are only built & published for master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ sudo: false
 
 language: rust
 cache: cargo
-env:
-  - RUSTFLAGS="-C debuginfo=0" RUST_TEST_THREADS=1 RUST_TEST_PATIENCE_MS=200
 
 branches:
   only:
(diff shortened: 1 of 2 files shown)


olix0r:
Note that `CARGO_VERBOSE` has been replaced with `travis_wait` (which is somewhat more reliable for the purpose of not timing out builds)

PR #194, opened 2019-02-21
Update Router to wait for inner poll_ready before calling inner call

seanmonstar:
The contract to only call `Service::call` after seeing ready from `poll_ready` was just [made strong](https://github.com/tower-rs/tower/pull/161). So, this one step towards making sure we do just that.

Since a router can't call `poll_ready` on all of its inner routes, instead it requires each service be `Clone`, and calls `poll_ready` in its `ResponseFuture`.

### DIFF
--- a/lib/router/src/lib.rs
+++ b/lib/router/src/lib.rs
@@ -3,7 +3,7 @@ extern crate indexmap;
 extern crate linkerd2_stack as stack;
 extern crate tower_service as svc;
 
-use futures::{Future, Poll};
+use futures::{Async, Future, Poll};
 
 use std::hash::Hash;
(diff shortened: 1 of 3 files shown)


lib/router/src/lib.rs
```diff
@@ -64,22 +64,21 @@ where
...
 where
-    F: Future,
+    Svc: svc::Service<Req>,
 {
-    Inner(F),
-    RouteError(E),
-    NoCapacity(usize),
-    NotRecognized,
-    Invalid,
+    NotReady(Req, Svc),
```
carllerche: These state transitions can be extracted to a shared `Oneshot` utility in `tower-util`. (not saying this PR should do it).

PR #195, opened 2019-02-22
fs-watch: Use a properly sized buffer for inotify events

hawkw:
The buffer passed to `read(2)` to read inotify events from the inotify
fd must be long enough to fit events with filenames of the maximum
allowable length. The `inotify-rs` library previously always constructed
such a buffer, but as of 1fe4e3885039b9e66f9d93bf0cec11748253c39e, we
are required to provide our own buffer. The buffer we were providing was
insufficiently large for some filenames.

See also the `man` page for `inotify`:
> The behavior when the buffer given to read(2) is too small to return
> information about the next event depends on the kernel version: in
> kernels before 2.6.21, read(2) returns 0; since kernel 2.6.21,
> read(2) fails with the error EINVAL.  Specifying a buffer of size
>
>         sizeof(struct inotify_event) + NAME_MAX + 1
>
> will be sufficient to read at least one event.

(http://man7.org/linux/man-pages/man7/inotify.7.html)

This branch updates the `fs-watch` lib to calculate the necessary buffer
length correctly, and use a properly-sized buffer. I've also added tests
that fail when the buffer is too short.

Fixes linkerd/linkerd2#2331

### DIFF
--- a/lib/fs-watch/src/lib.rs
+++ b/lib/fs-watch/src/lib.rs
@@ -177,14 +177,21 @@ impl PathAndHash {
 #[cfg(target_os = "linux")]
 pub mod inotify {
     extern crate inotify;
+    extern crate inotify_sys;
+    extern crate libc;
 
     use self::inotify::{Event, EventMask, EventStream, Inotify, WatchMask};
+    use self::inotify_sys as ffi;
+
     use futures::{Async, Poll, Stream};
-    use std::{io, path::PathBuf};
+    use std::{io, mem, path::PathBuf};
+
+    const EVENT_BUF_SZ: usize =
+        mem::size_of::<ffi::inotify_event>() + (libc::FILENAME_MAX as usize) + 1;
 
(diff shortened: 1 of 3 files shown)


lib/fs-watch/src/lib.rs
```diff
@@ -812,4 +819,26 @@ mod tests {
     fn inotify_nonexistent_files_dont_file_delete_events() {
         Fixture::new().test_inotify(test_detects_delete_and_recreate)
     }
+
+    const LONG: &str = "loooooooooooooooooooooooooooooooooooooooooooooooongboi";
```
olix0r: this might be a good job for quickcheck

hawkw: Yeah, that's a good point. I'm going to go ahead and merge the bugfix first, though, so it's not blocked on writing a QuickCheck generator for valid filenames?

PR #197, opened 2019-02-23
Update *ring* to 0.14.6

hawkw:


hawkw:
AFAICT the `winapi` and `spin` version bumps are required by the latest `*ring*`.

PR #198, opened 2019-02-24
Remove the CONTROLLER_NAMESPACE configuration

olix0r:
linkerd/linkerd2#2371 removes the need for the proxy to be aware of the
controller namespace by moving the TLS-identity checking logic into the
destination service.

With this change, when the destination service returns an identity, the
proxy will try to establish TLS'd communication to the endpoint if the
proxy has trust anchors.


olix0r:
fixed in another branch

PR #199, opened 2019-02-24
Accept identity configuration without templating

olix0r:
Previously, the proxy accepted an identity template and namespace and
merged them during configuration. We can rely on external configuration
systems to do this templating, and so we simplify the proxy by no longer
taking a pod namespace environment and instead taking a
`LINKERD2_PROXY_LOCAL_IDENTITY`.

### DIFF
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -118,8 +118,6 @@ pub struct Config {
 
 #[derive(Clone, Debug)]
 pub struct Namespaces {
-    pub pod: String,
-
     /// `None` if TLS is disabled; otherwise must be `Some`.
     pub tls_controller: Option<String>,
 }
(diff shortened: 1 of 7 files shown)


src/app/config.rs
```diff
@@ -430,14 +426,7 @@ impl<'a> TryFrom<&'a Strings> for Config {
...
                 .unwrap_or(DEFAULT_DNS_CANONICALIZE_TIMEOUT);
-
-        let pod_namespace = strings.get(ENV_POD_NAMESPACE).and_then(|maybe_value| {
-            // There cannot be a default pod namespace, and the pod namespace is required.
-            maybe_value.ok_or_else(|| {
-                error!("{} is not set", ENV_POD_NAMESPACE);
-                Error::InvalidEnvVar
-            })
-        });
+;
```
seanmonstar: This little guy looks lost.

tests/discovery.rs
```diff
@@ -813,7 +813,7 @@ mod proxy_to_proxy {
         env.put(app::config::ENV_TLS_PRIVATE_KEY, key);
         env.put(app::config::ENV_TLS_TRUST_ANCHORS, trust_anchors);
         env.put(
-            app::config::ENV_TLS_POD_IDENTITY,
+            app::config::ENV_TLS_local_identITY,
```
adleong: I think the casing is messed up here

tests/transparency.rs
```diff
@@ -151,7 +151,7 @@ fn tcp_server_first_tls() {
     env.put(app::config::ENV_TLS_PRIVATE_KEY, key);
     env.put(app::config::ENV_TLS_TRUST_ANCHORS, trust_anchors);
     env.put(
-        app::config::ENV_TLS_POD_IDENTITY,
+        app::config::ENV_TLS_local_identITY,
```
adleong: and here

tests/discovery.rs
hawkw: Yeah, this looks like a case-insensitive find-and-replace accident or something.

olix0r:
fixed in  another PR

PR #200, opened 2019-02-25
Set `l5d-remote-ip` on inbound requests and outbound responses

kleimkuhler:
## Problem

[Context](https://github.com/linkerd/linkerd2/issues/2310)

## Solution

We now add the `l5d-remote-ip` header to all inbound requests and outbound responses. We also make sure to strip the header before setting so that clients may not set this header.

We were already adding a header in similar style, so in order to deduplicate code, `add_header.rs` holds the generic logic for adding a header to a `http::Request` or `http::Response`. `l5d-client-id` and `l5d-server-id` now also use the same logic in order to add their respective headers.

## Validation

An additional test module was added to `tests/discovery.rs` that tests the following:
- outbound responses strip `l5d-remote-ip`
- inbound requests strip `l5d-remote-ip`
- outbound responses set `l5d-remote-ip` to the endpoint IP
- inbound requests set `l5d-remote-ip` to the source IP

Closes linkerd/linkerd2#2310

### DIFF
--- a/src/app/inbound.rs
+++ b/src/app/inbound.rs
@@ -388,6 +388,122 @@ pub mod client_id {
     }
 }
 
+/// Adds `l5d-remote-ip` headers to http::Requests derived from the
+/// `remote` of a `Source`.
+pub mod remote_ip {
+    use std::marker::PhantomData;
+
+    use futures::Poll;
+    use http::{self, header::HeaderValue};
+
+    use proxy::server::Source;
+    use svc;
+
+    #[derive(Debug)]
+    pub struct Layer<B>(PhantomData<fn() -> B>);
+
+    #[derive(Debug)]
+    pub struct Stack<M, B> {
+        inner: M,
+        _marker: PhantomData<fn() -> B>,
+    }
+
+    #[derive(Debug)]
+    pub struct Service<S, B> {
+        inner: S,
+        value: HeaderValue,
+        _marker: PhantomData<fn() -> B>,
+    }
+
+    pub fn layer<B>() -> Layer<B> {
+        Layer(PhantomData)
+    }
+
+    impl<B> Clone for Layer<B> {
+        fn clone(&self) -> Self {
+            Layer(PhantomData)
+        }
+    }
+
+    impl<M, B> svc::Layer<Source, Source, M> for Layer<B>
+    where
+        M: svc::Stack<Source>,
+    {
+        type Value = <Stack<M, B> as svc::Stack<Source>>::Value;
+        type Error = <Stack<M, B> as svc::Stack<Source>>::Error;
+        type Stack = Stack<M, B>;
+
+        fn bind(&self, inner: M) -> Self::Stack {
+            Stack {
+                inner,
+                _marker: PhantomData,
+            }
+        }
+    }
+
+    // === impl Stack ===
+
+    impl<M: Clone, B> Clone for Stack<M, B> {
+        fn clone(&self) -> Self {
+            Stack {
+                inner: self.inner.clone(),
+                _marker: PhantomData,
+            }
+        }
+    }
+
+    impl<M, B> svc::Stack<Source> for Stack<M, B>
+    where
+        M: svc::Stack<Source>,
+    {
+        type Value = Service<M::Value, B>;
+        type Error = M::Error;
+
+        fn make(&self, source: &Source) -> Result<Self::Value, Self::Error> {
+            let svc = self.inner.make(source)?;
+            let value = HeaderValue::from_str(&source.remote.ip().to_string()).unwrap();
+
+            return Ok(Service {
+                inner: svc,
+                value,
+                _marker: PhantomData,
+            });
+        }
+    }
+
+    // === impl Service ===
+
+    impl<S: Clone, B> Clone for Service<S, B> {
+        fn clone(&self) -> Self {
+            Service {
+                inner: self.inner.clone(),
+                value: self.value.clone(),
+                _marker: PhantomData,
+            }
+        }
+    }
+
+    impl<S, B> svc::Service<http::Request<B>> for Service<S, B>
+    where
+        S: svc::Service<http::Request<B>>,
+    {
+        type Response = S::Response;
+        type Error = S::Error;
+        type Future = S::Future;
+
+        fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+            self.inner.poll_ready()
+        }
+
+        fn call(&mut self, mut req: http::Request<B>) -> Self::Future {
+            req.headers_mut()
+                .insert(super::super::L5D_REMOTE_IP, self.value.clone());
+
+            self.inner.call(req)
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use http;
--- a/src/app/inbound.rs
+++ b/src/app/inbound.rs
@@ -261,7 +261,7 @@ pub mod rewrite_loopback_addr {
 
 /// Adds `l5d-client-id` headers to http::Requests derived from the
 /// TlsIdentity of a `Source`.
-pub mod client_id {
+pub mod set_client_id_on_req {
     use super::super::L5D_CLIENT_ID;
     use http::header::HeaderValue;
 
@@ -292,7 +292,7 @@ pub mod client_id {
 
 /// Adds `l5d-remote-ip` headers to http::Requests derived from the
 /// `remote` of a `Source`.
-pub mod remote_ip {
+pub mod set_remote_ip_on_req {
     use super::super::L5D_REMOTE_IP;
     use bytes::Bytes;
     use http::header::HeaderValue;
--- a/src/app/inbound.rs
+++ b/src/app/inbound.rs
@@ -262,245 +262,47 @@ pub mod rewrite_loopback_addr {
 /// Adds `l5d-client-id` headers to http::Requests derived from the
 /// TlsIdentity of a `Source`.
 pub mod client_id {
-    use std::marker::PhantomData;
+    use http::header::HeaderValue;
 
-    use futures::Poll;
-    use http::{self, header::HeaderValue};
-
-    use proxy::server::Source;
-    use svc;
+    use proxy::{
+        http::add_header::{self, request::ReqHeader, Layer},
+        server::Source,
+    };
     use Conditional;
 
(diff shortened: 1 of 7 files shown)


src/proxy/http/add_header.rs
```diff
@@ -0,0 +1,229 @@
...
+        Stack {
+            header: self.header.clone(),
+            retrieve: self.retrieve,
+            inner,
+            _req_or_res: PhantomData,
+        }
+    }
+}
+
+impl<H, T, R> fmt::Debug for Layer<H, T, R>
```
hawkw: Can we get this to display whether the layer is a request layer or a response layer?

I *think* you can do something like this:

```rust
impl<H, T> fmt::Debug for Layer<H, T, ReqHeader>
where
    H: fmt::Debug,
    T: fmt::Debug,
{
        f.debug_struct("Layer")
            .field("header", &self.header)
            .field("retrieve", &format_args!("{}", "..."))
            .field("req_or_res", &format_args!("{}", "Request"))
            .finish()
    }
}


impl<H, T> fmt::Debug for Layer<H, T, ResHeader> 
where
    ...
```

tests/discovery.rs
```diff
@@ -328,6 +328,83 @@ macro_rules! generate_tests {
             assert_eq!(client.get("/"), "hello");
         }
 
+        mod remote_header {
+            use http::header::HeaderValue;
+            use super::super::*;
+
+            const REMOTE_IP_HEADER: &'static str = "l5d-remote-ip";
+            const FOO: &'static str = "0.0.0.0";
```
hawkw: Can we call these, like, `IP_1` and `IP_2` or something?

tests/discovery.rs
```diff
@@ -328,6 +328,83 @@ macro_rules! generate_tests {
...
+                let header = HeaderValue::from_static(FOO);
+
+                let srv = $make_server().route_fn("/strip", |_req| {
+                    Response::builder().header(REMOTE_IP_HEADER, FOO).body(Default::default()).unwrap()
+                }).run();
+
+                let ctrl = controller::new().destination_and_close("disco.test.svc.cluster.local", srv.addr);
+                let proxy = proxy::new().controller(ctrl.run()).outbound(srv).run();
+                let client = $make_client(proxy.outbound, "disco.test.svc.cluster.local");
+                let rsp = client.request(&mut client.request_builder("/strip"));
```
hawkw: Nit/TIOLI: since you're not doing anything special with the request, I think this can just be
```suggestion
                let rsp = client.get("/strip");
```

src/proxy/http/add_header.rs
olix0r: When do we expect these layers to be logged?  they really only exist in main.rs ephemerally...

i'd rather keep our boilerplate code to a minimum if there's no upsid

hawkw: @olix0r that's fair, I suppose it's really the _service_ that ought to be able to indicate this. Maybe consider replacing its derived `Debug` impl, and remove this one entirely since we're not currently logging layers?

tests/discovery.rs
kleimkuhler: Since `.get` returns a `String`, it was either this, or change `Response::default()` to `builder()` and setting a body (in not wanting to just assert equality on `""`). I'll keep it how it is since changing it just introduces a small nit somewhere else.

src/proxy/http/add_header.rs
kleimkuhler: @hawkw Are you suggesting `ReqHeader` and `ResHeader` `Debug` impls, but for `Service` instead?

hawkw: @kleimkuhler yeah, that's correct --- I think the `Service` types _may_ on occasion show up in the logs though I'm not sure if that's the case or not.

tests/discovery.rs
hawkw: 👍

src/app/inbound.rs
```diff
@@ -262,129 +262,49 @@ pub mod rewrite_loopback_addr {
...
-
-        fn call(&mut self, mut req: http::Request<B>) -> Self::Future {
-            req.headers_mut()
-                .insert(super::super::L5D_CLIENT_ID, self.value.clone());
-
-            self.inner.call(req)
-        }
+/// Adds `l5d-remote-ip` headers to http::Requests derived from the
+/// `remote` of a `Source`.
+pub mod remote_ip {
```
olix0r: It's pretty confusing that we have inbound::remote_ip and outbound::remote_ip and that they do subtle different things. Can we choose clearer names that differentiate the behavior?

`set_remote_ip_on_{req,rsp}`?

src/proxy/http/add_header.rs
```diff
@@ -0,0 +1,229 @@
+use std::{fmt, marker::PhantomData};
+
+use http::header::{AsHeaderName, HeaderValue};
+
+use svc;
+
+type Ret<T> = fn(&T) -> Option<HeaderValue>;
```
olix0r: This probably warrants a description.  Also, I have no idea what `Ret` is supposed to be short for. Can this be an actual name?

src/proxy/http/add_header.rs
```diff
@@ -0,0 +1,229 @@
...
+
+/// Wraps HTTP `Service` `Stack<T>`s so that a given header is removed from a
+/// request or response.
+#[derive(Clone)]
+pub struct Layer<H, T, R> {
+    header: H,
+    retrieve: Ret<T>,
+    _req_or_res: PhantomData<fn(R)>,
```
olix0r: Hm??? What does this mean?

This definitely falls within the realm of "surprisingly clever", in my opinion, and so it warrants an explanation.

src/proxy/http/add_header.rs
```diff
@@ -0,0 +1,229 @@
...
+
+// === impl Stack ===
+
+impl<H, T, M, R> svc::Stack<T> for Stack<H, T, M, R>
+where
+    H: AsHeaderName + Clone + fmt::Debug,
+    T: fmt::Debug,
+    M: svc::Stack<T>,
+{
+    type Value = svc::Either<Service<H, M::Value, R>, M::Value>;
```
olix0r: Why is this necessary? Can't we just always return a wrapped service?

src/proxy/http/add_header.rs
hawkw: Note that this is based on code that already exists in the proxy (in the analogous `strip_header` layer): https://github.com/linkerd/linkerd2-proxy/blob/6cc755881994cba967619a0999dd7b2339c986e2/src/proxy/http/strip_header.rs#L12

I suggested that @kleimkuhler base his implementation off of that layer when we were discussing how to make this change...

src/proxy/http/add_header.rs
hawkw: I suppose the alternative approach would just be to wrap with a service that has an `Option<HeaderValue>` and just not insert if it's `None`. Might be a little simpler...

-

src/app/inbound.rs
kleimkuhler: The naming we choose is reflected here in [main.rs](https://github.com/linkerd/linkerd2-proxy/blob/0bd13492d55dbce44b44e64d53351c539c54c3eb/src/app/main.rs#L493)

Both `client_id` and `remote_ip` serve the same purpose for their respective header. If I change the naming on this, I'll change the naming on both.

I like `set_{header}_on_{req, rsp}`, so it should help clarify what exactly the line `.push(client_id::layer()` is doing when building the stack. It also just introduces another naming convention for these modules as headers are added (which may not be often!).

src/proxy/http/add_header.rs
kleimkuhler: The `Either` style was adapted from the current way we add `client_id` [right now](https://github.com/linkerd/linkerd2-proxy/blob/dc00685639c50c5cf608bb9c0d587b87b9a83282/src/app/inbound.rs#L331). I think changing to an `Option<HeaderValue>` could simplify it, but we are pretty explicit right now when it fails.

olix0r: It's probably fine as it is, I'm just looking for opportunities to simplify.

src/app/inbound.rs
olix0r: I think it makes sense to clarify the naming in both cases.

kleimkuhler:
Okay I change the naming of our modules in `inbound.rs` and `outbound.rs`. `Ret` is now `RetrieveHeader` so that should clarify to the reader it is a function that retrieves a header value for a given Stack target.

src/proxy/http/add_header.rs
```diff
@@ -77,7 +77,7 @@ where
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.debug_struct("Layer")
             .field("header", &self.header)
-            .field("retrieve", &format_args!("{}", "..."))
+            .field("get_header", &format_args!("{}", "..."))
```
olix0r: i'll just note that this my complaint with adding debug impls to types we never try to format as debug... it's just more code that has to be maintained when we make small changes.

While we're here.. what's the deal with the format_args macro? We can't use "..." as a literal here?

kleimkuhler: It's because `get_header` is not going to format to something helpful since it is a function pointer. It's comparable to `fmt::Debug` for [`Resolver`](https://github.com/linkerd/linkerd2-proxy/blob/6cc755881994cba967619a0999dd7b2339c986e2/src/dns.rs#L195). I'm not pointing this out as a reason to keep it; I think it proves your point more.

I'm okay removing the debug impl since it already does not serve a purpose in the code.

hawkw: @olix0r the `format_args!` was my suggestion --- we're not using a literal because it will be formatted using its `Debug` implementation, and the `Debug` implementation for strings includes the quotation marks, so it'll be formatted like this: `"..."`. However, formatting a string with `Display` omits the quotes. 

I thought we should avoid having the quotes so that the formatted output doesn't imply that the `get_header` field's value is a string.

With that said, I'm happy to agree that types which will not be formatted don't need `Debug` impls --- in the past we've generally tried to adhere to a practice of providing `Debug` impls for all types?

olix0r: @hawkw thanks, makes sense.

PR #201, opened 2019-02-26
replace Error::cause with Error::source (#2370)

DebugSteven:
proxy: replace `Error::cause` with `Error::source`

Currently, linkerd2-proxy uses and implements `Error::cause`
for error handling.

This branch removes the use of the deprecated
`Error::cause` method and implements and uses the
`Error::source` method, as [recommended by the
documentation](https://doc.rust-lang.org/std/error/trait.Error.html#method.cause). To implement `Error::source`, there are
static lifetimes added to the `Error` traits. However, this
doesn't remove the [use of `cause2`](https://github.com/DebugSteven/linkerd2-proxy/blob/231cb67b1a2178627bbb304055d408cee1007fd9/src/proxy/http/glue.rs#L226) because
`hyper::Error` doesn't provide a `source` yet:
https://docs.rs/hyper/0.12.24/src/hyper/error.rs.html#289-333.
A separate issue could be opened to use `source` for
`hyper::Error` when it becomes available.

Fixes: https://github.com/linkerd/linkerd2/issues/2370

### DIFF
--- a/lib/metrics/src/serve.rs
+++ b/lib/metrics/src/serve.rs
@@ -113,7 +113,7 @@ impl fmt::Display for ServeError {
             f,
             "{}: {}",
             self.description(),
-            self.cause().expect("ServeError must have cause")
+            self.source().expect("ServeError must have cause")
         )
     }
 }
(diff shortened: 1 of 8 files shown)


lib/metrics/src/serve.rs
```diff
@@ -113,7 +113,7 @@ impl fmt::Display for ServeError {
             f,
             "{}: {}",
             self.description(),
-            self.cause().expect("ServeError must have cause")
+            self.source().expect("ServeError must have cause")
```
hawkw: Take it or leave it: maybe it's worth changing the `expect` message to
```suggestion
            self.source().expect("ServeError must have source")
```
but it's not super important.

DebugSteven:
Great! Thank you for the review!

PR #203, opened 2019-03-01
perf(router): Increase inbound/outbound router cap.

kleimkuhler:
In both server and client case, the default value of
MAX_CONCURRENT_STREAMS is `u32::MAX`. However, we currently cap the
inbound and outbound router capacity limits to 100.

This change increases the capacity of both to 10_000.

This will close linkerd/linkerd2#2395 and could possibly address the
issue in linkerd/linkerd2#2363.

### DIFF
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -299,8 +299,8 @@ const DEFAULT_RESOLV_CONF: &str = "/etc/resolv.conf";
 
 /// It's assumed that a typical proxy can serve inbound traffic for up to 100 pod-local
 /// HTTP services and may communicate with up to 10K external HTTP domains.
-const DEFAULT_INBOUND_ROUTER_CAPACITY: usize = 100;
-const DEFAULT_OUTBOUND_ROUTER_CAPACITY: usize = 100;
+const DEFAULT_INBOUND_ROUTER_CAPACITY: usize = 10000;
+const DEFAULT_OUTBOUND_ROUTER_CAPACITY: usize = 10000;
 
 const DEFAULT_INBOUND_ROUTER_MAX_IDLE_AGE: Duration = Duration::from_secs(60);
 const DEFAULT_OUTBOUND_ROUTER_MAX_IDLE_AGE: Duration = Duration::from_secs(60);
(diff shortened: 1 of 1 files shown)


src/app/config.rs
```diff
@@ -299,8 +299,8 @@ const DEFAULT_RESOLV_CONF: &str = "/etc/resolv.conf";
 
 /// It's assumed that a typical proxy can serve inbound traffic for up to 100 pod-local
```
hawkw: This comment is definitely out of date now and should be updated (or just removed).

hawkw:
Er, sorry, I ment to hit the "comment" button rather than the "approve" button --- I think this is correct, but I'll have to think about it for a sec.

src/app/config.rs
```diff
@@ -299,8 +299,8 @@ const DEFAULT_RESOLV_CONF: &str = "/etc/resolv.conf";
...
 /// HTTP services and may communicate with up to 10K external HTTP domains.
-const DEFAULT_INBOUND_ROUTER_CAPACITY: usize = 100;
-const DEFAULT_OUTBOUND_ROUTER_CAPACITY: usize = 100;
+const DEFAULT_INBOUND_ROUTER_CAPACITY: usize = 10000;
```
dadjeibaah: Just out of curiosity, how did we pick this number?

src/app/config.rs
seanmonstar: Does it make sense that inbound would really need 10,000 different authorities?

src/app/config.rs
kleimkuhler: After talking with @olix0r about it, it is an arbitrary number that seemed more sensible than 100.

src/app/config.rs
kleimkuhler: I think it would be okay if inbound had a smaller capacity than outbound, but I didn't want to just assume "inbound can always be half of outbound" or something like that. We could even keep it at 100 probably. The reason for this change is really coming from the outbound side of things.

src/app/config.rs
olix0r: Yeah, let's keep inbound at 100 for now. I can't think of a good reason for it to be so large...

kleimkuhler: Sounds good.

src/app/config.rs
kleimkuhler: @hawkw With adjusting inbound capacity back to 100, if my understanding is correct then this comment is still accurate.

hawkw: @kleimkuhler yup, let's leave it as-is 👍

PR #204, opened 2019-03-04
Configure HTTP/2 window size

kleimkuhler:
Expose an environment configuration to override the default stream or
connection level window size of 65,535.

This allows us to test alternate values without having to recompile.

Closes linkerd/linkerd2#2396

### DIFF
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -123,7 +123,13 @@ pub struct Config {
 
     pub dns_canonicalize_timeout: Duration,
 
-    pub h2_settings: H2Config,
+    pub h2_settings: H2Settings,
+}
+
+#[derive(Copy, Clone, Debug, Default)]
+pub struct H2Settings {
+    pub initial_stream_window_size: Option<u32>,
+    pub initial_connection_window_size: Option<u32>,
 }
 
 /// Configuration settings for binding a listener.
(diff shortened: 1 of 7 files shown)


src/app/config.rs
```diff
@@ -611,6 +630,11 @@ impl<'a> TryFrom<&'a Strings> for Config {
             dns_max_ttl: dns_max_ttl?,
 
             dns_canonicalize_timeout,
+
+            h2_settings: H2Config {
```
hawkw: Pretty minor nit, but I don't like that the field is named `h2_settings` but the type is `h2_config` (this pattern shows up elsewhere in the PR as well). Can we call both either `H2Settings` or `h2_config`?

src/app/config.rs
```diff
@@ -284,6 +292,12 @@ const ENV_DNS_MAX_TTL: &str = "LINKERD2_PROXY_DNS_MAX_TTL";
 /// an uncanonicalized address.
 const ENV_DNS_CANONICALIZE_TIMEOUT: &str = "LINKERD2_PROXY_DNS_CANONICALIZE_TIMEOUT";
 
+/// Configure the stream or connection level flow control setting for HTTP2.
+///
+/// If unspecified, the default value of 65,535 is used.
+const ENV_INITIAL_STREAM_WINDOW_SIZE: &str = "LINKERD2_PROXY_INITIAL_STREAM_WINDOW_SIZE";
+const ENV_INITIAL_CONNECTION_WINDOW_SIZE: &str = "LINKERD2_PROXY_INITIAL_CONNECTION_WINDOW_SIZE";
```
hawkw: As much as I don't like making the already long env var names even longer, I feel like these should be `LINKERD2_PROXY_HTTP2_INITIAL_STREAM_WINDOW_SIZE`/`LINKERD2_PROXY_HTTP2_INITIAL_CONNECTION_WINDOW_SIZE`, to differentiate them from the TCP flow control window?

src/app/control.rs
```diff
@@ -442,7 +443,7 @@ pub mod client {
                 .clone()
                 .with_remote(target.connect.addr)
                 .executor();
-            Ok(http::h2::Connect::new(c, e))
+            Ok(http::h2::Connect::new(c, e, H2Config::default()))
```
hawkw: TIOLI: might be worthwhile making a separate `h2::Connect::new` and `h2::Connect::new_with_config` or something so this doesn't have to change? I dunno.

src/app/config.rs
kleimkuhler: I agree this looks awkward. I like `H2Settings` because it reflects that that these values are HTTP/2 settings that we are overriding: https://http2.github.io/http2-spec/#SETTINGS_INITIAL_WINDOW_SIZE

olix0r:
meta-note: I know that hyper uses the angular-style commit messages (e.g. `feat(client server)`), but I'd prefer not to bring that boilerplate into this project unless it serves a specific purpose.

src/app/control.rs
kleimkuhler: I don't know if I like adding a new method on `Connect` only to reduce the diff here. I prefer this way, but I can definitely make the change if I'm in the minority here.

kleimkuhler:
> meta-note: I know that hyper uses the angular-style commit messages (e.g. `feat(client server)`), but I'd prefer not to bring that boilerplate into this project unless it serves a specific purpose.

@olix0r Sounds good. I'll remove the boilerplate in the "Squash and Merge" part of this PR, and not use that style in the future.

kleimkuhler:
@olix0r Okay the merge conflicts have been resolved. It may show the last commit as a whole new set of changes, but it's the same as it was before. CI is kicked off!

PR #206, opened 2019-03-06
OpenCensus tracing WIP PR

manimaul:
This PR is a draft / conversation starter so that we can discuss the details around adding OpenCensus tracing and to make sure we have alignment on direction and strategy. 

Some open questions:

- Should we assemble a Layer and place it in the request response pipeline?
(This draft PR assumes this approach adding a scaffolding "layer cake" `Layer, Stack, Service <- Tracer`)
- Or should we use some of the existing metrics mechanisms rather than adding a pipeline layer?

This change set is the scaffolding of such a Layer. Eventually, the layer would provide an implementation of a Sampler and Tracer which would employ an Async Exporter. Perhaps the sample could be an enum with associated values eg:
Never,
Always,
Rate(rps),
Percentage(value)

Next would come the evolution of the control plane configuring tracing TBD.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -574,6 +574,7 @@ dependencies = [
  "linkerd2-stack 0.1.0",
  "linkerd2-task 0.1.0",
  "linkerd2-timeout 0.1.0",
+ "linkerd2-tracing 0.1.0",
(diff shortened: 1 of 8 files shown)


olix0r:
An opencensus exporter was added to the proxy last year

PR #207, opened 2019-03-10
Replace futures::sync::mpsc with tokio::sync::mpsc channels

zaharidichev:
Fixes https://github.com/linkerd/linkerd2/issues/2435

### DIFF
--- a/src/drain.rs
+++ b/src/drain.rs
@@ -1,8 +1,8 @@
 use std::mem;
 
 use futures::future::Shared;
-use futures::sync::{mpsc, oneshot};
 use futures::{Async, Future, Poll, Stream};
+use tokio::sync::{mpsc, oneshot};
 
 use never::Never;
 
@@ -12,7 +12,7 @@ use never::Never;
 /// when a drain is signaled.
 pub fn channel() -> (Signal, Watch) {
     let (tx, rx) = oneshot::channel();
-    let (drained_tx, drained_rx) = mpsc::channel(0);
+    let (drained_tx, drained_rx) = mpsc::channel(1);
     (
         Signal { drained_rx, tx },
         Watch {
@@ -136,7 +136,7 @@ impl Future for Drained {
     type Error = ();
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-        match try_ready!(self.drained_rx.poll()) {
+        match try_ready!(self.drained_rx.poll().map_err(|_| ())) {
             Some(never) => match never {},
             None => Ok(Async::Ready(())),
         }
--- a/src/app/profiles.rs
+++ b/src/app/profiles.rs
@@ -1,10 +1,10 @@
-use futures::sync::mpsc;
 use futures::{Async, AsyncSink, Future, Poll, Sink, Stream};
 use http;
 use regex::Regex;
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::executor::{DefaultExecutor, Executor};
+use tokio::sync::mpsc;
 use tokio_timer::{clock, Delay};
 use tower_grpc::{self as grpc, generic::client::GrpcService, Body, BoxBody};
 use tower_retry::budget::Budget;
--- a/src/proxy/canonicalize.rs
+++ b/src/proxy/canonicalize.rs
@@ -9,10 +9,11 @@
 //! DNS TTLs are honored and, if the resolution changes, the inner stack is
 //! rebuilt with the updated value.
 
-use futures::{future, sync::mpsc, Async, Future, Poll, Stream};
+use futures::{future, Async, Future, Poll, Stream};
 use std::time::Duration;
 use std::{error, fmt};
 use tokio::executor::{DefaultExecutor, Executor};
+use tokio::sync::mpsc;
 use tokio_timer::{clock, Delay, Timeout};
 
 use dns;
@@ -180,7 +181,7 @@ impl Future for Task {
                             let resolved = NameAddr::new(refine.name, self.original.port());
                             if self.resolved.get() != Some(&resolved) {
                                 let err = self.tx.try_send(resolved.clone()).err();
-                                if err.map(|e| e.is_disconnected()).unwrap_or(false) {
+                                if err.map(|e| e.is_closed()).unwrap_or(false) {
                                     return Ok(().into());
                                 }
 
@@ -199,7 +200,7 @@ impl Future for Task {
                                     e,
                                 );
                                 let err = self.tx.try_send(self.original.clone()).err();
-                                if err.map(|e| e.is_disconnected()).unwrap_or(false) {
+                                if err.map(|e| e.is_closed()).unwrap_or(false) {
                                     return Ok(().into());
                                 }
 
--- a/src/tap/daemon.rs
+++ b/src/tap/daemon.rs
@@ -1,6 +1,6 @@
-use futures::sync::{mpsc, oneshot};
 use futures::{Async, Future, Poll, Stream};
 use never::Never;
+use tokio::sync::{mpsc, oneshot};
 
 use super::iface::Tap;
 
@@ -86,7 +86,7 @@ impl<T: Tap> Future for Daemon<T> {
                 let err = svc.try_send(tap.clone()).err();
 
                 // If the service has been dropped, make sure that it's not added.
-                dropped = err.as_ref().map(|e| e.is_disconnected()).unwrap_or(false);
+                dropped = err.as_ref().map(|e| e.is_closed()).unwrap_or(false);
 
                 // If service can't receive any more taps, stop trying.
                 if err.is_some() {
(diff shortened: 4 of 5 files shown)


src/drain.rs
```diff
@@ -12,7 +12,7 @@ use never::Never;
 /// when a drain is signaled.
 pub fn channel() -> (Signal, Watch) {
     let (tx, rx) = oneshot::channel();
-    let (drained_tx, drained_rx) = mpsc::channel(0);
+    let (drained_tx, drained_rx) = mpsc::channel(1);
```
zaharidichev: Will be interesting to know whats the significance of a zero sized mpsc::channel. Tokio does not allow that. Would a chan of size one be ok here ?

seanmonstar: The futures channel always had 1 buffer space per sender, so even 0 realistically meant 1. Tokio-sync doesn't add buffer space per sender.

This channel is never actually sent on, it's just used since senders can be cloned, and the receiver can be notified when the last clone drops.

zaharidichev: Ah, I see. Thanks for the clarification

hawkw:
@zaharidichev it looks like the CI build is currently failing because of the rustfmt formatting check. Can you run rustfmt on this branch and push the reformatted code?

As an aside, if you're using an editor that supports this, you might find it helpful to configure your editor to run rustfmt on save, at least for the `linkerd2-proxy` repository.

olix0r:
It looks like CI is failing because `rustfmt` needs to be run in a few places.

zaharidichev:
Sorry was a bit busy the last day. I will run it through rustfmt and try to collect some benchmarks as well as I am interested in the results too.

zaharidichev:
@hawkw 

I ran the benchmarks that you suggested and here are the results I get

bench


Here is the raw data: https://gist.github.com/zaharidichev/6407e3b8a9c0579b3a477b3596f10641

This is averaged across samples. It will be great to know whether the performance regression isexpected. Is it something that is acceptable and moreover is there adequate explanation for it ?

hawkw:
@zaharidichev That's definitely interesting --- since you've averaged across multiple runs, I'd really like to see standard deviations for those measurements.

zaharidichev:
@hawkw  I hope this helps

| BC                         | S0    | S1    | S2    | AVG   | STDEV |
|----------------------------|-------|-------|-------|-------|-------|
| Before H1 Request rate     | 992   | 992   | 995   | 993   | 1     | 
| Before H1 p50 latency (us) | 1844  | 1800  | 1789  | 1811  | 29    |
| Before H1 p75 latency (us) | 2473  | 2429  | 2413  | 2438  | 31    |
| Before H1 p90 latency (us) | 3255  | 3217  | 3181  | 3218  | 37    |
| Before H1 p95 latency (us) | 4137  | 4049  | 3942  | 4043  | 98    |
| Before H1 p99 latency (us) | 7467  | 7051  | 6785  | 7101  | 344   |
| Before H1 p99 latency (us) | 19304 | 16941 | 14810 | 17018 | 2248  |
| Before H2 Request rate     | 1000  | 1000  | 1000  | 1000  | 0     |
| Before H2 p50 latency (us) | 1604  | 1573  | 1514  | 1564  | 46    |
| Before H2 p75 latency (us) | 2209  | 2162  | 2048  | 2140  | 83    |
| Before H2 p90 latency (us) | 3084  | 3030  | 2878  | 2997  | 107   |
| Before H2 p95 latency (us) | 3865  | 3776  | 3484  | 3708  | 199   |
| Before H2 p99 latency (us) | 6220  | 6000  | 5765  | 5995  | 228   |
| Before H2 p99 latency (us) | 18954 | 17208 | 16144 | 17435 | 1419  |
| After H1 Request rate      | 998   | 998   | 1000  | 999   | 1     |
| After H1 p50 latency (us)  | 1998  | 1995  | 2221  | 2071  | 130   |
| After H1 p75 latency (us)  | 2831  | 2827  | 2920  | 2859  | 53    |
| After H1 p90 latency (us)  | 3812  | 3816  | 4050  | 3893  | 136   |
| After H1 p95 latency (us)  | 4396  | 4404  | 5082  | 4627  | 394   |
| After H1 p99 latency (us)  | 9114  | 9175  | 8014  | 8768  | 653   |
| After H1 p99 latency (us)  | 18803 | 18224 | 21225 | 19417 | 1592  |
| After H2 Request rate      | 1000  | 1000  | 1000  | 1000  | 0     |
| After H2 p50 latency (us)  | 2188  | 2193  | 2221  | 2201  | 18    |
| After H2 p75 latency (us)  | 2808  | 2832  | 2920  | 2853  | 59    |
| After H2 p90 latency (us)  | 3824  | 3887  | 4050  | 3920  | 117   |
| After H2 p95 latency (us)  | 4627  | 4786  | 5082  | 4832  | 231   |
| After H2 p99 latency (us)  | 7740  | 7806  | 8014  | 7853  | 143   |
| After H2 p99 latency (us)  | 18134 | 19091 | 21225 | 19483 | 1582  |

carllerche:
As a general note, the `tokio-sync` mpsc channel's `buffer` argument means something different than `futures::sync::mpsc`. If the buffer size is small (less than the number of `Sender` handles), this could be the cause for the performance regression.

The old channel had a capacity of `buffer + num-sender-handles` where as the new channel has a buffer size of `buffer`.

However, off the top of my head, I am not sure which channels are being hit during the benchmark.

zaharidichev:
Just wondering, is this still needed ?

olix0r:
@zaharidichev Sorry for the silence. We still want this! We'd like to do a bit more to diagnose the performance differences, but we've been focused on the upcoming 2.3 release recently.

zaharidichev:
@olix0r  No worries, just let me know whether I can dig into something in order to help out with that. Probably will need quite a bit of help, but I am keen

zaharidichev:
@carllerche Sorry I find it a bit hard to dissect. Where is exactly this pre - allocation happening. Do oyu think you can point me to this part of the code ?

carllerche:
The channel is a [linked list of blocks](https://github.com/tokio-rs/tokio/blob/master/tokio-sync/src/mpsc/block.rs). Each block holds [32 messages](https://github.com/tokio-rs/tokio/blob/master/tokio-sync/src/mpsc/block.rs#L42).

The idea is that usually, when using a channel, many messages will be sent. By allocating storage for messages in blocks, allocations can be reduced.

You can kind of think of the channel as a lazy, infinite, array. When sending into the channel. First [an index for the message is acquired by incrementing a counter](https://github.com/tokio-rs/tokio/blob/master/tokio-sync/src/mpsc/list.rs#L62). Next, the [block for that index is found](https://github.com/tokio-rs/tokio/blob/master/tokio-sync/src/mpsc/list.rs#L65). If the block does not exist yet, it [is allocated](https://github.com/tokio-rs/tokio/blob/master/tokio-sync/src/mpsc/list.rs#L125). Finally, the message is written into [its index](https://github.com/tokio-rs/tokio/blob/master/tokio-sync/src/mpsc/list.rs#L69).

This provides significant gain when sending many messages, but regresses in some other cases.

zaharidichev:
Thanks for the explanation @carllerche. That makes sense. So I ran the benchmarks again and modified the block cap size to be equal to 2 in one of my runs.  Here are the results of my runs. Do you think there is enough performance improvement in the version with BLOCK_CAP=2 to tell us that the preallocation logic is the root of the perf regression ? 

BEFORE: 
---------
    p50 latency (us): 1778
    p75 latency (us): 2561
    p90 latency (us): 3702
    p95 latency (us): 4499
    p99 latency (us): 7396
    p99 latency (us): 13383

AFTER (Block Cap 32):
-----------------------
    p50 latency (us): 2207
    p75 latency (us): 2851
    p90 latency (us): 3777
    p95 latency (us): 4422
    p99 latency (us): 8077
    p99 latency (us): 21796

AFTER (Block Cap 2):
-----------------------
    p50 latency (us): 1987
    p75 latency (us): 2597
    p90 latency (us): 3565
    p95 latency (us): 4345
    p99 latency (us): 7581
    p99 latency (us): 20620

seanmonstar:
It *looks* to me like there may be something else, since the change from 32 to 2 was tiny (though noticeable).

carllerche:
Which channel in the codebase is the benchmark hitting?

zaharidichev:
What do you think is a good way to go forward and find out where the regression comes from? Sorry for all the questions but I am fairly new to this code base and rust and its a tad hard for me navigating it :)

carllerche:
@zaharidichev another thing to try would be to switch all the channels to unbounded channels and see how that impacts the benchmark

zaharidichev:
@carllerche I have updated the results with the run that is using unbounded chans everywhere.

carllerche:
Hmm. Ok at this point, I guess profiling is in order.

olix0r:
We've basically done this organically at this point. Thanks for all of the help @zaharidichev

PR #208, opened 2019-03-11
Update tower-grpc and tower Error types

seanmonstar:
- upgrade from grpc::Error to grpc::Status
- Fix many bounds of `E: Error` to `E: Into<Box<dyn Error>>`

### DIFF
--- a/src/proxy/http/mod.rs
+++ b/src/proxy/http/mod.rs
@@ -29,12 +29,25 @@ pub trait HasH2Reason {
     fn h2_reason(&self) -> Option<::h2::Reason>;
 }
 
+impl HasH2Reason for Box<dyn std::error::Error> {
+    fn h2_reason(&self) -> Option<::h2::Reason> {
+        None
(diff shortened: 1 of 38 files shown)


src/proxy/http/mod.rs
```diff
@@ -29,12 +29,25 @@ pub trait HasH2Reason {
     fn h2_reason(&self) -> Option<::h2::Reason>;
 }
 
+impl HasH2Reason for Box<dyn std::error::Error> {
+    fn h2_reason(&self) -> Option<::h2::Reason> {
+        None
```
olix0r: Should this check `self.source` to see if it implements HasH2Reason? is that even possible?

seanmonstar: Woah, good catch! I meant to fill this in, shoulda left myself a better note.

olix0r:
Though, it appears that CI has some compilation errors...

PR #212, opened 2019-03-14
add rust-toolchain file and bash script to update Docker, Travis and rustup-toolchain files with same version

jxs:
closes [linkerd/2482](https://github.com/linkerd/linkerd2/issues/2482)

Travis file is using the `stable` version, is there a special reason or can just be updated to the matching versions from Dockerfile and rustup-toolchain?

thanks 🙂


update-rust-version.sh
```diff
@@ -0,0 +1,5 @@
+VERSION=$1
```
hawkw: This seems like it should probably have a `#!/bin/bash` shebang?

.travis.yml
```diff
@@ -12,7 +12,7 @@ branches:
 jobs:
   include:
     # Quickly run tests in dev mode.
-    - rust: stable
+    - rust: 1.32.0
```
hawkw: Can we add a comment here that this should be kept in sync with the `rust-toolchain` file and the Dockerfile? It could say something like "rather than updating this manually, run `update-rust-version.sh`...

rust-toolchain
```diff
@@ -0,0 +1 @@
+1.32.0
```
hawkw: Again, a comment pointing to the `update-rust-version` script would probably be nice (if `rust-toolchain` supports comments? it may not, now that I think of it...)

update-rust-version.sh
jxs: yes totally forgot it 😅

rust-toolchain
jxs: from the docs, and the code I saw doesn't seem to be possible

.travis.yml
jxs: added :) also added to the Dockerfile

rust-toolchain
hawkw: 👍 okay, nevermind, then! thanks for looking into it!

olix0r:
> It's probably best for Travis to be using the same version as the Dockerfile , so I think you made the right call.

I think I disagree with this... I'd rather that we test and publish from the latest master so that we discover issues eagerly, and also benefit from fixes. There may be a point in the future we we'd prefer to have more process around version updates, but it feels like a step backwards now.

jxs:
@olix0r so, want me to update the bash script and remove the `travis.yml` file update?

olix0r:
@jxs Yeah, let's do that for now, and we can add that back later if it seems important.

hawkw:
> > It's probably best for Travis to be using the same version as the Dockerfile , so I think you made the right call.
> 
> I think I disagree with this... I'd rather that we test and publish from the latest stable so that we discover issues eagerly, and also benefit from fixes. There may be a point in the future we we'd prefer to have more process around version updates, but it feels like a step backwards now.

@olix0r What about having a travis build against the latest stable _and_ one against the version built in the Dockerfile? We could use the latest stable build to be aware of upcoming breaking changes, and the pinned version build to verify that the dockerfile will actually build a working proxy image.

olix0r:
@hawkw and I discussed this in realtime... I'll summarize:

The problem we care most about addressing is when a new developer is (1) running a very old rust version, so they cannot compile a recent proxy or (2, more likely) the user has a nightly that supports features we don't accept in our build.

For this case, we want a `rust-toolchain` to force users in case (1) to a newer minimum rust version and we want to force users in case (2) to use a stable rust.

To this end, I think it's okay if we continue to use `stable` in CI so that release artifacts are published from the most-recent Rust stable. We don't expect that newer rusts in CI will break existing code, so this is okay.

Then, we can manually bump rust-toolchain versions with the script when we take a new minimum version for development.

In summary: let's leave the travis.yml as it was.

jxs:
thanks for the feedback loop
updated!

PR #213, opened 2019-03-14
Replace fs-watch based TLS with new Identity service

olix0r:
This change replaces the proxy's TLS configuration logic so that certificates
are dynamically provisioned by the proxy via a new _Identity_ controller service.

When the proxy processes its configuration, it now requires that a
`LINKERD2_PROXY_IDENTITY_DISABLED` environment is set; oterhwise,
a valid identity configuration must be provided. An identity configuration
includes:

- PEM-encoded trust anchors provided via the
  `LINKERD2_PROXY_IDENTITY_TRUST_ANCHORS` environment;
-  A directory, specified by `LINKERD2_PROXY_IDENTITY_DIR`, containing two
  files: `key.p8` and `csr.der`. The key includes a PKCS#8-encoded private
  key; and the CSR includes a certificate request, signed by the private key,
  that can be sent to the identity service;
- A serviceaccount token specified by `LINKERD2_PROXY_IDENTITY_TOKEN_FILE`
  that the proxy will read each time it requests a new certificate.

The `linkerd2_fs_watch` crate is no longer used and should probably be split
out into a utility (perhaps tokio-fs?).

Various other configuration options have changed for consistency:
- URLs are no longer accepted when we only care about addresses.
- Configuration specific to the destination service (as opposed to all
  controller clients) has been renamed appropriately.

Once the configuration is processed---if Identity is not completely
disabled---then a client is established to the identity service and its
identity is validated. As a certificate is acquired, it is validated and
published into the _listeners_ and _connect stacks_ (rather than into the
routing stack). This removes the need for rebinding watch layers, since these
connection layers lazily acquire the proper configuration.

When a certificate has not yet been acquired, then ALL inbound TLS
connections that should be terminated by the process fail, and all outbound
connections are server-validated but no client identity is provided. Once the
certificate is acquired, client identities are provided on subsequent
connections and the listener will properly accept connections.

---

Depends on https://github.com/linkerd/linkerd2-proxy-api/pull/25


Dockerfile
```diff
@@ -8,7 +8,7 @@
 # runtime performance.
 
 ARG RUST_IMAGE=rust:1.32.0
-ARG RUNTIME_IMAGE=gcr.io/linkerd-io/base:2017-10-30.01
+ARG RUNTIME_IMAGE=gcr.io/linkerd-io/proxy:git-820f9cd8
```
olix0r: Note that this dockerfile now uses a proxy image so that we can just overwrite the binary. This is now necessary because we require utilities built in the other repo.

olix0r:
I should note that three tests are now disabled... two of which require a mock identity service implementation for integration tests

src/app/identity.rs
```diff
@@ -0,0 +1,220 @@
...
+// === impl Config ===
+
+impl Config {
+    fn refresh(&self, expiry: SystemTime) -> Delay {
+        let now = clock::now();
+
+        let refresh = match expiry
+            .duration_since(SystemTime::now())
+            .ok()
+            .map(|d| d * 7 / 10)
```
seanmonstar: Mind explaining what this math wizardry is doing? Or else we'll forget in a couple months.

olix0r: Ah, yes. The intent is to get to ~70% of the expiration time. I'll add a comment.

Is there a better way to accomplish that math?

src/app/identity.rs
```diff
@@ -0,0 +1,220 @@
...
+                    Inner::ShouldRefresh
+                }
+                Inner::ShouldRefresh => match self.config.token.load() {
+                    Ok(token) => {
+                        let req = grpc::Request::new(api::CertifyRequest {
+                            token,
+                            identity: self.config.local_name.as_ref().to_owned(),
+                            certificate_signing_request: self.config.csr.to_vec(),
+                        });
+                        let f = self.client.certify(req);
```
seanmonstar: Recently, the poll-ready-before-call contract was made strong in tower. While there's likely a `buffer` service protecting against most cases, many services will now *panic* instead return an error future if called and they weren't ready...

olix0r: This service is intended to be buffered, though it's not documented in the type system, so I agree we should poll-ready

src/app/identity.rs
seanmonstar: Oh sorry, yes I recognized it as 70%. I just didn't get the *why*.

I guess when we get a token, it has an expiration point. And this is saying to start checking for a new token once we're within 30% of it expiring?

hawkw: +1 for a comment

src/app/main.rs
```diff
@@ -260,38 +243,80 @@ where
...
+                );
+
+                Conditional::Some(local_identity)
+            }
+        };
+
+        let dst_svc = config.destination_addr.as_ref().map(|addr| {
+            use super::control;
+
+            //let tls_server_identity = config.destination_identity.clone();
```
hawkw: Is there a reason we commented this out rather than removing it? AFAICT this is no longer necessary since it's now provided by the `ControlAddr` type?

src/app/main.rs
```diff
@@ -703,14 +722,14 @@ where
                             .future(resolver_bg_rx.map_err(|_| {}).flatten()),
                     );
 
-                    rt.spawn(::logging::admin().bg("tls-config").future(tls_cfg_bg));
+                    //rt.spawn(::logging::admin().bg("tls-config").future(tls_cfg_bg));
```
hawkw: Nit: another commented out line that I think needs to just be deleted?

src/identity/mod.rs
```diff
@@ -0,0 +1,462 @@
...
+pub use dns::InvalidName;
+
+pub trait LocalIdentity {
+    fn name(&self) -> &Name;
+    fn credentials(&self) -> Option<&CrtKey>;
+}
+
+/// A DER-encoded X.509 certificate signing request.
+#[derive(Clone, Debug)]
+pub struct CSR(Arc<Vec<u8>>);
```
hawkw: Super nitpicky, but it's considered idiomatic in Rust to not make acronyms all upper case in type names, so this should be `Csr`...

src/lib.rs
```diff
@@ -1,11 +1,9 @@
-#![cfg_attr(feature = "cargo-clippy", allow(clone_on_ref_ptr))]
-#![cfg_attr(feature = "cargo-clippy", allow(new_without_default_derive))]
 #![deny(warnings)]
+#![allow(dead_code, unused_imports)] // FIXME
```
hawkw: I'm guessing you were going to remove this prior to merging?

src/transport/tls/accept.rs
```diff
@@ -0,0 +1,102 @@
+
```
hawkw: It's hard to tell what changed here, since it was also split out into a new file..

src/app/config.rs
```diff
@@ -404,155 +354,64 @@ impl<'a> TryFrom<&'a Strings> for Config {
...
-        let tls_private_key = parse(strings, ENV_TLS_PRIVATE_KEY, parse_path);
-        let tls_pod_identity_template = strings.get(ENV_TLS_POD_IDENTITY);
-        let tls_controller_identity = strings.get(ENV_TLS_CONTROLLER_IDENTITY);
+        // DNS
 
         let resolv_conf_path = strings.get(ENV_RESOLV_CONF);
 
-        let metrics_retain_idle = parse(strings, ENV_METRICS_RETAIN_IDLE, parse_duration);
-
         let dns_min_ttl = parse(strings, ENV_DNS_MIN_TTL, parse_duration);
```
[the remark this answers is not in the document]
olix0r: Yeah, I think this is a good general direction -- each app module should have its own config struct, and then the main config i sresponsible for parsing/validating them.

src/app/config.rs
```diff
@@ -786,6 +642,171 @@ fn parse_dns_suffix(s: &str) -> Result<dns::Suffix, ParseError> {
...
+                p.set_extension("der");
+
+                fs::read(p)
+                    .map_err(|e| {
+                        error!("Failed to read Csr: {}", e);
+                        Error::InvalidEnvVar
+                    })
+                    .and_then(|b| {
+                        identity::Csr::from_der(b).ok_or_else(|| {
+                            error!("No Csr found");
```
hawkw: In the logs, the acronym should probably still be upper-case...I'm assuming this is due to a find and replace?
```suggestion
                            error!("No CSR found");
```

PR #217, opened 2019-03-17
Instrument identity with logging

olix0r:
This changes initialization logic so that a task is spawn that logs when
identity is provisioned (as observed by stacks). Identity's daemon task is
now spawned on the admin thread with context logging.

Furthermore, the processes's uptime is now logged in log messages so that the
relative time between events is easily observeable. This is borrowed from the
Linux kernel's behavior (i.e. in /var/log/kern.log).

### DIFF
--- a/src/app/identity.rs
+++ b/src/app/identity.rs
@@ -104,6 +104,10 @@ impl Local {
         (l, s)
     }
 
+    pub fn name(&self) -> &Name {
+        &self.name
+    }
+
     pub fn await_crt(self) -> AwaitCrt {
         AwaitCrt(Some(self))
     }
--- a/src/app/identity.rs
+++ b/src/app/identity.rs
@@ -11,6 +11,7 @@ use never::Never;
 pub use identity::{Crt, CrtKey, Csr, InvalidName, Key, Name, TokenSource, TrustAnchors};
 use transport::tls;
 
+/// Configures the Identity service and local identity.
 #[derive(Clone, Debug)]
 pub struct Config {
     pub svc: super::control::ControlAddr,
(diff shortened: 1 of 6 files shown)


src/app/identity.rs
```diff
@@ -164,63 +185,95 @@ where
...
+                                leaf_certificate,
+                                intermediate_certificates,
+                                valid_until,
+                            } = rsp.into_inner();
+
+                            match valid_until
+                                .and_then(|d| Result::<SystemTime, Duration>::from(d).ok())
+                            {
+                                None => error!(
+                                    "Identity service did not specify a ceritificate expiration."
```
hawkw: Typo: "ceritificate"
```suggestion
                                    "Identity service did not specify a certificate expiration."
```

src/logging.rs
```diff
@@ -17,20 +17,24 @@ thread_local! {
...
+        .format(move |fmt, record| {
+            CONTEXT.with(move |ctxt| {
                 let level = match record.level() {
                     Level::Trace => "TRCE",
                     Level::Debug => "DBUG",
                     Level::Info => "INFO",
                     Level::Warn => "WARN",
                     Level::Error => "ERR!",
                 };
+                let uptime = clock::now() - start_time;
```
hawkw: How concerned are we about the performance impact of adding timestamps? 

I believe that (unlike `SystemTime::now`) whether or not `clock::now()` will require a syscall every time may be platform-dependant, but there's still some added overhead.

If this is a concern, we could instead include timestamps in the messages of certain log records where knowing the time of the event matters (such as the identity daemon's log messages)...

olix0r: I'd like to have the uptime counter if possible--it helps to understand relative timing.

How can we determine whether `clock::now` invokes a syscall? We have a pretty specific target platform...

hawkw: By default, I believe it depends on the behaviour of `Instant::now`. IIRC, someone (possibly @carllerche?) told me at one point that in some cloud provider's environment, `Instant::now` always invoked a syscall but in others it didn't. Sorry I don't recall all the details around this OTTOMH.

carllerche: I would not worry about it until it is a problem.

PR #218, opened 2019-03-17
Replace the metrics server with an admin server

olix0r:
The so-called "metrics" server need not only serve metrics. It can also
serve other administrative endpoints, especially those that may be used
as readiness and liveness probes.

A /ready endpoint is exposed that responds successfully when identity
has been loaded.

The `METRICS_LISTEN_ADDR` configuration has been renamed as
`ADMIN_LISTEN_ADDR`.

### DIFF
--- a/src/app/admin/readiness.rs
+++ b/src/app/admin/readiness.rs
@@ -1,0 +1,25 @@
+use std::sync::{Arc, Weak};
+
+#[derive(Clone, Debug)]
+pub struct Readiness(Weak<()>);
+
+#[derive(Clone, Debug)]
+pub struct Latch(Arc<()>);
+
+impl Readiness {
+    pub fn new() -> (Readiness, Latch) {
+        let r = Arc::new(());
+        (Readiness(Arc::downgrade(&r)), Latch(r))
+    }
+
+    pub fn ready(&self) -> bool {
+        self.0.upgrade().is_none()
+    }
+}
+
+impl Default for Readiness {
+    fn default() -> Self {
+        // Is immediately ready.
+        Self::new().0
+    }
+}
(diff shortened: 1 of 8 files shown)


src/app/main.rs
```diff
@@ -204,7 +205,7 @@ where
         );
         info!(
             "serving Prometheus metrics on {:?}",
```
hawkw: Nit: perhaps this should be
```suggestion
            "serving admin endpoint metrics on {:?}",
```
or something now?

src/app/main.rs
```diff
@@ -298,6 +303,7 @@ where
                         .clone()
                         .await_crt()
                         .map(move |id| {
+                            drop(ready_latch);
```
hawkw: TIOLI: the fact that dropping the latch indicates readiness seems a little non-obvious without actually reading the `ready` module.

What about adding a function like 
```rust
impl Latch {
    pub fn set_ready(self) {
        drop(self);
    }
}
```
so that this could be `ready_latch.set_ready()` instead? 

Admittedly, this adds some more code that isn't _strictly_ necessary, but it communicates the semantic meaning of this drop a little better. Up to you!

olix0r: yeah that seems fine. the main goal was to _require_ acting on the latch to signal readiness. if you drop them all, it should just be ready. but this is clearer as a golden-path api.

src/app/admin/readiness.rs
```diff
@@ -12,14 +16,21 @@ impl Readiness {
...
 impl Default for Readiness {
     fn default() -> Self {
-        // Is immediately ready.
         Self::new().0
     }
 }
+
+impl Latch {
+    /// Releases this readiness latch.
+    pub fn release(self) {
```
hawkw: 👍

PR #220, opened 2019-03-21
Add a oneshot to notify the profiles daemon if the stream is dropped

adleong:
Each service in the router cache maintains a profile watch on the destination service which is driven by the profile daemon future.  If that service is dropped from the cache, the profile daemon will not realize this until it gets a profile update and attempts to send the profile to the service.  As long as the profile grpc stream is NotReady (i.e. as long as there are no profile updates), the daemon future will live on, even after the service has been dropped.

As services churn into and out of the router cache, this results in more and more daemon futures accumulating.  Each future is associated with a grpc stream to the destination service so this causes an unbounded number of grpc streams to be open to the destination service.  This results in a leak of goroutines and therefore memory in the destination service.

We solve this by adding a oneshot from the stream to the daemon so that the daemon will be notified when the stream is dropped.  This will cause the future to complete and the grpc stream to be closed.

Fixes https://github.com/linkerd/linkerd2/issues/2523

### DIFF
--- a/src/app/profiles.rs
+++ b/src/app/profiles.rs
@@ -24,7 +24,7 @@ pub struct Client<T> {
 
 pub struct Rx {
     rx: mpsc::Receiver<profiles::Routes>,
-    _hangup: oneshot::Sender<()>,
+    _hangup: oneshot::Sender<Never>,
 }
 
 struct Daemon<T>
@@ -37,7 +37,7 @@ where
     state: State<T>,
     tx: mpsc::Sender<profiles::Routes>,
     context_token: String,
-    hangup: oneshot::Receiver<()>,
+    hangup: oneshot::Receiver<Never>,
 }
 
 enum State<T>
@@ -129,7 +129,7 @@ where
     fn proxy_stream(
         rx: &mut grpc::Streaming<api::DestinationProfile, T::ResponseBody>,
         tx: &mut mpsc::Sender<profiles::Routes>,
-        hangup: &mut oneshot::Receiver<()>,
+        hangup: &mut oneshot::Receiver<Never>,
     ) -> Async<StreamState> {
         loop {
             match tx.poll_ready() {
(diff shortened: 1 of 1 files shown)


src/app/profiles.rs
```diff
@@ -127,7 +139,24 @@ where
             }
 
             match rx.poll() {
-                Ok(Async::NotReady) => return Async::NotReady,
+                Ok(Async::NotReady) => match hangup.poll() {
+                    Ok(Async::Ready(_)) => {
+                        // We should never get here, the hangup tx should never
+                        // be written to.
+                        warn!("profile stream hangup");
+                        return StreamState::SendLost.into();
```
olix0r: Consider making this `unreachable!()`, since it's technically a "programming error" if we enter this branch?

src/app/profiles.rs
```diff
@@ -22,7 +22,10 @@ pub struct Client<T> {
     context_token: String,
 }
 
-pub struct Rx(mpsc::Receiver<profiles::Routes>);
+pub struct Rx {
+    rx: mpsc::Receiver<profiles::Routes>,
+    _hangup: oneshot::Sender<()>,
```
olix0r: It would be useful to add an internal (`//` not `///`) comment explaining that it's just there to be dropped and we'll never actually send on it.

src/app/profiles.rs
seanmonstar: Or you could make it literally impossible by making the message type `Never`. You can `match never {}`, the compiler is happy, and eliminates the branch entirely.

src/app/profiles.rs
olix0r: If you follow @seanmonstar's `never` suggestion, the comment isn't really necessary as the type signature will be clearer about its use.

src/app/profiles.rs
adleong: I love types

src/app/profiles.rs
```diff
@@ -206,13 +230,15 @@ where
                         State::Backoff(Delay::new(clock::now() + self.backoff))
                     }
                 },
-                State::Streaming(ref mut s) => match Self::proxy_stream(s, &mut self.tx) {
-                    Async::NotReady => return Ok(Async::NotReady),
-                    Async::Ready(StreamState::SendLost) => return Ok(().into()),
-                    Async::Ready(StreamState::RecvDone) => {
-                        State::Backoff(Delay::new(clock::now() + self.backoff))
+                State::Streaming(ref mut s) => {
+                    match Self::proxy_stream(s, &mut self.tx, &mut self.hangup) {
```
hawkw: It took me a moment to figure out why the whole match had "changed" in the diff", but I realised that adding `&mut self.hangup` to the call to `proxy_stream` had made the line too long, triggering `rustfmt` to reformat this block. So that's fine actually.

PR #222, opened 2019-03-25
Add an http::errors::layer to translate errors

seanmonstar:
This is just the beginning of it, to be used by other pieces later.

### DIFF
--- a/src/app/errors.rs
+++ b/src/app/errors.rs
@@ -1,0 +1,118 @@
+//! Layer to map HTTP service errors into appropriate `http::Response`s.
+
+use futures::{Future, Poll};
+use http::{header, Request, Response, StatusCode};
+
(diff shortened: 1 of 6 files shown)


seanmonstar:
Hmmm, should this be part of `app` instead? Seems like a similar role to `app::classify`...

olix0r:
@seanmonstar wdyt about moving `map_err_to_5xx` to `app` and making the layer take a `ToStatus` (or something) generic argument -- or even just a pure `fn` if that works out?

seanmonstar:
I've been planning as well to add support to create an appropriate h2 reason if the request was h2 (like refused).

seanmonstar:
For now, I've just moved it entire to `app::errors`.

PR #225, opened 2019-03-26
Add load shedding when over max-in-flight requests.

seanmonstar:
Also adds configuration for inbound and outbound max-in-flight requests.

### DIFF
--- a/src/app/errors.rs
+++ b/src/app/errors.rs
@@ -100,9 +100,13 @@ where
 
 fn map_err_to_5xx(e: Error) -> StatusCode {
     use proxy::http::router::error as router;
+    use proxy::shed::error as shed;
 
     if let Some(ref c) = e.downcast_ref::<router::NoCapacity>() {
         warn!("router at capacity ({})", c.0);
+        http::StatusCode::SERVICE_UNAVAILABLE
+    } else if let Some(_) = e.downcast_ref::<shed::Overloaded>() {
+        warn!("server overloaded, max-in-flight reached");
         http::StatusCode::SERVICE_UNAVAILABLE
     } else if let Some(ref r) = e.downcast_ref::<router::MakeRoute>() {
         error!("router error: {:?}", r);
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -23,7 +23,7 @@ use proxy::{
         client, insert_target, metrics as http_metrics, normalize_uri, profiles, router, settings,
         strip_header,
     },
-    limit, reconnect,
+    limit, reconnect, shed,
 };
 use svc::{
     self, shared,
@@ -186,8 +186,6 @@ where
             outbound_listener,
             admin_listener,
         } = self;
-
-        const MAX_IN_FLIGHT: usize = 10_000;
 
         const EWMA_DEFAULT_RTT: Duration = Duration::from_millis(30);
         const EWMA_DECAY: Duration = Duration::from_secs(10);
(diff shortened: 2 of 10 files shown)


src/app/errors.rs
```diff
@@ -100,10 +100,14 @@ where
 
 fn map_err_to_5xx(e: Error) -> StatusCode {
     use proxy::http::router::error as router;
+    use proxy::shed::error as shed;
 
     if let Some(ref c) = e.downcast_ref::<router::NoCapacity>() {
         warn!("router at capacity ({})", c.0);
         http::StatusCode::SERVICE_UNAVAILABLE
+    } else if let Some(_) = e.downcast_ref::<shed::Overloaded>() {
+        warn!("server overloaded, max-in-flight reached");
```
hawkw: TIOLI: should we log the inflight limit here?

src/app/main.rs
```diff
@@ -460,7 +459,7 @@ where
             // 6. Strips any `l5d-server-id` that may have been received from
             //    the server, before we apply our own.
             let endpoint_stack = client_stack
-                .push(buffer::layer(MAX_IN_FLIGHT))
+                .push(buffer::layer(max_in_flight))
```
kleimkuhler: TIOLI: I find the the ordering that we push layers in the `inbound`/`outbound` routers to be helpful in understanding what a layer should do in the stack. I think this layer makes more sense right next to the `router`, instead of having the `strip_header`s in between.

When I first started looking at this part of the proxy, I placed a lot of significance on the ordering of layers when sometimes there it may not matter. I would have been confused why we had `strip_header` layers come in between this and the `router` layer.

src/app/main.rs
```diff
@@ -555,8 +554,7 @@ where
             // 4. Finally, if the Source had an SO_ORIGINAL_DST, this TCP
             // address is used.
             let addr_router = addr_stack
-                .push(buffer::layer(MAX_IN_FLIGHT))
-                .push(limit::layer(MAX_IN_FLIGHT))
+                .push(buffer::layer(max_in_flight))
```
kleimkuhler: Same TIOLI about understanding.

src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![deny(warnings)]
+//#![deny(warnings)]
```
kleimkuhler: Looks like this snuck in here.

src/app/main.rs
```diff
@@ -574,6 +572,9 @@ where
                         })
                         .ok()
                 }))
+                // `limit` must be shared to apply to all connections...
+                .push(limit::layer(max_in_flight))
```
kleimkuhler: If I understand what is meant by applying to all connections, wouldn't this *outer* limit layer load shed before any of the routers below it since they are all the same value? Is it redundant to have limit layers in any of the routers above this?

seanmonstar: > wouldn't this outer limit layer load shed before any of the routers below it since they are all the same value?

Yes, that's the goal, since the routers can siphon off requests in different buckets, but we want the *total* limit.

src/app/errors.rs
seanmonstar: We don't have the limit available here... 🤷‍♂️

hawkw: Ahh yeah, nevermind. Might eventually be nice to add that to the `shed::Overloaded` error type?

seanmonstar:
> Does not bound the number of open streams?

Correct, the concurrency-limit releases when the response future completes, not when the stream completes: https://github.com/tower-rs/tower/blob/979c1399127bdb269c751b99c12b3adfad55463c/tower-limit/src/concurrency/future.rs#L33

PR #227, opened 2019-03-28
Add test support infrastructure for identity

hawkw:
This branch adds a mock Identity service implementation to the proxy's
test-support module, and wires up the rest of the test support code to
allow mocking the Identity service in tests. The existing tests for the
`l5d-server-id` header have been updated to work with the identity
service.

Note that this branch does *not* add any new tests for the proxy's
behaviour with the Identity service. I was planning to add these in a
follow-up PR.

See also linkerd/linkerd2#2505

Signed-off-by: Eliza Weisma <eliza@buoyant.io>


tests/support/identity.rs
```diff
@@ -0,0 +1,164 @@
...
+    expect_calls: Arc<Mutex<VecDeque<Certify>>>,
+}
+
+type Certify = Box<
+    Fn(
+            pb::CertifyRequest,
+        )
+            -> Box<Future<Item = grpc::Response<pb::CertifyResponse>, Error = grpc::Status> + Send>
+        + Send,
+>;
```
hawkw: I want to note here that I did not choose this formatting, this is what `rustfmt` thought would look good.

tests/support/controller.rs
```diff
@@ -245,28 +264,29 @@ fn run(
...
+                .map_err(move |e| println!("{} error: {:?}", name, e));
 
             runtime.spawn(serve);
-            runtime.block_on(rx).expect("support controller run");
+            runtime.block_on(rx).expect(name);
         })
         .unwrap();
 
     listening_rx.wait().expect("listening_rx");
+    println!("{} listening; addr={:?}", name, addr);
```
olix0r: rm or make a proper debug line

olix0r: nevermind, i see that this already does println!s for what i assume are good reasons.

src/app/config.rs
```diff
@@ -158,6 +158,7 @@ pub trait Strings {
 /// An implementation of `Strings` that reads the values from environment variables.
 pub struct Env;
 
+#[derive(Clone)]
```
olix0r: just curious.... can this be `#[cfg(test)]`?

tests/support/controller.rs
hawkw: @olix0r most of the test support code uses `println!`; the idea is that it should be captured and only printed out if the test fails (if printed from the main test thread). I tried to follow the existing convention with the rest of the test support code. 

Happy to change it though.

src/app/config.rs
seanmonstar: I don't think so, since the tests are compiled as the binary in test mode with the proxy as just a regular dependency.

tests/support/identity.rs
seanmonstar: wat

PR #230, opened 2019-04-01
Support endpoint weights

olix0r:
This change modifies the proxy to honor weights provided by the
destination service. When the destination service replies with a
weight, this value is divided by 10,000 to produce a weight on
[0.0, ~400000.0]. This weight is used by load the load balancer
to modify load interpretation and therefore request distribution.

A weight of 0.0 will cause the endpoint's load to be effectively infinite
so that requests will only be sent to the endpoint when no other endpoints
exists or when the other endpoints that were considered had 0-weights.


src/app/outbound.rs
```diff
@@ -66,6 +67,17 @@ impl connect::HasPeerAddr for Endpoint {
...
 }
 
+impl HasWeight for Endpoint {
+    fn weight(&self) -> Weight {
+        let w =self.metadata.weight();
+        if w == 0 {
+            1.0.into()
+        } else {
+            ((w as f64) / 1000.0).into()
+        }
```
olix0r: this mapping isn't right -- we shouldn't prevent 0 values

src/control/destination/mod.rs
```diff
@@ -180,18 +182,21 @@ impl Metadata {
             labels: IndexMap::default(),
             protocol_hint: ProtocolHint::Unknown,
             identity: None,
+            weight: 10_000,
```
kleimkuhler: Could we parameterize the `10_000`? I was little confused why the number was chosen in [here](https://github.com/linkerd/linkerd2-proxy/pull/230/files#diff-259de4b6a0f288cb0b854d2c3e6062ffR73).

src/proxy/http/balance.rs
```diff
@@ -131,3 +133,69 @@ where
...
+
+    pub fn layer() -> Layer {
+        Layer(())
+    }
+
+    impl<M> svc::Layer<M> for Layer {
+        type Service = MakeSvc<M>;
+
+        fn layer(&self, inner: M) -> Self::Service {
+            Self::Service { inner }
```
hawkw: huh, TIL that you can do this!

src/proxy/http/balance.rs
```diff
@@ -131,3 +133,69 @@ where
...
+        inner: M,
+    }
+
+    #[derive(Debug)]
+    pub struct MakeFuture<F> {
+        inner: F,
+        weight: Weight,
+    }
+
+    pub fn layer() -> Layer {
```
hawkw: TIOLI: It might be somewhat more concise to just use the [`svc::layer::mk` fn](https://github.com/linkerd/linkerd2-proxy/blob/master/src/svc.rs#L5), like how `pending` does:
https://github.com/linkerd/linkerd2-proxy/blob/61db2e77a247f7b0235b67581f60e8a92f8543cb/src/proxy/pending.rs#L23-L25

this avoids some of the boilerplate around having a layer type and impl that don't really do any _behavior_.

src/control/destination/mod.rs
olix0r: we can constify it... but this just equals "1.0" -- the grpc API sends weight as a uint32. We represent 1.0 as 10_000 so the smallest value we can represent is 0.0001.  Practically, it doesn't matter at all because we'll _always_ set a value from the API when balancing is used (once eliza's changes are in, anyway)

olix0r: I've added a comment and consolidated the logic in that file.

PR #231, opened 2019-04-01
Disable potentially info-leaking headers

olix0r:
The proxy has been instrumented to expose various informational `l5d-`
headers to expose identity and network information that would otherwise
be unaccessible to the application.

However, as described in linkerd/linkerd2#2597, when linkerd is injected
into an ingress pod, its easy to accidentally leak these details to
external applications.

Until we have a better mechanism for flagging external-facing pods,
these headers should be disabled.

### DIFF
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -413,7 +413,9 @@ where
 
         let outbound = {
             use super::outbound::{
-                add_remote_ip_on_rsp, add_server_id_on_rsp, discovery::Resolve, orig_proto_upgrade,
+                //add_remote_ip_on_rsp, add_server_id_on_rsp,
+                discovery::Resolve,
+                orig_proto_upgrade,
                 Endpoint,
             };
             use proxy::{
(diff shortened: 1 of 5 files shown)


src/app/main.rs
```diff
@@ -728,9 +733,9 @@ where
             let source_stack = dst_router
                 .push(orig_proto_downgrade::layer())
                 .push(insert_target::layer())
-                .push(set_remote_ip_on_req::layer())
+                //.push(set_remote_ip_on_req::layer())
+                //.push(set_client_id_on_req::layer())
```
hawkw: Is it necessary to no longer set inbound client IDs? AFAICT that should _not_ make it out of the cluster even when injected into an ingress controller?

olix0r: I don't think there's any value in leaving this stuff half-on.

hawkw: Fair enough!

hawkw:
FWIW, I do recall having a conversation on Linkerd slack with at least one user who was (at least planning on) using the `remote-ip` header, so this should probably be noted as a breaking change.

kleimkuhler:
@olix0r Do we want to pull this in for the edge?

PR #233, opened 2019-04-11
Do not block listener on TLS detection

olix0r:
Currently, the listener task does not multiplex TLS discovery, so one
slow/idle connection can prevent a proxy from transmitting all requests.

This change alters the listener to use buffer_unordered and repurposes
the listener's "connection limit" to control this buffer's capacity
(rather than the total number of connections that may be served).

### DIFF
--- a/src/transport/tls/listen.rs
+++ b/src/transport/tls/listen.rs
@@ -117,13 +117,13 @@ impl<L: HasConfig, G> Listen<L, G> {
         <Fut as IntoFuture>::Future: Send,
         Self: GetOriginalDst + Send + 'static,
     {
-        self.listen_and_fold_inner(std::u64::MAX, initial, f)
+        self.listen_and_fold_inner(std::usize::MAX, initial, f)
     }
 
     #[cfg(test)]
     pub fn listen_and_fold_n<T, F, Fut>(
         self,
-        connection_limit: u64,
+        connection_limit: usize,
         initial: T,
         f: F,
     ) -> impl Future<Item = (), Error = io::Error> + Send + 'static
@@ -141,7 +141,7 @@ impl<L: HasConfig, G> Listen<L, G> {
 
     fn listen_and_fold_inner<T, F, Fut>(
         mut self,
-        connection_limit: u64,
+        connection_limit: usize,
         initial: T,
         f: F,
     ) -> impl Future<Item = (), Error = io::Error> + Send + 'static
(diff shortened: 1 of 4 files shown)


src/transport/tls/listen.rs
```diff
@@ -173,8 +173,7 @@ impl<L: HasConfig, G> Listen<L, G> {
             });
 
             incoming
-                .take(connection_limit)
```
hawkw: I believe that at one point, the ability to limit the number of accepted connections over the lifetime of the process was required for some kind of testing thing? I can no longer remember the rationale, but is it still necessary?

src/transport/tls/listen.rs
```diff
@@ -117,13 +117,13 @@ impl<L: HasConfig, G> Listen<L, G> {
...
     {
-        self.listen_and_fold_inner(std::u64::MAX, initial, f)
+        self.listen_and_fold_inner(std::usize::MAX, initial, f)
     }
 
     #[cfg(test)]
     pub fn listen_and_fold_n<T, F, Fut>(
         self,
-        connection_limit: u64,
+        connection_limit: usize,
```
hawkw: Is the change from `u64` to `usize` related to the fix here? Just curious.

src/transport/tls/listen.rs
olix0r: Given that we run this with a MAX value, no.

src/transport/tls/listen.rs
olix0r: yeah, buffered takes usize

tests/identity.rs
```diff
@@ -12,6 +12,41 @@ use std::{
...
+    let proxy = proxy.inbound(srv).run_with_test_env(env);
+    let client = client::tcp(proxy.inbound);
+
+    // Create an idle connection and then an active connection. Ensure that
+    // protocol detection on the idle connection does not block communication on
+    // the active connection.
+    let _idle = client.connect();
+    let active = client.connect();
+    active.write(msg1);
+    assert_eq!(active.read_timeout(Duration::from_secs(2)), msg2.as_bytes());
```
kleimkuhler: Thoughts on making this timeout a little shorter? The probes were sometimes getting responses in under 2 seconds, but that was still too high.

olix0r: We need this to be high enough to withstand flakey CI.

In the case you were seeing, other connections were closing. In this case, we're holding the idle connection open so the active one can't be processed.

kleimkuhler: Oh okay makes sense!

PR #234, opened 2019-04-17
Add support for mocking DNS refinement

hawkw:
This branch adds support for mocking the source of DNS name refinement
used by the proxy. A new trait, `dns::Refine`, is added, which is
implemented by the `dns::resolver` type, and the appropriate plumbing is
added to the test support code.

Fixes linkerd/linkerd2##2706

### DIFF
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -54,12 +54,12 @@ use super::profiles::Client as ProfilesClient;
 ///
 /// The private listener routes requests to service-discovery-aware load-balancer.
 ///
-pub struct Main<G, R=dns::DefaultRefine> {
+pub struct Main<G, R = dns::DefaultRefine> {
     proxy_parts: ProxyParts<G, R>,
     runtime: task::MainRuntime,
 }
 
-struct ProxyParts<G, R=dns::DefaultRefine> {
+struct ProxyParts<G, R = dns::DefaultRefine> {
     config: Config,
     identity: tls::Conditional<(identity::Local, identity::CrtKeyStore)>,
 
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -134,7 +134,7 @@ where
     R: dns::NewResolver + Send,
     R::Resolver: Clone + Send + Sync + 'static,
 {
-    pub fn with_refine<R2>(self, make_refine: R2) -> Main<G, R2>
+    pub fn with_dns<R2>(self) -> Main<G, R2>
     where
         R2: dns::NewResolver + Send,
         R2::Resolver: Clone + Send + Sync + 'static,
@@ -210,7 +210,6 @@ where
             inbound_listener,
             outbound_listener,
             admin_listener,
-            make_refine,
         } = self;
 
         const MAX_IN_FLIGHT: usize = 10_000;
@@ -247,8 +246,6 @@ where
                 // FIXME: DNS configuration should be infallible.
                 panic!("invalid DNS configuration: {:?}", e);
             });
-
-        let refine = make_refine.make(&dns_resolver);
 
         let (tap_layer, tap_grpc, tap_daemon) = tap::new();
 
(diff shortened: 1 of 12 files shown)


olix0r:
needs fmt

src/dns/mod.rs
```diff
@@ -26,6 +26,26 @@ pub trait ConfigureResolver {
...
+    /// result, instead returning the `Name` that was resolved.
+    ///
+    /// For example, a name like `web` may be refined to `web.example.com.`,
+    /// depending on the DNS search path.
+    fn refine(&self, name: &Name) -> Self::Future;
+}
+
+pub trait MakeRefine: Send + 'static {
+    type Refine: Refine + Clone + Send + Sync + 'static;
+    fn make(self, resolver: &Resolver) -> Self::Refine;
```
olix0r: It's odd that we need to pass a resolver in here? I would expect this to encapsulate creation of the resolver?

src/app/main.rs
```diff
@@ -116,13 +118,41 @@ where
             outbound_listener,
             control_listener,
             admin_listener,
+            make_refine: dns::DefaultRefine,
```
olix0r: If you instantiate the resolver in `new`, you can crate the refine with it, rather than making it a parameter on the trait.

hawkw:
As a note, the intention behind this was to allow backporting a test for service profiles for names outside the destination search suffixes to master. However, it turns out that this test won't work on master regardless, as it also relies on DNS for routing. Instead of this change, we might want to consider making _all_ the DNS resolver functionality mockable, rather than just refinement.

src/dns/mod.rs
hawkw: This is a hack and I don't care for it either. It's a workaround for the issue that the current DNS resolver's `new` returns both a resolver _and_ a background task. We can't just immediately spawn the background task in this function because we want to run it on the control thread, but we need the resolver before we start running the control thread.

Perhaps the right thing would be to do a deeper refactor of how this stuff works, but that requires further thought.

src/app/main.rs
hawkw: This doesn't really work for the reasons I described in https://github.com/linkerd/linkerd2-proxy/pull/234#discussion_r276359677 --- `dns::Resolver::from_system_config_with` returns a background future. Unfortunately, we cannot construct the background future in `new` and store it as a struct field to be spawned later, since it's an `impl Future`, a constraint imposed on us by `trust-dns`.

In fairness, the background future being an `impl Future` _is_ [my fault](https://github.com/bluejekyll/trust-dns/commit/84e6f1df30eda892663553949837f6c7dd788f69), so I suspect Ben would accept a PR changing it to a named type, but I wasn't sure if it was worth it.

olix0r: It would be fine to NewType(Box) it for now to unblock yourself

src/app/main.rs
```diff
@@ -54,9 +54,10 @@ use super::profiles::Client as ProfilesClient;
 ///
 /// The private listener routes requests to service-discovery-aware load-balancer.
 ///
-pub struct Main<G> {
+pub struct Main<G, R = dns::DefaultResolver> {
     proxy_parts: ProxyParts<G>,
     runtime: task::MainRuntime,
+    dns: R
```
kleimkuhler: We have three different names going on for DNS resolvers. Here we're calling it `dns`, but we call it `dns_resolver` and `resolver` in other parts of the proxy. Maybe we could consolidate down one name? Otherwise looks good!

tests/support/proxy.rs
```diff
@@ -62,6 +72,26 @@ impl Proxy {
         self
     }
 
+    pub fn dns<R2>(self, dns: R2) -> Proxy<R2>
```
kleimkuhler: Are we able to do something with implementing `default()` for a `Resolver` so that we don't need separate types for what should be the same thing?

hawkw: Hmm, if I'm understanding what you mean correctly, I don't think so --- we need to have a `NewResolver` trait rather than requiring a bound like `trait Resove: Default` because some resolvers (i.e. the actual dns resolver) also return a background future from their constructors.

PR #235, opened 2019-04-18
Convert from stack to tower-layer

seanmonstar:
This converts from the previously used `linkerd2-stack` to `tower-layer`. This means that *most* things that were `Stack<T>` have been changed to `Service<T>`.

Some specific things changed to support the new pattern:

- The `linkerd2-router` would hold a `Stack<T>` to build services when a new target was recognized. The change in this PR is to create a specific `trait Make<T>` in `linkerd2-router` which can make a service immediately. In order to support this with the various `router::layer`s, a `pending::layer` has been created, which creates an `impl Service<Req>` immediately from a `MakeService<T, Req>`. The `Pending` service returns `NotReady` while the `MakeService::Future` is still pending. Once it completes, it acts just like the underlying service.
- `proxy::canonicalize` was holding onto a `Stack<T>` internally, and would build a new stack whenever DNS resolved to a new address. In to not need to add a buffer + pending layer just to support this, it was made a little more specific. It's moved to `proxy::http::canonicalize`, because it will insert the refined `Addr` into the `http::Request` extensions directly. Previously, since it held a `Stack<T>`, we had a tiny stack that was essentially just `insert_target`.

----

There's some `FIXME(sean)`s in here that still need some thought. And it's still depending on the builder-hax branch, thus: a draft! The discovery and transparency tests all pass for me.

Tracking FIXMEs:

- [ ] `http::profiles` has extra pending to apply the per-profile layer
- [ ] pattern uplift a layer of `Service<Req>` to a layer of `MakeService<_, Req>`


src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![warn(warnings)] // FIXME(sean) don't merge
```
olix0r: ;)

lib/router/src/lib.rs
```diff
@@ -39,22 +39,39 @@ pub trait Recognize<Request> {
     fn recognize(&self, req: &Request) -> Option<Self::Target>;
 }
 
+pub trait Make<Target> {
```
hawkw: might be overly nitpicky but i'm not a huge fan of having a trait that's just called `Make`, esp. in a world where we already have a bunch of types and traits named `MakeFoo`...

lib/stack/src/layer.rs
```diff
@@ -1,95 +1,20 @@
...
-impl<T, M: super::Stack<T>> Layer<T, T, M> for () {
-    type Value = M::Value;
-    type Error = M::Error;
-    type Stack = M;
-
-    fn bind(&self, inner: M) -> M {
-        inner
-    }
+/// Make a `Layer` from a closure.
+pub fn mk<F>(f: F) -> LayerFn<F> {
```
hawkw: this is handy (not sold on the name though)

src/proxy/buffer.rs
```diff
@@ -1,28 +1,36 @@
...
 pub struct Layer<Req> {
     capacity: usize,
     _marker: PhantomData<fn(Req)>,
 }
 
-/// Produces `Service`s wrapped with a `Buffer`
+/// Produces `MakeService`s where the output `Service` is wrapped with a `Buffer`
 #[derive(Debug)]
-pub struct Stack<M, Req> {
+pub struct MakeBuffer<M, Req> {
```
hawkw: This is fairly minor but I notice there are some cases where these are named after what they make (i.e. `MakeBuffer`, `MakePending`, other cases where they're named `Make` or `MakeSvc`, and still others where they're still named `Stack`. Would be nice if this was more consistent.

lib/stack/src/layer.rs
seanmonstar: I dunno, I copied the `mk` function from finagle, where there seems to be `Service.mk`, etc. 🤷‍♂️

src/proxy/buffer.rs
seanmonstar: I swayed back and forth on what they should be called, and eventually stopped caring that some are still called `Stack`. 🤷‍♂️

lib/stack/src/layer.rs
hawkw: fair, i don't actually care & can't think of anything better.

src/proxy/buffer.rs
hawkw: yeah, that's fair. i don't really care what they end up being called (though I'm somewhat in favour of the more information-rich `MakeWhatever`) but they should probably eventually be consistent?

Cargo.toml
```diff
@@ -105,3 +101,8 @@ debug = false
 
 [patch.crates-io]
 webpki = { git = "https://github.com/seanmonstar/webpki", branch = "cert-dns-names" }
+
+[patch."https://github.com/tower-rs/tower"]
```
olix0r: needs a fix before merge -- is this dependent on a tower branch?

seanmonstar:
Closing because travis-ci status is borked?

PR #236, opened 2019-04-22
Replace linkerd2-stack with tower-layer

seanmonstar:
This converts from the previously used `linkerd2-stack` to `tower-layer`. This means that *most* things that were `Stack<T>` have been changed to `Service<T>`.

Some specific things changed to support the new pattern:

- The `linkerd2-router` would hold a `Stack<T>` to build services when a new target was recognized. The change in this PR is to create a specific `trait Make<T>` in `linkerd2-router` which can make a service immediately. In order to support this with the various `router::layer`s, a `pending::layer` has been created, which creates an `impl Service<Req>` immediately from a `MakeService<T, Req>`. The `Pending` service returns `NotReady` while the `MakeService::Future` is still pending. Once it completes, it acts just like the underlying service.
- `proxy::canonicalize` was holding onto a `Stack<T>` internally, and would build a new stack whenever DNS resolved to a new address. In to not need to add a buffer + pending layer just to support this, it was made a little more specific. It's moved to `proxy::http::canonicalize`, because it will insert the refined `Addr` into the `http::Request` extensions directly. Previously, since it held a `Stack<T>`, we had a tiny stack that was essentially just `insert_target`.


olix0r:
now:
```
    Finished release [optimized] target(s) in 6m 23s
CARGO_RELEASE=1 make clean build  1082.30s user 33.68s system 288% cpu 6:26.26 total
```

master:
```
    Finished release [optimized] target(s) in 5m 28s
CARGO_RELEASE=1 make clean build  1013.46s user 29.43s system 316% cpu 5:29.04 total
```

It's interesting that this seems to have a substantial impact on build time. I wonder why?

seanmonstar:
Oh that's interesting, I timed now and master but not in `--release` mode, and it was only a couple seconds difference, which I just assumed was noise.

olix0r:
Seems about 15% hit in release mode, and about 25% in dev mode, if my math is right:

tower-layer:
```
    Finished dev [unoptimized] target(s) in 3m 40s
make clean build  396.19s user 25.32s system 190% cpu 3:41.68 total
```

master:
```
    Finished dev [unoptimized] target(s) in 2m 45s
make clean build  340.30s user 29.89s system 213% cpu 2:53.30 total
```

src/proxy/pending.rs
```diff
@@ -0,0 +1,73 @@
...
+type Error = Box<dyn std::error::Error + Send + Sync>;
+
+#[derive(Clone, Debug)]
+pub struct MakePending<M> {
+    inner: M,
+}
+
+/// Creates a `Service` immediately, even while the future making the service
+/// is still pending.
+pub enum Pending<F, S> {
```
olix0r: TIOLI: it seems like "pending" is a little ambiguous  --- one might assume it has something to do with tracking pending requests, for instance, especially since it's typically paired with a buffer... Would this be more accurately named `make_service_as_service` or something? That names probably no better. I think this is what finagle calls FactoryToService, fwiw...

seanmonstar: `FactoryToService` creates a `Service[Req, Rep]` where every `Req` will acquire a new service from the factory. I believe this is more akin to [`Service.pending`](http://twitter.github.io/finagle/docs/com/twitter/finagle/Service$.html#pending[Req,Rep](svc:com.twitter.util.Future[com.twitter.finagle.Service[Req,Rep]]):com.twitter.finagle.Service[Req,Rep]).

olix0r:
It looks like a few tests now hang... I assume these tests are flakey, but I'd be curious if this breakage is new
```
:; cargo test --test tap -- inbound_http1
    Finished dev [unoptimized] target(s) in 0.32s
     Running target/debug/deps/tap-74153255a2dbc7e3

running 1 test
test inbound_http1 ... test inbound_http1 has been running for over 60 seconds
```

olix0r:
That test passes without issue on master:
```
running 1 test
test inbound_http1 ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out
```

seanmonstar:
Hm, they're passing for me. Not sure what could be causing that.

olix0r:
Okay, `make test` succeeds; so I'm happy to dismiss the test failures to flakiness.

PR #237, opened 2019-04-23
Replace fixed reconnect backoff with exponential one

zaharidichev:
This replaces the `Backoff::Fixed` with `Backoff::Exponential(min, max, max_jitter)` and 
fixes https://github.com/linkerd/linkerd2/issues/2512

Hope I am not stepping on someone's toes


olix0r:
Thanks @zaharidichev! This looks like a great change at a first glance!

src/app/config.rs
```diff
@@ -40,11 +40,21 @@ pub struct Config {
...
+
+    pub inbound_connect_exp_backofff_max_jitter: f64,
+
+    /// Settings for the exponential back-off used to determine the time the
+    /// amount of time to wait between connection attempts.
+    pub outbound_connect_exp_backoff_min: Duration,
+
+    pub outbound_connect_exp_backoff_max: Duration,
+
+    pub outbound_connect_exp_backofff_max_jitter: f64,
```
olix0r: Take it or leave it: we're slowly in the process of "un-flattening" this large config structure... What do you think about defining a `struct ExpBackoff` in this module, and then we can simply store `outbound_connect_backoff: ExpBackoff` in the config struct.

In fact... what do you think about making `reconnect`'s `Backoff` struct public? Then we could just store that structure here; and the reconnect module could be changed to have a `with_backoff(Backoff)`...

src/proxy/reconnect.rs
```diff
@@ -46,7 +49,7 @@ where
 #[derive(Clone, Debug)]
 enum Backoff {
     None,
-    Fixed(Duration),
+    Exponential(Duration, Duration, f64),
```
olix0r: I think it would be a little clearer to use named params here, especially if we make this struct `pub`, e.g.:
```rust
pub enum Backoff {
    None,
    Exponential { min: Duration, max: Duration, jitter: f64 },
}
```

src/app/main.rs
```diff
@@ -276,7 +276,11 @@ where
                     .layer(http_metrics::layer::<_, classify::Response>(
                         ctl_http_metrics.clone(),
                     ))
-                    .layer(reconnect::layer().with_fixed_backoff(config.control_backoff_delay))
+                    .layer(reconnect::layer().with_exp_backoff(
+                        config.control_exp_backoff_min,
+                        config.control_exp_backoff_max,
+                        config.control_exp_backoff_max_jitter,
+                    ))
```
olix0r: and if we follow the above recommendation, all of these turn into the less-verbose 
```rust
reconnect::layer().with_backoff(config.control_backoff.clone())
```

src/proxy/reconnect.rs
```diff
@@ -169,9 +174,21 @@ where
...
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        fn apply_jitter(duration: Duration, max_jitter: f64) -> Duration {
+            let jitter = rand::random::<f64>() * max_jitter;
+            let secs = (duration.as_secs() as f64) * jitter;
+            let nanos = (duration.subsec_nanos() as f64) * jitter;
+            let millis = (secs * 1000f64) + (nanos / 1000000f64);
+            duration + Duration::from_millis(millis as u64)
+        }
+
+        fn next_backoff(failed_attempts: u32, min: &Duration, max: &Duration) -> Duration {
```
olix0r: What do you think about moving the logic from `next_backoff` & `apply_jitter` into a method on `Backoff`?

For example, something like the following:

```rust
impl Backoff {
  fn for_failures<R: rand::Rng>(&self, failures: u32, rng: R) -> Duration {
    match self {
      Backoff::None => Duration::from_secs(0),
      Backoff::Exponential { max, min, jitter } => { ... }
    }
  }
}
```

Then, we can write unit tests on the `Backoff` type independently of the reconnect logic!

src/app/config.rs
```diff
@@ -40,11 +40,21 @@ pub struct Config {
     /// The maximum amount of time to wait for a connection to a remote peer.
     pub outbound_connect_timeout: Duration,
 
-    /// The amount of time to wait between connection attempts.
-    pub inbound_connect_backoff: Duration,
+    /// Settings for the exponential back-off used to determine the time the
```
kleimkuhler:
```suggestion
    /// Settings for the exponential back-off used to determine the amount of time
    /// to wait between connection attempts.
```

src/app/config.rs
```diff
@@ -40,11 +40,21 @@ pub struct Config {
...
+    /// amount of time to wait between connection attempts.
+    pub inbound_connect_exp_backoff_min: Duration,
 
-    /// The amount of time to wait between connection attempts.
-    pub outbound_connect_backoff: Duration,
+    pub inbound_connect_exp_backoff_max: Duration,
+
+    pub inbound_connect_exp_backofff_max_jitter: f64,
```
kleimkuhler: TIOLI: How about we call this `inbound_connect_exp_backoff_jitter`? Since we don't have a min/max here, the "max" may be unnecessary.

src/proxy/reconnect.rs
```diff
@@ -169,9 +174,21 @@ where
     type Future = <Reconnect<M, T> as svc::Service<Req>>::Future;
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        fn apply_jitter(duration: Duration, max_jitter: f64) -> Duration {
+            let jitter = rand::random::<f64>() * max_jitter;
+            let secs = (duration.as_secs() as f64) * jitter;
+            let nanos = (duration.subsec_nanos() as f64) * jitter;
+            let millis = (secs * 1000f64) + (nanos / 1000000f64);
```
kleimkuhler: Consider using [mul_add](https://doc.rust-lang.org/std/primitive.f64.html#method.mul_add) here especially since this is something that could be executed quite frequently!

zaharidichev:
Thanks for the comphensive reviews, very educational.

src/proxy/reconnect.rs
```diff
@@ -44,9 +47,34 @@ where
...
+    fn for_failures<R: rand::Rng>(&self, failures: u32, mut rng: R) -> Option<Duration> {
+        match self {
+            Backoff::None => None,
+            Backoff::Exponential { max, min, jitter } => {
+                let backoff = Duration::min(min.mul(2_u32.pow(failures)), *max);
+
+                if *jitter != 0.0 {
+                    Some(backoff)
+                } else {
+                    let rand_jitter = rng.gen::<f64>() * jitter;
```
seanmonstar: A random f64 could be negative, right? I *think* that'd be a problem, probably when casting the millis back to a u64.

src/app/config.rs
```diff
@@ -705,6 +727,38 @@ fn parse_dns_suffix(s: &str) -> Result<dns::Suffix, ParseError> {
...
+
+    if disabled {
+        Ok(Backoff::None)
+    } else {
+        let min_env = format!("LINKERD2_PROXY_{}_EXP_BACKOFF_MIN", base);
+        let min = parse(strings, &min_env, parse_duration);
+        let max_env = format!("LINKERD2_PROXY_{}_EXP_BACKOFF_MAX", base);
+        let max = parse(strings, &max_env, parse_duration);
+        let jitter_env = format!("LINKERD2_PROXY_{}_EXP_BACKOFF_JITTER", base);
+        let jitter = parse(strings, &jitter_env, parse_number::<f64>);
```
seanmonstar: While it'd be a bad idea, this could parse a negative jitter. Perhaps this should require positive jitter?

src/app/config.rs
```diff
@@ -705,6 +727,38 @@ fn parse_dns_suffix(s: &str) -> Result<dns::Suffix, ParseError> {
...
+
+        match (min?, max?, jitter?) {
+            (None, None, None) => Ok(default),
+            (Some(min), Some(max), Some(jitter)) => Ok(Backoff::Exponential { min, max, jitter }),
+            _ => {
+                error!("You need to specify either all of {} {} {} or non of them to use the default backoff", min_env, max_env,jitter_env );
```
seanmonstar: Typo "non" missing an "e".

src/proxy/reconnect.rs
seanmonstar: Maybe a `quickcheck!` for `Backoff::for_failures` would trigger something bad?

zaharidichev: While I also expected that to be the case, I found out the [docs](https://github.com/rust-random/rand/blob/master/src/lib.rs#L32) pointing out that this actually should give you a random float between 0.0 and 1.0. Will run it through quickcheck though...

seanmonstar: Oh my, I should have looked, but that's not obvious at all. I suppose adding a `debug_assert` would help document this behavior for us.

src/app/config.rs
```diff
@@ -190,9 +194,8 @@ pub const ENV_ADMIN_LISTEN_ADDR: &str = "LINKERD2_PROXY_ADMIN_LISTEN_ADDR";
 pub const ENV_METRICS_RETAIN_IDLE: &str = "LINKERD2_PROXY_METRICS_RETAIN_IDLE";
 const ENV_INBOUND_CONNECT_TIMEOUT: &str = "LINKERD2_PROXY_INBOUND_CONNECT_TIMEOUT";
 const ENV_OUTBOUND_CONNECT_TIMEOUT: &str = "LINKERD2_PROXY_OUTBOUND_CONNECT_TIMEOUT";
-const ENV_INBOUND_CONNECT_BACKOFF: &str = "LINKERD2_PROXY_INBOUND_CONNECT_BACKOFF";
-const ENV_OUTBOUND_CONNECT_BACKOFF: &str = "LINKERD2_PROXY_OUTBOUND_CONNECT_BACKOFF";
-
+const ENV_INBOUND_CONNECT_BASE: &str = "INBOUND_CONNECT";
```
kleimkuhler: Because this is not an environment variable in the sense that the other variables are in this file, I think we should call this something different -- perhaps just `INBOUND_CONNECT_BASE`.

This constant is used as part of generating the configuration as a whole, but it is not something that is dependent on the environment.

This, along with outbound connect and control base could be put at the top of the file like other constants are throughout the project ([e.g.](https://github.com/linkerd/linkerd2-proxy/blob/5f58883a7e34bc449f0b0e8c6f997018c5eda32a/src/app/mod.rs#L21-L22))

Let me know if that makes sense or if you think otherwise!

PR #238, opened 2019-04-26
tap: remove need to clone Services

seanmonstar:
This refactors the tap system to not require intermediary channels to
register matches and taps when a request comes through. The Dispatcher
that used to exist in order to prevent tapping more requests than the
limit asked for has been removed. In its place is a shared atomic
counter to keep the count under the limit.

The resulting behavior should be the same. There should be improved
performance as tap registration doesn't need go through a second
channel, and requests don't need to be delayed waiting for the
dispatcher to be able to process its queue.

----

I noticed this because of the need for `S: Service + Clone` on the tap layer, which I'm trying to remove for another reason. However, this seems orthogonal enough to be a separate PR.

### DIFF
--- a/src/tap/grpc/server.rs
+++ b/src/tap/grpc/server.rs
@@ -1,8 +1,7 @@
 use bytes::Buf;
-use futures::sync::{mpsc, oneshot};
+use futures::sync::mpsc;
 use futures::{future, Async, Future, Poll, Stream};
 use hyper::body::Payload;
-use never::Never;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, Weak};
 use std::time::Instant;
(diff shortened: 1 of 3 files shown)


olix0r:
Seems likely that this wasn't related to this change, but i see some test hangs when i don't filter flakey... (iirc inbound_http1 already had this behavior on my linux host):

```
     Running target/debug/deps/tap-1b40e6a274444a4b

running 3 tests
test support::assert_eventually ... ok
test inbound_http1 ... test inbound_http1 has been running for over 60 seconds
test grpc_headers_end ... test grpc_headers_end has been running for over 60 seconds```

seanmonstar:
Those same tests pass for me. I really need to find a way to detect the taps have been registered to unflake those tests...

PR #239, opened 2019-04-30
Remove buffers from endpoint stacks

seanmonstar:
Due to the `http::settings::router`, a `buffer` was needed in each
endpoint stack. This meant that the service was always ready, even if
the client were falling over (and reconnecting). In turn, this meant
that the balancer would pick one of these endpoint stacks, because it
was always ready!

This change includes a test of a failing endpoint, that the balancer no
longer assumes it is ready, and has the following functional changes:

- Removed `http::settings::router`, instead the client HTTP settings are
  detected as part of the `DstAddr`. This means that each balancer only
  has endpoints with the same HTTP settings.
- Removed `buffer` layer from inside the endpoint stacks.

Closes https://github.com/linkerd/linkerd2/issues/2550

### DIFF
--- a/src/app/inbound.rs
+++ b/src/app/inbound.rs
@@ -7,7 +7,7 @@ use std::sync::Arc;
 use super::classify;
 use super::dst::DstAddr;
 use super::identity;
-use proxy::http::router;
+use proxy::http::{router, settings};
 use proxy::server::Source;
 use tap;
 use transport::{connect, tls};
@@ -17,6 +17,7 @@ use {Conditional, NameAddr};
 pub struct Endpoint {
     pub addr: SocketAddr,
     pub dst_name: Option<NameAddr>,
+    pub http_settings: settings::Settings,
     pub tls_client_id: tls::PeerIdentity,
 }
 
@@ -32,6 +33,7 @@ impl From<SocketAddr> for Endpoint {
         Self {
             addr,
             dst_name: None,
+            http_settings: settings::Settings::NotHttp,
             tls_client_id: Conditional::None(tls::ReasonForNoPeerName::NotHttp.into()),
         }
     }
@@ -46,6 +48,12 @@ impl connect::HasPeerAddr for Endpoint {
 impl tls::HasPeerIdentity for Endpoint {
     fn peer_identity(&self) -> tls::PeerIdentity {
         Conditional::None(tls::ReasonForNoPeerName::Loopback.into())
+    }
+}
+
+impl settings::HasSettings for Endpoint {
+    fn http_settings(&self) -> &settings::Settings {
+        &self.http_settings
     }
 }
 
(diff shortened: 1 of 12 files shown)


src/app/inbound.rs
```diff
@@ -126,16 +134,20 @@ impl<A> router::Recognize<http::Request<A>> for RecognizeEndpoint {
...
             .extensions()
             .get::<DstAddr>()
-            .and_then(|a| a.as_ref().name_addr())
-            .cloned();
+            .expect("request extensions should have DstAddr");
+
+        let dst_name = dst_addr.as_ref().name_addr().cloned();
+        let http_settings = dst_addr.http_settings;
+
         debug!("inbound endpoint: dst={:?}", dst_name);
```
olix0r: I assume `Settings` is `Debug` -- if so, worth including in this log?

src/proxy/reconnect.rs
```diff
@@ -169,20 +169,15 @@ where
...
-                    }
+        if let Some(delay) = self.active_backoff.as_mut() {
+            match delay.poll() {
+                Ok(Async::NotReady) => return Ok(Async::NotReady),
+                Ok(Async::Ready(())) => {}
+                Err(e) => {
+                    error!("timer failed; continuing without backoff: {}", e);
                 }
             }
-        };
```
olix0r: just to confirm... this was unrelated cleanup?

src/app/inbound.rs
```diff
@@ -32,6 +33,7 @@ impl From<SocketAddr> for Endpoint {
         Self {
             addr,
             dst_name: None,
+            http_settings: settings::Settings::Http2,
```
olix0r: Would it make more sense to have a distinct Settings variant for "NotHttp"? I think it might be less surprising next time someone comes through this -- it's subtle when/how this is used...  And we could fail to build clients when this settings variant is used.

tests/discovery.rs
```diff
@@ -661,6 +661,56 @@ mod http2 {
 
     generate_tests! { server: server::new, client: client::new }
 
+
+    #[test]
+    fn outbound_blah_blah_blah() {
```
seanmonstar: Oh yea, I should get a better name XD

src/proxy/reconnect.rs
seanmonstar: Yea, just something I noticed when I was adding a debug log below...

seanmonstar:
Review addressed, rebased onto master.

src/proxy/http/normalize_uri.rs
```diff
@@ -23,6 +22,10 @@ pub struct Service<S> {
     inner: S,
 }
 
+fn should_normalize_uri(settings: &Settings) -> bool {
```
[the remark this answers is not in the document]
olix0r: I thought that first too, but Settings already has public methods that express the settings -- the logic of "should i normalize" is properly locally to this module, imo.

olix0r:
@hawkw 

> The one potential concern I have is how this would affect #240 --- since that branch adds a fallback router to the lb layer, and IIRC the router requires the underlying service to be Clone, I think the buffer may be the only thing that currently makes that service Clone. Since having a buffer under the lb is problematic, we may have to rethink this a bit?

We'll probably need to use a buffer when the router is used, but not when the balancer is used.

hawkw:
@olix0r yup, see also 767b8c69d0e50b8a4e14806d3bf3ede66694d6ad :)

PR #240, opened 2019-04-30
Route requests to original dsts when no endpoints exist in the lb

hawkw:
Currently, when no endpoints exist in the load balancer for a
destination, we fail the request. This is because we expect endpoints to
be discovered by both destination service queries _and_ DNS lookups, so
if there are no endpoints for a destination, it is assumed to not exist.

In linkerd/linkerd2#2661, we intend to remove the DNS lookup from the
proxy and instead fall back to routing requests for which no endpoints
exist in the destination service to their SO_ORIGINAL_DST IP address.
This means that the current approach of failing requests when the load
balancer has no endpoints will no longer work.

This branch introduces a new `balance::fallback::layer`, which marries a
load balancer with a router. When the load balancer has no endpoints, we
fall back to the router, rather than failing the request at the
balancer. We can use this router to recognize endpoints based on the
request's original destination address.

A new trait, `HasEndpointStatus`, is added in order to allow the
discovery lookup to communicate the "no endpoints" state to the
balancer. In addition, we add a new `Update::NoEndpoints` variant to
`proxy::resolve::Update`, so that when the control plane sends a no
endpoints update, we switch from the balancer to the fallback router
_immediately_, rather than waiting for all the endpoints to be
individually removed.

A future commit will remove the DNS lookups from the discovery module.


hawkw:
I've marked this PR as a draft as I'd still like to add integration tests for this change if possible. However, I thought @olix0r and @kleimkuhler would like to have a look at it sooner rather than later.

src/app/main.rs
```diff
@@ -663,7 +692,7 @@ where
             // implementations can use the route-specific configuration.
             let dst_route_stack = svc::builder()
                 .layer(buffer::layer(max_in_flight))
-                .layer(pending::layer())
+                .layer(pending::layer::<_, dst::Route, _>())
```
hawkw: This type annotation is necessary as a result of changing the `pending::layer` function to bound the target type.

hawkw:
@olix0r and @kleimkuhler, I've made the changes we discussed offline earlier today, and rebased this onto the latest master.

I've marked this PR as "ready for review", as it won't really be possible to write new integration tests for this until we make the second part of the change and remove the DNS lookups from discovery as wekk,

src/app/main.rs
```diff
@@ -474,14 +476,32 @@ where
...
                 .layer(insert_target::layer());
 
+            let balancer = svc::builder()
+                .layer(
+                    balance::layer(
+                        EWMA_DEFAULT_RTT,
+                        EWMA_DECAY,
+                        resolve::layer(Resolve::new(resolver)),
+                    )
+                    .with_fallback(max_in_flight, |req: &http::Request<_>| {
```
hawkw: having to pass the in-flight limit here is a little unfortunate, but it's necessary as the buffer is only applied to the stack under the router (and not to the stack under the lb)

src/proxy/http/balance.rs
```diff
@@ -118,19 +152,426 @@ where
...
+                <buffer::Layer<http::Request<A>> as svc::stack::Layer<M>>::Service,
+            >>::Service,
+            Bal::Service,
+            A,
+        >;
+
+        fn layer(&self, inner: M) -> Self::Service {
+            let balance = self.balance_layer.layer(inner.clone());
+            let inner = buffer::layer(self.max_in_flight).layer(inner);
+            let fallback = router::layer(self.recognize.clone()).layer(inner);
```
hawkw: Initially, I tried to use `ServiceBuilder` here but couldn't figure out the bounds to make that work.

src/app/main.rs
olix0r: The buffer is only necessary so that we have a cloneable sender, not to actually hold requests in a queue... Can we Use a Buffer with a capacity of 1 to force backpressure? when the route is in use?

hawkw: I double-checked the `tower-buffer` behaviour and that should work. Will do.

src/control/destination/background/destination_set.rs
```diff
@@ -230,6 +230,10 @@ where
         );
         match self.addrs.take() {
             Exists::Yes(mut cache) => {
+                self.responders.retain(|r| {
+                    let sent = r.update_tx.unbounded_send(Update::NoEndpoints);
```
kleimkuhler: This will send the new `Update::NoEndpoints` state that we poll for in resolve now, but why does it need to be `retain`? Shouldn't all the responders be ready to receive?

src/app/main.rs
olix0r: As we discussed in person, we decided that pushing backpressure into the executor is not how we want to handle this, so instead we should be willing to buffer here.

hawkw: yup, 00d0c47433b585f32f4109d18814885d4a1f60ac reverted the change that added a capacity 1 buffer.

src/control/destination/background/destination_set.rs
hawkw: the `retain` will drop any responders for which the receive end has been dropped. `on_change` also uses `retain`: https://github.com/linkerd/linkerd2-proxy/blob/767b8c69d0e50b8a4e14806d3bf3ede66694d6ad/src/control/destination/background/destination_set.rs#L267-L271

src/proxy/http/balance.rs
```diff
@@ -188,3 +650,164 @@ pub mod weight {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use proxy::{
+        http::router::rt,
+        pending,
+        resolve::{self, Resolution, Resolve},
+    };
+
+    fn assert_make<A, T: Clone>(_: A)
```
kleimkuhler: Are these from the original bounds debugging? Looks like you can remove some of this if these tests still apply.

src/proxy/http/balance.rs
```diff
@@ -118,19 +152,447 @@ where
...
+            >,
+            future::Map<Bal::Future, fn(Bal::Response) -> Self::Response>,
+        >;
+
+        fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+            let ready = self.balance.poll_ready()?;
+            if !self.status.is_empty() {
+                // destroy the fallback router
+                self.fallback.destroy();
+            }
```
olix0r: don't we technically need to call poll_ready on the fallback here if status.is_empty?

olix0r: And, in fact, won't this prevent us from ever dispatching requests to the fallback while balance is not ready?

src/proxy/http/balance.rs
hawkw: yes, but i thought it was better to leave the tests to guard against problems in the future?

kleimkuhler: Ah my fault the test is on the actual test signatures. I was seeing these as empty tests. All good -- definitely good to keep these.

src/proxy/http/balance.rs
hawkw: @olix0r whoops, you're right --- fixed in 19cda744071c21b91decfe51e16c00d186df2559

src/proxy/http/balance.rs
```diff
@@ -36,43 +41,71 @@ pub struct MakeSvc<M, A, B> {
...
         }
     }
 }
 
+impl<A, B, D> Layer<A, B, D> {
+    pub fn with_fallback<Rec>(
+        self,
+        max_in_flight: usize,
+        recognize: Rec,
+    ) -> fallback::Layer<Rec, Self, A>
```
olix0r: just thinking out loud -- I wonder if this could just take another layer type, so the router/buffering/etc logic need not be fixed in this file (and could potentially support arbitrary other layers on the fallback route)

src/proxy/http/balance.rs
```diff
@@ -118,19 +152,471 @@ where
...
+
+impl fmt::Display for NoEndpoints {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        fmt::Display::fmt("load balancer has no endpoints", f)
+    }
+}
+
+impl ::std::error::Error for NoEndpoints {}
+
+pub mod fallback {
```
olix0r: this module doubles the size of this file... can it be split into its own file? maybe `balance` needs to become a directory...

hawkw: yeah, I wasn't sure --- will split it out!

src/proxy/http/balance/fallback.rs
```diff
@@ -0,0 +1,409 @@
...
+            }
+        }
+    }
+
+    fn call(&mut self, req: http::Request<A>) -> Self::Future {
+        loop {
+            if let Some(ref mut router) = self.router {
+                return svc::Service::call(router, req);
+            } else {
+                self.create();
```
olix0r: These seems problematic, as we'll loop into call without having invoked poll_ready. In fact, this seems unreachable.

hawkw: the intention was to guard against a situation where the no endpoints is received between `poll_ready` and `call`. I'm not sure if that's actually a concern. but i think you're right that this is incorrect --- the state transition should only ever occur in poll_ready. 

in the current situation I don't think this is a problem since routers are always ready? but if that changes this would be a problem

PR #241, opened 2019-05-02
router: Fail requests when the route is not ready

olix0r:
In linkerd/linkerd2#2779, we plan to expire requests while they are
buffered. However, the router _implicitly_ buffers requests in the
executor when the inner service is not ready.

This change alters the router's respnse future to, instead, fail
requests when the inner service is not ready, much like the load
shedding layer.

### DIFF
--- a/lib/router/src/lib.rs
+++ b/lib/router/src/lib.rs
@@ -61,7 +61,7 @@ where
     Svc: svc::Service<Req>,
     Svc::Error: Into<error::Error>,
 {
-    state: State<Req, Svc>,
+    state: State<Req, LoadShed<Svc>>,
 }
 
 struct Inner<Req, Rec, Mk>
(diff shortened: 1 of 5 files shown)


seanmonstar:
Not sure what I'm about to say is better, but: this could also be achieved by putting a `load_shed` layer under the router, right?

olix0r:
@seanmonstar yeah, i considered that and i think it's totally equivalent. i thought it seemed like a bigger foot-gun that the router can do this buffering implicitly, though.

lib/router/src/lib.rs
```diff
@@ -71,17 +72,17 @@ where
...
 enum State<Req, Svc>
 where
     Svc: svc::Service<Req>,
+    Svc::Error: Into<error::Error>,
 {
-    NotReady(Req, Svc),
-    Called(Svc::Future),
-    Error(error::Error),
-    Tmp,
+    Init(Option<Req>, LoadShed<Svc>),
```
hawkw: TIOLI: Does `State` need to know about the `LoadShed` middleware? It seems like it could've been left the same and `ResponseFuture` just be changed to have
```rust
state: State<Req, LoadShed<Svc>>,
```

lib/router/src/lib.rs
```diff
@@ -34,8 +34,8 @@ where
 pub trait Recognize<Request> {
     /// Identifies a Route.
     type Target: Eq + Hash;
-
     /// Determines the target for a route to handle the given request.
+
```
kleimkuhler: Extra newline here

PR #243, opened 2019-05-03
Assert that TLS connection is refused if identity is not certified yet

zaharidichev:
This branch adds tls capability to the support cient used in tests. In addition to that it adds two tests verifying that a TLS connection is refused in case the identity is not certified yet. This attempts to fix #https://github.com/linkerd/linkerd2/issues/2598 and provide facility to write tests for https://github.com/linkerd/linkerd2/issues/2676. 

As these are still some of my first lines of Rust code, it is advised to approach everything with a healthy dose of doubt :)

### DIFF
--- a/tests/identity.rs
+++ b/tests/identity.rs
@@ -48,66 +48,93 @@ fn nonblocking_identity_detection() {
     assert_eq!(active.read_timeout(Duration::from_secs(2)), msg2.as_bytes());
 }
 
-#[test]
+macro_rules! generate_tls_accept_test {
+    ( client_non_tls: $make_client_non_tls:path, client_tls: $make_client_tls:path) => {
+        let _ = env_logger_init();
+        let id = "foo.ns1.serviceaccount.identity.linkerd.cluster.local";
+        let id_svc = identity::Identity::new("foo-ns1", id.to_string());
+        let proxy = proxy::new()
+            .identity(id_svc.service().run())
+            .run_with_test_env(id_svc.env);
 
-fn accepts_tls_after_identity_is_certified() {
-    let _ = env_logger_init();
-    let id = "foo.ns1.serviceaccount.identity.linkerd.cluster.local";
-    let id_svc = identity::Identity::new("foo-ns1", id.to_string());
-    let proxy = proxy::new()
-        .identity(id_svc.service().run())
-        .run_with_test_env(id_svc.env);
+        let non_tls_client = $make_client_non_tls(proxy.metrics, "localhost");
+        assert_eventually!(
+            non_tls_client
+                .request(non_tls_client.request_builder("/ready").method("GET"))
+                .status()
+                == http::StatusCode::OK
+        );
 
-    let non_tls_client = client::http1(proxy.metrics, "localhost");
-    assert_eventually!(
-        non_tls_client
-            .request(non_tls_client.request_builder("/ready").method("GET"))
-            .status()
-            == http::StatusCode::OK
-    );
+        let tls_client = $make_client_tls(
+            proxy.metrics,
+            "localhost",
+            client::TlsConfig::new(id_svc.client_config, id),
+        );
+        assert_eventually!(
+            tls_client
+                .request(tls_client.request_builder("/ready").method("GET"))
+                .status()
+                == http::StatusCode::OK
+        );
+    };
+}
 
-    let tls_client = client::http1_tls(
-        proxy.metrics,
-        "localhost",
-        client::TlsConfig::new(id_svc.client_config, id),
-    );
-    assert_eventually!(
-        tls_client
-            .request(tls_client.request_builder("/ready").method("GET"))
-            .status()
-            == http::StatusCode::OK
-    );
+macro_rules! generate_tls_reject_test {
+    ( client: $make_client:path) => {
+        let _ = env_logger_init();
+        let id = "foo.ns1.serviceaccount.identity.linkerd.cluster.local";
+        let identity::Identity {
+            env,
+            mut certify_rsp,
+            client_config,
+        } = identity::Identity::new("foo-ns1", id.to_string());
+
+        certify_rsp.valid_until = Some((SystemTime::now() + Duration::from_secs(666)).into());
+
+        let (_tx, rx) = oneshot::channel();
+        let id_svc = controller::identity().certify_async(move |_| rx).run();
+
+        let proxy = proxy::new().identity(id_svc).run_with_test_env(env);
+
+        let client = $make_client(
+            proxy.metrics,
+            "localhost",
+            client::TlsConfig::new(client_config, id),
+        );
+
+        assert!(client
+            .request_async(client.request_builder("/ready").method("GET"))
+            .wait()
+            .err()
+            .unwrap()
+            .is_connect());
+    };
 }
 
 #[test]
-fn rejects_tls_before_identity_is_certified() {
-    let _ = env_logger_init();
-    let id = "foo.ns1.serviceaccount.identity.linkerd.cluster.local";
-    let identity::Identity {
-        env,
-        mut certify_rsp,
-        client_config,
-    } = identity::Identity::new("foo-ns1", id.to_string());
+fn http1_accepts_tls_after_identity_is_certified() {
+    generate_tls_accept_test! {
+       client_non_tls:  client::http1,
+       client_tls: client::http1_tls
+    }
+}
 
-    certify_rsp.valid_until = Some((SystemTime::now() + Duration::from_secs(666)).into());
+#[test]
+fn http1_rejects_tls_before_identity_is_certified() {
+    generate_tls_reject_test! {client: client::http1_tls}
+}
 
-    let (_tx, rx) = oneshot::channel();
-    let id_svc = controller::identity().certify_async(move |_| rx).run();
+#[test]
+fn http2_accepts_tls_after_identity_is_certified() {
+    generate_tls_accept_test! {
+       client_non_tls:  client::http2,
+       client_tls: client::http2_tls
+    }
+}
 
-    let proxy = proxy::new().identity(id_svc).run_with_test_env(env);
-
-    let client = client::http1_tls(
-        proxy.metrics,
-        "localhost",
-        client::TlsConfig::new(client_config, id),
-    );
-
-    assert!(client
-        .request_async(client.request_builder("/ready").method("GET"))
-        .wait()
-        .err()
-        .unwrap()
-        .is_connect());
+#[test]
+fn http2_rejects_tls_before_identity_is_certified() {
+    generate_tls_reject_test! {client: client::http2_tls}
 }
 
 #[test]
--- a/tests/identity.rs
+++ b/tests/identity.rs
@@ -20,6 +20,7 @@ fn nonblocking_identity_detection() {
     let identity::Identity {
         env,
         mut certify_rsp,
+        client_config: _,
     } = identity::Identity::new("foo-ns1", id.to_string());
     certify_rsp.valid_until = Some((SystemTime::now() + Duration::from_secs(666)).into());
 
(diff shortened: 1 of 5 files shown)


tests/support/client.rs
```diff
@@ -8,43 +8,57 @@ use self::tokio::{
...
 pub struct BytesBody(hyper::Body);
 
-pub fn new<T: Into<String>>(addr: SocketAddr, auth: T) -> Client {
-    http2(addr, auth.into())
+pub fn new<T: Into<String>>(addr: SocketAddr, auth: T, tls: Option<TlsConfig>) -> Client {
+    http2(addr, auth.into(), tls)
 }
 
-pub fn http1<T: Into<String>>(addr: SocketAddr, auth: T) -> Client {
+pub fn http1<T: Into<String>>(addr: SocketAddr, auth: T, tls: Option<TlsConfig>) -> Client {
```
kleimkuhler: Let's consider making a new function that takes a `tls` parameter instead of an `Option` like we are doing here. It will make this change significantly smaller since we will not need to update all the calls to this function in the rest of the tests.

I'm think something like `pub fn tls_http1<T: Into<String>>(..., tls_config: TlsConfig) -> Client {}`.

The current `http1`/`http2` functions can just pass `None` into `new` so that we can continue making one `Client`.

tests/support/mod.rs
```diff
@@ -22,9 +22,11 @@ extern crate tokio;
 extern crate tokio_connect;
 extern crate tokio_current_thread;
 pub extern crate tokio_io;
+pub extern crate tokio_rustls;
```
hawkw: I'm not sure if this needs to be `pub` --- it looks like it's only used within the `support` module and need not be exposed to the test modules that depend on `support`.
```suggestion
extern crate tokio_rustls;
```

tests/support/mod.rs
```diff
@@ -22,9 +22,11 @@ extern crate tokio;
...
 extern crate tower_grpc;
 extern crate tower_http_service;
 extern crate tower_service;
+pub extern crate webpki;
```
hawkw: Same as above, I don't think this needs to be `pub`.
```suggestion
extern crate webpki;
```

tests/support/client.rs
hawkw: +1 for @kleimkuhler's comment --- I'd really prefer to not have to pass `None` into all test client constructors in the future.

tests/support/client.rs
```diff
@@ -95,12 +111,12 @@ impl Client {
...
         builder: &mut http::request::Builder,
-    ) -> Box<Future<Item = Response, Error = String> + Send> {
+    ) -> Box<Future<Item = Response, Error = ClientError> + Send> {
         self.send_req(builder.body(Bytes::new()).unwrap())
     }
 
-    pub fn request(&self, builder: &mut http::request::Builder) -> Response {
-        self.request_async(builder).wait().expect("response")
+    pub fn request(&self, builder: &mut http::request::Builder) -> Result<Response, ClientError> {
+        self.request_async(builder).wait()
```
hawkw: I'm not sure if I understand why the `expect` was removed here. Removing it requires changing a bunch of other code (e.g., a lot of the `transparency` tests) that otherwise would not need to be removed. 

If the intention is to assert that some requests will fail (such as when the proxy cannot yet accept TLS connections), can't we just use 
```rust
client.request_async(...).wait()
```
 in those specific cases?

tests/support/client.rs
```diff
@@ -197,45 +231,37 @@ fn run(addr: SocketAddr, version: Run) -> (Sender, Running) {
     (tx, running_rx)
 }
 
-/// The "connector". Clones `running` into new connections, so we can signal
-/// when all connections are finally closed.
-struct Conn {
+struct CustomConnector {
```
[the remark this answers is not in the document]
hawkw: Leaving it `Conn` would be fine, too.

tests/support/client.rs
```diff
@@ -8,43 +8,57 @@ use self::tokio::{
...
 use support::hyper::body::Payload;
+use tokio_rustls::TlsStream;
+use webpki::DNSName;
 
+type ClientError = hyper::Error;
 type Request = http::Request<Bytes>;
 type Response = http::Response<BytesBody>;
-type Sender = mpsc::UnboundedSender<(Request, oneshot::Sender<Result<Response, String>>)>;
+type Sender = mpsc::UnboundedSender<(Request, oneshot::Sender<Result<Response, ClientError>>)>;
+type TlsConfig = (Arc<ClientConfig>, DNSName);
```
hawkw: Take it or leave it, but I'd prefer if this was a `struct` with named fields, rather than a type alias for a tuple. I think the code would be easier to read in that case.

tests/support/client.rs
```diff
@@ -144,8 +175,12 @@ enum Run {
     Http2,
 }
 
-fn run(addr: SocketAddr, version: Run) -> (Sender, Running) {
-    let (tx, rx) = mpsc::unbounded::<(Request, oneshot::Sender<Result<Response, String>>)>();
+fn run(
+    addr: SocketAddr,
+    version: Run,
+    tls: Option<(Arc<ClientConfig>, DNSName)>,
```
hawkw: Nit: If we continue using a type alias for this, I'd prefer to always see `Option<TlsConfig>` rather than `Option<(...)>`
```suggestion
    tls: Option<TlsConfig>,
```

tests/support/client.rs
```diff
@@ -197,45 +231,37 @@ fn run(addr: SocketAddr, version: Run) -> (Sender, Running) {
...
-            .expect("support client cannot connect more than once");
-        let c = TcpStream::connect(&self.addr)
-            .and_then(|tcp| tcp.set_nodelay(true).map(move |_| tcp))
-            .map(move |tcp| RunningIo {
-                inner: tcp,
-                running: running,
-            });
-        Box::new(c)
+impl CustomConnector {
+    fn initialize_connection(
```
hawkw: Not sure why this function was renamed?

tests/support/client.rs
```diff
@@ -245,40 +271,135 @@ impl hyper::client::connect::Connect for Conn {
...
 }
 
-/// A wrapper around a TcpStream, allowing us to signal when the connection
-/// is dropped.
-struct RunningIo {
-    inner: TcpStream,
-    /// When this drops, the related Receiver is notified that the connection
-    /// is closed.
-    running: oneshot::Sender<()>,
+pub enum ConnectorFuture<F: Future> {
```
hawkw: I don't think this needs to be `pub` --- AFAICT it's not exposed outside this module?

tests/support/client.rs
```diff
@@ -245,40 +271,135 @@ impl hyper::client::connect::Connect for Conn {
...
+    Init {
+        future: F,
+        tls: Option<(Arc<ClientConfig>, DNSName)>,
```
hawkw: Again, would prefer to see the type alias used consistently if we're going to continue using one:
```suggestion
        tls: Option<TlsConfig>,
```

tests/support/client.rs
```diff
@@ -245,40 +271,135 @@ impl hyper::client::connect::Connect for Conn {
...
+where
+    F: Future,
+    F::Item: AsyncRead + AsyncWrite + Send + 'static,
+    F::Error: From<io::Error>,
+{
+    type Item = RunningIo<F::Item>;
+    type Error = F::Error;
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        fn take_running(r: &Mutex<Option<oneshot::Sender<()>>>) -> oneshot::Sender<()> {
```
hawkw: Take it or leave it: if this was a closure rather than a function, it could just capture `running`...

tests/support/client.rs
```diff
@@ -245,40 +271,135 @@ impl hyper::client::connect::Connect for Conn {
...
+                    let io = try_ready!(future.poll());
+                    return Ok(Async::Ready(RunningIo::Https(io, take_running(running))));
+                }
+            }
+        }
+    }
 }
 
-impl io::Read for RunningIo {
+pub enum RunningIo<T> {
```
hawkw: Again, not sure if this needs to be `pub`. 

Also, I think the underlying IO will always be a `TcpStream`, so I don't think it really needs to be generic over a `T`...

hawkw:
It looks like the CI build is failing because the client constructor is being called without the new TLS config parameter in `tests/discovery.rs`. 

IMO that's a further argument for trying to make this change without modifying the existing API that the rest of the tests rely on, if possible.

tests/support/identity.rs
```diff
@@ -51,6 +52,20 @@ where
...
+        use std::io::Cursor;
+
+        let mut roots = rustls::RootCertStore::empty();
+        roots
+            .add_pem_file(&mut Cursor::new(s))
+            .expect("add pem file");
+
+        let mut c = rustls::ClientConfig::new();
+        c.root_store = roots;
+        c.enable_tickets = false;
```
kleimkuhler: We shouldn't need to worry about setting this in tests; it should be fine to leave out.

tests/support/client.rs
```diff
@@ -8,12 +8,38 @@ use self::tokio::{
     io::{AsyncRead, AsyncWrite},
     net::TcpStream,
 };
+use self::tokio_rustls::TlsStream;
+use self::webpki::DNSName;
+use self::webpki::DNSNameRef;
```
kleimkuhler: We can consolidate this to:

```diff
diff --git a/tests/support/client.rs b/tests/support/client.rs
index 7a3147b..1b03bf7 100644
--- a/tests/support/client.rs
+++ b/tests/support/client.rs
@@ -9,8 +9,7 @@ use self::tokio::{
     net::TcpStream,
 };
 use self::tokio_rustls::TlsStream;
-use self::webpki::DNSName;
-use self::webpki::DNSNameRef;
+use self::webpki::{DNSName, DNSNameRef};
 use bytes::{Buf, BufMut};
 use rustls::{ClientConfig, ClientSession};
 use std::io::{Read, Write};
```

tests/support/client.rs
```diff
@@ -29,6 +55,18 @@ pub fn http1<T: Into<String>>(addr: SocketAddr, auth: T) -> Client {
         Run::Http1 {
             absolute_uris: false,
         },
+        None,
+    )
+}
+
+pub fn http1_tls<T: Into<String>>(addr: SocketAddr, auth: T, tls: TlsConfig) -> Client {
```
kleimkuhler: Any reason we're only testing TLS over http1? It could be helpful to add a similar `http2_tls` function.

tests/support/client.rs
```diff
@@ -110,23 +151,38 @@ impl Client {
...
-            let absolute = format!("http://{}{}", self.authority, req.uri().path())
-                .parse()
-                .unwrap();
-            *req.uri_mut() = absolute;
+            if self.tls.is_some() {
+                *req.uri_mut() = format!("https://{}{}", self.authority, req.uri().path())
+                    .parse()
+                    .unwrap();
+            } else {
+                *req.uri_mut() = format!("https://{}{}", self.authority, req.uri().path())
```
kleimkuhler: Looks like this should be: `http://{}{}`

tests/support/client.rs
```diff
@@ -249,36 +294,129 @@ impl hyper::client::connect::Connect for Conn {
...
+    }
+}
+
+enum RunningIo {
+    Http(TcpStream, oneshot::Sender<()>),
+    Https(TlsStream<TcpStream, ClientSession>, oneshot::Sender<()>),
+}
+
+impl Read for RunningIo {
+    #[inline]
```
kleimkuhler: Is there a specific reason you added the `#[inline]` attributes? If not, I'm inclined to suggest we leave them out. If we start addressing test performance, there are probably some larger module changes we could make before worrying about hinting inlines to the compiler!

tests/support/client.rs
```diff
@@ -249,36 +294,129 @@ impl hyper::client::connect::Connect for Conn {
...
-        self.inner.flush()
+        match *self {
+            RunningIo::Http(ref mut s, _) => s.flush(),
+            RunningIo::Https(ref mut s, _) => s.flush(),
+        }
     }
 }
 
-impl AsyncRead for RunningIo {}
+impl AsyncRead for RunningIo {
```
kleimkuhler: Unless you had a reason to add this, we should still be okay with `impl AsyncRead for Running IO {}`. Removing this I was still able to run all the tests, but let me know if you were thinking it will be helpful for certain future tests!

tests/support/client.rs
```diff
@@ -249,36 +294,129 @@ impl hyper::client::connect::Connect for Conn {
...
 impl AsyncWrite for RunningIo {
     fn shutdown(&mut self) -> Poll<(), io::Error> {
-        AsyncWrite::shutdown(&mut self.inner)
+        match *self {
+            RunningIo::Http(ref mut s, _) => s.shutdown(),
+            RunningIo::Https(ref mut s, _) => s.shutdown(),
+        }
+    }
+
+    fn write_buf<B: Buf>(&mut self, buf: &mut B) -> Poll<usize, io::Error> {
```
kleimkuhler: I have the same comment for this method about `AsyncRead`. We should be fine not adding this, but let me know if there were future intentions for it.

tests/support/client.rs
zaharidichev: True, it will also be needed for the tap client as as its http2

tests/identity.rs
```diff
@@ -47,13 +48,103 @@ fn nonblocking_identity_detection() {
     assert_eq!(active.read_timeout(Duration::from_secs(2)), msg2.as_bytes());
 }
 
+macro_rules! generate_tls_accept_test {
```
hawkw: +1 for generating multiple tests with macros!

tests/support/identity.rs
```diff
@@ -84,10 +98,17 @@ impl Identity {
 
         env.put(app::config::ENV_IDENTITY_DIR, id_dir);
         env.put(app::config::ENV_IDENTITY_TOKEN_FILE, token);
-        env.put(app::config::ENV_IDENTITY_TRUST_ANCHORS, trust_anchors);
+        env.put(
+            app::config::ENV_IDENTITY_TRUST_ANCHORS,
+            trust_anchors.clone(),
```
hawkw: TIOLI: since `Identity::from_pem` takes an `&str`, you could call that function with `&trust_anchors` prior to inserting it into the test env, obviating the need for the clone. Since this is not called in a hot path, performance here isn't really important, but I thought I'd mention it.

tests/support/tap.rs
```diff
@@ -185,6 +185,10 @@ where
 
             buf.freeze()
         });
-        Box::new(self.0.request_body_async(req))
+        Box::new(
+            self.0
+                .request_body_async(req)
+                .map_err(|err| err.to_string()),
```
hawkw: TIOLI: I think this could also be 
```suggestion
                .map_err(ClientError::to_string),
```

zaharidichev: The compiler is not quite happy about that: 

```

tests/support/identity.rs
zaharidichev: Nice ! Thanks :)

tests/support/tap.rs
hawkw: ah whoops, nevermind then!

hawkw:
aaand...merged! thanks again for the contribution @zaharidichev!

PR #244, opened 2019-05-03
Remove Destination service query concurrency limit

hawkw:
Currently, the proxy enforces a limit on the number of concurrent
queries (i.e., the number of gRPC streams) to the Destination service.
This limit was added based on information about the behaviour of the
Destination service that is now known to be incorrect.

This branch removes the limit on concurrent queries from the proxy's
`control::destination` module. Although it should now be possible to
simplify this code as a result of this change, I've refrained from doing
any major refactoring in this branch --- my intention is to do this
after the DNS fallback behaviour has also been removed, as together with
this change, that will result in a _significant_ simplification of the
module. Additionally, I've removed the tests for the concurrency limit,
as they are no longer relevant.

The `LINKERD2_PROXY_DESTINATION_CLIENT_CONCURRENCY_LIMIT` 
environment variable was also removed; this is not a breaking change as
neither the CLI nor the proxy injector will currently set this env var.

### DIFF
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -242,14 +242,6 @@ pub const ENV_DESTINATION_GET_SUFFIXES: &str = "LINKERD2_PROXY_DESTINATION_GET_S
 /// If unspecified, a default value is used.
 pub const ENV_DESTINATION_PROFILE_SUFFIXES: &str = "LINKERD2_PROXY_DESTINATION_PROFILE_SUFFIXES";
 
-/// Limits the buffer capacity of the destination service client.
-///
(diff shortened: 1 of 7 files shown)


hawkw:
@olix0r just for clarity's sake, I'm assuming you mean renaming the field on `Config`, _not_ the env var, as renaming the env var would technically be a breaking change?

hawkw:
I've removed the env var entirely (as suggested by @olix0r offline), as it turns out that neither the proxy injector nor the CLI ever actually set it.

PR #246, opened 2019-05-04
Introduce dispatch timeouts around buffers

olix0r:
The proxy has several buffers, especially where it routes requests over
shared stacks. If any of these routes is unavailable, then a request may
remain buffered indefinitely. Previously, before service profiles were
introduced, there was a default _response_ timeout that would cause
these requests to fail; but since this response timeout is now optional
(and is only applied once the request has been routed within a proxy),
then we need a new mechanism to prevent requests from getting "stuck".

This change does the following:
- all proxied requests are annotated with a dispatch deadline;
- each time a request is bufered, a timeout is registered.
- if the timeout fires, the response exception fails, a 503 is returned,
  and the request is dropped.
- if the request is processed into the inner stack, the timeout is
  ignored.

The dispatch timeout limits the _time a request is buffered in a proxy_.
This is distinct from the response timeout, as the server's response may
naturally be delayed for any number of (non-proxy-related) reasons.

The `insert_target` module has been generalized to `insert` to support
setting the DispatchDeadline extension.

The `buffer` module has been augmented with generic deadline-extraction
logic.

The `svc` module now exposes its own builder type that notably adds
a `buffer_pending` helper. It's helpful to pull a builder type into the
proxy to assist debugging type errors when modifying stacks.

Fixes linkerd/linkerd2#2779 linkerd/linkerd2#2795


src/proxy/buffer.rs
```diff
@@ -1,91 +1,138 @@
...
+use futures::{Async, Future, Poll};
+use tokio::timer::Delay;
+use tower::buffer;
 
 use logging;
 use proxy::Error;
 use svc;
 
+pub trait ExtractDeadline<Req>: Clone {
+    fn extract(&self, req: &Req) -> Option<Instant>;
```
hawkw: I'm not sure how I feel about this name --- I think if I saw it elsewhere I wouldn't be clear what `extract` means? Maybe `fn extract_deadline`? Or, rename the trait to `HasDeadline` and just make this `fn deadline`?

src/proxy/buffer.rs
```diff
@@ -95,72 +142,335 @@ where
...
+        self.0.poll_ready().map_err(Into::into)
+    }
+
+    fn call(&mut self, req: Stealer<Req>) -> Self::Future {
+        match req.upgrade() {
+            None => DequeueFuture::Lost,
+            Some(l) => match l.lock() {
+                Err(_) => DequeueFuture::Lost,
+                Ok(ref mut v) => match v.take() {
+                    None => DequeueFuture::Lost,
```
hawkw: I feel like there may be a cleaner way to express this with `and_then`? Maybe
```rust
req.upgrade()
    .and_then(|l| { 
        l.lock().ok()?.take()
    })
    .map(|req| DequeueFuture::Inner(self.0.call(req)))
    .unwrap_or(DequeueFilter::Lost)
```

src/proxy/buffer.rs
olix0r: i'm -1 on a raw fn -- the signature gets gnarly with lifetimes

i'm +0 on Has -- We tend to use `Has..` with targets

`Extract` is what tower-web et al use to refer to pulling data out of requests, so it seemed relevant.

I do tink the fn should probably be called `deadline`...

olix0r: settled on just Deadline

src/proxy/pending.rs
```diff
@@ -19,8 +22,18 @@ pub enum Pending<F, S> {
 
 pub type Svc<M, T> = Pending<svc::Oneshot<M, T>, <M as svc::Service<T>>::Response>;
 
-pub fn layer<M>() -> impl svc::Layer<M, Service = MakePending<M>> + Copy {
-    svc::layer::mk(|inner| MakePending { inner })
+pub fn layer() -> Layer {
+    Layer(())
```
olix0r: Added so it has a nameable type for the Builder

src/proxy/buffer.rs
hawkw: > i'm -1 on a raw fn -- the signature gets gnarly with lifetimes

I didn't mean a raw fn, I just meant renaming the trait fn to `extract_deadline`. 

`deadline` seems fine :+1:

src/proxy/http/insert.rs
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    // === impl Stack ===
+
+    impl<T, M> svc::Service<T> for Make<M>
+    where
+        T: Clone + Send + Sync + 'static,
+        M: svc::Service<T>,
+    {
+        type Response = super::Service<M::Response, super::ValLazy<T>, T>;
```
kleimkuhler: Not sure I'm following why the `Response` is a `super::Service` that takes a `lazy: ValLazy` instead of `lazy: FnLazy`. Is the use of `ValLazy` here mainly to work with the fact that `insert::layer()` takes a closure, and `insert::target::layer()` does not?

src/svc.rs
```diff
@@ -1,12 +1,63 @@
...
+use proxy::{buffer, pending};
+
+#[derive(Clone, Debug)]
+pub struct Builder<L>(ServiceBuilder<L>);
+
+pub fn builder() -> Builder<Identity> {
+    Builder(ServiceBuilder::new())
+}
+
+impl<L> Builder<L> {
```
kleimkuhler: I think I like this? Seems like it will keep noise down in `app/main.rs` when we're building stacks, but it also introduces another place where we can define layer logic.

src/app/main.rs
```diff
@@ -472,14 +491,20 @@ where
...
+                .buffer_pending(max_in_flight, DispatchDeadline::extract)
                 .layer(classify::layer())
                 .layer(metrics::layer::<_, classify::Response>(route_http_metrics))
                 .layer(proxy::http::timeout::layer())
                 .layer(retry::layer(retry_http_metrics.clone()))
                 .layer(metrics::layer::<_, classify::Response>(retry_http_metrics))
-                .layer(insert_target::layer());
+                .layer(insert::target::layer());
+
+            let balancer_stack = svc::builder()
```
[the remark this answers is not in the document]
olix0r: Eh, this was split out so that it was easier to pin down the errors when it didn't compile. It could be reverted, but also seemed nice and readable this way.

src/proxy/buffer.rs
```diff
@@ -95,72 +147,351 @@ where
...
+impl<Req, F> Future for EnqueueFuture<F, Req>
+where
+    F: Future,
+    F::Error: Into<Error>,
+{
+    type Item = F::Item;
+    type Error = Error;
+
+    fn poll(&mut self) -> Poll<F::Item, Self::Error> {
+        if let Async::Ready(v) = self.inner.poll()? {
```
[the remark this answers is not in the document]
olix0r: We _do_ only return `Async::NotReady` when the `inner` future has returned `NotReady`

the logic is:
* poll for a response; if it fails, fail. if it succeeds, return it.
* otherwise, if there's no response, and the request still hasn't been dispatched, and if the timeout has expired, return an error.
* otherwise, return not ready. we'll get notified when either the response or timeout future is satisfied.

src/proxy/buffer.rs
```diff
@@ -95,72 +147,351 @@ where
...
+        write!(f, "the request could not be dispatched in a timely fashion")
+    }
+}
+
+impl error::Error for Aborted {}
+
+// === impl Deadline ===
+
+impl<Req> Deadline<Req> for () {
+    fn deadline(&self, _: &Req) -> Option<Instant> {
```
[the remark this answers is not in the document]
olix0r: Yep.

src/app/main.rs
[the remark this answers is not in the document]
hawkw: I did pretty much the same thing in #240, so I'm definitely +1 on this.

src/proxy/http/insert.rs
olix0r: Yep. target can't take a fn because it needs the ... target. it's probably not strictly necessary to have `ValLazy`

src/svc.rs
olix0r: Yeah, I see it as a place for capturing repeated patterns we use in the main. It should be used sparingly, though we certainly may want to split out more common stuff into this.

src/app/config.rs
```diff
@@ -371,9 +386,16 @@ impl Config {
         let admin_listener_addr = parse(strings, ENV_ADMIN_LISTEN_ADDR, parse_socket_addr);
         let inbound_forward = parse(strings, ENV_INBOUND_FORWARD, parse_socket_addr);
 
+        let inbound_dispatch_timeout = parse(strings, ENV_INBOUND_DISPATCH_TIMEOUT, parse_duration);
```
hawkw: Nitpicky: Should we have `parse_timeout` function for these that takes INBOUND_BASE/OUTBOUND_BASE/CONTROL_BASE and a "kind" parameter (connect/dispatch)? It seems like other stuff that's configurable separately for inbound/outbound/ctrl is moving in that direction...

src/app/main.rs
```diff
@@ -516,8 +536,7 @@ where
                     debug!("outbound dst={:?}", addr);
                     addr
                 }))
-                .layer(buffer::layer(max_in_flight))
-                .layer(pending::layer())
+                .buffer_pending(max_in_flight, DispatchDeadline::extract)
```
hawkw: TIOLI: this seems to be repeated enough that maybe there should be a `buffer_with_deadline` helper or something? on the other hand that might not really make the code any simpler.

src/proxy/buffer.rs
```diff
@@ -95,72 +147,383 @@ where
...
+    type Error = Error;
+
+    fn poll(&mut self) -> Poll<F::Item, Self::Error> {
+        if let Async::Ready(v) = self.inner.poll()? {
+            return Ok(Async::Ready(v));
+        }
+
+        // If the request hasn't been consumed by `Dequeue`, then steal it and
+        // drop it when the timeout fires.
+        if let Ok(mut h) = self.holder.lock() {
```
hawkw: If this returns `Err`, then that means the mutex is poisoned. I don't think silently doing nothing is the right way to handle poisons here --- we'll fall through and return `NotReady`, leaking the task. Instead, we should _probably_ propagate the panic, I think.

src/proxy/buffer.rs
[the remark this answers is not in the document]
kleimkuhler: Okay cool thanks for the explanation.

src/proxy/buffer.rs
```diff
@@ -95,72 +147,383 @@ where
...
+                1,
+                &mut logging::context_executor("test"),
+            );
+
+            assert!(svc.poll_ready().ok().map(|r| r.is_ready()).unwrap_or(false));
+            let call = svc.call(());
+            drop(svc);
+
+            let call = tokio::timer::Timeout::new(call, Duration::from_millis(100))
+            .then(|r| match r {
```
hawkw: Nit: weird indentation here?

src/proxy/buffer.rs
```diff
@@ -95,72 +147,383 @@ where
...
+                Duration::from_secs(0),
+                1,
+                &mut logging::context_executor("test"),
+            );
+
+            assert!(svc.poll_ready().ok().map(|r| r.is_ready()).unwrap_or(false));
+            let call = svc.call(());
+            drop(svc);
+
+            let call = tokio::timer::Timeout::new(call, Duration::from_millis(100))
```
hawkw: TIOLI: Now that the [`tokio-test` crate](https://github.com/tokio-rs/tokio/tree/master/tokio-test) has been released, we might want to look into rewriting this (and other tests) to use mock timers instead. Maybe a follow-up?

src/proxy/http/insert.rs
```diff
@@ -0,0 +1,226 @@
...
+use std::marker::PhantomData;
+
+use svc;
+
+pub trait Lazy<V>: Clone {
+    fn value(&self) -> V;
+}
+
+/// Wraps an HTTP `Service` so that the `T -typed value` is cloned into
+/// each request's extensions.
```
hawkw: The way you generalized this is extremely cool :)

src/proxy/buffer.rs
olix0r: let's get an issue in the backlog for this?

src/proxy/buffer.rs
olix0r: I mean, this is all moot right now because this mutex isn't actually shared across threads, but point taken.

src/app/config.rs
olix0r: Sure, but let's make that it's own change if/when we do it.

src/proxy/buffer.rs
hawkw: sounds good, i opened linkerd/linkerd2#2802

src/proxy/buffer.rs
```diff
@@ -310,15 +310,17 @@ where
...
+        let mut h = self.holder.lock().expect("inner service panicked");
+        if h.is_some() {
+            if let Some(t) = self.timeout.as_mut() {
+                if t.poll().map_err(Error::from)?.is_ready() {
+                    drop(h.take());
+                    return Err(Aborted.into());
                 }
             }
+        } else {
+            // Drop the timeout future so the timer doesn't need to track it.
```
hawkw: 👍

src/app/errors.rs
```diff
@@ -107,6 +108,9 @@ fn map_err_to_5xx(e: Error) -> StatusCode {
     } else if let Some(_) = e.downcast_ref::<shed::Overloaded>() {
         warn!("server overloaded, max-in-flight reached");
         http::StatusCode::SERVICE_UNAVAILABLE
+    } else if let Some(_) = e.downcast_ref::<buffer::Aborted>() {
+        warn!("dispatch aborted");
```
seanmonstar: Should this message say a little more? Such a terse message could be confusing in logs.

PR #247, opened 2019-05-04
Eagerly drop services from the router cache

kleimkuhler:
This change introduces an eviction task for router caches. Few changes were made
in how a router uses a cache; we still use `access` and `reserve` handles for
cache interactions.

### Current issue

The issue this change addresses is that services are only dropped from a router
cache when we have reached that cache's capacity.

We drop a service when it has not received a request within the `max_idle_age`.
When the cache is full, we scan through all the services and check their last
access time. If a service's last access time is longer than `max_idle_age`, then
it is removed.

This strategy is problematic because a service is only removed once the capacity
has been reached. If the capacity of a cache is never reached, then a service's
stack is kept around for the lifetime of that router.

### Proposed solution

When a router is created, we now also create a handle to that router's purge
task (the `PurgeCache` struct). If a router should attempt to purge its services
in the background, it can spawn that handle when the router is created.

A cache's state is now internally maintained by a
[tokio_timer::DelayQueue](https://docs.rs/tokio/0.1.19/tokio/timer/struct.DelayQueue.html).
When `PurgeCache` is able to pop services from the queue, we know that those
services have not received a request within an `expires` span of time -- even if
the cache has not reached it's capacity.

An important thing to note about this strategy is we still rely on locking the
*entire* cache when holding an `access`/`reserve` handle, as well as now purging
it in the background. The purge task will only ever attempt to lock the cache
with `try_lock!`, this ensures we do not cause lock contention with a more
important `access`/`reserve` interaction.

Another important note is that `PurgeCache` only has a weak reference to the
router cache. If the router is dropped and a purge task was previously spawned
on the runtime, then it will not continue to try and purge cache values.

Note: Implementation originally sketched out
[here](https://github.com/kleimkuhler/tec/)

### DIFF
--- a/lib/router/src/cache.rs
+++ b/lib/router/src/cache.rs
@@ -2,7 +2,7 @@ use futures::{Async, Future, Poll, Stream};
 use indexmap::IndexMap;
 use std::{hash::Hash, time::Duration};
 use tokio::sync::lock::Lock;
-use tokio_timer::{DelayQueue, Error, Interval};
+use tokio_timer::{delay_queue, DelayQueue, Error, Interval};
 
 /// A cache that is internally maintained by a `tokio_timer::DelayQueue`.
 ///
@@ -13,12 +13,12 @@ where
 {
     capacity: usize,
     expires: Duration,
-    /// Elements are keys into `values`. As elements become Ready, we can remove
-    /// the key and corresponding value from the cache.
+    /// Elements are keys into `values`. As elements become Ready, we can
+    /// remove the key and corresponding value from the cache.
     expirations: DelayQueue<K>,
-    /// Cache access is coordinated through `values`. This field represents the
-    /// current state of the cache.
-    values: IndexMap<K, V>,
+    /// Cache access is coordinated through `values`. This field represents
+    /// the current state of the cache.
+    values: IndexMap<K, Node<V>>,
 }
 
 /// A background future that eagerly removes expired cache values.
@@ -32,19 +32,15 @@ where
     interval: Interval,
 }
 
+/// A handle to a cache value.
+pub struct Node<T> {
+    dq_key: delay_queue::Key,
+    value: T,
+}
+
 #[derive(Debug, PartialEq)]
 pub struct CapacityExhausted {
     pub(crate) capacity: usize,
-}
-
-/// A handle to a cache that has capacity for at least one additional value.
-pub struct Reserve<'a, K, V>
-where
-    K: Clone + Eq + Hash,
-{
-    expirations: &'a mut DelayQueue<K>,
-    expires: Duration,
-    values: &'a mut IndexMap<K, V>,
 }
 
 // ===== impl Cache =====
@@ -71,29 +67,44 @@ where
         (cache, bg_purge)
     }
 
-    pub fn access(&mut self, key: &K) -> Option<V> {
-        let value = self.values.get_mut(key)?;
-        Some(value.clone())
-    }
-
     pub fn capacity(&self) -> usize {
         self.capacity
     }
 
-    pub fn reserve(&mut self) -> Poll<Reserve<K, V>, ()> {
+    pub fn get(&mut self, key: &K) -> Option<V> {
+        if let Some(node) = self.values.get_mut(key) {
+            self.expirations.reset(node.dq_key_ref(), self.expires);
+            return Some(node.value_ref().clone());
+        }
+
+        None
+    }
+
+    pub fn insert(&mut self, key: K, value: V) -> Option<Node<V>> {
+        let node = {
+            let dq_key = self.expirations.insert(key.clone(), self.expires);
+            Node::new(dq_key, value)
+        };
+
+        self.values.insert(key, node)
+    }
+
+    pub fn poll_insert(&mut self) -> Poll<(), ()> {
+        // When checking capacity, only try to remove values if the cache is
+        // at capacity
         if self.values.len() == self.capacity {
             match self.expirations.poll() {
-                // The cache is at capacity but we are able to remove a value.
-                Ok(Async::Ready(Some(entry))) => {
-                    self.values.remove(entry.get_ref());
+                // The cache is at capacity, but we are able to remove a value
+                Ok(Async::Ready(Some(expired))) => {
+                    self.values.remove(expired.get_ref());
                 }
 
-                // `Ready(None)` can only be returned when expirations is
-                // empty. We know `expirations` is not empty because `values` is
-                // not empty and capacity does not equal zero.
+                // `Ready(None)` can only be returned when `expirations` is
+                // empty; we know `expirations` is not empty because `values`
+                // is not empty and capacity does not equal zero.
                 Ok(Async::Ready(None)) => unreachable!("cache expirations cannot be empty"),
 
-                // The cache is at capacity and no values can be removed.
+                // The cache is at capacity and no values can be removed
                 Ok(Async::NotReady) => {
                     return Ok(Async::NotReady);
                 }
@@ -102,16 +113,12 @@ where
             }
         }
 
-        Ok(Async::Ready(Reserve {
-            expirations: &mut self.expirations,
-            expires: self.expires,
-            values: &mut self.values,
-        }))
+        Ok(Async::Ready(()))
     }
 
     fn poll_purge(&mut self) -> Poll<(), Error> {
-        while let Some(entry) = try_ready!(self.expirations.poll()) {
-            self.values.remove(entry.get_ref());
+        while let Some(expired) = try_ready!(self.expirations.poll()) {
+            self.values.remove(expired.get_ref());
         }
 
         Ok(Async::Ready(()))
@@ -133,34 +140,27 @@ where
             panic!("PurgeCache.interval Interval::poll must not fail: {}", e);
         }));
 
-        let mut acquired = try_ready!(Ok(self.cache.poll_lock()));
+        let mut cache = try_ready!(Ok(self.cache.poll_lock()));
 
-        acquired.poll_purge().map_err(|e| {
+        cache.poll_purge().map_err(|e| {
             panic!("PurgeCache.cache Cache::poll_purge must not fail: {}", e);
         })
     }
 }
 
-// ===== impl Access =====
+// ===== impl Node =====
 
-// impl<'a, K, V> Drop for Access<'a, K, V>
-// where
-//     K: Clone + Eq + Hash,
-// {
-//     fn drop(&mut self) {
-//         self.expirations.reset(&self.node.key, self.expires);
-//     }
-// }
+impl<T> Node<T> {
+    pub fn new(dq_key: delay_queue::Key, value: T) -> Self {
+        Node { dq_key, value }
+    }
 
-// ===== impl Reserve =====
+    pub fn dq_key_ref(&self) -> &delay_queue::Key {
+        &self.dq_key
+    }
 
-impl<'a, K, V> Reserve<'a, K, V>
-where
-    K: Clone + Eq + Hash,
-{
-    pub fn store(self, key: K, value: V) {
-        let _delay = self.expirations.insert(key.clone(), self.expires);
-        self.values.insert(key, value);
+    pub fn value_ref(&self) -> &T {
+        &self.value
     }
 }
 
@@ -171,7 +171,7 @@ mod tests {
     use tokio::runtime::current_thread::{self, Runtime};
 
     #[test]
-    fn reserve_and_store() {
+    fn check_capacity_and_insert() {
         current_thread::run(future::lazy(|| {
             let (mut cache, _cache_purge) = Cache::new(2, Duration::from_millis(10));
 
@@ -181,26 +181,26 @@ mod tests {
             };
 
             {
-                let slot = match cache.reserve() {
-                    Ok(Async::Ready(slot)) => slot,
+                match cache.poll_insert().unwrap() {
+                    Async::Ready(()) => (),
                     _ => panic!("cache should be Ready to reserve"),
                 };
-                slot.store(1, 2);
+                cache.insert(1, 2);
             }
             assert_eq!(cache.values.len(), 1);
 
             {
-                let slot = match cache.reserve() {
-                    Ok(Async::Ready(slot)) => slot,
+                match cache.poll_insert().unwrap() {
+                    Async::Ready(()) => (),
                     _ => panic!("cache should be Ready to reserve"),
                 };
-                slot.store(2, 3);
+                cache.insert(2, 3);
             }
             assert_eq!(cache.values.len(), 2);
 
             {
-                match cache.reserve() {
-                    Ok(Async::NotReady) => (),
+                match cache.poll_insert().unwrap() {
+                    Async::NotReady => (),
                     _ => panic!("cache should not be Ready to reserve"),
                 }
             }
@@ -211,7 +211,7 @@ mod tests {
     }
 
     #[test]
-    fn store_access_value() {
+    fn insert_and_access_value() {
         current_thread::run(future::lazy(|| {
             let (mut cache, _cache_purge) = Cache::new(2, Duration::from_millis(10));
 
@@ -220,38 +220,38 @@ mod tests {
                 _ => panic!("cache lock should be Ready"),
             };
 
-            assert!(cache.access(&1).is_none());
-            assert!(cache.access(&2).is_none());
+            assert!(cache.get(&1).is_none());
+            assert!(cache.get(&2).is_none());
 
             {
-                let slot = match cache.reserve() {
-                    Ok(Async::Ready(slot)) => slot,
+                match cache.poll_insert().unwrap() {
+                    Async::Ready(()) => (),
                     _ => panic!("cache should be Ready to reserve"),
                 };
-                slot.store(1, 2);
+                cache.insert(1, 2);
             }
-            assert!(cache.access(&1).is_some());
-            assert!(cache.access(&2).is_none());
+            assert!(cache.get(&1).is_some());
+            assert!(cache.get(&2).is_none());
 
             {
-                let slot = match cache.reserve() {
-                    Ok(Async::Ready(slot)) => slot,
+                match cache.poll_insert().unwrap() {
+                    Async::Ready(()) => (),
                     _ => panic!("cache should be Ready to reserve"),
                 };
-                slot.store(2, 3);
+                cache.insert(2, 3);
             }
-            assert!(cache.access(&1).is_some());
-            assert!(cache.access(&2).is_some());
+            assert!(cache.get(&1).is_some());
+            assert!(cache.get(&2).is_some());
 
-            assert_eq!(cache.access(&1).take().unwrap(), 2);
-            assert_eq!(cache.access(&2).take().unwrap(), 3);
+            assert_eq!(cache.get(&1).take().unwrap(), 2);
+            assert_eq!(cache.get(&2).take().unwrap(), 3);
 
             Ok::<_, ()>(())
         }))
     }
 
     #[test]
-    fn reserve_does_nothing_when_capacity_exists() {
+    fn poll_insert_does_nothing_when_capacity_exists() {
         current_thread::run(futures::future::lazy(|| {
             let (mut cache, _cache_purge) = Cache::new(2, Duration::from_millis(10));
 
@@ -261,17 +261,17 @@ mod tests {
             };
 
             {
-                let slot = match cache.reserve() {
-                    Ok(Async::Ready(slot)) => slot,
+                match cache.poll_insert().unwrap() {
+                    Async::Ready(()) => (),
                     _ => panic!("cache should be Ready to reserve"),
                 };
-                slot.store(1, 2);
+                cache.insert(1, 2);
             }
             assert_eq!(cache.values.len(), 1);
 
             {
-                let _slot = match cache.reserve() {
-                    Ok(Async::Ready(slot)) => slot,
+                match cache.poll_insert().unwrap() {
+                    Async::Ready(()) => (),
                     _ => panic!("cache should be Ready to reserve"),
                 };
             }
@@ -282,7 +282,7 @@ mod tests {
     }
 
     #[test]
-    fn store_and_self_purge() {
+    fn insert_and_self_purge() {
         let mut rt = Runtime::new().unwrap();
 
         let (mut cache, _cache_purge) = rt
@@ -299,18 +299,18 @@ mod tests {
             };
 
             {
-                let slot = match cache_1.reserve() {
-                    Ok(Async::Ready(slot)) => slot,
+                match cache_1.poll_insert().unwrap() {
+                    Async::Ready(()) => (),
                     _ => panic!("cache should be Ready to reserve"),
                 };
-                slot.store(1, 2);
+                cache_1.insert(1, 2);
             }
             {
-                let slot = match cache_1.reserve() {
-                    Ok(Async::Ready(slot)) => slot,
+                match cache_1.poll_insert().unwrap() {
+                    Async::Ready(()) => (),
                     _ => panic!("cache should be Ready to reserve"),
                 };
-                slot.store(2, 3);
+                cache_1.insert(2, 3);
             }
             assert_eq!(cache_1.values.len(), 2);
 
@@ -330,14 +330,14 @@ mod tests {
             };
 
             {
-                let slot = match cache_2.reserve() {
-                    Ok(Async::Ready(slot)) => slot,
+                match cache_2.poll_insert().unwrap() {
+                    Async::Ready(()) => (),
                     _ => panic!("cache should be Ready to reserve"),
                 };
-                slot.store(3, 4);
+                cache_2.insert(3, 4);
             }
             assert_eq!(cache_2.values.len(), 2);
-            assert!(cache_2.access(&3).is_some());
+            assert!(cache_2.get(&3).is_some());
 
             Ok::<_, ()>(())
         }))
@@ -345,7 +345,7 @@ mod tests {
     }
 
     #[test]
-    fn store_and_background_purge() {
+    fn isnert_and_background_purge() {
         let mut rt = Runtime::new().unwrap();
 
         let (mut cache, cache_purge) = rt
@@ -365,18 +365,18 @@ mod tests {
             };
 
             {
-                let slot = match cache.reserve() {
-                    Ok(Async::Ready(slot)) => slot,
+                match cache.poll_insert().unwrap() {
+                    Async::Ready(()) => (),
                     _ => panic!("cache should be Ready to reserve"),
                 };
-                slot.store(1, 2);
+                cache.insert(1, 2);
             }
             {
-                let slot = match cache.reserve() {
-                    Ok(Async::Ready(slot)) => slot,
+                match cache.poll_insert().unwrap() {
+                    Async::Ready(()) => (),
                     _ => panic!("cache should be Ready to reserve"),
                 };
-                slot.store(2, 3);
+                cache.insert(2, 3);
             }
             assert_eq!(cache.values.len(), 2);
 
@@ -393,73 +393,5 @@ mod tests {
             _ => panic!("cache lock should be Ready"),
         };
         assert_eq!(cache.values.len(), 0);
-    }
-
-    #[test]
-    fn drop_access_resets_expiration() {
-        let mut rt = current_thread::Runtime::new().unwrap();
-
-        let (mut cache, cache_purge) = rt
-            .block_on(futures::future::lazy(|| {
-                Ok::<_, ()>(Cache::new(2, Duration::from_millis(10)))
-            }))
-            .unwrap();
-
-        // Spawn a background purge task on the runtime
-        rt.spawn(cache_purge);
-
-        {
-            let mut cache = match cache.poll_lock() {
-                Async::Ready(acquired) => acquired,
-                _ => panic!("cache lock should be Ready"),
-            };
-
-            // Hold on to an access handle the cache
-            let _access = rt
-                .block_on(future::lazy(|| {
-                    {
-                        let slot = match cache.reserve() {
-                            Ok(Async::Ready(slot)) => slot,
-                            _ => panic!("cache should be Ready to reserve"),
-                        };
-                        slot.store(1, 2);
-                    }
-                    assert!(cache.access(&1).is_some());
-
-                    Ok::<_, ()>(cache.access(&1))
-                }))
-                .unwrap();
-
-            // Sleep for enough time that the background purge would remove the value
-            rt.block_on(tokio_timer::sleep(Duration::from_millis(100)))
-                .unwrap();
-
-            // Drop both the access and cache handles. Dropping the
-            // access handle will reset the expiration on the value in the cache.
-            // Dropping the cache handle will unlock the cache and allow a
-            // background purge to occur.
-        }
-
-        // Ensure a background purge is polled so that it can expire any
-        // values.
-        rt.block_on(future::lazy(|| {
-            tokio_timer::sleep(Duration::from_millis(1))
-        }))
-        .unwrap();
-
-        rt.block_on(future::lazy(|| {
-            let mut cache = match cache.poll_lock() {
-                Async::Ready(acquired) => acquired,
-                _ => panic!("cache lock should be Ready"),
-            };
-
-            // The cache value should still be present since it was reset after
-            // the value expiration. We ensured a background purge occurred but
-            // that it did not purge the value.
-            assert!(cache.access(&1).is_some());
-
-            Ok::<_, ()>(())
-        }))
-        .unwrap()
     }
 }
--- a/lib/router/src/cache.rs
+++ b/lib/router/src/cache.rs
@@ -8,13 +8,13 @@ use tokio_timer::{delay_queue, DelayQueue, Error, Interval};
 /// An LRU cache that can eagerly remove values in a background task.
 ///
 /// Assumptions:
-///     - `access` is common
-///     - `insert` is less common
-///     - Values should have an `expires` span of time greater than 0
+/// - `access` is common
+/// - `insert` is less common
+/// - Values should have an `expires` span of time greater than 0
 ///
 /// Complexity:
-///     - `access` in **O(1)** time (amortized average)
-///     - `insert` in **O(1)** time (average)
+/// - `access` in **O(1)** time (amortized average)
+/// - `insert` in **O(1)** time (average)
 ///
 /// The underlying data structure of Cache is a [`DelayQueue`]. This allows
 /// the background task to remove values by polling for elements that have
@@ -30,7 +30,8 @@ where
 {
     capacity: usize,
     expires: Duration,
-    /// Elements are keys into `values`. As elements become Ready, we can
+    /// A queue of keys into `values` that become ready when the corresponding
+    /// cache entry expires. As elements become ready, we can
     /// remove the key and corresponding value from the cache.
     expirations: DelayQueue<K>,
     /// Cache access is coordinated through `values`. This field represents
--- a/lib/router/src/cache.rs
+++ b/lib/router/src/cache.rs
@@ -88,7 +88,7 @@ impl<K: Clone + Eq + Hash, V> Cache<K, V> {
         })
     }
 
-    pub fn reserve(&mut self) -> Result<Reserve<K, V>, CapacityExhausted> {
+    pub fn poll_reserve(&mut self) -> Result<Reserve<K, V>, CapacityExhausted> {
         // If the cache capacity is zero, then we have no space to reserve a slot
         if self.capacity == 0 {
             return Err(CapacityExhausted {
@@ -215,14 +215,14 @@ mod tests {
 
             let mut cache = cache.lock().unwrap();
 
-            cache.reserve().expect("reserve").store(1, 2);
+            cache.poll_reserve().expect("reserve").store(1, 2);
             assert_eq!(cache.vals.len(), 1);
 
-            cache.reserve().expect("reserve").store(2, 3);
+            cache.poll_reserve().expect("reserve").store(2, 3);
             assert_eq!(cache.vals.len(), 2);
 
             assert_eq!(
-                cache.reserve().err(),
+                cache.poll_reserve().err(),
                 Some(CapacityExhausted { capacity: 2 })
             );
             assert_eq!(cache.vals.len(), 2);
@@ -241,11 +241,11 @@ mod tests {
             assert!(cache.access(&1).is_none());
             assert!(cache.access(&2).is_none());
 
-            cache.reserve().expect("reserve").store(1, 2);
+            cache.poll_reserve().expect("reserve").store(1, 2);
             assert!(cache.access(&1).is_some());
             assert!(cache.access(&2).is_none());
 
-            cache.reserve().expect("reserve").store(2, 3);
+            cache.poll_reserve().expect("reserve").store(2, 3);
             assert!(cache.access(&1).is_some());
             assert!(cache.access(&2).is_some());
 
@@ -263,10 +263,10 @@ mod tests {
 
             let mut cache = cache.lock().unwrap();
 
-            cache.reserve().expect("capacity").store(1, 2);
+            cache.poll_reserve().expect("capacity").store(1, 2);
             assert_eq!(cache.vals.len(), 1);
 
-            assert!(cache.reserve().is_ok());
+            assert!(cache.poll_reserve().is_ok());
             assert_eq!(cache.vals.len(), 1);
 
             Ok(())
@@ -287,8 +287,8 @@ mod tests {
         rt.block_on(future::lazy(|| {
             let mut cache_1 = cache.lock().unwrap();
 
-            cache_1.reserve().expect("reserve").store(1, 2);
-            cache_1.reserve().expect("reserve").store(2, 3);
+            cache_1.poll_reserve().expect("reserve").store(1, 2);
+            cache_1.poll_reserve().expect("reserve").store(2, 3);
             assert_eq!(cache_1.vals.len(), 2);
 
             Ok::<_, ()>(())
@@ -303,7 +303,7 @@ mod tests {
         rt.block_on(future::lazy(|| {
             let mut cache_2 = cache.lock().unwrap();
 
-            cache_2.reserve().expect("reserve").store(3, 4);
+            cache_2.poll_reserve().expect("reserve").store(3, 4);
             assert_eq!(cache_2.vals.len(), 2);
             assert!(cache_2.access(&3).is_some());
 
@@ -329,8 +329,8 @@ mod tests {
         rt.block_on(future::lazy(|| {
             let mut cache = cache.lock().unwrap();
 
-            cache.reserve().expect("reserve").store(1, 2);
-            cache.reserve().expect("reserve").store(2, 3);
+            cache.poll_reserve().expect("reserve").store(1, 2);
+            cache.poll_reserve().expect("reserve").store(2, 3);
             assert_eq!(cache.vals.len(), 2);
 
             Ok::<_, ()>(())
@@ -363,7 +363,7 @@ mod tests {
             // Hold on to an access handle the cache
             let _access = rt
                 .block_on(future::lazy(|| {
-                    cache.reserve().expect("reserve").store(1, 2);
+                    cache.poll_reserve().expect("reserve").store(1, 2);
                     assert!(cache.access(&1).is_some());
 
                     Ok::<_, ()>(cache.access(&1))
--- a/lib/router/src/cache.rs
+++ b/lib/router/src/cache.rs
@@ -1,10 +1,7 @@
 use futures::{Async, Future, Poll, Stream};
 use indexmap::IndexMap;
-use std::{
-    hash::Hash,
-    sync::{Arc, Mutex, TryLockError, Weak},
-    time::Duration,
-};
+use std::{hash::Hash, time::Duration};
+use tokio::sync::lock::Lock;
 use tokio_timer::{delay_queue, DelayQueue, Error, Interval};
 
 /// A cache that is internally maintained by a `tokio_timer::DelayQueue`.
@@ -16,7 +13,7 @@ use tokio_timer::{delay_queue, DelayQueue, Error, Interval};
 /// can be polled in the background and remove expired values.
 ///
 /// All values in the cache will expire after a `expires` span of time.
-pub struct Cache<K: Clone + Eq + Hash, V> {
+pub struct Cache<K: Clone + Eq + Hash + Send, V> {
     capacity: usize,
     expires: Duration,
     expirations: DelayQueue<K>,
@@ -28,15 +25,15 @@ pub struct Cache<K: Clone + Eq + Hash, V> {
 ///
 /// This contains a weak reference to a cache so that if use of the cache is
 /// dropped, it will not continue to be polled in the background.
-pub struct PurgeCache<K: Clone + Eq + Hash, V> {
-    cache: Weak<Mutex<Cache<K, V>>>,
+pub struct PurgeCache<K: Clone + Eq + Hash + Send, V> {
+    cache: Lock<Cache<K, V>>,
     interval: Interval,
 }
 
 /// Wraps a cache value so that a lock is held on the entire cache. When the
 /// access is dropped, the associated expiration time of the value in
 /// `expirations` is reset.
-pub struct Access<'a, K: Clone + Eq + Hash, V> {
+pub struct Access<'a, K: Clone + Eq + Hash + Send, V> {
     expires: Duration,
     expirations: &'a mut DelayQueue<K>,
     pub(crate) node: &'a mut Node<V>,
@@ -54,7 +51,7 @@ pub struct CapacityExhausted {
 }
 
 /// A handle to a cache that has capacity for at least one additional value.
-pub struct Reserve<'a, K: Clone + Eq + Hash, V> {
+pub struct Reserve<'a, K: Clone + Eq + Hash + Send, V> {
     expirations: &'a mut DelayQueue<K>,
     expires: Duration,
     vals: &'a mut IndexMap<K, Node<V>>,
@@ -62,17 +59,17 @@ pub struct Reserve<'a, K: Clone + Eq + Hash, V> {
 
 // ===== impl Cache =====
 
-impl<K: Clone + Eq + Hash, V> Cache<K, V> {
-    pub fn new(capacity: usize, expires: Duration) -> (Arc<Mutex<Cache<K, V>>>, PurgeCache<K, V>) {
+impl<K: Clone + Eq + Hash + Send, V> Cache<K, V> {
+    pub fn new(capacity: usize, expires: Duration) -> (Lock<Cache<K, V>>, PurgeCache<K, V>) {
         let cache = Self {
             capacity,
             expires,
             expirations: DelayQueue::with_capacity(capacity),
             vals: IndexMap::default(),
         };
-        let cache = Arc::new(Mutex::new(cache));
+        let cache = Lock::new(cache);
         let bg_purge = PurgeCache {
-            cache: Arc::downgrade(&cache),
+            cache: cache.clone(),
             interval: Interval::new_interval(expires),
         };
 
@@ -137,7 +134,7 @@ impl<K: Clone + Eq + Hash, V> Cache<K, V> {
 
 // ===== impl PurgeCache =====
 
-impl<K: Clone + Eq + Hash, V> Future for PurgeCache<K, V> {
+impl<K: Clone + Eq + Hash + Send, V> Future for PurgeCache<K, V> {
     type Item = ();
     type Error = ();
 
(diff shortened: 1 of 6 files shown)


lib/router/src/lib.rs
```diff
@@ -150,8 +148,8 @@ where
     ///
     /// TODO Attempt to free capacity in the router.
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
-        // FIXME(kleimkuhler): impl Send for Target and Svc?
-        // tokio::spawn(self.inner._cache_bg);
+        let inner = self.inner.clone();
+        tokio::spawn(inner.cache_bg);
```
hawkw: I don't understand why every `poll_ready` spawns a new `cache_bg`? My expectation is that each router would run a single background task...

lib/router/src/lib.rs
```diff
@@ -117,14 +115,14 @@ where
...
-        let (cache, _cache_bg) = Cache::new(capacity, max_idle_age);
+        let (cache, cache_bg) = Cache::new(capacity, max_idle_age);
 
         Router {
             inner: Arc::new(Inner {
                 recognize,
                 make,
                 cache,
-                _cache_bg,
+                cache_bg,
```
hawkw: I don't really get why the cache purging task is a field on `Inner` --- I think it makes more sense to have `Router::new` return `(Router, PurgeCache)` instead, so that the caller can spawn the purging task...

lib/router/src/cache.rs
```diff
@@ -1,426 +1,367 @@
...
+            match self.expirations.poll() {
+                // The cache is at capacity but we are able to remove a value.
+                Ok(Async::Ready(Some(entry))) => {
+                    self.vals.remove(entry.get_ref());
+                }
+
+                // `Ready(None)` can only be returned when expirations is
+                // empty. We know `expirations` is not empty because `vals` is
+                // not empty.
+                Ok(Async::Ready(None)) => unreachable!(),
```
hawkw: Extremely nitpicky: it might be nice to add a textual description to the `unreachable!()` macro, rather than explaining in a comment? That way, if this ever _is_ hit, the panic message will help explain the bug. Take it or leave it.

lib/router/src/cache.rs
```diff
@@ -1,426 +1,367 @@
...
+
+                // The cache is at capacity and no values can be removed.
+                Ok(Async::NotReady) => {
+                    return Err(CapacityExhausted {
+                        capacity: self.capacity,
+                    });
+                }
+
+                Err(_e) => {
+                    return Err(CapacityExhausted {
```
hawkw: I'm not sure if it's correct to return `CapacityExhausted` here --- the error would be a timer error in this case, and I think it would be confusing to return `CapacityExhausted` when that isn't _actually_ the problem. Instead we should either propagate this error or panic, depending on whether or not the situations where the timer can error should be considered fatal...

lib/router/src/cache.rs
```diff
@@ -1,426 +1,367 @@
...
 
-impl<'a, T: 'a, N: Now + 'a> Deref for Access<'a, T, N> {
-    type Target = T;
-    fn deref(&self) -> &Self::Target {
-        &self.node
-    }
-}
+        let lock = match self.cache.upgrade() {
+            Some(lock) => lock,
+            None => return Ok(Async::Ready(())),
```
hawkw: A comment here stating that failing to upgrade the weak ref means that the cache is dropped, so the purging cache should end might be good?

lib/router/src/cache.rs
```diff
@@ -1,426 +1,367 @@
...
-}
+impl<K: Clone + Eq + Hash, V> Future for PurgeCache<K, V> {
+    type Item = ();
+    type Error = ();
 
-// ===== impl Access =====
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        try_ready!(self.interval.poll().map_err(|e| {
+            println!("error polling purge interval: {:?}", e);
+            ()
```
hawkw: Note that the return type of `println!(...);` is also `()` so you don't need an explicit unit here.

lib/router/src/cache.rs
```diff
@@ -1,426 +1,367 @@
...
+                return Ok(Async::NotReady);
+            }
+        };
 
-impl<'a, T: 'a, N: Now + 'a> DerefMut for Access<'a, T, N> {
-    fn deref_mut(&mut self) -> &mut Self::Target {
-        &mut self.node
+        cache.poll_purge().map_err(|e| {
+            println!("error purging cache: {:?}", e);
+            ()
```
hawkw: Same as above, you don't actually need this.

lib/router/src/cache.rs
```diff
@@ -1,426 +1,367 @@
...
+            }
 
-        // Capacity at 3+s.
-        clock.advance(Duration::from_secs(1));
-        assert!(cache.reserve().is_ok());
-        assert_eq!(cache.vals.len(), 0);
+            // Sleep for an amount of time greater than expiry duration so
+            // that a value can be purged.
+            tokio::spawn(
+                tokio_timer::sleep(Duration::from_millis(100))
```
hawkw: Now that the [`tokio-test` crate](https://github.com/tokio-rs/tokio/tree/master/tokio-test) has been released, we might want to look into rewriting this (and other tests) to use mock timers instead. It would be fine to do this in a follow up branch though.

lib/router/src/cache.rs
```diff
@@ -1,426 +1,367 @@
...
 
     #[test]
-    fn node_access_updated_on_drop() {
-        let mut clock = Clock::default();
-        let t0 = clock.now();
-        let mut node = Node::new(123, t0);
+    fn drop_access_resets_expiration() {
+        use tokio::runtime::current_thread::Runtime;
+
+        let mut runtime = Runtime::new().unwrap();
```
hawkw: TIOLI, but I might recommend rewriting the tests above that use `tokio::run`/`tokio::spawn` to also use the current_thread `Runtime`, for consistency? Also, the pattern of `block_on` -> assert -> `block_on` rather than chaining together futures that make assertions might lead to less "rightward drift" in indentation...

lib/router/src/cache.rs
```diff
@@ -1,426 +1,367 @@
...
+            let _access = cache_1.access(&1);
+
+            // Sleep for an amount of time greater than the expiry duration so
+            // that the background purge would purge the cache iff values can
+            // be expired.
+            runtime
+                .block_on(tokio_timer::sleep(Duration::from_millis(100)))
+                .unwrap();
+
+            // Explicity drop both the access and cache handles. Dropping the
```
hawkw: Nit: I wouldn't say "Explicitly" here unless you actually call `drop(cache_1)` and so on...

src/proxy/http/profiles.rs
```diff
@@ -477,13 +478,15 @@ pub mod router {
...
-        T::Output: Eq + Hash,
+        T::Output: Clone + Eq + Hash,
         R: rt::Make<T::Output> + Clone,
         R::Value: svc::Service<http::Request<B>> + Clone,
     {
         fn update_routes(&mut self, routes: Routes) {
             let slots = routes.len() + 1;
-            self.router = Router::new(
+
+            // We never want to purge routes from the Profile's router cache
```
hawkw: Nit:
```suggestion
            // We never want to purge routes from the profiles router cache
```

src/proxy/http/profiles.rs
```diff
@@ -416,7 +416,8 @@ pub mod router {
             let inner = self.inner.make(&target);
             let stack = self.route_layer.clone().service(svc::shared(inner));
 
-            let router = Router::new(
+            // We never want to purge routes from the Profile's router cache
```
hawkw: Nit:
```suggestion
            // We never want to purge routes from the profiles router cache
```

src/proxy/http/profiles.rs
```diff
@@ -416,7 +416,8 @@ pub mod router {
...
+            let (router, _cache_bg) = Router::new(
```
hawkw: Nit: I think this can just be 
```suggestion
            let (router, _) = Router::new(
```
which will immediately drop the background future

lib/router/src/cache.rs
kleimkuhler: No not nitpicky all all! I intended to add a better description here than just keeping the explanation as a comment.

Something this has me thinking about though is we can make a `Cache` with `capacity = 0`. If that was the case, we could hit this branch because `vals.len() == capacity` and `expirations.poll()` always return `Async::Ready`.

I'm not sure if we want to prevent a cache capacity of zero, so maybe we should handle this?

lib/router/src/cache.rs
kleimkuhler: Yea good call. I think depending on what I just mentioned above about `Async::Ready(None)`, we may want to be able to handle errors other than `CapacityExhausted`.

lib/router/src/cache.rs
hawkw: Ahh, hm. I suppose we should probably put a
```rust
if capacity == 0 {
    return Err(CapacityExhausted {
        capacity: self.capacity,
    });                   
}
```

at the top of this function?

lib/router/src/cache.rs
kleimkuhler: We probably don't want to `println!` here in the first place.

lib/router/src/cache.rs
kleimkuhler: I'd like to include this in linkerd/linkerd2#2802

lib/router/src/cache.rs
hawkw: we should probably log with `error!(...)` or `warn!(...)` instead

kleimkuhler:
@hawkw As we discussed, your comments led to me realizing some these tests should have been panicking when instead they were not. This was because they were not being run in a tokio runtime.

I've fixed them all to do that as well as switched the ones that need to block to all use `block_on`.

lib/router/src/cache.rs
```diff
@@ -108,11 +115,7 @@ impl<K: Clone + Eq + Hash, V> Cache<K, V> {
                     });
                 }
 
-                Err(_e) => {
-                    return Err(CapacityExhausted {
-                        capacity: self.capacity,
-                    });
-                }
+                Err(e) => panic!("Cache.expirations DelayQueue::poll must not fail: {}", e),
```
hawkw: TIOLI: this could also be an `expect` on `self.expirations.poll`, rather than an explicit case. Not sure if it's any clearer that way.

kleimkuhler: @hawkw I lean towards keeping the error case explicit here. It's how we seem to handle a lot of `Err(...)` cases in `match` and I think it's a little more clear this way!

lib/router/src/cache.rs
```diff
@@ -1,426 +1,402 @@
...
+            Ok(lock) => lock,
+
+            // If we were unable to lock the cache, panic if the cause of error
+            // was a poisoned lock
+            Err(TryLockError::Poisoned(e)) => panic!("lock poisoned: {:?}", e),
+
+            // If the lock is not poisoned, it is being held by another
+            // thread. Schedule this thread to be polled in the near future.
+            Err(_) => {
+                futures::task::current().notify();
```
seanmonstar: If you'd like to use a lock that doesn't block futures, consider using [`tokio::sync::lock::Lock`](https://docs.rs/tokio/0.1.*/tokio/sync/lock/struct.Lock.html) instead. Otherwise, this could cause quite a few spins.

lib/router/src/cache.rs
```diff
@@ -1,426 +1,402 @@
...
+                capacity: self.capacity,
             });
+        }
 
-            if self.vals.len() == self.capacity {
-                return Err(CapacityExhausted {
-                    capacity: self.capacity,
-                });
+        if self.vals.len() == self.capacity {
+            match self.expirations.poll() {
```
seanmonstar: The `reserve` method doesn't return a `Poll`, so I assumed it was a method that could be called from outside a task. However, calling `poll()` here will trigger a panic if so.

lib/router/src/cache.rs
hawkw: I forgot this had merged! :tada:

src/proxy/http/router.rs
```diff
@@ -42,8 +42,9 @@ pub struct Stack<Req, Rec: Recognize<Req>, Mk> {
 
 pub struct Service<Req, Rec, Mk>
 where
-    Rec: Recognize<Req>,
-    Mk: rt::Make<Rec::Target>,
+    Rec: Recognize<Req> + Clone + Send,
+    <Rec as Recognize<Req>>::Target: Clone,
```
olix0r: so, it seems that all uses of Recognize requie that `Target: Clone`... You could probably reduce boilerplate by moving this contraint onto the trait definition?

kleimkuhler:
In addressing @seanmonstar comment about using `tokio::sync::lock::Lock`, I ended up making some significant changes compared to the original implementation.

The core difference is that locking the router cache has moved into the `impl Future for ResponseFuture` [here](https://github.com/linkerd/linkerd2-proxy/pull/247/files#diff-4b1c749938240031b52806d3cf04f838L225).

The states of `ResponseFuture` are now:
* `Init`: The initial state where the router cache is locked
* `Acquired`: We have acquired the lock to the router cache
* `Call`: We are preparing to route the request to the service
* `Called`: We have routed the request
* `Error`: There has been an error in polling `ResponseFuture`

In summary, the router cache lock acquisition has been moved into the tokio runtime and should result in less spins and lock contention (when we someday default to a multithreaded runtime).

lib/router/src/cache.rs
```diff
@@ -1,426 +1,476 @@
...
     }
 
-    /// Ensures that there is capacity to store an additional route.
-    ///
-    /// Returns a handle that may be used to store an ite,. If there is no available
-    /// capacity, idle entries may be evicted to create capacity.
-    ///
-    /// An error is returned if there is no available capacity.
-    pub fn reserve(&mut self) -> Result<Reserve<K, V, N>, CapacityExhausted> {
+    pub fn reserve(&mut self) -> Async<Reserve<K, V>> {
```
hawkw: What if we changed this from a function returning `Async` to a function returning a `Future`, that returns `Ok(Async::Ready(Reserve<K, V>))` when we're able to reserve an entry? I think that could move some logic out of the complex state machine in `Router::call`

lib/router/src/lib.rs
```diff
@@ -208,43 +260,113 @@ where
...
+                                // Make a new service for the route
+                                let make = state.make.take().expect("polled after ready");
+                                let service = LoadShed::new(make.make(&target));
+
+                                // Cache the service and send the request on the route
+                                reserve.store(target, service.clone());
+                                State::Call(Some(request), Some(service))
+                            }
+                            Async::NotReady => {
+                                State::Error(Some(error::NoCapacity(capacity).into()))
```
hawkw: I'm not sure if this should fail when there's no capacity. Instead, I think the future should return `NotReady` here, since additional capacity might be released, and enter a state where we keep trying to reserve a slot in the cache. If we spend too long trying to reserve a router slot, the request will still eventually get killed by the dispatch timeout, so this won't hang forever.

lib/router/src/cache.rs
```diff
@@ -1,426 +1,476 @@
...
         if self.vals.len() == self.capacity {
```
olix0r: Does this mean that we only try to evict items when the router is at capacity?

kleimkuhler: Yes, but this only applies when trying to reserve a slot. The eager eviction comes from spawning the background task that will poll the DelayQueue without checking current length first. If we don’t spawn the background task, then the only way values are dropped is right here.

lib/router/src/cache.rs
```diff
@@ -1,426 +1,476 @@
...
+        (cache, bg_purge)
+    }
+
+    pub fn access(&mut self, key: &K) -> Option<Access<K, V>> {
+        let node = self.vals.get_mut(key)?;
+        Some(Access {
+            expires: self.expires,
+            expirations: &mut self.expirations,
+            node,
+        })
```
olix0r: I think this can all be quite a bit simpler...

As it is, `lib.rs` calls this and then almost immediately drops the access, cloning the value out of the cache.  Can we get rid of all of the borrowing semantics (and the Access type entirely) and instead move the expiration-reset logic into this function? I think this would eliminate the needed for `State::Acquired`...

lib/router/src/lib.rs
```diff
@@ -208,43 +260,113 @@ where
...
         loop {
             self.state = match self.state {
-                State::Init(ref mut req, ref mut svc) => {
-                    match svc.poll_ready().map_err(error::Error::from)? {
-                        Async::NotReady => {
-                            unreachable!("load shedding services must always be ready");
+                State::Init(ref mut state) => {
+                    let request = state.request.take().expect("polled after ready");
+                    let target = state.target.take().expect("polled after ready");
+                    let make = state.make.take().expect("polled after ready");
```
olix0r: This is a bug ;)  If the lock is not ready, you'll have taken these values and they won't be available on the next invocation!  This should happen after you've acquired the cache.

olix0r:
Also, be sure to merge master into this when you get a chance. It merges cleanly.

lib/router/src/lib.rs
kleimkuhler: Ah good call, thanks! I have a few changes to roll up so I’ll make sure to fix this.

lib/router/src/lib.rs
hawkw: As per our offline conversation w/ @olix0r, we've determined that failing fast when there's no capacity is the right thing to do here. Commenting just so that it's recorded for posterity.

lib/router/src/cache.rs
hawkw: Since we want to fail the request when capacity is exhausted (see https://github.com/linkerd/linkerd2-proxy/pull/247#discussion_r293508066), this should probably return `Result<Reserve<...>, CapacityExhausted>` rather than `Async`, as it did previously.

hawkw:
Looks like the build failure is just due to [DCOBot seeing commits signed with the wrong email](https://github.com/linkerd/linkerd2-proxy/runs/148076222).

kleimkuhler:
> Looks like the build failure is just due to [DCOBot seeing commits signed with the wrong email](https://github.com/linkerd/linkerd2-proxy/runs/148076222).

Should be fixed now!

lib/router/src/cache.rs
```diff
@@ -1,426 +1,397 @@
...
+            let dq_key = self.expirations.insert(key.clone(), self.expires);
+            Node::new(dq_key, value)
+        };
+
+        self.values.insert(key, node)
     }
-}
 
-// ===== impl Reserve =====
+    pub fn poll_insert(&mut self) -> Poll<(), ()> {
```
hawkw: I don't think it makes a lot of sense for this to return a `Poll`. Looking at how it's used, it looks like we `expect` that it will always be `Ok`, and return a `NoCapacity` error if it's `NotReady`: 
https://github.com/linkerd/linkerd2-proxy/blob/39f00f9753b8c1affb070cf1a8cc8ef24b2ae3af/lib/router/src/lib.rs#L283-L295

This seems unnecessarily complex to me. In particular, I notice that this method will never return `Err` --- we panic if polling the `DelayQueue` failed, and there's no other case where we would return `Err`, so wrapping this in a `Result` that we expect will never be `Err` is unnecessary. In the router code where we call this method, rather than yielding when we see a `NotReady`, we return an error, so perhaps _that_ case ought to be represented using `Err`.

I think we should consider changing this from `poll_insert` to something like
```rust
pub fn can_insert(&mut self) -> Result<(), NoCapacity>
```
If this had that signature, the case in `ResponseFuture::poll` could just look like this:
```rust
if let Some(service) = cache.get(&target) {
    State::Calling(Some(request), Some(service))
} else {
    // Ensure that there is capacity for a new slot
    cache.can_insert()?;
    // Make a new service for the target
    let make = rr.make.take().expect("polled after ready");
    let service = LoadShed::new(make.make(&target));

    cache.insert(target, service.clone());
    State::Calling(Some(request), Some(service))
}
```

lib/router/src/cache.rs
```diff
@@ -1,426 +1,397 @@
...
 impl<T> Node<T> {
-    pub fn new(value: T, last_access: Instant) -> Self {
-        Node { value, last_access }
+    pub fn new(dq_key: delay_queue::Key, value: T) -> Self {
+        Node { dq_key, value }
     }
 
-    pub fn access<'a, N: Now + 'a>(&'a mut self, now: &'a N) -> Access<'a, T, N> {
-        Access { now, node: self }
+    pub fn dq_key_ref(&self) -> &delay_queue::Key {
```
hawkw: I don't really get why this needs to be `pub`. AFAICT, it's not used anywhere outside of the cache module. Actually, now that I think of it, it looks like the `Node` type need not be `pub`?

lib/router/src/lib.rs
```diff
@@ -208,43 +247,93 @@ where
...
+                                // Make a new service for the target
+                                let make = rr.make.take().expect("polled after ready");
+                                let service = LoadShed::new(make.make(&target));
+
+                                cache.insert(target, service.clone());
+                                State::Calling(Some(request), Some(service))
+                            }
+                            Async::NotReady => {
+                                let capacity = cache.capacity();
+                                State::Error(Some(error::NoCapacity(capacity).into()))
```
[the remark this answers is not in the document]
hawkw: I think the purpose of `State::Error` is just so that the router can return a future that's already failed. Looking at the previous code, when the future fails from another state, it looks like we just return the error immediately, rather than advancing to the `Error` state. That might simplify this code a bit.

lib/router/src/lib.rs
```diff
@@ -208,43 +247,93 @@ where
...
+                            let request = request.take().expect("polled after ready");
+                            State::Called(service.call(request))
                         }
-                        Async::Ready(()) => {
-                            let req = req.take().expect("polled after ready");
-                            State::Called(svc.call(req))
+                        Ok(Async::NotReady) => {
+                            unreachable!("load shedding services must always be ready");
                         }
+                        Err(err) => State::Error(Some(err)),
```
hawkw: Similarly, if we just returned the error directly here, I think we could say `match service.poll_ready()?` above, avoiding this match arm...

lib/router/src/cache.rs
kleimkuhler: I'll look into your suggestion and see how it changes things, but before doing so I'll leave a quick explanation on why I ultimately thought `Poll` was a good choice.

In a previous state of this PR, `poll_insert()` (basically what `reserve()` was) did return a `Result`. However Sean pointed out [here](https://github.com/linkerd/linkerd2-proxy/pull/247#discussion_r284857061) that it internally calls `poll()` and could therefore panic if called outside a task. I felt like that was an important point, and it still holds true even though a lot else has changed.

I know the cases of this being used elsewhere where the implementer doesn't know to only call it within a task, but maybe something to keep in mind.

lib/router/src/cache.rs
kleimkuhler: Yep I don't think these even needed to be public in the original implementation. I'll remove them in my next set of changes.

kleimkuhler: Well `Cache::insert` does return an `Option<Node<V>>` so it would need to be pub for that. Field access through the impl doesn't need to be pub for how it is used right now, but maybe something consider in the future? I was basing some of that off [delay_queue::Expired](https://docs.rs/tokio-timer/0.2.10/src/tokio_timer/delay_queue.rs.html#830-845)

lib/router/src/cache.rs
hawkw: > However Sean pointed out [here](https://github.com/linkerd/linkerd2-proxy/pull/247#discussion_r284857061) that it internally calls `poll()` and could therefore panic if called outside a task. I felt like that was an important point, and it still holds true even though a lot else has changed.

That's a good point from @seanmonstar that I hadn't thought of. I think there's some tension here between indicating that the function should not be called outside of a task context, and simplifying the signature of this function. It definitely seems strange to have a function called `poll_XXX` that returns a `Result`, so the return type of `Poll` makes sense in that context, but it also seems to introduce some complexity downstream that's otherwise unnecessary.

I think either way, there should be a doc comment stating that this shouldn't be called outside a task context, whether it remains `poll_insert` or not.

lib/router/src/cache.rs
hawkw: Ah, I missed that `Cache::insert` returned a `Node`. Is it necessary for it to return `Option<Node<V>>` rather than `Option<V>`? As far as I can tell, the `Node` is never actually used. Can we just map over the removed node and extract the `V`?

lib/router/src/cache.rs
kleimkuhler: > I think what we should probably do is leave the function poll_insert but change it to return Poll<(), NoCapacity>. 

Yep I like that approach. It will be nice to use the `?` operator downstream, but keep a signature that indicates this should be called within a task.

lib/router/src/cache.rs
kleimkuhler: Yep `Node` is an implementation detail that is only used in _cache.rs_. It makes sense to just return the value instead of keeping it coupled with the `delay_queue::Key`

lib/router/src/lib.rs
```diff
@@ -208,43 +247,80 @@ where
...
+                    let request = rr.request.take().expect("polled after ready");
+                    let target = rr.target.take().expect("polled after ready");
+
+                    // If the target is already cached, route the request to
+                    // the service; otherwise, try to insert it
+                    if let Some(service) = cache.get(&target) {
+                        State::Calling(Some(request), Some(service))
+                    } else {
+                        // Ensure that there is capacity for a new slot
+                        cache.poll_insert()?;
```
olix0r: What if this returns NotReady?

lib/router/src/lib.rs
```diff
@@ -208,43 +247,80 @@ where
...
+                        let service = LoadShed::new(make.make(&target));
+
+                        cache.insert(target, service.clone());
+                        State::Calling(Some(request), Some(service))
                     }
                 }
-                State::Called(ref mut fut) => return fut.poll().map_err(error::Error::from),
+                State::Calling(ref mut request, ref mut service) => {
+                    let mut service = service.take().expect("polled after ready");
+                    service.poll_ready()?;
```
olix0r: What if this returns NotReady?

lib/router/src/lib.rs
olix0r: It appears that poll_insert can't actually return `NotReady`... But the type system should actually reflect the possible outcomes (this is more important than the naming lining up, imo)...

I would suggest changing poll_insert to return a `Poll<(), Never>` and, instead, handling the not-ready->Error case in the ResponseFuture impl

hawkw: My [initial suggestion](https://github.com/linkerd/linkerd2-proxy/pull/247#discussion_r293576909) was that `poll_insert` should be renamed to `can_insert` (or something) and return `Result` rather than `Poll`, since (as written) it will never be `NotReady`. 

@kleimkuhler [pointed out](https://github.com/linkerd/linkerd2-proxy/pull/247) that @seanmonstar suggested _changing_ this function to be named `poll_insert` and return `Poll`, since it polls the timer internally and will thus fail if called outside of a task context, so the current approach is a bit of a compromise between our suggestions. 

IMO, I still think this should just return `Result` --- I'd rather not have the type signature include a case that will never actually be returned. But, I think that at this point, @olix0r, @seanmonstar, and I should probably try to come to an agreement on what the right thing to do here is, so that we can stop telling @kleimkuhler to change his code one way and then change it back...

lib/router/src/lib.rs
```diff
@@ -105,37 +126,58 @@ impl<Req, Rec, Mk> Router<Req, Rec, Mk>
...
+                },
+            },
+            cache_bg,
+        )
+    }
+
+    pub fn new_lazy(recognize: Rec, make: Mk, capacity: usize, max_idle_age: Duration) -> Self {
+        let (cache, _) = Cache::new(capacity, max_idle_age);
+
+        (Router {
```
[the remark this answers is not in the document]
hawkw: Are the parens here because this returned a tuple at one point?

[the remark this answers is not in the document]
kleimkuhler: Yep that would be the reason. I used `new()` which does require parens because of the return type.

lib/router/src/lib.rs
kleimkuhler: We know that `service` will resolve to a `LoadShed` which should always be ready. The [current implementation](https://github.com/linkerd/linkerd2-proxy/blob/35df8ab451b1dd6f7c8999e10d0e5928ddcc6231/lib/router/src/lib.rs#L231) is explicit about that and I probably should have kept that in there. I've added it back in

lib/router/src/lib.rs
kleimkuhler: I think we are approaching the right decision. I feel it should return a `Poll` based off @seanmonstar comments, but I also like @olix0r's suggestion about a `Poll<(), Never>`. I have that change in there, and `NoCapacity` is constructed at the router now instead of the cache.

If `Result` does make sense after further discussion I can change it back! I don't mind the back and forth; good to see the reasons for each.

lib/router/src/cache.rs
```diff
@@ -1,426 +1,461 @@
...
 
-    /// Overrides the time source for tests.
-    #[cfg(test)]
-    fn with_clock<M: Now>(self, now: M) -> Cache<K, V, M> {
-        Cache {
-            now,
-            vals: self.vals,
-            capacity: self.capacity,
-            max_idle_age: self.max_idle_age,
+    pub fn get(&mut self, key: &K) -> Option<V> {
```
olix0r: tioli: I prefer `access` to get here, since it's `&mut self` --- `get`s are typically `&self`; we actually mutate state, so it would be clearest to use a verb that makes that clearer.

Also, would be nice to document the semantics of this function.... Something like:

```
/// Attempts to access an item by key.
///
/// If a value is returned, this key will not be considered for eviction for... (how long?)
```

lib/router/src/cache.rs
```diff
@@ -1,426 +1,461 @@
...
-    pub fn new(value: T, last_access: Instant) -> Self {
-        Node { value, last_access }
+        Ok(Async::Ready(()))
     }
 
-    pub fn access<'a, N: Now + 'a>(&'a mut self, now: &'a N) -> Access<'a, T, N> {
-        Access { now, node: self }
-    }
+    fn poll_purge(&mut self) -> Poll<(), Error> {
+        while let Some(expired) = try_ready!(self.expirations.poll()) {
```
olix0r: Hmm.. so, I'm having a little trouble trying to understand the semantics of `expirations` -- it seems that it is polled by two tasks, which seems unsafe.

Specifically, I worry that a `poll_insert` can prevent a `poll_purge` from being notified...

Why does `expirations` need to be polled in `poll_insert`?

kleimkuhler: If we choose not to spawn `cache_bg`, then we have the current behavior of no values being removed from the cache. In these cases, `poll_insert` will at some point [branch](https://github.com/linkerd/linkerd2-proxy/pull/247/files/ccb320e266c984268c22eae14b2358e66765f78c#diff-230615b3316b2a089b0d6563c3f74944R109) into the cache being at capacity.

In order to make capacity, we can check if any values should be removed and therefore need to `poll` `expirations`.

This is the reason that there is a lock around `Cache`. The `Future:poll` impl for `PurgeCache` will not `poll_purge` until it has acquired a lock for the cache.

lib/router/src/cache.rs
```diff
@@ -1,426 +1,461 @@
...
-    #[cfg(test)]
-    fn with_clock<M: Now>(self, now: M) -> Cache<K, V, M> {
-        Cache {
-            now,
-            vals: self.vals,
-            capacity: self.capacity,
-            max_idle_age: self.max_idle_age,
+    pub fn get(&mut self, key: &K) -> Option<V> {
+        if let Some(node) = self.values.get_mut(key) {
+            self.expirations.reset(&node.dq_key, self.expires);
```
olix0r: We should validate that it's safe to call `expirations.reset` outside the context of the polling task... I expect that it is, given its naming; but worth double checking...

lib/router/src/lib.rs
```diff
@@ -208,43 +246,86 @@ where
...
+                    let request = rr.request.take().expect("polled after ready");
+                    let target = rr.target.take().expect("polled after ready");
+
+                    // If the target is already cached, route the request to
+                    // the service; otherwise, try to insert it
+                    if let Some(service) = cache.access(&target) {
+                        State::Calling(Some(request), Some(service))
+                    } else {
+                        // Ensure that there is capacity for a new slot
+                        cache.at_capacity()?;
```
olix0r: I would expect something called at_capacity/is_at_capacity to return only a boolean indicating whether there is capacity in the cache; and I still think it would be clearer if the error logic was handled here rather than in the cache.

What do you think about changing the cache to export `pub fn can_insert(&self) -> bool`; and then having this code just be `if !cache.can_insert() { return Err(NoCapacity) }` -- I don't think there's any value to returning `self.capacity`...

lib/router/src/lib.rs
```diff
@@ -208,43 +246,86 @@ where
...
+                        let service = LoadShed::new(make.make(&target));
+
+                        cache.insert(target, service.clone());
+                        State::Calling(Some(request), Some(service))
+                    }
+                }
+                State::Calling(ref mut request, ref mut service) => {
+                    let mut service = service.take().expect("polled after ready");
+
+                    match service.poll_ready().map_err(error::Error::from)? {
```
olix0r: tioli: slightly clearer to only handle the exceptional case in a block; also a prime case for `assert!`:

```rust
assert!(service.poll_ready()?.is_ready(), "...");

let request = request.take().expect("polled after ready");
State::Called(service.call(request))
```

lib/router/src/lib.rs
hawkw: My argument was that having this method return `Result` made the callsite simpler, since it wasn't responsible for constructing the error, which feels like the cache's responsibility to me. In that case, though, I wouldn't call it `at_capacity` --- `try_reserve` or something would make sense for a method returning `Result`.

Regardless, I'm fine with @olix0r's suggestion instead; this isn't a big deal to me.

olix0r: This would be fine if called `try_reserve`, imo, as well... I could go either way on this.

lib/router/src/lib.rs
```diff
@@ -72,30 +82,63 @@ where
 {
     recognize: Rec,
     make: Mk,
-    cache: Mutex<Cache<Rec::Target, LoadShed<Mk::Value>>>,
+    cache: Lock<Cache<Rec::Target, LoadShed<Mk::Value>>>,
 }
 
-enum State<Req, Svc>
+struct RouteRequest<Req, Rec, Mk>
```
olix0r: nit: I'd be inclined to embed this into the enum directly to reduce boilerplate... e.g.

```rust
enum State<Req, Rec, Mk>
where ...
{
    Inserting {
        request: Option<Req>,
        target: Option<Rec::Target>,
        make: Option<Mk>,
        cache: Lock<Cache<Rec::Target, LoadShed<Mk::Value>>>
    }
    ...
}
```

lib/router/src/lib.rs
```diff
@@ -105,37 +148,62 @@ impl<Req, Rec, Mk> Router<Req, Rec, Mk>
...
 
-impl<Req, Rec, Mk, Svc> svc::Service<Req> for Router<Req, Rec, Mk>
+impl<Req, Rec, Svc> Router<Req, Rec, FixedMake<Rec::Target, Svc>>
 where
     Rec: Recognize<Req>,
-    Mk: Make<Rec::Target, Value = Svc>,
     Svc: svc::Service<Req> + Clone,
-    Svc::Error: Into<error::Error>,
+{
+    pub fn new_fixed(recognize: Rec, routes: IndexMap<Rec::Target, Svc>) -> Self {
```
olix0r: This could use a docstring, at least documenting the assumption that `recognize` only produce targets that exist in `routes`.

lib/router/src/lib.rs
```diff
@@ -105,37 +148,62 @@ impl<Req, Rec, Mk> Router<Req, Rec, Mk>
...
+        let capacity = routes.len();
+        let (router, _) = Self::new(
+            recognize,
+            FixedMake(routes),
+            capacity,
+            Duration::from_secs(1),
```
olix0r: Probably worth noting in a comment that the drop of the daemon task ensures that eviction is not performed and that the max_idle_age is ignored, etc.

lib/router/src/lib.rs
```diff
@@ -72,30 +82,63 @@ where
...
+}
+
+enum State<Req, Rec, Mk>
+where
+    Rec: Recognize<Req>,
+    Mk: Make<Rec::Target>,
+    Mk::Value: svc::Service<Req>,
+    <Mk::Value as svc::Service<Req>>::Error: Into<error::Error>,
+{
+    Inserting(RouteRequest<Req, Rec, Mk>),
```
olix0r: Is `Inserting` the best name for this still? We are not necessarily inserting anything in this state...

Perhaps rename these enum states as `Acquire/Call/Respond`?

lib/router/src/lib.rs
hawkw: +1, there should definitely be a comment explaining what this does.

lib/router/src/lib.rs
```diff
@@ -149,34 +217,19 @@ where
...
-                return ResponseFuture::no_capacity(capacity);
-            }
-        };
-
-        // Bind a new route, send the request on the route, and cache the route.
-        let service = LoadShed::new(self.inner.make.make(&target));
-
-        reserve.store(target, service.clone());
-        ResponseFuture::new(request, service)
+        return ResponseFuture::new(
```
hawkw: I think the return here is not necessary.
```suggestion
       ResponseFuture::new(
```

lib/router/src/lib.rs
```diff
@@ -72,30 +82,63 @@ where
...
+    Rec: Recognize<Req>,
+    Mk: Make<Rec::Target>,
+    Mk::Value: svc::Service<Req>,
+    <Mk::Value as svc::Service<Req>>::Error: Into<error::Error>,
 {
-    Init(Option<Req>, Svc),
-    Called(Svc::Future),
+    request: Option<Req>,
+    target: Option<Rec::Target>,
+    make: Option<Mk>,
```
hawkw: TIOLI: this is a _bit_ awkward, maybe putting all the things that get `take`n in a `struct` so you can have one `Option` you take in one place would be nicer? Or wrapping this whole struct in an `Option` in the State` enum? Not a big deal.

olix0r: fwiw, this was my suggestion and i think it's fine for now...

lib/router/src/cache.rs
```diff
@@ -1,426 +1,306 @@
...
+    K: Clone + Eq + Hash,
+{
+    cache: Lock<Cache<K, V>>,
+    interval: Interval,
 }
 
-#[derive(Clone, Debug, PartialEq)]
-pub struct CapacityExhausted {
-    pub capacity: usize,
+/// A handle to a cache value.
```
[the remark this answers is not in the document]
kleimkuhler: I don't think I had the Wikipedia definition in mind when describing it, but I use "handle" because it is essentially a wrapper around the actual value so that `dq_key` can be used to do an `expirations` lookup during resets. It's also a private struct so I think handle is fine in this case!

lib/router/src/lib.rs
kleimkuhler: There is also the case where we don’t want to take the make if we don’t need to. If the route is already in the cache, we don’t end up using it that field.

olix0r:
I'm trying to devise a manual test plan for this...

Essentially, we need to have a client that sends a request to a service and then sleeps for twice the max idle age before repeating the process. We'd want to manually confirm (probably via logs, for now?) that we see the service dropped and re-inserted upon the next request...

This also raises some follow-up issues we should open:
* How should the router's state be reflected in metrics? I expect we want counters for each (expiring) router's inserts/evictions/rejections... What else?
* What is a sensible default max-idle-age? I expect whatever value we have now is too high...

lib/router/src/lib.rs
hawkw: Okay, that's fair. Carry on.

hawkw:
It looks like your [most recent build](https://travis-ci.org/linkerd/linkerd2-proxy/builds/548260027?utm_source=github_status&utm_medium=notification) failed due to rustfmt reasons.

lib/router/src/lib.rs
```diff
@@ -102,37 +135,68 @@ impl<Req, Rec, Mk> Router<Req, Rec, Mk>
...
+impl<Req, Rec, Svc> Router<Req, Rec, FixedMake<Rec::Target, Svc>>
 where
     Rec: Recognize<Req>,
-    Mk: Make<Rec::Target, Value = Svc>,
     Svc: svc::Service<Req> + Clone,
-    Svc::Error: Into<error::Error>,
+{
+    /// A router that is created with a fixed set of known routes.
+    ///
+    /// `recognize` will only produce targets that exist in `routes`. We never
```
hawkw: Nit: I might change this to say
```suggestion
    /// `recognize` is expected to only produce targets that exist in `routes`. We never
```
we can't actually _enforce_ this; it's a contract that we expect the caller to adhere to.

lib/router/src/lib.rs
```diff
@@ -102,37 +135,68 @@ impl<Req, Rec, Mk> Router<Req, Rec, Mk>
...
+    /// want to expire routes from the fixed set; the explicit drop of the
```
hawkw: Nit/TIOLI: I might also rephrase this to focus on the _behaviour_ of the fixed router --- something like
"Because the set of routes will not change, routes will not be expired from the router cache, and no background task is necessary to purge it." --- I think Oliver's suggestion that we comment about the explicit drop of the purge task is more of an implementation detail that I'd put in the function body.

This is a nit & non-blocking, I think the comment is okay as-is.

src/proxy/http/router.rs
```diff
@@ -114,27 +114,31 @@ where
...
 {
     pub fn make(&self) -> Service<Req, Rec, Mk> {
-        let inner = Router::new(
+        let (inner, cache_bg) = Router::new(
             self.recognize.clone(),
             self.inner.clone(),
             self.config.capacity,
             self.config.max_idle_age,
         );
+        tokio::spawn(cache_bg);
```
hawkw: It looks like the purge task doesn't currently log anything, but if it did, we _might_ want to wrap this bg task in a log context, so that we would know which cache's purge task any log messages came from.

Not a suggestion, since we're not currently logging from that task; just a note for the future.

src/proxy/http/profiles.rs
```diff
@@ -477,23 +479,35 @@ pub mod router {
...
         fn update_routes(&mut self, routes: Routes) {
-            let slots = routes.len() + 1;
-            self.router = Router::new(
+            let default_route = self.target.clone().with_route(self.default_route.clone());
+
+            let capacity = routes.len() + 1;
+            let mut make = IndexMap::with_capacity(capacity);
+            make.insert(default_route.clone(), self.stack.make(&default_route));
+
+            for (_, route) in routes.clone().into_iter() {
```
hawkw: Not a big deal, but I would probably change this to
```rust
for (_, route) in &routes {
    let route = self.target.clone().with_route(route.clone());
    // ...
}
```
What you have currently is cloning the entire `IndexMap`, implying a heap allocation (and cloning the map's keys as well as the values, although the cloned keys are never used). On the other hand, iterating by reference and cloning each `route` will clone just the `Route`, which AFAIK is trivial to clone.

I dunno how hot this function is, so it may not be a big deal.

kleimkuhler:
@olix0r @hawkw Above you will see the change I had to make to ensure the cache purge task continues to poll in the background.

I confirmed services are being purged from the router cache as we would expect with this change. I setup an environment based off slow-cooker. There is a client that makes requests a single curl request every 4 seconds. I set the `LINKERD2_PROXY_OUTBOUND_ROUTER_MAX_IDLE_AGE` of the injected slow-cooker pod to `2s` so that services would be expired after being idle for 2 seconds.

The following abbreviated logs show the repeated scenario:
```
DBUG [    28.904693s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_proxy::app::main outbound addr=Ok(Name(NameAddr { name: "bb-p2p", port: 7070 }))
DBUG [    28.904742s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router aquiring lock for the router cache...
DBUG [    28.904747s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router aquired lock for the router cache!
DBUG [    28.904751s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router current cache capacity: 10000; length: 0
DBUG [    28.904756s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router::cache accessing cache...
DBUG [    28.904761s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router target NOT cached!
DBUG [    28.904775s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router inserting new target into cache...
DBUG [    28.904782s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router::cache inserting into cache...
DBUG [    28.904785s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router::cache inserting into: expirations; expires: 2s
DBUG [    28.904794s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router::cache inserting into: values
DBUG [    28.904805s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router insert new target; cache length: 1
...
DBUG [    28.905493s] bb-p2p:7070 linkerd2_proxy::app::main outbound dst=Some(DstAddr { addr: Name(NameAddr { name: "bb-p2p.default.svc.cluster.local.", port: 7070 }), direction: Out, http_settings: Http1 { keep_alive: true, wants_h1_upgrade: false, was_absolute_form: false } })
DBUG [    28.905539s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router aquiring lock for the router cache...
DBUG [    28.905547s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router aquired lock for the router cache!
DBUG [    28.905552s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router current cache capacity: 10000; length: 0
DBUG [    28.905556s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router::cache accessing cache...
DBUG [    28.905561s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router target NOT cached!
DBUG [    28.905580s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router inserting new target into cache...
DBUG [    28.905584s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router::cache inserting into cache...
DBUG [    28.905588s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router::cache inserting into: expirations; expires: 2s
DBUG [    28.905596s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router::cache inserting into: values
DBUG [    28.905604s] proxy={server=out listen=127.0.0.1:4140 remote=10.48.0.44:44200} linkerd2_router insert new target; cache length: 1
```

[Here](https://gist.github.com/kleimkuhler/93bd3fc3d4fdf8c3e65f43649a55668b) is the gist with the deployment I used. To replicate you'll need to:
1. `linkerd install --proxy-version "router-cache-purge-8" --proxy-log-level "warn,linkerd2_proxy=debug,linkerd2_router=trace" | kubectl apply -f -`
2. `cat purge-deployment.yml | linkerd inject - | kubectl apply -f -`
3. `kubectl edit` the purge deployment and add:
```
- name: LINKERD2_PROXY_OUTBOUND_ROUTER_MAX_IDLE_AGE
    value: 2s
```

After that you can `kubectl logs -f slow-cooker-... linkerd-proxy` and observe [these](https://github.com/linkerd/linkerd2-proxy/blob/8928fdab7868d7c1700dc81923957becbca96ab1/lib/router/src/cache.rs#L128-L129) trace log lines.

kleimkuhler:
I also have this proxy image continuing to run this purge deployment on GKE `kleimkuhler-dev-0` and lifecycle on GKE `kleimkuhler-dev-1`. I'll check those in the morning for any interesting behavior.

lib/router/src/cache.rs
```diff
@@ -1,426 +1,322 @@
...
+            }));
+            trace!("purge cache interval reached");
+
+            let mut cache = try_ready!(Ok(self.cache.poll_lock()));
+
+            trace!("purge cache is polling for evictions...");
+            match cache.poll_purge() {
+                Ok(Async::NotReady) => return Ok(Async::NotReady),
+                Ok(Async::Ready(())) => (),
+                Err(e) => {
```
olix0r: If you make the error type of `poll_purge` `Never`, you can avoid the panic by matching on Never...

kleimkuhler: I'm inclined to get rid of the match statement since we don't actually need it and change it to `cache.poll_purge().expect("Cache::poll_purge must not fail");`

kleimkuhler: Plus can't `poll_purge` can error if `expirations.poll()` errors?

hawkw: I think it's reasonable to move the `expect` into `poll_purge`, and have it return `Poll<(), Never>`. Not a blocker IMO.

src/proxy/http/router.rs
```diff
@@ -114,27 +115,34 @@ where
...
     pub fn make(&self) -> Service<Req, Rec, Mk> {
-        let inner = Router::new(
+        let (inner, cache_bg) = Router::new(
             self.recognize.clone(),
             self.inner.clone(),
             self.config.capacity,
             self.config.max_idle_age,
         );
+
+        let ctx = logging::Section::Proxy.bg(self.config.proxy_name);
```
hawkw: Nit: this log context will format like this:
`proxy={bg=foo_router}`

I think it might be better to wrap the `proxy_name` field in a type that formats like
```
"cache_purge router={}", proxy_name
```
so that the resulting log lines look like
```
proxy={bg=cache_purge router=foo_router}
```
or something. This might be useful in case there are ever other background contexts associated with routers.

On the other hand, we can always change this later.

kleimkuhler: There is definitely some follow up around this change regarding metrics around the router. I will probably address this then when I do that. These would also be at the `trace` level so I don't think it's too big an issue right now as searching off the `trace` messages is fairly helpful.

lib/router/src/cache.rs
kleimkuhler: This introduces some further complexity into `poll_purge` as the `try_ready!` has to change to a match. I'm going to punt on this for now and consider in the follow-up.

lib/router/src/cache.rs
```diff
@@ -1,426 +1,321 @@
...
-}
+    /// Evict expired values from the cache.
+    ///
+    /// Polls the underlying `DelayQueue`. When elements are returned from the
+    /// queue, remove the associated key from `values`.
+    fn poll_purge(&mut self) -> Poll<(), Error> {
+        while let Some(expired) = try_ready!(self.expirations.poll()) {
+            trace!("expired an entry from expirations");
+            self.values.remove(expired.get_ref());
+            trace!("removed an entry from values!");
```
olix0r: This log line is redundant given the immedately preceding log line.

lib/router/src/cache.rs
```diff
@@ -1,426 +1,321 @@
...
-        &self.value
-    }
-}
+    /// Evict expired values from the cache.
+    ///
+    /// Polls the underlying `DelayQueue`. When elements are returned from the
+    /// queue, remove the associated key from `values`.
+    fn poll_purge(&mut self) -> Poll<(), Error> {
+        while let Some(expired) = try_ready!(self.expirations.poll()) {
+            trace!("expired an entry from expirations");
```
olix0r: What do you think about moving this up to debug and rewording it slightly:
```rust
debug!("expiring an item from the cache")
```

lib/router/src/cache.rs
```diff
@@ -1,426 +1,321 @@
...
+    pub fn insert(&mut self, key: K, value: V) -> Option<V> {
+        let node = {
+            trace!("inserting value into expirations");
+            let dq_key = self.expirations.insert(key.clone(), self.expires);
+            Node { dq_key, value }
+        };
 
-    pub fn last_access(&self) -> Instant {
-        self.last_access
+        trace!("inserting node into values");
```
olix0r: Again, this is fairly redundant with the above log line.

lib/router/src/cache.rs
```diff
@@ -1,426 +1,321 @@
...
-    pub fn access<'a, N: Now + 'a>(&'a mut self, now: &'a N) -> Access<'a, T, N> {
-        Access { now, node: self }
-    }
+    /// Attempts to insert an item by key.
+    ///
+    /// If a value is returned, this key has been set to expire after an
+    /// `expires` span of time.
+    pub fn insert(&mut self, key: K, value: V) -> Option<V> {
+        let node = {
+            trace!("inserting value into expirations");
```
olix0r: Similarly, I'd suggest that we move this up to debug and make the phrasing uniform with the expiration:

```rust
debug!("inserting an item from the cache")
```

lib/router/src/cache.rs
kleimkuhler: Its logging the fact that we are removing a value from both the internal `expirations`, and then router facing state of the cache `values`, but I see your point. I'll remove one

PR #248, opened 2019-05-08
Load balancers fall back to ORIG_DST when no endpoints exist

hawkw:
Currently, when no endpoints exist in the load balancer for a
destination, we fail the request. This is because we expect endpoints to
be discovered by both destination service queries _and_ DNS lookups, so
if there are no endpoints for a destination, it is assumed to not exist.

In linkerd/linkerd2#2661, we intend to remove the DNS lookup from the
proxy and instead fall back to routing requests for which no endpoints
exist in the destination service to their SO_ORIGINAL_DST IP address.
This means that the current approach of failing requests when the load
balancer has no endpoints will no longer work.

This branch introduces a generic `fallback` layer, which composes a
primary and secondary service builder into a new layer. The primary 
service can fail requests with an error type that propages the original
request, allowing the fallback middleware to call the fallback service
with the same request. Other errors returned by the primary service are
still propagated upstream.

In contrast to the approach used in #240, this fallback middleware is 
generic and not tied directly to a load balancer or a router, and can
be used for other purposes in the future. It relies on the router cache
eviction added in #247 to drain the router when it is not being used, 
rather than proactively destroying the router when endpoints are
available for the lb, and re-creating it when they exist again.

A new trait, `HasEndpointStatus`, is added in order to allow the
discovery lookup to communicate the "no endpoints" state to the
balancer. In addition, we add a new `Update::NoEndpoints` variant to
`proxy::resolve::Update`, so that when the control plane sends a no
endpoints update, we switch from the balancer to the no endpoints state
_immediately_, rather than waiting for all the endpoints to be
individually removed. When the balancer has no endpoints, it fails all
requests with a fallback error, so that the fallback middleware

A future commit will remove the DNS lookups from the discovery module.

Closes #240.

### DIFF
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -20,7 +20,7 @@ use logging;
 use metrics::FmtMetrics;
 use never::Never;
 use proxy::{
-    self, accept,
+    self, accept, buffer,
     http::{
         client, insert, metrics as http_metrics, normalize_uri, profiles, router, settings,
         strip_header,
@@ -419,12 +419,13 @@ where
 
         let outbound = {
             use super::outbound::{
-                //add_remote_ip_on_rsp, add_server_id_on_rsp,
+                self,
                 discovery::Resolve,
                 orig_proto_upgrade,
+                //add_remote_ip_on_rsp, add_server_id_on_rsp,
             };
             use proxy::{
-                http::{balance, canonicalize, header_from_target, metrics, retry},
+                http::{balance, canonicalize, fallback, header_from_target, metrics, retry},
                 resolve,
             };
 
@@ -499,9 +500,23 @@ where
                 .layer(metrics::layer::<_, classify::Response>(retry_http_metrics))
                 .layer(insert::target::layer());
 
-            let balancer_stack = svc::builder()
+            let balancer = svc::builder()
                 .layer(balance::layer(EWMA_DEFAULT_RTT, EWMA_DECAY))
-                .layer(resolve::layer(Resolve::new(resolver)))
+                .layer(resolve::layer(Resolve::new(resolver)));
+
+            let fallback = svc::builder()
+                .layer(router::layer(
+                    router::Config::new("out ep", capacity, max_idle_age),
+                    |req: &http::Request<_>| {
+                        let ep = outbound::Endpoint::from_orig_dst(req);
+                        debug!("outbound ep={:?}", ep);
+                        ep
+                    },
+                ))
+                .layer(buffer::layer(max_in_flight, DispatchDeadline::extract));
+
+            let balancer_stack = svc::builder()
+                .layer(fallback::layer(balancer, fallback))
                 .layer(pending::layer())
                 .layer(balance::weight::layer())
                 .service(endpoint_stack);
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -504,7 +504,9 @@ where
                 .layer(balance::layer(EWMA_DEFAULT_RTT, EWMA_DECAY))
                 .layer(resolve::layer(Resolve::new(resolver)));
 
-            let fallback = svc::builder()
+            // Routes requests to their original destination endpoints. Used as
+            // a fallback when service discovery has no endpoints for a destination.
+            let orig_dst_router = svc::builder()
                 .layer(router::layer(
                     router::Config::new("out ep", capacity, max_idle_age),
                     |req: &http::Request<_>| {
@@ -516,7 +518,7 @@ where
                 .layer(buffer::layer(max_in_flight, DispatchDeadline::extract));
 
             let balancer_stack = svc::builder()
-                .layer(fallback::layer(balancer, fallback))
+                .layer(fallback::layer(balancer, orig_dst_router))
                 .layer(pending::layer())
                 .layer(balance::weight::layer())
                 .service(endpoint_stack);
--- a/src/proxy/http/balance.rs
+++ b/src/proxy/http/balance.rs
@@ -5,7 +5,7 @@ extern crate tower_discover;
 use std::marker::PhantomData;
 use std::time::Duration;
 
-use futures::{Future, Poll};
+use futures::{Async, Future, Poll};
 use hyper::body::Payload;
 
 use self::tower_discover::Discover;
@@ -15,6 +15,7 @@ pub use self::tower_balance::{
     choose::PowerOfTwoChoices, load::WithPeakEwma, Balance, HasWeight, Weight, WithWeighted,
 };
 
+use proxy::resolve::{HasEndpointStatus, EndpointStatus};
 use http;
 use svc;
 
@@ -34,6 +35,12 @@ pub struct MakeSvc<M, A, B> {
     default_rtt: Duration,
     inner: M,
     _marker: PhantomData<fn(A) -> B>,
+}
+
+#[derive(Debug)]
+pub struct Service<S> {
+    balance: S,
+    status: EndpointStatus,
 }
 
 // === impl Layer ===
@@ -89,13 +96,13 @@ impl<M: Clone, A, B> Clone for MakeSvc<M, A, B> {
 impl<T, M, A, B> svc::Service<T> for MakeSvc<M, A, B>
 where
     M: svc::Service<T>,
-    M::Response: Discover,
+    M::Response: Discover + HasEndpointStatus,
     <M::Response as Discover>::Service:
         svc::Service<http::Request<A>, Response = http::Response<B>>,
     A: Payload,
     B: Payload,
 {
-    type Response = Balance<WithPeakEwma<M::Response, PendingUntilFirstData>, PowerOfTwoChoices>;
+    type Response = Service<Balance<WithPeakEwma<M::Response, PendingUntilFirstData>, PowerOfTwoChoices>>;
     type Error = M::Error;
     type Future = MakeSvc<M::Future, A, B>;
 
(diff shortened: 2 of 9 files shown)


src/app/main.rs
```diff
@@ -499,9 +500,27 @@ where
...
+                .layer(resolve::layer(Resolve::new(resolver)));
+
+            let fallback = svc::builder()
+                .layer(
+                    router::layer(|req: &http::Request<_>| {
+                        let ep = outbound::Endpoint::from_orig_dst(req);
+                        debug!("outbound ep={:?}", ep);
+                        ep
+                    })
+                    .configured(router::Config::new(
```
hawkw: This is necessary as the router layer cannot make a `Service<Config>` here --- we need the primary and fallback layers to take the same target types.

I've considered refactoring router construction so that all router layers to be constructed with a config rather than binding to it after the stack is built, but that would change a lot of unrelated code & ought to be a separate branch. This works for now.

src/proxy/http/router.rs
```diff
@@ -97,6 +103,65 @@ where
     }
 }
 
+impl<Req, Rec> Clone for Layer<Req, Rec>
+where
+    Rec: Recognize<Req> + Clone + Send + Sync + 'static,
+{
```
hawkw: The previous derived `Clone` implementation for  `Layer` (and for `Stack`) required that the request type also be `Clone`, even though it's a `PhantomData`; this had to be fixed since the fallback layer requires cloning the fallback service.

src/app/main.rs
olix0r: Can we just move all of this into the `router::layer` constructor? Is there a good reason to have two versions of the router layer, really?  Perhaps that should be its own setup change (to change router configuration to use a Target type of `()` and to require the layer have the required configuration).

src/proxy/http/fallback.rs
```diff
@@ -0,0 +1,336 @@
...
+
+use std::{fmt, marker::PhantomData, mem};
+
+#[derive(Debug)]
+pub struct Error<A> {
+    error: proxy::Error,
+    fallback: Option<http::Request<A>>,
+}
+
+pub type Layer<P, F, A> = MakeFallback<svc::Builder<P>, svc::Builder<F>, A>;
```
olix0r: that we do this for Layer and not the others is.... unclear? I'd just redefine it? but i don't care deeply

src/proxy/http/fallback.rs
```diff
@@ -0,0 +1,336 @@
...
+    _p: PhantomData<fn(A)>,
+}
+
+pub struct MakeFuture<P, F, A>
+where
+    P: Future,
+    F: Future,
+{
+    primary: Making<P>,
+    fallback: Making<F>,
```
olix0r: What's going on here? Isn't this just Pending? Can the Service just hold pending services?

src/proxy/http/fallback.rs
```diff
@@ -0,0 +1,336 @@
...
+    T: Clone,
+{
+    type Response = Service<P::Response, F::Response, A>;
+    type Error = proxy::Error;
+    type Future = MakeFuture<P::Future, F::Future, A>;
+
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        let primary_ready = self.primary.poll_ready().map_err(Into::into);
+        let fallback_ready = self.fallback.poll_ready().map_err(Into::into);
+        try_ready!(fallback_ready);
```
olix0r: So, this means that if the fallback isn't ready, we'll never attempt to use the primary... is that appropriate?

src/proxy/http/fallback.rs
```diff
@@ -0,0 +1,336 @@
...
+    P: Future,
+    P::Error: Into<proxy::Error>,
+    F: Future,
+    F::Error: Into<proxy::Error>,
+{
+    type Item = Service<P::Item, F::Item, A>;
+    type Error = proxy::Error;
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        let done =
+            self.primary.poll().map_err(Into::into)? && self.fallback.poll().map_err(Into::into)?;
```
olix0r: Is there something subtle going on here with `&&` and `Async::Ready(())` ?! This isn't clear to me... Would be clearer with `.is_ready()` or whatever.

src/proxy/http/fallback.rs
```diff
@@ -0,0 +1,336 @@
...
+impl<P, F, A> Future for MakeFuture<P, F, A>
+where
+    P: Future,
+    P::Error: Into<proxy::Error>,
+    F: Future,
+    F::Error: Into<proxy::Error>,
+{
+    type Item = Service<P::Item, F::Item, A>;
+    type Error = proxy::Error;
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
```
olix0r: we tend to leave a newline between type and fn definitions....

src/proxy/http/fallback.rs
```diff
@@ -0,0 +1,336 @@
...
+                        .poll()
+                        .map(|p| p.map(|rsp| rsp.map(Body::B)))
+                        .map_err(Into::into)
+                }
+            }
+        }
+    }
+}
+
+impl<A, B> Payload for Body<A, B>
```
olix0r: `// === impl Body ===` would be helpful above here, or we should split into separate files

src/proxy/http/fallback.rs
```diff
@@ -0,0 +1,336 @@
...
+    fallback: F,
+    state: State<P, F::Future>,
+}
+
+pub enum Body<A, B> {
+    A(A),
+    B(B),
+}
+
+enum State<P, F> {
```
olix0r: prefer `ResponseInner` or something that makes it clearer how/where it's used.. `State` would be fine if this was in a file with only the responsefuture...

src/proxy/http/fallback.rs
olix0r: I see.. `poll_ready` doesn't actually return a `Poll`... it should have a different name or signature.

src/proxy/http/fallback.rs
hawkw: personally i felt like "state" conveys at least slightly more meaning than "inner" --- how about `ResponseState`?

src/proxy/http/fallback.rs
hawkw: I was initially going to reuse one struct for more of the types but it didn't work for the two Service impls due to coherence. Can replace this with a separate struct.

src/proxy/http/fallback.rs
hawkw: This is less like `Pending` and more like `future::Join` --- the idea is that the future is complete when both the wrapped services complete. Reimplementing `join` felt like less work than making `join(...).map(...)` work w/o a closure

src/proxy/http/fallback.rs
hawkw: > So, this means that if the fallback isn't ready, we'll never attempt to use the primary... is that appropriate?

The alternative is to fail the request if we try to fall back and the fallback service isn't ready. Is that preferable?

src/proxy/http/fallback.rs
hawkw: No, the "poll" fn on these returns `Result<bool, ...>`. Maybe it should be renamed?

src/proxy/http/fallback.rs
hawkw: (i just wanted to save on the `impl Clone` boilerplate honestly)

src/proxy/http/fallback.rs
hawkw: > I see.. `poll_ready` doesn't actually return a `Poll`... it should have a different name or signature.

I think this comment was intended as a reply to https://github.com/linkerd/linkerd2-proxy/pull/248#discussion_r283576128?

olix0r: I had imagined this working by cloneing the fallback service into the response future and then using an error response to cause fallback, at which point the inner service would be polled for readiness... I don't really have my head around the EndpointStatus stuff yet enough to compare...

olix0r: This probably isn't a hard blocker given that we'll use a router in the fallback route.. but we'll probably need to do something like my suggestion to make this a truly generic component.

hawkw: > I had imagined this working by cloneing the fallback service into the response future and then using an error response to cause fallback, 

That _is_ how this works...

> at which point the inner service would be polled for readiness... 

We could do it that way rather than polling the fallback service here...in that case, we would return the error if the fallback service isn't ready. Is that preferable over exerting backpressure here?

src/proxy/http/fallback.rs
hawkw: changed it to be a proper poll in 123b2b70f6fb082a0a2cdf4850e768fee8be4fc2, thanks! reads much better now IMO

src/proxy/http/fallback.rs
```diff
@@ -0,0 +1,334 @@
...
+            Making::Ready(_) => return Ok(Async::Ready(())),
+            Making::Done => panic!("polled after ready"),
+        };
+        Ok(Async::Ready(()))
+    }
+
+    fn take(&mut self) -> T::Item {
+        match mem::replace(self, Making::Done) {
+            Making::Ready(a) => a,
+            _ => panic!(),
```
kleimkuhler: A comment here could be useful. Something like `tried to take service after ready`?

src/app/main.rs
hawkw: @olix0r I made the change you suggested in #253, and rebased this branch to be on top of that branch. Thanks!

hawkw:
@olix0r and @kleimkuhler, I've made all the changes you requested. I think we still need to agree on what the correct thing to do when the fallback service is not ready is.

src/proxy/http/fallback.rs
olix0r: > we would return the error if the fallback service isn't ready.

Why? Can't the response future wait until this service is ready? It sounds like ResonseState needs another state

hawkw: Oh, wow, I hadn't thought of that but it's obvious in retrospect. Will change it around.

src/proxy/http/balance.rs
```diff
@@ -118,19 +132,45 @@ where
...
+        future::MapErr<S::Future, fn(proxy::Error) -> Self::Error>,
+        future::FutureResult<http::Response<B>, fallback::Error<A>>,
+    >;
+
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        self.balance.poll_ready().map_err(From::from)
+    }
+
+    fn call(&mut self, req: http::Request<A>) -> Self::Future {
+        if self.status.is_empty() {
```
olix0r: Where is this set?

olix0r: Oh, I see, this is configured in the constructor.

olix0r: My main question about this... it _feels_ like something we should be polling, rather than switching on. But, given that this will be driven by balance.poll_ready, it's probably fine as it is.

PR #251, opened 2019-05-13
Assert that outbound TLS works before identity is certified

zaharidichev:
This commit introduces TLS capabilities to the support server as well as
tests to ensure that outbound TLS works even when there is no verified
certificate for the proxy yet. 

Fixes https://github.com/linkerd/linkerd2/issues/2599

### DIFF
--- a/tests/identity.rs
+++ b/tests/identity.rs
@@ -21,6 +21,7 @@ fn nonblocking_identity_detection() {
         env,
         mut certify_rsp,
         client_config: _,
+        server_config: _,
     } = identity::Identity::new("foo-ns1", id.to_string());
     certify_rsp.valid_until = Some((SystemTime::now() + Duration::from_secs(666)).into());
 
@@ -87,6 +88,7 @@ macro_rules! generate_tls_reject_test {
             env,
             mut certify_rsp,
             client_config,
+            server_config: _,
         } = identity::Identity::new("foo-ns1", id.to_string());
 
         certify_rsp.valid_until = Some((SystemTime::now() + Duration::from_secs(666)).into());
@@ -137,6 +139,61 @@ fn http2_rejects_tls_before_identity_is_certified() {
     generate_tls_reject_test! {client: client::http2_tls}
 }
 
+macro_rules! generate_outbound_tls_accept_not_cert_identity_test {
+    (server: $make_server:path, client: $make_client:path) => {
+        let _ = env_logger_init();
+        let proxy_id = "foo.ns1.serviceaccount.identity.linkerd.cluster.local";
+        let app_id = "bar.ns1.serviceaccount.identity.linkerd.cluster.local";
+
+        let mut proxy_identity = identity::Identity::new("foo-ns1", proxy_id.to_string());
+
+        proxy_identity.certify_rsp.valid_until =
+            Some((SystemTime::now() + Duration::from_secs(666)).into());
+
+        let (_, rx) = oneshot::channel();
+        let id_svc = controller::identity().certify_async(move |_| rx).run();
+
+        let app_identity = identity::Identity::new("bar-ns1", app_id.to_string());
+        let srv = $make_server(app_identity.server_config)
+            .route("/", "hello")
+            .run();
+
+        let proxy = proxy::new()
+            .outbound(srv)
+            .identity(id_svc)
+            .run_with_test_env(proxy_identity.env);
+
+        let client = $make_client(
+            proxy.outbound,
+            app_id,
+            client::TlsConfig::new(app_identity.client_config, app_id),
+        );
+
+        assert_eventually!(
+            client
+                .request(client.request_builder("/").method("GET"))
+                .status()
+                == http::StatusCode::OK
+        );
+    };
+}
+
+#[test]
+fn http1_outbound_tls_works_before_identity_is_certified() {
+    generate_outbound_tls_accept_not_cert_identity_test! {
+    server: server::http1_tls,
+    client : client::http1_tls
+    }
+}
+
+#[test]
+fn http2_outbound_tls_works_before_identity_is_certified() {
+    generate_outbound_tls_accept_not_cert_identity_test! {
+    server: server::http2_tls,
+    client : client::http2_tls
+    }
+}
+
 #[test]
 fn ready() {
     let _ = env_logger_init();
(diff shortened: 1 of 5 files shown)


tests/support/identity.rs
```diff
@@ -29,44 +30,88 @@ type Certify = Box<
...
-        leaf_certificate,
-        intermediate_certificates,
-        ..Default::default()
-    };
-    Ok(rsp)
+const TLS_VERSIONS: &[rustls::ProtocolVersion] = &[rustls::ProtocolVersion::TLSv1_2];
+
+struct Certificates {
+    pub leaf_certificate: Vec<u8>,
+    pub intermediate_certificates: Vec<Vec<u8>>,
```
kleimkuhler: Since the struct is `Certificates`, maybe we call the fields `leaf` and `intermediates`?

tests/identity.rs
```diff
@@ -137,6 +139,61 @@ fn http2_rejects_tls_before_identity_is_certified() {
...
+        let _ = env_logger_init();
+        let proxy_id = "foo.ns1.serviceaccount.identity.linkerd.cluster.local";
+        let app_id = "bar.ns1.serviceaccount.identity.linkerd.cluster.local";
+
+        let mut proxy_identity = identity::Identity::new("foo-ns1", proxy_id.to_string());
+
+        proxy_identity.certify_rsp.valid_until =
+            Some((SystemTime::now() + Duration::from_secs(666)).into());
+
+        let (_, rx) = oneshot::channel();
```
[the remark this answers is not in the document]
zaharidichev: Ahh, nice catch. Thank you

tests/support/mod.rs
```diff
@@ -163,6 +169,46 @@ pub mod server;
 pub mod tap;
 pub mod tcp;
 
+enum RunningIo<T> {
+    Plain(TcpStream, Option<oneshot::Sender<()>>),
+    Tls(TlsStream<TcpStream, T>, Option<oneshot::Sender<()>>),
+}
+impl<T: Session> Read for RunningIo<T> {
```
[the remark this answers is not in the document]
zaharidichev: had hopes that rustfmt would take care of that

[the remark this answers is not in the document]
kleimkuhler: Yes it takes care of a lot, but will not try to do anything with newlines :(

tests/identity.rs
```diff
@@ -21,6 +21,7 @@ fn nonblocking_identity_detection() {
         env,
         mut certify_rsp,
         client_config: _,
+        server_config: _,
```
hawkw: TIOLI: you could also just say
```rust
    let identity::Identity { 
        env, 
        mut certify_rsp, 
        ..  
    } = identity::Identity::new("foo-ns1", id.to_string());
```
which is a little bit more concise (and if we imported `Identity`, would probably fit on one line!)

(ref: https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html#ignoring-remaining-parts-of-a-value-with-)

tests/identity.rs
```diff
@@ -137,6 +139,58 @@ fn http2_rejects_tls_before_identity_is_certified() {
...
+                == http::StatusCode::OK
+        );
+    };
+}
+
+#[test]
+fn http1_outbound_tls_works_before_identity_is_certified() {
+    generate_outbound_tls_accept_not_cert_identity_test! {
+    server: server::http1_tls,
+    client : client::http1_tls
```
hawkw: Nit: weird whitespace here?

tests/identity.rs
```diff
@@ -137,6 +139,58 @@ fn http2_rejects_tls_before_identity_is_certified() {
...
+    }
+}
+
+#[test]
+fn http2_outbound_tls_works_before_identity_is_certified() {
+    generate_outbound_tls_accept_not_cert_identity_test! {
+    server: server::http2_tls,
+    client : client::http2_tls
```
hawkw: ...and here?

tests/support/client.rs
```diff
@@ -4,14 +4,10 @@ use std::io;
...
 
 use self::futures::sync::{mpsc, oneshot};
-use self::tokio::{
-    io::{AsyncRead, AsyncWrite},
-    net::TcpStream,
-};
-use self::tokio_rustls::TlsStream;
+use self::tokio::net::TcpStream;
 use self::webpki::{DNSName, DNSNameRef};
+use self::RunningIo;
```
hawkw: I don't think this is necessary -- was it a find-and-replace?

tests/identity.rs
```diff
@@ -145,6 +199,7 @@ fn ready() {
         env,
         mut certify_rsp,
         client_config: _,
+        server_config: _,
```
hawkw: Again, could use `..` here.

tests/support/client.rs
zaharidichev: most likely

tests/identity.rs
zaharidichev: hmm... so rustfmt does not work for macro invocations?

hawkw: yeah, rustfmt will not format macro invocations as they're not guaranteed to contain valid Rust syntax.

tests/identity.rs
```diff
@@ -179,15 +178,15 @@ macro_rules! generate_outbound_tls_accept_not_cert_identity_test {
 fn http1_outbound_tls_works_before_identity_is_certified() {
     generate_outbound_tls_accept_not_cert_identity_test! {
     server: server::http1_tls,
-    client : client::http1_tls
+    client: client::http1_tls
```
hawkw: Sorry to keep picking formatting nits, but I think `server` and `client` should also be indented one more level since they're inside the macro invocation's curly braces.

tests/identity.rs
```diff
@@ -179,15 +178,15 @@ macro_rules! generate_outbound_tls_accept_not_cert_identity_test {
...
     }
 }
 
 #[test]
 fn http2_outbound_tls_works_before_identity_is_certified() {
     generate_outbound_tls_accept_not_cert_identity_test! {
     server: server::http2_tls,
-    client : client::http2_tls
+    client: client::http2_tls
```
hawkw: same as above

kleimkuhler:
@zaharidichev Quick reminder that it looks like you saw my [comment](https://github.com/linkerd/linkerd2-proxy/pull/251/files#r283942388), but that still needs to be fixed so that the channel is not closed right away!

zaharidichev:
@kleimkuhler one too many cmd&z presses

PR #252, opened 2019-05-13
Allow to disable tap via setting an env var

zaharidichev:
This PR aims to fix https://github.com/linkerd/linkerd2/issues/2811. Now if `LINKERD2_PROXY_TAP_DISABLED` is set, the tap is not served at all. The approach taken is that  the `ProxyParts` is changed so the `control_listener` is now an `Option` that will be None if tap is disabled as this control_listener seems to be exclusively used to serve the tap. Feel free to suggest a better approach. 

### DIFF
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -145,6 +145,9 @@ pub struct Config {
     pub dns_canonicalize_timeout: Duration,
 
     pub h2_settings: H2Settings,
+
+    /// When set, tap is not served
+    pub tap_disabled: bool,
 }
 
 #[derive(Copy, Clone, Debug, Default)]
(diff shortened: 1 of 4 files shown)


olix0r:
Thanks, @zaharidichev!

src/app/config.rs
```diff
@@ -145,6 +145,9 @@ pub struct Config {
     pub dns_canonicalize_timeout: Duration,
 
     pub h2_settings: H2Settings,
+
+    /// When set, tap is not served
+    pub tap_disabled: bool,
```
olix0r: What if, instead of this flag, we make the `control_listener` hold an `Option` so that it only exists when tap is enabled?

src/app/config.rs
```diff
@@ -464,6 +464,19 @@ impl Config {
...
+            .map(|d| !d.is_empty())
+            .unwrap_or(false);
+
+        let control_listener = if tap_disabled {
+            None
+        } else {
+            let addr = parse(strings, ENV_CONTROL_LISTEN_ADDR, parse_socket_addr)?
+                .unwrap_or_else(|| parse_socket_addr(DEFAULT_CONTROL_LISTEN_ADDR).unwrap());
+            Some(Listener { addr })
+        };
```
hawkw: TIOLI: Maybe the logic for parsing the control listener if enabled should be factored out into its own function? `config::parse` is quite long...

tests/support/proxy.rs
```diff
@@ -273,7 +273,10 @@ fn run(proxy: Proxy, mut env: app::config::TestEnv) -> Listening {
     // printlns will show if the test fails...
     println!(
         "proxy running; destination={}, identity={:?}, inbound={}{}, outbound={}{}, metrics={}",
-        control_addr,
+        control_addr
+            .as_ref()
+            .map(|i| format!("{}", i))
```
hawkw: TIOLI: I think you could also use
```suggestion
            .map(SocketAddr::to_string)
```

src/app/main.rs
```diff
@@ -379,7 +384,10 @@ where
                     ));
 
                     rt.spawn(tap_daemon.map_err(|_| ()));
```
kleimkuhler: Because we are disabling tap for the entire proxy, we could have a similar `if let` check here and conditionally spawn the `tap_daemon`.

PR #255, opened 2019-05-15
Fix load balancer not being ready when it wants to fall back

hawkw:
The `tower-balance` load balancer's `poll_ready` function returns
`NotReady` when the load balancer has no ready endpoints. However, our
fallback strategy added in #248 requires the load balancer to accept a
request when it has no endpoints, so that it can return the error
indicating that the fallback router should be used. This branch changes
the proxy's `http::balance::Service`'s `poll_ready` implementation to
return `Ok(Async::Ready(()))` when it is in the no endpoints state, so
that the request will fall back.

This fix was backported from #245.

### DIFF
--- a/src/proxy/http/balance.rs
+++ b/src/proxy/http/balance.rs
@@ -164,6 +164,11 @@ where
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
         let ready = self.balance.poll_ready().map_err(fallback::Error::from)?;
         if self.status.is_empty() {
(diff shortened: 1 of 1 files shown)


hawkw:
Looking at the test results on CI, I think this _does_ only make sense in the context of removing the DNS fallback (#245); this appears to break stuff on its own.

olix0r:
Shall we close this then?

hawkw:
@olix0r yes, sorry --- I meant to click "close and comment"

PR #258, opened 2019-05-20
Use the proxy's log formatting in tests

hawkw:
The proxy implements custom log formatting which includes timestamps and
formatted contexts describing the context in which the log line was
recorded. Currently, however, when running the proxy's integration
tests, the default log formatting from the `env_logger` crate rather
than the proxy's format. There are two main reasons why it would be
preferable to use the proxy's log formatting style instead:

* Information in the log contexts and timestamps can be valuable for
  debugging a failing test. For example, many log lines don't include
  information such as the destination address for a request as they
  assume this will be provided by the log context. This can make
  understanding log output from tests difficult since the log contexts
  are not included.
* In order to debug changes to the log formatting, it may be easier to
  run a test with a log level set, rather than exercise certain
  behaviours manually. This is my original motivation behind this change
  --- I wanted to add a new log context to the background task driving
  destination service discovery, and it takes significantly less work to
  simply run the discovery tests with logging enabled rather than
  actually spinning up a new Kubernetes cluster to test this with a live
  destination service.

This branch changes the integration test support module's
`env_logger_init` function to use the proxy's log formatter rather than
the default.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -53,7 +53,7 @@ pub mod convert;
 mod dns;
 mod drain;
 mod identity;
-mod logging;
+pub mod logging;
 mod proxy;
 mod svc;
 mod tap;
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -18,31 +18,36 @@ thread_local! {
     static CONTEXT: RefCell<Vec<*const fmt::Display>> = RefCell::new(Vec::new());
 }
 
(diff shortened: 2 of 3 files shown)


hawkw:
I originally made this change to test out changes on a different in-progress branch, but I figured that I'd split it into a separate branch for the git history's sake.

PR #260, opened 2019-05-28
[GSoC 2019] Add test case to detect Kafka traffic

ruig2:
This PR is **for reviewing and feedback purpose** only.

This is a very first step to the GSoC 2019 project (Kafka detector and analyzer). The logic of Kafka codec, in the end, will be moved to a new dedicated Rust lib and Linkerd will call that lib to minimize the changes on Linkerd side.

I have a few questions about how to code in the Linkerd codebase and commented them in the PR.
Thanks for your suggestion and feedback ;)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ linkerd2-timeout   = { path = "lib/timeout" }
 
 linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", tag = "v0.1.8" }
 
+byteorder = "1.3.1"
 bytes = "0.4"
 env_logger = { version = "0.5", default-features = false }
 futures = "0.1"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,7 @@
 #![deny(warnings)]
 #![recursion_limit = "128"]
 
+extern crate byteorder;
 extern crate bytes;
 extern crate env_logger;
 #[macro_use]
--- a/src/proxy/protocol.rs
+++ b/src/proxy/protocol.rs
@@ -1,4 +1,5 @@
 use httparse;
+use byteorder::{BigEndian, ByteOrder};
 
 /// Transport protocols that can be transparently detected by `Server`.
 #[derive(Debug)]
@@ -17,6 +18,7 @@ impl Protocol {
         // http2 is easiest to detect
         if bytes.len() >= H2_PREFACE.len() {
             if &bytes[..H2_PREFACE.len()] == H2_PREFACE {
+                println!("^^^^^ is H2");
                 return Some(Protocol::Http2);
             }
         }
@@ -35,11 +37,49 @@ impl Protocol {
             // We didn't want to keep parsing headers, just validate that
             // the first line is HTTP1.
             Ok(_) | Err(httparse::Error::TooManyHeaders) => {
+                println!("^^^^^ is H1");
                 return Some(Protocol::Http1);
             }
             _ => {}
         }
 
+        // Detect if this is a Kafka traffic
+
+        // Header:
+        //     Request/response Size => INT32
+        //     api_key => INT16
+        //     api_version => INT16
+        //     correlation_id => INT32
+        //     client_id => NULLABLE_STRING (first two bytes as the length of the client_id string)
+        const KAFKA_REQUEST_HEADER_LENGTH: usize = 4 + 2 + 2 + 4 + 2;
+        println!("length {}", bytes.len());
+        if bytes.len() >= KAFKA_REQUEST_HEADER_LENGTH {
+            let request_size =      BigEndian::read_i32(&bytes);
+            let api_key =           BigEndian::read_i16(&bytes[4..]);
+            let api_version =       BigEndian::read_i16(&bytes[6..]);
+            let correlation_id =    BigEndian::read_i32(&bytes[8..]);
+            let len_client_id =     BigEndian::read_i16(&bytes[12..]);
+            let mut client_id = "null";
+
+            if len_client_id > -1 {
+                use std::str;
+
+                client_id = match str::from_utf8(&bytes[14..(14+len_client_id as usize)]) {
+                    Ok(v) => v,
+                    Err(e) => panic!("Invalid UTF-8 sequence: {}", e),
+                };
+            }
+
+            println!("checking if is kafka...");
+            println!("\trequest_size: {}", request_size);
+            println!("\tapi_key: {}", api_key);
+            println!("\tapi_version: {}", api_version);
+            println!("\tcorrelation_id: {}", correlation_id);
+            println!("\tlen_client_id: {}", len_client_id);
+            println!("\tclient_id: {}", client_id);
+        }
+
+        println!("^^^^^ is None");
         None
     }
 }
--- a/src/proxy/protocol.rs
+++ b/src/proxy/protocol.rs
@@ -66,7 +66,7 @@ impl Protocol {
 
                 client_id = match str::from_utf8(&bytes[14..(14+len_client_id as usize)]) {
                     Ok(v) => v,
-                    Err(e) => panic!("Invalid UTF-8 sequence: {}", e),
+                    Err(_) => return None,
                 };
             }
 
--- a/tests/transparency.rs
+++ b/tests/transparency.rs
@@ -93,13 +93,13 @@ fn outbound_kafka() {
         // client_id => NULLABLE_STRING
         0, 10,
         99, 111, 110, 115, 117, 109, 101, 114, 45, 49]; // value: consumer-1
-    let _msg1 = "custom tcp hello";
-    let _msg2 = "custom tcp bye";
+    // Just echo to make things easier for now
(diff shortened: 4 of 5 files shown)


Cargo.toml
```diff
@@ -33,6 +33,7 @@ linkerd2-timeout   = { path = "lib/timeout" }
 
 linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", tag = "v0.1.8" }
 
+byteorder = "1.3.1"
```
ruig2: The `byteorder` lib may be removed from Linkerd if we implement a new and dedicated Kafka Rust codec. If we do need such a lib in Linkerd, then is it correct to put it here?

src/lib.rs
```diff
@@ -1,6 +1,7 @@
 #![deny(warnings)]
 #![recursion_limit = "128"]
 
+extern crate byteorder;
```
ruig2: Same question as above.

src/proxy/protocol.rs
```diff
@@ -17,6 +18,7 @@ impl Protocol {
         // http2 is easiest to detect
         if bytes.len() >= H2_PREFACE.len() {
             if &bytes[..H2_PREFACE.len()] == H2_PREFACE {
+                println!("^^^^^ is H2");
```
ruig2: I'll remove the `println`s later. This PR is for reviewing purpose only.

ruig2:
The output of the test is:

```
➜  linkerd2-proxy git:(rui/add_kafka_as_protocol) cargo test --color=always --package linkerd2-proxy --test transparency outbound_kafka -- --nocapture
    Finished dev [unoptimized] target(s) in 0.55s
     Running target/debug/deps/transparency-cce5f0bc06c29c72

running 1 test
*******************
tcp server (addr=127.0.0.1:53619): running
support destination service listening; addr=V4(127.0.0.1:53620)
proxy running; destination=127.0.0.1:53621, identity=None, inbound=127.0.0.1:53624, outbound=127.0.0.1:53623 (SO_ORIGINAL_DST=127.0.0.1:53619), metrics=127.0.0.1:53622
tcp client (addr=127.0.0.1:53623) thread running
tcp client (addr=127.0.0.1:53623): connected
tcp client (addr=127.0.0.1:53623): write
tcp client (addr=127.0.0.1:53623): read
length 24
checking if is kafka...
	request_size: 20
	api_key: 18
	api_version: 2
	correlation_id: 1
	len_client_id: 10
	client_id: consumer-1
^^^^^ is None
server Listening dropped; addr=127.0.0.1:53619
test outbound_kafka ... ok
```

hawkw:
@ruig2 I'm excited to take a look at this PR! 

Before I do, here are a couple things you may find useful when working in this repo:
- Please note that any commits merged to `linkerd2-proxy` must have a [Developer Certificate of Origin sign-off](https://probot.github.io/apps/dco/) for legal reasons. Commits without the DCO sign-off will [fail a check on your PR](https://github.com/linkerd/linkerd2-proxy/pull/260/checks?check_run_id=136938623) that must pass in order to merge. 

  This may not matter here if you don't intend to merge this branch, but I just thought it could save you some time later to point that out.
- When opening a pull request for review that you don't intend to merge, it's helpful to [mark the PR as a draft](https://github.blog/2019-02-14-introducing-draft-pull-requests/). This makes it easier to determine which PRs should be merged and which should not, which will make my life a little bit easier.

src/proxy/protocol.rs
hawkw: It's absolutely fine to use `println!` when debugging. If we want to print out diagnostic information that we expect to be useful in the future and don't intend to remove, though, we prefer to use the [`log` crate's macros](https://docs.rs/log/0.4.6/log/#macros), which will log records that include metadata and can be filtered, rather than printing directly to stdout.

src/proxy/protocol.rs
```diff
@@ -35,11 +37,49 @@ impl Protocol {
...
+        //     correlation_id => INT32
+        //     client_id => NULLABLE_STRING (first two bytes as the length of the client_id string)
+        const KAFKA_REQUEST_HEADER_LENGTH: usize = 4 + 2 + 2 + 4 + 2;
+        println!("length {}", bytes.len());
+        if bytes.len() >= KAFKA_REQUEST_HEADER_LENGTH {
+            let request_size =      BigEndian::read_i32(&bytes);
+            let api_key =           BigEndian::read_i16(&bytes[4..]);
+            let api_version =       BigEndian::read_i16(&bytes[6..]);
+            let correlation_id =    BigEndian::read_i32(&bytes[8..]);
+            let len_client_id =     BigEndian::read_i16(&bytes[12..]);
```
hawkw: Just so you're aware, CI builds on this repository are checked using [rustfmt](https://github.com/rust-lang/rustfmt), a Rust code formatting tool. Branches which are not formatted according to rustfmt's rules will fail on CI. 

It may make your life easier to download `rustfmt` and run it on changes branches prior to pushing them. Personally, I have VS Code configured to run `rustfmt` every time I save a file.

src/proxy/protocol.rs
```diff
@@ -35,11 +37,49 @@ impl Protocol {
...
+            }
+
+            println!("checking if is kafka...");
+            println!("\trequest_size: {}", request_size);
+            println!("\tapi_key: {}", api_key);
+            println!("\tapi_version: {}", api_version);
+            println!("\tcorrelation_id: {}", correlation_id);
+            println!("\tlen_client_id: {}", len_client_id);
+            println!("\tclient_id: {}", client_id);
+        }
```
hawkw: Eventually, I expect that this would return a `Protocol::Kafka` variant?

src/lib.rs
hawkw: Yeah, the `extern crate` statements should be in alphabetical order.

Cargo.toml
hawkw: Yes, if we need to add an additional dependency, this is the correct place to put it.

However, it's worth noting that we're already using the [`bytes` crate](https://docs.rs/crate/bytes/0.4.12), which also provides functions for reading bytes in big- and little-endian orders (I believe it uses `byteorder` internally). So, I think we may be able to implement the Kafka protocol detection using functions from the `bytes` crate, without adding a new dependency. If possible, that would probably be preferable.

src/proxy/protocol.rs
```diff
@@ -35,11 +37,49 @@ impl Protocol {
...
+        // Header:
+        //     Request/response Size => INT32
+        //     api_key => INT16
+        //     api_version => INT16
+        //     correlation_id => INT32
+        //     client_id => NULLABLE_STRING (first two bytes as the length of the client_id string)
+        const KAFKA_REQUEST_HEADER_LENGTH: usize = 4 + 2 + 2 + 4 + 2;
+        println!("length {}", bytes.len());
+        if bytes.len() >= KAFKA_REQUEST_HEADER_LENGTH {
+            let request_size =      BigEndian::read_i32(&bytes);
```
hawkw: As I mentioned above, I think we can implement this using the `bytes` crate, without adding a new dependency on `byteorder`. 

We could implement this parsing using the [`Buf`](https://docs.rs/bytes/0.4.12/bytes/trait.Buf.html) trait in `bytes`, which is implemented for `Cursor<&[u8]>`.
```rust
if bytes.len() >= KAFKA_REQUEST_HEADER_LENGTH {
    use bytes::Buf;
    use std::io::Cursor;

    let mut buf = Cursor::new(bytes);

    let request_size = buf.get_i32_be();
    let api_key = buf.get_i16_be();
    let api_version = buf.get_i16_be();
    // ... and so on
}
```
The nice thing about using `Cursor` and `Buf` is that the cursor position is advanced for you every time you read a number, so you don't have to slice the buffer from each offset. 

Alternatively, you could use the `bytes` crate's [re-export of `BigEndian`](https://docs.rs/bytes/0.4.12/bytes/enum.BigEndian.html) and this code would look identical, but we wouldn't need to add a new `extern crate` statement for `byteorder`.

tests/transparency.rs
```diff
@@ -76,6 +76,42 @@ fn inbound_tcp() {
     assert_eq!(tcp_client.read(), msg2.as_bytes());
 }
 
+#[test]
+fn outbound_kafka() {
```
hawkw: The tests in the `tests` directory are intended as end-to-end tests of the whole proxy. I think this test is focused enough that it might eventually be better off as a test specifically of the `Protocol::detect` method. In that case, it would live in that module, and the test could be much simpler, since you wouldn't have to create an entire proxy, a tcp client, and so on --- you could just call `Protocol::detect(KAFKA_REQUEST_API_VERSIONS)` and assert that you get `Some(Protocol::Kafka)` back.

With that said, having an end-to-end test for kafka proxying and detection is probably also valuable...

src/proxy/protocol.rs
ruig2: Of course.

ruig2:
Thanks for the detailed comment, and it is super helpful. I'll close this PR and continue working.

hawkw:
Great! Please don't hesitate to open more draft PRs in the future any time you need feedback, I'm always happy to answer questions.

PR #263, opened 2019-06-02
 discovery: Only fall back on InvalidArgument

hawkw:
When the Destination service returns a `NoEndpoints` response, a field
`exists` is set if the destination _does_ exist in service discovery
but has no endpoints. The API spec states that:

> `no_endpoints{exists: false}` indicates that the service does not exist
> and the client MAY try an alternate service discovery method (e.g. DNS).
>
> `no_endpoints(exists: true)` indicates that the service does exist and
> the client MUST NOT fall back to an alternate service discovery
> method.

When the DNS fallback behavior was removed from the proxy in #259, this
field was overlooked, and the proxy was changed to fall back to original
destination routing _any_ time the control plane indicates that no
endpoints exist for a destination. This is incorrect, as the proxy
should always treat the destination service as authoritative.
Additionally, this means that requests to endpoints which are known to
not exist will construct an unnecessary client service, and eventually
fail with a 502 error when the upstream client connection  to the
non-existent endpoint ultimately fails.

This branch changes the `control::destination::resolution::Daemon`
future to only send `Update::NoEndpoints` (which indicates that the load
balancer should fall back) to the load balancer when the Destination
service response has `exists: false`, or on `InvalidArgument` errors,
and _not_ when a `no_endpoints{exists: true}` response is received.
These requests will now fail fast with a 503 error rather than
attempting an ultimately futile fallback request.

Previously, one of the discovery tests expected the incorrect behavior.
I've changed this test to now expect that destinations known to not
exist do _not_ fall back, and renamed it from
`outbound_falls_back_to_orig_dst_when_destination_has_no_endpoints` to
`outbound_fails_fast_when_destination_has_no_endpoints`. I've also
confirmed that the updated test fails against master and passes after
this change.

Fixes linkerd/linkerd2#2880

### DIFF
--- a/src/control/destination/resolution.rs
+++ b/src/control/destination/resolution.rs
@@ -161,8 +161,8 @@ where
                                 let addrs = r_set.addrs.into_iter().filter_map(pb_to_sock_addr);
                                 try_send!(self.updater.remove(addrs));
                             }
-                            Some(PbUpdate2::NoEndpoints(_)) => {
-                                try_send!(self.updater.no_endpoints())
+                            Some(PbUpdate2::NoEndpoints(eps)) => {
+                                try_send!(self.updater.no_endpoints(eps.exists))
                             }
                             None => (),
                         };
@@ -179,7 +179,7 @@ where
                         // requested name should *not* query the destination
                         // service. In this case, do not attempt to reconnect.
                         debug!("Destination.Get stream ended with Invalid Argument",);
-                        let _ = self.updater.no_endpoints();
+                        let _ = self.updater.no_endpoints(false);
                         return Ok(Async::Ready(()));
                     }
                     Err(err) => {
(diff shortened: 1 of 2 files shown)


olix0r:
Per discussion in https://github.com/linkerd/linkerd2/issues/2880, should this be changed to never fallback when a NoEndpoints is returned?

hawkw:
@olix0r I'm happy to make that change, but I do wonder about one potential case: suppose that kube-dns sees the creation of a service before the Destination service does, so an app resolves the DNS name and gets back an IP, and sends the request to the proxy. The proxy looks up that name in the Destination service, and gets back a `no_endpoints { exists: false }`, so we wouldn't fall back. Is it worth only falling back when `exists: true` to avoid this kind of potential race?

hawkw:
@olix0r as we discussed offline, f20846c9d1281d02490b4b3b229e5d4de3ac876b changes this branch so that we now only fall back on `InvalidArgument`. I've also changed the log messages from `Resolution` to better reflect the distinction between `InvalidArgument` (should not query destination) and `NoEndpoints` responses, and fixed a test that expected the prior behaviour.

src/control/destination/resolution.rs
```diff
@@ -282,30 +278,28 @@ impl Updater {
...
-            trace!("remove {} (no endpoints)", addr);
+            trace!("remove {} ({})", addr, reason);
             self.tx
                 .unbounded_send(Update::Remove(addr))
                 .map_err(|_| ())?;
         }
         Ok(())
     }
+
+    fn does_not_exist(&mut self) -> Result<(), ()> {
```
olix0r: Hmmm, should this be called something else to distinguish from the other not-exists cases? Perhaps `unresolveable` or something like that?

hawkw: changed in fe9d612a8d4f48494a17a776cc80757c52d0f834, good call

hawkw:
@olix0r #268 moves the fallback from per-request to per-`MakeService`, as you suggested. Let me know what you think!

PR #266, opened 2019-06-05
Use a constant-time load balancer

olix0r:
tower-rs/tower#288 changed the Balance implementation substantially so
that it is no longer responsible for driving services to readiness.
Instead, a new layer, `spawn_ready` is inserted around endpoint stack to
ensure that readiness is driven on a background task.

In support of this change, the `pending` layer was removed from the
enpdoint stack and, instead, the discovery system is now responsible for
driving pending services to be materialized.


hawkw:
Looks like CI is angry for `rustfmt` reasons.

src/app/main.rs
```diff
@@ -484,6 +484,28 @@ where
                 .layer(strip_header::response::layer(super::L5D_REMOTE_IP))
                 .service(client_stack);
 
+            let balancer = svc::builder()
```
hawkw: Is there a reason this was moved higher up in the file, or is this change unrelated?

src/proxy/resolve.rs
```diff
@@ -58,13 +58,18 @@ pub struct MakeSvc<R, M> {
 
 /// Observes an `R`-typed resolution stream, using an `M`-typed endpoint stack to
 /// build a service for each endpoint.
-#[derive(Clone, Debug)]
-pub struct Discover<R, M> {
+pub struct Discover<R: Resolution, M: svc::Service<R::Endpoint>> {
     resolution: R,
     make: M,
+    pending: FuturesUnordered<MakeFuture<M::Future>>,
```
hawkw: Using `FuturesUnordered` as a scheduler here is smart 👍

src/proxy/resolve.rs
```diff
@@ -129,25 +136,28 @@ where
...
 {
     type Key = SocketAddr;
-    type Service = M::Value;
+    type Service = M::Response;
     type Error = Error;
 
     fn poll(&mut self) -> Poll<Change<Self::Key, Self::Service>, Self::Error> {
         loop {
+            if let Async::Ready(Some((addr, svc))) = self.pending.poll().map_err(Into::into)? {
+                self.is_empty.store(false, Ordering::Release);
```
hawkw: If my understanding is correct, this means that we will only consider the `Discover` "ready" when it has actually built at least one service from its discovery updates, rather than considering it ready as soon as an `Add` update is received?

src/proxy/resolve.rs
```diff
@@ -129,25 +136,28 @@ where
...
-                    // insertions of new endpoints and metadata changes for
-                    // existing ones can be handled in the same way.
-                    let svc = self.make.make(&target);
-                    self.is_empty.store(false, Ordering::Release);
-                    return Ok(Async::Ready(Change::Insert(addr, svc)));
+                    let inner = self.make.call(target);
+                    self.pending.push(MakeFuture { addr, inner });
                 }
                 Update::Remove(addr) => return Ok(Async::Ready(Change::Remove(addr))),
                 Update::NoEndpoints => {
```
hawkw: When we see a `NoEndpoints` update, should we cancel any pending services?

olix0r: This is an extremely good point ;/ I think we'd need something basically like what's described in https://github.com/tower-rs/tower/issues/292... I can probably come up with an approach that doesn't require this futures-unorderd until we have the proper stream data structure...

hawkw: If the balancer always polls discovery until it recieves a NotReady, I don't think the pending services are actually *leaked* --- we should see `Update::Remove` events for them & remove them from the balancer. But, it seems inefficient to drive them to readiness when they'll immediately be torn down...

olix0r: Not really... If we have an Add(A), put it in the pending, Remove(A), and then the pending becomes ready, we'll fire the add out-of-order!

hawkw: oh. yeah, you're right. that's potentially leaky for sure.

src/proxy/resolve.rs
```diff
@@ -160,8 +190,74 @@ where
...
+    type Error = F::Error;
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        loop {
+            return match self.futures.poll() {
+                Err(MakeError::Canceled) => continue,
+                Err(MakeError::Inner(err)) => Err(err),
+                Ok(Async::Ready(Some((addr, svc)))) => {
+                    let _rm = self.cancellations.remove(&addr);
+                    debug_assert!(_rm.is_some(), "cancellation missing");
```
hawkw: TIOLI: since we have the address anyway, maybe
```suggestion
                    debug_assert!(_rm.is_some(), "cancellation missing for {:?}", addr);
```
would be worth including? could make debugging easier if we ever actually hit this assertion...

src/proxy/resolve.rs
```diff
@@ -58,13 +59,29 @@ pub struct MakeSvc<R, M> {
 
 /// Observes an `R`-typed resolution stream, using an `M`-typed endpoint stack to
 /// build a service for each endpoint.
-#[derive(Clone, Debug)]
-pub struct Discover<R, M> {
+pub struct Discover<R: Resolution, M: svc::Service<R::Endpoint>> {
     resolution: R,
     make: M,
+    makes: MakeStream<M::Future>,
```
hawkw: TIOLI: personally, I find `makes` a bit weird (and, I notice, just one typo away from the similarly-named `make`. WDYT of `make_futures` or `make_stream` instead? it's admittedly a bit longer.

src/svc.rs
```diff
@@ -38,7 +45,12 @@ impl<L> Builder<L> {
         D: buffer::Deadline<Req>,
         Req: Send + 'static,
     {
-        self.layer(buffer::layer(bound, d)).layer(pending::layer())
+        self.layer(buffer::layer(bound, d)).pending()
+    }
+
+    /// Buffer requests when when the next layer is out of capacity.
```
kleimkuhler: This comment is the same as `fn pending`. Should this be describing a layer that ensures readiness is driven on the inner service?

kleimkuhler: This comment is the same as fn pending. Should this be describing a layer that ensures readiness is driven on the inner service?

src/proxy/resolve.rs
```diff
@@ -107,61 +125,442 @@ where
...
+    M::Error: Into<Error>,
 {
-    type Key = SocketAddr;
-    type Service = M::Value;
-    type Error = Error;
-
-    fn poll(&mut self) -> Poll<Change<Self::Key, Self::Service>, Self::Error> {
+    fn poll_resolution(&mut self) -> Poll<Change<SocketAddr, M::Response>, Error> {
         loop {
+            try_ready!(self.make.poll_ready().map_err(Into::into));
```
kleimkuhler: We need to check for `make`'s readiness, because this is the readiness of the Discover for the endpoint stack? Once we know that is readiness, we can poll for updates?

src/proxy/resolve.rs
```diff
@@ -107,61 +125,442 @@ where
...
+
             let up = try_ready!(self.resolution.poll().map_err(Into::into));
```
kleimkuhler: Nit: I think calling this `update` would be a helpful change. This may be confused for meaning that the `resolution` stream is "up" -- or something like that.

src/proxy/resolve.rs
```diff
@@ -58,13 +60,29 @@ pub struct MakeSvc<R, M> {
...
+pub struct Discover<R: Resolution, M: svc::Service<R::Endpoint>> {
     resolution: R,
     make: M,
+    makes: MakeStream<M::Future>,
     is_empty: Arc<AtomicBool>,
 }
 
+struct MakeStream<F> {
+    futures: FuturesUnordered<MakeFuture<F>>,
+    cancelations: IndexMap<SocketAddr, oneshot::Sender<()>>,
```
kleimkuhler: Typo in field name

src/proxy/resolve.rs
```diff
@@ -107,61 +125,442 @@ where
...
+                .send(Svc(vec![rsp1_rx]))
+                .expect("make must receive service");
+            match discover.poll().expect("discover can't fail") {
+                Async::NotReady => panic!("not processed"),
+                Async::Ready(Change::Remove(..)) => panic!("unexpected remove"),
+                Async::Ready(Change::Insert(a, mut svc)) => {
+                    assert_eq!(a, addr1);
+
+                    assert!(svc.poll_ready().unwrap().is_ready());
+                    let mut fut = svc.call(());
```
kleimkuhler: Checking my understanding here: Would this `svc.call(())` pop the last element out of the Vec -- which is `make0_rx`? I'm looking at how `fn call` is implicted for `Svc`, and see that we just pop last element off of the Vec of services. It was put in though as `make1_rx` and then `make0_rx`

src/proxy/resolve.rs
hawkw: I _definitely_ spelled `cancellations` right --- it looks like a80396b6c17f1e4e871bb252488e31582c7b61b0 broke this :P

olix0r: Both spellings are correct

olix0r: Notice, for instance that we use "Canceled" elsewhere in this file. A quick search says that the single el is an Americanism. Let it henceforth be policy that we use Americanized spelling in Linkerd.

src/proxy/resolve.rs
olix0r: The discover.poll() returns the service from `make1_rx1`... Then, this `call` returns `rsp1_rx`.

src/proxy/resolve.rs
olix0r: I've added a comment:
```
            // Before polling the resolution, where we could potentially receive
            // an `Add`, poll_ready to ensure that `make` is ready to build new
            // services. Don't process any updates until we can do so.
```

src/proxy/resolve.rs
olix0r: Enter my confusion around `is_empty` and what it is intended to convey...

I think you're right about this change in behavior, though, yes.

src/app/main.rs
olix0r: unrelated, mostly to make it easier to comment pieces out

src/proxy/resolve.rs
hawkw: TIL that single-el is valid!

src/proxy/resolve.rs
```diff
@@ -107,61 +125,440 @@ where
...
-    fn poll(&mut self) -> Poll<Change<Self::Key, Self::Service>, Self::Error> {
+    fn poll_resolution(&mut self) -> Poll<Change<SocketAddr, M::Response>, Error> {
         loop {
-            let up = try_ready!(self.resolution.poll().map_err(Into::into));
-            trace!("watch: {:?}", up);
-            match up {
+            // Before polling the resolution, where we could potentially receive
+            // an `Add`, poll_ready to ensure that `make` is ready to build new
+            // services. Don't process any updates until we can do so.
+            try_ready!(self.make.poll_ready().map_err(Into::into));
```
hawkw: Is this fixing one of the potential race conditions you mentioned?

src/proxy/resolve.rs
```diff
@@ -107,61 +125,440 @@ where
...
-                    // by replacing the old endpoint with the new one, so
-                    // insertions of new endpoints and metadata changes for
-                    // existing ones can be handled in the same way.
-                    let svc = self.make.make(&target);
-                    self.is_empty.store(false, Ordering::Release);
-                    return Ok(Async::Ready(Change::Insert(addr, svc)));
+                    // Start building the service and continue.
+                    // Remove any pending additions for this addr so that they
+                    // cannot race.
+                    self.make_futures.remove(&addr);
```
hawkw: TIOLI: Seems like this should just be part of `MakeFutures::push` to me --- the main reason I made it a separate struct rather than a new field on `Discover` was that it seemed neater to bundle all the logic for managing the pending service futures together...

src/proxy/resolve.rs
```diff
@@ -107,61 +125,440 @@ where
...
+    fn new() -> Self {
+        Self {
+            futures: FuturesUnordered::new(),
+            cancelations: IndexMap::new(),
+        }
+    }
+
+    fn push(&mut self, addr: SocketAddr, inner: F) {
+        let (cancel, canceled) = oneshot::channel();
+        self.cancelations.insert(addr, cancel);
```
hawkw: Also, in re: my earlier comment about moving the removal of a previous pending future to `MakeFutures::push`, it also saves us a hashmap lookup, since `IndexMap::insert` returns the prior value if there was one. We could just say
```rust
if let Some(cancel_prior) = self.cancelations.insert(addr, cancel) {
    let _ = cancel_prior.send(());
}
```

This is not too hot so that kind of optimization may not matter a whole lot, but it seems neater overall to me...

src/proxy/resolve.rs
hawkw: you may be pleased to hear that all the `is_empty` stuff is going away in #268

src/proxy/resolve.rs
olix0r: The fix was mostly about factoring out poll_resolution and calling it before polling the pending futures.

Previously, when the pending services were polled first, you could get into odd states when a cancelation had been fired but we didn't actually poll pending services to drive the cancelation.

Also, it was necessary to remove existing makes on Add so that we could guard against the cases in the tests.

src/proxy/resolve.rs
```diff
@@ -107,61 +125,440 @@ where
...
+
+// === impl EndpointStatus ===
+
 impl EndpointStatus {
     pub fn is_empty(&self) -> bool {
         self.0.load(Ordering::Acquire)
     }
 }
+
+// === impl MakeFutures MakeFutures===
```
kleimkuhler: Extra MakeFutures here

olix0r:
Docker images pushed as `config.linkerd.io/proxy-version: ver-update-tower-balance-3`

olix0r:
```
:; linkerd stat -n strest-new-balance po --to po -t 600s                   
NAME                                                  MESHED   SUCCESS         RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99   TCP_CONN
strest-client-edge-19-7-3-655bccb976-swzgc               1/1   100.00%   1867.8rps           1ms           3ms          32ms          6
strest-client-update-tower-balance-5f84fd9ff7-cz882      1/1   100.00%   2052.4rps           1ms           3ms          19ms          6
```

olix0r:
I've also tried this without the spawn-ready layer for comparison. Throughput/latency is roughly the same, but spawn-ready does appear to reduce CPU utilization:
```
:; kubectl top po --containers -n strest-new-balance |awk '$1 ~ /^strest-client-/ && $2 ~ /^linkerd-proxy$/ { print $0 }'
strest-client-edge-19-7-3-655bccb976-nq8zl            linkerd-proxy   68m          2Mi             
strest-client-no-spawn-ready-79ff7b98fd-9hhp7         linkerd-proxy   75m          3Mi             
strest-client-update-tower-balance-5f84fd9ff7-ktpmn   linkerd-proxy   59m          2Mi
:; linkerd stat -n strest-new-balance po --to po
NAME                                                  MESHED   SUCCESS        RPS   LATENCY_P50   LATENCY_P95   LATENCY_P99   TCP_CONN
strest-client-edge-19-7-3-655bccb976-nq8zl               1/1   100.00%   138.9rps           1ms         464ms         888ms          6
strest-client-no-spawn-ready-79ff7b98fd-9hhp7            1/1   100.00%   161.2rps           1ms         407ms         816ms          6
strest-client-update-tower-balance-5f84fd9ff7-ktpmn      1/1   100.00%   160.1rps           1ms         403ms         855ms          6

```

PR #267, opened 2019-06-07
Honor `l5d-override-dst` for inbound service profiles

hawkw:
When the proxy receives an inbound request, it strips the
`l5d-dst-override` header and ignores its value. Instead, this header
value should be honored in the inbound router as it is in the outbound
router. This will enable service profile discovery from ingresses, for
instance.

This branch changes the inbound router to use the `l5d-override-dst`
header preferentially, with only the `l5d-canonical-dst` header taking
priority over it if it is present. The layer that strips the override
header is moved to under the inbound dst router.

In addition, I've added new tests in the discovery module that assert
that the value of the `l5d-override-dst` header is used for profile
discovery when present, on both the inbound and outbound sides of the
proxy. The inbound tests fail prior to the other changes on this branch.

Closes linkerd/linkerd2#2910.

### DIFF
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -771,7 +771,7 @@ where
                             .or_else(|| super::http_request_authority_addr(req).ok())
                             .or_else(|| super::http_request_host_addr(req).ok())
                             .or_else(|| super::http_request_orig_dst_addr(req).ok());
-                            debug!("inbound dst={:?}", dst);
+                        debug!("inbound dst={:?}", dst);
 
                         dst.map(|addr| {
                             let settings = settings::Settings::from_request(req);
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -573,15 +573,18 @@ where
 
             // Routes requests to an `Addr`:
             //
-            // 1. If the request is HTTP/2 and has an :authority, this value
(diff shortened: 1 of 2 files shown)


src/app/main.rs
```diff
@@ -753,10 +760,19 @@ where
                         debug!("inbound canonical={:?}", canonical);
 
                         let dst = canonical
+                            .or_else(|| {
```
hawkw: This is the one thing in this PR I wasn't completely sure about --- should `l5d-override-dst` take priority over `l5d-canonical-dst`, or the other way around? 

In practice, I think we won't see both on the same inbound request, since `l5d-canonical-dst` is expected to only be set by outbound linkerd proxies, while `l5d-override-dst` is expected to be set only by clients that are _not_ linkerd proxies (as the outbound proxy will strip that header after using it). So, the ordering may not actually _matter_ all that much.

olix0r: Good question. i _think_ `canonical-dst` should take precedence, since it would only have been set by a calling proxy. If `l5d-dst-override` had also been set, they should in theory be identical (though the override could potentially be non-canonical? maybe?)

hawkw: @olix0r yup, that confirms what i was thinking! thanks.

PR #268, opened 2019-06-11
discovery: Fall back only on InvalidArgument, in MakeService

hawkw:
When the DNS fallback behavior was removed from the proxy in #259, the
proxy was changed to fall back to original destination routing _any_
time the control plane indicates that no endpoints exist for a
destination. This is incorrect, as the proxy should always treat the
destination service as authoritative. Additionally, this means that
requests to endpoints which are known to not exist will construct an
unnecessary client service, and eventually fail with a 502 error when
the upstream client connection to the non-existent endpoint ultimately
fails. Instead, the proxy should only fall back when the destination is
outside the search suffixes or the Destination service returns an
`InvalidArgument` response.

This is a much larger change than #263. In particular, the fallback
behavior has been moved from when the the actual client service is 
called to when the client service is _constructed_, as we do not expect
to recieve an `InvalidArgument` response on a Destination query that
has previously recieved updates. 

This was implemented by changing the `Resolve` trait to return a
`Future` whose `Item` type is a `Resolution`, rather than returning a
`Resolution`. When the Destination query returns `InvalidArgument`, the
future will fail. The `proxy::http::fallback` middleware has been
changed so that rather than making both the primary and fallback service
and falling back on a per-request basis, it tries to make the primary
service, and falls back if the primary `MakeService` future fails. 

Although this is a large change, I think the resulting code is simpler
and more elegant than the previous approach.

Previously, one of the discovery tests expected the incorrect behavior.
I've changed this test to now expect that destinations known to not
exist do _not_ fall back, and renamed it from
`outbound_falls_back_to_orig_dst_when_destination_has_no_endpoints` to
`outbound_fails_fast_when_destination_has_no_endpoints`. I've also
confirmed that the updated test fails against master and passes after
this change.

Fixes linkerd/linkerd2#2880
Closes #263 

### DIFF
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -429,7 +429,7 @@ where
                 //add_remote_ip_on_rsp, add_server_id_on_rsp,
             };
             use proxy::{
-                http::{balance, canonicalize, fallback, header_from_target, metrics, retry},
+                http::{balance, canonicalize, header_from_target, metrics, retry},
                 resolve,
             };
 
@@ -504,10 +504,6 @@ where
                 .layer(metrics::layer::<_, classify::Response>(retry_http_metrics))
                 .layer(insert::target::layer());
 
-            let balancer = svc::builder()
-                .layer(balance::layer(EWMA_DEFAULT_RTT, EWMA_DECAY))
-                .layer(resolve::layer(Resolve::new(resolver)));
-
             // Routes requests to their original destination endpoints. Used as
             // a fallback when service discovery has no endpoints for a destination.
             let orig_dst_router = svc::builder()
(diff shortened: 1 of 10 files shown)


src/app/main.rs
```diff
@@ -522,7 +522,10 @@ where
                 .layer(buffer::layer(max_in_flight, DispatchDeadline::extract));
 
             let balancer_stack = svc::builder()
-                .layer(fallback::layer(balancer, orig_dst_router))
+                .layer(
+                    fallback::layer(balancer, orig_dst_router)
+                        .on_error::<control::destination::Unresolvable>(),
+                )
```
olix0r: tioli: consider adding a `fallback_on` combinator to the svc::Builder so this could be...

```
￼                 .layer(balancer)
                 .fallback_on::<control::destination::Unresolvable>(orig_dst_router)
```

perhaps not worth it since it's only used once atm, but seems like a small readability improvement.

hawkw: oh, yeah, that is rather nice. good call!

hawkw:
@olix0r I've kicked off a lifecycle test to run overnight. Will report back.

olix0r:
@hawkw How does this interact with suffix matching? Do we throw an unresolveable error when the suffix doesn't match?

hawkw:
> @hawkw How does this interact with suffix matching? Do we throw an unresolveable error when the suffix doesn't match?

Yeah, that's right. If the requested name isn't in the search suffixes, `Resolver::resolve` returns `ResolveFuture::unresolvable`:
https://github.com/linkerd/linkerd2-proxy/blob/b5964b531d9686100334537e1103342d08305031/src/control/destination/mod.rs#L116-L125

`ResolveFuture::unresolvable` won't query the control plane, doesn't spawn a background daemon task, and will fail immediately with `Unresolvable` when polled.

hawkw:
I deployed a burn-in test overnight, and everything seems fine. I'm going to go ahead and merge this if there are no objections?

PR #270, opened 2019-06-17
Add concrete dst router

adleong:
Add a concrete dst router to the dst stack which routes over balancers based on the ConcreteDst request extension metadata.  The ConcreteDst extension is by the profiles router by reading the `dst_overrides` field of the service profile and picking a ConcreteDst from that list.  If the `dst_overrides` field is empty, no ConcereteDst is set and the concrete dst router will route based on the existing DstAddr.

The net effect of this is that service profile lookups will be done on the existing DstAddr (typically from the request authority) where as destination address lookups will be done on the concrete dst from the service profile's dst_overrides.  This allows the dst_overrides field to shift traffic to other destinations.


Cargo.toml
```diff
@@ -31,7 +31,7 @@ linkerd2-stack     = { path = "lib/stack" }
 linkerd2-task      = { path = "lib/task" }
 linkerd2-timeout   = { path = "lib/timeout" }
 
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", tag = "v0.1.8" }
+linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", rev = "ddbc3a4f7f8b0058801f896d27974d19ee98094c" }
```
hawkw: Should there have been a proxy-api tag published for this?

src/app/concrete_dst_router.rs
```diff
@@ -0,0 +1,206 @@
...
+use svc;
+use super::dst::DstAddr;
+use proxy::http::settings;
+use proxy::http::profiles::ConcreteDst;
+
+extern crate linkerd2_router as rt;
+
+pub use self::rt::{error, Router};
+
+#[allow(dead_code)]
```
hawkw: I'm guessing this `allow` is necessary as the compiler is not able to determine this is used? In that case, it would be nice to have a comment stating that.

src/app/concrete_dst_router.rs
```diff
@@ -0,0 +1,206 @@
...
+type Error = Box<dyn std::error::Error + Send + Sync>;
+
+#[derive(Clone, Debug)]
+pub struct Config {
```
hawkw: Nit: This looks like the same type as [`router::Config`](https://github.com/linkerd/linkerd2-proxy/blob/b70c68d4504a362eac6a7828039a2e5c7fcd308a/src/proxy/http/router.rs#L17-L21), maybe we should just reuse that?

src/app/main.rs
```diff
@@ -542,6 +543,10 @@ where
                     dst_route_layer,
                 ))
                 .buffer_pending(max_in_flight, DispatchDeadline::extract)
+                // Route to a balancer based on the ConcreteDst extention metadata
```
hawkw: Typo:
```suggestion
                // Route to a balancer based on the ConcreteDst extension metadata
```

src/app/profiles.rs
```diff
@@ -159,7 +159,14 @@ where
                         .routes
                         .into_iter()
                         .filter_map(move |orig| convert_route(orig, retry_budget.as_ref()));
-                    match tx.start_send(routes.collect()) {
+                    let overrides = profile
+                        .dst_overrides
+                        .into_iter()
+                        .filter_map(|orig| convert_dst_override(orig));
```
hawkw: Nit: I think this can use point-free style, like
```suggestion
                        .filter_map(convert_dst_override);
```

src/proxy/http/profiles.rs
```diff
@@ -482,25 +494,37 @@ pub mod router {
...
                 .as_mut()
                 .and_then(|ref mut s| s.poll().ok())
         }
+
+        fn pick_dst_override(&self) -> Option<NameAddr> {
+            if !self.dst_overrides.is_empty() {
+                let weights = self.dst_overrides.iter().map(|dst| dst.1);
+                let mut rng = rand::thread_rng();
+                let dist = WeightedIndex::new(weights).unwrap();
+                Some(self.dst_overrides[dist.sample(&mut rng)].0.clone())
```
hawkw: Nit/TIOLI: Personally, I find this expression a little complex and hard to follow. What about something like:

```rust
let idx = WeightedIndex::new(weights)
    .unwrap()
    .sample(&mut rng);
Some(self.dst_overrides[idx].0.clone())
```

src/proxy/http/profiles.rs
```diff
@@ -482,25 +494,37 @@ pub mod router {
...
         }
 
         fn poll_route_stream(&mut self) -> Option<Async<Option<Routes>>> {
             self.route_stream
                 .as_mut()
                 .and_then(|ref mut s| s.poll().ok())
         }
+
+        fn pick_dst_override(&self) -> Option<NameAddr> {
+            if !self.dst_overrides.is_empty() {
```
hawkw: Nit/TIOLI: I might structure this as 
```rust
if self.dst_overrides.is_empty() {
    return None;
}

// ... rest of method body
```

src/proxy/http/profiles.rs
```diff
@@ -16,7 +20,13 @@ use never::Never;
 use super::retry::Budget;
 use NameAddr;
 
-pub type Routes = Vec<(RequestMatch, Route)>;
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct ConcreteDst(pub NameAddr);
+
+pub struct Routes {
+    pub routes: Vec<(RequestMatch, Route)>,
+    pub dst_overrides: Vec<(NameAddr, u32)>,
```
hawkw: Nit/TIOLI: What do you think about changing these tuples to structs? Code consuming this might read better with named field accesses rather than `.1`/`.0`.

src/proxy/http/profiles.rs
```diff
@@ -16,7 +20,13 @@ use never::Never;
 use super::retry::Budget;
 use NameAddr;
 
-pub type Routes = Vec<(RequestMatch, Route)>;
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct ConcreteDst(pub NameAddr);
```
hawkw: Nit: I don't find this name terribly obvious, but I can't come up with any better suggestions OTTOMH. To me, `concrete` doesn't really convey "final destination after applying weighting" but I'm not sure what would.

src/proxy/http/profiles.rs
```diff
@@ -482,25 +494,37 @@ pub mod router {
...
             self.route_stream
                 .as_mut()
                 .and_then(|ref mut s| s.poll().ok())
         }
+
+        fn pick_dst_override(&self) -> Option<NameAddr> {
+            if !self.dst_overrides.is_empty() {
+                let weights = self.dst_overrides.iter().map(|dst| dst.1);
+                let mut rng = rand::thread_rng();
+                let dist = WeightedIndex::new(weights).unwrap();
```
seanmonstar: How often will this be called? If rather frequently, then we may care that `WeightedIndex::new` allocates an internal `Vec` each time, and instead just store that weighted vec instead? If not frequently, it won't matter then.

Also, same line: assuming the `unwrap` is because the error case can never happen, changing to `expect("why this error cannot happen")` can be more self-documenting.

hawkw: AFAICT, this is called per-request. I think caching the `WeightedIndex` is probably a good idea.

Perhaps we should build it in `update_routes`, since I think that's called any time the destination overrides would change?

src/app/concrete_dst_router.rs
olix0r: I don't think we should use the same type, as these routers are different (and will diverge more soon).  Specifically, why does this type need a capacity or max_idle_age? Don't we know the capacity from the dst overrides? We don't intend to idle these out, so a configurable max_idle_age is unnecessary...

src/app/concrete_dst_router.rs
```diff
@@ -0,0 +1,206 @@
...
+{
+    fn clone(&self) -> Self {
+        layer(self.config.clone())
+    }
+}
+
+// === impl Recognize ===
+
+impl<A> self::rt::Recognize<http::Request<A>> for Recognize {
+
```
olix0r: inconsistent lonely whitespace

src/proxy/http/profiles.rs
```diff
@@ -3,6 +3,10 @@ extern crate tower_discover;
 use futures::Stream;
 use http;
 use indexmap::IndexMap;
+// Import is used by WeightedIndex::sample.
+#[allow(unused_imports)]
```
olix0r: This shouldn't be necessary... let's figure this out before merging.

src/app/concrete_dst_router.rs
```diff
@@ -0,0 +1,206 @@
...
+use never::Never;
+
+use svc;
+use super::dst::DstAddr;
+use proxy::http::settings;
+use proxy::http::profiles::ConcreteDst;
+
+extern crate linkerd2_router as rt;
+
+pub use self::rt::{error, Router};
```
olix0r: Is it necessary to reexport these?

src/proxy/http/profiles.rs
```diff
@@ -16,7 +20,13 @@ use never::Never;
 use super::retry::Budget;
 use NameAddr;
 
-pub type Routes = Vec<(RequestMatch, Route)>;
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct ConcreteDst(pub NameAddr);
+
+pub struct Routes {
```
olix0r: This name feels slightly inappropriate now, since it's no longer just a list of routes...

src/proxy/http/profiles.rs
```diff
@@ -482,25 +494,37 @@ pub mod router {
...
 
         fn poll_route_stream(&mut self) -> Option<Async<Option<Routes>>> {
             self.route_stream
                 .as_mut()
                 .and_then(|ref mut s| s.poll().ok())
         }
+
+        fn pick_dst_override(&self) -> Option<NameAddr> {
+            if !self.dst_overrides.is_empty() {
+                let weights = self.dst_overrides.iter().map(|dst| dst.1);
```
olix0r: prefer `.map(|(_, svc)| svc)` destructuring to `.0/.1` references

src/app/profiles.rs
```diff
@@ -267,6 +274,12 @@ fn convert_route(
     Some((req_match, route))
 }
 
+fn convert_dst_override(orig: api::WeightedDst) -> Option<(NameAddr, u32)> {
+    NameAddr::from_str(orig.authority.as_str())
+        .ok()
+        .map(|addr| (addr, orig.weight))
+}
```
olix0r: I suggest that this (or another filter) drop any route with a 0-weight?

Or, rather, what is the expected behavior with zero-weights? What if all overrides have 0 weights? is that the same as having no weights at all?

src/app/main.rs
```diff
@@ -542,6 +543,10 @@ where
...
+                // if it has been set by the profiles router.
+                .layer(concrete_dst_router::layer(concrete_dst_router::Config::new("concrete dst", capacity, max_idle_age)))
```
kleimkuhler: Thoughts on putting this in `Builder` in `app/svc.rs` similar to `buffer_pending` right below? I'm a little surprised rustfmt is not adding line breaks, but it seems like something to consider.

src/app/concrete_dst_router.rs
adleong: My understanding is that there will be a concrete dst router per logical dst and these routers will not get rebuilt when the service profile changes.  so I don't think the capacity can be known ahead of time.

src/app/main.rs
hawkw: I think removing the `capacity` and `max_idle_age` (as @olix0r suggested in https://github.com/linkerd/linkerd2-proxy/pull/270#discussion_r294905445) from the `concrete_dst_router::Config` type would help a bit. 

If we removed those fields from `Config`, we would only take a name:
```rust
  .layer(concrete_dst_router::layer(concrete_dst_router::Config::new("concrete dst")))
```
In that case, we could probably just pass the name directly to `layer` and have _it_ construct the `Config`:
```rust
  .layer(concrete_dst_router::layer("concrete dst"))
```
but we may want to add more configs to `Config` later, so 🤷‍♀

adleong:
Closing in favor of #276

PR #273, opened 2019-06-18
Fix HTTP2 requests that don't have an :authority

seanmonstar:
Closes https://github.com/linkerd/linkerd2/issues/2942

### DIFF
--- a/src/proxy/http/h2.rs
+++ b/src/proxy/http/h2.rs
@@ -171,7 +171,21 @@ where
         self.tx.poll_ready().map_err(From::from)
     }
 
-    fn call(&mut self, req: http::Request<B>) -> Self::Future {
+    fn call(&mut self, mut req: http::Request<B>) -> Self::Future {
+        debug_assert_eq!(
+            req.version(),
+            http::Version::HTTP_2,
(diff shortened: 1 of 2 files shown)


seanmonstar:
I don't quite understand the question.

olix0r:
@seanmonstar when the outbound proxy receives an h2 request without an `:authority`, will an orig-proto header be set on such a request? What value would be set?

When an inbound proxy receives such a request, how will the request that is issued to the target server differ from the original request sent by the client?

seanmonstar:
> when the outbound proxy receives an h2 request without an :authority, will an orig-proto header be set on such a request?

No, orig-proto only upgrades outbound h1. If the proxy received an h2 request (with or without an :authority), it never goes through the orig-proto middleware.

olix0r:
@seanmonstar okay, so the http version mutation occurs inside of the client and not the server -- so linkerd is dealing with the h2 version until the request is completely handled by the proxy... I was unclear whether this change impacted the server or client side of the h2 stack.

seanmonstar:
Yes, this version mutation is a line before handing it over to hyper's client.

PR #274, opened 2019-06-19
Update lib directory to Rust 2018

seanmonstar:


lib/hyper-balance/src/lib.rs
```diff
@@ -38,15 +36,15 @@ where
...
-        let (parts, body) = rsp.into_parts();
-        let handle = if body.is_end_stream() {
-            drop(handle);
-            None
-        } else {
-            Some(handle)
-        };
-        let body = PendingUntilFirstDataBody { handle, body };
-        http::Response::from_parts(parts, body)
+        rsp.map(move |body| {
```
hawkw: purely out of interest: is the change from `into/from_parts` to `map` related to the change to Rust 2018?

seanmonstar: Nah, something I just noticed when looking through the file 🤷‍♂ .

PR #276, opened 2019-06-20
Concrete dst routing

adleong:
Add a concrete dst router to the profiles router to honor any override_dsts in the destination profile.


src/app/profiles.rs
```diff
@@ -159,7 +159,14 @@ where
...
                         .into_iter()
                         .filter_map(move |orig| convert_route(orig, retry_budget.as_ref()));
-                    match tx.start_send(routes.collect()) {
+                    let overrides = profile
+                        .dst_overrides
+                        .into_iter()
+                        .filter_map(|orig| convert_dst_override(orig));
+                    match tx.start_send(profiles::Routes {
+                        routes: routes.collect(),
+                        dst_overrides: overrides.collect(),
```
hawkw: Nit/TIOLI: I _might_ split the `Routes` initializer out into a separate `let` binding so this reads a little more linearly?

src/app/profiles.rs
```diff
@@ -159,7 +159,14 @@ where
                         .routes
                         .into_iter()
                         .filter_map(move |orig| convert_route(orig, retry_budget.as_ref()));
-                    match tx.start_send(routes.collect()) {
+                    let overrides = profile
+                        .dst_overrides
+                        .into_iter()
+                        .filter_map(|orig| convert_dst_override(orig));
```
hawkw: Nit/TIOLI: This can just be
```suggestion
                        .filter_map(convert_dst_override);
```

src/app/profiles.rs
```diff
@@ -267,6 +274,15 @@ fn convert_route(
     Some((req_match, route))
 }
 
+fn convert_dst_override(orig: api::WeightedDst) -> Option<profiles::WeightedAddr> {
+    NameAddr::from_str(orig.authority.as_str())
+        .ok()
+        .map(|addr| profiles::WeightedAddr {
+            addr,
+            weight: orig.weight,
```
hawkw: Should we test that the weight is non-zero here?

src/proxy/http/profiles.rs
```diff
@@ -295,49 +311,55 @@ pub mod router {
...
-    pub struct Recognize<T> {
+    pub struct RouteRecognize<T> {
         target: T,
-        routes: Routes,
+        routes: Vec<(RequestMatch, Route)>,
         default_route: Route,
     }
 
-    impl<B, T> rt::Recognize<http::Request<B>> for Recognize<T>
+    pub struct ConcreteDstRecognize<T> {
```
hawkw: Nit/TIOLI: WDYT of making these `RecognizeRoute`/`RecognizeConcreteDst`? Seems...verb-ier.

src/proxy/http/profiles.rs
```diff
@@ -357,12 +379,38 @@ pub mod router {
...
+                .iter()
+                .map(|dst| dst.weight)
+                .filter(|weight| *weight > 0)
+                .peekable();
+            if weights.peek().is_none() {
+                // Weights list is empty.
+                return Some(self.target.clone());
+            }
+            let mut rng = rand::thread_rng();
+            let idx = WeightedIndex::new(weights)
```
hawkw: Is constructing a new `WeightedIndex` expected to be costly? If so, should we move it out of the request path so that it's only rebuilt when routes change?

src/proxy/http/profiles.rs
```diff
@@ -392,19 +440,31 @@ pub mod router {
...
         <T as WithRoute>::Output: Eq + Hash + Clone,
-        M: rt::Make<T>,
-        M::Value: Clone,
+        M: rt::Make<T, Value = MSvc> + Clone,
+        MSvc: svc::Service<http::Request<MBody>> + Clone,
+        MSvc::Error: Into<Error>,
         G: GetRoutes,
-        R: svc::Layer<svc::shared::Shared<M::Value>, Service = RMk> + Clone,
+        R: svc::Layer<
+                svc::shared::Shared<rt::Router<http::Request<MBody>, ConcreteDstRecognize<T>, M>>,
```
hawkw: Nit/TIOLI: This seems like something that should maybe be type-aliased?

src/proxy/http/profiles.rs
```diff
@@ -473,23 +547,54 @@ pub mod router {
         }
     }
 
-    impl<G, T, R, B> Service<G, T, R, B>
+    impl<G, T, R, RMk, M, RBody, MBody>
+        Service<
+            G,
+            T,
+            R,
+            svc::shared::Shared<rt::Router<http::Request<MBody>, ConcreteDstRecognize<T>, M>>,
```
hawkw: oh cool here's that one type again 🙃

src/app/profiles.rs
adleong: We discard zero-weighted items when constructing the weighed distribution, so I don't think we need to test that here as well.

src/proxy/http/profiles.rs
adleong: neither way seems particularly better or worse than the other to me.  I can change it if you feel strongly.

src/proxy/http/profiles.rs
```diff
@@ -294,50 +311,69 @@ pub mod router {
         }
     }
 
+    type ConcreteRouter<T, Svc, B> =
```
kleimkuhler: Nit: It seems like we usually put type aliases at the top of the module, so could we put this above `fn layer()`?

src/proxy/http/profiles.rs
```diff
@@ -294,50 +311,69 @@ pub mod router {
...
+        M::Value: svc::Service<http::Request<MBody>> + Clone,
+        R: svc::Layer<S, Service = RMk>,
+        RMk: rt::Make<T::Output>,
+        RMk::Value: svc::Service<http::Request<RBody>> + Clone,
     {
         target: T,
-        stack: R,
+        inner: M,
+        route_layer: svc::Builder<R>,
         route_stream: Option<G>,
```
kleimkuhler: We call this field `route_stream` in `struct Service`, and `get_routes` in `MakeSvc`. I see it is an `Option<..>` here, but could we use the same name? I think it helps for readability.

src/proxy/http/profiles.rs
```diff
@@ -392,19 +472,30 @@ pub mod router {
         }
     }
 
-    impl<T, G, M, R, B, RMk, RSvc> svc::Service<T> for MakeSvc<G, M, R, B>
+    impl<T, G, M, R, RBody, MBody, MSvc, RMk, RSvc> svc::Service<T> for MakeSvc<G, M, R, RBody, MBody>
     where
-        T: CanGetDestination + WithRoute + Clone,
+        T: CanGetDestination + WithRoute + WithAddr + Eq + Hash + Clone,
```
kleimkuhler: I know this occurs a few places in the proxy code, but seeing this `where` makes me think we should try and have a comment explaining the difference at least between the the `R..` and non-`R..` type bounds. It wasn't very obvious to me when first working on this area of the code.

src/proxy/http/profiles.rs
```diff
@@ -413,17 +504,23 @@ pub mod router {
...
             let default_route = target.clone().with_route(self.default_route.clone());
 
             let mut make = IndexMap::with_capacity(1);
             make.insert(default_route.clone(), stack.make(&default_route));
 
-            // Create a new fixed router; we can eagerly make the services
-            // and never expire the routes from the profile router cache
-            let router = Router::new_fixed(
-                Recognize {
+            let router = rt::Router::new_fixed(
```
kleimkuhler: I know this method wasn't heavily commented to start, but keeping the existing comment about `new_fixed` and adding additional ones about why we need to make two routers now would be helpful.

lib/router/src/lib.rs
```diff
@@ -180,6 +180,10 @@ where
 
         router
     }
+
+    pub fn drain(self) -> IndexMap<Rec::Target, Svc> {
```
kleimkuhler: Nit: I would expect a `drain()` method to return an iterator. I think `get_make` or something similar could serve as a better name for what this is used for.

src/proxy/http/profiles.rs
```diff
@@ -475,33 +575,82 @@ pub mod router {
...
+        RMk: rt::Make<T::Output> + Clone,
+        RMk::Value: svc::Service<http::Request<RBody>> + Clone,
+        M: rt::Make<T> + Clone,
+        M::Value: svc::Service<http::Request<MBody>> + Clone,
     {
         fn update_routes(&mut self, routes: Routes) {
+            let capacity = routes.dst_overrides.len() + 1;
+
+            let mut make = IndexMap::with_capacity(capacity);
+            let mut drained = self
```
kleimkuhler: Similar to my comment about finding a better name for `Router::drain()`, I think this could use a change in name as well. We don't use this value as an iterator -- just a lookup for targets that we expect to be there.

src/app/profiles.rs
```diff
@@ -159,7 +159,14 @@ where
                         .routes
                         .into_iter()
                         .filter_map(move |orig| convert_route(orig, retry_budget.as_ref()));
-                    match tx.start_send(routes.collect()) {
+                    let overrides = profile
+                        .dst_overrides
+                        .into_iter()
+                        .filter_map(convert_dst_override);
+                    match tx.start_send(profiles::Routes {
+                        routes: routes.collect(),
```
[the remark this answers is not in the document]
adleong: cool, I didn't realize the type unifier was smart enough to know which `collect` to use without type annotations, but this totally works. :+1:

src/proxy/http/profiles.rs
adleong: I'm inclined to leave this alone as it wasn't touched in this PR.

src/proxy/http/profiles.rs
hawkw: Changing the type parameters from single letters to whole words makes the bounds _longer_, but WDYT of something like this:
```rust
    impl<T, Routes, InnerMk, RouteLayer, RouteBody, InnerBody, InnerSvc, RouteMk, RouteSvc> svc::Service<T> 
        for MakeSvc<Routes, InnerMk, RouteLayer, RouteBody, InnerBody>
    where
        T: CanGetDestination + WithRoute + WithAddr + Eq + Hash + Clone,
        <T as WithRoute>::Output: Eq + Hash + Clone,
        InnerMk: rt::Make<T, Value = InnerSvc> + Clone,
        InnerSvc: svc::Service<http::Request<InnerBody>> + Clone,
        InnerSvc::Error: Into<Error>,
        Routes: GetRoutes,
        RouteLayer: svc::Layer<svc::shared::Shared<ConcreteRouter<T, InnerMk::Value, InnerBody>>, Service = RouteMk>
            + Clone,
        RouteMk: rt::Make<<T as WithRoute>::Output, Value = RouteSvc> + Clone,
        RouteSvc: svc::Service<http::Request<RouteBody>> + Clone,
        RouteSvc::Error: Into<Error>,
    {
```

In particular, I think making the two "families" of types all share a common prefix might make this a little easier to understand

src/proxy/http/profiles.rs
```diff
@@ -392,19 +472,30 @@ pub mod router {
...
+        T: CanGetDestination + WithRoute + WithAddr + Eq + Hash + Clone,
         <T as WithRoute>::Output: Eq + Hash + Clone,
-        M: rt::Make<T>,
-        M::Value: Clone,
+        M: rt::Make<T, Value = MSvc> + Clone,
+        MSvc: svc::Service<http::Request<MBody>> + Clone,
+        MSvc::Error: Into<Error>,
         G: GetRoutes,
-        R: svc::Layer<svc::shared::Shared<M::Value>, Service = RMk> + Clone,
+        R: svc::Layer<svc::shared::Shared<ConcreteRouter<T, M::Value, MBody>>, Service = RMk>
```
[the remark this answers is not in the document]
adleong: While this is shorter, adding a type alias here adds a layer of indirection that makes this harder for me to understand.  I'll defer to those who know the style better than me, though.

tests/support/controller.rs
```diff
@@ -15,6 +15,10 @@ pub fn new() -> Controller {
     Controller::new()
 }
 
+pub fn new_unordered() -> Controller {
```
kleimkuhler: Was this added for the case of an applied traffic split when we don't know which leaf service the request is routed to, but do know that which ever one was chosen there should be a destination lookup for it?

src/proxy/http/profiles.rs
kleimkuhler: I still think longer type parameter names could be helpful for the short term, but don't like we need to block on it!

tests/profiles.rs
```diff
@@ -340,3 +340,127 @@ fn timeout() {
...
+    // 6. Send `n` requests to apex service
+    for _ in 0..n {
+        let rsp = client.get("/traffic-split");
+        assert!(rsp == "leaf-a" || rsp == "leaf-b");
+    }
+
+    // 7. Apex proxy metrics should assert there are >0 responses
+    assert!(leaf_a_responses.load(Ordering::SeqCst) > 0);
+    assert!(leaf_b_responses.load(Ordering::SeqCst) > 0);
+    assert_eq!(apex_responses.load(Ordering::SeqCst), 0);
```
kleimkuhler: It may be helpful to add one more set of steps to this test that tests changing an existing traffic split. If a new one is applied that sends all traffic to Leaf A, we could assert that `leaf_a_responses == n` and `leaf_b_responses == 0`. Also, if we just remove the traffic split we could instead assert steps 2 and 3.

src/proxy/http/profiles.rs
```diff
@@ -413,24 +525,37 @@ pub mod router {
...
         fn call(&mut self, target: T) -> Self::Future {
-            let inner = self.inner.make(&target);
-            let stack = self.route_layer.clone().service(svc::shared(inner));
+            // Initially there are no dst_overrides, so build a concrete router
+            // with only the default target.
+            let mut make = IndexMap::with_capacity(1);
+            make.insert(target.clone(), self.inner.make(&target));
+
+            let concrete_router =
+                rt::Router::new_fixed(ConcreteDstRecognize::new(target.clone(), Vec::new()), make);
```
olix0r: tioli:  i prefer breaking up complex constructors into multiple lines so they're easier to modify later:

```rust
let concrete_router = {
  let rec = ConcreteDstRecognize(...)
  rt::Router::new_fixed(rec, make)
};
```

tests/profiles.rs
```diff
@@ -340,3 +340,127 @@ fn timeout() {
         }
     }
 }
+
+#[test]
+fn traffic_split() {
```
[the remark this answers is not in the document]
olix0r: Also, for the sake of internal consistency, I’d prefer that we refer to dst overrides and not traffic split

adleong:
I have filed https://github.com/linkerd/linkerd2/issues/3005 and https://github.com/linkerd/linkerd2/issues/3006 as follow-ups and will address the remainder of the comments in this PR.

tests/profiles.rs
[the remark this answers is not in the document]
kleimkuhler: Metrics are polled for asserting the profile has been loaded. The response counts are used for actually testing the behavior we expect with a destination override applied. If there is a less flaky way to assert a profile has been loaded metrics can be completely removed.

PR #277, opened 2019-06-21
Replace `log` and `env-logger` with `tracing` and `tracing-fmt`

hawkw:
Currently, the proxy uses the `log` crate for log records, and formats
them using `env_logger`. However, `tracing` is capable of providing
significantly richer scoped, contextual diagnostics, which are
especially valuable in asynchronous systems like the proxy.

In addition, switching to `tracing` is a prerequisite for
dynamically reloading the log level (see linkerd/linkerd2#2005).

This branch replaces `env_logger` with `tracing-fmt`, and the `log`
crate's macros with `tracing`. Log records emitted by our
dependencies can still be consumed as `tracing` events using the
`tracing-log` adapter. A custom `tracing-fmt` formatter is used
to replicate the previous log line formatting; this currently uses the
proxy's log contexts in addition to `tracing` spans. Our homemade
log contexts will be phased out and replaced with `tracing` spans in
subsequent PRs, and the events emitted in the proxy will be made more
structured.

Closes linkerd/linkerd2#2004.

### DIFF
--- a/lib/router/src/cache.rs
+++ b/lib/router/src/cache.rs
@@ -2,10 +2,10 @@ use std::{hash::Hash, time::Duration};
 
 use futures::{try_ready, Async, Future, Poll, Stream};
 use indexmap::IndexMap;
-use log::debug;
+
 use tokio::sync::lock::Lock;
 use tokio_timer::{delay_queue, DelayQueue, Error, Interval};
-
+use tokio_trace::debug;
 /// An LRU cache that can eagerly remove values in a background task.
 ///
 /// Assumptions:
--- a/lib/router/src/cache.rs
+++ b/lib/router/src/cache.rs
@@ -6,6 +6,7 @@ use indexmap::IndexMap;
 use tokio::sync::lock::Lock;
 use tokio_timer::{delay_queue, DelayQueue, Error, Interval};
 use tokio_trace::debug;
+
 /// An LRU cache that can eagerly remove values in a background task.
 ///
 /// Assumptions:
--- a/lib/router/src/cache.rs
+++ b/lib/router/src/cache.rs
@@ -5,7 +5,7 @@ use indexmap::IndexMap;
 
 use tokio::sync::lock::Lock;
 use tokio_timer::{delay_queue, DelayQueue, Error, Interval};
-use tokio_trace::debug;
+use tracing::debug;
 
 /// An LRU cache that can eagerly remove values in a background task.
 ///
--- a/lib/router/src/cache.rs
+++ b/lib/router/src/cache.rs
@@ -3,9 +3,9 @@ use std::{hash::Hash, time::Duration};
 use futures::{task, try_ready, Async, Future, Poll, Stream};
 use indexmap::IndexMap;
 
-use tracing::trace;
 use tokio::sync::lock::Lock;
 use tokio_timer::{delay_queue, DelayQueue};
+use tracing::trace;
 
 /// An LRU cache that can eagerly remove values in a background task.
 ///
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -10,6 +10,7 @@ use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::runtime::current_thread;
 use tokio_timer::clock;
 use tower_grpc as grpc;
+use trace::futures::Instrument;
 
 use app::classify::{self, Class};
 use app::metric_labels::{ControlLabels, EndpointLabels, RouteLabels};
@@ -885,15 +886,17 @@ where
 
     let future = log.future(bound_port.listen_and_fold(
         (),
-        move |(), (connection, remote_addr)| {
-            let s = server.serve(connection, remote_addr, h2_settings);
+        move |(), (connection, remote)| {
+            let s = server.serve(connection, remote, h2_settings)
+                .instrument(info_span!("conn", %remote));
             // Logging context is configured by the server.
             let r = DefaultExecutor::current()
                 .spawn(Box::new(s))
                 .map_err(task::Error::into_io);
             future::result(r)
         },
-    ));
+    ))
+    .instrument(info_span!("server", server = %proxy_name, local = %listen_addr));
 
     let accept_until = Cancelable {
         future,
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -10,7 +10,6 @@ use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::runtime::current_thread;
 use tokio_timer::clock;
 use tower_grpc as grpc;
-use trace::futures::Instrument;
 
 use app::classify::{self, Class};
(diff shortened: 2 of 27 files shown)


lib/router/src/cache.rs
```diff
@@ -2,10 +2,10 @@ use std::{hash::Hash, time::Duration};
 
 use futures::{try_ready, Async, Future, Poll, Stream};
 use indexmap::IndexMap;
-use log::debug;
+
 use tokio::sync::lock::Lock;
 use tokio_timer::{delay_queue, DelayQueue, Error, Interval};
-
+use tokio_trace::debug;
```
kleimkuhler: Nit: Can we keep a newline between the use declarations and the doc comments?

src/app/main.rs
```diff
@@ -883,17 +883,18 @@ where
...
     let log = server.log().clone();
 
-    let future = log.future(bound_port.listen_and_fold(
-        (),
-        move |(), (connection, remote_addr)| {
-            let s = server.serve(connection, remote_addr, h2_settings);
+    let future = log.future(
+        bound_port.listen_and_fold((), move |(), (connection, remote)| {
+            let s = server.serve(connection, remote, h2_settings);
+            // .instrument(info_span!("conn", %remote));
```
kleimkuhler: Is this `.instrument()` a TODO or can it be removed?

src/app/main.rs
```diff
@@ -883,17 +883,18 @@ where
...
             // Logging context is configured by the server.
             let r = DefaultExecutor::current()
                 .spawn(Box::new(s))
                 .map_err(task::Error::into_io);
             future::result(r)
-        },
-    ));
+        }),
+    );
+    // .instrument(info_span!("proxy", server = %proxy_name, local = %listen_addr));
```
kleimkuhler: Same as above

src/app/mod.rs
```diff
@@ -25,9 +25,10 @@ const L5D_SERVER_ID: &'static str = "l5d-server-id";
 const L5D_CLIENT_ID: &'static str = "l5d-client-id";
 
 pub fn init() -> Result<config::Config, config::Error> {
-    use logging;
+    use trace;
 
-    logging::init();
+    // logging::init();
```
kleimkuhler: Looks like this can be removed

src/app/profiles.rs
```diff
@@ -197,10 +197,10 @@ where
                             Ok(Async::Ready(())) => svc.as_service(),
                             Ok(Async::NotReady) => return Ok(Async::NotReady),
                             Err(err) => {
+                                let err = err.into();
```
kleimkuhler: Curious if there is a reason the `err.into()` was moved out of `error!()`?

src/logging.rs
```diff
@@ -18,38 +14,105 @@ thread_local! {
...
+        let dispatch = Dispatch::new(
+            subscriber_builder()
+                .with_filter(filter::EnvFilter::from(filter))
+                .finish(),
+        );
+        dispatcher::set_global_default(dispatch)
+            .map_err(|_| "failed to set global default dispatcher")?;
+        let logger = tokio_trace_log::LogTracer::with_filter(log::LevelFilter::max());
+        log::set_boxed_logger(Box::new(logger)).map_err(|_| "failed to set global logger")?;
+        log::set_max_level(log::LevelFilter::max());
```
kleimkuhler: Is `set_max_level` still needed with the `with_filter(log::LevelFilter::max())` on :52?

src/proxy/http/router.rs
```diff
@@ -128,12 +130,17 @@ where
             self.config.capacity,
             self.config.max_idle_age,
         );
-
+        let span = debug_span!(
+            "router",
+            name = self.config.proxy_name,
+            // config.capacity = self.config.capacity,
```
kleimkuhler: Do these lines need to stick around or be marked as TODOs?

hawkw:
@kleimkuhler interesting, looks like CI is hitting the same problem. I _thought_ i pushed a working lockfile, let me double check.

src/logging.rs
hawkw: Yes, they are unrelatedd. `set_max_level` configures the _global_ max level for the `log` crate; `with_filter` configures a filter for the `LogTracer` adapter (which is not necessary as the trace subscriber manages filtering).

src/app/mod.rs
hawkw: yup, my bad.

src/app/main.rs
hawkw: This is a TODO for a future PR that will replace our homebrew log contexts with tokio-trace spans. I'll leave a comment here as well.

src/app/profiles.rs
hawkw: Yes, although it's no longer necessary. In an earlier iteration of this branch where we used the `tokio-trace` log back-compatibility feature to emit `log` records, `into()` would be called twice, once when the expression was used in the log format string and again when it was used as a `tokio-trace` field. We're no longer using the log compat feature so this can be removed.

lib/router/src/cache.rs
hawkw: I think this is rustfmt, but I'll double check.

src/logging.rs
```diff
@@ -18,38 +14,105 @@ thread_local! {
...
+        Box<
+            Fn(&Context<default::NewRecorder>, &mut fmt::Write, &Event) -> fmt::Result
+                + Send
+                + Sync
+                + 'static,
+        >,
+        filter::EnvFilter,
+    >;
+
+    pub fn init() {
```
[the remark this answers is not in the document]
hawkw: Made this fallible in f5d5595ce4d4ed2935c3190c2b0cae21c8751bf5

hawkw:
> I found the `SubscriberBuilder` and `FmtCtx` in src/logging.rs to not be that intuitive. Is that mostly because they are formatting the logs in a way that will not have a noticeable change with switching to trace?

@kleimkuhler I gave the naming a pass and added some comments in f5d5595ce4d4ed2935c3190c2b0cae21c8751bf5; let me know if that's any clearer?

kleimkuhler:
> @kleimkuhler I gave the naming a pass and added some comments in [f5d5595](https://github.com/linkerd/linkerd2-proxy/commit/f5d5595ce4d4ed2935c3190c2b0cae21c8751bf5); let me know if that's any clearer?

Yep the comments are helpful -- thanks! I've approved already, but still looks good to me.

hawkw:
I've updated this branch to depend on `tracing` rather than `tokio-trace`.

olix0r:
@hawkw CI failure looks like fmt as well as some straggler env_logger uses in tests

hawkw:
Ah, it looks like the `env_logger` use in tests slipped back in during a rebase --- I thought I'd gotten rid of that!

PR #279, opened 2019-06-25
Add `/proxy-log-level` endpoint to admin server

hawkw:
The linkerd proxy's log level can be configured by annotation; but this
means that it's not possible to update the log level on a running
process. In order to facilitate manual debugging of a running system,
there should be a way to update the log level of an individual pod at
runtime. It must not be possible for unauthorized users to alter the log
level of a proxy, as this can have serious performance and security
implications.

This branch adds a `/proxy-log-level` endpoint to the proxy's admin
server. The endpoint responds to `GET` requests by returning the current
log level string, and `PUT` requests by attempting to update the log
level. If the provided log level string is invalid, it returns a 400
error, and if the log level is valid, it returns 201 Created.

The endpoint only accepts requests on the loopback interface, and fails
with a 403 when requests are received from non-loopback addresses.

Note that this requires PR #277, as it is implemented using 
`tokio-trace-fmt`.

Closes linkerd/linkerd2#2005

<details>
<summary>Demonstration:</summary>

```
$ LINKERD2_PROXY_IDENTITY_DISABLED=true LINKERD2_PROXY_CONTROL_ADDR=localhost:8000 LINKERD2_PROXY_LOG=info,linkerd2_proxy::app::admin=trace cargo run &
[1] 95828

$     Finished dev [unoptimized] target(s) in 0.66s
INFO [     0.013175s] linkerd2_proxy::app::main using destination service at None
INFO [     0.013345s] linkerd2_proxy::app::main identity is DISABLED: disabled
INFO [     0.013382s] linkerd2_proxy::app::main routing on V4(127.0.0.1:4140)
INFO [     0.013419s] linkerd2_proxy::app::main proxying on V4(0.0.0.0:4143) to None
INFO [     0.013447s] linkerd2_proxy::app::main serving admin endpoint metrics on V4(127.0.0.1:4191)
INFO [     0.013474s] linkerd2_proxy::app::main protocol detection disabled for inbound ports {25, 3306}
INFO [     0.013545s] linkerd2_proxy::app::main protocol detection disabled for outbound ports {25, 3306}

$ curl localhost:4191/proxy-log-level

TRCE [    23.347619s] admin={server=admin listen=127.0.0.1:4191 remote=127.0.0.1:55600} admin{request.path=/proxy-log-level request.method=GET}: linkerd2_proxy::app::admin handling request
info,linkerd2_proxy::app::admin=trace%

$ curl localhost:4191/proxy-log-level -X PUT -d "info,linkerd2_proxy::app::admin=trace,linkerd2_proxy::logging=trace"

TRCE [    51.157556s] admin={server=admin listen=127.0.0.1:4191 remote=127.0.0.1:55630} admin{request.path=/proxy-log-level request.method=PUT}: linkerd2_proxy::app::admin handling request
TRCE [    51.157720s] admin={server=admin listen=127.0.0.1:4191 remote=127.0.0.1:55630} admin{request.path=/proxy-log-level request.method=PUT}: linkerd2_proxy::app::admin request.body="info,linkerd2_proxy::app::admin=trace,linkerd2_proxy::logging=trace"
INFO [    51.158078s] admin={server=admin listen=127.0.0.1:4191 remote=127.0.0.1:55630} admin{request.path=/proxy-log-level request.method=PUT}: linkerd2_proxy::logging::trace set new log level level=info,linkerd2_proxy::app::admin=trace,linkerd2_proxy::logging=trace

$ curl localhost:4191/proxy-log-level

TRCE [    72.513505s] admin={server=admin listen=127.0.0.1:4191 remote=127.0.0.1:55655} admin{request.path=/proxy-log-level request.method=GET}: linkerd2_proxy::app::admin handling request
info,linkerd2_proxy::logging=trace,linkerd2_proxy::app::admin=trace%

$ curl localhost:4191/proxy-log-level --head

TRCE [    31.791367s] admin={server=admin listen=127.0.0.1:4191 remote=127.0.0.1:55959} admin{request.path=/proxy-log-level request.method=HEAD}: linkerd2_proxy::app::admin handling request
HTTP/1.1 405 Method Not Allowed
allow: GET
allow: PUT
content-length: 0
date: Tue, 25 Jun 2019 21:38:59 GMT


$ curl localhost:4191/proxy-log-level -X PUT -d "info,linkerd2_proxy=debug"

TRCE [   133.613981s] admin={server=admin listen=127.0.0.1:4191 remote=127.0.0.1:55725} admin{request.path=/proxy-log-level request.method=PUT}: linkerd2_proxy::app::admin handling request
TRCE [   133.614121s] admin={server=admin listen=127.0.0.1:4191 remote=127.0.0.1:55725} admin{request.path=/proxy-log-level request.method=PUT}: linkerd2_proxy::app::admin request.body="info,linkerd2_proxy=debug"
INFO [   133.614341s] admin={server=admin listen=127.0.0.1:4191 remote=127.0.0.1:55725} admin{request.path=/proxy-log-level request.method=PUT}: linkerd2_proxy::logging::trace set new log level level=info,linkerd2_proxy=debug

$ curl localhost:4191/proxy-log-level -X PUT -d "i'm a bad log level, nothing should happen"

DBUG [   178.633996s] admin={server=admin listen=127.0.0.1:4191} linkerd2_proxy::transport::tls::listen accepted connection from 127.0.0.1:55777 to None; skipping TLS (disabled)
WARN [   178.634693s] admin={server=admin listen=127.0.0.1:4191 remote=127.0.0.1:55777} linkerd2_proxy::app::admin setting log level failed error=invalid filter directive 'i'm a bad log level'
invalid filter directive 'i'm a bad log level'%

$ fg
[1]  + 95828 running    LINKERD2_PROXY_IDENTITY_DISABLED=true LINKERD2_PROXY_CONTROL_ADDR= = cargo ru
^X^CINFO [   203.952663s] linkerd2_proxy::signal received SIGINT, starting shutdown
DBUG [   203.952778s] linkerd2_proxy::app::main shutdown signaled
DBUG [   203.953356s] linkerd2_proxy::app::main shutdown complete
```

</details>

### DIFF
--- a/src/app/admin/mod.rs
+++ b/src/app/admin/mod.rs
@@ -4,12 +4,12 @@
 //! * `/ready` -- returns 200 when the proxy is ready to participate in meshed traffic.
 
 
-use futures::future::{self, Future};
+use futures::future::{self, Future, FutureResult};
 use http::StatusCode;
 use hyper::{service::Service, Body, Request, Response};
 use std::io;
 
-use trace;
+use trace::{self, futures::Instrument};
 use metrics;
 use control::ClientAddr;
 
@@ -25,6 +25,11 @@ where
     trace_admin: trace::Admin,
     ready: Readiness,
 }
+
+pub type ResponseFuture = trace::futures::Instrumented<future::Either<
+    FutureResult<Response<Body>, io::Error>,
+    Box<Future<Item = Response<Body>, Error = io::Error> + Send + 'static>,
+>>;
 
 impl<M> Admin<M>
 where
@@ -51,6 +56,23 @@ where
                 .expect("builder with known status code must not fail")
         }
     }
+
+    fn proxy_log_level(&mut self, req: Request<Body>) -> ResponseFuture {
+        let is_loopback = req.extensions()
+            .get::<ClientAddr>()
+            .map(ClientAddr::is_loopback)
+            .unwrap_or(false);
+        if is_loopback {
+            future::Either::B(self.trace_admin.call(req))
+        } else {
+            future::Either::A(future::ok(
+                Response::builder()
+                    .status(StatusCode::FORBIDDEN)
+                    .body("access to /proxy-log-level only allowed from loopback addresses".into())
+                    .expect("builder with known status code must not fail")
+                ))
+        }
+    }
 }
 
 impl<M> Service for Admin<M>
@@ -63,31 +85,21 @@ where
     type Future = Box<Future<Item= Response<Body>, Error=Self::Error> + Send  + 'static>;
 
     fn call(&mut self, req: Request<Body>) -> Self::Future {
+        let path = req.uri().path();
+        let span = trace_span!("admin", request.path = %path, request.method = ?req.method());
+        let _enter = span.enter();
 
         match req.uri().path() {
-            "/metrics" => Box::new(self.metrics.call(req)),
-            "/proxy-log-level" => if req.extensions()
-                .get::<ClientAddr>()
-                .map(ClientAddr::is_loopback)
-                .unwrap_or(false)
-            {
-                Box::new(self.trace_admin.call(req))
-            } else {
-                Box::new(future::ok(
-                    Response::builder()
-                        .status(StatusCode::FORBIDDEN)
-                        .body("access to /proxy-log-level only allowed from loopback addresses".into())
-                        .expect("builder with known status code must not fail")
-                ))
-            },
+            "/metrics" => future::Either::A(self.metrics.call(req)),
+            "/proxy-log-level" => self.proxy_log_level(),
             "/ready" => Box::new(future::ok(self.ready_rsp())),
-            _ => Box::new(future::ok(
+            _ => future::Either::A(future::ok(
                 Response::builder()
                     .status(StatusCode::NOT_FOUND)
                     .body(Body::empty())
                     .expect("builder with known status code must not fail"),
             )),
-        }
+        }.instrument(span.clone())
     }
 }
 
--- a/src/app/admin/mod.rs
+++ b/src/app/admin/mod.rs
@@ -11,6 +11,7 @@ use std::io;
 
 use trace;
 use metrics;
+use control::ClientAddr;
 
 mod readiness;
 pub use self::readiness::{Latch, Readiness};
@@ -59,12 +60,26 @@ where
     type ReqBody = Body;
     type ResBody = Body;
     type Error = io::Error;
-    type Future =Box<Future<Item= Response<Body>, Error=Self::Error> + Send  + 'static>;
+    type Future = Box<Future<Item= Response<Body>, Error=Self::Error> + Send  + 'static>;
 
     fn call(&mut self, req: Request<Body>) -> Self::Future {
+
         match req.uri().path() {
             "/metrics" => Box::new(self.metrics.call(req)),
-            "/proxy-log-level" => Box::new(self.trace_admin.call(req)),
+            "/proxy-log-level" => if req.extensions()
+                .get::<ClientAddr>()
+                .map(ClientAddr::is_loopback)
+                .unwrap_or(false)
+            {
+                Box::new(self.trace_admin.call(req))
+            } else {
+                Box::new(future::ok(
+                    Response::builder()
+                        .status(StatusCode::FORBIDDEN)
+                        .body("access to /proxy-log-level only allowed from loopback addresses".into())
+                        .expect("builder with known status code must not fail")
+                ))
+            },
             "/ready" => Box::new(future::ok(self.ready_rsp())),
             _ => Box::new(future::ok(
                 Response::builder()
--- a/src/app/admin/mod.rs
+++ b/src/app/admin/mod.rs
@@ -3,15 +3,17 @@
 //! * `/metrics` -- reports prometheus-formatted metrics.
 //! * `/ready` -- returns 200 when the proxy is ready to participate in meshed traffic.
 
-
-use futures::future::{self, Future, FutureResult};
-use http::StatusCode;
+use futures::{
+    future::{self, Future, FutureResult},
+    Stream,
+};
+use http::{Method, StatusCode};
 use hyper::{service::Service, Body, Request, Response};
-use std::io;
+use std::{io, str};
 
-use trace::{self, futures::Instrument};
-use metrics;
 use control::ClientAddr;
+use metrics;
+use trace::{self, futures::Instrument, LevelHandle as TraceLevel};
 
 mod readiness;
 pub use self::readiness::{Latch, Readiness};
@@ -22,23 +24,25 @@ where
     M: metrics::FmtMetrics,
 {
     metrics: metrics::Serve<M>,
-    trace_admin: trace::Admin,
+    trace_level: TraceLevel,
     ready: Readiness,
 }
 
-pub type ResponseFuture = trace::futures::Instrumented<future::Either<
-    FutureResult<Response<Body>, io::Error>,
-    Box<Future<Item = Response<Body>, Error = io::Error> + Send + 'static>,
->>;
+pub type ResponseFuture = trace::futures::Instrumented<
+    future::Either<
+        FutureResult<Response<Body>, io::Error>,
+        Box<Future<Item = Response<Body>, Error = io::Error> + Send + 'static>,
+    >,
+>;
 
 impl<M> Admin<M>
 where
     M: metrics::FmtMetrics,
 {
-    pub fn new(m: M, trace_admin: trace::Admin, ready: Readiness) -> Self {
+    pub fn new(m: M, ready: Readiness, trace_level: TraceLevel) -> Self {
         Self {
             metrics: metrics::Serve::new(m),
-            trace_admin,
+            trace_level,
             ready,
         }
     }
@@ -56,23 +60,6 @@ where
                 .expect("builder with known status code must not fail")
         }
     }
-
-    fn proxy_log_level(&mut self, req: Request<Body>) -> ResponseFuture {
-        let is_loopback = req.extensions()
-            .get::<ClientAddr>()
-            .map(ClientAddr::is_loopback)
-            .unwrap_or(false);
-        if is_loopback {
-            future::Either::B(self.trace_admin.call(req))
-        } else {
-            future::Either::A(future::ok(
-                Response::builder()
-                    .status(StatusCode::FORBIDDEN)
-                    .body("access to /proxy-log-level only allowed from loopback addresses".into())
-                    .expect("builder with known status code must not fail")
-                ))
-        }
-    }
 }
 
 impl<M> Service for Admin<M>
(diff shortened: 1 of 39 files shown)


hawkw:
Okay, I've merged master and fixed the test support issues. Ideally, we'd also have automated tests for this functionality, and I can work on adding that as well.

src/logging.rs
```diff
@@ -97,6 +105,47 @@ pub mod trace {
...
+        /// do not exercise the `proxy-log-level` endpoint.
+        pub fn dangling() -> Self {
+            let builder = subscriber_builder()
+                .with_filter(filter::EnvFilter::default())
+                .with_filter_reloading();
+            let inner = builder.reload_handle();
+            LevelHandle { inner }
+        }
+
+        pub fn set_level(&self, level: impl AsRef<str>) -> Result<(), Error> {
```
olix0r: It's surprising that this isn't `&mut self`?

src/app/admin/mod.rs
```diff
@@ -55,22 +69,110 @@ where
...
                     .body(Body::empty())
                     .expect("builder with known status code must not fail"),
-            ),
+            )),
         }
     }
 }
 
+impl TraceLevel {
+    fn set_from(&self, chunk: hyper::Chunk) -> Result<(), String> {
```
olix0r: similarly, it's surprising that this doesn't take `&mut self`?

src/app/admin/mod.rs
```diff
@@ -55,22 +69,110 @@ where
...
-    type Future = FutureResult<Response<Body>, Self::Error>;
+    type Future = ResponseFuture;
 
     fn call(&mut self, req: Request<Body>) -> Self::Future {
+        let span = trace_span!(
+            "admin",
+            request.path = %req.uri().path(),
+            request.method = ?req.method(),
+        );
+        let _enter = span.enter();
```
olix0r: Would you mind splitting out the part of this change that adds spans to the admin endpoint to a followup?

src/control/serve_http.rs
```diff
@@ -21,8 +29,14 @@ where
         let log = log.clone();
         bound_port
             .listen_and_fold(Http::new(), move |hyper, (conn, remote)| {
+                let svc = service.clone();
```
olix0r: perhaps include a comment about why ClientAddr is set here..

src/control/serve_http.rs
```diff
@@ -41,3 +55,9 @@ where
 
     log.future(fut)
 }
+
+impl ClientAddr {
+    pub fn into_socket_addr(&self) -> SocketAddr {
+        self.0
+    }
+}
```
olix0r: Thoughts on this vs `impl Into<SocketAddr> for ClientAddr` ?

src/app/admin/mod.rs
```diff
@@ -55,22 +69,110 @@ where
...
+            "/metrics" => future::Either::A(self.metrics.call(req)),
+            "/proxy-log-level" => future::Either::B(self.trace_level.call(req)),
+            "/ready" => future::Either::A(future::ok(self.ready_rsp())),
+            _ => future::Either::A(future::ok(rsp(StatusCode::NOT_FOUND, Body::empty()))),
+        }
+        .instrument(span.clone())
+    }
+}
+
+impl Service for TraceLevel {
```
olix0r: Would you mind splitting TraceLevel into its own module, e.g. `app::admin::proxy_log_level`

src/control/serve_http.rs
hawkw: This began its life as `Into<SocketAddr>`, I recall there being type inference issues that required annotating it, so i thought I'd just add a function instead. It can be `Into` if you prefer.

src/logging.rs
hawkw: The `reload` method on `tracing-fmt::filter::reload::Handle` only requires `&self` because the actual filter state is behind a read-write lock; setting a new filter acquires the write lock. So, there's no _reason_ for this to take a mutable receiver.

hawkw:
@olix0r I believe I've made all the requested changes!

src/app/admin/mod.rs
```diff
@@ -3,32 +3,41 @@
...
 {
     metrics: metrics::Serve<M>,
+    trace_level: TraceLevel,
     ready: Readiness,
 }
 
+pub type ResponseFuture = future::Either<
+    FutureResult<Response<Body>, io::Error>,
+    Box<Future<Item = Response<Body>, Error = io::Error> + Send + 'static>,
+>;
```
olix0r: Do you think we should just make this all `Box<...>` without the either? I'm not particularly worried about boxing metrics responses...

src/app/admin/trace_level.rs
```diff
@@ -0,0 +1,89 @@
...
+    type ReqBody = Body;
+    type ResBody = Body;
+    type Error = io::Error;
+    type Future = Box<Future<Item = Response<Body>, Error = Self::Error> + Send + 'static>;
+
+    fn call(&mut self, req: Request<Body>) -> Self::Future {
+        // `/proxy-log-level` endpoint can only be called from loopback IPs
+        if let Some(addr) = req
+            .extensions()
+            .get::<ClientAddr>()
```
olix0r: fmt isn't happy about this

hawkw: whoops

src/app/admin/mod.rs
hawkw: I'm fine with boxing them all, my first pass at this branch did that. I just thought it would be trivial to avoid the box in the `/metrics` and `/ready` cases, so i went ahead and did it. happy to change it back.

src/logging.rs
```diff
@@ -97,6 +105,47 @@ pub mod trace {
...
+                .with_current(|f| format!("{}", f))
+                .map_err(Into::into)
+        }
+    }
+
+    impl fmt::Debug for LevelHandle {
+        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+            let current = self
+                .inner
+                // ugh...
```
olix0r: ???

src/logging.rs
```diff
@@ -97,6 +105,47 @@ pub mod trace {
...
+            let builder = subscriber_builder()
+                .with_filter(filter::EnvFilter::default())
+                .with_filter_reloading();
+            let inner = builder.reload_handle();
+            LevelHandle { inner }
+        }
+
+        pub fn set_level(&self, level: impl AsRef<str>) -> Result<(), Error> {
+            let level = level.as_ref();
+            let filter = level.parse::<filter::EnvFilter>()?;
```
olix0r: Where does this parsing code live? It seems like we should definitely have some testing around invalid input data.

src/app/admin/mod.rs
olix0r: we're clearly headed down the path of this signature getting more complex so i'd just simplify it to Box for now.

src/logging.rs
hawkw: The parsing code lives [here](https://github.com/tokio-rs/tracing/blob/0fe4d78d9f323079af1022dbc4d53dda87942134/tracing-fmt/src/filter/env.rs#L279-L379); there are [tests](https://github.com/tokio-rs/tracing/blob/0fe4d78d9f323079af1022dbc4d53dda87942134/tracing-fmt/src/filter/env.rs#L526-L804) later on in that file.

src/logging.rs
hawkw: ah, i had wanted to get rid of the unnecessary allocation from `format!` but i forgot. will fix it!

src/app/admin/trace_level.rs
```diff
@@ -0,0 +1,86 @@
...
+                let handle = self.clone();
+                let f = req
+                    .into_body()
+                    .concat2()
+                    .map(move |chunk| match handle.set_from(chunk) {
+                        Err(error) => {
+                            warn!(message = "setting log level failed", %error);
+                            rsp(StatusCode::BAD_REQUEST, format!("{}", error))
+                        }
+                        Ok(()) => rsp(StatusCode::CREATED, Body::empty()),
```
olix0r: Upon reflection, this probably isn't correct from the HTTP POV.  Since we're not creating a resource, this should probably be either a 200, if we want to round-trip the body; or 204 if we return without content.

hawkw: since the request is a `PUT`, i think it should probably just be a 204? but i am also not altogether up to date on my REST

src/logging.rs
hawkw: fixed in ccca5a

src/app/admin/trace_level.rs
hawkw: changed it to 204

src/logging.rs
```diff
@@ -17,32 +17,40 @@ thread_local! {
...
+
+    use std::{env, error, fmt, str, time::Instant};
     pub use tracing::*;
     pub use tracing_fmt::*;
 
     type SubscriberBuilder = Builder<default::NewRecorder, Format, filter::EnvFilter>;
     pub type Error = Box<error::Error + Send + Sync + 'static>;
 
+    #[derive(Clone)]
+    pub struct LevelHandle {
```
kleimkuhler: Could we call this `TraceLevel` so it doesn't need to be imported `as` in `src/app/admin/mod.rs`?

src/app/admin/trace_level.rs
```diff
@@ -0,0 +1,86 @@
+use futures::{
+    future::{self, Future},
+    Stream,
+};
+use http::{Method, StatusCode};
+use hyper::{service::Service, Body, Request, Response};
+use std::{io, net::SocketAddr, str};
+
+use control::ClientAddr;
+pub use trace::LevelHandle as TraceLevel;
```
kleimkuhler: Commenting here for my question about calling this `TraceLevel`

src/logging.rs
kleimkuhler: Meant to actually delete this comment on my review. After looking more at how `LevelHandle` is used here, I see the naming benefit of referring to it as a handle in this file. In other modules where it's easier to think of it just as the level makes sense.

src/app/admin/trace_level.rs
```diff
@@ -0,0 +1,86 @@
...
+
+use super::rsp;
```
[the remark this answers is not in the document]
hawkw: I _think_ `rustfmt` made them this way, but I'll double-check?

src/logging.rs
hawkw: I'm happy to just call this `TraceLevel` here, that might be clearer?

src/app/admin/trace_level.rs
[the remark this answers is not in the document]
hawkw: changed it in 553f4897cdeca554492f347271b3dd453d688257

src/logging.rs
kleimkuhler: I think keeping it `LevelHandle` is okay! `TraceLevel` is helpful in other files, but it's a handle everywhere in this file, so I'm happy with the naming.

src/app/admin/mod.rs
```diff
@@ -3,32 +3,38 @@
 //! * `/metrics` -- reports prometheus-formatted metrics.
 //! * `/ready` -- returns 200 when the proxy is ready to participate in meshed traffic.
```
adleong: I think the /proxy-log-level route should be included here

src/app/admin/mod.rs
```diff
@@ -55,22 +61,25 @@ where
...
-            "/metrics" => self.metrics.call(req),
-            "/ready" => future::ok(self.ready_rsp()),
-            _ => future::ok(
-                Response::builder()
-                    .status(StatusCode::NOT_FOUND)
-                    .body(Body::empty())
-                    .expect("builder with known status code must not fail"),
-            ),
+            "/metrics" => Box::new(self.metrics.call(req)),
+            "/proxy-log-level" => self.trace_level.call(req),
```
adleong: Is the name "proxy-log-level" redundant, given that this is on the proxy's admin server?  Could it just be called "log-level"?

src/app/admin/trace_level.rs
```diff
@@ -0,0 +1,85 @@
...
+                StatusCode::INTERNAL_SERVER_ERROR,
+                Body::empty(),
+            )));
+        }
+
+        match req.method() {
+            &Method::GET => match self.current() {
+                Ok(level) => Box::new(future::ok(rsp(StatusCode::OK, level))),
+                Err(error) => {
+                    warn!(message = "error getting proxy log level", %error);
```
adleong: What does the % symbol mean here?

src/app/admin/trace_level.rs
```diff
@@ -0,0 +1,85 @@
...
+            )),
+        }
+    }
+}
+
+impl TraceLevel {
+    fn set_from(&self, chunk: hyper::Chunk) -> Result<(), String> {
+        let bytes = chunk.into_bytes();
+        let body = str::from_utf8(&bytes.as_ref()).map_err(|e| format!("{}", e))?;
+        trace!(request.body = ?body);
```
adleong: What does the ? symbol mean here?

src/logging.rs
```diff
@@ -17,32 +17,40 @@ thread_local! {
...
-        let dispatch = Dispatch::new(
-            subscriber_builder()
-                .with_filter(filter::EnvFilter::from(filter))
-                .finish(),
-        );
+    pub fn init_with_filter<F: AsRef<str>>(filter: F) -> Result<LevelHandle, Error> {
+        // Set up the subscriber
+        let builder = subscriber_builder()
+            .with_filter(filter::EnvFilter::from(filter))
+            .with_filter_reloading();
```
adleong: I don't understand this change.  What is filter reloading and what effect does it have?

src/logging.rs
```diff
@@ -97,6 +105,50 @@ pub mod trace {
         }
     }
 
+    impl LevelHandle {
+        /// Returns a new `LevelHandle` without a corresponding filter.
```
adleong: What is the corresponding filter that is omitted here?  To me the builder here looks the same as the one in `init_with_filter`

adleong:
What do the trailing `%` symbols mean in the GET response bodies?

src/app/admin/mod.rs
olix0r: Given the annotation, etc is called `proxy-log-level`, the intent was to make it consistent. I don't feel strongly though

olix0r:
@adleong 

> What do the trailing % symbols mean in the GET response bodies?

That's the shell indicating that there's no newline at the end of the response

src/app/admin/mod.rs
hawkw: whoops, good eye!

src/app/admin/mod.rs
hawkw: Yeah, the route was in the original spec for this feature for the reason Oliver mentioned.

src/app/admin/trace_level.rs
hawkw: This is using the `tracing` crate's [structured key-value fields](https://docs.rs/tracing/0.1.2/tracing/macro.span.html#recording-fields); the `%` symbol means that the field should be recorded using its `fmt::Display` implementation.

src/app/admin/trace_level.rs
hawkw: `?` indicates that the field value should be recorded using its `fmt::Debug` implementation.

src/logging.rs
hawkw: The "filter" is what determines whether a span or event should be recorded. `with_filter_reloading` configures the builder to use a filter that can be dynamically changed after the subscriber is constructed. 

The method `reload_handle` that we [call on the subsequent line](https://github.com/linkerd/linkerd2-proxy/pull/279/files/553f4897cdeca55), which returns a handle that we can use to set a new filter, is only implemented when the filter on the subscriber being built has been wrapped to support reloading (which is what `with_filter_reloading` does).

src/logging.rs
hawkw: The difference is that, rather than installing the subscriber as the default, it's dropped at the end of the function, so the handle doesn't point at anything.

PR #280, opened 2019-06-25
Introduce `l5d-force-id` request header

kleimkuhler:
The `l5d-force-id` header is useful for forcing the TLS server identity of a
connection. When a request is made with this header, the value supplied is used
as the identity for the outbound endpoint that is created.

This is useful in the proxy when the Tap Service sends tap requests to
individual pods. The `Host` header of tap requests is the IP of the pod we want
to tap, so the proxy will not attempt to do service discovery for those
requests. Because service discovery is not performed, the connections will not
be TLS'd.

As part of the Tap hardening project, tap should only be served on connections
to the Tap service. In order to do this, the Tap service must initiate the
connection over TLS. This header will allow that to happen, and the pod serving
tap to verify the tap client identity.

Closes linkerd/linkerd2#2677


kleimkuhler:
Briefly closing to think a little more about design here. We should consider some of the details @olix0r listed in linkerd/linkerd2#2185

kleimkuhler:
After speaking with @adleong this is a good first step in the context of tap hardening. Two additional features were listed in linkerd/linkerd2#2185 -- neither of which is functionality that would be useful in this project or elsewhere in the proxy. They are still good things to consider down the road, but will require some more significant changes to `outbound::Endpoint` and `tls/client.rs`.

tests/identity.rs
```diff
@@ -267,3 +267,40 @@ fn refresh() {
...
+
+    let proxy = proxy::new()
+        .outbound(srv)
+        .identity(proxy_identity.service().run())
+        .run_with_test_env(proxy_identity.env);
+
+    // Make an http client with no TLS config
+    let client = client::http2(proxy.outbound, app_id);
+
+    // Send a request with a `l5d-force-id` header value of `app_id` so that
```
adleong: I think we should also test the failure scenarios: 
* When the l5d-force-id header is present and the server's TLS id doesn't match it, the connection should fail
* When the l5d-force-id header is present and the server does not have a TLS id, the connection should fail.

src/app/main.rs
```diff
@@ -506,11 +508,20 @@ where
 
             // Routes requests to their original destination endpoints. Used as
             // a fallback when service discovery has no endpoints for a destination.
+            //
+            // If the `l5d-force-id` header is present, then that identity is
+            // used as the server name when connecting to the endpoint.
             let orig_dst_router = svc::builder()
```
adleong: By adding this logic only to the orig_dst_router, I think the `l5d-force-id` header can only be used for non-Kubernetes services.  This is the case for tap requests, but I would have expected the `l5d-force-id` functionality to work for any outbound request, including ones to Kubernetes services.

olix0r: This is an interesting wrinkle... I think we'd probably want to preemptively fail the request in the case that the force-id doesn't match the discovered id?

kleimkuhler: The header logic is here because my understanding is we can only inspect request headers in a router. Because of this, I don't think we can check for `l5d-force-id` any lower (? - closer to the endpoint) in the stack. It is an interesting point; the current change is very specific to tap requests but not helpful for other use cases.

I could look into checking for this header in an earlier router and setting something on the request that could be passed down through the stack? I'll look into some different approaches to take and update this conversation.

kleimkuhler: I'm going to move the header check up into the destination router. We can then check the `l5d-force-id` value against the discovered endpoint identity and error if they are not equal. This will also allow more general use cases of the header instead of it's current location in the `original_dst_router`

kleimkuhler:
@adleong @olix0r I added a filter layer to the endpoint stack that will preemptively fail request with the `l5d-force-id` header and a different `outbound::Endpoint`'s `identity`.

When this layer makes the endpoint stack, it wraps the service in a new service that checks requests for the `l5d-force-id` header and errors if there is an unexpected identity.

I originally hoped to uset `tower_filter`, but due to the bounds of working the context of `Service` and not `MakeService`, it made sense to write a separate `MakeService` filter. This filter is specific to the identity check and is found [here](https://github.com/linkerd/linkerd2-proxy/blob/14f74d8a05c58bc42c2225310ba7e9b38a005eb6/src/app/outbound.rs#L531-L532)

@adleong I also extended the tests to assert the behavior of clients that connected to peers through `SO_ORIGINAL_DST` as well as the discovery service.

When the `l5d-force-id` header is present, there must be a matching `Endpoint` `identity`. If there is an unexpected `Endpoint` `identity` or totally absent, the service will error.

There work I would like to polish is the proper way for `IdentityCheckError` to return an `h2::error::Error` that indicates this client does not have permission to connect to the endpoint. Currently it is a `502` `BAD_GATEWAY`.

src/app/outbound.rs
```diff
@@ -398,6 +398,213 @@ pub mod orig_proto_upgrade {
...
+    impl std::error::Error for IdentityCheckError {}
+
+    impl std::fmt::Display for IdentityCheckError {
+        // TODO(kleimkuhler): Use force_identity and peer_identity fields?
+        fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+            write!(f, "force identity check failed")
+        }
+    }
+
+    impl HasH2Reason for IdentityCheckError {
```
kleimkuhler: @hawkw If you have some time to look at this tomorrow, I had some trouble implementing `HasH2Reason` for `IdentityCheckError` to return a permission denied status code instead of a `BAD_GATEWAY` like is currently asserted in the [tests](https://github.com/linkerd/linkerd2-proxy/pull/280/files#diff-dc35718beb77b0814d17cefbd6b98593R373)

tests/identity.rs
```diff
@@ -267,3 +267,165 @@ fn refresh() {
...
+    assert_eq!(
+        client
+            .request(
+                client
+                    .request_builder("/")
+                    .header("l5d-force-id", "hey-its-me")
+                    .method("GET")
+            )
+            .status(),
+        http::StatusCode::SERVICE_UNAVAILABLE
```
kleimkuhler: When a request is routed to `SO_ORIGINAL_DST`, the request seems to be bypassing the `force_identity_check::layer()` and therefore trying to connect to an identity that is incorrect or does not exist. The connection goes into a backoff and returns a `SERVICE_UNAVAILABLE`. I don't think this is the behavior we want, but maybe it is? It's an open question that I'm not totally sure how to resolve right now.

kleimkuhler:
Temporarily closing this to reopen as a draft because a lot has changed.

PR #283, opened 2019-07-03
Improve readability of profiles router

adleong:
Fixes https://github.com/linkerd/linkerd2/issues/3005

This improves the readability of the profiles router by breaking it into 3 files, and renaming the generic type parameters to longer and more self-descriptive names.


src/proxy/http/profiles/mod.rs
```diff
@@ -0,0 +1,286 @@
+extern crate linkerd2_router as rt;
+extern crate tower_discover;
+
+use futures::Stream;
+use http;
+use indexmap::IndexMap;
+// Import is used by WeightedIndex::sample.
+#[allow(unused_imports)]
+use rand::distributions::Distribution;
```
olix0r: I've confirmed that you can actually just delete this import from this file. It is needed in another file though.

src/proxy/http/profiles/recognize.rs
```diff
@@ -0,0 +1,90 @@
+use http;
+use std::hash::Hash;
+
+use super::*;
```
olix0r: Please enumerate the imports for this module

src/proxy/http/profiles/router.rs
```diff
@@ -0,0 +1,367 @@
...
+use std::hash::Hash;
+
+use never::Never;
+
+use dns;
+use proxy::Error;
+use svc;
+
+use super::recognize::*;
+use super::*;
```
olix0r: Similarly, we prefer to avoid glob imports for module code.

Also, I'd avoid so many newlines between import groups. It's fine to have one big group, especially if crate-local imports are e.g. `use crate::dns;`

src/proxy/http/profiles/recognize.rs
```diff
@@ -0,0 +1,91 @@
+use super::{rt, RequestMatch, Route, WeightedAddr, WithAddr, WithRoute};
+use http;
+use rand::distributions::Distribution;
+use rand::distributions::WeightedIndex;
```
olix0r: nit: can now be `...::{Distribution, WeightedIndex}`

src/proxy/http/profiles/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+// === impl Route ===
+
+impl Route {
+    pub fn new<I>(label_iter: I, response_classes: Vec<ResponseClass>) -> Self
+    where
+        I: Iterator<Item = (String, String)>,
+    {
+        let labels = {
+            let mut pairs = label_iter.collect::<Vec<_>>();
+            pairs.sort_by(|(k0, _), (k1, _)| k0.cmp(k1));
```
kleimkuhler: Is there a reason we need to sort the labels?

adleong: I'm not sure, this block of code was simply moved in this PR.  I'd guess that we sort these so that the prometheus output is predictable and consistent.

PR #285, opened 2019-07-03
canonicalize: stop resolving when the receiver is dropped

olix0r:
The proxy's canonicalization queries can persist after the observing
service is dropped.

This change modifies the background resolution task to use tokio-sync so
that it can eagerly cease polling for updates when corresponding service
is dropped.

### DIFF
--- a/src/proxy/http/canonicalize.rs
+++ b/src/proxy/http/canonicalize.rs
@@ -11,6 +11,7 @@
 
 use futures::{Async, Future, Poll, Stream};
 use http;
+use log::trace;
 use std::time::Duration;
 use tokio::{self, sync::mpsc};
 use tokio_timer::{clock, Delay, Timeout};
@@ -181,12 +182,16 @@ impl Future for Task {
         loop {
             // If the receiver has been dropped, stop watching for updates.
             let tx_ready = match self.tx.poll_ready() {
-                Err(_) => return Ok(Async::Ready(())),
                 Ok(r) => r,
+                Err(_) => {
+                    trace!("task complete; name={:?}", self.original);
+                    return Ok(Async::Ready(()));
+                }
             };
 
             self.state = match self.state {
                 State::Init => {
+                    trace!("task init; name={:?}", self.original);
                     let f = self.resolver.refine(self.original.name());
                     State::Pending(Timeout::new(f, self.timeout))
                 }
@@ -194,6 +199,7 @@ impl Future for Task {
                     // Only poll the resolution for updates when the receiver is
                     // ready to receive an update.
                     if tx_ready.is_not_ready() {
+                        trace!("task awaiting capacity; name={:?}", self.original);
                         return Ok(Async::NotReady);
                     }
 
@@ -202,6 +208,7 @@ impl Future for Task {
                             return Ok(Async::NotReady);
                         }
                         Ok(Async::Ready(refine)) => {
+                            trace!("task update; name={:?} refined={:?}", self.original, refine.name);
                             // If the resolved name is a new name, bind a
                             // service with it and set a delay that will notify
                             // when the resolver should be consulted again.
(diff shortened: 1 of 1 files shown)


src/proxy/http/canonicalize.rs
```diff
@@ -178,13 +178,25 @@ impl Future for Task {
...
         loop {
             self.state = match self.state {
                 State::Init => {
                     let f = self.resolver.refine(self.original.name());
                     State::Pending(Timeout::new(f, self.timeout))
                 }
                 State::Pending(ref mut fut) => {
+                    // Only poll the resolution for updates when the receiver is
+                    // ready to receive an update.
+                    if tx_ready.is_not_ready() {
```
seanmonstar: I don't have the full loop in my head: is possible to get bad to here after having sent once? If so, this poll would be old...

src/proxy/http/canonicalize.rs
```diff
@@ -179,33 +188,62 @@ impl Future for Task {
...
                             let resolved = NameAddr::new(refine.name, self.original.port());
                             if self.resolved.get() != Some(&resolved) {
-                                let err = self.tx.try_send(resolved.clone()).err();
-                                if err.map(|e| e.is_disconnected()).unwrap_or(false) {
-                                    return Ok(().into());
-                                }
-
+                                self.tx
+                                    .try_send(resolved.clone())
+                                    .expect("tx failed despite being ready");
```
adleong: Just to check my understanding: it is now impossible to get here if the receiver has been dropped because of the guard at the top of this function?

adleong: Or doubly impossible because of the the `poll_ready` as well?

src/proxy/http/canonicalize.rs
```diff
@@ -9,10 +9,12 @@
 //! DNS TTLs are honored and, if the resolution changes, the inner stack is
```
adleong: This isn't true, is it?  I don't see the inner stack getting rebuilt.  As far as I can tell, this module just sets the canonicalized addr into the request extensions.

src/proxy/http/canonicalize.rs
```diff
@@ -179,33 +188,62 @@ impl Future for Task {
 
     fn poll(&mut self) -> Poll<(), ()> {
```
adleong: Just curious why this is modeled as a Future instead of as a Stream?

src/proxy/http/canonicalize.rs
```diff
@@ -136,14 +141,16 @@ where
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         let inner = try_ready!(self.inner.poll());
         let svc = if let Some((na, resolver, timeout)) = self.task.take() {
-            let (tx, rx) = mpsc::channel(2);
+            let (tx, rx) = mpsc::channel(1);
```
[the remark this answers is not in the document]
olix0r: Before we didn't (couldn't?) `tx.poll_ready()` before polling the resolution future, so we allowed an extra buffer slot. Now we exert backpressure properly.

src/proxy/http/canonicalize.rs
olix0r: The poll_ready is the important part of it. It guarantees there's a slot to send into.

src/proxy/http/canonicalize.rs
olix0r: You can only spawn futures. As this is intended to be a spawned background task, it's a future that only completes when all work is done.

Or, to put it another way: what would consume the stream?

adleong: I'd imagine the service would consume it.  i.e. rather than doing `self.rx.poll()` doing `self.addr_stream.poll()`.

adleong: But I guess the difference is that this way the DNS resolutions are refreshed eagerly.

olix0r: Exactly. We rarely want to use poll_ready to drive work; almost always it should just check if other work has completed.

PR #286, opened 2019-07-04
router: Remove interval from router eviction

olix0r:
The router only purges its cache only once per `max_idle_age`, which
means that a route may be idle for ~2x `max_idle_age`.

By removing the interval completely, we can ensure that routes are
evicted in a timely fashion.

### DIFF
--- a/lib/router/src/cache.rs
+++ b/lib/router/src/cache.rs
@@ -2,9 +2,9 @@ use std::{hash::Hash, time::Duration};
 
 use futures::{try_ready, Async, Future, Poll, Stream};
 use indexmap::IndexMap;
-use log::debug;
+use log::trace;
 use tokio::sync::lock::Lock;
-use tokio_timer::{delay_queue, DelayQueue, Error, Interval};
+use tokio_timer::{delay_queue, DelayQueue};
 
 /// An LRU cache that can eagerly remove values in a background task.
 ///
@@ -43,13 +43,9 @@ where
 /// A background future that eagerly removes expired cache values.
 ///
 /// If the cache is dropped, this future will complete.
-pub struct PurgeCache<K, V>
+pub struct PurgeCache<K, V>(Lock<Cache<K, V>>)
 where
-    K: Clone + Eq + Hash,
-{
-    cache: Lock<Cache<K, V>>,
-    interval: Interval,
-}
+    K: Clone + Eq + Hash;
 
 /// A handle to a cache value.
 struct Node<T> {
@@ -66,19 +62,14 @@ where
 {
     pub fn new(capacity: usize, expires: Duration) -> (Lock<Cache<K, V>>, PurgeCache<K, V>) {
         assert!(capacity != 0);
-        let cache = Self {
+        let cache = Lock::new(Self {
             capacity,
             expires,
             expirations: DelayQueue::with_capacity(capacity),
             values: IndexMap::default(),
-        };
-        let cache = Lock::new(cache);
-        let bg_purge = PurgeCache {
-            cache: cache.clone(),
-            interval: Interval::new_interval(expires),
-        };
+        });
 
-        (cache, bg_purge)
+        (cache.clone(), PurgeCache(cache))
     }
 
     pub fn capacity(&self) -> usize {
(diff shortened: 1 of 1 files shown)


hawkw:
looks like the build failure may be travis-related

lib/router/src/cache.rs
```diff
@@ -122,14 +113,20 @@ where
...
-            debug!("expiring an item from the cache");
-            self.values.remove(expired.get_ref());
+    fn poll_purge(&mut self) -> Poll<(), ()> {
+        loop {
+            match try_ready!(self
+                .expirations
+                .poll()
+                .map_err(|e| unreachable!("expiration must not fail: {}", e)))
+            {
+                None => return Ok(Async::NotReady),
```
kleimkuhler: If `poll_purge` returns `NotReady` here, will the lock we acquired in `poll` [here](https://github.com/linkerd/linkerd2-proxy/pull/286/files#diff-230615b3316b2a089b0d6563c3f74944R144) get released? Or, will we continue to loop within this `poll_purge` and always hold on to the lock that we acquired?

kleimkuhler: Looking at this, I'm not immediately sure if this returning `NotReady` would pop up to the `poll` where the lock was acquired. And if that is not happening, then that would be an issue.

olix0r: I'm not sure I understand the question... Given the future impl:

```rust
impl<K, V> Future for PurgeCache<K, V>
where
    K: Clone + Eq + Hash,
    V: Clone,
{
    type Item = ();
    type Error = ();

        let mut cache = try_ready!(Ok(self.0.poll_lock()));
        cache.poll_purge()
    }
}
```

When poll_purge returns not ready, the lock is released (as the guard is dropped at the end of the scope).

kleimkuhler: Ok cool thanks for talking this through I’m following what’s going on and it makes sense.

olix0r: However, as it is, the purge future may never complete when the cache is dropped. I have a small modification to this branch that will fix this.

olix0r: Actually, since this doesn't pose a current problem for the proxy, let's defer this (it's a little more complicated than i initially thought).

lib/router/src/cache.rs
```diff
@@ -144,21 +141,8 @@ where
...
-            debug!("purge task locked the router cache");
-
-            // If poll_purge is not ready, we cannot expire any values and
-            // wait for the next interval. If poll_purge is ready, we have
-            // expired all values and wait for the next interval
-            debug!("purge task is polling for evictions...");
-            cache.poll_purge().expect("Cache::poll_purge must not fail");
-        }
+        let mut cache = try_ready!(Ok(self.0.poll_lock()));
+        cache.poll_purge()
```
hawkw: Nit/TIOLI: does
```rust
try_ready!(Ok(self.0.poll_lock())).poll_purge()
```
not work, did you think this was clearer?

olix0r: it works, but there are a lot of parens there, so i thought it clearer...

lib/router/src/cache.rs
adleong: What are the semantics of expirations returning None here?  Does that mean that there will never be any expirations ever again?  By my reading, returning NotReady here causes PurgeCache.poll to return NotReady but doesn't schedule this task to ever get polled again?

olix0r: Great question! From the DelayQueue docs:

> If no items are in the queue, i.e. is_empty() returns true, then poll returns Ready(None). This indicates that the stream has reached an end. However, if a new item is inserted after, poll will once again start returning items or `NotReady`.

But the real question is whether the delay queue notifies in this case...

FWIW, this seems to work fine in my manual tests of this behavior, but may not be correct by contract... We'd want to poll on something that will be notified when an item is inserted to be sure.

lib/router/src/cache.rs
```diff
@@ -120,7 +127,10 @@ where
                 .poll()
                 .map_err(|e| unreachable!("expiration must not fail: {}", e)))
             {
-                None => return Ok(Async::NotReady),
+                None => {
+                    self.purge_task = Some(task::current());
```
adleong: TIOLI: This could use a comment explaining that this is essentially registering the task to be notified when something is added to the delay queue.

lib/router/src/cache.rs
```diff
@@ -144,21 +151,8 @@ where
     type Error = ();
```
adleong: TIOLI: I think the Item type can now be Never.

olix0r: `spawn` takes a `Future<Item= (), Error = ()>`

PR #289, opened 2019-07-18
Introduce `l5d-require-id` request header

kleimkuhler:
This was originally implemented in #280, but has since changed enough that a
new PR is being opened

### Motivation

As part of the Tap hardening project, tap should only be served on TLS
connections. Currently, when the Tap service makes an outbound Tap request, the
`Host` header is the Pod IP of the resource. Service discovery is not attempted
and therefore the Tap service initiates a plaintext connection. The introduction
of the `l5d-require-id` header will allow an outbound request to require the
specified identity on the endpoint it connects to.

### Details

The `orig_dst_router` router on outbound requests will set the identity of the
`outbound::Endpoint` to the value of the `l5d-require-id` header if it is
present. This is useful specifically for the case explained above when request
are made with an IP as their `Host` header and need to be TLS connections.

In order for the header to be used in other scenarios, a new layer has been
added to the `endpoint_stack` stack on outbound requests. This layer is
implemented in `outbound::require_identity_on_endpoint`. The purpose of this
layer is to wrap the `MakeService` of `Endpoint`s with a new service that checks
for the presence of required identities, and if present, asserts the identity of
the endpoint is expected.

When constructing an `Endpoint`, if a request has the `l5d-require-id` header
then the endpoint will be constructed if it's identity is the expected identity
of the header value; if not, it will error and return a status code `FORBIDDEN`.

### Tests

Tests have been added for both http1 and http2. The tests cover the
`orig_dst_router` router as well as discovery. In both of those cases, clients
make requests with and without the `l5d-require-id` header, as well as correct
and inccorect `l5d-require-id` header values.

Closes linkerd/linkerd2#2677

### DIFF
--- a/src/app/errors.rs
+++ b/src/app/errors.rs
@@ -127,10 +127,10 @@ fn map_err_to_5xx(e: Error) -> StatusCode {
     } else if let Some(_) = e.downcast_ref::<router::NotRecognized>() {
         error!("could not recognize request");
         http::StatusCode::BAD_GATEWAY
-    } else if let Some(_) =
+    } else if let Some(err) =
         e.downcast_ref::<outbound::require_identity_on_endpoint::RequireIdentityError>()
     {
-        warn!("require identity check failed and downcasted");
+        error!("{}", err);
         http::StatusCode::FORBIDDEN
     } else {
         // we probably should have handled this before?
--- a/src/app/errors.rs
+++ b/src/app/errors.rs
@@ -110,6 +110,7 @@ where
 }
 
 fn map_err_to_5xx(e: Error) -> StatusCode {
+    use app::outbound;
     use proxy::buffer;
     use proxy::http::router::error as router;
     use tower::load_shed::error as shed;
@@ -126,6 +127,11 @@ fn map_err_to_5xx(e: Error) -> StatusCode {
     } else if let Some(_) = e.downcast_ref::<router::NotRecognized>() {
         error!("could not recognize request");
         http::StatusCode::BAD_GATEWAY
+    } else if let Some(_) =
+        e.downcast_ref::<outbound::require_identity_on_endpoint::RequireIdentityError>()
+    {
+        warn!("require identity check failed and downcasted");
+        http::StatusCode::FORBIDDEN
     } else {
         // we probably should have handled this before?
         error!("unexpected error: {}", e);
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -584,7 +584,7 @@ where
                 .layer(router::layer(
                     router::Config::new("out dst", capacity, max_idle_age),
                     |req: &http::Request<_>| {
-                        let addr = req.extensions().get::<Addr>().cloned().map(move |addr| {
+                        let addr = req.extensions().get::<Addr>().cloned().map(|addr| {
                             let settings = settings::Settings::from_request(req);
                             DstAddr::outbound(addr, settings)
                         });
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -530,7 +530,7 @@ where
                     |req: &http::Request<_>| {
                         let ep = outbound::Endpoint::from_orig_dst(req).and_then(|mut ep| {
                             if let Some(force_id) = identity_from_header(req, super::L5D_FORCE_ID) {
-                                debug!("outbound ep force identity={:?}", force_id);
+                                debug!("outbound ep server name={:?}; force-id", force_id);
                                 ep.identity = Conditional::Some(force_id);
                             }
                             Some(ep)
@@ -583,9 +583,10 @@ where
                 .layer(router::layer(
                     router::Config::new("out dst", capacity, max_idle_age),
                     |req: &http::Request<_>| {
-                        let addr = req.extensions().get::<Addr>().cloned().map(|addr| {
+                        let force_identity = identity_from_header(req, super::L5D_FORCE_ID);
+                        let addr = req.extensions().get::<Addr>().cloned().map(move |addr| {
                             let settings = settings::Settings::from_request(req);
-                            DstAddr::outbound(addr, settings)
+                            DstAddr::outbound(addr, settings, force_identity)
                         });
                         debug!("outbound dst={:?}", addr);
                         addr
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -434,10 +434,10 @@ where
             use super::outbound::{
                 self,
                 discovery::Resolve,
-
                 orig_proto_upgrade,
-                //add_remote_ip_on_rsp, add_server_id_on_rsp,
                 require_identity_on_endpoint,
+                //add_remote_ip_on_rsp, add_server_id_on_rsp,
+            };
             use proxy::{
                 http::{
                     balance, canonicalize, fallback, header_from_target, identity_from_header,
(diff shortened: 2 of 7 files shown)


olix0r:
tests fail in CI:
```
failures:
    require_id_header::http1::disco_client_connects_to_tls_server
    require_id_header::http1::orig_dst_client_connects_to_tls_server
test result: FAILED. 14 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out
```

src/app/outbound.rs
```diff
@@ -398,6 +398,223 @@ pub mod orig_proto_upgrade {
     }
 }
 
+pub mod require_identity_on_endpoint {
```
olix0r: nit: `on_endpoint` is redundant imo, since identity can only ever apply to endpoints.

src/app/main.rs
```diff
@@ -514,11 +520,22 @@ where
...
                 .layer(router::layer(
                     router::Config::new("out ep", capacity, max_idle_age),
                     |req: &http::Request<_>| {
-                        let ep = outbound::Endpoint::from_orig_dst(req);
+                        let ep = outbound::Endpoint::from_orig_dst(req).and_then(|mut ep| {
+                            if let Some(require_id) =
+                                identity_from_header(req, super::L5D_REQUIRE_ID)
+                            {
+                                debug!("outbound ep require id={:?}", require_id);
+                                ep.identity = Conditional::Some(require_id);
```
olix0r: How do you feel about moving this logic (and the `identity_from_header` utility) into the `outbound` module? This accomplishes a few things:

1. `main` is less verbose (there's already a lot in here_;
2. this logic only ever applies on the outbound side, so it need not be generalized;
3. identity is only constructed from a header in this context, so it need not be generalized

src/proxy/http/mod.rs
```diff
@@ -62,14 +63,32 @@ impl HasH2Reason for ::h2::Error {
...
-            if s.is_empty() {
-                None
-            } else {
-                s.parse::<Authority>().ok()
-            }
-        })
+        value
+            .to_str()
+            .ok()
+            .and_then(|s| if s.is_empty() { None } else { try_from(s) })
```
olix0r: I'm not sure that this helper buys all that much, but if we preserve it it would be clearer using the short-circuit syntax:
```rust
let value = req.headers().get(header)?.to_str().ok()?;
if value.is_empty()  {
    return None;
}
try_from(s)
```

src/app/outbound.rs
kleimkuhler: If it's just a nit I lean towards keep it, but I can remove it if you'd like. I still have #2543 in the back of my head, and am still under the impression this would add the option for identity on the source (or at least some way to represent both the local and peer identities in a connection).

src/app/main.rs
kleimkuhler: Something like this? I'd like to still show we're doing an additional check for the presence of a header, but agree it's overly verbose right now.

```
|req: &http::Request<_>| {
    let ep = outbound::Endpoint::from_orig_dst(req).and_then(|mut ep| {
        let ep = outbound::should_require_identity(ep, req);
        Some(ep)
    });
    debug!("outbound ep={:?}", ep);
    ep
},
```

src/proxy/http/mod.rs
kleimkuhler: I can change the syntax. I added it because when I added the other `fn`, I had both right next to each other and they did pretty much the exact same thing.

src/app/main.rs
kleimkuhler: Well something more like:

```
|req: &http::Request<_>| {
    let ep = outbound::Endpoint::from_orig_dst(req).map(|ep| {
        outbound::should_require_identity(ep, req)
    });
    debug!("outbound ep={:?}", ep);
    ep
},
```

src/app/main.rs
```diff
@@ -489,6 +494,7 @@ where
                 // disabled on purpose
                 //.layer(add_server_id_on_rsp::layer())
                 //.layer(add_remote_ip_on_rsp::layer())
+                .layer(strip_header::request::layer(super::L5D_REQUIRE_ID))
```
adleong: Just curious: does it make sense for the `require_identity_on_endpoint` layer to strip this header instead of doing it in a separate layer?  Presumably we would always want to strip the header after using it?

kleimkuhler:
So the tests did need to be fixed. The HTTP status code of the error when trying to connect with an incorrect `l5d-required-id` header is `BAD_GATEWAY` for http1, and `SERVICE_UNAVAILABLE` for http2. I extended the `generate_tests!` macro to take `connect_error` as well.

Also, there is a chatty server error as a result of http2 running the `orig_dst_client_cannot_connect_to_plaintext_server` test. It prints `support/server error: http2 error: protocol error: unspecific protocol error detected` a few times during the test run after trying to connect to a non-TLS server with an identity set on the endpoint.

I have not figured out the exact reason why the status code changes for http1 and http2. I'd like to get a better idea of that.

As for the chatty server, the test support and this file in general could use some general cleanup. I can address in a follow-up PR with some test support cleanup and figure out why the server prints errors only in the http2 case.

src/app/main.rs
olix0r: It probably could. The counter-argument is that it's better to have each layer do exactly one thing so that it's easier to reason about if, for instance, other layers want to act on or report headers...

olix0r:
I think a `fmt` is required, and it looks like the tests are still flakey in CI.

Before merging, we should either mark these tests as flakey or more deeply debug the flakiness...

But I think the code is in a good enough state to merge once the tests issues are sorted.

src/app/errors.rs
```diff
@@ -126,6 +127,11 @@ fn map_err_to_5xx(e: Error) -> StatusCode {
     } else if let Some(_) = e.downcast_ref::<router::NotRecognized>() {
         error!("could not recognize request");
         http::StatusCode::BAD_GATEWAY
+    } else if let Some(err) =
+        e.downcast_ref::<outbound::require_identity_on_endpoint::RequireIdentityError>()
+    {
+        error!("{}", err);
+        http::StatusCode::FORBIDDEN
```
olix0r: Actually, I know I said this should be FORBIDDEN in the issue, but that seems to imply something else -- that the server forbade requests from this client, when, in reality, the client's requirement on the server was not met. PRECONDITION FAILED seems inappropriate as well, though perhaps more in the direction of what we're looking for.

@seanmonstar any suggestions?

src/app/main.rs
kleimkuhler: I leaned towards a `strip_header` layer because it's still being explicit about the possibility of the`l5d-require-id` header being present. I think the argument could be made for other `l5d-*` headers, so if it's something we want to do I'd rather cover all those situations in a separate PR!

src/app/errors.rs
kleimkuhler: Hm good point... FORBIDDEN should be something returned by the server and that's not the case here. I can wait for input from @seanmonstar, but I like the suggestion.

olix0r: how about [`409 Conflict`](https://tools.ietf.org/html/rfc7231#section-6.5.8)

kleimkuhler:
@olix0r I've only seen these flakiness failures on CI. I'm going to try making separate clients for pass/fail requests and see if that fixes it. I'll push a commit with that change in a few.

src/app/errors.rs
seanmonstar: "428 Precondition Failed" is meant for when using headers like `If-Match: some-etag` in a `PUT`, `POST`, etc.

I've been thinking we should be able to send HTTP2 reset codes in here if the request was HTTP2, and `INADEQUATE_SECURITY` seems to match...

olix0r: https://github.com/linkerd/linkerd2/issues/3111

kleimkuhler:
I think I may mark these as flaky for now until I can figure out a better way to reproduce the failures locally or on Ares. I'll remove the previous commit that added `pass/fail_client`s.

PR #290, opened 2019-07-24
Tap server authorizes clients when identity is expected

kleimkuhler:
### Motivation

As part of the tap hardening project the tap server must serve tap requests
exclusively to the tap service. Clients must be both authenticated and
authorized. Currently, the tap server still can accept clients over plaintext,
and does not authorize clients over private connections.

### Details

In order for authenticated clients to be authorized, the `tap_identity` config
variable has been introduced. This value is used as the expected client identity
of incoming tap requests. When `tap_identity` is set, the tap server now checks
the peer identity of the client connection against this value. If they are equal
the connection is accept; otherwise is is rejected.

The tap server still accepts plaintext connections. This will be removed in a
future PR once the tap controller is sending tap requests with the required
`l5d-require-id` that is needed to set server identity on outbound requests to
IP addresses.

### Tests

Tests have been added to `tap::expected_identity` that assert:

* The tap server accepts plaintext connections when `tap_identity` is not set in
  the config
* The tap server rejects plaintext connections when `tap_identity` is set in the
  config
* The tap server rejects private connections when the client identity does not
  match the expected `tap_identity`
* The tap server accepts private connections when the client identity does match
  the expected `tap_identity`

### DIFF
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -103,6 +103,8 @@ pub struct Config {
     pub control_connect_timeout: Duration,
 
     pub identity_config: tls::Conditional<identity::Config>,
+
+    pub tap_identity: tls::PeerIdentity,
     //
     // Destination Config
     //
@@ -265,6 +267,7 @@ pub const ENV_IDENTITY_DISABLED: &str = "LINKERD2_PROXY_IDENTITY_DISABLED";
 pub const ENV_IDENTITY_DIR: &str = "LINKERD2_PROXY_IDENTITY_DIR";
 pub const ENV_IDENTITY_TRUST_ANCHORS: &str = "LINKERD2_PROXY_IDENTITY_TRUST_ANCHORS";
 pub const ENV_IDENTITY_IDENTITY_LOCAL_NAME: &str = "LINKERD2_PROXY_IDENTITY_LOCAL_NAME";
+pub const ENV_IDENTITY_TAP_IDENTITY: &str = "LINKERD2_PROXY_IDENTITY_TAP_IDENTITY";
 pub const ENV_IDENTITY_TOKEN_FILE: &str = "LINKERD2_PROXY_IDENTITY_TOKEN_FILE";
 pub const ENV_IDENTITY_MIN_REFRESH: &str = "LINKERD2_PROXY_IDENTITY_MIN_REFRESH";
 pub const ENV_IDENTITY_MAX_REFRESH: &str = "LINKERD2_PROXY_IDENTITY_MAX_REFRESH";
@@ -451,6 +454,8 @@ impl Config {
             parse_control_addr(strings, ENV_DESTINATION_SVC_BASE)
         };
 
+        let tap_identity = parse(strings, ENV_IDENTITY_TAP_IDENTITY, parse_identity);
+
         let dst_token = strings.get(ENV_DESTINATION_CONTEXT);
 
         let dst_get_suffixes = parse(strings, ENV_DESTINATION_GET_SUFFIXES, parse_dns_suffixes);
@@ -547,6 +552,10 @@ impl Config {
             destination_context: dst_token?.unwrap_or_default(),
 
             identity_config: identity_config?
+                .map(Conditional::Some)
+                .unwrap_or_else(|| Conditional::None(tls::ReasonForNoIdentity::Disabled)),
+
+            tap_identity: tap_identity?
                 .map(Conditional::Some)
                 .unwrap_or_else(|| Conditional::None(tls::ReasonForNoIdentity::Disabled)),
 
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -104,6 +104,7 @@ pub struct Config {
 
     pub identity_config: tls::Conditional<identity::Config>,
 
+    /// The expected identity of the Tap service.
     pub tap_identity: tls::PeerIdentity,
     //
     // Destination Config
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -104,8 +104,8 @@ pub struct Config {
 
     pub identity_config: tls::Conditional<identity::Config>,
 
-    /// The expected identity of the Tap service.
-    pub tap_identity: tls::PeerIdentity,
+    /// The expected name of the Tap service.
+    pub tap_svc_name: tls::PeerIdentity,
     //
     // Destination Config
     //
@@ -268,7 +268,6 @@ pub const ENV_IDENTITY_DISABLED: &str = "LINKERD2_PROXY_IDENTITY_DISABLED";
 pub const ENV_IDENTITY_DIR: &str = "LINKERD2_PROXY_IDENTITY_DIR";
 pub const ENV_IDENTITY_TRUST_ANCHORS: &str = "LINKERD2_PROXY_IDENTITY_TRUST_ANCHORS";
 pub const ENV_IDENTITY_IDENTITY_LOCAL_NAME: &str = "LINKERD2_PROXY_IDENTITY_LOCAL_NAME";
-pub const ENV_IDENTITY_TAP_IDENTITY: &str = "LINKERD2_PROXY_IDENTITY_TAP_IDENTITY";
 pub const ENV_IDENTITY_TOKEN_FILE: &str = "LINKERD2_PROXY_IDENTITY_TOKEN_FILE";
 pub const ENV_IDENTITY_MIN_REFRESH: &str = "LINKERD2_PROXY_IDENTITY_MIN_REFRESH";
 pub const ENV_IDENTITY_MAX_REFRESH: &str = "LINKERD2_PROXY_IDENTITY_MAX_REFRESH";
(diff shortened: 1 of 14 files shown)


src/app/config.rs
```diff
@@ -265,6 +268,7 @@ pub const ENV_IDENTITY_DISABLED: &str = "LINKERD2_PROXY_IDENTITY_DISABLED";
 pub const ENV_IDENTITY_DIR: &str = "LINKERD2_PROXY_IDENTITY_DIR";
 pub const ENV_IDENTITY_TRUST_ANCHORS: &str = "LINKERD2_PROXY_IDENTITY_TRUST_ANCHORS";
 pub const ENV_IDENTITY_IDENTITY_LOCAL_NAME: &str = "LINKERD2_PROXY_IDENTITY_LOCAL_NAME";
+pub const ENV_IDENTITY_TAP_IDENTITY: &str = "LINKERD2_PROXY_IDENTITY_TAP_IDENTITY";
```
adleong: This doesn't really have anything to do with proxy identity so I'd maybe put this with `ENV_TAP_DISABLED` instead of with the proxy identity vars. Also I'd consider naming it something like `LINKERD2_PROXY_TAP_NAME`.

src/app/tap.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    let fut = {
+        let log = log.clone();
+        bound_port
+            .listen_and_fold(new_service, move |mut new_service, (session, remote)| {
+                let log = log.clone().with_remote(remote);
+                let log_context = log.clone();
+
+                if let Conditional::Some(ref tap_identity) = tap_identity {
+                    debug!("expected Tap client identity: {:?}", tap_identity);
```
adleong: Should this message say "expecting" rather than "expected"?  to me, saying "expected" reads as if this expectation has failed

src/proxy/grpc/service.rs
```diff
@@ -111,3 +111,24 @@ pub mod res_body_as_payload {
...
+
+pub mod unauthenticated {
+    use api::tap as api;
+    use futures::{future, stream};
+    use tower_grpc::{Code, Request, Response, Status};
+
+    #[derive(Clone)]
+    pub struct Unauthenticated;
+
+    impl api::server::Tap for Unauthenticated {
```
adleong: When the client is not authenticated or authorized, is it better to accept the connection and serve back Unauthenticated errors like this?  Or simply reject the connection?

src/app/tap.rs
```diff
@@ -0,0 +1,99 @@
...
+use svc;
+use tokio::executor;
+use tower_grpc as grpc;
+use transport::{
+    tls::{self, HasPeerIdentity},
+    Listen,
+};
+use Conditional;
+
+macro_rules! tap_server {
```
adleong: honest question: are these macros worth it?  they're not very large and they're only invoked twice.

another honest question: why do these need to be macros rather than functions?

src/app/config.rs
kleimkuhler: Good point! I'll change that

src/proxy/grpc/service.rs
kleimkuhler: So we could just reject the connection, and it would remove a majority of the changes this PR introduces. However, mainly in testing and especially for debugging tap, I found it useful to be able to see that a client is unauthenticated from tapping a certain pod. I would prefer to keep this, but I'm open to reason on why it should be removed.

src/app/tap.rs
kleimkuhler: They may not be worth it. They do save some lines of code, but as a reader it does introduce some complexity in figuring out what exactly is going on.

They are macros, because as functions I felt like I was spending more time making the types generic enough to satisfy both call sites. They became more complex than I thought they needed to be.

I could remove these or take another go at making them functions, but either way it seemed like a good neutral point.

src/app/config.rs
```diff
@@ -280,6 +283,7 @@ pub const ENV_CONTROL_EXP_BACKOFF_MIN: &str = "LINKERD2_PROXY_CONTROL_EXP_BACKOF
 pub const ENV_CONTROL_EXP_BACKOFF_MAX: &str = "LINKERD2_PROXY_CONTROL_EXP_BACKOFF_MAX";
 pub const ENV_CONTROL_EXP_BACKOFF_JITTER: &str = "LINKERD2_PROXY_CONTROL_EXP_BACKOFF_JITTER";
 pub const ENV_TAP_DISABLED: &str = "LINKERD2_PROXY_TAP_DISABLED";
+pub const ENV_TAP_SVC_NAME: &str = "LINKERD2_PROXY_TAP_SVC_NAME";
```
hawkw: I think this should maybe be `..._TAP_SVC_IDENTITY`, rather than `..._TAP_SVC_NAME`, since (IIRC) this is the _service account_ of the tap service rather than the service _name_?

tests/support/proxy.rs
```diff
@@ -9,8 +9,12 @@ pub fn new() -> Proxy {
 pub struct Proxy {
     controller: Option<controller::Listening>,
     identity: Option<controller::Listening>,
-    inbound: Option<server::Listening>,
-    outbound: Option<server::Listening>,
+
+    inbound: Option<SocketAddr>,
+    outbound: Option<SocketAddr>,
+
+    inbound_server: Option<server::Listening>,
```
hawkw: I think it would be good to get a comment here explaining what the difference between `{inbound, outbound}` and `{inbound, outbound}_server` is.

src/app/tap.rs
```diff
@@ -0,0 +1,99 @@
...
+                if let Conditional::Some(ref tap_svc_name) = tap_svc_name {
+                    debug!("expecting Tap client name: {:?}", tap_svc_name);
+
+                    let is_expected_name = match session.peer_identity() {
+                        Conditional::Some(ref peer_name) => {
+                            debug!("found Tap client name: {:?}", peer_name);
+                            peer_name == tap_svc_name
+                        }
+                        _ => {
+                            debug!("did not find Tap client name");
```
hawkw: Nit/TIOLI: Since you're matching on the `Conditional::None` case, perhaps this should include the reason there isn't an identity for the tap client?

src/app/tap.rs
```diff
@@ -0,0 +1,99 @@
...
+    let log = logging::admin().server("tap", bound_port.local_addr());
+
+    let fut = {
+        let log = log.clone();
+        bound_port
+            .listen_and_fold(new_service, move |mut new_service, (session, remote)| {
+                let log = log.clone().with_remote(remote);
+                let log_context = log.clone();
+
+                if let Conditional::Some(ref tap_svc_name) = tap_svc_name {
```
hawkw: Nit/TIOLI: Throughout this function, you use `..._name` to refer to the tap service's identity. I think it would be more descriptive to use variable names like `..._identity` or `..._id` instead.

src/app/tap.rs
hawkw: Personally, I think the use of macros for repeated code is fine, since Rust macros don't really introduce new potential bugs the way that, say, the C preprocessor does. I'm not _convinced_ they're strictly necessary, since the code in question is repeated twice, but I do think it's good to factor it out, so that any future changes can be made in one place rather than two.

The main thing I'd change is that I don't _really_ understand why it's necessary to have two separate macros --- it looks like the output of `tap_service!` is only ever passed directly into `tap_task!`. It seems like we could combine them into one macro?

src/app/config.rs
kleimkuhler: I went with `..._TAP_SVC_NAME` since that the pattern we used for `..._DISCOVERY_SVC_NAME` and `..._IDENTITY_SVC_NAME`

src/app/tap.rs
kleimkuhler: We go back and forth on `..._name` and `..._id`/`..._identity` a decent amount. I ultimately went with `..._name` since the environment variable is `..._NAME`. The fact that we call it a proxy's `..._LOCAL_NAME` and such,  I decided to continue that naming (heh). If you feel any stronger than a nit let me know, but I will otherwise plan to keep it!

src/app/tap.rs
kleimkuhler: They were separated because of type inference errors that occur due to the difference in how we create `srv` between the two call sites--the `map_err` and `and_then` in the accept branch specifically.

src/app/config.rs
hawkw: Upon further investigation, turns out I'm wrong about this; the `DESTINATION_SVC_NAME` and `IDENTITY_SVC_NAME` are expected to be service account names, not service names, so this env var should follow that pattern. 👍

src/app/tap.rs
hawkw: I don't really care. IMO naming it `id` or `identity` makes it much clearer to someone looking at this file, and I don't think that being the same as the name of the env var is as important as a reader understanding what the variable _is_ without having to follow it back up the call tree. But, this is just a nit, so call it whatever you prefer.

kleimkuhler: Okay I'm going to stick with `..._name` here. I could flip back and forth just as easy, but the fact that the expected value is from an environment variable called `..._SVC_NAME` is enough for me to keep as is.

kleimkuhler:
@adleong @hawkw The comments I've left unresolved are there if you have any more questions/feedback, otherwise reviews have been addressed!

src/app/tap.rs
adleong: I'm more concerned about readability than introducing bugs

kleimkuhler: As functions they are no more readable than macros unfortunately. The signatures--similar to `serve_tap`--are longer than the actual bodies. I can remove them to help with readability.

adleong: Just so I understand... the reason why this can't be a function is because the return type is some horrible insane type?  And if we just shove that in a let binding, type inference will just silently know what type it is.  But if we make this into a function then suddenly we have to write down the name of that type and our eyeballs will melt out of our faces?

And you can't have a function with an inferred return type?

Rust is weird, y'all.

hawkw: @kleimkuhler I took a crack at replacing both macros with a single function. Here's the diff: https://github.com/linkerd/linkerd2-proxy/compare/kleimkuhler/tap-identity-chk...eliza/kleimkuhler/tap-identity-chk?expand=1#diff-477a4e74a1956a896487a859ef4fc481

If you're interested, I can open a PR against this branch to merge those changes.

kleimkuhler: @hawkw This is awesome thanks! It is merged now

PR #291, opened 2019-07-26
Properly fall back in the dst_router

kleimkuhler:
### Motivation

TLDR: Requests whose target authority is a socket address do not properly fall
back. Requests to unresolvable authorities which are DNS names do fall back as
expected.

In a control plane branch, I have the tap controller properly adding the header
metadata to tap requests, and the pod proxy’s expecting the tap identity the
requests would be coming from. After installing a local build of this control
plane and the proxy that supports `l5d-require-id`, I noticed that the tap
controller’s proxy was still not setting identity on `outbound::Endpoint`s when
the header was there.

I began looking for the `debug!` lines that would be hit in the
`orig_dst_router` because I would expect to be hitting that router, and did not
see them—indicating those requests were still going through the balancer.

It turns out that requests with an authority that we cannot resolve still get a
single slot in the balancer, and therefore do not fallback because they do not
return the `Unresolvable` error. We match on `Socket` [here](https://github.com/linkerd/linkerd2-proxy/blob/master/src/app/outbound.rs#L207) and make a new
slot in the balancer [here](https://github.com/linkerd/linkerd2-proxy/blob/master/src/app/outbound.rs#L276)

This is an issue because if these requests contain the `l5d-require-id` header,
the `Endpoint` `identity` is not set because we do not go through the
`orig_dst_router` and therefore do not have a TLS connection

### Solution

In `outbound::discover`, `Resolve` will now only attempt to resolve
`Addr::Name`, and return a `Unresolvable` error on `Addr::Socket`. For requests
that are `Addr::Socket`, we rely on getting the original dst of the request and
therefore routes through `orig_dst_router`.

### Tests

The tests that specifically test SO_ORIGINAL_DST have been removed because those
tests now result in errors which is the expected behavior now.

### DIFF
--- a/src/app/outbound.rs
+++ b/src/app/outbound.rs
@@ -166,7 +166,7 @@ pub mod discovery {
     use {Addr, Conditional, NameAddr};
 
     #[derive(Clone)]
-    pub struct Resolver<R: resolve::Resolve<NameAddr>>(R);
+    pub struct Resolve<R: resolve::Resolve<NameAddr>>(R);
 
     pub struct ResolveFuture<F> {
         resolving: Option<Resolution<F>>,
@@ -180,18 +180,19 @@ pub mod discovery {
 
     // ===== impl Resolver =====
 
-    impl<R> Resolver<R>
+    impl<R> Resolve<R>
     where
         R: resolve::Resolve<NameAddr, Endpoint = Metadata>,
     {
         pub fn new(resolver: R) -> Self {
-            Resolver(resolver)
+            Resolve(resolver)
         }
     }
 
-    impl<R> resolve::Resolve<DstAddr> for Resolver<R>
+    impl<R> resolve::Resolve<DstAddr> for Resolve<R>
     where
         R: resolve::Resolve<NameAddr, Endpoint = Metadata>,
+        <R as resolve::Resolve<NameAddr>>::Future: Future<Error = Unresolvable>,
     {
         type Endpoint = Endpoint;
         type Resolution = Resolution<R::Resolution>;
@@ -227,7 +228,7 @@ pub mod discovery {
 
     impl<F: Future> Future for ResolveFuture<F>
     where
-        F: Future,
+        F: Future<Error = Unresolvable>,
     {
         type Item = Resolution<F::Item>;
         type Error = Unresolvable;
@@ -239,14 +240,7 @@ pub mod discovery {
                 ref mut resolution,
             }) = self.resolving
             {
-                let resolution = match resolution.poll() {
-                    Ok(Async::Ready(resolution)) => resolution,
-                    Ok(Async::NotReady) => return Ok(Async::NotReady),
-                    Err(_) => {
-                        trace!("name={} is unresolvable", name);
-                        return Err(Unresolvable::new());
-                    }
-                };
+                let resolution = try_ready!(resolution.poll());
 
                 return Ok(Async::Ready(Resolution {
                     http_settings: http_settings.clone(),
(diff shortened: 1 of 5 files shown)


src/proxy/server.rs
```diff
@@ -95,11 +95,19 @@ pub struct NoOriginalDst;
 impl Source {
     pub fn orig_dst_if_not_local(&self) -> Option<SocketAddr> {
         match self.orig_dst {
-            None => None,
+            None => {
+                trace!("no SO_ORIGINAL_DST on source");
```
kleimkuhler: The `trace!`s in this file are not needed, but I found them helpful when testing and therefore probably a good reason to keep around.

hawkw:
> ### Motivation
> 
> TLDR: Fallback layer was not actually falling back on requests that cannot be
> resolved

This is _specifically_ in the case of requests whose target authority is a socket address. Requests to unresolvable authorities which are DNS names _do_ fall back as expected.

Just felt like that should be clarified.

src/app/outbound.rs
```diff
@@ -157,85 +157,111 @@ impl tap::Inspect for Endpoint {
...
+            if let Some(Resolution {
+                http_settings,
+                ref name,
+                ref mut resolution,
+            }) = self.resolving
+            {
+                let resolution = match resolution.poll() {
+                    Ok(Async::Ready(resolution)) => resolution,
+                    Ok(Async::NotReady) => return Ok(Async::NotReady),
+                    Err(_) => {
```
hawkw: Do we know that `Unresolvable` is the only error that the `control::destination` resolver will return?

src/app/outbound.rs
```diff
@@ -157,85 +157,111 @@ impl tap::Inspect for Endpoint {
...
-                // TODO: get rid of unnecessary clone
-                http_settings: self.http_settings.clone(),
-            }))
+            if let Some(Resolution {
+                http_settings,
+                ref name,
+                ref mut resolution,
+            }) = self.resolving
+            {
+                let resolution = match resolution.poll() {
```
hawkw: Can't this just be `try_ready!`?

kleimkuhler:
@hawkw Yes thank you for pointing out the more specific case of this! I updated the comment.

src/app/outbound.rs
```diff
@@ -157,85 +157,111 @@ impl tap::Inspect for Endpoint {
...
+    use control::destination::{Metadata, Unresolvable};
+    use futures::{future::Future, Async, Poll};
     use proxy::{http::settings, resolve};
     use transport::tls;
     use {Addr, Conditional, NameAddr};
 
-    #[derive(Clone, Debug)]
-    pub struct Resolve<R: resolve::Resolve<NameAddr>>(R);
+    #[derive(Clone)]
+    pub struct Resolver<R: resolve::Resolve<NameAddr>>(R);
```
hawkw: I feel like it _might_ be worth considering just moving all the remaining functionality in this module into `control::destination::Resolver`. It seems like this module is just a pretty thin layer over the `destination` resolver and we could probably put all that functionality in one place. I think a majority of the code here now is just re-implementing some future combinators, and there's only a few lines of actual _logic_ here.

On the other hand, moving that to `control::destination` might be a worse separation of concerns. 🤷‍♀

hawkw:
>   * Debug no longer impl'd

Ah, yeah, I missed this --- can we put debug back @kleimkuhler?

olix0r:
(Why) is it necessary to change the naming scheme for this module?

kleimkuhler:
@adleong Thanks for the great insight about changing the `Resolving` variants to get a minimal diff for this fix! The diff is much smaller now.

I will open a new issue that addresses what @hawkw has pointed out about making some larger changes to this module that remove some complexity around wrapping the destination resolver.

kleimkuhler:
I have also confirmed that this fixes the issue with a local build of this branch injected into my local build of the control plane.

PR #294, opened 2019-07-30
metrics: add `request_handle_us` histogram

hawkw:
As initially described in linkerd/linkerd2#730, there's a general need
to understand the proxy's latency overhead. We can begin by recording
the time a request spends in the proxy. 

linkerd/linkerd2#3098 proposes adding a new `request_handle_us`
histogram to the proxy with only a `direction` label. This histogram 
should store the elapsed time (in microseconds) from the moment a 
request reaches the source stack until the request is dropped.

This branch adds the `request_handle_us` histogram to the proxy's 
metrics. Handle time is recorded using a `Tracker` type which is 
inserted into each request's extensions map by a new layer at the
top of the stack. Each tracker has a reference count, and when 
requests are cloned for retries, the tracker is cloned into the new
request as well, incrementing the reference count. The reference 
count is decremented when trackers are dropped, and when it reaches
zero, the elapsed time is recorded.

Since this depends on a time measurement that's internal to the proxy,
it's difficult to write reliable tests for it. Unlike the latency
histograms, we cannot use a test service that waits for a period of time
to ensure that the recorded latencies have a lower bound. Instead, I did
some manual verification that the new metric is present in the proxy's
metrics endpoint, that its total count is equal to the number of
requests I've sent through the proxy, and that the handle time
measurements appear to be less than the total request latency
measurements.

Closes linkerd/linkerd2#3098 

### DIFF
--- a/lib/metrics/src/latency.rs
+++ b/lib/metrics/src/latency.rs
@@ -38,6 +38,33 @@ pub const BOUNDS: &Bounds = &Bounds(&[
 #[derive(Debug, Default, Clone)]
 pub struct Ms(Duration);
 
+/// A duration in microseconds.
+#[derive(Debug, Default, Clone)]
+pub struct Us(Duration);
+
+impl Into<u64> for Us {
+    fn into(self) -> u64 {
+        use std::convert::TryInto;
+        self.0.as_micros().try_into().unwrap_or_else(|_| {
+            // These measurements should never be long enough to overflow
+            // warn!("Duration::as_micros would overflow u64");
+            std::u64::MAX
+        })
+    }
+}
+
+impl From<Duration> for Us {
+    fn from(d: Duration) -> Self {
+        Us(d)
+    }
+}
+
+impl Default for Histogram<Us> {
+    fn default() -> Self {
+        Histogram::new(BOUNDS)
+    }
+}
+
 impl Into<u64> for Ms {
     fn into(self) -> u64 {
         self.0
--- a/lib/metrics/src/latency.rs
+++ b/lib/metrics/src/latency.rs
@@ -47,7 +47,7 @@ impl Into<u64> for Us {
         use std::convert::TryInto;
         self.0.as_micros().try_into().unwrap_or_else(|_| {
             // These measurements should never be long enough to overflow
-            // warn!("Duration::as_micros would overflow u64");
+            tracing::warn!("Duration::as_micros would overflow u64");
             std::u64::MAX
         })
     }
--- a/src/app/handle_time.rs
+++ b/src/app/handle_time.rs
@@ -3,22 +3,21 @@ use metrics::{Metric, FmtMetrics};
 use super::metric_labels::Direction;
 use std::{iter, fmt};
 
-pub struct Scopes {
+#[derive(Clone, Debug)]
+pub struct Metrics {
     inbound: handle_time::Scope,
     outbound: handle_time::Scope,
-    metric: Metric<'static, handle_time::Scope>,
 }
 
-impl Scopes {
+impl Metrics {
     pub const HELP: &'static str =
         "A histogram of the time in microseconds between when a request is received and when it is sent upstream.";
     pub const NAME: &'static str = "request_handle_us";
 
     pub fn new() -> Self {
-        Scopes {
+        Self {
             inbound: handle_time::Scope::new(),
             outbound: handle_time::Scope::new(),
-            metric: Metric::new(Self::NAME, Self::HELP),
         }
     }
 
@@ -30,6 +29,10 @@ impl Scopes {
         self.inbound.clone()
     }
 
+    fn metric(&self) -> Metric<handle_time::Scope> {
+        Metric::new(Self::NAME, Self::HELP)
+    }
+
     fn scopes<'a>(&'a self) -> impl Iterator<Item = (Direction, &'a handle_time::Scope)> {
         iter::once((Direction::In, &self.inbound))
             .chain(iter::once((Direction::Out, &self.outbound)))
@@ -37,9 +40,10 @@ impl Scopes {
 }
 
 
-impl FmtMetrics for Scopes {
+impl FmtMetrics for Metrics {
     fn fmt_metrics(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        self.metric.fmt_help(f)?;
-        self.metric.fmt_scopes(f, self.scopes(), |s| s)
+        let metric = self.metric();
+        metric.fmt_help(f)?;
+        metric.fmt_scopes(f, self.scopes(), |s| s)
     }
 }
--- a/src/app/handle_time.rs
+++ b/src/app/handle_time.rs
@@ -1,7 +1,7 @@
-use proxy::http::metrics::handle_time;
-use metrics::{Metric, FmtMetrics};
 use super::metric_labels::Direction;
-use std::{iter, fmt};
+use metrics::{FmtMetrics, Metric};
+use proxy::http::metrics::handle_time;
+use std::{fmt, iter};
 
 #[derive(Clone, Debug)]
 pub struct Metrics {
@@ -38,7 +38,6 @@ impl Metrics {
             .chain(iter::once((Direction::Out, &self.outbound)))
     }
 }
-
 
 impl FmtMetrics for Metrics {
     fn fmt_metrics(&self, f: &mut fmt::Formatter) -> fmt::Result {
--- a/src/app/handle_time.rs
+++ b/src/app/handle_time.rs
@@ -1,0 +1,45 @@
+use proxy::http::metrics::handle_time;
+use metrics::{Metric, FmtMetrics};
+use super::metric_labels::Direction;
+use std::{iter, fmt};
+
+pub struct Scopes {
+    inbound: handle_time::Scope,
+    outbound: handle_time::Scope,
+    metric: Metric<'static, handle_time::Scope>,
+}
+
+impl Scopes {
+    pub const HELP: &'static str =
+        "A histogram of the time in microseconds between when a request is received and when it is sent upstream.";
+    pub const NAME: &'static str = "request_handle_us";
+
+    pub fn new() -> Self {
+        Scopes {
+            inbound: handle_time::Scope::new(),
+            outbound: handle_time::Scope::new(),
+            metric: Metric::new(Self::NAME, Self::HELP),
+        }
+    }
+
+    pub fn outbound(&self) -> handle_time::Scope {
+        self.outbound.clone()
+    }
+
+    pub fn inbound(&self) -> handle_time::Scope {
+        self.inbound.clone()
+    }
+
+    fn scopes<'a>(&'a self) -> impl Iterator<Item = (Direction, &'a handle_time::Scope)> {
+        iter::once((Direction::In, &self.inbound))
+            .chain(iter::once((Direction::Out, &self.outbound)))
+    }
+}
+
+
+impl FmtMetrics for Scopes {
+    fn fmt_metrics(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        self.metric.fmt_help(f)?;
+        self.metric.fmt_scopes(f, self.scopes(), |s| s)
+    }
+}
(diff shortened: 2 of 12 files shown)


src/proxy/http/insert.rs
```diff
@@ -44,16 +44,26 @@ pub struct ValLazy<V>(V);
 
 pub fn layer<F, V>(f: F) -> Layer<FnLazy<F>, V>
 where
-    F: Fn() -> V,
+    F: Fn() -> V + Clone,
```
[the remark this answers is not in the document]
hawkw: "compiler said so" --- the `new` function I added has a `Lazy` bound:
https://github.com/linkerd/linkerd2-proxy/blob/0231f60979e9d988635c1f0f90b2a72a677681c3/src/proxy/http/insert.rs#L55-L57
and `FnLazy<F>` is only `Lazy` when `F: Clone`:
https://github.com/linkerd/linkerd2-proxy/blob/0231f60979e9d988635c1f0f90b2a72a677681c3/src/proxy/http/insert.rs#L168-L171

alternatively, we could get rid of the `new`, and not have this bound here, but it would allow constructing `Layer`s that don't _actually_ implement `Layer`, resulting in a much bigger & harder to debug type error later...

lib/metrics/src/latency.rs
```diff
@@ -38,6 +38,33 @@ pub const BOUNDS: &Bounds = &Bounds(&[
...
+/// A duration in microseconds.
+#[derive(Debug, Default, Clone)]
+pub struct Us(Duration);
+
+impl Into<u64> for Us {
+    fn into(self) -> u64 {
+        use std::convert::TryInto;
+        self.0.as_micros().try_into().unwrap_or_else(|_| {
+            // These measurements should never be long enough to overflow
+            // warn!("Duration::as_micros would overflow u64");
```
kleimkuhler: Looks like this should be uncommented

src/proxy/http/metrics/handle_time.rs
```diff
@@ -0,0 +1,235 @@
...
+    shared: Arc<Shared>,
+    idx: usize,
+    t0: Instant,
+}
+
+#[derive(Debug)]
+struct Shared {
+    // NOTE: this is inside a `Mutex` since recording a latency requires a mutable
+    // reference to the histogram. In the future, we could consider making the
+    // histogram counters `AtomicU64, so that the histogram could be updated
```
kleimkuhler: Missing the closing ` on AtomicU64

src/proxy/http/metrics/handle_time.rs
```diff
@@ -0,0 +1,235 @@
...
+    // with an immutable reference. Then, the mutex could be removed.
+    histogram: Mutex<Histogram<latency::Us>>,
+    // Stores the state of currently active `Tracker`s.
+    recorders: RwLock<Vec<Recorder>>,
+    // The next free index into `recorders`
+    next_recorder: AtomicUsize,
+}
+
+#[derive(Debug)]
+struct Recorder {
```
kleimkuhler: You mentioned this a little in the PR comment, but is the only purpose of `ref_count` in `Recorder` that fact that there can multiple references to it in the case of retries? I found this a little confusing to understand until coming back to that comment.

Also are there cases when next is not just `i + 1`?

src/proxy/http/metrics/handle_time.rs
```diff
@@ -0,0 +1,235 @@
...
+            }
+
+            // The recorder at `idx` was not actually free! Try again with a
+            // fresh index.
+            atomic::spin_loop_hint()
+        }
+    }
+
+    #[cold]
+    #[inline(never)]
```
kleimkuhler: I have an idea of what this attribute does, but what about this function made you add it?

src/proxy/http/metrics/handle_time.rs
hawkw: > You mentioned this a little in the PR comment, but is the only purpose of `ref_count` in `Recorder` that fact that there can multiple references to it in the case of retries? I found this a little confusing to understand until coming back to that comment.

Perhaps the name could be made clearer...would `trackers` or `clones` be better?

> Also are there cases when next is not just `i + 1`?

Yes, when all the `Tracker`s for a given index are dropped, it's reused so we don't have to allocate more storage for the next request. Here's the relevant code:
https://github.com/linkerd/linkerd2-proxy/blob/e8f5d136d4362e76ea678a1d89364e354b85ed8b/src/proxy/http/metrics/handle_time.rs#L140-L174

kleimkuhler: Oh okay I see [here](https://github.com/linkerd/linkerd2-proxy/blob/0231f60979e9d988635c1f0f90b2a72a677681c3/src/proxy/http/metrics/handle_time.rs#L225-L226) that the swap makes sure that `recorders.len()` always should return the first free index.

hawkw:
@kleimkuhler 30966f4f6c2626d9427f95172c32b84a4f9880d4 makes some naming changes & adds more comments, let me know if that clears up some of the stuff you had questions about?

kleimkuhler:
@hawkw Those comments and change in naming are helpful. Thanks for doing that!

src/proxy/http/metrics/handle_time.rs
```diff
@@ -0,0 +1,252 @@
...
+impl insert::Lazy<Tracker> for InsertTracker {
+    fn value(&self) -> Tracker {
+        self.0.clone().tracker()
+    }
+}
+
+// ===== impl Scope =====
+
+impl Scope {
+    pub fn new() -> Self {
```
olix0r: nit: i tend to prefer the module fn pattern of pub fn handle_time() -> (Scope, InsertLayer) -- using the scope to produce layers is fine, but also doesn't really seem necessary (since the layer should be clone)...

src/app/handle_time.rs
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+
+    pub fn outbound(&self) -> handle_time::Scope {
+        self.outbound.clone()
+    }
+
+    pub fn inbound(&self) -> handle_time::Scope {
+        self.inbound.clone()
+    }
```
olix0r: Similarly, I'd be inclined to make a constructor that returns a 3-tuple rather than to hide the scopes in this struct.

src/proxy/http/metrics/handle_time.rs
olix0r: I'm also curious.

src/proxy/http/metrics/handle_time.rs
```diff
@@ -0,0 +1,252 @@
...
+}
+
+impl Shared {
+    const INITIAL_RECORDERS: usize = 32;
+
+    fn new() -> Self {
+        let mut recorders = Vec::with_capacity(Self::INITIAL_RECORDERS);
+        Self::add_recorders(&mut recorders, Self::INITIAL_RECORDERS);
+        Self {
+            histogram: Mutex::new(Histogram::default()), // TODO(eliza): should we change the bounds here?
```
olix0r: to what? I don't understand the comment

src/proxy/http/metrics/handle_time.rs
```diff
@@ -0,0 +1,252 @@
...
+                let recorders = self
+                    .recorders
+                    .read()
+                    .ok()
+                    .filter(|recorders| idx < recorders.len())
+                    .unwrap_or_else(|| {
+                        // Slow path: if there are no free recorders in the
+                        // slab, extend it (acquiring a write lock temporarily).
+                        self.grow();
+                        self.recorders.read().unwrap()
```
olix0r: if we unwrap here, why not unwrap above (where you use `ok()`)?

src/proxy/http/metrics/handle_time.rs
```diff
@@ -0,0 +1,252 @@
...
+    /// and that recorder's next index will be set as the head. of the free
+    /// list.
+    ///
+    /// When an active recorder completes, this will be set to its index, and
+    /// the previous value will become the freed recorder's next pointer.
+    idle_head: AtomicUsize,
+}
+
+#[derive(Debug)]
+struct Recorder {
```
olix0r: I don't really understand the role of `Recorder` in this. My naive assumption is that this should all work without needing to store the rwlock of recorders. When would an individual Recorder have more than one tracker? Why would a scope have more than one recorder?

It would be helpful if the comments addressed this more directly.

src/proxy/http/metrics/handle_time.rs
hawkw: The `ok` above could also be an `unwrap`; the reason it's `ok()` is because we are optionally returning a recorder from the `filter`, and the `ok()` lets us nicely chain the option, without the `ok()` the code is a bit more complex. I can make this consistent if that's preferred.

olix0r: In general, I find the `filter` harder to read, and it's confusing because it at first seems like we're graceful to the lock being poisoned but, in fact, we just panic in the next access in that case.

```rust
let sz = self.recorders.read().unwrap().len();
if sz < idx {
  self.grow();
}
self.recorders.read().unwrap()
```

Is clearer about what it's doing, imo

src/proxy/http/metrics/handle_time.rs
hawkw: Requests may be cloned for retries, and we want to include any retries in the handle time for a request. The scope will have one `Recorder` for each request that is currently being tracked, and if that request is cloned, there will be a `Tracker` inserted into each clone's extension, incrementing the count. I changed some naming & added comments in 159b6730b9d928cb213cf4ee2b1f00657f385f76, hopefully the purpose of this is a little clearer.

The alternative to using the slab of counters is to create a new `Arc<()>` for each tracker, and record the handle time when that `Arc` is dropped; this would mean making a small heap allocation for every request. The slab lets us amortize the cost of storing the counters.

src/proxy/http/metrics/handle_time.rs
hawkw: Ah, I should probably have said "_bucket_ bounds" here, the type is just called `Bounds` in the histogram module. 

Basically, I'm not convinced that [these buckets](https://github.com/linkerd/linkerd2-proxy/blob/d315b9f6afb23eb3d0cf58577710989d1bd69944/lib/metrics/src/latency.rs#L7-L35) provide great resolution for handle times, given that it seems like a majority of them fall into the 1-2ms range; it might be worth tuning the buckets a bit.

olix0r: Yeah, I'm not convinced that those buckets are a good choice, either. Would you mind opening an issue in the proxy visibility backlog re: tuning histograms for fidelity.

src/app/handle_time.rs
hawkw: I'd be happy to change all this code to return tuples, but I'm not sure if it's the right thing in this specific case. I think using tuples as return types is fine when the types in the tuple are heterogeneous (the compiler will notice if you try to call a `Sender` method on a `Receiver`, for example), but I'm a little leery of relying on tuple ordering to differentiate two different instances of the same type, like inbound/outbound.

src/proxy/http/metrics/handle_time.rs
hawkw: The intention was to avoid reacquiring the read lock in the fast path. The code you posted is definitely more readable but will always acquire the read lock twice, even when we don't need to acquire the write lock. It is entirely possible that the performance hit from reacquiring it is not significant enough to justify making the code less clear, though.

src/proxy/http/metrics/handle_time.rs
hawkw: https://github.com/linkerd/linkerd2/issues/3202

src/proxy/http/metrics/handle_time.rs
hawkw: `#[cold]` is a hint to the compiler that it should use a calling convention that improves performance at the callsite under the assumption that a function is called infrequently: https://llvm.org/docs/LangRef.html#calling-conventions

in retrospect, I realise that `#[cold]` implies that a function should not be inlined, so I think the `#[inline(never)]` should be removed.

hawkw:
will merge this pending CI

PR #296, opened 2019-08-05
Add dst and route labels to request and response metrics

adleong:
The `request_total` and `response_total` metrics are measured in the endpoint stack and do not include a label which indicates the dst of the request.  These metrics do have an `authority` label, but this label's value is taken from the target of the endpoint stack.  In the case of a traffic-split, the `authority` label corresponds to the overridden (concrete) dst, and not to the authority of the request itself.  In order to know the logical dst of the request, a new label is needed.

We add a `dst` label by recording the target of the dst stack into the request extensions.  We then extract this value in the endpoint stack and use it as a label value.  Similarly, we also record the route (the target of the route stack) into the request extensions and then use this as a label in the endpoint stat metrics as well.  The route label is not related to traffic-splits, but seems like useful information anyway.

## Labels
```
authority: the concrete dst (authors-v1.default.svc.cluster.local:7001)
dst_service: the concrete service (authors-v1)
dst: the logical dst (authors.default.svc.cluster.local:7001) [NEW!]
rt_rotue: the logical route (GET /authors.json) [NEW!]
```

One side-effect of this change is that the default route now has a route label with an empty value.  This means that proxy metrics for requests without a route will now include the `rt_route=""` label.  Prometheus seems to filter out these empty value labels and these are not present in Prometheus.

### DIFF
--- a/src/app/metric_labels.rs
+++ b/src/app/metric_labels.rs
@@ -74,6 +74,17 @@ impl From<dst::Route> for RouteLabels {
     }
 }
 
+impl FmtLabels for dst::Route {
+    fn fmt_labels(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let labels = prefix_labels("rt", self.labels().as_ref().into_iter());
+        if let Some(labels) = labels.as_ref() {
+            write!(f, "{}", labels)?;
+        }
+
+        Ok(())
+    }
+}
+
 impl FmtLabels for RouteLabels {
     fn fmt_labels(&self, f: &mut fmt::Formatter) -> fmt::Result {
         self.dst.fmt_labels(f)?;
@@ -171,6 +182,12 @@ impl FmtLabels for dst::DstAddr {
         }
 
         write!(f, ",dst=\"{}\"", self.as_ref())
+    }
+}
+
+impl FmtLabels for Addr {
+    fn fmt_labels(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "dst=\"{}\"", self)
     }
 }
 
--- a/src/proxy/http/metrics/mod.rs
+++ b/src/proxy/http/metrics/mod.rs
@@ -5,6 +5,7 @@ use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 use tokio_timer::clock;
 
+use addr::Addr;
 use metrics::{latency, Counter, FmtLabels, Histogram};
 
 pub mod classify;
@@ -14,9 +15,10 @@ mod service;
 pub use self::report::Report;
 pub use self::service::layer;
 
-pub fn new<T, C>(retain_idle: Duration) -> (Arc<Mutex<Registry<T, C>>>, Report<T, C>)
+pub fn new<T, R, C>(retain_idle: Duration) -> (Arc<Mutex<Registry<T, R, C>>>, Report<T, R, C>)
 where
     T: FmtLabels + Clone + Hash + Eq,
+    R: FmtLabels + Hash + Eq,
     C: FmtLabels + Hash + Eq,
 {
     let registry = Arc::new(Mutex::new(Registry::default()));
@@ -24,12 +26,13 @@ where
 }
 
 #[derive(Debug)]
-pub struct Registry<T, C>
+pub struct Registry<T, R, C>
 where
     T: Hash + Eq,
+    R: Hash + Eq,
     C: Hash + Eq,
 {
-    by_target: IndexMap<T, Arc<Mutex<RequestMetrics<C>>>>,
+    by_target: IndexMap<T, Arc<Mutex<RequestMetrics<R, C>>>>,
 }
 
 pub trait Scoped<T> {
@@ -42,13 +45,31 @@ pub trait Stats {
 }
 
 #[derive(Debug)]
-pub struct RequestMetrics<C>
+pub struct RequestMetrics<R, C>
 where
+    R: Hash + Eq,
     C: Hash + Eq,
 {
     last_update: Instant,
-    total: Counter,
     by_retry_skipped: IndexMap<RetrySkipped, Counter>,
+    by_dst: IndexMap<Option<Addr>, DstMetrics<R, C>>,
+}
+
+#[derive(Debug)]
+struct DstMetrics<R, C>
+where
+    R: Hash + Eq,
+    C: Hash + Eq,
+{
+    by_route: IndexMap<Option<R>, RouteMetrics<C>>,
+}
+
+#[derive(Debug)]
+struct RouteMetrics<C>
+where
+    C: Hash + Eq,
+{
+    total: Counter,
     by_status: IndexMap<Option<http::StatusCode>, StatusMetrics<C>>,
 }
 
@@ -71,9 +92,10 @@ enum RetrySkipped {
     Budget,
 }
 
-impl<T, C> Default for Registry<T, C>
+impl<T, R, C> Default for Registry<T, R, C>
 where
     T: Hash + Eq,
+    R: Hash + Eq,
     C: Hash + Eq,
 {
     fn default() -> Self {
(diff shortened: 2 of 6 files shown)


scottcarol:
Excited to check this out. What's the best way to test this out and look at the Prometheus stats? This repo's README recommends using Cargo and `make build`, are there more steps I need to do after that to run with my local Linkerd2 branch?

adleong:
The easiest way would be to use the proxy image I've already pushed: `--proxy-version alex-dst-metrics`.  You can use this flag with either `linkerd install` or `linkerd upgrade`

scottcarol:
Can you confirm that I'm testing this correctly? I have a trafficsplit setup in booksapp; requests are being split between `authors` and `authors-clone` service. The apex is `authors`.

I installed:
```
linkerd upgrade --proxy-version alex-dst-metrics-3 | kubectl apply -f -
```
Then ran `linkerd check --proxy` which passed, with these warnings:
```
‼ data plane is up-to-date
    linkerd/linkerd-controller-6495cf7bc4-z724w: unsupported version channel: alex-dst-metrics-3
    see https://linkerd.io/checks/#l5d-data-plane-version for hints
‼ data plane and cli versions match
    linkerd/linkerd-web-6486cb6d7c-8jplx running alex-dst-metrics-3 but cli running edge-19.8.1
    see https://linkerd.io/checks/#l5d-data-plane-cli-version for hints
```

Then I port forwarded Prometheus:
```
kubectl -n linkerd port-forward $(kubectl --namespace=linkerd get po --selector=linkerd.io/control-plane-component=prometheus -o jsonpath='{.items[*].metadata.name}') 9090:9090
```

When I went to localhost:9090, this request:
```
sum(irate(request_total{direction="outbound",namespace="default",dst_service!=""}[1m]))by(namespace, dst_service, authority) 
```
gave me the results :
```
{authority="authors-clone.default.svc.cluster.local:7001",dst_service="authors-clone",namespace="default"} | 0.7000500070005
{authority="authors.default.svc.cluster.local:7001",dst_service="authors",namespace="default"} | 6.000160060001601
{authority="books.default.svc.cluster.local:7002",dst_service="books",namespace="default"} | 8.200740082007401
{authority="webapp.default.svc.cluster.local:7000",dst_service="webapp",namespace="default"} | 7.401480296059212
```
I was expecting one of the result rows to be going from authority `authors` to dst_service `authors-clone`.

adleong:
@scottcarol your setup is correct except that you should be looking for a result row that goes from `dst=authors.default.svc.cluster.local:7001` to `dst_service=authors-clone`.  The (unfortunately named) `authority` label and the `dst_service` will always match because they both refer to the concrete destination.

scottcarol:
👍 Got it, just needed to delete the pods in my namespace so they could restart. The correct data is showing up!
```
{dst="authors.default.svc.cluster.local:7001",dst_service="authors-clone",namespace="default"} | 0.7002811659432091
{dst="authors.default.svc.cluster.local:7001",dst_service="authors",namespace="default"} | 5.806794395816288
{dst="books.default.svc.cluster.local:7002",dst_service="books",namespace="default"} | 7.89981257158902
{dst="webapp.default.svc.cluster.local:7000",dst_service="webapp",namespace="default"} | 7.22383866760309
```

scottcarol:
Nice, and I'm seeing `rt_route` in my Prometheus metrics when I created a service profile. 💯

src/app/metric_labels.rs
```diff
@@ -74,6 +74,17 @@ impl From<dst::Route> for RouteLabels {
     }
 }
 
+impl FmtLabels for dst::Route {
+    fn fmt_labels(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let labels = prefix_labels("rt", self.labels().as_ref().into_iter());
```
scottcarol: What was your thinking behind the label name `rt_route`? Is the `rt` for route as well? `target_route` seems easier to understand to me, would that be technically correct?

adleong: The `rt_route` label already exists on the route (`route_response_total` et al.) metrics.  This change just puts that label on the endpoint (`response_total` et al.) metrics as well.

The `rt` prefix indicates that this is a route label and `route` is the name of the route label.  The naming is not terribly obvious, but since the metrics are only user-facing for advanced users who are interacting directly with prometheus, the standards for clarity can be somewhat relaxed.

scottcarol: I see, thanks for the explanation. Do we have a doc anywhere where we list the available user-facing labels, and what they mean? Happy to do this if there isn't one.

src/proxy/http/profiles/mod.rs
```diff
@@ -280,3 +280,11 @@ impl fmt::Debug for Labels {
         self.0.fmt(f)
     }
 }
+
+impl Default for Labels {
+    fn default() -> Self {
+        let mut labels = IndexMap::with_capacity(1);
+        labels.insert("route".to_string(), "".to_string());
```
olix0r: Without more context, this is super weird. Why is there a magic string in here? Why is this necessary at all?

olix0r: Specifically, why is it necessary to hydrate an empty label?

src/proxy/http/metrics/mod.rs
```diff
@@ -42,13 +45,31 @@ pub trait Stats {
...
-pub struct RequestMetrics<C>
+pub struct RequestMetrics<R, C>
 where
+    R: Hash + Eq,
     C: Hash + Eq,
 {
     last_update: Instant,
-    total: Counter,
     by_retry_skipped: IndexMap<RetrySkipped, Counter>,
+    by_dst: IndexMap<Option<Addr>, DstMetrics<R, C>>,
```
olix0r: I don't really understand why the metrics registry has to be aware of any of this... It seems like a it should all be a detail of the stack targets...

Do we not have both the logical and concrete dsts in the endpoint stack target? If not, can we?

olix0r:
Yakshave:

* `authority` is _intended_ to reflect the logical destination (i.e. what the request refers to). That it refers to the concrete dst is an accident introduced by traffic split.
* `dst_*` is generally used as a prefix in our metrics, so i'd prefer to continue that with i.e. `dst_concrete`
* I don't understand what `rt_route` has to do with this change. Can it be split into a separate change? Or is this coupled in some way?

src/proxy/http/profiles/mod.rs
adleong: The problem is that if a route has no labels, its FmtLabels implementation will format empty which will cause the total label string to include `,,` which breaks Prometheus.  Adding an empty label is hack to work around this.

src/proxy/http/metrics/mod.rs
adleong: We don't have the logical and concrete dsts in the endpoint stack target, just the concrete.  We could have the logical dst in there, but this would make it so that concrete dst load balancers would not be shared between logical dsts.  

This is why we propagate the logical dst in the request extensions.  Having the logical dst in the target of the endpoint stack would certainly simplify this a lot.  Not sure how to evaluate the tradeoff, though.

adleong:
`rt_route` isn't necessarily coupled to this change.  While plumbing logical dst through the request extensions, I just though I would bring `rt_route` along for the ride since it seems like potential useful information and is similar to logical dst in that they are both from the dst-route stack target.  I can take it off this PR if you think it doesn't make sense here.

src/proxy/http/metrics/mod.rs
olix0r: Per offline discussion: because a concrete lb is not shared across multiple logical names, we actually can have all of the needed metadata in the endpoint stack's target, which should greatly simplify the change.

olix0r:
> `rt_route` isn't necessarily coupled to this change.

let's split out that change, then. there's likely a simplification for that change as well, but it will be easier to address in a follow-up change.

adleong:
Closed in favor of https://github.com/linkerd/linkerd2-proxy/pull/300

PR #300, opened 2019-08-06
Authority label should reflect logical dst

adleong:
Supersedes https://github.com/linkerd/linkerd2-proxy/pull/296

The authority label reflects the concrete dst of the request in the endpoint stack.  In the case of a traffic split, this may be different from the actual authority of the request (i.e. the logical dst).

We add the logical dst as a field on the endpoint target and use this value for the authority label instead of the concrete dst.

The net result is that the authority field reflects the apex service of a traffic split and the `dst_service` label reflects the leaf service.  

E.g.:

```
response_total{authority="authors.default.svc.cluster.local:7001",direction="outbound",dst_control_plane_ns="linkerd",dst_deployment="authors-v2",dst_namespace="default",dst_pod="authors-v2-7cffd4894-crhbw",dst_pod_template_hash="7cffd4894",dst_service="authors-v2",dst_serviceaccount="default",tls="true",server_id="default.default.serviceaccount.identity.linkerd.cluster.local",status_code="201",classification="success"} 5
```

This PR can be tested with the `alex-dst-metrics-4` proxy tag.

### DIFF
--- a/src/app/dst.rs
+++ b/src/app/dst.rs
@@ -32,8 +32,8 @@ pub struct Retry {
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct DstAddr {
-    logical_dst: Addr,
-    concrete_dst: Addr,
+    dst_logical: Addr,
+    dst_concrete: Addr,
     direction: Direction,
     pub(super) http_settings: settings::Settings,
 }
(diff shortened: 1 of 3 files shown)


olix0r:
I'm curious if we should make the `authority` label _only_ refer to the host/authority value and introduce a new label, dst_logical, that includes the value of l5d-dst-override, etc...  But this is probably unnecessary and may make things more complex than they need to be...

scottcarol:
When I tested this, it worked as expected when I queried Prometheus for outbound requests while I had a trafficsplit running. 👍 
```
{authority="authors.default.svc.cluster.local:7001", dst_service="authors"}
{authority="authors.default.svc.cluster.local:7001",dst_service="authors-clone"}
```

olix0r:
@adleong nm. i see flakey test failures that appear unrelated to your change.

PR #302, opened 2019-08-08
linkerd2-proxy: Adopt Rust 2018

olix0r:
Much of this repo was moved to Rust 2018 in 1d4eac..011c00. This
change updates the primary proxy project.

More information about Rust 2018 can be found at
https://doc.rust-lang.org/stable/edition-guide/rust-2018/edition-changes.html

While pursuing this change, I've tried to consolidate `use` statements into a
single block at the beginning of each module. This will help to make modules
more uniform.


src/app/config.rs
```diff
@@ -1,3 +1,12 @@
+use super::control::ControlAddr;
+use super::identity;
+use crate::addr;
```
hawkw: Can/should this be
```rust
use crate::{
    addr,
    convert::TryFrom,
    ... 
};
```

src/app/control.rs
```diff
@@ -142,23 +139,22 @@ pub mod add_origin {
...
     }
 }
 
 /// Resolves the controller's `addr` once before building a client.
 pub mod resolve {
-    use futures::{Future, Poll};
+    use super::{client, ControlAddr};
+    use crate::dns;
+    use crate::svc;
+    use crate::Addr;
```
hawkw: similarly, should this be
```rust
use crate::{dns, svc, Addr};
```

src/app/control.rs
```diff
@@ -272,7 +268,7 @@ pub mod resolve {
             let target = client::Target {
                 addr,
                 server_name: dst.identity.clone(),
-                log_ctx: ::logging::admin().client("control", dst.addr.clone()),
+                log_ctx: crate::logging::admin().client("control", dst.addr.clone()),
```
hawkw: IMO, this would look nicer if we imported the `logging` module here (and elsewhere), but TIOLI.

src/app/main.rs
```diff
@@ -1,6 +1,34 @@
...
+use super::config::{Config, H2Settings};
+use super::dst::DstAddr;
+use super::identity;
+use super::profiles::Client as ProfilesClient;
+use crate::app::classify::{self, Class};
+use crate::app::metric_labels::{ControlLabels, EndpointLabels, RouteLabels};
+use crate::app::tap::serve_tap;
+use crate::control;
+use crate::dns;
+use crate::drain;
```
hawkw: Again, seems like there can be some consolidation here?

src/app/config.rs
olix0r: it can be but, since this isn't something that is enforced by rustfmt, i don't particularly care one way or the other.

hawkw: fair; i only brought it up due to your comment about consolidating `use` statements, which i realise i kind of misread to imply that you tried to consolidate individual `use`s into blocks.

olix0r:
@kleimkuhler thanks, i had only tested compiling on linux. should work on macos now, as well.

PR #306, opened 2019-08-09
Split utilities into sub-crates

olix0r:
In preparation for breaking the large proxy library crate
into smaller components, this change moves several core
types and facilities into dedicated sub-crates:
* `linkerd2-dns-name` exposes the `Name` and `Suffix` types;
* `linkerd2-addr` exposes the core `Addr` and `NameAddr` types;
* `linkerd2-drain` exposes request-draining primitives used to notify server connections when shutdown is initiated.
* `linkerd2-signal` provides a facility for handling a shutdown signal from the OS, used only by the main binary.

There is no intended change in behavior.


lib/linkerd2-dns-name/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "linkerd2-dns-name"
```
hawkw: TIOLI: it looks like the `-addr` and `-dns-name` crates are small enough, unlikely to change, and related enough, that perhaps they ought to be combined?

hawkw:
looks like the Travis build of 42379670bfd27fa2863c5048815dd2b4e805dacf hung before it even started building linkerd, i'm gonna restart it

PR #307, opened 2019-08-09
Hangup destination resolutions when dropped

adleong:
When destinations idle out of the router cache, the corresponding destination lookup is not cancelled.  This can lead to a build-up of many subscriptions as entries idle out of the cache and are later re-added.  As these subscriptions accumulate, memory usage of the destination service grows, manifesting as a memory leak.

We use a hangup oneshot so that when the resolution is dropped, the daemon will be notified and will terminate.

### DIFF
--- a/src/control/destination/resolution.rs
+++ b/src/control/destination/resolution.rs
@@ -190,12 +190,12 @@ where
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         match self.updater.hangup.poll() {
             Ok(Async::Ready(never)) => match never {}, // unreachable!
-            Ok(Async::NotReady) => {},
+            Ok(Async::NotReady) => {}
             Err(_) => {
                 // Hangup tx has been dropped.
                 debug!("resolution cancelled");
                 return Ok(Async::Ready(()));
-            },
+            }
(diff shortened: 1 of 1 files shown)


src/control/destination/resolution.rs
```diff
@@ -175,7 +190,21 @@ where
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         loop {
             match self.query.poll() {
-                Ok(Async::NotReady) => return Ok(Async::NotReady),
+                Ok(Async::NotReady) => {
+                    match self.updater.hangup.poll() {
```
olix0r: Why is this only called in the unready state? Shouldn't this be called before polling the underlying client?

src/control/destination/resolution.rs
```diff
@@ -175,7 +190,21 @@ where
...
+                        Ok(Async::Ready(never)) => match never {}, // unreachable!
+                        Ok(Async::NotReady) => {
+                            // We are now scheduled to be notified if the hangup
+                            // tx is dropped.
```
olix0r: this comment is just documenting the futures contract, fwiw

PR #309, opened 2019-08-12
Split inbound and outbound proxy logic

olix0r:
Previously, `app::main` contained hundreds of lines of proxy stack
specification, making the file difficult to navigate.
    
The linkerd2-proxy-core crate is intended to expose the interfaces needed to implement linkerd2-proxy, so that implementation pieces may eventually be split into separate sub-crates.
    
Adds a common boxed `Error` type.
    
Adds `ListenAndSpawn` and `ServeConnection` traits. Each proxy (i.e., inbound or outbound) impements `ServeConnection`. The main is responsible for attaching the proxy server to a `ListenAndSpawn` handler.

These traits, above all else, allow modules to return an `impl Trait` for one of the above without having to name the type explicitly, which is especially cumbersome with layered service stacks.

While making this change, the proxy's `Connection` type was changed to include the remote address of the socket, as that data was missing and passed in separately via a tuple. In order to facilitate this, the `AddrInfo` API was extended to support `remote_addr`.


olix0r:
Setup has been split into:
* https://github.com/linkerd/linkerd2-proxy/pull/311
* https://github.com/linkerd/linkerd2-proxy/pull/312

PR #314, opened 2019-08-14
core: Formalize the listen/serve API

olix0r:
The `transport::tls::Listen` and `proxy::Server` types together formed a
sort of interface that facilitiates serving connections from tasks.

This change formalizes these interfaces so that the concrete
implementations can move/change more easily.

In pursuing this change, the remote `SocketAddr` is no longer passed
around in a tuple and is, instead, wired into the `Connection` type. The
`AddrInfo` trait has been updated to account for existince of the remote
socket addr on a socket.

No functional changes intended.

### DIFF
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -1,7 +1,7 @@
 use super::admin::{Admin, Readiness};
 use super::config::{Config, H2Settings};
 use super::profiles::Client as ProfilesClient;
-use super::{DispatchDeadline, dst::DstAddr, identity};
+use super::{dst::DstAddr, identity, DispatchDeadline};
 use crate::app::classify::{self, Class};
 use crate::app::handle_time;
 use crate::app::metric_labels::{ControlLabels, EndpointLabels, RouteLabels};
@@ -23,10 +23,10 @@ use crate::{
 use futures::{self, future, Future};
 use http;
 use hyper;
+use std::fmt;
 use std::net::SocketAddr;
 use std::thread;
 use std::time::{Duration, SystemTime};
-use std::fmt;
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::runtime::current_thread;
 use tracing::{debug, error, info, trace};
@@ -847,8 +847,7 @@ fn spawn_server<A, T, C, R, B, G>(
     router: R,
     h2_settings: H2Settings,
     drain_rx: drain::Watch,
-)
-where
+) where
     A: proxy::Accept<Connection> + Send + 'static,
     A::Io: transport::Peek + fmt::Debug + Send + 'static,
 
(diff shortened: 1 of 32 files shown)


olix0r:
https://github.com/linkerd/linkerd2-proxy/pull/316 illustrates the next change that uses these traits

olix0r:
https://github.com/linkerd/linkerd2-proxy/pull/317 follows this as well

src/app/main.rs
```diff
@@ -839,33 +824,30 @@ where
...
+            );
         };
-        task::spawn(inbound);
     }
 }
 
 type Error = Box<dyn std::error::Error + Send + Sync>;
 
-fn serve<A, T, C, R, B, G>(
+fn spawn_server<A, T, C, R, B, G>(
```
hawkw: What do you think about just getting rid of this function, and just constructing a `proxy::Server` and calling `proxy::spawn` with that server in  inbound/outbound? This doesn't seem to be doing much of anything anymore, and just creates another place where we have to have a big list of `where` bounds. It would be nice to reduce the number of places where we have to maintain a giant list of bounds, especially if those constraints are enforced elsewhere as well.

src/proxy/server.rs
```diff
@@ -190,164 +188,155 @@ impl fmt::Display for NoOriginalDst {
...
-        let connect = self.connect.clone();
-
-        if disable_protocol_detection {
-            trace!("protocol detection disabled for {:?}", orig_dst);
-            let fwd = tcp::forward(io, connect, source);
-            let fut = self.drain_signal.clone().watch(fwd, |_| {});
-            return log.future(Either::B(fut));
-        }
+        let accept_fut = if should_detect {
+            Either::A(
```
hawkw: Nit/TIOLI: IMO, this might read better as 
```rust
let accept_fut = if should_detect {
    let f = io.peek()
       .map_err(|e| debug!("peek error: {}", e))
       .map(|io| {
            ...
        });
    Either::A(f)
} else { 
   ... 
}
```
this way, I have to figure out slightly fewer grouping symbols.

Not a big deal though.

src/proxy/server.rs
```diff
@@ -190,164 +188,155 @@ impl fmt::Display for NoOriginalDst {
...
-                    trace!("detected HTTP/1");
-                    route
-                        .make_service(source)
-                        .map_err(|never| match never {})
-                        .and_then(move |s| {
+            Some(proto) => Either::B({
+                make_http
+                    .make_service(source)
+                    .map_err(|never| match never {})
+                    .and_then(move |http_svc| match proto {
```
hawkw: Nit/TIOLI: It would be nice if this was not nested so many indent levels deep...maybe this closure could factored out into a higher level `let` binding or something?

src/app/main.rs
olix0r: it's simplified substantially in the followup

src/proxy/server.rs
olix0r: I don't disagree. I'm going to have be back in this file soon and will make polish changes then

src/app/main.rs
hawkw: okay, cool; carry on then!

PR #315, opened 2019-08-15
travis: `make clean` after tests

olix0r:
In theory, travis's cache of build artifacts is useful for reducing CI
run times from ~30m to ~10m. However, in practice, the cache grows in
excess of 3GB after just a few runs. This frequently causes subsequent
runs to timeout while fetching the cache image, causing spurious
failures and, on master, requiring manual intervention to produce proxy
binaries.

CI needs to be predictable first and fast second; so, until we can
figure out a more robust caching scheme, I propose that we disable
caching from CI so that builds do not require manual intervention.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,11 +16,12 @@ jobs:
       install:
         - rustup component add rustfmt
         - export RUSTFLAGS="-C debuginfo=0" RUST_TEST_THREADS=1 RUST_TEST_PATIENCE_MS=200
+      # # Cache debugging:
+      # before_script:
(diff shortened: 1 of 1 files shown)


grampelberg:
fwiw, it probably isn't a travis thing, but buildkit has pretty great [compiler cache](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/experimental.md#run---mounttypecache) support now. I started to get it working in the web container as that'll really improve build times, just haven't solved some of the other buildkit issues.

olix0r:
@grampelberg once the linkerd2 repo is using prow, i'll be eager to move the proxy build over there (with buildkit, etc)

PR #318, opened 2019-08-15
Split linkerd2-proxy-resolve

olix0r:
`linkerd2-proxy::control` included the logic used to resolve names from
the control plane. Before this logic is altered to support IP-address
names, this change splits this module into a new subcrate with minimal
other changes.

No functional changes.


lib/linkerd2-proxy-resolve/src/destination/resolution.rs
```diff
@@ -165,10 +165,9 @@ where
                 .expect("resolution should not have been dropped");
 
             let query = self.query.take().expect("invalid state");
-            let ctx = logging::Section::Proxy.bg(LogCtx(query.authority().clone()));
 
-            let daemon = ctx.future(Daemon { updater, query });
-            tokio::spawn(Box::new(daemon));
+            // todo
```
adleong: todo what?

lib/linkerd2-proxy-resolve/src/destination/resolution.rs
```diff
@@ -165,10 +165,9 @@ where
                 .expect("resolution should not have been dropped");
 
             let query = self.query.take().expect("invalid state");
-            let ctx = logging::Section::Proxy.bg(LogCtx(query.authority().clone()));
 
-            let daemon = ctx.future(Daemon { updater, query });
-            tokio::spawn(Box::new(daemon));
+            // todo: set logging context within the daemon task.
+            task::spawn(Daemon { updater, query });
```
hawkw: This should be something like:
```rust
let daemon = Daemon { updater, query }
    .instrument(info_span!("resolve", authority = %query.authority())
task::spawn(daemon)
```
(assuming that the authority implements `fmt::Display`, which IIRC it does)

the imports to make this work will be something like
```rust
use tracing::info_span;
use tracing_futures::Instrument;
```
```

File: lib/linkerd2-proxy-resolve/src/destination/resolution.rs [hunk 3]
```diff
@@ -165,10 +166,11 @@ where
                 .expect("resolution should not have been dropped");
 
             let query = self.query.take().expect("invalid state");
-            let ctx = logging::Section::Proxy.bg(LogCtx(query.authority().clone()));
 
-            let daemon = ctx.future(Daemon { updater, query });
-            tokio::spawn(Box::new(daemon));
+            let authority = query.authority().clone();
+            let fut =
+                Daemon { updater, query }.instrument(info_span!("resolve", authority = %authority));
```
hawkw: Nit: this can actually just be
```suggestion
                Daemon { updater, query }.instrument(info_span!("resolve", %authority));
```
since `authority` is a variable in scope.

Also, `tracing` doesn't require the authority to be cloned, it can be recorded by reference

olix0r: but it's moved into `Daemon`, so it has to be

PR #319, opened 2019-08-17
classify: Assume success on missing grpc-status

olix0r:
When a request is canceled by a client, the server seems to just drop
the open response in a way that prevents the server code from being
notified of the reset reason, and so our classification code treats this
as if the stream ended without trailers.

This was previously treated as a failure, but this should be handled as
a success so that cancelations are not mistaken for application errors.

Addresses linkerd/linkerd2#3281
See also linkerd/linkerd2#3282


olix0r:
@adleong i don't think there's any situation where a client reset should be considered a server failure, but yes

adleong:
Ah, in the case of a server reset, we correctly get the reset code?

PR #320, opened 2019-08-17
ci: Only test libraries on PRs

olix0r:
Building the integration tests takes 2-3x the time as building tests for
the raw libs.

Split `make test-lib` and `make test-integration`, and change the Travis
config to only run `test-lib` on PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,14 @@ jobs:
       #   - du -sh . && find . -type f |xargs -n 1000 du -s |sort -rn |head |awk '{print $2}' |xargs du -sh
       script:
         - make check-fmt
-        - travis_wait 40 make test clean
+        - travis_wait 20 make test-lib clean
+
+    # On master, run the (much longer) set of integration tests.
+    - stage: publish
+      if: branch = master && type != pull_request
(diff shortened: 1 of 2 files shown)


olix0r:
```
:; make clean
cargo clean --target-dir target/debug

:; time make test-lib
cargo fetch --locked
cargo test --all --frozen  --lib --no-default-features
...
make test-lib  414.83s user 15.83s system 241% cpu 2:58.30 total

:; time make test-integration
cargo fetch --locked
cargo test --all --frozen  --tests --no-default-features
...
make test-integration  2902.36s user 26.63s system 667% cpu 7:18.50 total
```

hawkw:
I definitely agree that we need to fix the issue of 45-minute builds, but I'm a little concerned that in the past, broken `master` builds of PRs that have built successfully have gone unnoticed. I'm afraid that if we disable the integration tests on PRs, we might be more likely to merge broken builds and leave master in a broken state.

Have we looked into separating the integration tests into a separate crate that depends on the proxy crate? That might allow the proxy to be built once and cached, which could speed up the integration tests without having to disable them on PRs.

PR #321, opened 2019-08-17
[GSoC] [Draft PR] Added Kafka as a new traffic protocol

ruig2:
This PR is a draft and for discussion purpose only.

I added Kafka as a new traffic protocol and called the external lib kafka_codec to decode it. More details are given in the in-line comments.

Also, the header of Kafka **response** can be decoded.


Cargo.lock
```diff
@@ -438,6 +438,16 @@ name = "itoa"
 version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
+[[package]]
```
ruig2: It seems a few dependencies have newer versions and the `Cargo.lock` file is updated.

Cargo.toml
```diff
@@ -107,3 +108,5 @@ debug = false
 [patch.crates-io]
 webpki = { git = "https://github.com/seanmonstar/webpki", branch = "cert-dns-names" }
 
+[dependencies.kafka_codec]
+path = "../kafka_codec"
```
ruig2: This points to the codebase of https://github.com/ruig2/KafkaRustCodec/tree/rui/add_metadata_decode

src/proxy/server.rs
```diff
@@ -344,6 +344,13 @@ where
                                 .map(|_| ())
                                 .map_err(|e| trace!("http2 server error: {:?}", e))
                         })
+                })),
+                Protocol::Kafka => Either::B({
+                    trace!("detected Kafka");
+                    println!("detected Kafka");
+                    //let fwd = tcp::forward(io, connect, source);
+                    let fwd = tcp::forward(KafkaIo::new(io), connect, source);
```
ruig2: This is where I wrap the `io` as a `KafkaIo`

src/transport/io.rs
```diff
@@ -186,3 +186,101 @@ mod tests {
...
+where
+    I: AsyncRead + AsyncWrite,
+{
+    // Is it possible to read by frame? If so, then we need to handle the frame case
+    // for example, a Kafka message is splitted into two frames
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        let result = self.inner.read(buf);
+        match result {
+            Ok(n) => {
+                while self.start_pos_next_message <= n {
```
ruig2: `start_pos_next_message` is the state of the position of the next message to analyze.

PR #323, opened 2019-08-22
Increment Trace-Context spans

adleong:
This adds a `trace` layer which reads the `traceparent` HTTP header from the request.  If this header is absent, the request is fowarded unmodified.  If the header is present, a new span will be started in the current trace by creating a new random span id setting it into the `traceparent` header before forwarding the request.  If the sampled bit of the header was set, we emit metadata about the span to the returned channel when the span is complete, i.e. when we receive the response.

This layer is added to both the inbound and outbound proxies.  The span receivers are currently unused and emitted spans are simply dropped for now.  

Eventually, a gRPC client will be used to send the emitted spans to the OpenCensus collector.  

Also TODO is to add support for the `grpc-trace-bin` header in addition to the `traceparent` header.

### DIFF
--- a/src/app/inbound/mod.rs
+++ b/src/app/inbound/mod.rs
@@ -1,7 +1,7 @@
 use super::{classify, config::Config, dst::DstAddr, identity, DispatchDeadline};
 use crate::proxy::http::{
     client, insert, metrics as http_metrics, normalize_uri, profiles, router, settings,
-    strip_header,
+    strip_header, trace,
 };
 use crate::proxy::{accept, reconnect, Server};
 use crate::transport::{self, connect, keepalive, tls, Connection};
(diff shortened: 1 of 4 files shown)


src/app/inbound/mod.rs
```diff
@@ -54,9 +54,12 @@ where
         .layer(tls::client::layer(local_identity))
         .service(connect::svc());
 
+    let (_span_rx, trace_layer) = trace::layer(100);
```
olix0r: This looks generally good, but...

istm that you're going to want to pass the inbound/outbound server functions some interface that can receive traces, rather than returning an mpsc::Receiver from each.

As a very rough idea:

```rust
pub trait SpanSink {
  type Error: Into<Error>;

  fn poll_ready(&mut self) -> Poll<(), Self::Error>;

  fn send_span(&mut self, span: Span) -> Result<(), Self::Error>;
}
```

adleong:
Thanks for the feedback, @hawkw and @olix0r.  I will open a new PR taking this feedback into account.

PR #330, opened 2019-08-31
Stop using a Builder in the profile router

olix0r:
The builder type should only be used when composing layers. Now that we
have `Builder::into_inner`, we need not store a builder in the profile
router.

No functional changes.


hawkw:
(looks like you need to run rustfmt)

PR #331, opened 2019-08-31
Improve stack-related compiler error messages

olix0r:
Service stacks were built from "top" to bottom, so that the concrete
service type is satisied after all middlewares are composed toegether.
This leads to exceedingly complex compiler error messages.

This change adds a `svc::Stack` builder that aids composing layers
around a concrete inner service so that the compiler can more clearly
identity the layer that violates the type contract.

Note: this does *not* appear to improve compile times.

No functional changes are intended.

### DIFF
--- a/src/app/inbound/mod.rs
+++ b/src/app/inbound/mod.rs
@@ -84,12 +84,12 @@ where
     // extension into each request so that all lower metrics
     // implementations can use the route-specific configuration.
     let dst_route_layer = svc::layers()
-        .buffer_pending(max_in_flight, DispatchDeadline::extract)
-        .layer(classify::layer())
-        .layer(http_metrics::layer::<_, classify::Response>(
+        .and_then_buffer_pending(max_in_flight, DispatchDeadline::extract)
+        .and_then(classify::layer())
+        .and_then(http_metrics::layer::<_, classify::Response>(
             route_http_metrics,
         ))
-        .layer(insert::target::layer())
+        .and_then(insert::target::layer())
         .into_inner();
 
     // A per-`DstAddr` stack that does the following:
@@ -188,8 +188,8 @@ where
     // As the inbound proxy accepts connections, we don't do any
     // special transport-level handling.
     let accept = accept::builder()
-        .layer(transport_metrics.accept("inbound"))
-        .layer(keepalive::accept::layer(config.inbound_accept_keepalive));
+        .and_then(transport_metrics.accept("inbound"))
+        .and_then(keepalive::accept::layer(config.inbound_accept_keepalive));
 
     Server::new(
         "out",
--- a/src/app/inbound/mod.rs
+++ b/src/app/inbound/mod.rs
@@ -175,6 +175,9 @@ where
     let source_stack = svc::stack(svc::shared(admission_control))
         .push(orig_proto_downgrade::layer())
         .push(insert::target::layer())
+        // disabled due to information leagkage
+        //.push(set_remote_ip_on_req::layer())
+        //.push(set_client_id_on_req::layer())
         .push(strip_header::request::layer(super::L5D_REMOTE_IP))
         .push(strip_header::request::layer(super::L5D_CLIENT_ID))
         .push(strip_header::response::layer(super::L5D_SERVER_ID))
--- a/src/app/inbound/mod.rs
+++ b/src/app/inbound/mod.rs
@@ -54,7 +54,7 @@ where
         .push(rewrite_loopback_addr::layer());
 
     // Instantiates an HTTP client for a `client::Config`
-    let client_stack = svc::stack(connect.clone())
+    let client_stack = connect.clone()
         .push(client::layer("in", config.h2_settings))
         .push(reconnect::layer().with_backoff(config.inbound_connect_backoff.clone()))
         .push(normalize_uri::layer());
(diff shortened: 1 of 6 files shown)


olix0r:
@hawkw i'd encourage tower to mirror these changes. from the linkerd perspective, i'm not particularly bothered by maintaining our own builder types, especially if we continue ensure they are only used narrowly (i.e. code doesn't depend on concrete builder types).

hawkw:
> @hawkw i'd encourage tower to mirror these changes.

👍 @LucioFranco may be interested in this!

src/app/inbound/mod.rs
```diff
@@ -159,44 +157,38 @@ where
...
+        .push(strip_header::response::layer(super::L5D_SERVER_ID))
+        .push(insert::layer(move || {
             DispatchDeadline::after(dispatch_timeout)
         }))
-        .layer(strip_header::response::layer(super::L5D_SERVER_ID))
-        .layer(strip_header::request::layer(super::L5D_CLIENT_ID))
-        .layer(strip_header::request::layer(super::L5D_REMOTE_IP))
-        .layer(insert::target::layer())
-        .layer(orig_proto_downgrade::layer())
-        // disabled on purpose
```
kleimkuhler: Looks like these were removed here, but kept in `outbound.rs`. Can we either keep or remove in both files?

src/svc.rs
```diff
@@ -12,61 +11,120 @@ pub use tower::{service_fn as mk, MakeConnection, MakeService, Service, ServiceE
...
+    where
+        D: buffer::Deadline<Req>,
+        Req: Send + 'static,
+    {
+        self.push_pending().push(buffer::layer(bound, d))
     }
 
-    pub fn load_shed(self) -> Builder<Stack<LoadShedLayer, L>> {
-        Builder(self.0.load_shed())
+    // pub fn push_spawn_ready(self) -> Stack<tower_spawn_ready::MakeSpawnReady<S>> {
```
kleimkuhler: Can this be removed?

src/svc.rs
```diff
@@ -12,61 +11,120 @@ pub use tower::{service_fn as mk, MakeConnection, MakeService, Service, ServiceE
...
-    pub fn into_inner(self) -> L {
-        self.0.into_inner()
+    pub fn push_timeout(self, timeout: Duration) -> Stack<tower::timeout::Timeout<S>> {
+        self.push(TimeoutLayer::new(timeout))
     }
 
-    /// Wrap the service `S` with the layers.
-    pub fn service<S>(self, service: S) -> L::Service
+    /// Validates that this stack serves T-typed targets.
+    pub fn serves<T>(self) -> Self
```
kleimkuhler: What was the reason for adding this? It doesn't look like it is used anywhere.

LucioFranco:
@olix0r by chance, do you have an example of error messages and how they differ?

src/svc.rs
```diff
@@ -12,61 +11,120 @@ pub use tower::{service_fn as mk, MakeConnection, MakeService, Service, ServiceE
...
-        Builder(self.0.layer(l))
+impl<L> Layers<L> {
+    pub fn push<O>(self, outer: O) -> Layers<Pair<L, O>> {
+        Layers(Pair::new(self.0, outer))
     }
 
     /// Buffer requests when when the next layer is out of capacity.
-    pub fn pending(self) -> Builder<Stack<pending::Layer, L>> {
-        self.layer(pending::layer())
+    pub fn push_pending(self) -> Layers<Pair<L, pending::Layer>> {
```
kleimkuhler: This doesn't need to be `pub` right now. It was there before, but just noticing now.

src/svc.rs
```diff
@@ -12,61 +11,120 @@ pub use tower::{service_fn as mk, MakeConnection, MakeService, Service, ServiceE
...
+
+impl<S> Stack<S> {
+    pub fn push<L: Layer<S>>(self, layer: L) -> Stack<L::Service> {
+        Stack(layer.layer(self.0))
     }
 
     /// Buffer requests when when the next layer is out of capacity.
-    pub fn spawn_ready(self) -> Builder<Stack<SpawnReadyLayer, L>> {
-        self.layer(SpawnReadyLayer::new())
+    pub fn push_pending(self) -> Stack<pending::MakePending<S>> {
```
kleimkuhler: This also doesn't need to be `pub`.

src/svc.rs
olix0r: As the comment says, this is useful for validating the state of a stack -- i.e. that it actually implements a service for a given type. These assertions aren't necessarily made by the layer implementations, so this is useful during development... We allow(dead_code) here, more generally, because this stuff is here to be useful as you're moving the stack around. It's okay if there are things that aren't actively used...

olix0r:
@LucioFranco sorry, it's pretty annoying (time consuming) to generate these errors, so I'll have to leave that as an exercise to others. The problem is actually pretty easy to understand, though: If I construct a stack as `layerA.layer(layerB).layer(layerC).layer(layerD).service(svc)`, the type of this stack isn't bound until the `service()` call, and so the user has to trawl through a bunch of compiler errors to figure out which layer violated the contract.  If we invert this to be `svc.push(layerD).push(layerC).push(...)`, then we _always_ know the concrete type of the stack at each step, and I can make additional type assertions in the middle. So if there's an error, it should happen at the proper layer...

src/svc.rs
kleimkuhler: Cool makes sense--thanks!

LucioFranco:
@olix0r I'm onboard with that! I remember the discussion we had a while back. I do think with tower 0.1 we should keep it how it is now. Does linkerd have any plans around updating to futures 0.3? I guess there should be some talk around that...in  relation to tower.

PR #333, opened 2019-09-02
Split service discovery into layered components

olix0r:


src/svc.rs
```diff
@@ -103,6 +104,14 @@ impl<S> Stack<S> {
         self.push(TimeoutLayer::new(timeout))
     }
 
+    /// Validates that this stack serves T-typed targets.
+    pub fn makes<T>(self) -> Self
```
adleong: unused in this PR?

src/proxy/http/profiles/router.rs
```diff
@@ -133,24 +133,23 @@ where
     }
 }
 
-impl<G, Inner, RouteLayer, RouteBody, InnerBody, Target, InnerSvc, RouteMake, RouteSvc>
-    svc::Service<Target> for MakeSvc<G, Inner, RouteLayer, RouteBody, InnerBody>
+impl<G, Inner, RouteLayer, RouteBody, InnerBody, Target, RouteSvc> svc::Service<Target>
```
adleong: Nice cleanups in this file, but I don't see how it's related to the rest of the PR

src/svc.rs
olix0r: yeah, but it's useful. this module is for useful things when working with stacks.

src/proxy/http/profiles/router.rs
olix0r: i needed to make some changes below and was fighting type signatures to make it all happy... i can probably back some of this out...

olix0r:
Replaced by #341

PR #337, opened 2019-09-03
Improve stack-related compiler error messages

olix0r:
Service stacks were built from "top" to bottom, so that the concrete
service type is satisfied after all middlewares are composed together.
This leads to exceedingly complex compiler error messages.

This change adds a svc::Stack builder that aids composing layers
around a concrete inner service so that the compiler can more clearly
identity the layer that violates the type contract.

Note: this does not appear to improve compile times.

No functional changes are intended.


olix0r:
Resumission of #331, but against master instead of a development branch

PR #339, opened 2019-09-04
Add trace context crate

adleong:
Add a crate for reading the trace context from a request, incrementing the span id, and potentially emitting the span to a channel.  

Two trace propagation formats are supported:
* grpc-trace-bin: a trace propagation format used for gRPC requests where the trace context is stored in a base64 encoded binary format in the grpc-trace-bin header
* x-b3: the zipkin trace propagation format where the trace context is stored across the x-b3-trace-id, x-b3-spanid, and x-b3-sampled headers.


lib/linkerd2-trace-context/src/layer.rs
```diff
@@ -0,0 +1,184 @@
...
+{
+    type Item = Fut::Item;
+    type Error = Fut::Error;
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        let inner = try_ready!(self.inner.poll());
+        let mut span = self.span.take().expect("polled after ready");
+        span.end = SystemTime::now();
+        trace!(message = "emitting span", ?span);
+        self.sink.try_send(span).unwrap_or_else(|_| {
```
olix0r: also, nit, I find `if self.sin.try_send(span).is_err()` to be clearer than unwrap_or_else

lib/linkerd2-trace-context/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+pub struct Span {
+    pub trace_id: Id,
+    pub span_id: Id,
+    pub parent_id: Id,
+    pub span_name: String,
+    pub start: SystemTime,
+    pub end: SystemTime,
+}
+
+pub trait SpanSink {
```
olix0r: Consider using `stream::Sink<Span>` instead? Or if it's really necessary, to have a dedicate crate, provide a blanket impl for all Sinks... Note the impl of Sink for mpsc::Sender (https://github.com/tokio-rs/tokio/blob/master/tokio-sync/src/mpsc/bounded.rs#L255)

lib/linkerd2-trace-context/src/layer.rs
```diff
@@ -0,0 +1,184 @@
...
+                    // End time will be updated when the span completes.
+                    end: SystemTime::now(),
+                };
+                let span_fut = SpanFuture {
+                    span: Some(span),
+                    inner: f,
+                    sink: self.sink.clone(),
+                };
+                return futures::future::Either::B(span_fut);
+            }
```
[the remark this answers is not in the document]
adleong: > This all seems a bit tortured to me... 
This is how I always feel

I'm not sure I understand your suggestion, though.  We can't send the span in `call` because we don't know the span end time.  In a subsequent PR I'll be adding some metadata from the response to the span as well, which is not available until the future completes.

lib/linkerd2-trace-context/src/layer.rs
```diff
@@ -0,0 +1,184 @@
...
+            if context.is_sampled() {
+                path = Some(request.uri().path().to_string());
+            }
+        }
+
+        let f = self.inner.call(request);
+
+        if let Some(propagation::TraceContext {
+            propagation: _,
+            version: _,
```
olix0r: these are unused? you can elide these and include, `..` at the end (to account for any/all unknown fields)

lib/linkerd2-trace-context/src/layer.rs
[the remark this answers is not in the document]
olix0r: Ah, OK. Makes sense

I would probably personally opt to have a ResponseFuture type that hides the messy innards (basically, holding an `Option<SendSpan>`, but this is okay too.

But, please for the love of all things good import `Either` so we don't have to look at the `futures::future::` every time

lib/linkerd2-trace-context/src/lib.rs
adleong: I considered the existing Sink trait, but it's API is more complex when really all I want is the semantics of mpsc::Sender.

lib/linkerd2-trace-context/src/layer.rs
```diff
@@ -0,0 +1,183 @@
...
+        let mut trace_context = propagation::unpack_trace_context(&request);
+        let mut path = None;
+
+        if let Some(ref mut context) = trace_context {
+            trace!(message = "got trace context", ?context);
+            propagation::increment_span_id(&mut request, context);
+            // If we plan to sample this span, we need to copy the path from
+            // the request before dispatching it to inner.
+            if context.is_sampled() {
+                path = Some(request.uri().path().to_string());
```
kleimkuhler: @seanmonstar's comment [here](https://github.com/linkerd/linkerd2-proxy/pull/348#discussion_r322905732) may also be applicable here for sampling.

lib/linkerd2-trace-context/src/layer.rs
```diff
@@ -0,0 +1,183 @@
...
+
+impl<Fut, Sink> Future for SpanFuture<Fut, Sink>
+where
+    Fut: Future,
+    Sink: SpanSink,
+{
+    type Item = Fut::Item;
+    type Error = Fut::Error;
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
```
kleimkuhler: This entire layer has been really clear so far to read through!

lib/linkerd2-trace-context/src/propagation.rs
```diff
@@ -0,0 +1,248 @@
...
+}
+
+fn get_header_str<'a, B>(request: &'a http::Request<B>, header: &str) -> Option<&'a str> {
+    let hv = request.headers().get(header)?;
+    hv.to_str()
+        .map_err(|e| warn!("Invalid trace header {}: {}", header, e))
+        .ok()
+}
+
+fn parse_header_id<B>(request: &http::Request<B>, header: &str) -> Option<Id> {
```
kleimkuhler: Maybe something to just keep in mind, but it looks like there is some overlap with this and header retrieval functions in `src/proxy/http/mod.rs`. I don't think we need to consolidate right now, but maybe leave a comment about doing it in the future--maybe when we move some of that into `lib/` as well?

lib/linkerd2-trace-context/src/layer.rs
adleong: thanks!

lib/linkerd2-trace-context/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+// === impl Id ===
+
+impl Id {
+    fn new(len: usize) -> Self {
+        let mut rng = rand::thread_rng();
+        let mut bytes = vec![0; len];
+        rng.fill(bytes.as_mut_slice());
```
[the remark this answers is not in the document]
adleong: We use the ID type for both span ids (8 byte) and trace ids (16 byte).  However, we only ever generate span ids (as opposed to trace ids which are read from headers).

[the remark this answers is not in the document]
olix0r: perhaps `new_span_id` then? (with a const LEN internally)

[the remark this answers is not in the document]
olix0r: perhaps `new_span_id` (with a private const len), then?

lib/linkerd2-trace-context/src/layer.rs
```diff
@@ -0,0 +1,186 @@
...
+            if flags.is_sampled() {
+                trace!(message = "span will be sampled", ?span_id);
+                let span = Span {
+                    trace_id,
+                    span_id,
+                    parent_id,
+                    span_name: path.unwrap_or_else(String::new),
+                    start: SystemTime::now(),
+                    // End time will be updated when the span completes.
+                    end: SystemTime::now(),
```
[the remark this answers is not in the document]
adleong: Good call.  This actually allowed me to simplify some stuff as well.

lib/linkerd2-trace-context/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+
+impl fmt::Display for Flags {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:02x?}", self.0)
+    }
+}
+
+impl From<Bytes> for Flags {
+    fn from(buf: Bytes) -> Self {
+        Flags(buf[0])
```
olix0r: what's going on here? Is buf guaranteed to be non-empty? should this be a TryFrom instead?

lib/linkerd2-trace-context/src/layer.rs
```diff
@@ -0,0 +1,178 @@
...
+                });
+            }
+        }
+
+        let f = self.inner.call(request);
+
+        if let Some(span) = span {
+            let span_fut = SpanFuture {
```
olix0r: Again, I'd be reallly inclined to just return a SpanFuture here (always) and make the sink optional, and then handle the optionality in the SpanFuture impl (rather than the either).

But this is just cosmetic

hawkw: My guess is that always returning the same type, rather than using `Either`, may also reduce the impact on build times a bit (since `Either` results in a much longer type), but it's probably a drop in the bucket compared to other things that are making our builds long...

lib/linkerd2-trace-context/src/layer.rs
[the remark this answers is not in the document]
hawkw: Nit: If we're going to create spans with an `end` time when they haven't actually ended, we should probably try to avoid calling `SystemTime::now` twice --- this is going to result in a second unnecessary syscall. What about calling `SystemTime::now` once and then cloning that `SystemTime` for the placeholder "end" time? Or, we could use `SystemTime::UNIX_EPOCH` for the placeholder value?

lib/linkerd2-trace-context/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+
+pub mod layer;
+mod propagation;
+
+pub use layer::layer;
+
+const SPAN_ID_LEN: usize = 8;
+
+#[derive(Debug, Default)]
+pub struct Id(Vec<u8>);
```
hawkw: Correct me if I'm wrong, but this is only ever either an 8-byte span ID or a 16-byte trace ID, right? Why not have separate types for these, like
```rust
#[derive(Debug, Default)]
pub struct SpanId([u8; 8]);

#[derive(Debug, Default)]
pub struct TraceId([u8; 16]);
```
That way, we wouldn't have to allocate a `Vec` on the heap to represent each ID.

Alternatively, we could do
```rust
#[derive(Debug, Default)]
pub struct SpanId(u64);

#[derive(Debug, Default)]
pub struct TraceId(u128);
```
if we don't want to index the individual bytes.

lib/linkerd2-trace-context/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+impl Into<Vec<u8>> for Id {
+    fn into(self) -> Vec<u8> {
+        self.0
+    }
+}
+
+impl AsRef<Vec<u8>> for Id {
```
hawkw: Nit: I would probably make this `AsRef<[u8]>`, since `Vec` implements `AsRef<[u8]>` and most methods that one might need to pass a `&Vec<u8>` to will probably take a `&[u8]`.

lib/linkerd2-trace-context/src/propagation.rs
```diff
@@ -0,0 +1,248 @@
...
+#[derive(Debug)]
+pub enum Propagation {
+    Http,
+    Grpc,
+}
+
+#[derive(Debug)]
+pub struct TraceContext {
+    pub propagation: Propagation,
+    pub version: Id,
```
hawkw: As far as I can tell, a `version` is always a single byte, right? In that case, IMO, it doesn't really make sense for it to be an `Id`, which is internally a `Vec`. Why not define a newtype like
```rust
#[derive(Default, Clone, ...)]
struct Version(u8);
```
instead? This would save us an allocation, and (if you take my advice above about separating the `Id` type into `TraceId`s and `SpanId`s) let us enforce those known length constraints at the type level rather than at runtime...

lib/linkerd2-trace-context/src/layer.rs
adleong: This is much more clear to me, personally.  But I can change it if you want.

olix0r: here's my rationalization: as a user of this service/layer, i don't benefit at all from knowing about the eitherness of the response future -- it's all implementation detail. and it's not actually even two different potential futures! there's one (inner) future, and this layer decorates responses with its own logic (optionally doing some things as the future is processed). finally, its idiomatic (at least in our libs and tower) to have a `ResponseFuture` type for each layer (at least when it wraps its inner response future).

lib/linkerd2-trace-context/src/lib.rs
olix0r: I really like these suggestions! I don't want them to block merging this, though. Would you mind opening a p0 in the proxy backlog to remove the Vecs from the trace-context code, @hawkw ?

adleong: We use Vec throughout for Ids because `hex::decode` returns a Vec and the Span exporter gRPC client requires Vecs.  We could convert these to arrays in the meantime, but I don't think it saves on allocations.

hawkw: @adleong oh, that makes sense, never mind...there's been some talk in the past about patching `prost` to allow `Bytes` to be used as protobuf `bytes` fields; if we did that, we could probably just split the bytes we're parsing from into these types...

lib/linkerd2-trace-context/src/propagation.rs
adleong: Version was unused so I just removed it.

lib/linkerd2-trace-context/src/lib.rs
olix0r: the version suggestion seems like a fine followup, though probably not high priority

lib/linkerd2-trace-context/src/propagation.rs
```diff
@@ -85,11 +84,10 @@ fn unpack_grpc_trace_context<B>(request: &http::Request<B>) -> Option<TraceConte
 fn parse_grpc_trace_context_fields(buf: &mut Bytes) -> Option<TraceContext> {
     trace!(message = "reading binary trace context", ?buf);
 
-    let version = try_split_to(buf, 1).ok()?;
+    let _version = try_split_to(buf, 1).ok()?;
```
hawkw: Nit: do we not need to propagate the trace context version? It would be helpful to have a comment here explaining why we just ignore this...

lib/linkerd2-trace-context/src/layer.rs
```diff
@@ -0,0 +1,183 @@
...
+                span: Some(span),
+                inner: f,
+                sink: Some(self.sink.clone()),
+            }
+        } else {
+            ResponseFuture {
+                span: None,
+                inner: f,
+                sink: None,
+            }
```
olix0r: i think i agree with eliza's suggestion to store sink/span together... but if not that, this can be simplified as:

```rust
let sink = span.as_ref().map(|_| self.sink.clone());
let inner = self.inner.call(req);
Response Future { inner, span, sink }
```

PR #341, opened 2019-09-04
Split service discovery into composable components

olix0r:
The `linkerd2_proxy_resolve` crate, which was split out of the main
application in #318, contains a monilithic `Resolve` implementation.
Over time, this code has evolved in ways that have made it difficult to
modify and extend, especially and functionality cannot be tested
independently.

This change reorganizes the proxy's service discovery logic into a few
composable components. The `Resolve` API has been updated to facilitate
this.

Service discovery consists of two primary traits: `Resolve` and
`Discover`. `Resolve` is used within the proxy to model discovery
sources like the Destination API and DNS; and `Discover` updates a load
balancer with changes to its replica pool.

The former `Resolve` trait has been replaced with a trait alias for
`Service`, which allows for backpressure, etc. Furthermore, the API has
been updated so that it is easier to gracefully recover after a
resolution stream is lost (by changing `Update` to contain a `Vec`).

Now that `Resolve` is implemented as a `Service`, it becomes
straightforward to model service discovery in terms of a stack of
services, as we do throughout the project:

* The `linkerd2-proxy-api-resolve` crate provides an implementation of
  `Resolve` backed by the destination API. It does NOT handle
  reconnects, buffering/spawning, etc.
* The `linkerd2-request-filter` crate provides a simple
  middleware that can fail requests based on the request. This is used to
  implement the destination client's name/suffix restrictions.
* The `linkerd2-proxy-resolve` crate provides `Resolve`-middlewares that
  implement error recovery/backoff and endpoint-type-mapping.
* The `linkerd2-proxy-discover` crate provides a suite of middlewares
  that facilitate powering a balancer with a `Resolve`. Specifically:
  * `buffer` ensures that discovery updates are processed even when the
    balancer is not polling for updates.
  * `from_resolve` wraps a `Resolve` to produce `Discover` response.
  * `make_endpoint` wraps a `Discover`-producing service to build each
    inserted endpoint into a Service.

Almost all of this logic was happening in a single module, coupled to
our gRPC interface. Now, these components may be easily repurposed (for
instance, in order to back discovery by DNS).


olix0r:
In order to review this I recommend the following:
* Ignore the diff
* Review lib code as individual components
* Review new state of src/app
* If you're really really really brave, try to understand the prior discovery code for comparison.

In general, looking for places where this seem especially underdocumented (i.e. you really can't figure out what i'm trying to do or what a contract should be).

lib/linkerd2-proxy-core/src/recover.rs
```diff
@@ -0,0 +1,18 @@
+use super::Error;
+use futures::Stream;
+
+/// An error recovery strategy.
+pub trait Recover<E: Into<Error> = Error> {
```
hawkw: This is perhaps overly bikesheddy, but: why `Recover` rather than `Retry`? IMO, the name `Recover` implies a broader range of error recovery behavior, like fallbacks, as well as the backoff-and-retry behavior this trait implements.

I wonder if there's a way to come up with a more generic error-recovery trait that can encompass both this and the fallback layer? Not a blocker, just a thought.

lib/linkerd2-proxy-discover/src/buffer.rs
```diff
@@ -0,0 +1,161 @@
...
+    type Error = M::Error;
+    type Future = DiscoverFuture<T, M::Future, M::Response>;
+
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
+        self.inner.poll_ready()
+    }
+
+    fn call(&mut self, target: T) -> Self::Future {
+        let future = self.inner.call(target.clone());
+        Self::Future {
```
hawkw: :o TIL that you can do this!

lib/linkerd2-proxy-discover/src/buffer.rs
```diff
@@ -0,0 +1,161 @@
...
+            }
+
+            try_ready!(self
+                .tx
+                .poll_ready()
+                .map_err(|_| tracing::trace!("lost sender")));
+
+            let up = try_ready!(self.discover.poll().map_err(|e| {
+                let e: Error = e.into();
+                tracing::debug!("resoution lost: {}", e);
```
hawkw: Typo:
```suggestion
                tracing::debug!("resolution lost: {}", e);
```

hawkw: Also, nit: prefer structured fields:
```suggestion
                tracing::debug!(message = "resoution lost", error = %e);
```

lib/linkerd2-proxy-discover/src/buffer.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+pub struct Daemon<D: discover::Discover> {
+    discover: D,
+    disconnect_rx: oneshot::Receiver<Never>,
+    tx: mpsc::Sender<discover::Change<D::Key, D::Service>>,
+}
+
+#[derive(Clone, Debug)]
+pub struct Lost(());
```
hawkw: It's a little confusing to me that we have both this error and the `ResolutionLost` error in `proxy-core`. If I were downcasting errors upstream of this, I think it might be unclear what type to check for. What about adding a `new` to `ResolutionLost` and re-using it here? Or, if these errors have different meanings, a comment and/or name change could be good.

lib/linkerd2-proxy-discover/src/from_resolve.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+impl<R: Resolution> tower::discover::Discover for Discover<R> {
+    type Key = SocketAddr;
+    type Service = R::Endpoint;
+    type Error = R::Error;
+
+    fn poll(&mut self) -> Poll<Change<Self::Key, Self::Service>, Self::Error> {
+        loop {
+            if let Some(change) = self.pending.pop() {
```
hawkw: I notice that `pending` is essentially being treated as a stack here. Is that actually the correct behavior? 

If I understand correctly, it means that if we buffer a bunch of changes in `pending` and then get polled a bunch of times, we'll see the most recent changes _before_ earlier changes. If the `Discover` is then polled enough times to drain `pending`, then the most recently observed state will be the _oldest_ change.

Should we use a queue/dequeue rather than a stack here?

lib/linkerd2-proxy-resolve/src/map_endpoint.rs
```diff
@@ -0,0 +1,134 @@
...
+where
+    T: Clone,
+    R: resolve::Resolve<T>,
+    M: MapEndpoint<T, R::Endpoint> + Clone,
+{
+    type Response = Resolution<T, M, R::Resolution>;
+    type Error = R::Error;
+    type Future = ResolveFuture<T, R::Future, M>;
+
+    fn poll_ready(&mut self) -> Poll<(), Self::Error> {
```
hawkw: Should this be `#[inline]` to avoid pushing a stack frame for the outer `poll_ready`? IIRC rust won't inline functions across crate boundaries unless they're marked with the `#[inline]` attribute...

lib/linkerd2-proxy-resolve/src/recover.rs
```diff
@@ -0,0 +1,351 @@
...
+    T: fmt::Display + Clone,
+    R: resolve::Resolve<T>,
+    R::Endpoint: Clone + PartialEq,
+    E: Recover,
+{
+    fn poll_connected(&mut self) -> Poll<(), Error> {
+        loop {
+            self.state = match self.state {
+                State::Disconnected { ref mut backoff } => {
+                    tracing::trace!(message = "connecting", target = %self.target);
```
hawkw: Nit: IMO, it would be better for the `Resolution` type to hold a tracing `Span` with the target that we `enter` every time `Resolution::poll` is called. That way, we wouldn't have to manually include the target in these events, and any events emitted by methods we call would _also_ be in that span.

Also, eventually, it would allow us to do time profiling for resolutions...

lib/linkerd2-request-filter/src/lib.rs
```diff
@@ -0,0 +1,77 @@
+//! A `Service` middleware that applies arbirary-user provided logic to each target before f
```
hawkw: typo:
```suggestion
//! A `Service` middleware that applies arbitrary-user provided logic to each target before f
```

hawkw: Also, it looks like you started writing this comment and never finished it.

lib/linkerd2-request-filter/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "linkerd2-request-filter"
```
hawkw: AFAICT, the difference between this and [`tower-filter`](https://github.com/tower-rs/tower/blob/master/tower-filter/src/lib.rs) is that the predicate here is synchronous rather than a future?

It seems like eventually, this would be good to upstream into `tower-filter`. Not a blocker for now.

src/app/outbound/mod.rs
```diff
@@ -147,15 +148,15 @@ where
     // over all endpoints returned from the destination service.
     let balancer_layer = svc::layers()
         .push_spawn_ready()
-        .push(resolve::layer(discovery::Resolve::new(resolve)))
+        .push(discover::Layer::new(2, resolve))
```
hawkw: nit: reading this, the meaning of the `2` is not super obvious at all --- WDYT about making it a constant, or adding a builder-type API to `discover::Layer` so it's clearer what 2 means?

src/app/outbound/mod.rs
```diff
@@ -147,15 +148,15 @@ where
...
         .push(balance::layer(EWMA_DEFAULT_RTT, EWMA_DECAY));
 
     let distributor = endpoint_stack
         .push(
             // Attempt to build a balancer. If the service is
             // unresolvable, fall back to using a router that dispatches
             // request to the application-selected original destination.
-            fallback::layer(balancer_layer, orig_dst_router_layer).on_error::<Unresolvable>(),
+            fallback::layer(balancer_layer, orig_dst_router_layer), // TODO .on_error::<Unresolvable>(),
```
hawkw: i'm guessing this was supposed to be un-commented?

src/app/main.rs
```diff
@@ -402,7 +400,11 @@ where
             &config,
             local_identity.clone(),
             outbound_listener.local_addr(),
-            resolver,
+            outbound::resolve(
+                config.destination_get_suffixes.clone(),
+                outbound::ExponentialBackoff::default(),
```
hawkw: nit: is `outbound::resolve` ever called with a value for this that isn't the default? if not, maybe it should just always use default?

src/app/outbound/endpoint.rs
```diff
@@ -19,6 +21,9 @@ pub struct Endpoint {
     pub http_settings: settings::Settings,
 }
 
+#[derive(Copy, Clone, Debug)]
+pub struct FromMetadata;
```
hawkw: Nit/TIOLI: If there was an implementation of `MapEndpoint<A, B>` for `F where F: Fn(&A, SocketAddr, B) -> Out`, I think this could just be a function, and we could reference it as a function pointer instead of an empty struct?

lib/linkerd2-proxy-api-resolve/src/pb.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+fn to_id(pb: TlsIdentity) -> Option<identity::Name> {
+    use crate::api::destination::tls_identity::Strategy;
+
+    let Strategy::DnsLikeIdentity(i) = pb.strategy?;
+    match identity::Name::from_hostname(i.name.as_bytes()) {
+        Ok(i) => Some(i),
+        Err(_) => {
+            tracing::warn!("Ignoring invalid identity: {}", i.name);
```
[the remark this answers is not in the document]
olix0r: This is just moved code, going to defer all of this until later.

lib/linkerd2-proxy-api-resolve/src/resolve.rs
```diff
@@ -0,0 +1,144 @@
...
+use crate::core::resolve::Update;
+use crate::metadata::Metadata;
+use crate::pb;
+use futures::{future, Async, Future, Poll, Stream};
+use tower::Service;
+use tower_grpc::{self as grpc, generic::client::GrpcService, Body, BoxBody};
+use tracing::{debug, trace};
+
+#[derive(Clone)]
+pub struct Resolve<S> {
```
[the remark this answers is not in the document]
olix0r: i find that naming scheme to be pretty frustrating in practice.. how do you do that with Resolution, for instance?

src/app/outbound/mod.rs
olix0r: yeah, we don't need that anymore. any error that hits this should cause fallback.

src/app/main.rs
olix0r: it should probably be wired through to the config

src/app/outbound/endpoint.rs
olix0r: i like having the thing be nameable. is there a problem as it is?

lib/linkerd2-proxy-discover/src/buffer.rs
olix0r: in this case, i disagree? this is all one concatenated error message.

lib/linkerd2-proxy-discover/src/buffer.rs
olix0r: in one the resolution was lost because a stream completed, in the other, a task failed unexpectedly, so they are semantically different. ==

lib/linkerd2-proxy-discover/src/from_resolve.rs
olix0r: yeah, should be a vecdeque

lib/linkerd2-proxy-resolve/src/recover.rs
olix0r: If we're going to do this, that should be its own layer in a followup, but it doesn't seem high-priority at the moment, since this all gets put on a task with proper context

lib/linkerd2-proxy-core/src/recover.rs
olix0r: in our context, retry is an operation that occurs on an individual request. This _is_ a broader error recovery strategy.

src/app/outbound/endpoint.rs
hawkw: nope, that was a TIOLI, and i agree that having a name for it is valuable. carry on!

lib/linkerd2-proxy-discover/src/buffer.rs
hawkw: fair enough

lib/linkerd2-proxy-discover/src/buffer.rs
hawkw: > [..] they are semantically different.

in that case, i think we should probably rename them/change the `Display` impls/add comments to make that clearer...

lib/linkerd2-proxy-api-resolve/src/resolve.rs
```diff
@@ -0,0 +1,144 @@
...
+                path,
+                scheme: self.scheme.clone(),
+                context_token: self.context_token.clone(),
+            }))
+            .map(|rsp| {
+                debug!(metadata = ?rsp.metadata());
+                Resolution {
+                    inner: rsp.into_inner(),
+                }
+            })
```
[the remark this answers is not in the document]
olix0r: So, I'm not opposed to it, but this raises a larger point.. When is this necessary?

In this case, functionally, all of this is driven on the `discover::buffer`'s task, which is already instrumented with the target...

Now, on the other hand, api-resolve can be used without `discover::buffer`, so _something_ needs to be able to set this...

In general, I think this points to instrumentation being _never_ being done in arbitrary middlewares and _only_ being initiated through a dedicated tracing middleware.  Other middlewares shouldn't re-log, the target, though they can/should log the fact they are being called.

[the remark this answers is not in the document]
hawkw: > In this case, functionally, all of this is driven on the discover::buffer's task, which is already instrumented with the target...

ah, yeah, you're right; this would only need a span of its own if we wanted to record the time spent executing _this_ future in particular. i hadn't realised the buffer task was also instrumented because i was reading through the code top to bottom, which is probably a bug on my side :P

lib/linkerd2-proxy-resolve/src/recover.rs
olix0r: Simplified all of this by making target-tracing an explicit stack element.

olix0r: aaand undid that, but removed the target from here, since it's expected to be already part of the instrumented context

lib/linkerd2-request-instrument/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "linkerd2-request-instrument"
```
hawkw: fwiw, there is already a crate that implements this behavior: https://github.com/tokio-rs/tracing/blob/master/tracing-tower/src/request_span.rs

it's in the nursery still but could be stabilized (basically requires better docs/examples)

olix0r: @hawkw i backed that out in e970e0d -- it only covered the response future and not the discover/resolution stream; and it also misses any problem the buffer task might have, so i've restored it to where it was.

lib/linkerd2-proxy-discover/src/buffer.rs
olix0r: ftr: the blanket impl on Stream was removed, as it wasn't necessary

lib/linkerd2-request-instrument/Cargo.toml
hawkw: ah, okay. we should probably work on getting the `tracing-tower` layers in the stack separately, but that would be good to do...

olix0r:
Blocked on #347

lib/linkerd2-proxy-api-resolve/src/metadata.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// non 0-weighted endpoint.
+    ///
+    /// The default weight, corresponding to 1.0, is 10,000. This enables us to
+    /// specify weights as small as 0.0001 and as large as 400,000+.
+    ///
+    /// A float is not used so that this type can implement `Eq`.
+    weight: u32,
+
+    /// Arbitrary endpoint labels. Primarily used for telemetry.
+    labels: IndexMap<String, String>,
```
hawkw: This is probably something for another PR, but I wonder if it's worth reassessing our use of `IndexMap`s everywhere. IIRC we observed better performance with `IndexMap` (vs the standard library's `HashMap`) in the past. However, in Rust 1.36.0, the stdlib's `HashMap` was changed to be a reexport of `hashbrown::HashMap` (a SwissTable implementation) which should generally offer better performance. 

We might want to consider switching back to std's `HashMap` in cases where we don't rely on insertion ordering, since performance might be better (and it would maybe speed up the builds very slightly by not pulling another dependency).

src/app/outbound/mod.rs
```diff
@@ -45,10 +47,10 @@ pub fn server<R, P>(
...
 ) -> impl ServeConnection<Connection>
 where
-    R: Resolve<NameAddr, Endpoint = Metadata> + Clone + Send + Sync + 'static,
-    R::Future: futures::Future<Error = Unresolvable> + Send,
-    R::Resolution: Send,
-    <R::Resolution as Resolution>::Error: std::error::Error + Send + Sync + 'static,
+    R: Resolve<DstAddr, Endpoint = Endpoint> + Clone + Send + Sync + 'static,
+    R::Future: futures::Future + Send,
+    R::Resolution: Resolution + Send,
+    //<R::Resolution as Resolution>::Error: std::error::Error + Send + Sync + 'static,
```
hawkw: looks like the commented out code can be removed?

src/app/outbound/resolve.rs
```diff
@@ -0,0 +1,110 @@
...
+use linkerd2_proxy_resolve::{map_endpoint, recover};
+use linkerd2_request_filter as request_filter;
+use std::sync::Arc;
+use tower_grpc as grpc;
+
+pub fn resolve<R>(
+    suffixes: Vec<Suffix>,
+    backoff: ExponentialBackoff,
+    resolve: R,
+) -> map_endpoint::Resolve<
```
hawkw: TIOLI: What do you think about hiding this huge type behind a newtype? That would make errors shorter (and, I wonder if it would shave a little type inference time off?).

src/app/outbound/resolve.rs
```diff
@@ -0,0 +1,110 @@
...
+
+impl request_filter::RequestFilter<DstAddr> for PermitNamesInSuffixes {
+    type Error = Unresolvable;
+
+    fn filter(&self, dst: DstAddr) -> Result<DstAddr, Self::Error> {
+        if let Some(name) = dst.dst_concrete().name_addr() {
+            if self
+                .permitted
+                .iter()
+                .any(|suffix| suffix.contains(name.name()))
```
hawkw: Nit/TIOLI: I know the prior implementation used a `Vec`, but this seems like a job for a set to me?

src/app/outbound/resolve.rs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                tracing::debug!("suffix matches");
```
hawkw: Nit: should this include what the matched suffix was? or what the destination name was?

lib/linkerd2-request-filter/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+
+impl<T, I, S> tower::Service<T> for Service<I, S>
+where
+    I: RequestFilter<T>,
+    S: tower::Service<T>,
+    S::Error: Into<Error>,
+{
+    type Response = S::Response;
+    type Error = Error;
+    type Future = ResponseFuture<S::Future>;
```
adleong: is this preferable to `Either<S::Future, FutureResult<S::Response, Error>>`?

adleong: oh, I guess it's worse than that because of the `map_err`.  yuck.  nevermind.

lib/linkerd2-proxy-core/src/resolve.rs
```diff
@@ -1,31 +1,87 @@
 use futures::{Future, Poll};
+use linkerd2_error::Error;
 use std::net::SocketAddr;
 
 /// Resolves `T`-typed names/addresses as a `Resolution`.
 pub trait Resolve<T> {
```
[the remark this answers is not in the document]
olix0r: This trait alias exists to expose the Endpoint type so that type assertions are less crazy. Any service that returns a resolution automatically implements this. But, this permits
```rust
    R: Resolve<DstAddr, Endpoint = Endpoint>,
```
without this it would be
```rust
    R: Service<DstAddr>,
    R::Response: Resolution<Endpoint = Endpoint>
,
```

I find it especially useful to make aliases for special kinds of services, because otherwise every type bounds is just a huge mess of `Service` and it gets very confusing very fast. Though trait aliases aren't a panacea, being able to bundle a bunch of type assertions under a reusable handle is extremely helpful.

PR #342, opened 2019-09-05
update `tracing` and `tracing-subscriber`

hawkw:
This branch updates the proxy to depend on 0.1 versions of `tracing-log`
and `tracing-subscriber`, rather than alpha versions. The dependency on
`tracing-fmt` was removed, as that code was moved into
`tracing-subscriber` in tokio-rs/tracing#311.

Additionally, this should now allow us to use filter configuration that
involve dynamic filters on span fields, as well as on static targets.
Note that this may be of limited value *currently*, since the proxy
still uses our sui generis logging contexts rather than `tracing` spans,
but could be useful in the future.

### DIFF
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -16,15 +16,21 @@ thread_local! {
 pub mod trace {
     use super::{clock, Context as LegacyContext, CONTEXT as LEGACY_CONTEXT};
     use crate::Error;
-    use std::{env, fmt, str, time::Instant};
+    use std::{env, fmt, str, time::Instant, sync::{Arc, Mutex}};
+    use tracing_subscriber::{layer, fmt::{format, Builder, Context}};
     pub use tracing::*;
-    pub use tracing_fmt::*;
+    pub use tracing_subscriber::{Filter, reload, FmtSubscriber};
 
-    type SubscriberBuilder = Builder<format::NewRecorder, Format, filter::EnvFilter>;
+    type SubscriberBuilder = Builder<format::NewRecorder, Format, layer::Identity>;
+    type Subscriber = FmtSubscriber<format::NewRecorder, Format>;
 
     #[derive(Clone)]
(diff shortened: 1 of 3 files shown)


src/logging.rs
```diff
@@ -112,40 +124,39 @@ pub mod trace {
...
-            LevelHandle { inner }
+            LevelHandle { inner, current: Arc::new(Mutex::new(String::new())) }
         }
 
         pub fn set_level(&self, level: impl AsRef<str>) -> Result<(), Error> {
             let level = level.as_ref();
-            let filter = level.parse::<filter::EnvFilter>()?;
+            let filter = level.parse::<Filter>()?;
             self.inner.reload(filter)?;
+            let mut current = self.current.lock().unwrap();
```
adleong: Just curious why `set_level` has to take a `&self` instead of a `&mut self`.

olix0r:
@hawkw can you push to this branch again to rerun checks?

src/logging.rs
hawkw: We _could_ take `&mut self`, but we would still need to use locks internally. This is because admin service is `Clone`d, so we need to share one "current" state across all instances of this type. Therefore, taking `&mut self` isn't necessary, since we never actually mutate `self`.

hawkw:
tokio-rs/tracing#329 adds the missing `fmt::Display` impl for `Filter`, so I'll probably pick that up once merged & back out the change to store the filter string separately. That should trigger a new CI run.

hawkw:
> Looks good. It's a little unfortunate we need to add a mutex just to be able to display the current level, but makes sense for now.

mutex is no longer necessary in 0c7a39f48c23432b40779294e70f51cde65205a3

hawkw:
@olix0r i rebased master & pushed a new commit, but it still appears that CI is stuck?

PR #343, opened 2019-09-05
Use GitHub Actions for Pull Requests

olix0r:
Disables Travis on Pull Requests, replacing it with GitHub Actions.
Travis remains enabled for master builds for the moment.

### DIFF
--- a/.github/workflows/pull_request_rust.yml
+++ b/.github/workflows/pull_request_rust.yml
@@ -1,0 +1,29 @@
+name: Pull Request (Rust)
+
+on:
+  pull_request:
+    paths:
+    - 'Cargo.*'
+    - '*.rs'
+
+jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: docker://rust:1.37.0-slim-buster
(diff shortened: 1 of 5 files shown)


.github/workflows/pull_request_rust.yml
```diff
@@ -0,0 +1,28 @@
+name: Pull Request (Rust)
+
+on:
+  pull_request:
+    paths:
+    - 'Cargo.*'
+    - '*.rs'
```
hawkw: This is extremely cool! :D

.github/workflows/pull_request_rust.yml
```diff
@@ -0,0 +1,28 @@
+name: Pull Request (Rust)
+
+on:
+  pull_request:
```
hawkw: Will this be re-run for every push to a PR,  or just run once when the PR is created?

olix0r: all pushes to a pr, it seems

hawkw: 👍

olix0r:
I'm a bit miffed as to why the rust build is running on this PR and the docker build is not, but we'll fix it in post!

PR #344, opened 2019-09-05
CI: Run tests inside rust container

olix0r:
The first version of the action didn't actually use the rust containers
to drive the build. This change pins the ubuntu and rust versions.

Even though the base OS includes rust-1.37, we want to control our rust
version independently.

### DIFF
--- a/.github/workflows/pull_request_docker.yml
+++ b/.github/workflows/pull_request_docker.yml
@@ -1,15 +1,14 @@
-name: Pull Request (Docker)
+name: Docker PR
 
 on:
   pull_request:
     paths:
-    - '/Cargo.lock'
-    - '/Dockerfile'
+    - 'Cargo.lock'
+    - 'Dockerfile'
 
 jobs:
   docker:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/pull_request_docker.yml
```diff
@@ -1,15 +1,14 @@
-name: Pull Request (Docker)
+name: Docker PR
 
 on:
   pull_request:
     paths:
-    - '/Cargo.lock'
-    - '/Dockerfile'
+    - 'Cargo.lock'
+    - 'Dockerfile'
```
siggy: confirming this workflow is specifically for checking whether the docker build is working? is it possible that a change could cause the `Rust PR` workflow to pass but this workflow to fail? any reason not to run this all the time?

olix0r: yes, it's just checking that the dockerfile hasn't regressed. This can't be impacted by code changes, only by dependency/structure changes.

olix0r: also, this isn't even a required check currently, just informational, since that dockerfile isn't in the critical path for anything. starting with just the essentials in the critical path until i have a better handle on it all

PR #346, opened 2019-09-06
update tracing crates

hawkw:
This branch updates the proxy to depend on 0.1 versions of `tracing-log`
and `tracing-subscriber`, rather than alpha versions. The dependency on
`tracing-fmt` was removed, as that code was moved into
`tracing-subscriber` in tokio-rs/tracing#311.

Additionally, this should now allow us to use filter configuration that
involve dynamic filters on span fields, as well as on static targets.
Note that this may be of limited value *currently*, since the proxy
still uses our sui generis logging contexts rather than `tracing` spans,
but could be useful in the future.

Note that this is a duplicate of #342. I've submitted a new PR since CI
seems to be stuck refusing to build that branch. Therefore, this
closes #342.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -514,7 +514,7 @@ dependencies = [
  "linkerd2-dns-name 0.1.0",
  "ring 0.16.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "rustls 0.16.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


hawkw:
@adleong @kleimkuhler @olix0r this is a resubmission of #342, which you've previously reviewed --- this is just in order to get CI unstuck.

PR #347, opened 2019-09-07
core: Introduce the Recover trait

olix0r:
The `Recover` trait provides a means to inspect an error and, if it is
non-fatal, apply a backoff strategy. Backoff strategies are implemented
as a `Stream` that indicates when an operation may be
resumed/reattempted.

This affords an opportunity to make the `proxy::reconnect` module
decoupled from its backoff strategy (which will more easily allow the
strategy to be reused in other contexts, i.e. to support #341).

This change also corrects a logic error in the exponential backoff
strategy that prevented jitter from being applied. The code has been
refactored slightly to improve testability.


lib/linkerd2-error/src/recover.rs
```diff
@@ -0,0 +1,63 @@
+use super::{Error, Never};
+use futures::{stream, Stream};
+
+/// An error recovery strategy.
+pub trait Recover<E: Into<Error> = Error> {
```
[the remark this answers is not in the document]
hawkw: The equals sign means that there's a default value for the type parameter. So you can say `impl Recover for ...` (which is sugar for `impl Recover<Error> for ...`) or `impl<E> Recover<E> for ...`.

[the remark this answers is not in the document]
olix0r: it's a default, not an equality. it means it can be just referenced as `Recover` to mean `Recover<Error>`

lib/linkerd2-error/src/recover.rs
```diff
@@ -0,0 +1,63 @@
...
+    type Error: Into<Error>;
```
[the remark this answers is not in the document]
olix0r: practically, where this is used, we care about coercing this error and others into a unified type (see reconnect). If this assertion isn't made on the trait, it means that we'd need `R::Error: Into<Error>` everywhere it's used...

lib/linkerd2-exp-backoff/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+            if let Some(delay) = self.delay.as_mut() {
+                try_ready!(delay.poll());
+
+                self.delay = None;
+                self.iterations += 1;
+                return Ok(Some(()).into());
+            }
+
+            let backoff = {
+                let base = self.backoff.base(self.iterations + 1);
```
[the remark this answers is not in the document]
olix0r: good catch

lib/linkerd2-error/src/recover.rs
```diff
@@ -0,0 +1,63 @@
...
+    /// it signals that the caller should retry its operation. If the backoff is
+    /// polled agian, it is assumed that the operation failed and a new (possibly
+    /// longer) backoff is initated.
+    ///
+    /// If the error is not recoverable, it is returned immediately.
+    fn recover(&self, err: E) -> Result<Self::Backoff, E>;
+}
+
+#[derive(Copy, Clone, Debug, Default)]
+pub struct Immediately(());
```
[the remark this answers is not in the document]
olix0r: yes, this is done to prevent users from being able to expect/require that, though.

lib/linkerd2-error/src/recover.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    /// Given an E-typed error, determine if the error is recoverable.
+    ///
+    /// If it is, a backoff stream is returned. When the backoff becomes ready,
+    /// it signals that the caller should retry its operation. If the backoff is
+    /// polled agian, it is assumed that the operation failed and a new (possibly
+    /// longer) backoff is initated.
+    ///
+    /// If the error is not recoverable, it is returned immediately.
+    fn recover(&self, err: E) -> Result<Self::Backoff, E>;
```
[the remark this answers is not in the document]
olix0r: I understand your point but I don't know that I'm especially bothered by the current API... because backoff is lazy, it's basically free to return an unused backoff. I also like that this API allows the implementor to modify the Error type.

I'm -1 on `next_backoff` as proposed there, as there's nothing that requires a backoff strategy be strictly delay-based, and stream hides that detail nicely. Furthermore this would expose a much more stateful API, as the caller would have to inform the recover when the operation has been completed successfully! My proposed API does this all with RAII.

src/app/recover.rs
```diff
@@ -0,0 +1,20 @@
+use futures::Stream;
+use linkerd2_error::Error;
+pub use linkerd2_error::Recover;
+pub use linkerd2_exp_backoff::{ExponentialBackoff, ExponentialBackoffStream};
+
+pub fn always(backoff: ExponentialBackoff) -> Always {
```
hawkw: Nit/TIOLI: why is this in `app` rather than in the backoff crate?

lib/linkerd2-reconnect/src/service.rs
```diff
@@ -0,0 +1,152 @@
+use futures::{future, try_ready, Async, Future, Poll, Stream};
```
hawkw: It seems like it could be good to have some tests for the reconnect behavior?

src/app/recover.rs
olix0r: backoff doesn't have a dependency on on error

lib/linkerd2-error/src/recover.rs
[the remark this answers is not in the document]
adleong: I'm much less concerned about efficiency or allocations and much more worried about this being an API that is easy to use incorrectly.  A naive user (like myself) could easily call `recover` and use the result for each error, thus never actually backing off.  Or, I could just as easily assume that the backoff that I have is what I should use, but forget that I still need to call `recover` on each error to check that the error is recoverable.

lib/linkerd2-reconnect/src/service.rs
```diff
@@ -0,0 +1,152 @@
...
+                State::Recover {
+                    ref mut error,
+                    ref mut backoff,
+                } => {
+                    // Apply the recovery strategy. If the error isn't fatal,
+                    // prefer the existing backoff to the new one.
+                    let error = error.take().expect("error must be set");
+                    tracing::debug!(message="Service failed", %error);
+                    let new_backoff = self.recover.recover(error)?;
+                    tracing::debug!("Recovering");
```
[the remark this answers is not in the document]
olix0r: The reconnect module can't know this. It could be logged elsewhere, where it's knowable. (A given backoff impl may not be tied to a fixed duration, etc. Even this implementation doesn't know what it's delay will be until it actually starts doing work).

lib/linkerd2-reconnect/src/service.rs
```diff
@@ -0,0 +1,152 @@
...
+{
+    target: T,
+    recover: R,
+    make_service: M,
+    state: State<M::Future, R::Backoff>,
+}
+
+enum State<F: Future, B> {
+    Disconnected {
+        backoff: Option<B>,
```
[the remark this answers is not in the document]
olix0r: I think the clarity/correctness (esp, when a backoff can/should be held) is valuable. I'm at peace with Option/take.

src/app/recover.rs
olix0r: also, there's no reason this can't use other backoffs in the future, etc.... all of the concrete integrations will generally end up in app if they aren't generic.

lib/linkerd2-reconnect/src/service.rs
hawkw: none of these comments were blockers :)

lib/linkerd2-error/src/recover.rs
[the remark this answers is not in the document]
olix0r: > A naive user (like myself) could easily call recover and use the result for each error, thus never actually backing off.

Why would you think it appropriate to drop the first backoff stream in favor of the new one? That seems like an arbitrary, though intentional, decision. If we're saying that the docs can be improved to make the contract clearer, I agree. But as far as I can tell the other suggestions would have the same problem.

When recover gives you a backoff, you have to think "when should I poll this" and "when should I drop this". The answers to these questions are not ambiguous, though they may need to be more clearly spelled out in the docs.

My main argument is that if we changed this to be a more imperative style API where we had to call `set_recovered()` or some such to indicate that the backoff had completed, it would be just as easy to accidentally do the wrong thing if you don't read the full API contract. By virtue of returning a Stream, we can use the type system (as in reconnect) to be more clear about how the contract is enforced (for instance, it's impossible to reuse a backoff after a service has been connected successfully).

To eliza's point, we could try to decouple "checking" from creating the backoff, but it would proably need to be a bit funk, as well, in order to preserve some of the properties of the current api:

```rust
trait Recover<E> {
    type Error: Into<Error>;
    type Backoff: Stream<Item = (), Error = Self::Error>;

    fn check_recoverable(&self, err: E) -> Result<E, E>;
    fn new_backoff(&self, err: E) -> Self::Backoff;
}
```

That seems fine, I guess? but is it substantially better?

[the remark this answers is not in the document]
adleong: I'm fine with this API, I just wanted to give my perspective.

That said, I don't think I've clearly communicated my suggestion.  My intention is that `recover` would return a backoff object that has a `next_backoff` method.  The backoff object would be stateful, just like Stream is.  This utilizes RAII just like the current implementation does.  The difference is that `recover` is only called for the first error (i.e. when creating a new backoff) and then `next_backoff` is called on the backoff object for each error.  When service creation succeeds, the backoff object is dropped and the next error can call `recover` to create a new one.

To preserve the hiding of the backoff implementation, we could use a slightly different trait from Stream:

```
trait BackoffStream { 
  fn poll_backoff(&mut self, err: E) -> Result<Async<()>, E>;
}
```

But this has the problem that you have to clone the E each time you poll.  So really you'd want something more like...

```
trait BackoffStream {
  type Future: Future<Item=()>;
  fn next_backoff(&mut self, err: E) -> Result<Self::Future, E>;
}
```

But anyway, just sharing my thought process.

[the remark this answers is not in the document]
olix0r: @adleong yeah, i totally see where you're going. FWIW, I considered something like this but what we have here felt like the right balance. I reserve the right to change my mind at any point in the future ;p

lib/linkerd2-exp-backoff/src/lib.rs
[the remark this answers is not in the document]
kleimkuhler: I think it may be a little simpler to just pass in `self.iterations` without the ` + 1` and remove the `- 1` in `fn base`. I was wondering why that was there before seeing this conversation, but obviously not a blocker.

lib/linkerd2-reconnect/src/service.rs
```diff
@@ -0,0 +1,151 @@
...
+
+                State::Backoff(ref mut backoff) => {
+                    match backoff.as_mut().expect("backoff must be set").poll() {
+                        Ok(Async::NotReady) => return Ok(Async::NotReady),
+                        Ok(Async::Ready(more)) => State::Disconnected {
+                            // Only reuse the backoff if the backoff stream did not complete.
+                            backoff: more.and_then(move |()| backoff.take()),
+                        },
+                        Err(e) => {
+                            // If the backoff fails, try to recover, dropping the exisitng backoff.
```
kleimkuhler: I'm a little confused about situations where we have a backoff that has failed. Is this ever expected by the user, or is it only for recovering from situations where: if we have a `timer::Delay` that fails internally for some reason we can recover from that (or something like that)?

src/app/recover.rs
kleimkuhler: Any reason this is a function and not `fn new` method on an `Always` impl?

src/app/config.rs
```diff
@@ -810,38 +810,26 @@ fn parse_dns_suffix(s: &str) -> Result<dns::Suffix, ParseError> {
 pub fn parse_backoff<S: Strings>(
     strings: &S,
     base: &str,
-    default: Backoff,
-) -> Result<Backoff, Error> {
-    let env_disabled = format!("LINKERD2_PROXY_{}_EXP_BACKOFF_DISABLED", base);
```
kleimkuhler: Do we need to worry about still checking for this environment variable?

lib/linkerd2-error/src/recover.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+#[derive(Copy, Clone, Debug, Default)]
+pub struct Immediately(());
+
+// === impl Recover ===
+
+impl<E, B, F> Recover<E> for F
```
kleimkuhler: Is this impl here only for development use? I'm good with unused code and it doesn't seem to be used right now, but could possibly be helpful when writing a recover impl for a `Fn` or something?

lib/linkerd2-reconnect/src/service.rs
olix0r: reconnect can't make any assumptions about why a backoff stream might fail (or whether timers are used or what). it seems like a perfectly reasonable behavior for a recover/backoff to store the initial error and return it after some fixed number of attempts, for instance.

src/app/recover.rs
olix0r: everything else in our stack builders looks like this, because it's prettier

src/app/config.rs
olix0r: Executive decision: no. If you really don't want a backoff you can set min=0/max=0

lib/linkerd2-error/src/recover.rs
olix0r: Since `Recover` is really just a way to have a name for a `FnMut(E) -> Result<Backoff, E>`, it's convenient to provide an implementation. In fact, we can/should probably use this to get rid of the `always` stuff

lib/linkerd2-reconnect/src/service.rs
kleimkuhler: Yep makes sense. Sounds good just wanted to double check.

lib/linkerd2-error/src/recover.rs
kleimkuhler: Oh cool nice refactor to get rid of `always`. Looks good

lib/linkerd2-reconnect/src/service.rs
olix0r: after rereading what i just wrote, i realize that we should **not** try to recover if a backoff fails. It should be terminal.

PR #348, opened 2019-09-10
tap: add headers to tap events

hawkw:
PR linkerd/linkerd2-proxy-api#33 adds headers to the `RequestInit`,
`ResponseInit`, and `ResponseEnd` tap events. This branch updates the
proxy to populate these fields.

This updates the proxy-api dependency to the tag  `v0.1.10`, or
linkerd/linkerd2-proxy-api@34ab9c848f2cadb6d3e9dc47cca5b1c36adcd59f

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,7 @@ linkerd2-trace-context = { path = "lib/linkerd2-trace-context" }
 opencensus-proto       = { path = "lib/opencensus-proto" }
 
 # linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", rev = "ddbc3a4f7f8b0058801f896d27974d19ee98094c" }
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", branch = "kleimkuhler/tap-headers" }
+linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", rev = "1419992e32165e1b760d657205db4573890345b2" }
 
 bytes = "0.4"
 futures = "0.1"
@@ -129,7 +129,7 @@ quickcheck = { version = "0.8", default-features = false }
 linkerd2-metrics = { path = "./lib/linkerd2-metrics", features = ["test_util"] }
 linkerd2-task    = { path = "lib/linkerd2-task", features = ["test_util"] }
 # linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], rev = "ddbc3a4f7f8b0058801f896d27974d19ee98094c" }
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], branch = "kleimkuhler/tap-headers" }
+linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], rev = "1419992e32165e1b760d657205db4573890345b2" }
 flate2 = { version = "1.0.1", default-features = false, features = ["rust_backend"] }
 # `tokio-io` is needed for TCP tests, because `tokio::io` doesn't re-export
 # the `read` function.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,7 +63,7 @@ linkerd2-timeout = { path = "lib/linkerd2-timeout" }
 linkerd2-trace-context = { path = "lib/linkerd2-trace-context" }
 opencensus-proto       = { path = "lib/opencensus-proto" }
 
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", rev = "1419992e32165e1b760d657205db4573890345b2" }
+linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", rev = "4b788dd3d82564f0600a5904367e6d37827bb930" }
 
 bytes = "0.4"
 futures = "0.1"
@@ -95,7 +95,7 @@ tower-balance          = { git = "https://github.com/tower-rs/tower" }
 tower-load             = { git = "https://github.com/tower-rs/tower" }
 tower-request-modifier = { git = "https://github.com/tower-rs/tower-http" }
 tower-spawn-ready      = { git = "https://github.com/tower-rs/tower" }
-tower-grpc             = { git = "https://github.com/tower-rs/tower-grpc", default-features = false, features = ["protobuf"] }
+tower-grpc             = { version = "0.1", default-features = false, features = ["protobuf"] }
 
 # FIXME update to a release when available (>0.11)
 trust-dns-resolver = { git = "https://github.com/bluejekyll/trust-dns", rev = "7c8a0739dad495bf5a4fddfe86b8bbe2aa52d060", default-features = false }
@@ -127,7 +127,7 @@ net2 = "0.2"
 quickcheck = { version = "0.8", default-features = false }
 linkerd2-metrics = { path = "./lib/linkerd2-metrics", features = ["test_util"] }
 linkerd2-task    = { path = "lib/linkerd2-task", features = ["test_util"] }
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], rev = "1419992e32165e1b760d657205db4573890345b2" }
+linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], rev = "4b788dd3d82564f0600a5904367e6d37827bb930" }
 flate2 = { version = "1.0.1", default-features = false, features = ["rust_backend"] }
 # `tokio-io` is needed for TCP tests, because `tokio::io` doesn't re-export
 # the `read` function.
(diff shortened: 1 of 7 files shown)


src/tap/grpc/server.rs
```diff
@@ -265,14 +266,47 @@ impl iface::Tap for Tap {
...
+            let pseudos = vec![
+                http_types::headers::Header {
+                    name: ":method".to_owned(),
+                    value: req.method().as_str().as_bytes().into(),
+                },
+                http_types::headers::Header {
+                    name: ":scheme".to_owned(),
+                    value: req
+                        .uri()
+                        .scheme_part()
```
seanmonstar: If you want, there's `Uri::scheme_str`.

src/tap/grpc/server.rs
```diff
@@ -265,14 +266,47 @@ impl iface::Tap for Tap {
...
+                        .map(|scheme| scheme.as_str().as_bytes().into())
+                        .unwrap_or_default(),
+                },
+                http_types::headers::Header {
+                    name: ":authority".to_owned(),
+                    value: authority.as_bytes().into(),
+                },
+                http_types::headers::Header {
+                    name: ":path".to_owned(),
+                    value: req.uri().path().as_bytes().into(),
```
seanmonstar: `Uri::path` doesn't perfectly match `:path`. The `:path` header includes the query as well, so you probably want `Uri::path_and_query`.

hawkw:
thanks for the notes, @seanmonstar — I think 83ac0be addresses everything?

Cargo.toml
```diff
@@ -47,7 +47,8 @@ linkerd2-stack         = { path = "lib/linkerd2-stack" }
 linkerd2-task          = { path = "lib/linkerd2-task" }
 linkerd2-timeout       = { path = "lib/linkerd2-timeout" }
 
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", rev = "ddbc3a4f7f8b0058801f896d27974d19ee98094c" }
+# linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", rev = "ddbc3a4f7f8b0058801f896d27974d19ee98094c" }
```
kleimkuhler: This line can be removed

Cargo.toml
```diff
@@ -107,7 +108,8 @@ net2 = "0.2"
 quickcheck = { version = "0.8", default-features = false }
 linkerd2-metrics = { path = "./lib/linkerd2-metrics", features = ["test_util"] }
 linkerd2-task    = { path = "lib/linkerd2-task", features = ["test_util"] }
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], rev = "ddbc3a4f7f8b0058801f896d27974d19ee98094c" }
+# linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], rev = "ddbc3a4f7f8b0058801f896d27974d19ee98094c" }
```
kleimkuhler: This line as well!

grampelberg:
I'm good making it an opt-in feature (eg. when `tap -o json` is used). There will always be a need for a simple UI that just shows the basic data.

hawkw:
> Should we encode headers for all tap responses? Only a subset of queries can possibly use them. Should the the tap request object be modified to opt-in to header responses so that we don't do a bunch of needless allocation/serialization when the data can't possibly be used?

I'm +1 on this --- we should probably add something to the tap request indicating that headers should be included...

src/tap/grpc/server.rs
```diff
@@ -255,24 +284,70 @@ impl iface::Tap for Tap {
...
                     stream: next_id as u64,
                 };
-                Some((id, shared.events_tx.clone()))
-            } else {
-                None
+                // Is HTTP data being extracted from the request, and should
+                // headers be included?
+                match shared.extract {
+                    Extract::Http { headers } => return Some((id, shared.events_tx.clone(), headers)),
+                    // _ => {}
```
kleimkuhler: This can be removed

src/tap/grpc/server.rs
```diff
@@ -255,24 +284,70 @@ impl iface::Tap for Tap {
                     base: shared.base_id,
                     stream: next_id as u64,
                 };
-                Some((id, shared.events_tx.clone()))
-            } else {
-                None
+                // Is HTTP data being extracted from the request, and should
+                // headers be included?
+                match shared.extract {
+                    Extract::Http { headers } => return Some((id, shared.events_tx.clone(), headers)),
```
olix0r: So, this means that `Extract::Http` _must_ always be sent now, correct? While I agree with the intent, this is backwards-incompatible, so we should avoid it.  I think we should basically make this `let extract = shared.extract.map(|e| match e { Extract::Http { headers} => headers, _ => false }).unwrap_or(false)` ish thing... i.e. so that we only use this field to determine whether or not headers should be extracted. We can revisit the defaults later once we've reasonably expected all clients to take the new API.

hawkw: @olix0r That's not actually the case; `shared.extract` is a local type, _not_ the deserialized protobuf type: 
https://github.com/linkerd/linkerd2-proxy/blob/4b13908bad52f3d2aca84bebf4b941576f27fd29/src/tap/grpc/server.rs#L88-L91
which we construct from the received protobuf to make the representation simpler to match against internally.

If we receive an observe request with no `extract` field, we'll just construct `Extract::Http { headers: false }` by default, to preserve backwards compatibility:
https://github.com/linkerd/linkerd2-proxy/blob/4b13908bad52f3d2aca84bebf4b941576f27fd29/src/tap/grpc/server.rs#L140-L152

olix0r: OK, I think I understand. It's a little surprising given the overlapping names but I get where you're going. I think `Extract` could use a docstring explaining why it exists and how it relates to the protobuf concept of Extract. A name change could also help clarify (e.g., `ShouldExtractHttp`? idk) perhaps

hawkw: Okay, I can add some docs — I was basing this off of the similar pattern in the match module, where we have a separate internal match type that's constructed from the protobuf match type.

src/tap/grpc/server.rs
```diff
@@ -125,10 +142,24 @@ where
...
 
+        let extract = match req.extract.and_then(|ex| ExtractKind::try_from(ex).ok()) {
+            Some(ex) => ex,
+            None => {
+                // If there's no extract field, the request may have been sent
+                // by an older version of the Linkerd control plane. If this is
+                // the case, rather than failing the tap, just do the only
+                // behavior that older control planes know about --- extract
+                // HTTP data without headers.
+                let default = ExtractKind::Http { headers: false };
```
olix0r: consider a `Default` impl for `ExtractKind`? not a blocker

src/tap/grpc/server.rs
```diff
@@ -125,10 +142,24 @@ where
...
+                warn!(?default, "tap request with no extract field; using default");
```
olix0r: Ehhhh, I feel uneasy about emitting warnings to users with debug output like this in this case... It seems like something we should either panic on, because we expect never to encounter it, or to keep it at debug so users don't see it and get concerned, when they can't control anything that interacts with this.

src/tap/grpc/server.rs
```diff
@@ -255,24 +287,71 @@ impl iface::Tap for Tap {
...
+                // Is HTTP data being extracted from the request, and should
+                // headers be included?
+                match shared.extract {
+                    ExtractKind::Http { headers } => {
+                        return Some((id, shared.events_tx.clone(), headers))
+                    }
+                }
             }
+            None
         })?;
```
olix0r: tioli: This block may be clearer with more eager failure (there's basically no need for the `and_then`, right? we can just do `let shared = self.shared.upgrade()?;`

src/tap/grpc/server.rs
```diff
@@ -255,24 +287,71 @@ impl iface::Tap for Tap {
...
                     stream: next_id as u64,
                 };
-                Some((id, shared.events_tx.clone()))
-            } else {
-                None
+                // Is HTTP data being extracted from the request, and should
+                // headers be included?
+                match shared.extract {
+                    ExtractKind::Http { headers } => {
+                        return Some((id, shared.events_tx.clone(), headers))
```
olix0r: The `return` feels unnecessary here? Can the lower `None` not be moved up?

src/tap/grpc/server.rs
olix0r: Furthermore, are we really expecting that we'll upgrade all tap request to set this value? That was not what I expected, but we can do that if it's necessary.

src/tap/grpc/server.rs
```diff
@@ -125,10 +142,24 @@ where
             }
         };
 
+        let extract = match req.extract.and_then(|ex| ExtractKind::try_from(ex).ok()) {
+            Some(ex) => ex,
+            None => {
```
olix0r: nit, no match necessary if you `.unwrap_or_else` but, as noted, this coud just be `unwrap_or_default` if we did defaulting on ExtractKind.

src/tap/grpc/server.rs
hawkw: I thought about that — when I was writing this, I kind of thought this was clearer because it put the default behavior here in the one place it was actually used, rather than making a user hunt for it? But a default impl might be clearer.

PR #351, opened 2019-09-13
ci: Use a readymade release action

olix0r:
The first pass at releases used a container I found on the internet
called `github/ghx`, which provides a CLI interface to github that was
usable within actions. I've contacted the author, who says it is
unmaintained (and the sources for it are not public); so I've elected to
use softprops' release action.

In order to do this, I added a local action that handles the parsing of
the git `refs/*` syntax into the proper release metadata.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
     - name: git co
       uses: actions/checkout@v1
-    - name: Get Tag Info
+    - name: meta
       id: release-tag-meta
       uses: ./.github/actions/release-tag-meta
       with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,41 +18,12 @@ jobs:
       uses: ./.github/actions/release-tag-meta
       with:
         git-ref: ${{ github.ref }}
+    - runs: echo ${{ toJSON(steps) }}
     - name: make package
       env:
-        CARGO_RELEASE: 1
-      run: PACKAGE_VERSION=${{steps.release-tag-meta.name}} make package
-    - uses: actions/upload-artifact@v1
-      with:
-        name: package
-        path: target/release/package
-
-  test:
-    runs-on: ubuntu-latest
-    container:
-      image: docker://rust:1.37-buster
-    steps:
-    - name: git co
-      uses: actions/checkout@v1
-    - run: make test-integration
-
-  publish:
-    needs: [package, test]
-    runs-on: ubuntu-latest
-    steps:
-    - name: git co
-      uses: actions/checkout@v1
-    - name: Get Tag Info
-      id: release-tag-meta
-      uses: ./.github/actions/release-tag-meta
-      with:
-        git-ref: ${{ github.ref }}
-    - name: fetch
-      uses: actions/download-artifact@v1
-      with:
-        name: package
-        path: package
-    - runs: echo ${{ toJSON(steps) }}
+        CARGO_RELEASE: "1"
+        PACKAGE_VERSION: ${{ steps.release-tag-meta.name }}
+      run: make package
     - name: release
       uses: softprops/action-gh-release@b21b43d
       env:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,7 +18,7 @@ jobs:
       uses: ./.github/actions/release-tag-meta
       with:
         git-ref: ${{ github.ref }}
-    - runs: echo ${{ toJSON(steps) }}
+    - run: echo ${{ toJSON(steps) }}
     - name: make package
       env:
         CARGO_RELEASE: "1"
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -37,19 +40,23 @@ jobs:
...
+      id: release-tag-meta
+      uses: ./.github/actions/release-tag-meta
+      with:
+        git-ref: ${{ github.ref }}
     - name: fetch
       uses: actions/download-artifact@v1
       with:
         name: package
         path: package
+    - runs: echo ${{ toJSON(steps) }}
```
siggy: debug output?

.github/workflows/release.yml
```diff
@@ -13,12 +13,15 @@ jobs:
...
+      id: release-tag-meta
+      uses: ./.github/actions/release-tag-meta
+      with:
+        git-ref: ${{ github.ref }}
     - name: make package
       env:
         CARGO_RELEASE: 1
-      run: PACKAGE_VERSION=$(echo $GITHUB_REF | sed -E 's,^refs/tags/release/,,') make package
-
+      run: PACKAGE_VERSION=${{steps.release-tag-meta.name}} make package
```
siggy: how about:
```yaml
run: PACKAGE_VERSION=$(echo "$GITHUB_REF" | sed s,^refs/tags/release/,,) make package
```
...and then forego `Dockerfile`, `entrypoint.sh`, and `action.yml `? i could see utility in the custom action if it was doing more stuff.

.github/workflows/release.yml
```diff
@@ -37,19 +40,23 @@ jobs:
...
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      # ghx reads GITHUB_REF and then barfs on it, no matter its command-line
-      # args. so, we override it to the SHA so it all works out.
       with:
-        entrypoint: sh
-        args: -c "export TAG=$(echo $GITHUB_REF | sed -E 's,^refs/tags/,,') NAME=$(echo $GITHUB_REF | sed -E 's,^refs/tags/release/,,') ; export GITHUB_REF=$GITHUB_SHA ; cd package && ghx release create --verbose --tag=$TAG --name=$NAME *.tar.gz"
-
-
+        name: ${{ steps.release-tag-meta.name }}
```
siggy: ...i guess you'd need a step prior to this that does something like:
```yaml
- name: Get Tag Info
  id: release-tag-meta
  run: |
    name=$(echo "$GITHUB_REF" | sed s,^refs/tags/release/,,)
    echo ::set-output "name=name::$name"
```

.github/workflows/release.yml
```diff
@@ -13,43 +13,21 @@ jobs:
...
-    runs-on: ubuntu-latest
-    steps:
-    - name: fetch
-      uses: actions/download-artifact@v1
-      with:
-        name: package
-        path: package
+        CARGO_RELEASE: "1"
+        PACKAGE_VERSION: ${{ steps.release-tag-meta.name }}
+      run: make package
```
siggy: how about:
```yaml
run: PACKAGE_VERSION=$(echo "$GITHUB_REF" | sed s,^refs/tags/release/,,) make package
```
... and then forego `Dockerfile`, `entrypoint.sh`, and `action.yml`? i could see utility in the custom action setup if it was doing more stuff.

.github/workflows/release.yml
```diff
@@ -42,14 +45,17 @@ jobs:
...
+    - name: git co
+      uses: actions/checkout@v1
+    - name: Get Tag Info
+      id: release-tag-meta
+      uses: ./.github/actions/release-tag-meta
+      with:
+        git-ref: ${{ github.ref }}
     - name: release
-      uses: docker://github/ghx:master
+      uses: softprops/action-gh-release@b21b43d
```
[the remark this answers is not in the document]
olix0r: yeah, but a sha can't get updated from underneath us, and the tag will

.github/workflows/release.yml
olix0r: I don't think you can just do this without defining an output in an action.yml... am i wrong?

olix0r: From https://help.github.com/en/articles/development-tools-for-github-actions#set-an-output-parameter-set-output

> You will receive an error if you attempt to set an output value that was not declared in the action's metadata file.

.github/workflows/release.yml
olix0r: AFAIU: the action is needed so that i can define task outputs that can be referenced for the `name` input of the release action... i don't know of any other way of getting a modified value into that input...

olix0r:
OK, here's how a release turns out... 
image

The text is taken from the commit message at the referenced tag. I think if the tag has a signed message, it is preferred.

.github/workflows/release.yml
siggy: hmm, yeah it may be possible with https://github.com/actions/github-script but that's not much simpler than rolling our own action:
https://github.community/t5/GitHub-Actions/Pass-result-of-expression-to-custom-Github-Action-argument/m-p/30458#M435

PR #354, opened 2019-09-18
Add distributed tracing support

adleong:
This adds the trace context layer to the inbound and outbound stacks so that the proxy increments the span id on the server side and the client side.  Sampled traces are converted to the OpenCensus protobuf format and sent to the OpenCensus span exporter.

There will be additional followup work after this PR to enrich the span metadata with HTTP attributes such as method, path, status, etc.


src/app/inbound/mod.rs
```diff
@@ -31,6 +35,7 @@ pub fn server<P>(
     endpoint_http_metrics: super::HttpEndpointMetricsRegistry,
     route_http_metrics: super::HttpRouteMetricsRegistry,
     transport_metrics: transport::metrics::Registry,
+    span_sink: mpsc::Sender<oc::Span>,
```
olix0r: should this be a SpanSink?

olix0r: What happens when tracing is disabled?

adleong: This should not be a `SpanSink`.  That trait is for `trace_context::Span` whereas this channel is of `oc::Span` (i.e. the protobuf object).

lib/linkerd2-trace-context/src/layer.rs
```diff
@@ -142,7 +152,7 @@ where
         let f = self.inner.call(request);
 
         ResponseFuture {
-            trace: span.map(|span| (span, self.sink.clone())),
+            trace: span.map(|span| (span, sink.clone())),
```
hawkw: I think the `sink.clone()` here is no longer necessary, since you clone `self.sink` here:
https://github.com/linkerd/linkerd2-proxy/pull/354/files#diff-424b22001827f29238d4de967af40d6dR117

```suggestion
            trace: span.map(|span| (span, sink)),
```

src/app/inbound/mod.rs
```diff
@@ -1,13 +1,17 @@
...
 use crate::proxy::http::{
     client, insert, metrics as http_metrics, normalize_uri, profiles, router, settings,
     strip_header,
 };
 use crate::proxy::{accept, Server};
 use crate::transport::{self, connect, keepalive, tls, Connection};
-use crate::{core::listen::ServeConnection, svc, Addr};
+use crate::{core::listen::ServeConnection, svc, trace_context, Addr};
 use linkerd2_reconnect as reconnect;
+use opencensus_proto::trace::v1 as oc;
```
hawkw: Nit: it seems not great to me to make `app::inbound` and `app::outbound` be directly aware of OpenCensus — I think that ideally these modules would be written against interfaces that would abstract over the span collector. 

I'm not going to block the PR over this, but I'd like to hear if @olix0r has any thoughts.

src/app/inbound/mod.rs
```diff
@@ -54,6 +62,11 @@ where
         .push(transport_metrics.connect("inbound"))
         .push(rewrite_loopback_addr::layer());
 
+    let trace_context_layer = trace_context::layer(
+        span_sink
+            .as_ref()
```
hawkw: TIOLI: instead of 
```rust
span_sink
    .as_ref()
    .map(|span_sink| ... span_sink.clone() ... )
```
consider
```rust
span_sink.cloned().map(|span_sink| ... span_sink ...)
```

src/app/main.rs
```diff
@@ -403,6 +404,58 @@ where
...
+                .into_inner()
+                .make(addr.clone())
+        });
+
+        let (outbound_spans_tx, inbound_spans_tx) = match trace_collector_svc {
+            Some(trace_collector) => {
+                let (outbound_spans_tx, outbound_spans_rx) = mpsc::channel(100);
+                let (inbound_spans_tx, inbound_spans_rx) = mpsc::channel(100);
+
+                let node = oc::Node {
```
hawkw: Nit: again, it seems weird to me to do this much OpenCensus-specific setup in `main` --- I'd _prefer_ to see some of this in a function in a tracing module that we call with some configs.

src/app/main.rs
```diff
@@ -403,6 +404,58 @@ where
...
+                    config.destination_buffer_capacity,
+                    config.control_dispatch_timeout,
+                )
+                .into_inner()
+                .make(addr.clone())
+        });
+
+        let (outbound_spans_tx, inbound_spans_tx) = match trace_collector_svc {
+            Some(trace_collector) => {
+                let (outbound_spans_tx, outbound_spans_rx) = mpsc::channel(100);
```
hawkw: Nit: why 100?

src/app/spans.rs
```diff
@@ -0,0 +1,106 @@
...
+
+#[derive(Debug)]
+pub struct IdLengthError {
+    id: Vec<u8>,
+    expected_size: usize,
+    actual_size: usize,
+}
+
+impl error::Error for IdLengthError {
+    fn description(&self) -> &str {
```
hawkw: Nit: `Error::description` is [soft deprecated](https://doc.rust-lang.org/std/error/trait.Error.html#method.description), implementing it shouldn't be necessary.

src/app/spans.rs
```diff
@@ -0,0 +1,106 @@
...
+use opencensus_proto::trace::v1 as oc;
+use std::collections::HashMap;
+use std::{error, fmt};
+use tokio::sync::mpsc;
+
+const SPAN_KIND_SERVER: i32 = 1;
+const SPAN_KIND_CLIENT: i32 = 2;
+
+#[derive(Clone)]
+pub struct SpanConverter {
```
hawkw: It would be good to have a comment explaining what, exactly, this thing actually does...

src/app/spans.rs
```diff
@@ -0,0 +1,106 @@
...
+            start_time: Some(span.start.into()),
+            end_time: Some(span.end.into()),
+            attributes: None, // TODO: attributes
+            stack_trace: None,
+            time_events: None,
+            links: None,
+            status: None, // TODO: record this
+            resource: None,
+            same_process_as_parent_span: Some(self.kind == SPAN_KIND_CLIENT),
+            child_span_count: None,
```
hawkw: TIOLI: instead of having so many `None`s, you _could_ leave them empty and do
```rust
oc::Span {
    // fields...
   ...oc::Span::default()
}
```

though I'd leave the fields with `TODO`s on them.

src/app/mod.rs
```diff
@@ -14,6 +14,7 @@ mod metric_labels;
 mod outbound;
 mod profiles;
 mod proxy;
+mod spans;
```
hawkw: Nit: I'm not convinced that `app` is the right place for this code...not a big deal though.

src/app/main.rs
adleong: Picked a number out of a hat but this feels like it's the right ballpark.  It's also equal to the default max batch size, which is nice.

src/app/spans.rs
adleong: why, it converts spans, of course!

src/app/inbound/mod.rs
adleong: I think this has to be `span_sink.as_ref().cloned().map....` since `span_sink` is not an `Option<&'_ T>`

adleong: or maybe just `span_sink.clone().map...` ?

hawkw: ah, yeah, you're right --- the last one should work, i think

src/app/spans.rs
```diff
@@ -0,0 +1,106 @@
...
+use std::collections::HashMap;
+use std::{error, fmt};
+use tokio::sync::mpsc;
+
+const SPAN_KIND_SERVER: i32 = 1;
+const SPAN_KIND_CLIENT: i32 = 2;
+
+#[derive(Clone)]
+pub struct SpanConverter {
+    kind: i32,
```
kleimkuhler: Nit: If this is one of two values it probably doesn't need to be an `i32`?

kleimkuhler: ...Unless that size is needed for making an `oc::Span`

src/app/spans.rs
```diff
@@ -8,6 +8,11 @@ use tokio::sync::mpsc;
 const SPAN_KIND_SERVER: i32 = 1;
 const SPAN_KIND_CLIENT: i32 = 2;
 
+/// SpanConverter converts trace_context::Span objects into OpenCensus agent
```
hawkw: This is _extremely_ nitpicky, and feel free to ignore it, but: this reads like a Go doc comment to me. I think the preferred idiom in rust is:
- the first sentence of the doc comment should be a summary on a separate line
- RustDoc comments generally _don't_ start with the name of the type or function they document

See also: https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text

src/app/spans.rs
hawkw: I'm guessing the wire protocol probably defines this as 32 bits.

src/app/outbound/mod.rs
```diff
@@ -254,6 +268,11 @@ where
         .push_concurrency_limit(max_in_flight)
         .push_load_shed();
 
+    let trace_context_layer = trace_context::layer(
+        span_sink
+            .as_ref()
+            .map(|span_sink| SpanConverter::server(span_sink.clone(), trace_labels.clone())),
```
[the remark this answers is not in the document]
kleimkuhler: Looks like this line was changed from when I originally left this comment, but the unnecessary `.clone()` should still apply...

src/app/spans.rs
```diff
@@ -0,0 +1,107 @@
+use crate::trace_context;
+use linkerd2_error::Error;
+use opencensus_proto::trace::v1 as oc;
+use std::collections::HashMap;
+use std::{error, fmt};
+use tokio::sync::mpsc;
+
+const SPAN_KIND_SERVER: i32 = 1;
+const SPAN_KIND_CLIENT: i32 = 2;
```
hawkw: Nit/TIOLI: We could also represent this as
```rust
#[repr(i32)]
enum SpanKind {
    Server = 1,
    Client = 2,
}
```
and then use `as` to convert it into an i32 when it's used to construct the `oc::Span` below. (see https://doc.rust-lang.org/rust-by-example/custom_types/enum/c_like.html#c-like)

olix0r:
i'll echo what i think i see in the other feedback: it's preferable to not to make the layer itself optional or an either or something like that (at least if you like not waiting for a compiler). it's probably preferable to push the optionality into the layer...

adleong:
@olix0r [this commit](https://github.com/linkerd/linkerd2-proxy/pull/354/commits/c958f2a70caf46603ce2475b07a03e6db4c82514) pushed optionality into the layer by making the sink an `Option` and disabling the behavior when it's not present.

src/app/inbound/mod.rs
olix0r: I agree. I'd prefer that we re-export any of the OC types we want from the linkerd-opencensus crate -- that should be a complete API.

Not a hard blocker but the work's not done until that's cleaned up, imo.

src/app/spans.rs
olix0r: I don't understand why this is necessary. Why do we have to convert spans? Should this logic be part of linkerd-opencensus? What are we bridging that is best done in `app`?

adleong: `linkerd-opencensus` and `linkerd-trace-context` are independent crates which do not depend on each other.  each has their own Span struct, so something which depends on both needs to convert from one to the other.  In order for this conversion to happen in `linkerd-opencensus`, it would need to depend on `linkerd-trace-context`.

olix0r: @adleong ahh ok. is the intent that that trace-context can be used with non-opencensus spans? i'll dig in more to read the details, but curious on how you think about the relationship between the components (i.e. where is decoupling important, etc)

adleong: loose coupling is always a nice property to have, but realistically the crates are set up this way because it was the most expedient way to develop them in parallel.  will we ever replace opencensus with a different exporter? maybe.  but I don't want to over optimize for abstraction when we don't know what we want to abstract over.  

having self-contained library crates which are all tied together by the main crate is an architecture that makes sense to me.

src/app/config.rs
```diff
@@ -107,6 +107,8 @@ pub struct Config {
     /// Where to talk to the control plane.
     pub destination_addr: ControlAddr,
 
+    pub trace_collector_addr: Option<ControlAddr>,
```
olix0r: To be clear... we anticipate running the collector meshed with the proxy to support identity?

src/app/inbound/mod.rs
```diff
@@ -45,6 +50,9 @@ where
     let default_fwd_addr = config.inbound_forward.map(|a| a.into());
     let dispatch_timeout = config.inbound_dispatch_timeout;
 
+    let mut trace_labels = HashMap::new();
+    trace_labels.insert("direction".to_string(), "inbound".to_string());
```
olix0r: This feels weird to me; and, as discussed IRL, it feels weird to push this into the converter, specifically. I'd _expect_ all of the labels to be derived from the `Target` in the `trace_context` layer, but it sounds like some more work may be necessary to make that a reality.

src/app/main.rs
```diff
@@ -402,6 +403,58 @@ where
...
+                let (inbound_spans_tx, inbound_spans_rx) = mpsc::channel(100);
```
olix0r: It's not clear to me why we need two channels. What's the rational for not just one?

src/app/main.rs
```diff
@@ -402,6 +403,58 @@ where
...
+                    identifier: Some(oc::ProcessIdentifier {
+                        host_name: hostname::get_hostname().unwrap_or_default(),
+                        pid: 0,
+                        start_timestamp: Some(start_time.into()),
+                    }),
+                    library_info: None,
+                    service_info: Some(oc::ServiceInfo {
+                        name: "linkerd-proxy".to_string(),
+                    }),
+                    attributes: HashMap::new(),
```
olix0r: prefer `..oc::ProcessIdentifier::default()` rather than specifying unused fields.

src/app/config.rs
adleong: yes

PR #356, opened 2019-09-18
resolve: Do not send the 'k8s' scheme

olix0r:
The fact that the proxy sends resolve requests with a scheme value of
`k8s` is an accident of history. Given that this value is currently
unused by the controller, stop setting a scheme value entirely.

### DIFF
--- a/src/app/main.rs
+++ b/src/app/main.rs
@@ -339,7 +339,6 @@ where
         };
 
         let resolver = crate::api_resolve::Resolve::new(dst_svc.clone())
-            .with_scheme("k8s")
             .with_context_token(&config.destination_context);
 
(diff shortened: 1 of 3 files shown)


olix0r:
@kleimkuhler the "scheme" in tap should come from the http metadata -- if that's missing that sounds like a bug. this scheme is _like_ that, but not the same thing, really. we don't really have a use for the field at the moment, but we may want to start hydrating it i.e. "http", "https", "tcp", "tls", etc based on the context of the resolution (i.e. the http router vs the tcp forwarder, etc).

kleimkuhler:
@olix0r Okay thanks. I will look into it more

PR #357, opened 2019-09-19
Expose SNI values for opaque forwarded TLS streams

olix0r:
Fixes https://github.com/linkerd/linkerd2/issues/3446


olix0r:
This will probably stay in draft form until stable-2.6.0 is released, but will allow us to start testing to scope 2.7 work.

olix0r:
image at `gcr.io/linkerd-io/proxy:l2p-sni`

olix0r:
There are a number of issues that need to be addressed before this branch is suitable. For one, TLS  discovery is disabled on the outbound proxy (because it can never participate in mesh identity).  I'd like to decouple the listener into a stack as we address this.

PR #358, opened 2019-09-20
ci: Run integration tests on dedicated host

olix0r:
The linkerd org has dedicated hardware on which our integration tests
can be run.

This change reduces test time from ~80 minutes to...[pending]

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -46,7 +46,10 @@ jobs:
         echo "${{ secrets.DOCKER_PRIVATE_KEY }}" >~/.ssh/id
         # Use well-known public keys for the host to prevent middlemen.
         echo "${{ secrets.DOCKER_KNOWN_HOSTS }}" >~/.ssh/known_hosts
-        # Configure host with BatchMode to prevent idlelocking.
+        # Configure host with ServerAliveInterval to ensure that the client
+        # stays alive even when the server is busy emitting nothing.
+        # ServerAliveMaxCount ensures that server responds to these pings
+        # within ~5 minutes.
         (
           echo "Host linkerd-docker"
           echo "    User github"
@@ -54,6 +57,7 @@ jobs:
           echo "    IdentityFile ~/.ssh/id"
           echo "    BatchMode yes"
           echo "    ServerAliveInterval 60"
+          echo "    ServerAliveMaxCount 5"
         ) >~/.ssh/config
         # Build once with output, then build again to get the img ref (should
         # be a noop).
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -26,8 +26,54 @@ jobs:
...
+    - name: Test (Origin)
+      if: '!github.event.pull_request.head.repo.fork'
+      env:
+        DOCKER_HOST: "ssh://linkerd-docker"
+      run: docker run --rm $(cat image.ref) test-integration
+
+
+    # Try to delete the build image.
+    - name: Cleanup (Origin)
+      if: always() && !github.event.pull_request.head.repo.fork
```
kleimkuhler: Why is `always()` needed? I would think the same condition would be here as the previous non-forked steps.

.github/workflows/rust.yml
```diff
@@ -26,8 +26,54 @@ jobs:
...
+        # Use well-known public keys for the host to prevent middlemen.
+        echo "${{ secrets.DOCKER_KNOWN_HOSTS }}" >~/.ssh/known_hosts
+        # Configure host with BatchMode to prevent idlelocking.
+        (
+          echo "Host linkerd-docker"
+          echo "    User github"
+          echo "    Hostname ${{ secrets.DOCKER_ADDRESS }}"
+          echo "    IdentityFile ~/.ssh/id"
+          echo "    BatchMode yes"
+          echo "    ServerAliveInterval 60"
```
kleimkuhler: Nit: Is 60 a little often? I'm seeing when set this can be something like 5 minutes. Also looks like a `ServerAliveCountMax` pairs well with this if we care about checking aliveness.

.github/workflows/rust.yml
kleimkuhler: As in: `if: '!github.event.pull_request.head.repo.fork'`

olix0r: always makes this run when a prior step has failed

.github/workflows/rust.yml
olix0r:
```
     ServerAliveInterval
             Sets a timeout interval in seconds after which if no data has been received
             from the server, ssh(1) will send a message through the encrypted channel
             to request a response from the server.  The default is 0, indicating that
             these messages will not be sent to the server.
```

I've set this at 60s to ensure that we're sending messages on the connection even when a command being run stops emitting output (like the build does while building the proxy). WIthout this, the client can lose track of the server such that, even when the remote work completes, the client is never notified. Adding these heartbeats prevents that behavior. We can add `ServerAliveCountMax` if we run into situations where the network or build host dies and the client does not detect it, but that seems less of a concern.

PR #359, opened 2019-09-20
Add http metadata to spans as annotations

adleong:
We collect the following data from a sampled request/response and add it to the emitted span as an attribute map:

* http.method
* http.path
* http.authority OR http.host
* http.status_code


src/app/spans.rs
```diff
@@ -66,11 +85,14 @@ impl SpanConverter {
             kind: self.kind,
             start_time: Some(span.start.into()),
             end_time: Some(span.end.into()),
-            attributes: None, // TODO: attributes
+            attributes: Some(oc::span::Attributes {
+                attribute_map: attributes,
+                dropped_attributes_count: 0,
```
kleimkuhler: >  Attributes can be discarded because their keys are too long or because there are too many attributes

Are these things that need to be checked when making `attributes` above?

adleong: The list of attributes that we populate is hardcoded so I don't think we need to worry about this aspect of the API.

lib/linkerd2-trace-context/src/layer.rs
```diff
@@ -99,9 +100,9 @@ impl<F: Future, S> Future for MakeFuture<F, S> {
 
 // === impl Service ===
 
-impl<Svc, B, S> tower::Service<http::Request<B>> for Service<Svc, S>
+impl<Svc, B1, B2, S> tower::Service<http::Request<B1>> for Service<Svc, S>
```
olix0r: nit: What do we in other modules that have req and rsp body types? I believe it's typical to use `A` for the request body type.

lib/linkerd2-trace-context/src/layer.rs
```diff
@@ -180,3 +185,23 @@ where
         Ok(Async::Ready(inner))
     }
 }
+
+fn request_labels<Body>(labels: &mut HashMap<String, String>, req: &http::Request<Body>) {
+    labels.insert("http.method".to_string(), format!("{}", req.method()));
+    labels.insert("http.path".to_string(), req.uri().path().to_string());
```
olix0r: do you intend to omit query parameters?

lib/linkerd2-trace-context/src/layer.rs
```diff
@@ -180,3 +185,23 @@ where
         Ok(Async::Ready(inner))
     }
 }
+
+fn request_labels<Body>(labels: &mut HashMap<String, String>, req: &http::Request<Body>) {
```
olix0r: Can these maps be keyed on `&'static str` to avoid all of the needless `to_strings`?

lib/linkerd2-trace-context/src/layer.rs
adleong: It's inconsistent.  We have 11 files where we use `Request<A>` and 5 files where we use `Request<B1>`.  I find it pretty weird that when `B` is used alone it means "Body" but when it's used with `A` it means "the second letter of the alphabet", which is why I prefer `B1` `B2`.  Or `ReqBody` `RspBody`.

lib/linkerd2-trace-context/src/layer.rs
adleong: they can be but we have to eventually convert the keys to String to give it to prost.

lib/linkerd2-trace-context/src/layer.rs
olix0r: conversely, it's common to find `T` as a base generic with `S` and `U` in relation to it, but it's a total nit. fine with this too.

PR #360, opened 2019-09-20
Instrument OC exporter

adleong:
Fixes https://github.com/linkerd/linkerd2/issues/3453

Instrument the opencensus exporter with the following metrics:

```
# HELP span_export_streams Total count of opened span export streams
# TYPE span_export_streams counter
span_export_streams 1
# HELP span_export_requests Total count of span export request messages
# TYPE span_export_requests counter
span_export_requests 497
# HELP span_exports Total count of spans exported
# TYPE span_exports counter
span_exports 994
```

PR #361, opened 2019-09-24
fmt_metric should only borrow metric

adleong:
`Metric::fmt_metric` takes ownership of a metric parameter `M` even though it only needs to borrow this value to format it.  We change the trait definition to only borrow this parameter to allow it to be used in situations where you don't want `M` to be consumed.

### DIFF
--- a/lib/linkerd2-metrics/src/prom.rs
+++ b/lib/linkerd2-metrics/src/prom.rs
@@ -90,7 +90,7 @@ impl<'a, M: FmtMetric> Metric<'a, M> {
     }
 
     /// Formats a single metric without labels.
-    pub fn fmt_metric(&self, f: &mut fmt::Formatter<'_>, metric: M) -> fmt::Result {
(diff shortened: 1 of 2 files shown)


olix0r:
So, I'm not opposed to this; but it raises the question; why isn't this already necessary? Or why is it becoming necessary? Are we mistakenly cloning metrics? or are all of our metrics `Copy`?  Or do we implement `Metric` for lifetime-constrained types? What types of metrics are we introducing such that this current approach is no longer suitable?

olix0r:
Digging in a little, it looks like #360 is only passing counters to do this?  Here's counter:

```ryst
#[derive(Copy, Clone, Debug, Default, Eq, PartialEq)]
pub struct Counter(u64);
```

Copying counter is the same cost as taking it by ref (64b in either case).

olix0r:
I think I'm -0 on this change. All of our existing Metric implementations are Copy-friendly. I don't think this change is necessary?

adleong:
Ah, I didn't realize that Counter was Copy.  That was the missing piece.

PR #363, opened 2019-09-27
internal: Remove the SetKeepalive trait

olix0r:
In preparation of decoupling the listener logic from TLS discovery, I'm
trying to simplify the transport module.

This change is intended to make the `listen` and `connect` modules
primarily responsible for preparing a `TCPStream` before any other
transformations are applied to the IO stream. The `internal::Io` trait
can be narrowed and the various `Io` impls no longer need to proxy this
API.

In follow-up changes, the `listen` module will be decoupled from
identity/TLS; and some of the `transport` core types will be split out
into subcrates.


src/transport/mod.rs
```diff
@@ -19,11 +20,12 @@ pub use self::{
 
 fn set_nodelay_or_warn(socket: &::tokio::net::TcpStream) {
     if let Err(e) = socket.set_nodelay(true) {
-        tracing::warn!(
-            "could not set TCP_NODELAY on {:?}/{:?}: {}",
-            socket.local_addr(),
-            socket.peer_addr(),
-            e
-        );
+        tracing::warn!("failed to set nodelay: {}", e);
```
hawkw: TIOLI: record the error as a structured field:
```suggestion
        tracing::warn!(error = %e, "failed to set nodelay");
```

src/transport/mod.rs
```diff
@@ -19,11 +20,12 @@ pub use self::{
...
+    }
+}
+
+pub fn set_keepalive_or_warn(tcp: &mut TcpStream, ka: Option<Duration>) {
+    if let Err(e) = tcp.set_keepalive(ka) {
+        tracing::warn!("failed to set keepalive: {}", e);
```
hawkw: TIOLI: record the error as a structured field:
```suggestion
        tracing::warn!(error = %e, "failed to set keepalive");
```

src/transport/mod.rs
```diff
@@ -19,11 +20,12 @@ pub use self::{
 
 fn set_nodelay_or_warn(socket: &::tokio::net::TcpStream) {
```
hawkw: Nit: now that we've added an import for `tokio::net::TcpStream`, why not just change this to
```suggestion
fn set_nodelay_or_warn(socket: &TcpStream) {
```

src/transport/mod.rs
```diff
@@ -19,11 +20,12 @@ pub use self::{
...
-            "could not set TCP_NODELAY on {:?}/{:?}: {}",
-            socket.local_addr(),
-            socket.peer_addr(),
-            e
-        );
+        tracing::warn!("failed to set nodelay: {}", e);
+    }
+}
+
+pub fn set_keepalive_or_warn(tcp: &mut TcpStream, ka: Option<Duration>) {
```
hawkw: Nit: AFAICT this [doesn't require a mut reference to the `TcpStream`](https://docs.rs/tokio/0.1.12/tokio/net/struct.TcpStream.html#method.set_keepalive)?
```suggestion
pub fn set_keepalive_or_warn(tcp: &TcpStream, ka: Option<Duration>) {
```

src/transport/tls/listen.rs
```diff
@@ -172,14 +182,15 @@ impl<L: HasConfig, G> Listen<L, G> {
             });
 
             incoming
-                .map(move |(socket, remote_addr)| {
+                .map(move |(mut socket, remote_addr)| {
```
hawkw: As I mentioned above, I don't think we need a mut reference to set keepalive, so this shouldn't be necessary as far as I can tell?

PR #365, opened 2019-10-01
Handle 8 byte trace ids

adleong:
Nginx uses 8 byte trace ids whereas OpenCensus expects trace ids which are 16 bytes.  When we encounter an 8 byte trace id, we discard it, breaking the trace tree.

When reading trace id headers, we now left-pad trace ids with 0s to 16 bytes.  This is consistent with the way that OpenCensus handles trace ids less than 16 bytes: https://github.com/census-instrumentation/opencensus-go/blob/master/plugin/ochttp/propagation/b3/b3.go#L66

We also now log the appropriate error when we fail to emit a span.  This will allow us to distinguish between the scenario when the span receiver has been dropped vs when span conversion fails.

### DIFF
--- a/lib/linkerd2-trace-context/src/layer.rs
+++ b/lib/linkerd2-trace-context/src/layer.rs
@@ -178,8 +178,9 @@ where
             span.end = SystemTime::now();
             response_labels(&mut span.labels, &inner);
             trace!(message = "emitting span", ?span);
-            if sink.try_send(span).is_err() {
-                warn!("span dropped due to backpressure");
+            match sink.try_send(span) {
+                Ok(()) => {}
+                Err(e) => warn!(message = "span dropped", %e),
             }
         }
         Ok(Async::Ready(inner))
--- a/lib/linkerd2-trace-context/src/layer.rs
+++ b/lib/linkerd2-trace-context/src/layer.rs
@@ -180,7 +180,7 @@ where
             trace!(message = "emitting span", ?span);
             match sink.try_send(span) {
                 Ok(()) => {}
-                Err(e) => warn!(message = "span dropped", %e),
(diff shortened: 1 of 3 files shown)


adleong:
With this change, spans between Nginx and the rest of the trace tree are now properly connected: 
Screen Shot 2019-10-01 at 11 19 06 AM

lib/linkerd2-trace-context/src/propagation.rs
```diff
@@ -215,10 +215,20 @@ fn get_header_str<'a, B>(request: &'a http::Request<B>, header: &str) -> Option<
         .ok()
 }
 
-fn parse_header_id<B>(request: &http::Request<B>, header: &str) -> Option<Id> {
+fn parse_header_id<B>(request: &http::Request<B>, header: &str, pad_to: usize) -> Option<Id> {
```
olix0r: would feel much better about this with unit tests, but not a blocker to fixing the bug.

lib/linkerd2-trace-context/src/layer.rs
```diff
@@ -178,8 +178,9 @@ where
             span.end = SystemTime::now();
             response_labels(&mut span.labels, &inner);
             trace!(message = "emitting span", ?span);
-            if sink.try_send(span).is_err() {
-                warn!("span dropped due to backpressure");
+            match sink.try_send(span) {
+                Ok(()) => {}
+                Err(e) => warn!(message = "span dropped", %e),
```
olix0r: Is this error message actually helpful? If it is, consider renaming the field to `error` (because `e` is a weird key).

olix0r: also, prefer `if let Err(...) = ..`

PR #368, opened 2019-10-03
internal: Split Accept from Listen

olix0r:
This change reorganizes the listener logic. Now `transport::Listen` is soley
responsible for preparing and producing `TcpStreams`. These sockets are
then dispatched into an `Accept` service that takes a socket-like type and
returns a future that drives work on that connection.

`tls::Listen` is now `tls::AcceptTLS` and is _only_ responsible for
terminating TLS. In follow-up changes this type will be further
decomposed into an SNI-detecting piece and an Identity-terminating
piece. It's my hope that the `tls::Connection` type can be retired.

In the course of debugging, some logging has been switched from the
cumbersome internal logging module to `tracing`.

The admin thread's runtime has been reorganized so that `task::spawn` may be
used more liberally during initialization.


lib/linkerd2-proxy-core/src/listen.rs
```diff
@@ -1,27 +1,39 @@
-use crate::drain;
-use futures::Future;
+use futures::{Future, Poll};
 use linkerd2_error::Error;
+use tower::Service;
 
-pub trait ListenAndSpawn {
+pub trait Listen {
```
hawkw: This trait seems semantically similar to `Stream<Item = Connection>` — should there be a blanket impl or something?

src/app/inbound/mod.rs
```diff
@@ -25,19 +24,21 @@ mod set_remote_ip_on_req;
 
 pub use self::endpoint::{Endpoint, RecognizeEndpoint};
 
-pub fn server<P>(
+pub fn spawn<G, P>(
```
hawkw: It's not super obvious to me why this was changed to `spawn` rather than returning a task as it did previously — I'm not necessarily opposed but I'm curious about the rationale.

src/tap/accept.rs
```diff
@@ -0,0 +1,99 @@
...
+
+#[derive(Debug)]
+pub enum AcceptError {
+    Unauthorized(identity::Name),
+    NoPeerIdentity(tls::ReasonForNoIdentity),
+}
+
+impl AcceptPermittedClients {
+    pub fn new<I: Iterator<Item = identity::Name>>(
+        permitted_ids: I,
```
hawkw: style nit/tioli: could just be
```rust
pub fn new(permitted_ids: impl IntoIterator<Item = identity::Name>, server: super::Server)
```

src/tap/accept.rs
```diff
@@ -0,0 +1,99 @@
...
+            permitted_client_ids: permitted_ids.collect(),
+            server: TapServer::new(server),
+        }
+    }
+
+    fn serve(&mut self, conn: tls::Connection) -> ServeFuture {
+        let svc =
+            res_body_as_payload::Service::new(req_box_body::Service::new(self.server.clone()));
+        let fut = hyper::server::conn::Http::new()
+            //.with_executor(log2.executor())
```
hawkw: nit: probably remove commented-out code?

src/transport/tls/accept.rs
```diff
@@ -0,0 +1,280 @@
...
+    fn tls_server_config(&self) -> Arc<Config>;
+}
+
+/// Produces a server config that fails to handshake all connections.
+pub fn empty_config() -> Arc<Config> {
+    let verifier = rustls::NoClientAuth::new();
+    Arc::new(Config::new(verifier))
+}
+
+pub struct AcceptTLS<A: Accept<Connection>, T, G> {
```
hawkw: Nit/style: The Rust style guide suggests that
> In CamelCase, acronyms count as one word: use `Uuid` rather than `UUID`. In snake_case, acronyms are lower-cased: `is_xid_start`.

and elsewhere in the proxy, we tend to follow that convention. So I think this should be `AcceptTls` (and similar names elsewhere in this file ought to also be changed).

src/app/inbound/mod.rs
olix0r: I'm angling towards splitting inbound and outbound into subcrates so they can be compiled in parallel... I initially had this function returning and `impl Accept<...`, but it seemed better to just pass the listener in with it all and bind them together. I could have it return `impl Future`, but i figure this seals off any possibility of the compiler having to expose the inner stack type to the `main`... It's an experiment, I guess.

lib/linkerd2-proxy-core/src/listen.rs
olix0r: I'm not opposed to it, but we have no use for it at the moment. Less code wins ;)

src/app/main.rs
```diff
@@ -358,37 +335,62 @@ where
...
+                            if let Some((listener, tap_svc_name)) = control_listener {
+                                trace!("spawning tap server");
+                                task::spawn(tap_daemon.map_err(|_| ()));
+                                task::spawn(serve::serve(
+                                    "tap",
+                                    listener,
+                                    tls::AcceptTls::new(
+                                        get_original_dst.clone(),
+                                        local_identity,
+                                        tap::AcceptPermittedClients::new(
```
kleimkuhler: I think this addition of permitted clients is works really well with the accept logic in `src/tap/accept`. Was this with the intention of permitting more clients in the future though? I don't have a scenario in mind when we will permit more than just the tap controller.

src/app/main.rs
```diff
@@ -358,37 +335,62 @@ where
...
+                                            std::iter::once(tap_svc_name),
+                                            tap_grpc,
+                                        ),
+                                    ),
+                                    drain_rx.clone(),
+                                ));
+                            } else {
+                                drop((tap_daemon, tap_grpc));
```
kleimkuhler: This is good to see too. I don't think we were being explicit about this before.

src/app/admin/mod.rs
```diff
@@ -43,6 +42,10 @@ where
         }
     }
 
+    pub fn into_accept(self) -> Accept<M> {
```
kleimkuhler: Maybe put this into an `Into<Accept<M>>`?

src/app/admin/mod.rs
```diff
@@ -77,6 +77,35 @@ where
     }
 }
 
+impl<M: FmtMetrics + Clone + Send + 'static> svc::Service<tls::Connection> for Accept<M> {
```
kleimkuhler: I find it kind of odd that use two different types of `Service` here that seem to have the same purpose. Any reason `Admin` implements `hyper::Service` and not `tower`?

src/proxy/wrap_server_transport.rs
```diff
@@ -3,10 +3,10 @@ use tokio::io::{AsyncRead, AsyncWrite};
 use super::Source;
 
 /// Wraps serverside transports with additional functionality.
```
kleimkuhler: I didn't find the purpose of this very obvious right away. "Wrapping with additional functionality" seems like it could helpful, but if a case doesn't come to mind right away it seems a little vague?

Could this comment provide a quick explanation on what an `io` actually does with `Source`?

src/transport/tls/accept.rs
```diff
@@ -0,0 +1,280 @@
...
+
+pub struct AcceptTls<A: Accept<Connection>, T, G> {
+    accept: A,
+    tls: tls::Conditional<T>,
+    disable_protocol_detection_ports: IndexSet<u16>,
+    get_original_dst: G,
+}
+
+pub enum AcceptFuture<A: Accept<Connection>> {
+    TryTLS(Option<TryTLS<A>>),
```
kleimkuhler: nit: I think 3 letter naming in Rust tends to be first letter uppercase--rest lowercase.

src/tap/accept.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn call(&mut self, conn: tls::Connection) -> Self::Future {
+        match conn.peer_identity() {
+            tls::Conditional::Some(ref peer) if self.permitted_client_ids.contains(peer) => {
+                self.serve(conn)
+            }
+            tls::Conditional::None(tls::ReasonForNoIdentity::NoPeerName(
+                tls::ReasonForNoPeerName::Loopback,
+            )) => self.serve(conn),
+            tls::Conditional::Some(peer) => {
+                ServeFuture::Failed(Some(AcceptError::Unauthorized(peer).into()))
```
olix0r: This can't ship as-is; it needs to return a grpc-status: Unauthorized as it did previously

src/app/main.rs
olix0r: There are no concrete plans, but this will be a little more flexible if the tap controller identity changes, etc...

src/app/admin/mod.rs
olix0r: that will still be awkward... the users of this don't actually care about things being an `Accept`, they care that it's an `impl Accept...` So you'd have to do some weird explicit type specification..

src/app/admin/mod.rs
olix0r: I don't know. Unrelated to this change, though.

src/proxy/wrap_server_transport.rs
olix0r: Yeah, I'm removing this in the next PR.

kleimkuhler: Sounds good

src/app/inbound/mod.rs
hawkw: Okay, yeah, that makes sense! Just wondering what the motivation was.

PR #369, opened 2019-10-07
internal: Split transport utilities into a subcrate

olix0r:
The `transport` module contains utilities for working with transports that are
not otherwise coupled to the main application code.

This change replaces the `WrapServerTransport` trait with a `TransportLabels`
trait that is only responsible for producing metrics labels. The transport module
is otherwise (mostly) ignorant of transport labels. (It is responsible for setting an
Errno label).

`proxy::Server` has (temporarily) been tasked with instrumenting the connection
with telemetry, though this will likely be further decoupled in follow-up changes.


olix0r:
Note: I split https://github.com/linkerd/linkerd2-proxy/pull/370 out of this. It should be reviewed/merged first.

lib/linkerd2-proxy-transport/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+untrusted = "0.7"
+
+[target.'cfg(target_os = "linux")'.dependencies]
+libc = "0.2"
+procinfo = "0.4.2"
+
+[dev-dependencies]
+linkerd2-identity = { path = "../linkerd2-identity", features = ["test-util"] }
+tower-util = "0.1"
+tracing-subscriber = "0.1"
```
hawkw: TIOLI: if we opted out of default features on `tracing-subscriber` we could avoid having a few dependencies in the lockfile...but maybe having nicer formatting in test output is good 🤷‍♀

lib/linkerd2-proxy-transport/tests/tls_accept.rs
```diff
@@ -0,0 +1,320 @@
...
+    SF: Future<Item = SR, Error = io::Error> + Send + 'static,
+    SR: Send + 'static,
+{
+    {
+        use tracing_subscriber::{fmt, EnvFilter};
+        let sub = fmt::Subscriber::builder()
+            .with_env_filter(EnvFilter::from_default_env())
+            .finish();
+        if tracing::subscriber::set_global_default(sub).is_err() {
+            eprintln!("logging disabled");
```
hawkw: This isn't strictly true — `set_global_default` might fail if a global default subscriber has already been set, too, in which case logging is still enabled.

lib/linkerd2-proxy-transport/tests/tls_accept.rs
```diff
@@ -0,0 +1,320 @@
...
+}
+
+/// Runs a test for a single TCP connection. `client` processes the connection
+/// on the client side and `server` processes the connection on the server
+/// side.
+fn run_test<C, CF, CR, S, SF, SR>(
+    client_tls: tls::Conditional<(CrtKey, Name)>,
+    client: C,
+    server_tls: tls::Conditional<CrtKey>,
+    server: S,
```
hawkw: TIOLI: IMO passing this many params to this function makes it a little hard to tell what's going on when looking at the tests...WDYT about combining them into `Server` and `Client` structs or something, so that the purpose of the params is clearer at the callsite?

Not a big deal in any case.

src/transport/labels.rs
```diff
@@ -0,0 +1,114 @@
...
+/// Implements `FmtLabels`.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub struct Key {
+    direction: Direction,
+    peer: Peer,
+    tls_status: TlsStatus,
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+struct Direction(&'static str);
```
kleimkuhler: Was `Direction` an `Enum` before? If not, maybe a good time to consider doing so? I only seeing this being `Inbound` or `Outbound` and it could be helpful to have that in the type.

It would also make it more consistent with `Peer` being an `Enum` already.

src/app/serve.rs
```diff
@@ -1,44 +1,43 @@
...
 
 /// Spawns a task that binds an `S`-typed server with an `L`-typed listener until
 /// a drain is signaled.
-pub fn serve<L, A>(
-    server: &'static str,
-    listen: L,
-    accept: A,
-    drain: drain::Watch,
-) -> impl Future<Item = (), Error = ()>
+pub fn spawn<L, A>(server: &'static str, listen: L, accept: A, drain: drain::Watch)
```
kleimkuhler: nit: I think documentation is old. There is no longer an `S`, and `L` and `A` are now linked... `A` accepts `L::Connection`-typed connections.

src/transport/labels.rs
olix0r: There's another direction enum... all of the labels need to get cleaned up, but let's do that later.

PR #372, opened 2019-10-14
give each test proxy its own tracing subscriber

hawkw:
This branch reworks the test-support proxy so that each test gets its
own `tracing` subscriber, rather than sharing a single global
subscriber. Running a whole bunch of tests in parallel with verbose
tracing output turned on is now much snappier.

As a future extension of this, we might want to replace the test-mode
subscriber with a subscriber that, rather than printing to stdout,
writes to a buffer and dumps the whole buffer when the test completes,
along with a message identifying which test the output was collected
from, or with one that prefixes all log lines with the name of the test.
Ideally, this kind of test-specific functionality could be added 
upstream.


olix0r:
thanks!

PR #371, opened 2019-10-11
Ver/split proxies

olix0r:

PR #374, opened 2019-10-15
Adopt tokio-tracing contexts

olix0r:
We had our own poorly-implemented logging contexts. This change
replaces all uses of the old context implementation with `tracing`
(usually via `tracing-futures`).

Some types, like `server::Source`, `config::H2Settings`, etc, have been
moved in anticipation of #371.


src/app/control.rs
```diff
@@ -263,10 +268,10 @@ pub mod resolve {
             let target = client::Target {
                 addr,
                 server_name: dst.identity.clone(),
-                log_ctx: logging::admin().client("control", dst.addr.clone()),
             };
 
-            State::Inner(mk_svc.call(target))
+            info_span!("control", peer_addr=%addr, peer_identity=?dst.identity.clone())
```
hawkw: Cloning the identity shouldn't be necessary — it's recorded as a reference
```suggestion
            info_span!("control", peer_addr=%addr, peer_identity=?dst.identity)
```

hawkw: Nit/style: Typically prefer spaces around the `=` in `tracing` macros, the way you'd format a Rust assignment.
```suggestion
            info_span!("control", peer_addr = %addr, peer_identity = ?dst.identity.clone())
```

src/app/control.rs
```diff
@@ -263,10 +268,10 @@ pub mod resolve {
...
+                .in_scope(move || State::Inner(mk_svc.call(target).instrument(Span::current())))
```
hawkw: Nit/TIOLI: if we update to `tracing-futures` 0.1 (which we should do anyway) there is [a convenience method](https://docs.rs/tracing-futures/0.1.0/tracing_futures/trait.Instrument.html#method.in_current_span) for this:
```suggestion
                .in_scope(move || State::Inner(mk_svc.call(target).in_current_span()))
```

src/app/control.rs
hawkw: Nit/TIOLI: you can also namespace `tracing` field names with dots, so this could be
```suggestion
            info_span!("control", peer.addr=%addr, peer.identity=?dst.identity.clone())
```
future `tracing` APIs will have nice globbing support for field names based on dots.

src/app/errors.rs
```diff
@@ -124,12 +131,20 @@ fn map_err_to_5xx(e: Error) -> StatusCode {
     } else if let Some(_) = e.downcast_ref::<router::NotRecognized>() {
         error!("could not recognize request");
         http::StatusCode::BAD_GATEWAY
-    } else if let Some(err) = e.downcast_ref::<outbound::RequireIdentityError>() {
-        error!("{}", err);
-        http::StatusCode::FORBIDDEN
+    } else if let Some(err) = e.downcast_ref::<StatusError>() {
+        error!("{} {}", err.status, err.message);
```
hawkw: TIOLI: could record the status and message as separate structured fields, with
```suggestion
        error!(%err.status, %err.message);
```

src/app/inbound/mod.rs
```diff
@@ -123,7 +130,10 @@ pub fn spawn<P>(
             profiles_client,
             dst_route_layer,
         ))
-        .push(strip_header::request::layer(super::DST_OVERRIDE_HEADER));
+        .push(strip_header::request::layer(super::DST_OVERRIDE_HEADER))
+        .push(trace::layer(
+            |dst: &DstAddr| info_span!("logical", dst=%dst.dst_logical()),
```
hawkw: Nit/style: prefer spaces around `=`
```suggestion
            |dst: &DstAddr| info_span!("logical", dst = %dst.dst_logical()),
```

src/app/inbound/mod.rs
```diff
@@ -146,33 +156,30 @@ pub fn spawn<P>(
...
+                                a
+                            })
+                        })
+                    })
                     .or_else(|| {
                         super::http_request_l5d_override_dst_addr(req)
+                            .ok()
                             .map(|override_addr| {
-                                debug!("inbound dst={:?}; dst-override", override_addr);
+                                debug!("using l5d-dst-override");
```
hawkw: Nit/TIOLI: we use the value of `CANONICAL_DST_HEADER` const in the above message, should we use the dst override const here as well?

hawkw: Nit: should we still record the value of `override_addr` as well?
```suggestion
                                debug!(?override_addr, "using l5d-dst-override");
```

src/app/inbound/mod.rs
```diff
@@ -205,13 +213,15 @@ pub fn spawn<P>(
         .push(insert::layer(move || {
             DispatchDeadline::after(dispatch_timeout)
         }))
-        .push(super::errors::layer())
+        .push(errors::layer())
+        .push(trace::layer(
+            |src: &Source| info_span!("source", peer_ip = %src.remote.ip(), peer_id=?src.tls_peer),
```
hawkw: Nit: 
```suggestion
            |src: &Source| info_span!("source", peer_ip = %src.remote.ip(), peer_id = ?src.tls_peer),
```

hawkw: Also, consider:
```suggestion
            |src: &Source| info_span!("source", peer.ip = %src.remote.ip(), peer.id=?src.tls_peer),
```

src/app/main.rs
```diff
@@ -360,50 +365,61 @@ where
...
-                                    tls::AcceptTls::new(
-                                        get_original_dst.clone(),
-                                        local_identity,
-                                        tap::AcceptPermittedClients::new(
-                                            std::iter::once(tap_svc_name),
-                                            tap_grpc,
-                                        ),
-                                    ),
-                                    drain_rx.clone(),
+                                info_span!("tap", local_addr=%listener.local_addr()).in_scope(
```
hawkw: Nit:
```suggestion
                                info_span!("tap", local_addr = %listener.local_addr()).in_scope(
```

src/app/main.rs
```diff
@@ -360,50 +365,61 @@ where
...
                                 );
                             } else {
                                 drop((tap_daemon, tap_grpc));
                             }
 
-                            trace!("spawning dns resolver");
-                            task::spawn(logging::admin().bg("dns-resolver").future(dns_bg));
+                            info_span!("dns").in_scope(|| {
+                                trace!("spawning");
+                                tokio::spawn(dns_bg);
```
hawkw: Is the "dns" span actually propagated to `dns_bg` or does this need an `instrument`?

src/app/main.rs
```diff
@@ -476,47 +492,55 @@ where
                 ..oc::Node::default()
             };
             let span_exporter = SpanExporter::new(trace_collector, node, spans_rx, span_metrics);
-            task::spawn(span_exporter.map_err(|e| {
-                error!("span exporter failed: {}", e);
-            }));
+            tokio::spawn(
+                span_exporter
+                    .map_err(|e| {
+                        error!("span exporter failed: {}", e);
```
hawkw: Nit/TIOLI: consider a structured field for the error:
```suggestion
                        error!(error = %e, "span exporter failed");
```

src/app/main.rs
```diff
@@ -476,47 +492,55 @@ where
...
-            tap_layer.clone(),
-            outbound_handle_time,
-            endpoint_http_metrics.clone(),
-            route_http_metrics.clone(),
-            retry_http_metrics,
-            transport_metrics.clone(),
-            spans_tx.clone(),
-            drain_rx.clone(),
-        );
+        info_span!("out", local_addr=%outbound_listener.local_addr()).in_scope(|| {
```
hawkw: Nit: 
```suggestion
        info_span!("out", local_addr = %outbound_listener.local_addr()).in_scope(|| {
```

src/app/main.rs
```diff
@@ -476,47 +492,55 @@ where
...
-            profiles_client,
-            tap_layer,
-            inbound_handle_time,
-            endpoint_http_metrics,
-            route_http_metrics,
-            transport_metrics,
-            spans_tx,
-            drain_rx,
-        );
+        info_span!("in", local_addr=%inbound_listener.local_addr()).in_scope(move || {
```
hawkw: Nit:
```suggestion
        info_span!("in", local_addr = %inbound_listener.local_addr()).in_scope(move || {
```

src/app/outbound/mod.rs
```diff
@@ -174,7 +172,10 @@ pub fn spawn<R, P>(
     // application-selected original destination.
     let distributor = endpoint_stack
         .push(fallback::layer(balancer_layer, orig_dst_router_layer))
-        .serves::<DstAddr>();
+        .serves::<DstAddr>()
+        .push(trace::layer(
+            |dst: &DstAddr| info_span!("concrete", dst=%dst.dst_concrete()),
```
hawkw: Nit: 
```suggestion
            |dst: &DstAddr| info_span!("concrete", dst = %dst.dst_concrete()),
```

src/app/outbound/mod.rs
```diff
@@ -197,16 +198,17 @@ pub fn spawn<R, P>(
     // This is shared across addr-stacks so that multiple addrs that
     // canonicalize to the same DstAddr use the same dst-stack service.
     let dst_router = dst_stack
+        .push(trace::layer(
+            |dst: &DstAddr| info_span!("logical", dst=%dst.dst_logical()),
```
hawkw:
```suggestion
            |dst: &DstAddr| info_span!("logical", dst = %dst.dst_logical()),
```

src/app/outbound/mod.rs
```diff
@@ -237,22 +239,19 @@ pub fn spawn<R, P>(
...
+        .push(trace::layer(|addr: &Addr| info_span!("addr", %addr)))
         .push_buffer_pending(max_in_flight, DispatchDeadline::extract)
         .push(router::layer(
-            router::Config::new("out addr", capacity, max_idle_age),
+            router::Config::new(capacity, max_idle_age),
             |req: &http::Request<_>| {
                 super::http_request_l5d_override_dst_addr(req)
                     .map(|override_addr| {
-                        debug!("outbound addr={:?}; dst-override", override_addr);
+                        debug!("using dst-override");
```
hawkw: Nit: should this record the override addr as well?
```suggestion
                        debug!(?override_addr, "using dst-override");
```

src/proxy/buffer.rs
```diff
@@ -236,11 +227,10 @@ where
...
     Req: Send + 'static,
 {
-    pub fn new<E>(svc: S, deadline: D, capacity: usize, exec: &mut E) -> Self
-    where
-        E: buffer::WorkerExecutor<Dequeue<S>, Stealer<Req>>,
-    {
-        let inner = buffer::Buffer::with_executor(Dequeue(svc), capacity, exec);
+    pub fn new(svc: S, deadline: D, capacity: usize) -> Self {
+        let mut exec =
+            tokio::executor::DefaultExecutor::current().instrument(tracing::Span::current());
```
hawkw: Nit: if we update to `tracing-futures` 0.1, this can be:
```suggestion
            tokio::executor::DefaultExecutor::current().in_current_span();
```

hawkw: Also, I'm not sure if this is necessary — I think `tower-buffer` already propagates spans: https://github.com/tower-rs/tower/blob/e414b2b7d3ce8220aaf321d42af70cf37ec127d0/tower-buffer/src/service.rs#L96-L99

src/proxy/http/client.rs
```diff
@@ -131,20 +127,19 @@ where
...
-            .executor();
-
         match *config.http_settings() {
             Settings::Http1 {
                 keep_alive,
                 wants_h1_upgrade: _,
                 was_absolute_form,
             } => {
+                let exec = tokio::executor::DefaultExecutor::current()
+                    .instrument(info_span!("http1", %peer_addr));
```
hawkw: TIOLI: consider
```suggestion
                    .instrument(info_span!("http1", %peer.addr));
```

src/proxy/http/h2.rs
```diff
@@ -115,16 +112,20 @@ where
                 ConnectState::Connect(ref mut fut) => try_ready!(fut.poll().map_err(Into::into)),
                 ConnectState::Handshake(ref mut hs) => {
                     let (tx, conn) = try_ready!(hs.poll());
-                    let _ = self
-                        .executor
-                        .execute(conn.map_err(|err| debug!("http2 conn error: {}", err)));
+
+                    DefaultExecutor::current()
+                        .instrument(info_span!("h2", peer_addr=%self.peer_addr))
```
hawkw: Nit:
```suggestion
                        .instrument(info_span!("h2", peer_addr = %self.peer_addr))
```

hawkw: also, consider: 
```suggestion
                        .instrument(info_span!("h2", peer.addr=%self.peer_addr))
```

src/proxy/http/h2.rs
```diff
@@ -115,16 +112,20 @@ where
...
+                        .spawn(Box::new(conn.map_err(|error| debug!(%error, "failed"))))
+                        .map_err(Error::from)?;
 
                     return Ok(Connection { tx }.into());
                 }
             };
 
+            let exec =
+                DefaultExecutor::current().instrument(info_span!("h2", peer_addr=%self.peer_addr));
```
hawkw: Not:
```suggestion
                DefaultExecutor::current().instrument(info_span!("h2", peer_addr = %self.peer_addr));
```

src/proxy/http/router.rs
```diff
@@ -126,22 +113,18 @@ where
             self.config.capacity,
             self.config.max_idle_age,
         );
-        let span = debug_span!("router", name = self.config.proxy_name);
-        let ctx = logging::Section::Proxy.bg(self.config.proxy_name);
-        let cache_daemon = ctx.future(cache_bg);
-        tokio::spawn(cache_daemon);
-
-        Service { inner, span }
+        tokio::spawn(cache_bg.instrument(info_span!("router")));
```
hawkw: Should the background future be 
```suggestion
        tokio::spawn(cache_bg.instrument(info_span!("router_bg")));
```
or something?

src/proxy/http/upgrade.rs
```diff
@@ -157,47 +141,39 @@ impl Drop for Inner {
...
-
-            if let Err(_) = self.upgrade_executor.execute(fut) {
-                trace!("error spawning HTTP upgrade task");
-            }
+            tokio::spawn(
+                self.upgrade_drain_signal
+                    .take()
+                    .expect("only taken in drop")
+                    .watch(both_upgrades, |_| ())
+                    .instrument(Span::current()),
```
hawkw: Nit: if we update to `tracing-futures` 0.1, this can be 
```suggestion
                    .in_current_span(),
```

src/trace.rs
```diff
@@ -0,0 +1,309 @@
...
+fn subscriber_builder() -> SubscriberBuilder {
+    let start_time = clock::now();
+    FmtSubscriber::builder().on_event(Format { start_time })
+}
+
+struct Format {
+    start_time: Instant,
+}
+
+impl<N> tracing_subscriber::fmt::FormatEvent<N> for Format
```
hawkw: This custom implementation was really only necessary in order to format the custom log context along with the `tracing` span. Since we're no longer doing that, this can be replaced with the default implementation (with ANSI formatting disabled). If we want to keep the current formatting for timestamps, we'd need to provide a custom [timestamp formatter](https://docs.rs/tracing-subscriber/0.1.5/tracing_subscriber/fmt/time/trait.FormatTime.html) (although there is an [`Uptime` formatter](https://docs.rs/tracing-subscriber/0.1.5/tracing_subscriber/fmt/time/struct.Uptime.html), it formats the timestamps differently from ours).

src/trace.rs
```diff
@@ -0,0 +1,309 @@
...
+                write!(f, "{{{}}}", fields)?;
+            }
+            ":".fmt(f)
+        })?;
+        if seen {
+            f.pad(" ")?;
+        }
+        Ok(())
+    }
+}
```
hawkw: if we switch to the default ormatter as I mentioned above, this can all go away.

src/trace.rs
```diff
@@ -0,0 +1,309 @@
...
+
+pub trait GetSpan<T> {
```
hawkw: I'm fine with merging this, but I think we should try to move toward replacing it with the upstream version from `tracing-tower`. I'd prefer not to maintain our own version in-tree. 

Would you mind [opening feature requests](https://github.com/tokio-rs/tracing/issues) on the `tracing` issue tracker for anything that prevented the use of `tracing-tower` here?

src/app/control.rs
olix0r: i suppose rustfmt has trouble with the proc-macro...

hawkw: it's actually just a `macro_rules!` macro; AFAICT rustfmt doesn't really touch anything inside macro invocations

src/trace.rs
olix0r: :+1: let's address in a followup

src/trace.rs
olix0r: I'm not sure if there's actually any feature gap... I just hit a compile error that was annoying to debug and threw this together instead ;p I agree we should move it over to the upstream lib though

lib/linkerd2-proxy-transport/src/source.rs
```diff
@@ -0,0 +1,42 @@
+use super::tls::PeerIdentity;
```
kleimkuhler: Will `Endpoint` follow a similar change?

src/app/errors.rs
```diff
@@ -124,12 +131,20 @@ fn map_err_to_5xx(e: Error) -> StatusCode {
     } else if let Some(_) = e.downcast_ref::<router::NotRecognized>() {
         error!("could not recognize request");
         http::StatusCode::BAD_GATEWAY
-    } else if let Some(err) = e.downcast_ref::<outbound::RequireIdentityError>() {
-        error!("{}", err);
-        http::StatusCode::FORBIDDEN
+    } else if let Some(err) = e.downcast_ref::<StatusError>() {
+        error!(%err.status, %err.message);
```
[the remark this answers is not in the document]
olix0r: wdym? Isn't this structured?

lib/linkerd2-proxy-transport/src/source.rs
olix0r: Probably not. there are independent types for inbound and outbound endpoints and I don't think they should be combined. I'd prefer for this to go the other way... Source should just be an impl detail and not be relied on broadly.

src/app/errors.rs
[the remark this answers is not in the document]
kleimkuhler: Ah never mind then. I'm misunderstanding the syntax in some of these macros.

kleimkuhler:
> @kleimkuhler
> 
> > My general feedback is it seems like moving towards structured messages would be good in most places
> 
> I think that would probably make the already big diff even bigger — for lines that aren't already changed in this PR, can we maybe do that as a follow-up instead?

Oh yes that is what I was implying by it not being a blocker! I don't think it needs to be part of the PR.

PR #376, opened 2019-10-16
ci: Run docker builds on linkerd-docker host

olix0r:
Replicate our integration test configuration to also run docker builds
on our remote host.

Furthermore, run these builds in unoptimized mode (since we don't
consume the results).

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -57,7 +57,7 @@ jobs:
       env:
         DOCKER_HOST: "ssh://linkerd-docker"
       run: |
-        export DOCKER_TAG="proxy-ci:$(dd bs=64 count=1 if=/dev/urandom status=none | tr -dc 'a-zA-Z0-9')""
+        export DOCKER_TAG="proxy-ci:$(dd bs=64 count=1 if=/dev/urandom status=none | tr -dc 'a-zA-Z0-9')"
         make docker
         docker image rm -f "$DOCKER_TAG"
 
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -51,12 +51,11 @@ jobs:
         # Confirm that the SSH configuration works.
         ssh linkerd-docker docker version
 
-    # Use the previously built image to run tests.
+    # Use the previously built image to run tests. The image will be left on
+    # the build host; but orphaned image layers are pruned regularly.
     - name: Docker (Origin)
       if: '!github.event.pull_request.head.repo.fork'
       env:
         DOCKER_HOST: "ssh://linkerd-docker"
-      run:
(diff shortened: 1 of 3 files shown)


.github/workflows/docker.yml
```diff
@@ -12,9 +12,50 @@ on:
...
+        # Confirm that the SSH configuration works.
+        ssh linkerd-docker docker version
+
+    # Use the previously built image to run tests. The image will be left on
+    # the build host; but orphaned image layers are pruned regularly.
+    - name: Docker (Origin)
+      if: '!github.event.pull_request.head.repo.fork'
+      env:
+        DOCKER_HOST: "ssh://linkerd-docker"
+      run: make docker
```
siggy: change up `Use the previously built image to run tests`? the previous step doesn't build anything. also did you intend to run tests here?

if the intent of `docker.yml` is to just confirm that docker builds work, consider adding a comment to the top of the file.

.github/workflows/docker.yml
```diff
@@ -12,9 +12,50 @@ on:
...
     steps:
     - uses: actions/checkout@v1
-    - run: make docker
+
+    - name: Docker (Fork)
+      if: github.event.pull_request.head.repo.fork
+      run: make docker
+
+    # Create a build image on a Linkerd build host.
+    - name: Setup (Origin)
```
siggy: nit: consider moving `Docker (Fork)` step next to `Docker (Origin)`. no functional difference just cognitively that's where it sits in step ordering.

.github/workflows/docker.yml
```diff
@@ -12,9 +15,52 @@ on:
...
+          echo "    IdentityFile ~/.ssh/id"
+          echo "    BatchMode yes"
+          echo "    ServerAliveInterval 60"
+          echo "    ServerAliveCountMax 5"
+        ) >~/.ssh/config
+        # Confirm that the SSH configuration works.
+        ssh linkerd-docker docker version
+
+    # Use the previously built image to run tests. The image will be left on
+    # the build host; but orphaned image layers are pruned regularly.
```
siggy: what previously built image is this referring to?

.github/workflows/docker.yml
```diff
@@ -12,9 +15,50 @@ on:
...
+        # Confirm that the SSH configuration works.
+        ssh linkerd-docker docker version
+
+    - name: Docker (Origin)
+      if: '!github.event.pull_request.head.repo.fork'
+      env:
+        DOCKER_HOST: "ssh://linkerd-docker"
+      run: |
+        export DOCKER_TAG="proxy-ci:$(dd bs=64 count=1 if=/dev/urandom status=none | tr -dc 'a-zA-Z0-9')"
+        make docker
```
hawkw: Since this docker image is discarded and we're just testing if it can be built, should we maybe set PROXY_UNOPTIMIZED here? Might make it run slightly faster...

olix0r: We set it on the env higher up in the file

PR #377, opened 2019-10-16
Update to Rust 1.38.0

olix0r:


olix0r:
> it looks like there were some whitespace changes, i'm assuming that's due to the rustfmt version bump?

yep

PR #378, opened 2019-10-18
build: only deny warnings on CI, not locally

hawkw:
Currently, all crates in the proxy repo are configured to always deny
all Rust compiler warnings. This means that if the compiler reports any
warnings, the build will fail. This can be an issue when changes are
made locally that introduce warnings (i.e. unused code is added, imports
are no longer needed, etc) and the developer wishes to test an
incomplete state to ensure their changes are on the right path. With
warnings denied, tests will not run if the project contains any
warnings, so developers must either fix all warnings or disable the deny
attribute. Disabling the deny attribute when making changes locally has
become a fairly common practice, but it's error-prone: sometimes, the
deny attribute is commented out and then accidentally committed.

This branch is a proposal for a new way to handle Rust warnings. Rather
than always failing the build if any warnings are reported, we now
export an env var on CI that configures the compiler to deny warnings.
Local in-development builds will report warnings, but not *fail* the
build, making it easier to run tests for incomplete changes, but code
with warnings won't be allowed to merge to master. Hopefully, this
obsoletes the practice of commenting out `#![deny(warnings)]` when
making changes.

### DIFF
--- a/.github/actions/remote/Dockerfile
+++ b/.github/actions/remote/Dockerfile
@@ -4,4 +4,5 @@ FROM $RUST_IMAGE as build
 WORKDIR /usr/src/linkerd2-proxy
 COPY . ./
 RUN ["make", "fetch"]
+ENV RUSTFLAGS="-D warnings"
 ENTRYPOINT ["make"]
--- a/.github/actions/remote/Dockerfile
+++ b/.github/actions/remote/Dockerfile
@@ -4,5 +4,4 @@ FROM $RUST_IMAGE as build
 WORKDIR /usr/src/linkerd2-proxy
 COPY . ./
 RUN ["make", "fetch"]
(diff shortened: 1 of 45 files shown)


hawkw:
> I'm into this... but what do you think about setting this in the `Makefile` instead of in the CI configs?
> 
> I like the makefile to be "how ci will run" so i can replicate easily. When i'm testing during development, I run cargo directly (on a narrower crate).

I can do that. I thought that the makefile also might be used for local testing, but that makes sense!

olix0r:
I want to figure out if we can get this to apply for all cargo commands in the makefile....

Perhaps we can do something like `CARGO = env ... cargo` and then `CARGO_TEST = $(CARGO) ...` etc?

olix0r:
Let's revisit this post-migration if necessary.

PR #379, opened 2019-10-21
Remove the LINKERD2_PROXY_INBOUND_FORWARD config

olix0r:
The `LINKERD2_PROXY_INBOUND_FORWARD` configuration was implemented to support local testing; but the proxy has evolved to the point where it
needs a proper runtime harness (as in docker or the integration test
infrastructure) to run.

This option is just a bit of debt that should be removed to simplify 
code/diagnostics.

This also disables doctests in the opencensus crate to prevent spurious failures.

Furthermore, the proxy's initial logging has been clarified somewhat.

Closes linkerd/linkerd2#3609

### DIFF
--- a/linkerd/app/src/lib.rs
+++ b/linkerd/app/src/lib.rs
@@ -203,13 +203,13 @@ where
             Conditional::Some(config) => info!("using identity service at {:?}", config.svc.addr),
             Conditional::None(reason) => info!("identity is DISABLED: {}", reason),
         }
-        info!("admin on {}", admin_listener.local_addr(),);
+        info!("admin on {}", admin_listener.local_addr());
         info!(
             "tap on {:?}",
             control_listener.as_ref().map(|l| l.0.local_addr())
         );
         info!("outbound on {:?}", outbound_listener.local_addr());
-        info!("inbound on {}", inbound_listener.local_addr(),);
+        info!("inbound on {}", inbound_listener.local_addr());
         info!(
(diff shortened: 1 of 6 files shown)


linkerd/app/src/lib.rs
```diff
@@ -203,16 +203,13 @@ where
             Conditional::Some(config) => info!("using identity service at {:?}", config.svc.addr),
             Conditional::None(reason) => info!("identity is DISABLED: {}", reason),
         }
-        info!("routing on {:?}", outbound_listener.local_addr());
+        info!("admin on {}", admin_listener.local_addr(),);
```
kleimkuhler: nit: Odd that a comma was added here?

linkerd/app/src/lib.rs
```diff
@@ -203,16 +203,13 @@ where
...
-            config.inbound_forward
-        );
-        info!(
-            "serving admin endpoint metrics on {:?}",
-            admin_listener.local_addr(),
+            "tap on {:?}",
+            control_listener.as_ref().map(|l| l.0.local_addr())
         );
+        info!("outbound on {:?}", outbound_listener.local_addr());
+        info!("inbound on {}", inbound_listener.local_addr(),);
```
kleimkuhler: Here as well

olix0r: copypasta. fmt doesn't hit the tracing macros, apparently

PR #380, opened 2019-10-21
Remove the linkerd2-task crate

olix0r:
Most of the logic in `linkerd2-task` is not actually used.

This change moves the used part of this crate to `task-test-util`, and
removes the rest of the linkerd-specific task infrastructure in favor of
using tokio directly.


linkerd/app/src/lib.rs
```diff
@@ -155,28 +149,24 @@ where
...
+
+                    proxy_parts.build_proxy_task(drain_rx);
+                    trace!("main task spawned");
+
+                    shutdown_signal.and_then(move |()| {
+                        debug!("shutdown signaled");
+                        drain_tx.drain().map(|()| debug!("shutdown complete"))
+                    })
+                })
+                .instrument(info_span!("main")),
```
olix0r: shouldn't do this...

hawkw: ?

olix0r: i don't think i actually want `main:` prefixed on all of the log messages

linkerd/task-test-util/Cargo.toml
```diff
@@ -1,17 +1,10 @@
 [package]
-name = "linkerd2-task"
+name = "linkerd2-task-test-util"
```
hawkw: WDYT about just making this a general `test-util` crate? We might want to put some other stuff that's currently in the integration test crate in there as well, in the future, if it's used in unit tests as well...

linkerd/app/src/lib.rs
hawkw: ah, yeah. probably get rid of this...

PR #381, opened 2019-10-22
Record transport metrics after protocol detection

olix0r:
In order to include the detected protocol in transport metric labels,
protocol detection must be separated from dispatching a transport to the
HTTP server (i.e via `app::core::proxy::Server`).

This change accomplishes this as follows:

* The `io` type is split out of `proxy-transport`. It includes the core
  `BoxedIo`, `PrefixedIo`, and `Peek` types.
* The `transport::listen` module is now responsible for capturing all of
  an accepted socket's address info into an `Addrs` type. It, rather
  than the `tls::accept` module captures the original dst addr.
* The `Bind` type/trait is introduced to represent listener endpoint
  before it is bound.
* The `transport::AddrInfo` trait is removed. It's no longer necessary
  since addresses are determined predictably in the stack.
* The `tls::Connection` type is replaced by a `Peek`/`PrefixedIo`.
* The `detect` crate abstracts protocol detection using
  `Peek`/`PrefixedIo`. This is used to implement HTTP detection (in
  `Server`); and in the future it will be used for TLS/SNI detection.

Part of linkerd/linkerd2#3523


linkerd/app/core/src/proxy/server.rs
```diff
@@ -137,30 +164,32 @@ where
...
-            transport_metrics,
-            connect,
-            make_http,
-            drain,
-        }
+        skip_ports: Arc<IndexSet<u16>>,
+    ) -> detect::Accept<ProtocolDetect, Self> {
+        detect::Accept::new(
+            ProtocolDetect { skip_ports },
+            Self {
```
olix0r: This is really the whole crux of this PR.

linkerd/app/inbound/src/lib.rs
```diff
@@ -267,10 +276,10 @@ impl transport::metrics::TransportLabels<Endpoint> for TransportLabels {
...
 }
 
-impl transport::metrics::TransportLabels<Source> for TransportLabels {
+impl transport::metrics::TransportLabels<ServerProtocol> for TransportLabels {
     type Labels = transport::labels::Key;
 
-    fn transport_labels(&self, source: &Source) -> Self::Labels {
-        transport::labels::Key::accept("inbound", source.tls_peer.as_ref())
+    fn transport_labels(&self, proto: &ServerProtocol) -> Self::Labels {
+        transport::labels::Key::accept("inbound", proto.tls.peer_identity.as_ref())
```
olix0r: Note: this will change to encode additional metadata, including detected protocol and SNI names

linkerd/app/integration/tests/telemetry.rs
```diff
@@ -751,7 +751,7 @@ mod transport {
         drop(client);
         assert_eventually_contains!(
             metrics.get("/metrics"),
-            "tcp_close_total{direction=\"inbound\",peer=\"src\",tls=\"disabled\",errno=\"\"} 1"
+            "tcp_close_total{direction=\"inbound\",peer=\"src\",tls=\"disabled\",errno=\"ENOTCONN\"} 1"
```
olix0r: This appears to be the main functional change of moving the metrics layer.

linkerd/app/src/config.rs
```diff
@@ -0,0 +1,1058 @@
...
+        Ok(Config {
+            admin,
+            dns,
+            destination,
+            tap,
+            trace_collector,
+            identity,
+            outbound,
+            inbound,
+        })
```
olix0r: this is the part of the file that changed (other than removing unused env variables).

linkerd/io/src/peek.rs
```diff
@@ -0,0 +1,48 @@
...
+    where
+        Self: Sized + Future,
+    {
+        let buf = BytesMut::with_capacity(capacity);
+        Peek(Some(Inner { buf, io }))
+    }
+}
+
+impl<T: AsyncRead + AsyncWrite> Future for Peek<T> {
+    type Item = PrefixedIo<T>;
```
olix0r: Note that `Peek` has changed to be a concrete type that wraps an `Io` to produce a `PrefixedIo`

linkerd/proxy/detect/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    /// If the target can be determined by the target alone (i.e. because it's
+    /// known to be a server-speaks-first target), Otherwise, the target is
+    /// returned as an error.
+    fn detect_before_peek(&self, target: T) -> Result<Self::Target, T>;
+
+    /// If the target could not be determined without peeking, then used the
+    /// peeked prefix to determine the protocol.
+    fn detect_peeked_prefix(&self, target: T, prefix: &[u8]) -> Self::Target;
+}
```
olix0r: This is a generalization of the detection performed by the proxy Server previously. It can probably be used for TLS detection, also.

linkerd/proxy/http/src/retry.rs
```diff
@@ -187,6 +187,7 @@ where
     }
 }
 
+// TODO this needs to be moved up into the application!
```
olix0r: We don't want to encode all of the possible extensions in this module. In a followup, the retry layer should take a request-cloning strategy, rather than rely on the `Request` type implementing `TryClone` itself...

olix0r:
This proxy is testable via `config.linkerd.io/proxy-version: ver-transport-metrics-after-protocol.1`

olix0r:
Part of linkerd/linkerd2#3523

linkerd/app/core/src/config.rs
```diff
@@ -1,1137 +1,115 @@
...
-}
-
-/// Configuration settings for binding a listener.
-///
-/// TODO: Rename this to be more inline with the actual types.
 #[derive(Clone, Debug)]
-pub struct Listener {
-    /// The address to which the listener should bind.
-    pub addr: SocketAddr,
+pub struct ServerConfig<O: OrigDstAddr = NoOrigDstAddr> {
```
hawkw: I'm _very_ pleased to see that this refactor has finally happened!

It does seem like an enormous diff that could have potentially gone in its own branch, but if that's more work than having it in this PR, don't feel pressured to factor it out...

linkerd/app/core/src/config.rs
```diff
@@ -1,1137 +1,115 @@
...
-        })),
-        (Some(_), None) => {
-            error!("{} must be specified when {} is set", n_env, a_env);
-            Err(Error::InvalidEnvVar)
-        }
-        (None, Some(_)) => {
-            error!("{} must be specified when {} is set", a_env, n_env);
-            Err(Error::InvalidEnvVar)
+impl<O: OrigDstAddr> ServerConfig<O> {
+    pub fn with_orig_dst_addrs_from<P: OrigDstAddr>(self, orig_dst_addrs: P) -> ServerConfig<P> {
```
hawkw: TIOLI: this name feels pretty wordy and `with_orig_dst` seems equally descriptive to me? Not important to me though.

linkerd/app/core/src/proxy/server.rs
```diff
@@ -92,36 +127,28 @@ where
...
 }
 
-impl error::Error for NoOriginalDst {}
+impl error::Error for NoForwardTarget {}
 
-impl fmt::Display for NoOriginalDst {
+impl fmt::Display for NoForwardTarget {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "Missing SO_ORIGINAL_DST address")
+        write!(f, "Could not forward to a target address")
```
hawkw: Looks like this is only constructed if the target address is local, it might be good if the message indicated that for logs etc.

linkerd/app/core/src/proxy/server.rs
```diff
@@ -137,30 +164,32 @@ where
...
+                connect: ForwardConnect(connect),
+                make_http,
+                drain,
+            },
+        )
     }
 }
 
-impl<L, T, C, H, B> Service<Connection> for Server<L, T, C, H, B>
+impl<L, C, H, B> Service<(Protocol, BoxedIo)> for Server<L, C, H, B>
```
hawkw: AFAICT this can still be `Connection`, right?

linkerd/app/core/src/proxy/server.rs
```diff
@@ -187,103 +216,75 @@ where
...
-                                .map(|_| ())
-                                .map_err(Into::into)
-                        }),
-                    })
-            }),
-        });
-
-        Box::new(serve_fut)
+            HttpVersion::H2 => {
+                let exec = tokio::executor::DefaultExecutor::current().instrument(info_span!("h2"));
```
hawkw: Nit/TIOLI: this span is called `h2` while the HTTP 1 span is called `http1`, maybe they should be consistent with each other?
```suggestion
                let exec = tokio::executor::DefaultExecutor::current().instrument(info_span!("http2"));
```

linkerd/app/inbound/src/lib.rs
```diff
@@ -227,28 +232,33 @@ pub fn spawn<P>(
...
         }))
         .push(errors::layer())
-        .push(trace::layer(
-            |src: &Source| info_span!("source", peer_ip = %src.remote.ip(), peer_id=?src.tls_peer),
-        ))
+        .push(trace::layer(|src: &tls::accept::Meta| {
+            info_span!(
+                "source",
+                peer.ip = %src.addrs.peer().ip(),
+                peer.id = ?src.peer_identity,
```
hawkw: Nit/TIOLI: to me, `peer.id` _kind of_ implies an internal numeric identifier, while `peer.identity` suggests a cryptographic identity? but on the other hand, it's more characters in the logs, so maybe this is better...

linkerd/app/inbound/src/lib.rs
```diff
@@ -227,28 +232,33 @@ pub fn spawn<P>(
...
+                target.addr = %src.addrs.target_addr(),
```
hawkw: Nit/TIOLI: i notice we have just the peer's IP, but the target's `addr`, which includes both the IP and port; should these be formatted more consistently (e.g. `{ peer.addr, target.addr }` or `{ peer.ip, peer.port, target.ip, target.port }`)?

linkerd/app/integration/src/proxy.rs
```diff
@@ -147,41 +147,43 @@ struct DstInner {
...
+fn run(proxy: Proxy, mut env: TestEnv) -> Listening {
     let controller = proxy.controller.unwrap_or_else(|| controller::new().run());
     let inbound = proxy.inbound;
     let outbound = proxy.outbound;
     let identity = proxy.identity;
     let mut mock_orig_dst = DstInner::default();
 
     env.put(
-        app::config::ENV_DESTINATION_SVC_ADDR,
+        "LINKERD2_PROXY_DESTINATION_SVC_ADDR",
```
hawkw: nit: it would be nice if this could still use a const...i won't block the PR over that, though...

linkerd/app/integration/tests/telemetry.rs
hawkw: Seems good to me!

linkerd/app/src/config.rs
```diff
@@ -0,0 +1,1058 @@
...
+}
+
+impl Config {
+    pub fn try_from_env() -> Result<Self, EnvError> {
+        Self::parse(&Env)
+    }
+
+    /// Load a `Config` by reading ENV variables.
+    pub fn parse<S: Strings>(strings: &S) -> Result<Self, EnvError> {
+        // Parse all the environment variables. `parse` will log any errors so
```
hawkw: Not a blocker, really, but it seems to me like this massive function out to be split into smaller `parse(strings: &S)` methods on the various `FooConfig` structs?

linkerd/app/src/config.rs
hawkw: by "this", do you mean the `parse` method, or everything below the parse method?

linkerd/dns/src/lib.rs
```diff
@@ -60,15 +62,12 @@ impl Resolver {
     /// NOTE: It would be nice to be able to return a named type rather than
     ///       `impl Future` for the background future; it would be called
     ///       `Background` or `ResolverBackground` if that were possible.
```
hawkw: I think this comment can be removed now that we have a named return type rather than `impl Future`?

linkerd/proxy/detect/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+
+#[derive(Debug, Clone)]
+pub struct Accept<D, A> {
+    detect: D,
+    accept: A,
+    peek_capacity: usize,
+}
+
+pub enum AcceptFuture<T, D: Detect<T>, A: core::listen::Accept<(D::Target, BoxedIo)>> {
```
hawkw: Nit: could these bounds maybe go in a `where` clause? this is a lot for one line...

linkerd/proxy/detect/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+use linkerd2_proxy_core as core;
+
+/// A strategy for detectin values out of a client transport.
+pub trait Detect<T> {
+    type Target;
+
+    /// If the target can be determined by the target alone (i.e. because it's
+    /// known to be a server-speaks-first target), Otherwise, the target is
+    /// returned as an error.
+    fn detect_before_peek(&self, target: T) -> Result<Self::Target, T>;
```
hawkw: Hmm, I'm a little bit on the fence about using `Result` to indicate incomplete rather than erroneous states, but the `Conditional` type has kind of scared me out of defining new enums for cases that can reasonably use `Result`s...this is probably the best approach...

linkerd/proxy/http/src/h1.rs
```diff
@@ -19,23 +20,27 @@ pub fn normalize_our_view_of_uri<B>(req: &mut http::Request<B>) {
...
+    if let Some(addr) = req
         .extensions()
-        .get::<Source>()
-        .and_then(|ctx| ctx.orig_dst_if_not_local());
-
-    if let Some(orig_dst) = orig_dst {
+        .get::<tls::accept::Meta>()
+        .map(|s| s.addrs.target_addr())
+    {
+        trace!(target.addr = %addr, "using target address");
```
hawkw: nit: should this say
```suggestion
        trace!(target.addr = %addr, "using original dst target address");
```
or something?

linkerd/proxy/transport/src/listen.rs
```diff
@@ -1,39 +1,111 @@
...
+        Bind {
+            orig_dst_addr,
+            bind_addr: self.bind_addr,
+            keepalive: self.keepalive,
+        }
     }
 
-    pub fn local_addr(&self) -> SocketAddr {
-        self.local_addr
+    pub fn with_orig_dst_addrs_from_system(self) -> Bind<SysOrigDstAddr> {
```
hawkw: nit: should this maybe fail fast (panic?) if not on Linux?

linkerd/proxy/transport/src/tls/client.rs
```diff
@@ -101,36 +101,25 @@ where
...
                 ConnectFuture::Init { future, tls } => {
                     let io = try_ready!(future.poll());
-                    let remote_addr = io.remote_addr().expect("socket must have remote addr");
 
                     match tls {
-                        Conditional::Some((server_name, local_tls)) => {
-                            trace!("initiating TLS to {}", server_name.as_ref());
-                            let future =
+                        Conditional::Some((peer_identity, local_tls)) => {
+                            trace!(sni = %peer_identity.as_ref(), "initiating TLS");
```
hawkw: nit/tioli: should this be 
```suggestion
                            trace!(peer.sni = %peer_identity.as_ref(), "initiating TLS");
```

linkerd2-proxy/src/main.rs
```diff
@@ -5,18 +5,27 @@
...
-    let (config, trace_admin) = match linkerd2_app::init() {
-        Ok(c) => c,
+    // Load configuration from the environment without binding ports.
+    let main = match Main::try_from_env() {
+        Ok(main) => main,
         Err(e) => {
-            eprintln!("configuration error: {:#?}", e);
-            std::process::exit(64)
+            eprintln!("Configuration error: {}", e);
+            std::process::exit(64) // EX_USAGE
```
hawkw: Nit/TIOLI: if you add a const, this could be `std::process:exit(EX_USAGE);` and the comment wouldn't be necessary...not a big deal though.

linkerd/io/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// This trait is private, since it's purpose is for creating a dynamic
+    /// trait object, but doing so without care can lead not getting vectored
+    /// writes.
+    ///
+    /// Instead, use the concrete `BoxedIo` type.
+    pub trait Io: AsyncRead + AsyncWrite + Send {
+        fn shutdown_write(&mut self) -> Result<()>;
+
+        /// This method is to allow using `Async::write_buf` even through a
```
adleong: I don't understand this song and dance.  Where is the non-vectorized implementation of write_buf?  Is it the default implementation on the `AsyncWrite` trait?  why would a call to `write_buf` on a `Box<dyn AsyncWrite>` dispatch to the default implementation rather than the overriding implementation?

olix0r: > Is it the default implementation on the AsyncWrite trait?

Yep. If an impl doesn't override it, perf suffers _severely_. This isn't new code, fwiw.

linkerd/proxy/detect/src/lib.rs
olix0r: And it's not _wrong_, here, really. It's like how sender uses Result to return the item if the receiver is gone.

linkerd/app/core/src/proxy/server.rs
olix0r: This is consistent with how the protocols are referred to in RFCs.

linkerd/app/core/src/config.rs
olix0r: I began to try to split things apart but it's extremely annoying, as it touches changes in transport, which touches changes in io... I can give it another shot, because I'd like to changes to be smaller, but I'm not all that optimistic.

linkerd/app/inbound/src/lib.rs
olix0r: We don't care about the peer's port because it's an ephemeral client port. Just as the inbound omits all of the peer information (because it's always localhost, and an ephemeral port.).  The logs need to be useful to humans, not machines.

linkerd/app/integration/src/proxy.rs
olix0r: this const existed only for this usage, and was redundant/decoupled with the templatized consts that actually control how configs are parsed. in light of all that, 0 value in having a const, imo

linkerd/proxy/transport/src/listen.rs
```diff
@@ -1,39 +1,111 @@
...
+
+#[derive(Clone, Debug)]
+pub struct Bind<O: OrigDstAddr = NoOrigDstAddr> {
+    bind_addr: SocketAddr,
+    keepalive: Option<Duration>,
+    orig_dst_addr: O,
+}
+
+#[derive(Debug)]
+pub struct Listen<O: OrigDstAddr = NoOrigDstAddr> {
```
kleimkuhler: I feel like the naming for this struct is better suited for a trait. How about `Listener`?

linkerd/proxy/transport/src/listen.rs
```diff
@@ -1,39 +1,111 @@
...
+            keepalive: self.keepalive,
+            orig_dst_addr: self.orig_dst_addr,
+            state: State::Init(Some(tcp)),
+        })
     }
 }
 
-impl linkerd2_proxy_core::listen::Listen for Listen {
-    type Connection = (tokio::net::TcpStream, SocketAddr);
+impl<O> listen::Listen for Listen<O>
```
kleimkuhler: I.e. implementing `Listen` for `Listen` reads a little weird.

linkerd/proxy/detect/src/lib.rs
hawkw: yeah, i think this is probably the best thing here, i just had to think about it for a minute.

linkerd/proxy/transport/src/tls/accept.rs
```diff
@@ -204,22 +202,24 @@ impl<A: Accept<Connection>> Future for AcceptFuture<A> {
...
-                    let AcceptMeta {
-                        accept,
-                        remote_addr,
-                        orig_dst_addr,
-                    } = meta.take().expect("polled after complete");
-                    let conn = Connection::tls(BoxedIo::new(io), remote_addr, client_id)
-                        .with_original_dst(orig_dst_addr);
-                    AcceptFuture::ReadyAccept(accept, Some(conn))
+                    let AcceptMeta { accept, addrs } = meta.take().expect("polled after complete");
+                    // FIXME the connection doesn't know about TLS connections
```
kleimkuhler: This is cool to see that this is a possibility now though!

linkerd/proxy/transport/src/listen.rs
olix0r: This is pretty idiomatic in this codebase (we don't have any Connector types, for instance). I'm not opposed to changing the idiom, but don't want to do so ad-hoc.

kleimkuhler: Hm okay good point. No problem

olix0r:
I'm going to resubmit this without most of the application/config changes. Then I'll followup with a branch that includes the config changes as well as some other application cleanup (including logging improvements).

olix0r:
I've backed out most of the unneeded changes to app. I'll follow up with a branch that addresses configuration, logging contexts, etc.

linkerd/proxy/transport/src/tls/client.rs
olix0r: actually, this is all redundant with value already in the context...

linkerd/proxy/http/src/h1.rs
olix0r: I'm trying to set up most of the uses around accept::Meta/listen::Addrs to be agnostic about whether the orig dst is set (it's kind of an impl detail) to support serving local services from the proxy directly

linkerd/app/src/config.rs
olix0r: that's a yak for another day

linkerd/app/core/src/proxy/server.rs
kleimkuhler: I found the return value of this to be a little surprising, because it returns an `detect::Accept` and then `server` is passed into `TlsAccept::new` (on inbound). The trait bound for `accept` in `TlsAccept::new` is that it implements `listen::Accept`, but `detect::Accept` only implements `Service`. Then I remembered the blanket `Accept` impl for all `Services`s that meet fairly simple bounds which is the case here.

All this to say I'm just writing this down now for my understanding, but I did not find it to be immediately obvious.

linkerd/app/core/src/tap/accept.rs
```diff
@@ -2,32 +2,33 @@ use crate::identity;
...
 use futures::{Future, Poll};
 use indexmap::IndexSet;
 use linkerd2_error::Error;
 use linkerd2_proxy_api::tap::server::{Tap, TapServer};
+use std::sync::Arc;
 
 #[derive(Clone, Debug)]
 pub struct AcceptPermittedClients {
-    permitted_client_ids: IndexSet<identity::Name>,
+    permitted_client_ids: Arc<IndexSet<identity::Name>>,
```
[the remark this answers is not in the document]
olix0r: Given the way that detection works, Accept is cloned for each connection. This prevents having to clone the set itself.

PR #383, opened 2019-10-24
Eliminate tower::util::Either from add_header

olix0r:
I have a hypothesis that layers with `Either` services are especially costly for our compile-times.

This change removes one such usage in the `add_header` module, which is used several times in each stack.


olix0r:
Hypothesis does not hold

PR #386, opened 2019-10-31
internal: Move suffix-filtering into profiles API

olix0r:
Currently, the profile_suffixes are wired through into the inbound and
outbound proxies to configure the profile layer.

This change decouples inbound/outbound proxy initialization from this
detail of the profile client.

### DIFF
--- a/linkerd/app/core/src/profiles.rs
+++ b/linkerd/app/core/src/profiles.rs
@@ -1,3 +1,4 @@
+use crate::dns;
 use crate::proxy::http::{profiles, retry::Budget};
 use futures::sync::{mpsc, oneshot};
 use futures::{Async, AsyncSink, Future, Poll, Sink, Stream};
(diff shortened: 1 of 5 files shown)


linkerd/app/core/src/profiles.rs
```diff
@@ -79,11 +86,16 @@ where
     type Stream = Rx;
 
     fn get_routes(&self, dst: &NameAddr) -> Option<Self::Stream> {
-        let (tx, rx) = mpsc::channel(1);
+        if !self.suffixes.iter().any(|s| s.contains(dst.name())) {
+            debug!("name not in profile suffixes");
```
hawkw: should this include what the name was, or is this happening inside a span that already includes that?

olix0r: it's safe to assume all of that metadata will be on the context if it needs to be.

linkerd/app/core/src/profiles.rs
```diff
@@ -79,11 +86,16 @@ where
     type Stream = Rx;
 
     fn get_routes(&self, dst: &NameAddr) -> Option<Self::Stream> {
-        let (tx, rx) = mpsc::channel(1);
+        if !self.suffixes.iter().any(|s| s.contains(dst.name())) {
```
adleong: I know this is just a move from elsewhere, but I'm confused by this logic.  We check that the suffix contains the name?  Isn't that backwards?

olix0r: we can change the word "contains", but a domain suffix contains its subdomains. they are constituents.

PR #387, opened 2019-10-31
internal: Split tap out of app-core

olix0r:
Tap doesn't have any dependencies on the app infrastructure, so split it
into its own proxy subcrate.

Also, reexport opencensus::proto to streamline dependencies.


linkerd/proxy/tap/src/grpc/server.rs
```diff
@@ -570,12 +572,15 @@ fn base_event<B, I: Inspect>(req: &http::Request<B>, inspect: &I) -> api::TapEve
...
-                    .insert("server_id".to_owned(), id.as_ref().to_owned());
+            match inspect.dst_tls(req) {
+                Conditional::None(reason) => {
+                    m.labels.insert("tls".to_owned(), reason.to_string());
+                }
+                Conditional::Some(id) => {
+                    m.labels.insert("tls".to_owned(), "true".to_owned());
+                    m.labels
+                        .insert("server_id".to_owned(), id.as_ref().to_owned());
+                }
```
olix0r: Note: We've decoupled tap from metric labels (which is probably a good thing), so we replicate the prior tap-labeling behavior here.

linkerd/proxy/tap/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "linkerd2-proxy-tap"
+version = "0.1.0"
+authors = ["Linkerd Developers <cncf-linkerd-dev@lists.cncf.io>"]
+edition = "2018"
+publish = false
+
+[dependencies]
```
hawkw: I'm assuming you've checked that tap requires all of these deps?

PR #388, opened 2019-11-04
Resolve private IP addresses via the destination service

olix0r:
This enables discovery (and mTLS) for HTTP traffic to private-network IP addresses, on the assumption that they are more than likely part of the local cluster. The set of discoverable networks may be configured via the environment.

This depends on https://github.com/linkerd/linkerd2/pull/3595 returning resolutions for addresses.

A proxy version is published for testing at `gcr.io/linkerd-io/proxy:ver-2019-11-04.0`


linkerd/app/src/dst/mod.rs
```diff
@@ -11,8 +12,9 @@ use tower_grpc::{generic::client::GrpcService, Body, BoxBody};
 pub struct Config {
     pub control: ControlConfig,
     pub context: String,
-    pub get_suffixes: Vec<dns::Suffix>,
-    pub profile_suffixes: Vec<dns::Suffix>,
+    pub get_suffixes: IndexSet<dns::Suffix>,
+    pub get_networks: IndexSet<ipnet::IpNet>,
+    pub profile_suffixes: IndexSet<dns::Suffix>,
```
hawkw: TIOLI: might be worth having doc comments explaining what all these similarly-named fields actually _do_?

linkerd/app/src/dst/resolve.rs
```diff
@@ -1,22 +1,24 @@
...
 use std::sync::Arc;
 use tower_grpc::{generic::client::GrpcService, Body, BoxBody, Code, Status};
 
 pub type Resolve<S> = request_filter::Service<
-    PermitNamesInSuffixes,
+    PermitConfiguredDsts,
     recover::Resolve<BackoffUnlessInvalidArgument, api::Resolve<S>>,
 >;
 
 pub fn new<S>(
```
hawkw: Nit/TIOLI: this is kind of a lot of arguments...I know this method isn't called very often, but the callsite might be easier to read if there was a builder, or at least a struct with a bunch of named fields? maybe it's not worth it...

linkerd/app/src/env.rs
```diff
@@ -86,6 +87,17 @@ pub const ENV_OUTBOUND_MAX_IN_FLIGHT: &str = "LINKERD2_PROXY_OUTBOUND_MAX_IN_FLI
...
 
+/// Constrains which destination addresses are resolved through the destination
+/// service.
+///
+/// The value is a comma-separated list of networks that may be
+/// resolved via the destination service.
+///
+/// If specified and empty, the destination service is not used for resolution.
+///
+/// If unspecified, a default value is used
```
hawkw: Nit:
```suggestion
/// If unspecified, a default value is used.
```

PR #389, opened 2019-11-05
internal: Decouple TCP forwarding from protocol dispatch

olix0r:
In preparation of adding discovery logic to outbound TCP forwarding,
this change extracts the TCP forwarding logic as an `Accept` so that the
outbound proxy will be able to provide an alternate implementation.


linkerd/app/inbound/src/endpoint.rs
```diff
@@ -123,7 +123,7 @@ impl<A> router::Recognize<http::Request<A>> for RecognizeEndpoint {
     fn recognize(&self, req: &http::Request<A>) -> Option<Self::Target> {
         let src = req.extensions().get::<tls::accept::Meta>();
         debug!("inbound endpoint: src={:?}", src);
-        let addr = src.and_then(|s| s.addrs.target_addr_if_not_local())?;
+        let addr = src.and_then(|m| m.addrs.target_addr_if_not_local())?;
```
hawkw: ?

linkerd/proxy/core/src/listen.rs
```diff
@@ -42,8 +42,19 @@ pub trait Accept<C> {
     fn poll_ready(&mut self) -> Poll<(), Self::Error>;
 
     fn accept(&mut self, connection: C) -> Self::Future;
+
+    #[inline]
```
hawkw: the compiler is probably smart enough to inline this for you :)

linkerd/proxy/detect/src/lib.rs
```diff
@@ -3,7 +3,7 @@ use linkerd2_error::Error;
 use linkerd2_io::{BoxedIo, Peek};
 use linkerd2_proxy_core as core;
 
-/// A strategy for detecting values out of a client transport.
+/// A strategy for detectin values out of a client transport.
```
hawkw: Nit: 
```suggestion
/// A strategy for detecting values out of a client transport.
```
also, didn't this get fixed in an earlier PR? perhaps a rebase is needed?

PR #392, opened 2019-11-09
Update tower-balance

olix0r:
`tower-balance` has been refactored and updated to use a new version of
the `rand` crate.

This change udpates the project to use the latest tower-balance, rand,
and quickcheck dependencies

I've soak-tested this over the weekend and the new balancer
implementation is at least as good as the prior version in terms of
latency and CPU usage.

### DIFF
--- a/linkerd2-proxy/src/main.rs
+++ b/linkerd2-proxy/src/main.rs
@@ -7,6 +7,7 @@
 use futures::{future, Future};
 use linkerd2_app::{trace, Config};
 use linkerd2_signal as signal;
+pub use tracing::{debug, error, info, warn};
 
 fn main() {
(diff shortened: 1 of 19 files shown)


linkerd2-proxy/src/main.rs
```diff
@@ -7,6 +7,7 @@
 use futures::{future, Future};
 use linkerd2_app::{trace, Config};
 use linkerd2_signal as signal;
+pub use tracing::{debug, error, info, warn};
```
olix0r: the compiler decided this was necessary. who am i to argue with a compiler

olix0r:
The soak test I used is available at https://gist.github.com/olix0r/f4c64ca5b2e64067ac339a704f64dce2

PR #393, opened 2019-11-13
Make fallback HTTP-agnostic

olix0r:
The fallback module is currently HTTP-aware, as it must make response
body types uniform.

This change modifies fallback to instead require that both service
types are identical. This is accomplished by boxing the services, using
a new http boxed service type.

This has the side-effect of dramatically improving compilation times for
the outbound proxy.


olix0r:
master:
```
cargo build -p linkerd2-app-outbound  209.80s user 1.03s system 104% cpu 3:22.24 total
```

ver/agnostic-fallback
```
cargo build -p linkerd2-app-outbound  175.25s user 0.98s system 105% cpu 2:47.59 total
```

linkerd/fallback/src/lib.rs
```diff
@@ -174,132 +167,9 @@ where
                 // We've called the fallback service and are waiting for its
                 // future to complete.
                 FallbackState::Fallback(ref mut f) => {
-                    return f.poll().map(|a| a.map(Either::B)).map_err(Into::into)
+                    return f.poll().map(|a| a.map(Into::into)).map_err(Into::into);
```
olix0r: probably not needed

olix0r: it is

olix0r:
Additional boxing appears to shave compile time even more; but we'll keep that option in our back pocket for now.

PR #394, opened 2019-11-13
Make the router HTTP-agnostic

olix0r:
There's nothing HTTP-specific about the router.

This change moves the layer implementations--which unnecessarily enforce
HTTP-specific type requirements--into the generic router crate, relaxing
its type constraints.

### DIFF
--- a/linkerd/router/src/layer.rs
+++ b/linkerd/router/src/layer.rs
@@ -1,8 +1,6 @@
+use crate::{Recognize, Router};
 use futures::Poll;
-use http;
 use linkerd2_error::{Error, Never};
-use linkerd2_router as rt;
-pub use linkerd2_router::{error, Recognize, Router};
 use std::marker::PhantomData;
 use std::time::Duration;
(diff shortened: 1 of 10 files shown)


linkerd/router/src/layer.rs
```diff
@@ -56,29 +54,30 @@ impl Config {
...
-    Layer {
-        config,
-        recognize,
-        _p: PhantomData,
+    pub fn new(config: Config, recognize: Rec) -> Self {
+        Self {
+            config,
+            recognize,
+            _p: PhantomData,
+        }
```
hawkw: what motivated the change from `router::layer` to `router::Layer::new`? seems like the diff would've been smaller without this?

not a big deal, just curious.

olix0r: Eh, it just gets awkward since it would actually be `router::layer::layer` or it would overload the name or something... As we split things out I think it just generally makes sense to adopt tower best practices.

PR #395, opened 2019-11-18
Classify some gRPC status codes as non-errors

tegioz:
Linkerd classifies all gRPC status codes except `OK` as errors. This can negatively affect a gRPC server's success rate, even when it is only returning things like `NOT_FOUND` or `INVALID_ARGUMENT`.

This change narrows down the list of gRPC status codes that are considered an error to:

* UNKNOWN (2)
* DEADLINE_EXCEEDED (4)
* INTERNAL (13)
* UNAVAILABLE (14)
* DATA_LOSS (15)

Please see [this linkerd2 issue](https://github.com/linkerd/linkerd2/issues/3729) for more details.

Signed-off-by: Sergio Castaño Arteaga <tegioz@icloud.com>

### DIFF
--- a/linkerd/app/core/src/classify.rs
+++ b/linkerd/app/core/src/classify.rs
@@ -3,6 +3,7 @@ use linkerd2_error::Error;
 pub use linkerd2_proxy_http::metrics::classify::{self, layer, CanClassify};
 use linkerd2_proxy_http::{profiles, timeout, HasH2Reason};
 use std::borrow::Cow;
+use tower_grpc::{self as grpc};
 use tracing::trace;
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


linkerd/app/core/src/classify.rs
```diff
@@ -183,10 +183,9 @@ fn grpc_class(headers: &http::HeaderMap) -> Option<Class> {
         .and_then(|v| v.to_str().ok())
         .and_then(|s| s.parse::<u32>().ok())
         .map(|grpc_status| {
-            let ok = if grpc_status == 0 {
-                SuccessOrFailure::Success
-            } else {
-                SuccessOrFailure::Failure
+            let ok = match grpc_status {
+                0 | 3 | 5 => SuccessOrFailure::Success,
```
seanmonstar: It can be hard to remember what the numbers mean exactly. How about a comment here reminding us?

olix0r: Are we able to use the const's from the gRPC crate? https://github.com/tower-rs/tower-grpc/blob/master/tower-grpc/src/status.rs#L26

tegioz: Sure, I was going to ask what did you think about going this way:

```rust
let ok = match grpc::Code::from_i32(grpc_status as i32) {
    grpc::Code::Ok | grpc::Code::InvalidArgument | grpc::Code::NotFound => {
        SuccessOrFailure::Success
    }
    _ => SuccessOrFailure::Failure,
};
```

I saw the `0` before and I intended to follow the same approach. First contact with Rust, any feedback is more than welcome :)

olix0r: I think we may not have had those consts available to us when this code was initially written :)  This change seems good to me.

tegioz: Thanks 👍

olix0r:
It seems to me that other grpc status codes should not be considered errors as well. In fact, it seems like only a few should be considered errors: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

Specifically, UNAVAILABLE and INTERNAL seem to be clear errors. I'm not sure that _any_ of the other error codes actually correspond to 5XX server failures.

PR #396, opened 2019-12-06
router: Ensure that the purge task completes

olix0r:
The router's cache purging tasks never complete. This means that
when a router is dropped, it's purge task is leaked into the executor,
possibly holding services.

This change uses an mpsc to ensure that the purge task is notified when
all of the routers using this cache have been dropped.

### DIFF
--- a/linkerd/router/src/cache.rs
+++ b/linkerd/router/src/cache.rs
@@ -1,9 +1,6 @@
-use std::{hash::Hash, time::Duration};
-
-use futures::{task, try_ready, Async, Future, Poll, Stream};
+use futures::{task, Async, Stream};
 use indexmap::IndexMap;
-
-use tokio::sync::lock::Lock;
+use std::{hash::Hash, time::Duration};
 use tokio_timer::{delay_queue, DelayQueue};
 use tracing::trace;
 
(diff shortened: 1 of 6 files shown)


linkerd/router/src/purge.rs
```diff
@@ -0,0 +1,54 @@
...
+pub struct Purge<K: Clone + Eq + Hash, V> {
+    cache: Lock<Cache<K, V>>,
+    hangup: mpsc::Receiver<Never>,
+}
+
+/// Ensures that `Purge` runs until all handles are dropped.
+#[derive(Clone)]
+pub struct Handle(mpsc::Sender<Never>);
+
+// ===== impl PurgeCache =====
```
kleimkuhler: "impl Purge" now

linkerd/router/src/cache.rs
```diff
@@ -121,18 +109,16 @@ where
     ///
     /// Polls the underlying `DelayQueue`. When elements are returned from the
     /// queue, remove the associated key from `values`.
-    fn poll_purge(&mut self) -> Poll<(), ()> {
+    pub fn purge(&mut self) {
```
adleong: If we made this function always return NotReady (instead of returning nothing), I think it would make it more clear from the API that calling this method will schedule this task to be notified.

olix0r: I think I'd rather rename the function that have a return type of `Poll<Never, Never>` -- it doesn't make the callsite any clearer. Suggestions?

adleong: Maybe I got the wrong idea about the implicit contract of Poll.  My understanding was that if a method returns NotReady, then the contract is that it *must* schedule the task to be notified.

This method *does* schedule the task to be notified, but that fact isn't communicated by the function signature.  I'm not sure what name would communicate that other than `poll` or `poll_purge`.

olix0r: You are correct that when something returns not ready it must schedule itself to be notified. However, the inverse of this is not true (all functions that schedule notification need not return notready -- especially for crate-internals).

PR #397, opened 2019-12-09
profiles: Eagerly read profiles off the wire

olix0r:
All communication on a destination client can become stalled if an individual profile client does not consume profile updates in a timely manner.

This change alters the buffering strategy for profile updates to ensure that the resolution stream is always polled, independently of whether the service is actively being used.

### DIFF
--- a/linkerd/app/core/src/profiles.rs
+++ b/linkerd/app/core/src/profiles.rs
@@ -11,7 +11,7 @@ use std::time::Duration;
 use tokio::sync::{oneshot, watch};
 use tokio_timer::{clock, Delay};
 use tower_grpc::{self as grpc, generic::client::GrpcService, Body, BoxBody};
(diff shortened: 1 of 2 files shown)


linkerd/app/core/src/profiles.rs
```diff
@@ -216,21 +213,14 @@ where
...
                     };
-
-                    let req = api::GetDestination {
-                        path: self.dst.clone(),
-                        context_token: self.context_token.clone(),
-                        ..Default::default()
-                    };
-                    debug!("getting profile: {:?}", req);
-                    let rspf = self.service.get_profile(grpc::Request::new(req));
+                    debug!("getting profile");
```
adleong: should we render the request is this message?

olix0r: the dst is already in the context now

PR #399, opened 2019-12-11
Remove default for destination lookup subnets

kleimkuhler:
Related to #3807, but serves as a more temporary fix until the proxy injector
provides a way for this to be configured from an install.

### DIFF
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -196,10 +196,6 @@ const DEFAULT_OUTBOUND_MAX_IN_FLIGHT: usize = 10_000;
 const DEFAULT_DESTINATION_GET_SUFFIXES: &str = "svc.cluster.local.";
 const DEFAULT_DESTINATION_PROFILE_SUFFIXES: &str = "svc.cluster.local.";
 
-// RFC6890 private networks
-// Doesn't make sense to include the IPv6 network yet (fd00::/8).
-const DEFAULT_DESTINATION_GET_NETWORKS: &str = "10.0.0.0/8,172.16.0.0/12,192.168.0.0/16";
-
 const DEFAULT_IDENTITY_MIN_REFRESH: Duration = Duration::from_secs(10);
 const DEFAULT_IDENTITY_MAX_REFRESH: Duration = Duration::from_secs(60 * 60 * 24);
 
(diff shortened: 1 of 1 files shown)


linkerd/app/src/env.rs
```diff
@@ -408,8 +404,7 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
             context: dst_token?.unwrap_or_default(),
             get_suffixes: dst_get_suffixes?
                 .unwrap_or(parse_dns_suffixes(DEFAULT_DESTINATION_GET_SUFFIXES).unwrap()),
-            get_networks: dst_get_networks?
-                .unwrap_or(parse_networks(DEFAULT_DESTINATION_GET_NETWORKS).unwrap()),
+            get_networks: dst_get_networks?.unwrap_or(IndexSet::new()),
```
olix0r: tioli: `unwrap_or_default()`

kleimkuhler: 👍

linkerd/app/src/env.rs
```diff
@@ -408,8 +404,7 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
             context: dst_token?.unwrap_or_default(),
             get_suffixes: dst_get_suffixes?
                 .unwrap_or(parse_dns_suffixes(DEFAULT_DESTINATION_GET_SUFFIXES).unwrap()),
-            get_networks: dst_get_networks?
-                .unwrap_or(parse_networks(DEFAULT_DESTINATION_GET_NETWORKS).unwrap()),
+            get_networks: dst_get_networks?.unwrap_or_default(),
```
adleong: Just for my understanding, `dst_get_networks` is a `Result<Option>`, yeah?  so the `unwrap_or_default` returns `None` when the environment variable is not provided?

olix0r: It's an IndexSet<IpNet> (so it's an empty set)

kleimkuhler: `dst_get_networks` returns a `Result<Option<IndexSet<T>>, Err>`. The `?` unwraps it to `Option<IndexSet<T>>`. The `unwrap_or_default` returns an `IndexSet<T>` if there is one, or the default (empty `IndexSet<T>`)

adleong: wait no, the `?` unwraps the `Result`.  So the `unwrap_or_default` returns the empty string?

adleong: :+1:

PR #401, opened 2019-12-17
discover: Timeout stalled resolutions

olix0r:
If a load balancer is not polled, i.e. because it is idle/leaked, its
resolution stream can fill up, causing backpressure onto the destination
controller client.

This change implements a timeout for this scenario. When the resolution
buffer is at capacity for a full idle timeout, the background resolution
completes to free its resources.

### DIFF
--- a/linkerd/proxy/discover/src/buffer.rs
+++ b/linkerd/proxy/discover/src/buffer.rs
@@ -1,13 +1,16 @@
 use futures::{try_ready, Async, Future, Poll, Stream};
 use linkerd2_error::{Error, Never};
 use std::fmt;
+use std::time::{Duration, Instant};
 use tokio::sync::{mpsc, oneshot};
+use tokio::timer::Delay;
 use tower::discover;
 use tracing_futures::Instrument;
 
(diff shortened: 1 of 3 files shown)


linkerd/proxy/discover/src/buffer.rs
```diff
@@ -111,10 +124,30 @@ where
...
+                    tracing::trace!("lost sender");
+                    return Err(());
+                }
+                Ok(Async::NotReady) => {
+                    let mut watchdog = self
+                        .watchdog
+                        .take()
+                        .unwrap_or_else(|| Delay::new(Instant::now() + self.watchdog_timeout));
+                    if watchdog.poll().expect("timer must not fail").is_ready() {
+                        tracing::warn!("dropping resolution due to watchdog timeout");
```
hawkw: Nit/TIOLI: should this include the duration of the timeout?

olix0r:
If the balanacer were to start consuming from the queue again, it would get an error when the queue is fully consumed and fail (which is what we'd hope).

PR #402, opened 2019-12-17
router: Move the `Make` trait to `stack` crate

olix0r:
The router's `Make` trait is generally useful when we have to build
middlewares. In preparation for improvements to the profile router, this
change moves the `Make` trait into the `linkerd2-stack` crate, with some
API improvements.

Furthermore, the Router is changed to avoid requiring that its services
implement `Clone`.

In general, we expect targets to be cheaply cloneable, but we should
only require that services are cloneable when absolutely necessary.

### DIFF
--- a/linkerd/app/core/src/proxy/buffer.rs
+++ b/linkerd/app/core/src/proxy/buffer.rs
@@ -1,7 +1,7 @@
 use crate::svc;
 use futures::{try_ready, Async, Future, Poll};
 use linkerd2_error::Error;
-use linkerd2_router as rt;
+use linkerd2_stack::Make;
 use std::marker::PhantomData;
 use std::sync::{Arc, Mutex, Weak};
 use std::time::{Duration, Instant};
@@ -25,7 +25,7 @@ pub struct Layer<D, Req> {
 
 /// Produces `MakeService`s where the output `Service` is wrapped with a `Buffer`
 #[derive(Debug)]
-pub struct Make<M, D, Req> {
+pub struct MakeBuffer<M, D, Req> {
     capacity: usize,
     deadline: D,
     inner: M,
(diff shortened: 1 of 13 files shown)


linkerd/router/src/lib.rs
```diff
@@ -281,66 +266,65 @@ where
...
-                    let target = target.take().expect("polled after ready");
 
                     // If the target is already cached, route the request to
                     // the service; otherwise, try to insert it
                     if let Some(service) = cache.access(&target) {
-                        trace!("target already cached");
-                        State::Call(Some(request), Some(service))
-                    } else {
-                        debug!("target not cached");
+                        let ready = service.poll_ready().map_err(|e| {
```
kleimkuhler: Thoughts on:

```
let ready = service
    .poll_ready()
    .map_err(Into::<error::Error>::into)?;
```

linkerd/router/src/lib.rs
```diff
@@ -281,66 +266,65 @@ where
...
-                }
-                State::Call(ref mut request, ref mut service) => {
-                    let mut service = service.take().expect("polled after ready");
+                        let mut service = LoadShed::new(make.make(target.clone()));
 
-                    assert!(
-                        service.poll_ready()?.is_ready(),
-                        "load shedding services must always be ready"
-                    );
+                        let ready = service.poll_ready().map_err(|e| {
```
kleimkuhler: Same suggestion here as above ^

linkerd/router/src/layer.rs
```diff
@@ -155,39 +162,3 @@ where
         }
     }
 }
-// === impl Service ===
```
kleimkuhler: 👍

linkerd/app/core/src/proxy/buffer.rs
```diff
@@ -120,7 +120,7 @@ impl<M: Clone, D: Clone, Req> Clone for Make<M, D, Req> {
     }
 }
 
-impl<T, M, D, Req> svc::Service<T> for Make<M, D, Req>
+impl<T, M, D, Req> svc::Service<T> for MakeBuffer<M, D, Req>
```
kleimkuhler: Thoughts on renaming `Self::Future` here `MakeBufferFuture` instead of keeping it `MakeFuture`?

olix0r:
I'm going to revive this after some other things land

PR #403, opened 2020-01-07
Decouple service acquisition from request processing

olix0r:
## Overview

At the outset of the proxy's life, we introduced a `Router`. The router has a
few jobs:

1. To identify a target for each request;
2. To provision a service for that target; and
3. To maintain a cache of services to be reused across connections.
4. To evict "idle" services that have not been used, i.e. so that the cache
   does not leak defunct services.

Because the router cannot know which inner service will be used before the
request is received, the router's `poll_ready` _always_ returns ready,
meaning that it cannot exert backpressure to its callers.

So, in order to ensure that the router's inner services can be shared by an
arbitrary number of callers---and to ensure that the inner service is driven
to readiness---we had to add a buffering layer within the router. The buffer
holds requests while the inner service is driven to readiness on a dedicated
task.

But requests could remain queued indefinitely, so we introduced a "deadline"
feature to the buffer so that the request could be stolen from the queue and
failed if the request was not processed in a given amount of time.

And then we added Service Profiles with destination overrides; and with these
features, a slew of new routers and buffers.

As we've diagnosed recent issue reports, it's become apparent that the
system, as its grown organically, does not properly handle backpressure. This
most frequently manifests as 503s when requests are timed out of buffers,
though it is undoubtedly related to a plethora of other user-reported issues
(like Service Discovery staleness).

### Buffers

Buffers are deployed to solve two problems: (1) Clonability and (2)
Readiness.

#### Clonability

Without getting too into the weeds of Rust's memory model or Tokio's
execution model, _clonability_ basically refers to the ability to share a
single service across multiple tasks. For example, if an application
initiates multiple connections to an outbound proxy, we don't want to create
new routers/caches/load balancers for each connection. Instead, we want to
share the cached load balancers across all of these connections, so we need
to _clone_ the cache into each connection's task. The buffer allows multiple
tasks to send messages to a single service (via an Multi-producer Single
Consumer queue).

#### Readiness

In Tower, a service's _readiness_ indicates its ability to process requests.
Before a request can be dispatched, the caller must invoke
`Service::poll_ready` to ensure that the inner service is able to accept a
request. This is how backpressure works. If an HTTP server's inner service is
not ready, it won't attempt to read a request from the socket, and so the
remote client's write calls will block once the kernel buffers are full.
Backpressure magic, baby!

When we use buffers to "ensure readiness", we are effectively disabling
backpressure. We are signing up to handle requests and have to deal with
timeouts, etc. We should only do this in rare and exceptional circumstances.
But, as discussed above, our current routing strategy explicitly requires
that we do not exert backpressure: routers must always be ready; and so must
each inner service. Otherwise requests are dropped on the floor. So... not
great.

#### Cancelation

If an inner service does not become ready in a timely fashion, requests can
get "stucK" in the buffer. If a caller cancels the request (i.e. by dropping
the response future), we don't have any means to eagerly evict the request
from the buffer. We've taken pains to be able to "steal" requests from the
buffer after a timeout; but this behavior has proven to be complex,
imperfect, and difficult to diagnose/explain.

### Routers

As discussed above, routers do a few jobs; and we currently use routers for a
few different things.

#### Destination routing

The primary use is when we receive a request and want to send it through a
service that is configured by the control plane. We don't want to query the
control plane for each request, and so we want to cache a service that holds
the proper configuration for the target service. Also, as mentioned above,
these services need to be garbage collected once they are no longer in use,
otherwise the proxy is prone to consume memory without bounds.

#### Service Profile routing

Routers are used somewhat differently in the context of Service profiles,
though. Service profile routing is substantially different from destination
routing:

* All routes are known a priori, and hence do not need to be discovered for
  each request;
* Because all routes are provided by the control plane, garbage collection is
  unnecessary and undesirable.
* Because all routes in a service profile operate over the same inner
  service(s), and because none of the layers in service profile routes can
  actually implement backpressure, there's no reason a service profile router
  _really_ needs to guarantee readiness.

The concrete destination router similarly abuses routers, and therefore busts
our ability to exert backpressure.

### Summary

1. We _only_ care about clonability. We should **~never** synthesize
   readiness in the data path.
2. We need a simpler mechanism for bounding the time a request remains in the
   proxy before being dispatched to a client.
3. The routing layer is not a one-size-fits-all solution. Be wary the siren
   song of code reuse.

## Solution

Armed with this fresh knowledge, I've refactored the proxy stacks to (1)
eliminate all buffer layers from the data path and (2) enforce a single
_service acquisition timeout_ that bounds the amount of time a request waits
in the proxy to be dispatched to a client.

### Clonability without Buffer

Assuming that we can relax the readiness constraints, what size should the
buffers be to support clonability? If we used a buffer with only a single
slot, for instance, we could limit the how many requests can get stuck in a
buffer (to 1), but the problem would remain. We can't create a buffer with a
capacity of zero... but if we could, we would have a Mutex. So why not just
use one of those?

My proposed change replaces use of a `Buffer` with a new (clonable) `Lock`
layer. `Lock:poll_ready` first acquires a lock guarnteeing exclusive access
to the inner service before the service is polled to ready. Then, the locked
state is consumed and dropped as a request is dispatched on this service,
permitting other calls to obtain exclusive access to the service.

The one important subtlety here is that we need to be careful about services
that invoke `poll_ready` before polling another source (like the socket) for
a request. In these cases, the service can hold the lock on the inner
service, preventing other services from being stalled.

Services that are in this situation may either clone the service to drop the
locked state; or a `Oneshot` layer can be used to push backpressure into the
service's response (i.e. so that poll_ready is not invoked until the request
is materialized).

This is all to say that using a lock comes with some complexity. We have to
be careful about how we use services that may contain a lock, otherwise tasks
may be starved. This seems to be an acceptable tradeoff, though we'll need to
find ways to detect/test improper access patterns.

### Breaking the Readiness Requirement

Fundamentally, we need to do routing without requiring that all routes are
ready; and we need to be able to limit the amount of time a request spends
waiting for a ready service.

In order to accomplish this, we need to formalize two distinct "phases" of
proxying. The first phase is **service acquisition**, which can be loosely
described by the `tower::MakeService` API: we first (asynchronously) build a
service for a target type so that the request can be dispatched to the target
service. We can set a _service acquisition timeout_ on the future that
acquires the service (i.e. without setting a _response timeout_).

This is all accomplished by decoupling the _routing_ from _caching_. Routing
selects _which target to use_. The cache implements `tower::MakeService`,
_returning_ a service for each target. With this decoupling, we can insert
timeout layers within the top-level logical router to ensure that service
acquisition is bounded.

With this new strategy, the Tokio runtime becomes the buffer. We can bound
the total number of requests in flight to constrain this buffer; but this
means that backpressure, cancelation, etc can be achieved more naturally.

## Areas of Focus

Look, this change is huge. A significant number of superficial changes are
co-mingled here. On the upside, compile time has been reduced substantially,
and some organically diverging idioms have been consolidated.

1. The proxy stacks;
2. The lock layer;
3. The caching layer;
4. The routing layer;
5. The profile router;

Beyond this, I've endeavored to limit PhantomData uses, especially in layers.
They make it so the stack cannot be reused and, I believe, negatively impact
compile times. In general, I've tried to remove unnecesscary type constraints.
They make the system more resistant to change. Such changes include:

* The `router::Make` type has been renamed to `stack::NewService` -- `Make` and
  `MakeService` were too close for comfort.

* Many layers now implement both `MakeService` (via `Service`) and `NewService`
* `Stack` types are now generally named `MakeFoo` or `NewFoo`;
* Types in the retry module have been renamed (so that the Service could be called Retry,
  basically).
* The profile router is just totally new.
* Target types, especially `DstAddr` are not reused. Purpose-specific target types are used.
* Probably other things, too. It's been a trip.

---

This is going to be difficult to review. My advice to you, the reviewer, is to first focus on the above description. Please ask questions & poke at implementation until most of the above makes some sense. I'd like to schedule a longer in-person review for later this week so that we can get enough confidence to move forward with this approach.


olix0r:
Resubmitting this in smaller chunks

PR #407, opened 2020-01-17
update the proxy to tokio 0.2

hawkw:
This updates only the proxy, not the test-support code (as suggested by @olix0r).

I'm opening this as a draft since we likely will want to wait on merging it until the performance-testing harness is in place (#406) to compare performance with tokio 0.1.


hawkw:
Rebased and updated this to the latest master. Going to do some performance testing & hopefully we can get this on the path to actually merging.

olix0r:
@hawkw Thanks! I wouldn't spend too much time on perf analysis-- I want to play around with this a little before advancing it. We're going to hold this at least until 2.7.1 is released.

PR #409, opened 2020-01-21
internal: Split retry, http-classify, and http-metrics

olix0r:
The retry middleware is coupled to both the HTTP metrics infrastructure
as well as the various request extensions that are required to route a
request (i.e. via `TryClone`). Because of this, retry metrics are
present for all HTTP metric scopes, even where retries can't apply.

This change extracts several components from the `proxy-http`
conglomerate subcrate:

* `http-classify` includes classification traits and a middleware layer;
* `http-metrics` wraps `tower::retry` with a `MakeService` configuration layer;
* `retry` wraps `tower::retry` with a `MakeService` configuration layer;
* `app-core` now handles retry-configuration

Additionally, the `metrics::Metric` type is now generic over the metric
name type so that prefixed names do not need to be formatted eagerly.

The `route_actual_retry_skipped_total` counter has been replaced with a
`route_retryable_total` counter. This new value is only present when
retries are enabled on a route; and a `skipped` labels is added when
retryability is ignored (i.e. due to budget being exhausted). This
metric is not consumed by the control plane, so it should be safe to
change.


linkerd/http-classify/src/layer.rs
```diff
@@ -0,0 +1,89 @@
+use super::{CanClassify, Classify};
+use futures::{try_ready, Future, Poll};
+
+#[derive(Debug, Clone)]
+pub struct Layer(());
+
+#[derive(Clone, Debug)]
+pub struct Make<N> {
```
hawkw: Nit/TIOLI: Elsewhere, it seems like we've been changing the naming for `Service`s/`MakeService`s to include the service name. While we're here, should this be changed to `MakeHttpClassify`? (and the `Service` struct `HttpClassify`?)

linkerd/http-metrics/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+h2 = "0.1"
+http = "0.1"
+http-body = "0.1"
+httparse = "1.2"
+hyper = "0.12.3"
+indexmap = "1.0"
+linkerd2-error = { path  = "../error" }
+linkerd2-http-classify = { path  = "../http-classify" }
+linkerd2-metrics = { path  = "../metrics" }
+rand = "0.7"
```
hawkw: what do we use `rand` for in this crate? i didn't see it in the code.

linkerd/http-metrics/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+edition = "2018"
+publish = false
+
+[dependencies]
+bytes = "0.4"
+futures = "0.1"
+h2 = "0.1"
+http = "0.1"
+http-body = "0.1"
+httparse = "1.2"
```
hawkw: where do we use `httparse` in this crate? AFAICT, we only use it in [version.rs](https://github.com/linkerd/linkerd2-proxy/blob/8c8134800b614820df75ba47632987b9f663faeb/linkerd/proxy/http/src/version.rs) which doesn't seem to have moved in this change?

linkerd/http-metrics/Cargo.toml
hawkw: in general, i think we may want to run whatever the cargo tool that tells you if dependencies are actually used?

linkerd/http-metrics/src/requests/layer.rs
```diff
@@ -1,17 +1,16 @@
-use super::super::retry::TryClone;
-use super::classify::{ClassifyEos, ClassifyResponse};
-use super::{ClassMetrics, Registry, RequestMetrics, StatusMetrics};
+//use super::super::retry::TryClone;
```
kleimkuhler: Can remove this

olix0r:
There are followup changes coming to these modules. Naming will change, etc.

linkerd/http-metrics/Cargo.toml
olix0r: For what it's worth, I did run the tool and it didn't catch any of this :man_shrugging:

PR #414, opened 2020-01-23
patch `prost` 0.5 to pick up security fix

hawkw:
This branch updates the proxy's `prost` dependency to a patched version
of `prost` 0.5 that incorporates the cahnges in danburkert/prost#268.
This patch fixes a security issue where a malicious protobuf message
could be used to trigger a stack overflow.

We are unfortunately unable to easily update to `prost` 0.6.1, which
includes this fix, as 0.6 updates the `bytes` dependency to 0.5. The
`tokio` 0.1 ecosystem that the proxy currently uses still depends on
0.4, and the breaking changes in 0.5 are quite significant. Therefore,
updating to `bytes` 0.5 would require a lot of fairly large changes to
legacy versions of...pretty much everything (`tokio-io`, `tokio-buf`,
`hyper`, `http-body`...). As we intend to update to `tokio` 0.2 in the
near future, patching all these legacy dependencies is a bit of a waste
of time. Therefore, I opted to backport the security fix to a compatible
`prost` version instead.

Closes linkerd/linkerd2#3963

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1480,10 +1480,8 @@ dependencies = [
 [[package]]
 name = "prost"
 version = "0.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "96d14b1c185652833d24aaad41c5832b0be5616a590227c1fbff57c616754b23"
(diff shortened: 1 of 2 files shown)


hawkw:
unfortunately, `cargo audit` will still think we are vulnerable because the version hasn't changed (and, we can't change it since it's necessary to patch the version _that other crates depend on_)

PR #415, opened 2020-01-26
metrics: Make Counter & Gauge atomic

olix0r:
The `Counter` and `Gauge` metrics currently hold `u64` values. This
means that all uses need to gaurd access with a `Mutex`.

By changing both types to hold an `AtomicU64`, we can avoid using locks
in some cases.

### DIFF
--- a/linkerd/app/core/src/handle_time.rs
+++ b/linkerd/app/core/src/handle_time.rs
@@ -259,7 +259,7 @@ impl Shared {
         if counter.clones.fetch_sub(1, Ordering::Release) == 1 {
             let elapsed = t0.elapsed();
 
-            let mut hist = match self.histogram.lock() {
+            let hist = match self.histogram.lock() {
                 Ok(lock) => lock,
                 // Avoid double panicking in drop.
                 Err(_) if panicking => return,
--- a/linkerd/metrics/src/counter.rs
+++ b/linkerd/metrics/src/counter.rs
@@ -1,4 +1,4 @@
-use super::prom::{FmtLabels, FmtMetric};
+use super::prom::{FmtLabels, FmtMetric, MAX_PRECISE_VALUE};
 use std::fmt::{self, Display};
 use std::ops;
 use std::sync::atomic::{AtomicU64, Ordering};
@@ -18,13 +18,6 @@ use std::sync::atomic::{AtomicU64, Ordering};
 #[derive(Debug, Default)]
 pub struct Counter(AtomicU64);
 
-/// Largest `u64` that can fit without loss of precision in `f64` (2^53).
-///
-/// Wrapping is based on the fact that Prometheus models counters as f64 (52-bits
-/// mantissa), thus integer values over 2^53 are not guaranteed to be correctly
-/// exposed.
-pub(crate) const MAX_PRECISE_COUNTER: u64 = 0x20_0000_0000_0000;
-
 // ===== impl Counter =====
 
 impl Counter {
@@ -40,7 +33,6 @@ impl Counter {
     pub fn value(&self) -> u64 {
         self.0
             .load(Ordering::Acquire)
-            .wrapping_rem(MAX_PRECISE_COUNTER + 1)
     }
 }
 
@@ -87,7 +79,7 @@ impl FmtMetric for Counter {
     const KIND: &'static str = "counter";
 
     fn fmt_metric<N: Display>(&self, f: &mut fmt::Formatter<'_>, name: N) -> fmt::Result {
-        writeln!(f, "{} {}", name, self.value())
+        writeln!(f, "{} {}", name, self.value().wrapping_rem(MAX_PRECISE_VALUE))
     }
 
     fn fmt_metric_labeled<N, L>(
@@ -102,7 +94,7 @@ impl FmtMetric for Counter {
     {
         write!(f, "{}{{", name)?;
         labels.fmt_labels(f)?;
-        writeln!(f, "}} {}", self.value())
+        writeln!(f, "}} {}", self.value().wrapping_rem(MAX_PRECISE_VALUE))
     }
 }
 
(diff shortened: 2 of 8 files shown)


olix0r:
Hmm, I probably need to correct this per #139? I'm not sure how to handle this with atomics...

linkerd/metrics/src/gauge.rs
```diff
@@ -1,50 +1,44 @@
...
     /// Decrement the gauge by one.
-    pub fn decr(&mut self) {
-        if let Some(new_value) = self.0.checked_sub(1) {
-            (*self).0 = new_value;
-        } else {
-            warn!("Gauge underflow");
-        }
+    pub fn decr(&self) {
+        self.0.fetch_sub(1, Ordering::SeqCst);
+    }
```
hawkw: I don't think these need to be `SeqCst`. This means that they will be part of a global order with _other_ atomic values, which I don't think is really necessary. They probably only need to be `Release`, so that any loads in calls to `Gauge::value` will happen after the increment/decrement ops.

linkerd/metrics/src/counter.rs
```diff
@@ -15,75 +15,70 @@ use super::prom::{FmtLabels, FmtMetric};
 /// [`rate()`]: https://prometheus.io/docs/prometheus/latest/querying/functions/#rate()
 /// [`irate()`]: https://prometheus.io/docs/prometheus/latest/querying/functions/#irate()
 /// [`resets()`]: https://prometheus.io/docs/prometheus/latest/querying/functions/#resets
-#[derive(Copy, Clone, Debug, Default, Eq, PartialEq)]
-pub struct Counter(u64);
-
-/// Largest `u64` that can fit without loss of precision in `f64` (2^53).
```
hawkw: I think we need to retain this behavior for prometheus compatibility. I think that rather than `fetch_add`, we need to change this code to do a load, 52-bit wrapping add, and try to CAS the new value back. This probably needs to be a CAS loop.

linkerd/metrics/src/counter.rs
```diff
@@ -15,75 +15,70 @@ use super::prom::{FmtLabels, FmtMetric};
...
-    ///
-    /// This function wraps on 52-bit overflows.
-    pub fn incr(&mut self) {
-        *self += 1;
+    pub fn incr(&self) {
+        self.add(1)
+    }
+
+    pub fn add(&self, n: u64) {
+        self.0.fetch_add(n, Ordering::SeqCst);
```
hawkw: Since a counter is only ever incremented, the fetch_add only needs to be `Release`, to establish a happens-before relationship with the `Acquire` load operations.

linkerd/metrics/src/gauge.rs
```diff
@@ -1,50 +1,44 @@
...
-
 use super::{FmtLabels, FmtMetric};
+use std::fmt::{self, Display};
+use std::sync::atomic::{AtomicU64, Ordering};
 
 /// An instaneous metric value.
-#[derive(Copy, Clone, Debug, Default, Eq, PartialEq)]
-pub struct Gauge(u64);
+#[derive(Debug, Default)]
+pub struct Gauge(AtomicU64);
```
hawkw: Hmm, for prometheus compatibility, do gauges also need to implement 52-bit precision? I know they don't _currently_...

linkerd/metrics/src/counter.rs
```diff
@@ -15,75 +15,72 @@ use super::prom::{FmtLabels, FmtMetric};
...
+    pub fn add(&self, n: u64) {
+        self.0.fetch_add(n, Ordering::Release);
     }
 
-    /// Return current counter value.
+    /// Return current counter value, wrapped to be safe for use with Prometheus.
     pub fn value(&self) -> u64 {
         self.0
+            .load(Ordering::Acquire)
+            .wrapping_rem(MAX_PRECISE_VALUE + 1)
```
hawkw: oh, just wrapping here is much neater than doing a load-wrap-CAS loop as i suggested, 👍

linkerd/metrics/src/counter.rs
```diff
@@ -15,75 +15,72 @@ use super::prom::{FmtLabels, FmtMetric};
...
 
 impl ops::Add<u64> for Counter {
     type Output = Self;
     fn add(self, rhs: u64) -> Self::Output {
-        let wrapped = self
-            .0
-            .wrapping_add(rhs)
-            .wrapping_rem(MAX_PRECISE_COUNTER + 1);
-        Counter(wrapped)
+        self.0.fetch_add(rhs, Ordering::SeqCst);
```
hawkw: should probably be `Release`, like `add` does
```suggestion
        self.0.fetch_add(rhs, Ordering::Release);
```

hawkw: or, actually...can this just use the inherent `add` method so we don't need to keep the implementation in sync?
```suggestion
        self.add(rhs)
```

linkerd/metrics/src/counter.rs
```diff
@@ -15,75 +15,72 @@ use super::prom::{FmtLabels, FmtMetric};
...
-        self + rhs
+    fn add(self, rhs: Self) -> Self::Output {
+        self + rhs.value()
     }
 }
 
 impl ops::AddAssign<u64> for Counter {
     fn add_assign(&mut self, rhs: u64) {
-        *self = *self + rhs
+        self.0.fetch_add(rhs, Ordering::SeqCst);
```
hawkw:
```suggestion
        self.add(rhs);
```

linkerd/app/core/src/handle_time.rs
```diff
@@ -259,7 +259,7 @@ impl Shared {
         if counter.clones.fetch_sub(1, Ordering::Release) == 1 {
             let elapsed = t0.elapsed();
 
-            let mut hist = match self.histogram.lock() {
+            let hist = match self.histogram.lock() {
```
hawkw: since we no longer need to mutate the histogram, should we consider using a rwlock? (could be a follow-up)

olix0r: @hawkw sounds reasonable but yeah, let's save that for a followup

PR #416, opened 2020-01-27
Set grpc-status headers on dispatch errors

olix0r:
When the proxy encounters errors, it sets the HTTP response code.

However, gRPC applications do not honor HTTP response codes and,
instead, rely on the `grpc-status` response header to be set. Since the
proxy has specialized gRPC support in some other places (namely
metrics/classification), we should attempt to behave well with gRPC
applications.

This change extracts the `errors` module from app-core into a new
`error-respond` crate, which exposes a simplified trait interface.
`app::core::errors` now provides an implementation for these traits that
is able to translate proxy-internal errors to gRPC status codes and
messages.

Fixes linkerd/linkerd2#3493

### DIFF
--- a/linkerd/app/core/src/errors.rs
+++ b/linkerd/app/core/src/errors.rs
@@ -1,150 +1,177 @@
-//! Layer to map HTTP service errors into appropriate `http::Response`s.
-
-use crate::svc;
-use futures::{Future, Poll};
-use http::{header, Request, Response, StatusCode, Version};
+use crate::proxy::identity;
+use http::{header::HeaderValue, StatusCode};
 use linkerd2_error::Error;
+use linkerd2_error_respond as respond;
 use linkerd2_proxy_http::HasH2Reason;
-use tracing::{debug, error, warn};
+use linkerd2_router::error as router;
+use tower::load_shed::error as shed;
+use tower_grpc::{self as grpc, Code};
+use tracing::debug;
 
-/// Layer to map HTTP service errors into appropriate `http::Response`s.
-pub fn layer() -> Layer {
-    Layer
+pub fn layer<B: Default>() -> respond::RespondLayer<NewRespond<B>> {
+    respond::RespondLayer::new(NewRespond(std::marker::PhantomData))
 }
-
-#[derive(Clone, Debug)]
-pub struct Layer;
-
-#[derive(Clone, Debug)]
-pub struct Stack<M> {
-    inner: M,
-}
-
-#[derive(Clone, Debug)]
-pub struct Service<S>(S);
 
 #[derive(Debug)]
-pub struct ResponseFuture<F> {
-    inner: F,
-    is_http2: bool,
+pub struct NewRespond<B>(std::marker::PhantomData<fn() -> B>);
+
+#[derive(Copy, Clone, Debug)]
+pub enum Respond<B> {
+    Http1(std::marker::PhantomData<fn() -> B>),
+    Http2 { is_grpc: bool },
 }
 
-#[derive(Clone, Debug)]
-pub struct StatusError {
-    pub status: http::StatusCode,
-    pub message: String,
-}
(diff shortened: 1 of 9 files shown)


linkerd/app/core/src/errors.rs
```diff
@@ -1,150 +1,177 @@
...
+fn http_status(error: Error) -> StatusCode {
+    if error.is::<router::NoCapacity>() {
         http::StatusCode::SERVICE_UNAVAILABLE
-    } else if let Some(_) = e.downcast_ref::<shed::Overloaded>() {
-        warn!("server overloaded, max-in-flight reached");
+    } else if error.is::<shed::Overloaded>() {
         http::StatusCode::SERVICE_UNAVAILABLE
-    } else if let Some(_) = e.downcast_ref::<buffer::Aborted>() {
-        warn!("request aborted because it reached the configured dispatch deadline");
+    } else if error.is::<tower::timeout::error::Elapsed>() {
```
olix0r: This should be the DispatchDeadline error instead for now

linkerd/app/core/src/errors.rs
```diff
@@ -1,150 +1,177 @@
...
-pub fn layer() -> Layer {
-    Layer
-}
-
-#[derive(Clone, Debug)]
-pub struct Layer;
+use linkerd2_router::error as router;
+use tower::load_shed::error as shed;
+use tower_grpc::{self as grpc, Code};
+use tracing::debug;
```
hawkw: Teensy style nit: I notice we import `debug!` by name, but refer to `warn!` as `tracing::warn!`...is this just because there are multiple `debug!`s and only a single `warn!`?

linkerd/app/core/src/errors.rs
```diff
@@ -1,150 +1,177 @@
...
-                        debug!("propagating http2 response error: {:?}", err);
-                        return Err(err);
-                    }
-                }
-
-                let response = Response::builder()
-                    .status(map_err_to_5xx(err))
-                    .header(header::CONTENT_LENGTH, "0")
+            if *is_grpc {
+                debug!("Handling error with gRPC status code");
```
hawkw: should we move this to somewhere where we can record what status we're returning?

linkerd/app/core/src/errors.rs
```diff
@@ -1,150 +1,177 @@
...
-
-                Ok(response.into())
+                set_grpc_status(error, rsp.headers_mut());
+                return Ok(rsp);
             }
         }
+
+        let status = http_status(error);
+        debug!(%status, "Handling error with HTTP response");
+        return Ok(http::Response::builder()
```
hawkw: nit: since the function body ends with this, I don't think the explicit `return` statement is really necessary (if we removed the semicolon)

linkerd/app/core/src/errors.rs
```diff
@@ -1,150 +1,187 @@
...
+    } else if error.is::<buffer::Aborted>() {
+        let code = Code::Unavailable;
+        headers.insert(GRPC_STATUS, code_header(code));
+        headers.insert(
+            GRPC_MESSAGE,
+            HeaderValue::from_static("proxy dispatch timed out"),
+        );
+        code
+    } else if error.is::<IdentityRequired>() {
+        let code = Code::FailedPrecondition;
```
kleimkuhler: I don't think anything relied on the `Forbidden` status code, but is this changing the error type for required tap identity? Before in `require_identity_on_endpoint`, if the condition failed we returned an `error::StatusError` with a `Status = Forbidden`. Now it's a gRPC status code `FailedPrecondition`

olix0r: It's still a `FORBIDDEN` for HTTP requests; but is now a `FailedPrecondition` for gRPC requests.. reading through the descriptions i thought this was the closest ft.

PR #417, opened 2020-01-28
Record a metric of proxy errors

olix0r:
The proxy handles errors by synthesizing responses; however, depending
on where such an error occurs, it may not be recorded by a metrics
layer.

This change introduces a `request_errors_total` counter that records
the number/kind of errors encountered. This is immediately helpful for
debugging and diagnostics, but could also be helpful to view in the UI.
We may want to revisit this metric, especialy its labels, before relying
on it more prominently. But in the meantime it's extraordinarily useful
as it is.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -138,10 +138,11 @@ checksum = "b486ce3ccf7ffd79fdeb678eac06a9e6c09fc88d33836340becb8fffe87c5e33"
 
 [[package]]
 name = "chrono"
-version = "0.4.10"
+version = "0.4.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "31850b4a4d6bae316f7a09e691c944c28299298837edc0a03f755618c23cbc01"
+checksum = "e8493056968583b0193c1bb04d6f7684586f3726992d6c573261941a895dbd68"
 dependencies = [
+ "libc",
  "num-integer",
  "num-traits 0.2.6",
  "time",
@@ -628,6 +629,7 @@ dependencies = [
  "linkerd2-drain",
  "linkerd2-duplex",
  "linkerd2-error",
+ "linkerd2-error-metrics",
  "linkerd2-error-respond",
  "linkerd2-exp-backoff",
  "linkerd2-fallback",
@@ -784,6 +786,17 @@ name = "linkerd2-error"
 version = "0.1.0"
 dependencies = [
  "futures",
+]
+
+[[package]]
+name = "linkerd2-error-metrics"
+version = "0.1.0"
+dependencies = [
+ "futures",
+ "indexmap",
+ "linkerd2-error",
+ "linkerd2-metrics",
+ "tower",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -138,11 +138,10 @@ checksum = "b486ce3ccf7ffd79fdeb678eac06a9e6c09fc88d33836340becb8fffe87c5e33"
 
 [[package]]
 name = "chrono"
-version = "0.4.9"
+version = "0.4.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e8493056968583b0193c1bb04d6f7684586f3726992d6c573261941a895dbd68"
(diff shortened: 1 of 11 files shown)


olix0r:
Depends on #416

Cargo.lock
```diff
@@ -138,10 +138,11 @@ checksum = "b486ce3ccf7ffd79fdeb678eac06a9e6c09fc88d33836340becb8fffe87c5e33"
 
 [[package]]
 name = "chrono"
-version = "0.4.10"
+version = "0.4.9"
```
hawkw: Is there a reason this downgrade was necessary? Just curious.

linkerd/app/core/src/errors.rs
```diff
@@ -193,3 +214,55 @@ impl std::fmt::Display for IdentityRequired {
...
+    }
+}
+
+impl metrics::FmtLabels for Reason {
+    fn fmt_labels(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "message=\"{}\"",
+            match self {
+                Reason::CacheFull => "router full",
```
kleimkuhler:
```suggested
Reason::CacheFull => "router cache full",
```

I think it may be helpful to clarify that `router full` reasons happen *when* the router cache is full--especially for someone that may not be familiar with the proxy's router implementation and caching.

linkerd/error-metrics/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+pub struct RecordErrorLayer<L, K: Hash + Eq> {
+    label: L,
+    errors: Arc<Mutex<IndexMap<K, Counter>>>,
+}
+
+pub struct RecordError<L, K: Hash + Eq, S> {
+    label: L,
+    errors: Arc<Mutex<IndexMap<K, Counter>>>,
+    inner: S,
+}
```
kleimkuhler: What are your thoughts on changing these structs to:

```
pub struct RecordErrorLayer<L, K: Hash + Eq> {
    label: L,
    registry: Registry<K>,
}

pub struct RecordError<L, K: Hash + Eq, S> {
    label: L,
    registry: Registry<K>,
    inner: S,
}
```

When reading through this change, this change is what I expected to see and how I thought about it when understanding the rest of the `impl`s.

linkerd/app/core/src/errors.rs
olix0r: In the upcoming branch that this change was derived from, caching is explicitly separated from routing. So I'll probably leave as-is on the assumption that lands soon.

linkerd/error-metrics/src/lib.rs
kleimkuhler: Alternatively, this could be similar to the `Registry` in #428 where:

```
type Registry<L> = Arc<Mutex<IndexMap<L, Arc<Metrics>>>>;
```

olix0r: Eh, the `Registry` type is really there as a public handle for getting layers and reporting the results... It's a bit confusing for Registry to produce layers but those layers just hold Registries?

This might all be clearer if I split the types up into separate files?

kleimkuhler: Yea it helps, but I think what I'm reacting to is it seems like each `*-metrics` crate has it's own way of implementing what it's concept of a `Registry` is and they are just different enough that I feel like I'm reviewing new impls.

Reviewing this change, the other open #428 metrics change, and the previous `http-metrics` crate change all have a similar purpose but don't really make their metrics recording services the same way.

I'm okay merging this as is, but it may just be something to be aware of that as more metrics are added, see if it makes sense to introduce a reusable "metrics registry".

linkerd/error-metrics/src/service.rs
```diff
@@ -0,0 +1,95 @@
+use crate::LabelError;
+use futures::{Future, Poll};
+use indexmap::IndexMap;
+use linkerd2_metrics::{Counter, FmtLabels};
+use std::hash::Hash;
+use std::sync::{Arc, Mutex};
+
+/// A middlware taht records errors.
```
kleimkuhler: typo: `taht`

PR #420, opened 2020-01-31
trace: Use debug_span! instead of info_span!

olix0r:
linkerd/linkerd2#3998 describes an issue where the proxy's memory grows
with traffic. After testing, we've identified that this is caused by
logging emitted by `tracing`. 07667b8b upgraded the `tracing-subscriber`
trait, which reduced memory pressure; however, we continue to observe
heap usage grow in large leaps when running with the default log level
`linkerd=info,warn`. This appears to be due to span creation, which
occurs on every new connection.

By changing uses of `info_span!` to `debug_span!`, we can avoid this
allocation path at the expense of losing contextual logging. This seems
like a suitable tradeoff until we can address the underling issues in
`tracing`.

I've tested this overnight and memory usage remains effectively flat.

### DIFF
--- a/linkerd/app/core/src/control.rs
+++ b/linkerd/app/core/src/control.rs
@@ -152,7 +152,7 @@ pub mod resolve {
     use linkerd2_dns as dns;
     use std::net::SocketAddr;
     use std::{error, fmt};
-    use tracing::info_span;
+    use tracing::debug_span;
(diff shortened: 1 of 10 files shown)


hawkw:
Once we're able to verify that #423 fixes the leak, we should probably close this in favour of that?

PR #421, opened 2020-01-31
router: Avoid large up-front allocations

olix0r:
We initially opted to pre-allocate the delay queue to avoid allocation
at runtime. However, practically, router capacities are _much_ higher
than their typical usage; so we incur an extra ~1MB+ of allocation that
will never really be used (for outbound routers).

This change initializes the router's DelayQueue with only a single slot
by default. The capacity will be increased at runtime as needed.

### DIFF
--- a/linkerd/router/src/cache.rs
+++ b/linkerd/router/src/cache.rs
@@ -58,7 +58,7 @@ where
         Self {
             capacity,
             expires,
-            expirations: DelayQueue::with_capacity(1),
+            expirations: DelayQueue::new(),
             values: IndexMap::default(),
             purge_task: None,
         }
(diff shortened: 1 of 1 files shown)


olix0r:
@kleimkuhler the thought was that we'll ~never create routers that have _no_ services.  Or that will be pretty rare. But it's not a particularly important optimization.

kleimkuhler:
> @kleimkuhler the thought was that we'll ~never create routers that have _no_ services. Or that will be pretty rare. But it's not a particularly important optimization.

Yep makes sense. It was more coming from the fact that we already have an allocation to make when a service is added since `values` defaults to `0` capacity. So it may make sense to change values to be `IndexMap::with_capacity(1)` as well.

PR #424, opened 2020-02-02
errors: Set correct HTTP version on responses

olix0r:
When synthesizing error responses, all errors were sent with the default
HTTP/1.1 response. This change ensures that the request's version is
used when synthesizing a response.

### DIFF
--- a/linkerd/app/core/src/errors.rs
+++ b/linkerd/app/core/src/errors.rs
@@ -52,27 +52,27 @@ impl<B: Default> respond::Respond for Respond<B> {
     fn respond(&self, error: Error) -> Result<Self::Response, Error> {
         warn!("Failed to proxy request: {}", error);
 
+        if let Respond::Http2 { is_grpc } = self {
+            if let Some(reset) = error.h2_reason() {
+                debug!(%reset, "Propagating HTTP2 reset");
+                return Err(error);
+            }
+
+            if *is_grpc {
+                let mut rsp = http::Response::builder()
+                    .version(http::Version::HTTP_2)
+                    .header(http::header::CONTENT_LENGTH, "0")
+                    .body(B::default())
+                    .expect("app::errors response is valid");
+                let code = set_grpc_status(error, rsp.headers_mut());
+                debug!(?code, "Handling error with gRPC status");
+                return Ok(rsp);
+            }
+        };
+
         let version = match self {
             Respond::Http1(ref version, _) => version.clone(),
-            Respond::Http2 { is_grpc } => {
-                if let Some(reset) = error.h2_reason() {
-                    debug!(%reset, "Propagating HTTP2 reset");
-                    return Err(error);
-                }
-
-                if *is_grpc {
-                    let mut rsp = http::Response::builder()
-                        .version(http::Version::HTTP_2)
-                        .header(http::header::CONTENT_LENGTH, "0")
-                        .body(B::default())
-                        .expect("app::errors response is valid");
-                    let code = set_grpc_status(error, rsp.headers_mut());
-                    debug!(?code, "Handling error with gRPC status");
-                    return Ok(rsp);
-                }
-
-                http::Version::HTTP_2
-            }
+            Respond::Http2 { .. } => http::Version::HTTP_2,
         };
 
         let status = http_status(error);
--- a/linkerd/app/core/src/errors.rs
+++ b/linkerd/app/core/src/errors.rs
@@ -17,7 +17,7 @@ pub struct NewRespond<B>(std::marker::PhantomData<fn() -> B>);
 
 #[derive(Copy, Clone, Debug)]
 pub enum Respond<B> {
-    Http1(std::marker::PhantomData<fn() -> B>),
+    Http1(http::Version, std::marker::PhantomData<fn() -> B>),
     Http2 { is_grpc: bool },
 }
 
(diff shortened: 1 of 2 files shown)


linkerd/app/core/src/errors.rs
```diff
@@ -26,15 +26,16 @@ impl<A, B: Default> respond::NewRespond<http::Request<A>> for NewRespond<B> {
...
-            Respond::Http2 { is_grpc }
-        } else {
-            Respond::Http1(self.0)
+        match req.version() {
+            http::Version::HTTP_2 => {
+                let is_grpc = req
+                    .headers()
+                    .get(http::header::CONTENT_TYPE)
+                    .and_then(|v| v.to_str().ok().map(|s| s.starts_with("application/grpc")))
+                    .unwrap_or(false);
```
hawkw: Nit/TIOLI: might be slightly neater to express this as
```rust
let is_grpc = req
    .headers()
    .get(http::header::CONTENT_TYPE)
    .and_then(|v| v.to_str().ok())
    .iter()
    .contains(|s| s.starts_with("application/grpc"));
```
it's the same amount of code, but i think separating the conversion to a string and the testing the string for the predicate shows the flow more clearly.

admittedly, this doesn't really matter at all! just a thought

linkerd/app/core/src/errors.rs
```diff
@@ -51,26 +52,33 @@ impl<B: Default> respond::Respond for Respond<B> {
...
+                if *is_grpc {
+                    let mut rsp = http::Response::builder()
+                        .version(http::Version::HTTP_2)
+                        .header(http::header::CONTENT_LENGTH, "0")
+                        .body(B::default())
+                        .expect("app::errors response is valid");
+                    let code = set_grpc_status(error, rsp.headers_mut());
+                    debug!(?code, "Handling error with gRPC status");
+                    return Ok(rsp);
+                }
```
hawkw: I find it a little weird that all this logic, with multiple early returns, happens in the assignment to `version`...but, not sure if there is a clearer way to express it, so this is fine

linkerd/proxy/http/src/orig_proto.rs
```diff
@@ -76,24 +77,43 @@ where
...
+            .and_then(|orig_proto| {
                 if orig_proto == "HTTP/1.1" {
-                    http::Version::HTTP_11
+                    Some(http::Version::HTTP_11)
                 } else if orig_proto == "HTTP/1.0" {
-                    http::Version::HTTP_10
+                    Some(http::Version::HTTP_10)
                 } else {
-                    warn!("unknown {} header value: {:?}", L5D_ORIG_PROTO, orig_proto);
-                    res.version()
```
hawkw: hmm, should we still warn when the orig_proto header had a weird value? we will record a "Downgrading response to None" event, but since it's at the DEBUG level, it probably won't show up in the logs for most users, and this is probably a sign that something is wrong...

linkerd/proxy/http/src/orig_proto.rs
```diff
@@ -107,6 +107,7 @@ where
                 } else if orig_proto == "HTTP/1.0" {
                     Some(http::Version::HTTP_10)
                 } else {
+                    debug!(%orig_proto, "Malformed header value");
```
hawkw: IMO this should probably be
```suggestion
                    warn!(%orig_proto, "Malformed header value");
```
The reason I suggested we continue to emit a separate event this case is because the event won't be seen with the default filter configuration.

olix0r: What information does this provide a user? What if we change the values a server may emit and an older client talks to a newer server? I think it's useful to be able to know this when debugging; but I don't think this adds any value to users.

hawkw: Okay, that's fair. I was mostly just thinking that if we thought this should be a warning at some point in the past, it should continue to be one, but your rationale makes sense.

linkerd/proxy/http/src/orig_proto.rs
```diff
@@ -76,24 +77,44 @@ where
...
+{
+    type Item = F::Item;
+    type Error = F::Error;
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        let mut res = try_ready!(self.inner.poll());
+        let version = res
+            .headers_mut()
+            .remove(L5D_ORIG_PROTO)
+            .and_then(|orig_proto| {
```
kleimkuhler: From above it looks like `orig_proto` could be either `HTTP/1.1` or `HTTP/1.1; absolute-form` (same for 1.0).

If the `L5D_ORIG_PROTO` header has the `..; absolute-form` ending then the `else` statement would set it back to `None`. That behavior isn't changed here, but is that an issue?

hawkw: I also noticed this, but after taking a closer look, I believe this isn't an issue. This future is the _response_ future, not the request, and "absolute form" only applies to requests. The request handling side _does_ handle the presence of `; absolute-form` in the `l5d-orig-proto` header.

It would be nice if there was a comment here so future readers don't have to hunt for this...

kleimkuhler: Okay cool I agree it doesn't sound like an issue if "absolute form" only applies to requests. Thanks! And yes a comment on this definitely wouldn't hurt.

PR #426, opened 2020-02-14
trace: update `tracing-subscriber` dependency to 0.2.1

hawkw:
This release makes a number of significant performance improvements when
filtering events emitted through the `log` compatibility layer (such as
from `hyper` and `h2`). See tokio-rs/tracing#586 for details.

Since these performance improvements apply to cases where a filter
_disables_ an event, as well as to enabling events, this should improve
the proxy's performance with the default log configuration, especially
given that `h2` and `hyper` emit a _lot_ of `trace`-level logs in hot
paths. Ignoring those logs should now have a much lower overhead.

In addition, 0.2.1 fixes a bug where `EnvFilter`'s filter selection
differed from the `env_logger` behavior it was intended to emulate. If
the most specific filter directive that applies to a given event did not
enable that  event's level, filter selection would continue to try
decreasingly-specific filters, and would only disable the event if
**no** filters enabled it (see tokio-rs/tracing#512).

In practice, this means that if you set a filter like

```
LINKERD2_PROXY_LOG=warn,linkerd=debug,linkerd2_metrics=info
```

in an attempt to disable the `debug`-level events in the
`linkerd2_metrics` crate, they would still be enabled by the
`linkerd=debug` directive.

This is now fixed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2565,9 +2565,9 @@ dependencies = [
 
 [[package]]
 name = "tracing-subscriber"
-version = "0.2.0-alpha.6"
+version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8178ab65c2ae72cd431fcdb51e7317fb0b48d64599e66d52c7b64cd86ee5b482"
+checksum = "c6bbfcf5ab0799649189b0df4bd7f730f1929a3846d33ae56cb0ad6e0bbdad09"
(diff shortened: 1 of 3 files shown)


hawkw:
preliminary latency benchmarks:
master-vs-subscriber-0_2_1latency
although, these were both run with `LINKERD2_PROXY_LOG=""`, and we might see a bigger difference with more filters.

PR #427, opened 2020-02-17
Reimplement the Lock middleware with tokio::sync

olix0r:
As described in tokio-rs/tokio#2237, the `tokio::sync::Semaphore` can
hold unbounded memory, especially when the semaphor is being contested
and consumers drop interest. Unfortunately, this use case is common in
the proxy, especially when a destination service is unavailable and the
proxy is timing out requests.

This change reimplements the Lock middleware without using
`tokio::sync::Semaphore`. This implementation is in some ways more
naive and inefficient, but it appears to be better suited for the
proxy's needs. Specifically, waiters are stored in a LIFO stack, which
optimizes for minimizing latency. Under certain high-load scenarios,
this Lock could be forced to grow its waiters set without cleaning up
expired watchers. If this becomes a more serious concern, we could
change the implementation to use a FIFO queue of waiters.


olix0r:
Though the lock isn't yet used in the proxy, I've tested another branch that uses the lock and confirmed that we don't see the unbounded memory growth observed with the previous version.

linkerd/lock/src/service.rs
```diff
@@ -0,0 +1,177 @@
+use crate::error::{Error, Poisoned, ServiceError};
+use crate::shared::{Shared, Wait};
+use futures::{future, Async, Future, Poll};
+use std::sync::{Arc, Mutex};
+use tracing::trace;
+
+/// Guards access to an inner service with a `tokio::sync::lock::Lock`.
```
seanmonstar: This is no longer the case, right?

linkerd/lock/src/service.rs
```diff
@@ -0,0 +1,177 @@
...
+                            shared.fail(error.clone());
+                        }
+                        State::Failed(error)
+                    }
+                },
+
+                // This instance has not registered interest in the lock.
+                State::Released => match self.shared.lock() {
+                    Err(_) => return Err(Poisoned::new().into()),
+                    // First, try to acquire the lock withotu creating a waiter.
```
seanmonstar:
```suggestion
                    // First, try to acquire the lock without creating a waiter.
```

linkerd/lock/src/shared.rs
```diff
@@ -0,0 +1,166 @@
...
+            self.0.register();
+        }
+
+        /// Returns true iff there is currently a `Notify` handle for this waiter.
+        pub fn is_waiting(&self) -> bool {
+            Arc::weak_count(&self.0) == 1
+        }
+
+        /// Returns true iff there is not currently a `Notify` handle for this waiter.
+        pub fn is_not_waiting(&self) -> bool {
```
seanmonstar: Is a weak count every expected to be something other than 1 or 0? If so, we should comment somewhere so we remember. If not, is this function needed? It's just `!wait.is_waiting()`. Or the implementation could just be that, `!self.is_waiting()`.

linkerd/lock/src/shared.rs
```diff
@@ -0,0 +1,166 @@
...
+                debug_assert!(wait.is_waiting());
+                Ok(Async::NotReady)
+            }
+            Err(error) => Err(error),
+        }
+    }
+
+    pub fn release_and_notify(&mut self, value: T) {
+        trace!(waiters = self.waiters.len(), "Releasing");
+        debug_assert!(match self.state {
```
seanmonstar: This assertion is to catch misuse of `Shared`? Or sanity-checking internals?

If it's to catch misuse, I prefer a regular `assert!`, since there may some cases it *is* misused and only noticed under load. If it's for internals, ignore me :D

linkerd/lock/src/shared.rs
olix0r: That's right. There may be at most one instance of `Notify` per `Wait` at any one time, as enforced by `get_notify`. I'll add `debug_assert`s to make this clearer.

linkerd/lock/src/test.rs
```diff
@@ -0,0 +1,254 @@
...
+                trace!("Dropping");
+                return Ok(Async::Ready(()));
+            }
+            self.0.poll_ready().map_err(|_| ())
+        }
+    }
+}
+
+#[test]
+fn fuzz() {
```
hawkw: You may want to consider using `loom` for this test. `loom` 0.1.x has [a mock version of `AtomicTask`](https://github.com/tokio-rs/loom/blob/2f735ffc5485edd98cb8158212dcf89bad1ad98b/src/futures/atomic_task.rs#L8-L15), which can simulate spurious wakeups, and a mock version of `std::thread`, which will run every permutation of threads being preempted (with bounding).

Using `loom` rather than randomness would make this task deterministic, which could make debugging failures significantly easier.

linkerd/lock/src/shared.rs
```diff
@@ -0,0 +1,166 @@
...
+        self.notify_next_waiter();
+    }
+
+    pub fn notify_next_waiter(&mut self) {
+        while let Some(waiter) = self.waiters.pop() {
+            if waiter.notify() {
+                trace!("Notified waiter");
+                return;
+            }
+        }
```
hawkw: If I'm correctly understanding this strategy, I think we may still be potentially susceptible to some memory leakiness when a very large number of waiters are in the stack. Since this is a LIFO, if there are a lot of waiters on the lock, those towards the end of the stack may never be woken up. If the delay is long enough, those waiters will time out, causing the futures to be dropped and thus dropping the `Arc`-ed waker, but the `Weak` pointer will remain in the stack. If a lot of new waiters are always being pushed, those `Weak`s will remain at the back of the stack forever.

Since we're dropping the actual waker when a future is dropped, the _size_ of the data remaining in the stack per waiter is significantly smaller than in `tokio_sync`'s `Semaphore` (I believe a `Weak` is just a single pointer). However, it _is_ still possible for the vec to have unbounded growth.

One thought is that we may want to periodically call `Vec::retain` or something, to drop any `Notify` instances in the stack where the waiter has been dropped. However, this is expensive — it requires iterating the whole vec — so for latency reasons, we probably don't want to do this when acquiring or releasing the lock. We could probably consider putting this work in a background task that will run when the lock is not busy?

We may also want to periodically call `shrink_to_fit` so that the actual allocated capacity of the vec is released when load decreases enough that we don't need as much capacity. However, we definitely don't want to do this _too_ often, as we could essentially thrash allocations. If we're occasionally cleaning out dead waiters, vec capacity is probably much less of an issue...

linkerd/lock/src/service.rs
```diff
@@ -4,7 +4,7 @@ use futures::{future, Async, Future, Poll};
 use std::sync::{Arc, Mutex};
 use tracing::trace;
 
-/// Guards access to an inner service with a `tokio::sync::lock::Lock`.
+/// A middlware that safely shares an inner service among clones.
```
hawkw: typo
```suggestion
/// A middleware that safely shares an inner service among clones.
```

linkerd/lock/src/shared.rs
```diff
@@ -76,15 +76,25 @@ impl<T> Shared<T> {
...
             _ => false,
         });
         self.state = State::Unclaimed(value);
         self.notify_next_waiter();
     }
 
-    pub fn notify_next_waiter(&mut self) {
+    pub fn release_waiter(&mut self, wait: Wait) {
+        // If a waiter is being released and it does not have a notify, then it must be being
+        // relased after being notified. Notify the next waiter to prevent deadlock.
```
hawkw: typo:
```suggestion
        // released after being notified. Notify the next waiter to prevent deadlock.
```
(are you having keyboard issues? :wink:)

linkerd/lock/src/test.rs
olix0r: Sounds cool! I'd rather not block the PR on this unless we have specific concerns.

linkerd/lock/src/shared.rs
olix0r: Yep. This is all correct. My current assumption is that what we have is Good Enough. We could certainly try to address these issues, but solutions all come with complexity. We can only really hit this situation when the lock is _always_ under contention. As soon as it becomes idle, the entire set of waiters will be cleaned up. It seems like it would be unlikely to encounter this in practice, though I'm sure that we could write a test that triggers this...

For now, I'd prefer to add a note about this and revisit if we see this type of usage in practice.

linkerd/lock/src/shared.rs
```diff
@@ -0,0 +1,166 @@
...
+            Err(error) => Err(error),
+        }
+    }
+
+    pub fn release_and_notify(&mut self, value: T) {
+        trace!(waiters = self.waiters.len(), "Releasing");
+        debug_assert!(match self.state {
+            State::Claimed => true,
+            _ => false,
+        });
```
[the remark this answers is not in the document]
olix0r: unless i'm mistaken, this would require that the inner service is PartialEq, which is not something i want to require

linkerd/lock/src/shared.rs
hawkw: > As soon as it becomes idle, the entire set of waiters will be cleaned up

Where does this happen? I must've missed it.

linkerd/lock/src/service.rs
```diff
@@ -0,0 +1,173 @@
...
+            State::Acquired(svc) => svc,
+            _ => panic!("Called before ready"),
+        };
+
+        let fut = svc.call(req);
+
+        // Return the service to the shared state, notifying waiters as needed.
+        //
+        // The service is dropped if the inner mutex has been poisoned, and
+        // subsequent calsl to poll_ready will return a Poisioned error.
```
hawkw: typos:
```suggestion
        // subsequent calls to poll_ready will return a Poisoned error.
```

linkerd/lock/src/service.rs
```diff
@@ -0,0 +1,173 @@
...
+        fut.map_err(Into::into)
+    }
+}
+
+impl<S> Drop for Lock<S> {
+    fn drop(&mut self) {
+        let state = std::mem::replace(&mut self.state, State::Released);
+        trace!(?state, "Dropping");
+        match state {
+            // If this lock was holding the service, return it back back to the
```
hawkw: typo:
```suggestion
            // If this lock was holding the service, return it back to the
```

linkerd/lock/src/shared.rs
hawkw: Oh, you mean that if the lock is released when there are no "live" waiters in the stack, we continue popping until we hit a "live" waiter, so if there are none, we'll pop everything.  Yeah, that means we probably won't see issues around the stack's capacity except in extreme edge cases. 👍

linkerd/lock/src/shared.rs
olix0r: more like brainworms

linkerd/lock/src/shared.rs
olix0r: Added a note onto the `Shared` docstring.

linkerd/lock/src/service.rs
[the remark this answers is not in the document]
kleimkuhler: Already fixed

PR #428, opened 2020-02-18
Expose the number of service instances in the proxy

olix0r:
When diagnosing issues, it can be difficult to reason about the states
of the various caches in the proxy.

This change introduces a set of `stack` metrics to help proxy developers
diagnose issues. The metrics track stack creations, stack deletions, and
the amount of time a given service is not available (via
`stack_poll_total_ms`)

```
:; kubectl exec  -n burn-http deploy/client-burst -c client -- curl -s localhost:4191/metrics |grep -e ^stack
stack_create_total{direction="outbound",name="fallback.endpoint"} 0
stack_create_total{direction="outbound",name="balance.endpoint"} 1
stack_create_total{direction="outbound",name="logical.dst"} 1
stack_create_total{direction="outbound",name="addr"} 1
stack_create_total{direction="outbound",name="source"} 100
stack_create_total{direction="inbound",name="endpoint"} 0
stack_create_total{direction="inbound",name="route"} 0
stack_create_total{direction="inbound",name="logical"} 0
stack_create_total{direction="inbound",name="source"} 0
stack_drop_total{direction="outbound",name="fallback.endpoint"} 0
stack_drop_total{direction="outbound",name="balance.endpoint"} 0
stack_drop_total{direction="outbound",name="logical.dst"} 0
stack_drop_total{direction="outbound",name="addr"} 0
stack_drop_total{direction="outbound",name="source"} 0
stack_drop_total{direction="inbound",name="endpoint"} 0
stack_drop_total{direction="inbound",name="route"} 0
stack_drop_total{direction="inbound",name="logical"} 0
stack_drop_total{direction="inbound",name="source"} 0
stack_poll_total{direction="outbound",name="fallback.endpoint",ready="true"} 0
stack_poll_total{direction="outbound",name="balance.endpoint",ready="true"} 70371
stack_poll_total{direction="outbound",name="logical.dst",ready="true"} 23600
stack_poll_total{direction="outbound",name="addr",ready="true"} 23600
stack_poll_total{direction="outbound",name="source",ready="true"} 23600
stack_poll_total{direction="inbound",name="endpoint",ready="true"} 0
stack_poll_total{direction="inbound",name="route",ready="true"} 0
stack_poll_total{direction="inbound",name="logical",ready="true"} 0
stack_poll_total{direction="inbound",name="source",ready="true"} 0
stack_poll_total{direction="outbound",name="fallback.endpoint",ready="false"} 0
stack_poll_total{direction="outbound",name="balance.endpoint",ready="false"} 23175
stack_poll_total{direction="outbound",name="logical.dst",ready="false"} 0
stack_poll_total{direction="outbound",name="addr",ready="false"} 1
stack_poll_total{direction="outbound",name="source",ready="false"} 0
stack_poll_total{direction="inbound",name="endpoint",ready="false"} 0
stack_poll_total{direction="inbound",name="route",ready="false"} 0
stack_poll_total{direction="inbound",name="logical",ready="false"} 0
stack_poll_total{direction="inbound",name="source",ready="false"} 0
stack_poll_total{direction="outbound",name="fallback.endpoint",ready="error"} 0
stack_poll_total{direction="outbound",name="balance.endpoint",ready="error"} 0
stack_poll_total{direction="outbound",name="logical.dst",ready="error"} 0
stack_poll_total{direction="outbound",name="addr",ready="error"} 0
stack_poll_total{direction="outbound",name="source",ready="error"} 0
stack_poll_total{direction="inbound",name="endpoint",ready="error"} 0
stack_poll_total{direction="inbound",name="route",ready="error"} 0
stack_poll_total{direction="inbound",name="logical",ready="error"} 0
stack_poll_total{direction="inbound",name="source",ready="error"} 0
stack_poll_total_ms{direction="outbound",name="fallback.endpoint"} 0
stack_poll_total_ms{direction="outbound",name="balance.endpoint"} 40
stack_poll_total_ms{direction="outbound",name="logical.dst"} 0
stack_poll_total_ms{direction="outbound",name="addr"} 2
stack_poll_total_ms{direction="outbound",name="source"} 0
stack_poll_total_ms{direction="inbound",name="endpoint"} 0
stack_poll_total_ms{direction="inbound",name="route"} 0
stack_poll_total_ms{direction="inbound",name="logical"} 0
stack_poll_total_ms{direction="inbound",name="source"} 0
```

### DIFF
--- a/linkerd/app/core/src/svc.rs
+++ b/linkerd/app/core/src/svc.rs
@@ -51,6 +51,10 @@ impl<L> Layers<L> {
         self.push_pending().push(buffer::layer(bound, d))
     }
 
+    pub fn push_per_make<T: Clone>(self, layer: T) -> Layers<Pair<L, stack::per_make::Layer<T>>> {
+        self.push(stack::per_make::layer(layer))
+    }
+
     pub fn push_spawn_ready(self) -> Layers<Pair<L, SpawnReadyLayer>> {
         self.push(SpawnReadyLayer::new())
     }
--- a/linkerd/stack-metrics/src/lib.rs
+++ b/linkerd/stack-metrics/src/lib.rs
@@ -1,12 +1,15 @@
 #![deny(warnings, rust_2018_idioms)]
 
-use futures::{Async, Poll};
+mod layer;
+mod service;
+
+pub use self::layer::TrackServiceLayer;
+pub use self::service::TrackService;
 use indexmap::IndexMap;
 use linkerd2_metrics::{metrics, Counter, FmtLabels, FmtMetrics};
 use std::fmt;
 use std::hash::Hash;
 use std::sync::{Arc, Mutex};
-use std::time::Instant;
 
 metrics! {
     stack_create_total: Counter { "Total number of services created" },
@@ -15,23 +18,10 @@ metrics! {
     stack_poll_total_ms: Counter { "Total number of milliseconds this service has spent awaiting readiness" }
 }
 
-type Registry<L> = Arc<Mutex<IndexMap<L, Arc<Metrics>>>>;
+type Shared<L> = Arc<Mutex<IndexMap<L, Arc<Metrics>>>>;
 
 #[derive(Debug)]
-pub struct NewLayer<L: Hash + Eq> {
-    registry: Registry<L>,
-}
-
-/// Reports metrics for prometheus.
-#[derive(Debug)]
-pub struct Report<L: Hash + Eq> {
-    registry: Registry<L>,
-}
-
-#[derive(Clone, Debug)]
-pub struct Layer {
-    metrics: Arc<Metrics>,
-}
+pub struct Registry<L: Hash + Eq>(Shared<L>);
 
 #[derive(Debug, Default)]
 struct Metrics {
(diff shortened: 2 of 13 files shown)


hawkw:
This seems like something that could potentially be a good use-case for the `tracing` metrics integration that some folks in the community are supposedly working on, but that's not here yet, and this works now...

linkerd/stack-metrics/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+                    self.metrics.poll_millis.add(not_ready.as_millis() as u64);
+                }
+                Ok(Async::Ready(()))
+            }
+            Err(e) => {
+                self.metrics.error_total.incr();
+                if let Some(t0) = self.blocked_since.take() {
+                    let not_ready = Instant::now() - t0;
+                    self.metrics.poll_millis.add(not_ready.as_millis() as u64);
+                }
```
hawkw: Any concerns about the performance implications of recording all these timestamps? It seems like if we are instrumenting every stack in the proxy with one of these layers, we are potentially adding a fair number of syscalls in the request path. 

It could be good to do some benchmarking and see if there's a noticeable perf impact from this? We might want to add a config to turn off the stack metrics, or something?

linkerd/app/core/src/svc.rs
```diff
@@ -51,6 +51,10 @@ impl<L> Layers<L> {
         self.push_pending().push(buffer::layer(bound, d))
     }
 
+    pub fn push_per_make<T: Clone>(self, layer: T) -> Layers<Pair<L, stack::per_make::Layer<T>>> {
```
hawkw: TIOLI: It might be worth adding a comment to explain the difference between `push` and `push_per_make`?

linkerd/stack-metrics/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+    stack_create_total: Counter { "Total number of services created" },
+    stack_drop_total: Counter { "Total number of services dropped" },
+    stack_poll_total: Counter { "Total number of stack polls" },
+    stack_poll_total_ms: Counter { "Total number of milliseconds this service has spent awaiting readiness" }
+}
+
+type Registry<L> = Arc<Mutex<IndexMap<L, Arc<Metrics>>>>;
+
+#[derive(Debug)]
+pub struct NewLayer<L: Hash + Eq> {
```
kleimkuhler: Similar to comment on #417, it feels like this can be similar to the types in that [change](https://github.com/linkerd/linkerd2-proxy/blob/d5cf282ab3297dba0e83e7c2054096c81b6fd46d/linkerd/error-metrics/src/lib.rs#L29-L37). `NewLayer` is a little surprising to see in `core` code when all the other metrics are not named this way.

linkerd/stack-metrics/src/lib.rs
olix0r: How is this different from any of the other latency metrics we record?

olix0r: Furthermore, this is only triggred when a service is NotReady, which is fairly rare in the scheme of things. I'm not particularly worried about this? Unless you really think we should be?

linkerd/app/core/src/svc.rs
olix0r: This will change more. We'll document it better as it stabilizes.

PR #430, opened 2020-02-18
http-box: Box HTTP payloads via middleware

olix0r:
The previous `http::BoxedService` implementation (1) boxed the inner
service, (2) boxed the service's errors, and (3) boxed the service's
HTTP response payload type. However, sometimes it is necessary to box
HTTP _request_ payload types; and other times it may not be
necessary/desirable to box the service type when boxing a response
payload.

This change extracts a new `http-box` crate with HTTP-payload-boxing
middlewares, and the prior general BoxService is moved to a dedicated
`box` crate.


linkerd/http-box/src/payload.rs
```diff
@@ -0,0 +1,113 @@
...
+        B: hyper::body::Payload + 'static,
+        B::Error: Into<Error>,
+    {
+        Self {
+            inner: Box::new(Inner(inner)),
+        }
+    }
+}
+
+impl hyper::body::Payload for Payload {
```
seanmonstar: May wish to override `content_length` as well, though for forwarded messages, the original headers take priority. :shrug:

PR #431, opened 2020-02-19
lock: Generalize to protect a guarded value

olix0r:
We used Tokio's Lock implementation in the router's cache
implementation, though we know it can leak under contention.

This change generalizes the Lock to return a guarded value (like Tokio's
lock). This change simplifies the Lock's state management: the Lock may
no longer hold a value, nor can it fail.

The `lock::Service` implementation now holds a `Result<Service,
ServiceError>` so that lock services may still broadcast the inner
service's failure.

### DIFF
--- a/linkerd/lock/src/lock.rs
+++ b/linkerd/lock/src/lock.rs
@@ -1,0 +1,122 @@
+use crate::shared::{Shared, Wait};
+use futures::Async;
+use std::sync::{Arc, Mutex};
+
+/// A middleware that safely shares an inner service among clones.
+///
+/// As the service is polled to readiness, the lock is acquired and the inner service is polled. If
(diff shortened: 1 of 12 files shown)


linkerd/lock/src/lock.rs
```diff
@@ -0,0 +1,122 @@
+use crate::shared::{Shared, Wait};
+use futures::Async;
+use std::sync::{Arc, Mutex};
+
+/// A middleware that safely shares an inner service among clones.
+///
+/// As the service is polled to readiness, the lock is acquired and the inner service is polled. If
+/// the service is cloned, the service's lock state isnot retained by the clone.
```
seanmonstar: Looks like these docs are meant for `LockService`.

PR #432, opened 2020-02-20
fallback: Operate on Services instead of Layers

olix0r:
The fallback middleware currently operates by applying a primary layer
to an underlying service and, if that primary service does not produce a
result, the fallback layer is applied to the underlying service to
produce a result.

This change modifies the middleware to no longer require that the
underlying service be shared by the primary and fallback layers.
Instead, the middleware attempts to produce a result from the underlying
service and, if that fails, the fallback service is used.

This will allow the fallback middleware to be used more flexibly in
upcoming changes.

### DIFF
--- a/linkerd/app/outbound/src/lib.rs
+++ b/linkerd/app/outbound/src/lib.rs
@@ -15,8 +15,8 @@ use linkerd2_app_core::{
     http_request_l5d_override_dst_addr, http_request_orig_dst_addr, metric_labels,
     opencensus::proto::trace::v1 as oc,
     proxy::{
-        self, core::resolve::Resolve, discover, fallback, http, identity, resolve::map_endpoint,
-        tap, tcp, Server,
+        self, core::resolve::Resolve, discover, http, identity, resolve::map_endpoint, tap, tcp,
+        Server,
     },
     reconnect, retry, router, serve,
     spans::SpanConverter,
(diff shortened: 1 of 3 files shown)


linkerd/fallback/src/lib.rs
```diff
@@ -1,175 +1,149 @@
...
 use linkerd2_error::Error;
-use tracing::trace;
+use tower::util::{Either, Oneshot, ServiceExt};
+use tracing::debug;
 
 /// A fallback layer composing two service builders.
 ///
 /// If the future returned by the primary builder's `MakeService` fails with
 /// an error matching a given predicate, the fallback future will attempt
 /// to call the secondary `MakeService`.
```
hawkw: I think this comment needs to be updated?

linkerd/fallback/src/lib.rs
```diff
@@ -1,175 +1,149 @@
...
-        self.primary.poll_ready().map_err(Into::into)
+        self.inner.poll_ready().map_err(Into::into)
     }
 
     fn call(&mut self, target: T) -> Self::Future {
-        MakeFuture {
-            fallback: self.fallback.clone(),
+        MakeFuture::A {
+            primary: self.inner.call(target.clone()),
+            fallback: Some(self.fallback.clone().oneshot(target)),
```
hawkw: Using `oneshot` here makes sense, good call!

linkerd/app/outbound/src/lib.rs
```diff
@@ -201,7 +182,8 @@ impl<A: OrigDstAddr> Config<A> {
             // Resolves the target via the control plane and balances requests
             // over all endpoints returned from the destination service.
             const DISCOVER_UPDATE_BUFFER_CAPACITY: usize = 10;
-            let balancer_layer = svc::layers()
+            let balancer_stack = endpoint_stack
+                .serves::<Endpoint>()
```
kleimkuhler: `endpoint_stack` has already verified it serves `Endpoint` typed targets. I'm not sure a redundant check is needed here.

PR #433, opened 2020-02-20
internal: Extract a service-profile crate

olix0r:
Service profile logic is unnecessarily split up among several crates.
This change centralizes it all in a `service-profiles` crate.

While doing this, I've cleaned up various stale/unneeded dependencies in
Cargos toml.

No functional changes.


Cargo.lock
```diff
@@ -1066,11 +1066,9 @@ dependencies = [
  "linkerd2-identity",
  "linkerd2-metrics",
  "linkerd2-proxy-transport",
- "linkerd2-router",
  "linkerd2-stack",
  "linkerd2-timeout",
  "rand 0.7.2",
- "regex 1.0.0",
```
hawkw: Is there a tool that you're using to prune unused deps from the Cargo.toml? If so, we might want to think about running it on CI. Not a blocker, just a thought — if we add this, it would be in a separate branch.

linkerd/service-profiles/src/client.rs
```diff
@@ -224,15 +224,15 @@ where
                     }
                     Err(e) => {
                         warn!("error fetching profile: {:?}", e);
-                        State::Backoff(Delay::new(clock::now() + self.backoff))
+                        State::Backoff(Delay::new(Instant::now() + self.backoff))
```
hawkw: At one point, I believe we were trying to use `tokio_timer::clock::now()` rather than `Instant::now()` everywhere. Is this something we still care about? IIRC, we aren't doing any large-scale integration testing with the mock clock...are there other reasons to prefer `clock::now()`?

Cargo.lock
olix0r: Yeah, we've talked about this before. `cargo-udeps` appears to require nightly, and its output is a bit unwieldy, so i'd want to get it integrated as a proper action or something...

linkerd/service-profiles/src/client.rs
olix0r: I think it only makes sense in contexts where we are actually mocking time. We generally don't do that.

PR #434, opened 2020-02-21
Increase inbound router capacity default to 10000

adleong:
Even though we expect there to be only a small number of pod-local HTTP services, clients may address the pod by a wide variety of names.  We increase the inbound router capacity so as to not fail requests when this is the case.


seanmonstar:
10k incoming variants? Wow!

adleong:
If the pod is part of a NodePort service, this should be at least the number of nodes in the cluster.  10k should generously cover it.

PR #435, opened 2020-02-22
Generalize router::Make as stack::NewService

olix0r:
The `router` crate provides a trait, `Make`, that supports immediate,
infallible service creation. We've typically used a `pending` layer to
modify `MakeService` instances as a `Make`.

These needs are not really router-specific, though: many middlewares are
infallible and immediate.

This change extracts `router::Make` as `stack::NewService` and replaces
the `pending` layer with `stack::new_service::FromMakeServiceLayer`.

Several other superficial changes have been made:

* The `stack::per_make` layer is now `stack::on_response`.
* The `buffer` middleware no longer provides a `MakeService`
  implementation and is, instead, applied through `OnResponse`.
* The unused `timeout::stack` is removed.
* The `stack::layer::LayerExt` trait is removed, in favor of the `svc`
  builders.
* The `fallback` crate is removed and the fallback middleware is moved
  into `stack::fallback`.

No functional changes.


linkerd/app/core/src/svc.rs
```diff
@@ -36,26 +33,22 @@ impl<L> Layers<L> {
...
         D: buffer::Deadline<Req>,
         Req: Send + 'static,
     {
-        self.push_pending().push(buffer::layer(bound, d))
+        self.push(buffer::layer(bound, d))
     }
 
-    pub fn push_per_make<T: Clone>(self, layer: T) -> Layers<Pair<L, stack::per_make::Layer<T>>> {
-        self.push(stack::per_make::layer(layer))
+    pub fn push_on_response<U>(self, layer: U) -> Layers<Pair<L, stack::OnResponseLayer<U>>> {
```
hawkw: 👍 this naming is a lot clearer IMO

linkerd/app/core/src/svc.rs
```diff
@@ -36,26 +33,22 @@ impl<L> Layers<L> {
         Layers(Pair::new(self.0, outer))
     }
 
-    /// Buffer requests when when the next layer is out of capacity.
-    pub fn push_pending(self) -> Layers<Pair<L, pending::Layer>> {
-        self.push(pending::layer())
+    /// Wraps an inner `MakeService` to be a `NewService`.
+    pub fn push_into_new_service(self) -> Layers<Pair<L, new_service::FromMakeServiceLayer>> {
+        self.push(new_service::FromMakeServiceLayer::default())
```
hawkw: bikeshed: IMO `pending` described the behavior of this thing a little clearer...yes, it turns a `MakeService` into a `NewService`, but that seems incidental to the underlying behavior of allowing a pending `Future<Service>` to be treated as a `Service`...I can learn to live with this terminology, though... not a blocker!

linkerd/stack/src/future_service.rs
```diff
@@ -0,0 +1,56 @@
...
+            };
+        }
+    }
+
+    fn call(&mut self, req: Req) -> Self::Future {
+        if let Inner::Service(ref mut svc) = self.inner {
+            return svc.call(req).map_err(Into::into);
+        }
+
+        panic!("pending not ready yet");
```
hawkw: nit: if we're renaming this, should probably be
```suggestion
        panic!("FutureService not ready yet");
```

linkerd/app/outbound/src/lib.rs
```diff
@@ -233,12 +237,18 @@ impl<A: OrigDstAddr> Config<A> {
...
 
                 distributor
-                    .serves::<DstAddr>()
-                    .push_buffer_pending(buffer.max_in_flight, DispatchDeadline::extract)
-                    .makes::<DstAddr>()
+                    .check_service::<DstAddr>()
+                    .into_new_service()
+                    .push_on_response(
+                        svc::layers().push_buffer(buffer.max_in_flight, DispatchDeadline::extract),
+                    )
```
hawkw: Is there a difference between 
```rust
.push_on_response(
    svc::layers().push_buffer(buffer.max_in_flight, DispatchDeadline::extract),
)
```
and 
```rust
.push_on_response(
    buffer::layer(buffer.max_in_flight, DispatchDeadline::extract)
)
```
it seems like these _should_ be equivalent, but I wanted to be sure this is correct?

olix0r: they're the same

linkerd/app/core/src/svc.rs
olix0r: > but that seems incidental to the underlying behavior of allowing a pending Future<Service> to be treated as a Service

That's not really what it does though? It explicitly modeled a MakeService as a Make (now, NewService). It's an adapter type.

"pending", to me, sounds much more about readiness

linkerd/router/src/lib.rs
```diff
@@ -115,45 +99,45 @@ where
...
-impl<Req, Rec, Mk> Router<Req, Rec, Mk>
+impl<Req, Rec, N> Router<Req, Rec, N>
 where
     Rec: Recognize<Req>,
-    Mk: Make<Rec::Target>,
-    <Mk as Make<Rec::Target>>::Value: Clone,
-    Mk::Value: tower::Service<Req>,
+    N: NewService<Rec::Target>,
+    <N as NewService<Rec::Target>>::Service: Clone,
+    N::Service: tower::Service<Req>,
```
kleimkuhler: Can just be:

```
N::Service: tower::Service<Req> + Clone,
```

linkerd/app/core/src/svc.rs
```diff
@@ -177,22 +171,34 @@ impl<S> Stack<S> {
         self.push(http::boxed::response::Layer::new())
     }
 
+    pub fn push_map_target<M: Clone>(self, map: M) -> Stack<stack::MapTargetService<S, M>> {
+        self.push(stack::MapTargetLayer::new(map))
+    }
+
     /// Validates that this stack serves T-typed targets.
-    pub fn makes<T>(self) -> Self
+    pub fn check_new_service<T>(self) -> Self
```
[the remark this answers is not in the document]
olix0r: I considered that, but `assert` is a runtime failure, whereas this is a static assertion.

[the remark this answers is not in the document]
olix0r: Could use `ensure_`...

PR #436, opened 2020-02-23
integration: Make the test controller more realistic

olix0r:
Many of the integration tests relied on `destination_and_close`; but
the real controller doesn't work this way. Instead, it keeps response
streams open indefinitely in non-error cases.

This change modifies uses of the mock controller to behave more
realistically. Some other improvements (logging, removing a
potentially infinite loop, etc) have crept in here, too.


linkerd/app/integration/tests/telemetry.rs
```diff
@@ -79,7 +80,12 @@ impl TcpFixture {
     }
 
     fn inbound() -> Self {
-        let proxy = proxy::new().inbound(TcpFixture::server()).run();
+        let ctrl = controller::new();
+        //ctrl.profile_tx_default("tele.test.svc.cluster.local");
```
hawkw: should this be removed?

linkerd/app/integration/tests/telemetry.rs
```diff
@@ -91,7 +97,12 @@ impl TcpFixture {
     }
 
     fn outbound() -> Self {
-        let proxy = proxy::new().outbound(TcpFixture::server()).run();
+        let ctrl = controller::new();
+        //ctrl.profile_tx_default("tele.test.svc.cluster.local");
```
hawkw: similarly, should this be removed?

linkerd/app/profiling/src/bin/profile.rs
```diff
@@ -25,9 +25,11 @@ fn main() {
 
     let srv = server::mock_listening(addr.clone());
     let srv2 = server::mock_listening(addr.clone());
-    let ctrl = controller::new()
-        .destination_and_close("transparency.test.svc.cluster.local", srv.addr)
-        .run();
+
+    let ctrl = controller::new();
+    let transparency_tx = ctrl.destination_tx("transparency.test.svc.cluster.local");
+    transparency_tx.send_addr(srv.addr);
```
hawkw: should we also call `profile_tx_default` here to avoid churning service profile resolutions?

linkerd/app/integration/src/controller.rs
```diff
@@ -327,40 +326,43 @@ where
...
+    let (trace, _) = trace_init();
+    std::thread::Builder::new()
         .name(name.into())
         .spawn(move || {
-            if let Some(delay) = delay {
-                let _ = listening_tx.take().unwrap().send(());
-                delay.wait().expect("support server delay wait");
-            }
-            let mut runtime = runtime::current_thread::Runtime::new().expect("support runtime");
+            tracing::dispatcher::with_default(&trace, move || {
```
hawkw: we could also use `set_default` here, which returns a drop guard, and avoid indenting everything...it's the same change, but makes a smaller diff. I should've mentioned that in my earlier review.

hawkw: (not a blocker, just a note)

linkerd/app/integration/tests/discovery.rs
```diff
@@ -450,6 +465,10 @@ macro_rules! generate_tests {
                             .label("load_profile", "bar"),
                     ], None, vec![]));
 
+                    ctrl
+                        .destination_tx(FOO).send_addr(foo.addr);
```
kleimkuhler: Odd formatting?

PR #441, opened 2020-02-25
stack: Introduce the Proxy trait

olix0r:
The `stack::Proxy` trait models middlewares that may modify the request
or response types but do not need to participate in the inner Service's
readiness.

This will be an important component of a new `service-profiles`
implementation: each request route does not need an independent clone of
the underlying concrete service (i.e., that it must drive to readiness
independently). Instead, the profile router can be responsible for
driving a single concrete router to readiness and invoke the ready
service through the appropriate route's `Proxy`.

Implementations will follow.


linkerd/stack/src/proxy.rs
```diff
@@ -0,0 +1,87 @@
+use futures::{future, Future, Poll};
+use linkerd2_error::Error;
+
+/// A middleware type that cannot exert backpressure.
+///
+/// Typically used to modify requests or responses.
+pub trait Proxy<Req, S: tower::Service<Self::Request>> {
```
hawkw: This is, admittedly, extremely bikesheddy, but I'm not a huge fan of calling this a `Proxy`...my immediate assumption upon seeing something called a `Proxy` is that it's bundling together all the behavior of a inbound or outbound side of the proxy, or an entire linkerd instance.

I'm not sure what to suggest as an alternative though. My first thought on reading "A middleware type that cannot exert backpressure" is `ReadyService`, but that's been used in the past to mean something different...it's basically like a `map` operation for requests/responses, but I don't know if it makes sense to call the trait `Map`...I almost want to suggest `Filter`...

I'm probably overthinking it, and the name `Proxy` is fine...

linkerd/stack/src/proxy.rs
```diff
@@ -0,0 +1,87 @@
...
+    type Error: Into<Error>;
+
+    /// The Future type returned to callers.
+    type Future: Future<Item = Self::Response, Error = Self::Error>;
+
+    /// Usually invokes `S::call`, potentially modifying requests or responses.
+    fn proxy(&self, inner: &mut S, req: Req) -> Self::Future;
+
+    /// Wraps an `S` typed service with the proxy.
+    fn into_service(self, inner: S) -> ProxyService<Self, S>
```
hawkw: also admittedly very bikesheddy, but I kind of expect `into_` methods to be conversions that only take `self`. i might've called this `wrap` or `and_then` or something...

linkerd/stack/src/proxy.rs
olix0r: Is `ServiceProxy` any better?

hawkw: wordier but clearer, IMO!

kleimkuhler: Then is the service it produces `ServiceProxyService`?

kleimkuhler: I kind of lean towards the original `Proxy`, but definitely not a blocker for `ServiceProxy`

hawkw: `ServiceProxy` is the trait, `ProxyService` is the `Service`, obviously! :P

PR #442, opened 2020-02-26
timeout: Do not synthesize HTTP response

olix0r:
The `proxy::http::timeout` middleware currently creates HTTP responses
when a timeout is encountered. Instead, we should allow the error to
propagate through the stack as an error so that all synthesized
responses happen in one place.

This change means that we no longer annotate a `status_code` label in
the `route_response_total` metric, though the `classification` and
`error` labels remain unchanged.

We can also avoid using a `tower::util::Either` by encoding the optional
behavior within the implementation (cutting down on compile-time type
bloat).

### DIFF
--- a/linkerd/proxy/http/src/timeout.rs
+++ b/linkerd/proxy/http/src/timeout.rs
@@ -1,9 +1,8 @@
-use futures::{future, try_ready, Future, Poll};
-use http::{Request, Response, StatusCode};
-use linkerd2_error::Error;
-use linkerd2_timeout::{error, Timeout};
+use futures::{try_ready, Future, Poll};
+use linkerd2_stack::NewService;
+pub use linkerd2_timeout::error;
+use linkerd2_timeout::Timeout;
 use std::time::Duration;
-use tracing::{debug, error};
 
 /// Implement on targets to determine if a service has a timeout.
 pub trait HasTimeout {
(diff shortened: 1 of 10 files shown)


hawkw:
> This change means that we no longer annotate a `status_code` label in
> the `route_response_total` metric

Do we know if anything was actually consuming this label?

olix0r:
@hawkw Not that I know of. We use `classification` for success rate calculations, etc.

linkerd/proxy/http/src/timeout.rs
```diff
@@ -34,28 +29,35 @@ pub struct MakeFuture<F> {
...
-impl<T, M> tower::Service<T> for Stack<M>
+impl<T, M> tower::Service<T> for MakeTimeout<M>
 where
     M: tower::Service<T>,
     T: HasTimeout,
 {
-    type Response = tower::util::Either<Service<M::Response>, M::Response>;
+    type Response = Timeout<M::Response>;
     type Error = M::Error;
     type Future = MakeFuture<M::Future>;
```
kleimkuhler: To go along with the renamings taking place, maybe call hits `MakeTimeoutFuture` now?

olix0r: I think _MakeFuture_ is descriptive enough -- we typically use _ResponseFuture_ in other contexts... I may change my mind on this later ;)

PR #444, opened 2020-02-26
orig-proto: Be more flexible to stack placement

olix0r:
The orig-proto Upgrade middleware unnecessarily captures a PhantomData;
and does URI normalization even though we have another dedicated
middleware for that.

This change modifies orig-proto::Upgrade to use the stack configuration
to determine whether the request's URI was absolute (these requests may
not share a Hyper client, so they need to be on different stacks). This
module no longer normalizes URIs.

The inbound::orig_proto_downgrade module is removed, as it mostly
boilerplate.


linkerd/app/outbound/src/orig_proto_upgrade.rs
```diff
@@ -1,68 +1,72 @@
...
+        Self::default()
     }
 }
 
-impl<M, A, B> svc::Layer<M> for Layer<A, B>
-where
-    M: svc::MakeService<Endpoint, http::Request<A>, Response = http::Response<B>>,
-{
-    type Service = MakeSvc<M, A, B>;
+impl<M> svc::Layer<M> for OrigProtoUpgradeLayer {
```
kleimkuhler: Not a blocker, but any reason `app::core` rexports `tower::layer::Layer`? We could just use `tower::layer::Layer` direclty here and down below for `Either`. There's only a few other places we use the rexported types/traits.

olix0r: Classically, `svc` provided all of the service primitives to app::core. It doesn't really need to stay that way, though.

linkerd/proxy/http/src/normalize_uri.rs
```diff
@@ -32,6 +32,22 @@ pub fn layer<M>() -> impl tower::layer::Layer<M, Service = MakeNormalizeUri<M>>
...
+impl<T, M> NewService<T> for MakeNormalizeUri<M>
+where
+    T: ShouldNormalizeUri,
+    M: NewService<T>,
+{
+    type Service = NormalizeUri<M::Service>;
+
+    fn new_service(&self, target: T) -> Self::Service {
+        let authority = target.should_normalize_uri();
+        tracing::trace!(?authority, "new");
```
hawkw: nit: here, we use `fmt::Debug` to record the authority, but in `NormalizeUri::call`, we record it with `fmt::Display`: https://github.com/linkerd/linkerd2-proxy/blob/c58ffc9ab43ab5ad05526180c6e82529fe37adb2/linkerd/proxy/http/src/normalize_uri.rs#L118

we should probably pick one and stick with it, to make this more consistent (so you can grep an authority in the logs, etc)?

linkerd/app/outbound/src/orig_proto_upgrade.rs
```diff
@@ -1,68 +1,72 @@
...
+    fn new_service(&self, mut endpoint: Endpoint) -> Self::Service {
+        if !endpoint.can_use_orig_proto() {
+            trace!("Endpoint does not support transparent HTTP/2 upgrades");
+            return svc::Either::B(self.inner.new_service(endpoint));
         }
+
+        let was_absolute = endpoint.http_settings.was_absolute_form();
+        trace!(
+            header = %orig_proto::L5D_ORIG_PROTO,
+            %was_absolute,
```
hawkw: nit: `was_absolute` is a `bool`, right? `bool` implements `tracing::field::Value`, so we don't need to use `fmt::Display`:
```suggestion
            was_absolute,
```

linkerd/proxy/http/src/normalize_uri.rs
olix0r: This probably should go away anyway, as it's all on the trace context already

olix0r:
Note that this branch should have 2d68d8500e1b0017aae2d15e571bfcf96339e101, but github seems to be having some consistency issues at the moment.

PR #445, opened 2020-02-26
inbound: Box client transports to ensure shutdown

olix0r:
The inbound client may never initiate a TLS connection, but we use a TLS
middleware to ensure the underlying transport is boxed. The BoxedIo type
enforces shutdown semantics differently than the raw TcpStream, and we
rely on this behavior in TCP forwarding.

This change introduces a stack::map_response middleware; and replaces
use of the inbound TLS client with a layer that boxes transports.

### DIFF
--- a/linkerd/app/core/src/svc.rs
+++ b/linkerd/app/core/src/svc.rs
@@ -185,6 +185,10 @@ impl<S> Stack<S> {
         self.push(stack::MapTargetLayer::new(map))
     }
 
+    pub fn push_map_response<M: Clone>(self, map: M) -> Stack<stack::MapResponse<S, M>> {
+        self.push(stack::MapResponseLayer::new(map))
+    }
+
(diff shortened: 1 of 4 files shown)


linkerd/app/core/src/svc.rs
```diff
@@ -185,6 +185,10 @@ impl<S> Stack<S> {
         self.push(stack::MapTargetLayer::new(map))
     }
 
+    pub fn push_map_response<M: Clone>(self, map: M) -> Stack<stack::MapResponse<S, M>> {
```
hawkw: It would be nice if we could bound `M` here, but I'm guessing that would mean introducing a bunch of other bounds so we can know the response type, which is less nice?

olix0r: Yeah. Practically, i've been adding type constraints when i debug compilation issues but remove them before merging, since, for instance, parameterizing on the target type or response (body) makes the layer useful in only a single stack context.

PR #447, opened 2020-02-26
outbound: Relax type constraints in require_identity_on_endpoint

olix0r:
Makes the naming uniform with newer stack modules and relaxes type
constraints to avoid capturing body types in the Layer (i.e. so the
layer can be used across stacks of varying body types).

### DIFF
--- a/linkerd/app/outbound/src/require_identity_on_endpoint.rs
+++ b/linkerd/app/outbound/src/require_identity_on_endpoint.rs
@@ -10,69 +10,75 @@ use linkerd2_app_core::{
     transport::tls::{self, HasPeerIdentity},
     Conditional, Error, L5D_REQUIRE_ID,
 };
-use std::marker::PhantomData;
 use tracing::debug;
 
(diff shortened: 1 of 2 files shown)


linkerd/app/outbound/src/require_identity_on_endpoint.rs
```diff
@@ -10,69 +10,75 @@ use linkerd2_app_core::{
...
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
         self.inner.poll_ready()
     }
 
-    fn call(&mut self, target: Endpoint) -> Self::Future {
+    fn call(&mut self, target: T) -> Self::Future {
         // After the inner service is made, we want to wrap that service
         // with a filter that compares the target's `peer_identity` and
         // `l5d_require_id` header if present
```
kleimkuhler: Looks like you can remove this comment as part of this if you want. The comment below it is basically a duplicate but more accurate.

PR #451, opened 2020-03-04
Introduce a backpressure-propagating buffer

olix0r:
The tower buffer implementation uses the underlying channel for
backpressure. This admits requests before the inner service becomes
ready, and doesn't fully propagate error conditions.

This replacement buffer implementation exposes readiness via a `Watch`.
This prevents callers from sending requests when the inner service is
not ready and allows service errors to be propagated reliably.  The
buffer's background task drops the inner service as soon as it fails.

The effect of this is that the buffer's channel is only used for
requests that race to submit requests while the inner service is ready,
and so it seems likely that they only need enough slots to satisfy the
proxy's expected concurrency (i.e. in terms of cores).


olix0r:
@carllerche's feedback is that the delay-based probing may be unnecessary if inner services are responsible for notifying the task when they need to change (even when they return Ready).

olix0r:
I've removed the probing logic from the buffer and moved it into a [dedicated middleware](https://github.com/linkerd/linkerd2-proxy/pull/452) -- there's no reason it needs to be coupled to the buffer.

linkerd/buffer/src/dispatch.rs
```diff
@@ -0,0 +1,127 @@
...
+                // If a request was ready, spawn its response future.
+                Ok(Async::Ready(Some(InFlight { request, tx }))) => {
+                    trace!("Dispatching a request");
+                    let inner = self.inner.as_mut().expect("Service must not be dropped");
+                    tokio::spawn(inner.call(request).then(move |res| {
+                        let _ = tx.send(res.map_err(Into::into));
+                        Ok(())
+                    }));
+                }
+            }
```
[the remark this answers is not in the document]
olix0r: good idea!

linkerd/buffer/src/dispatch.rs
```diff
@@ -112,14 +112,15 @@ where
                 // The sender has been dropped, complete.
                 Err(_) | Ok(Async::Ready(None)) => return Ok(Async::Ready(())),
 
-                // If a request was ready, spawn its response future.
+                // If a request was ready return it to the caller.
```
hawkw: nice, i'm glad this actually worked!

linkerd/buffer/src/dispatch.rs
```diff
@@ -0,0 +1,136 @@
...
+
+                // If a request was ready return it to the caller.
+                Ok(Async::Ready(Some(InFlight { request, tx }))) => {
+                    trace!("Dispatching a request");
+                    let fut = self
+                        .inner
+                        .as_mut()
+                        .expect("Service must not be dropped")
+                        .call(request);
+                    let _ = tx.send(Ok(fut));
```
hawkw: how often do we think requesters will hang up after requests are dispatched? is this something that we would want to record a TRACE/DEBUG event for?

linkerd/buffer/src/service.rs
```diff
@@ -0,0 +1,104 @@
...
+pub struct Buffer<Req, F> {
+    /// Updates with the readiness state of the inner service. This allows the buffer to reliably
+    /// exert backpressure and propagate errors, especially before the inner service has been
+    /// initialized.
+    ready: watch::Receiver<Poll<(), ServiceError>>,
+
+    /// The queue on which in-flight requests are sent to the inner service.
+    ///
+    /// Because the inner service's status is propagated via `ready` watch, this is here to
+    /// accomodate for when multiple services race to send a requesto
```
kleimkuhler: typo: `requesto`

linkerd/buffer/src/dispatch.rs
```diff
@@ -0,0 +1,136 @@
...
+                // requests and then complete.
+                Err(error) => {
+                    let shared = ServiceError(Arc::new(error.into()));
+                    trace!(%shared, "Inner service failed");
+
+                    // First, notify services of the readiness change to prevent new requests from
+                    // being buffered.
+                    let is_active = self.ready.broadcast(Err(shared.clone())).is_ok();
+
+                    // Propagate the error to all in-
```
kleimkuhler: Propogate the error to all in-flight requests?

linkerd/buffer/src/dispatch.rs
```diff
@@ -0,0 +1,136 @@
...
+                    }
+
+                    // Drop the inner Service to free its resources. It won't be used again.
+                    self.inner = None;
+
+                    // Ensure the task remains active until all services have observed the error.
+                    return_ready_if!(!is_active);
+
+                    // This is safe because ready.poll_close has returned NotReady.
+                    return Ok(Async::NotReady);
```
kleimkuhler: Once a service has errored will it enter a non-error state? I would expect this to return `Ok(Async::Ready)`;

kleimkuhler: On further thought this may make sense since as the comment mentions, `ready.poll_close()` returned `NotReady` above so after broadcasting an error to receivers this will be polled again once everything closes.

linkerd/buffer/src/service.rs
olix0r: No, I'm pretty sure that's right.

linkerd/buffer/src/dispatch.rs
olix0r: That's right!

linkerd/buffer/src/dispatch.rs
olix0r: This will happen whenever a request is canceled (i.e. the client resets the stream or drops the connection). Could go either way

PR #453, opened 2020-03-04
Overhaul buffering & caching to better-support backpressure

olix0r:
This encompasses a number of related changes:

* The `router` crate has been split up, decoupling its `cache` from its
  request-routing functionality. The `router` crate now _only_ handles
  request routing and not caching.
* The `cache` crate now exposes a `MakeService<T, Req>` (rather than a
  `Service<Req>`, returning clones of each inner service).
* `lock` is now used in place of a buffer in many places. This is done
  only to satisfy cloneability and will likely be changed (again) in a
  followup.
* Service profiles no longer use routers/buffers internally. The
  concrete router picks an
* Many middlewares have new Proxy implementations, some of which replace
  prior Service implementations (liek retry).
* DNS refinement is now modled as a `Service` so it can use the same
  caching infrastructure.
* Fallback is restructured to, for instance, avoid using a service
  profile stack when it cannot be created successfully.

This change dramatically improves compilation times, likely due to
simpler type constraints in the service profile middleware.


hawkw:
@olix0r i notice that there were a number of changes to the `Addr` types and to metrics labels that don't seem _directly_ linked to any of the changes you mentioned in the PR description. It's not a blocker but I'm curious about what changed & why.

linkerd/app/core/src/svc.rs
```diff
@@ -187,42 +278,50 @@ impl<S> Stack<S> {
         self.push(http::boxed::response::Layer::new())
     }
 
-    pub fn push_map_target<M: Clone>(self, map: M) -> Stack<stack::MapTargetService<S, M>> {
-        self.push(stack::MapTargetLayer::new(map))
-    }
-
-    pub fn push_map_response<M: Clone>(self, map: M) -> Stack<stack::MapResponse<S, M>> {
-        self.push(stack::MapResponseLayer::new(map))
+    /// Validates that this stack serves T-typed targets.
```
hawkw: "serves", or "makes new services for"?

linkerd/app/core/src/trace.rs
```diff
@@ -9,6 +7,8 @@ use tracing_subscriber::{
     reload, EnvFilter, FmtSubscriber,
 };
 
+const ENV_LOG: &str = "LINKERD2_PROXY_LOG";
```
hawkw: did rustfmt decide to move this around?

linkerd/app/inbound/src/endpoint.rs
```diff
@@ -116,117 +212,119 @@ impl tap::Inspect for Endpoint {
...
+                None => info_span!(
+                    "http2",
+                    port = %self.addr.port(),
+                ),
+                Some(name) => info_span!(
+                    "http2",
+                    %name,
+                    port = %self.addr.port(),
+                ),
+            },
```
hawkw: if we update to the latest `tracing` release, we could do
```rust
let span = info_span!("http2", port = %self.addr.port(), name = tracing::field::Empty);
span.record("name", &tracing::field::display(name));
```
which would let us have a single span that may or may not have a `name` field. not sure if the resultant code is actually any clearer, though....

linkerd/app/integration/tests/discovery.rs
```diff
@@ -615,16 +585,15 @@ macro_rules! generate_tests {
...
                 assert_eventually_contains!(metrics.get("/metrics"), "rt_hello=\"foo\"");
 
-                // Request 2 --- with override header
-                let res = override_req(&client);
-                assert_eq!(res.status(), http::StatusCode::OK);
-                assert_eventually_contains!(metrics.get("/metrics"), "rt_hello=\"bar\"");
+                // // Request 2 --- with override header
+                // let res = override_req(&client);
+                // assert_eq!(res.status(), http::StatusCode::OK);
+                // assert_eventually_contains!(metrics.get("/metrics"), "rt_hello=\"bar\"");
```
hawkw: is there a reason this is commented out?

linkerd/cache/src/cache.rs
```diff
@@ -109,7 +105,7 @@ where
     ///
     /// Polls the underlying `DelayQueue`. When elements are returned from the
     /// queue, remove the associated key from `values`.
-    pub fn purge(&mut self) {
+    pub fn poll_purge(&mut self) {
```
hawkw: nit/tioli: it still seems to me like a method named `poll_foo` should return a `Poll`...

linkerd/cache/src/error.rs
```diff
@@ -0,0 +1,12 @@
+use std::fmt;
+
+#[derive(Debug)]
+pub struct NoCapacity(pub usize);
+
+impl fmt::Display for NoCapacity {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "router capacity reached ({})", self.0)
```
hawkw: nit: should this now be
```suggestion
        write!(f, "cache capacity reached ({})", self.0)
```

linkerd/cache/src/layer.rs
```diff
@@ -0,0 +1,72 @@
...
+struct Config {
+    capacity: usize,
+    max_idle_age: Duration,
+}
+
+/// A layer that that builds a routing service.
+///
+/// A `Rec`-typed `Recognize` instance is used to produce a target for each
+/// `Req`-typed request. If the router doesn't already have a `Service` for this
+/// target, it uses a `Mk`-typed `Service` stack.
```
hawkw: nit: this looks like the same comment the old router layer...maybe:
```suggestion
/// A layer that that builds a cache service.
```
(and perhaps add some notes on the cache's eviction policy or something?)

linkerd/proxy/api-resolve/src/resolve.rs
```diff
@@ -74,15 +71,15 @@ where
 
     fn call(&mut self, target: T) -> Self::Future {
         let path = target.to_string();
-        trace!("resolve {:?}", path);
+        debug!(dst = %path, context=%self.context_token, "Resolving");
```
hawkw: nit: weird spacing
```suggestion
        debug!(dst = %path, context = %self.context_token, "Resolving");
```

linkerd/proxy/api-resolve/src/resolve.rs
```diff
@@ -112,6 +109,7 @@ where
                             .filter_map(|addr| pb::to_addr_meta(addr, &metric_labels))
                             .collect::<Vec<_>>();
                         if !addr_metas.is_empty() {
+                            debug!(endpoints = %addr_metas.len(), "Add");
```
hawkw: nit: if `len` is a usize, we shouldn't need to use `Display`. not a big deal in practice, but we could theoretically use the typed integer value for e.g. a metric of endpoint counts or something, in the future...
```suggestion
                            debug!(endpoints = addr_metas.len(), "Add");
```

linkerd/service-profiles/src/http/concrete.rs
```diff
@@ -0,0 +1,181 @@
...
+            Routes::Forward(Some(addr)) => write!(f, "Routes::Forward({})", addr),
+            Routes::Override { overrides, .. } => {
+                write!(f, "Routes::Override(")?;
+                let mut addrs = overrides.iter();
+                if let Some(a) = addrs.next() {
+                    write!(f, "{}", a)?;
+                }
+                for a in addrs {
+                    write!(f, ", {}", a)?;
+                }
```
hawkw: tioli: i think this could also be accomplished with `Formatter::debug_list`?

linkerd/app/inbound/src/lib.rs
```diff
@@ -103,210 +94,188 @@ impl<A: OrigDstAddr> Config<A> {
...
-                .push(insert::target::layer())
-                // disabled due to information leagkage
-                //.push(set_remote_ip_on_req::layer())
-                //.push(set_client_id_on_req::layer())
+                // Caches profile stacks.
+                .check_new_clone_service::<Profile>()
+                .spawn_cache(cache_capacity, cache_max_idle_age)
+                .instrument(|p: &Profile| info_span!("profile", addr = %p.addr()))
+                .check_service::<Profile>()
+                .push(router::Layer::new(|()| ProfileTarget))
```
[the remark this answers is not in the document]
kleimkuhler: Answered directly.

linkerd/router/src/lib.rs
```diff
@@ -1,577 +1,145 @@
...
-pub trait Recognize<Request> {
-    /// Identifies a Route.
-    type Target: Clone + Eq + Hash;
+pub trait Key<T> {
+    type Key: Clone;
 
-    /// Determines the target for a route to handle the given request.
-    fn recognize(&self, request: &Request) -> Option<Self::Target>;
+    fn key(&self, t: &T) -> Self::Key;
 }
```
hawkw: nitpicky, but: using the name `Key` for the trait that returns a key is not exactly my favorite naming...especially since the associated type is _also_ called `Key`...IMO, the old `Recognize` was more descriptive for the trait. that said...i acknowledge that this may not actually matter.

linkerd/service-profiles/src/http/service.rs
```diff
@@ -0,0 +1,379 @@
...
+            concrete: Override { service, update },
+        };
+
+        trace!("overriding profile service ready");
+        Ok(svc.into())
+    }
+}
+
+mod sealed {
+    pub trait PollUpdate {
```
hawkw: it would be nice if there was a comment here explaining what this is for...its purpose is not very obvious.

PR #456, opened 2020-03-06
cache: Let services self-evict

olix0r:
This change modifies the `cache` crate to remove its background task
which is responsible for tracking service use and evicting idle
services.

Now, the cache (1) has no capacity limit and (2) evicts services from
the cache when they drop their underlying service.

This is accomplished by modifying the `CacheLayer` to accept a layer
that is responsible for making the inner service shareable and for
dropping its inner service when it should be considered defunct. This is
accomplished by moving an `Arc<()>` into the inner stack and holding a
`Weak<()>` reference to it in the cache module. Then, before a request
is processed by the cache, it attempts to upgrade each cache entry and
drops services that have dropped the inner service (holding the strong
reference).


linkerd/app/inbound/src/lib.rs
```diff
@@ -181,14 +197,24 @@ impl<A: OrigDstAddr> Config<A> {
...
+                            // TODO: self-evicting buffers
+                            // // If the service has been ready & unused for `cache_max_idle_age`,
+                            // // fail it.
+                            // .push_idle_timeout(cache_max_idle_age)
+                            // // If the service has been unavailable for an extend time, eagerly
+                            // // fail requests.
+                            // .push_failfast(Duration::from_secs(10))
+                            // // Shares the service, ensuring discovery errors are propagated.
+                            // .push_spawn_buffer(10, Duration::from_secs(1))
+                            .push_lock()
```
olix0r: before this merges, i'd like to figure out why tests fail when this changes

linkerd/app/inbound/src/lib.rs
```diff
@@ -115,12 +116,20 @@ impl<A: OrigDstAddr> Config<A> {
...
+                    svc::layers().push_on_response(
+                        svc::layers()
+                            // If the service has been ready & unused for `cache_max_idle_age`,
+                            // fail it.
+                            .push_idle_timeout(cache_max_idle_age)
+                            // If the service has been unavailable for an extend time, eagerly
+                            // fail requests.
+                            .push_failfast(Duration::from_secs(10))
+                            // Shares the service, ensuring discovery errors are propagated.
+                            .push_spawn_buffer(10)
```
olix0r: These values should at least be constified, and probably exposed via configuration

hawkw:
Testing this branch vs b61b044ead8025c9553d3a1fcd7dcf3db56b9a74, it seems like the p99.9 latency regression I observed on #453 now only exists in the outbound direction:
ver-cache-latency

linkerd/app/inbound/src/lib.rs
```diff
@@ -159,13 +168,20 @@ impl<A: OrigDstAddr> Config<A> {
...
```
hawkw: seems like this shows up enough that we might want to have a `push_idle_timeout_failfast_and_buffer` or something (but please don't call it that!)

linkerd/app/outbound/src/lib.rs
```diff
@@ -153,10 +153,9 @@ impl<A: OrigDstAddr> Config<A> {
 
                 tcp_connect
                     .clone()
-                    // Initiates an HTTP client on the underlying transport.
-                    // Prior-knowledge HTTP/2 is typically used (i.e. when
-                    // communicating with other proxies); though HTTP/1.x fallback
-                    // is supported as needed.
+                    // Initiates an HTTP client on the underlying transport. Prior-knowledge HTTP/2
+                    // is typically used (i.e. when communicating with other proxies); though
+                    // HTTP/1.x fallback is supported as needed.
```
hawkw: it looks like all the comments in this file were re-wrapped...is this rustfmt's fault?

linkerd/app/inbound/src/lib.rs
hawkw: to wit, if  the _only_ time we push these layers is all together, it doesn't really seem like they all need their own `push_` methods on `Stack`, but it _does_ seem like we might want a `Stack::push_all_these_layers_together` method...

linkerd/cache/src/lib.rs
```diff
@@ -1,113 +1,116 @@
...
-    guard: Option<Guard<Cache<T, M::Service>>>,
-    _hangup: purge::Handle,
+    new_service: N,
+    lock: Lock<Services<T, N::Service>>,
+    guard: Option<Guard<Services<T, N::Service>>>,
 }
 
-// === impl Service ===
+#[derive(Clone, Debug)]
+pub struct Handle(Arc<()>);
```
hawkw: might be good to have a brief comment explaining how this is used?

linkerd/cache/src/lib.rs
```diff
@@ -1,113 +1,116 @@
 #![deny(warnings, rust_2018_idioms)]
 
-use self::cache::Cache;
-pub use self::layer::Layer;
-pub use self::purge::Purge;
 use futures::{future, Async, Poll};
-use linkerd2_error::Error;
+use indexmap::IndexMap;
```
hawkw: do we actually use `IndexMap`'s ordered property here? AFAICT, we don't, so it might be better to just use `HashMap`...

linkerd/app/outbound/src/lib.rs
olix0r: I updated an editor setting that was inconsistent. I'd prefer 100char wrapping to 80.

linkerd/app/inbound/src/lib.rs
olix0r: I'd rather not for now, as this makes it all more brittle to change. If it stays this way for a while, we can formalize it.

hawkw: fair enough. there is definitely a  tradeoff between, on one hand, wanting to be able to change every instance of this in one place, and on the other hand, not wanting that. :)

linkerd/app/inbound/src/lib.rs
hawkw: @olix0r what was the answer to this?

PR #457, opened 2020-03-09
Use authority_override data to set Host header

zaharidichev:
This PR uses auth override data coming from the destinations service to set the Host header for oubound http requests


zaharidichev:
Closed in favor of #458

PR #458, opened 2020-03-10
Use authority override from metadata

zaharidichev:
Overwrite the authority of the outbound req based on the data received from dst service. Whenever we averwrite the URI auth, we also overwrite the canonical_dst header and set the request to have an absolute form instead of origin one. 


olix0r:
I have some style comments... but before that: can you elaborate on why this is related to absolute-form? It’s not immediately obvious to me how the two concepts are related.

Cc @seanmonstar

zaharidichev:
@olix0r Maybe that should be made more explicit, but my reasoning is that if we overwrite the authority, we are sending that to a proxy (nginx). If that is the case the [RFC](https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html) states that: 

>  The absoluteURI form is REQUIRED when the request is being made to a proxy. The proxy is requested to forward the request or service it from a valid cache, and return the response.

olix0r:
How does that interact with HTTP2 upgrading though?

My gut instinct is that we should decouple “overwrite the authority” and “force absolute form “...

If we just overwrite the header without forcing absolute form, do things still work?

Also keep in mind that the gateway is a linkerd implementation detail and not a generalized proxy

zaharidichev:
> How does that interact with HTTP2 upgrading though?

As far as I understand, this only pertains to HTTP1. So if H1 is upgraded to H2, not much changes, then the request is downgraded on the other side it comes in abs form. Is that what you are asking about? 

> If we just overwrite the header without forcing absolute form, do things still work?

Yes.

> Also keep in mind that the gateway is a linkerd implementation detail and not a generalized proxy

Yes that is correct. The only reason I did is because based on my conversations with @seanmonstar I reached the conclusion that setting to abs form is the semantically correct thing to do. That being said, this does not mean my conclusion was right. And I also think we can live without it.

olix0r:
@zaharidichev my concern about absolute-form is mostly about how it may interact with client caching (since absolute-form is a client-wide setting). Let me dig in a little more today so I can make a reasonable recommendation.

olix0r:
@zaharidichev In order to get things slightly easier for me to move this forward, is it possible to split this PR such that we have 1 change that only wires the `AuthorityOverride` metadata through discovery? This change will be easy to ship, and it will also help me to review a narrower change about authority overwriting.

I think I remain -1 on coupling this feature with absolute-form, but it will be easier to propose an alternative once we've got the baseline changes onto master.

zaharidichev:
@olix0r Thanks for the feedback. I have split it into two PRs. The first one that just wires the metadata coming from the dst API is #462. Then we have this which is  a PR agains it that uses the change. I have also addressed the feedback from @hawkw

linkerd/proxy/http/src/overwrite_authority.rs
```diff
@@ -0,0 +1,157 @@
...
+use http::uri::Authority;
+use tracing::trace;
+
+pub trait ExtractAuthority<T> {
+    fn extract(&self, target: &T) -> Option<Authority>;
+}
+
+pub trait ForceAbsForm {
+    fn is_abs_form(&self) -> bool;
+}
```
olix0r, line 15: If we called this `ShouldOverwriteHeader`, would that change anything semantically? I'm still not 100% comfortable with this being so closely relate to absolute-form.

linkerd/proxy/http/src/overwrite_authority.rs
```diff
@@ -0,0 +1,157 @@
...
+#[derive(Debug, Clone)]
+pub struct Layer<E, H> {
+    extractor: E,
+    canonical_dst_header: H,
+}
+
+#[derive(Clone, Debug)]
+pub struct MakeSvc<E, H, M> {
+    extractor: E,
+    canonical_dst_header: H,
```
olix0r, line 26: It's surprising to me that this is coupled to the canonical dst header. Why?

zaharidichev: So.. yes in theory we can do that. It just made sense to me that we have this separate layer that overwrites `authoirty` and the canonical_dst header as a separate step, even though the `canonical_dst` was already set. Also, seems to me the `canonical_dst` is set from the `Target.Addr` at the moment via `header_from_target`, which is not modified by the authority_overwrite layer.  We can split all of that up.  I am not sure coupling the authority overwrite to the canonical_dst  troubles me so much as we actually want to overwrite both, right?

olix0r: My view is that this module shouldn't know anything about canonical dst headers. What other headers does it need to know about? This shouldn't have to grow to know about arbitrary Linkerd headers. Should this module take a list of arbitrary headers to overwrite? How do we decouple this from this particular linkerd feature.

olix0r: Is it even appropriate to set the canonical dst header, since we're setting a canonical authority anyway? Is it more appropriate to clear it entirely? How do we expect service profiles to interact with this in general?

zaharidichev: @olix0r Good questions! So the reason I am overwriting the canonical dst and host at that point is because of the situation that arises when doing traffic splitting. What happens is the following: 

Lets say we have a ts that looks like: 

apex-svc: backend-zero-svc (on the local cluster)
leafsvc: backend-two-svc-remote (on the remote cluster). 

If we do not set the canonical dst the request will leave the proxy with the floowing values for these headers: 

```
l5d-dst-canonical: backend-zero-svc.default.svc.cluster.local:8888
host: backend-zero-svc:8888
```

So when this request arrives at the remote gateway, it cannot be routed. This is why I decided to overwrite these at that point. Maybe I am missing the context as to whether this is appropriate or not, but it seemed correct to me. Wrt to service profiles, I am not sure thie interferes with them too much. The profiles on the outbound side a resolved before this rewriting is applied. The profiles on the incoming side (the remote cluster) are resolved based on the canonical_dst that we set, which will be equal to the name of the service that actually exists on the remote cluster. So this seems correct to me. I guess the same thing will work if we blank the value of this header, because then the incoming Recognize will rely on the authority of the request.  So any suggestion as to how we want to handle that without coupling are more than welcome :)

olix0r: Okay, I follow what you're saying. Here's what's tripping me up: within a cluster, service profiles are only ever applied on a logical name (i.e. in the face of a traffic split, the inbound proxy uses the canonical-form logical dst to resolve the service profile). Typically, both the outbound and inbound proxies refer to the same service profile data (though there's some added complexity about why that's not always the case). So, I guess my question is: as a user, what sort of service name would i reference when configuring a service profile for inbound traffic on the remote cluster? Do we have a concrete example of the routing logic in play for a traffic-split that goes through a gateway? That would help me be clearer on this.

My general concern about the coupling with canonical-dst is that if there's one such case, there are likely to be others. Does it make sense to take a list of headers to clear in this case? Or should we support adding a `Proxy` (as we do for service profile routes) to support adding arbitrary logic to gateway traffic?

linkerd/proxy/http/src/client.rs
```diff
@@ -128,13 +130,16 @@ where
...
                     .executor(exec)
                     .keep_alive(keep_alive)
-                    // hyper should never try to automatically set the Host
-                    // header, instead always just passing whatever we received.
-                    .set_host(false)
-                    .build(HyperConnect::new(connect, target, was_absolute_form));
+                    // hyper should should only try to automatically
+                    // set the host if the request was in absolute_form
+                    .set_host(abs_form)
+                    .build(HyperConnect::new(connect, target, abs_form));
```
olix0r, line 142: OK, Help me understand thsi change in behavior. We always created this with `false` before, but now it's not only conditioned on overwriting, but also on the `was_absolute_form` value? I'm still getting caught up on this.

olix0r: 1. Why is it always appropriate to do this when we are overwriting?
2. It feels leaky that the http client has to know about authority overwriting at all. I'd expect the target's `HasSettings` value to be enough to configure whatever is needed by the client.

zaharidichev: yeah I was a bit confused by that myself, but let me explain my logic. First of all... the gateway itself at the moment routes based on the host header of a request. So really what we want to happen is for that host header value to be equal to the fqdn of the servie on the other side. So what happens in our case when we have proxy to proxy (the gateway is meshed as well): 

On the outbound side:
- We get the overwrite metadata
- In the overwrite_authority we blank the canonical_dst and host headers AND we set the authority of the request.
- We upgrade the request of the to H2 from H1 (because we can :). 
- When we build the client, we enable automatic host header derivation (based on authority) IF we have overwritten the authority. Now keep in mind that this is a setting that only applies to H1 clients. Because we have upgraded to H2 this is effectivelly a no-op in this scenario.
- The request goes on to the remote cluster gateway with the correct authority

On the inbound side: 
- We downgrade the requests to H1.
- Because this H1 request arrived in abs form on the inbound side, we do .set_host(true) for hyper
- Now hyper will derive the Host header based on the authority and that request will arrive at the local application (the gateway) with the Host header set to the FQDN name of the service that need sto receive it at the end. 

I made the changes that you suggested so we pipe less dependencies into the client

linkerd/app/outbound/src/lib.rs
```diff
@@ -156,6 +157,7 @@ impl Config {
                     }))
                     .push(observability.clone())
                     .push(identity_headers.clone())
+                    .push(http::overwrite_authority::layer(endpoint::FromMetadata,  vec![HOST.as_str(), CANONICAL_DST_HEADER]))
```
olix0r, line 160: Why does this need `FromMetadata`? hasn't the endpoint already been constructed at this point?

zaharidichev: Yeah, I added an `ExtractAuthority` impl to `FromMetadata`. Does that introduce tangling of concepts? 

```
impl ExtractAuthority<Target<HttpEndpoint>> for FromMetadata {
    fn extract(&self, target: &Target<HttpEndpoint>) -> Option<Authority> {
        target.inner.metadata.authority_override().cloned()
    }
}
```

olix0r: I sketched out an alternate approach here https://github.com/linkerd/linkerd2-proxy/pull/481/files (note that this is based on #480)

linkerd/app/outbound/src/orig_proto_upgrade.rs
```diff
@@ -75,7 +77,9 @@ where
     fn call(&mut self, mut endpoint: Target<HttpEndpoint>) -> Self::Future {
         let can_upgrade = endpoint.inner.can_use_orig_proto();
 
-        let was_absolute = endpoint.inner.settings.was_absolute_form();
+        let was_absolute = endpoint.inner.metadata.authority_override().is_some()
+            || endpoint.inner.settings.was_absolute_form();
+
```
olix0r, line 82: https://github.com/linkerd/linkerd2-proxy/pull/485

PR #459, opened 2020-03-18
Use strong_count instead of upgrade on weak Arcs in cache

zaharidichev:
I was trying to understand the caching code a bit more today. Unless I am missing something, it seems to me that we can use `strong_count` instead of trying to upgrade a `Weak` to determine eviction eligibility. At least in the case where we are not evicting, we swap a CAS loop and a few branches for a `SeqCst` load, which seems to be making a difference. Heare are the bench results. Let me know if I am missing something obvious
```

+++master+++
test always_evict    ... bench:      22,313 ns/iter (+/- 1,882)
test never_evict     ... bench:     116,827 ns/iter (+/- 4,593)
test sometimes_evict ... bench:      71,082 ns/iter (+/- 2,954)


+++this branch+++
test always_evict    ... bench:      24,546 ns/iter (+/- 7,296)
test never_evict     ... bench:      36,306 ns/iter (+/- 3,261)
test sometimes_evict ... bench:      29,291 ns/iter (+/- 4,164)
```


benches/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+version = "0.0.0"
+publish = false
+edition = "2018"
+
+[dependencies]
+linkerd2-cache = { path = "../linkerd/cache" }
+linkerd2-error = { path = "../linkerd/error" }
+linkerd2-lock = { path = "../linkerd/lock" }
+linkerd2-stack = { path = "../linkerd/stack" }
+bencher = "0.1.5"
```
hawkw: Elsewhere, we have benchmarks that are written with the libtest benchmark framework. I'm not opposed to switching to `bencher`, especially since it can run on stable; however, I think it would be good to pick one and use it consistently. If we decide to adopt `bencher` here, we should probably open a follow-up to change other benchmarks to use `bencher` as well.

We may also want to consider `criterion` rather than `bencher`. Although I have my doubts about the accuracy of `criterion`'s regression detection (it seems way too sensitive to noise IMO), I think some of the additional benchmarking APIs `criterion` provides, like [`bench_with_input`](https://docs.rs/criterion/0.3.1/criterion/struct.BenchmarkGroup.html#method.bench_with_input) and [`iter_custom`](https://docs.rs/criterion/0.3.1/criterion/struct.Bencher.html#method.iter_custom) are very useful when writing benchmarks...just a thought!

benches/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "benches"
```
hawkw: Curious about the motivation to create a new `benches` crate, rather than putting the cache benchmarks in `linkerd/cache/benches`? Not a blocker, I just want to know if there's a reason we can't do the former.

linkerd/cache/src/lib.rs
```diff
@@ -79,7 +79,7 @@ where
                     // Drop defunct services before interacting with the cache.
                     let n = services.len();
                     services.retain(|_, (_, weak)| {
-                        if weak.upgrade().is_some() {
+                        if weak.strong_count() > 0 {
```
hawkw: :+1:

I think some of the performance advantage of this change is also that we don't increment the strong count and then immediately decrement it again (when dropping the returned `Arc` from `upgrade`).

hawkw:
CI failure doesn't _seem_ related, I'm giving it another try.

olix0r:
@hawkw the test certs are expired and need to be updated on master

benches/Cargo.toml
olix0r: All other things equal, I think I'd prefer that the tests are "close" to the code they test.. I.e. rather than a crate for all benches, I think they should be within the crate they test.  My workflow is to run `cargo test -p $crate`.

zaharidichev: No particular reason really. I was looking at other crates on github and how they organize things and its was mostly 50/50. Makes sense to be closer to code. I moved it and switched to libtest. Feel free to take another look. I am not sure whether the `unstable` feature flag is how it should be setup.

linkerd/cache/src/lib.rs
```diff
@@ -9,6 +10,8 @@ use std::hash::Hash;
 use std::sync::{Arc, Weak};
 use tracing::{debug, trace};
 
+#[cfg(all(feature = "unstable", test))]
+mod benches;
```
hawkw: Instead of a module, let's put the benchmarks in a separate `benches` directory inside the `cache` crate? That's generally considered more conventional, and would let ets us run them with `cargo bench`, as well as removing the necessity of an `unstable` feature flag.

benches/Cargo.toml
hawkw: >  I was looking at other crates on github and how they organize things and its was mostly 50/50

In general, I think putting benchmarks in a separate crate is something that libraries will do, usually so that projects that depend on that library don't need to download test-only dependencies. In this case, I don't think it's necessary, since there aren't any downstream consumers of linkerd-internal crates.

PR #461, opened 2020-03-24
Add protocol detection timeout

zaharidichev:
This change changes the protocol detection functionality to be a layer. This allows us to push a timeout on it. The reason for the timeout is that in situations when we have server speaks first protocols, we can hang forever in a "peeking" state. In order to reproduce: 

1. Install Linkerd
2. Create the following config map

```yaml
apiVersion: v1
items:
- apiVersion: v1
  data:
    config-file.cnf: |
      [mysqld]
      port=15800
  kind: ConfigMap
  metadata:
    name: config
    namespace: memtest
kind: List
```
\
3. Add the following resources: https://gist.github.com/zaharidichev/fed5a5d9d7338fc960de478dd90e93ff

4. Inject the mysql and mysql-client deployments.
5. Ssh into the mysql-client container and run a script that periodically connects to the DB: 

```bash
#!/bin/bash
for value in {1..1000000}
do
   echo "Connection" $value
   mysql --port=15800 -h mysql -ppassword -e "SELECT version()" &
   sleep 1
done
```

You can see the proxy memory growing steadily. Here are the results before and after this change: 

Screenshot 2020-03-24 at 11 43 25
Screenshot 2020-03-24 at 11 31 43

Fixes https://github.com/linkerd/linkerd2/issues/4069

### DIFF
--- a/linkerd/app/inbound/src/lib.rs
+++ b/linkerd/app/inbound/src/lib.rs
@@ -17,9 +17,10 @@ use linkerd2_app_core::{
     profiles,
     proxy::{
         self,
+        detect::DetectProtocolLayer,
         http::{self, normalize_uri, orig_proto, strip_header},
         identity,
-        server::{Protocol as ServerProtocol, Server},
+        server::{Protocol as ServerProtocol, ProtocolDetect, Server},
         tap, tcp,
     },
     reconnect, router, serve,
@@ -82,6 +83,7 @@ impl<A: OrigDstAddr> Config<A> {
                     disable_protocol_detection_for_ports,
                     dispatch_timeout,
                     max_in_flight_requests,
+                    detect_protocol_timeout,
                 },
         } = self;
 
@@ -290,11 +292,16 @@ impl<A: OrigDstAddr> Config<A> {
                 http_server.into_inner(),
                 h2_settings,
                 drain.clone(),
-                disable_protocol_detection_for_ports.clone(),
             );
 
+            let svc = svc::stack(tcp_server)
+                .push(DetectProtocolLayer::new(ProtocolDetect::new(
+                    disable_protocol_detection_for_ports.clone(),
+                )))
+                .push_timeout(detect_protocol_timeout);
+
             // Terminate inbound mTLS from other outbound proxies.
-            let accept = tls::AcceptTls::new(local_identity, tcp_server)
+            let accept = tls::AcceptTls::new(local_identity, svc.into_inner())
                 .with_skip_ports(disable_protocol_detection_for_ports);
 
             info!(listen.addr = %listen.listen_addr(), "Serving");
--- a/linkerd/app/outbound/src/lib.rs
+++ b/linkerd/app/outbound/src/lib.rs
@@ -17,13 +17,13 @@ use linkerd2_app_core::{
     opencensus::proto::trace::v1 as oc,
     profiles,
     proxy::{
-        self, core::resolve::Resolve, discover, http, identity, resolve::map_endpoint, tap, tcp,
-        Server,
+        self, core::resolve::Resolve, detect::DetectProtocolLayer, discover, http, identity,
+        resolve::map_endpoint, server::ProtocolDetect, tap, tcp, Server,
     },
     reconnect, retry, router, serve,
     spans::SpanConverter,
     svc::{self, NewService},
-    transport::{self, tls, OrigDstAddr, SysOrigDstAddr},
+    transport::{self, io::BoxedIo, tls, OrigDstAddr, SysOrigDstAddr},
     Conditional, DiscoveryRejected, Error, ProxyMetrics, TraceContextLayer, CANONICAL_DST_HEADER,
     DST_OVERRIDE_HEADER, L5D_CLIENT_ID, L5D_REMOTE_IP, L5D_REQUIRE_ID, L5D_SERVER_ID,
 };
@@ -101,6 +101,7 @@ impl<A: OrigDstAddr> Config<A> {
                     disable_protocol_detection_for_ports,
                     dispatch_timeout,
                     max_in_flight_requests,
+                    detect_protocol_timeout,
                 },
         } = self;
 
@@ -427,13 +428,19 @@ impl<A: OrigDstAddr> Config<A> {
                 http_server.into_inner(),
                 h2_settings,
                 drain.clone(),
-                disable_protocol_detection_for_ports.clone(),
             );
+
+            let svc = svc::stack(tcp_server)
+                .push(DetectProtocolLayer::new(ProtocolDetect::new(
+                    disable_protocol_detection_for_ports.clone(),
+                )))
+                .push_timeout(detect_protocol_timeout)
+                .check_service_response::<(tls::accept::Meta, BoxedIo), ()>();
 
             // The local application does not establish mTLS with the proxy.
             let no_tls: tls::Conditional<identity::Local> =
                 Conditional::None(tls::ReasonForNoPeerName::Loopback.into());
-            let accept = tls::AcceptTls::new(no_tls, tcp_server)
+            let accept = tls::AcceptTls::new(no_tls, svc.into_inner())
                 .with_skip_ports(disable_protocol_detection_for_ports);
 
             serve::serve(listen, accept, drain)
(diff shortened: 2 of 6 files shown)


linkerd/proxy/detect/src/lib.rs
```diff
@@ -17,41 +17,26 @@ pub trait Detect<T>: Clone {
...
+
+// === impl DetectProtoLayer ===
+impl<D> DetectProtocolLayer<D> {
     const DEFAULT_CAPACITY: usize = 8192;
 
-    /// Creates a new `Detect`.
-    pub fn new(detect: D, accept: A) -> Self {
-        Self {
+    pub fn new(detect: D) -> DetectProtocolLayer<D> {
+        DetectProtocolLayer {
```
olix0r: Generally, we prefer the `Self` form... minimizes churn if names change.
```suggestion
    pub fn new(detect: D) -> Self {
        Self {
```

linkerd/proxy/detect/src/lib.rs
```diff
@@ -62,25 +47,63 @@ impl<D, A> Accept<D, A> {
...
+        inner: PeekAndDetect<T, D>,
+    },
+}
+
+impl<T, D, S> tower::Service<(T, BoxedIo)> for DetectProtocol<D, S>
 where
     D: Detect<T>,
-    A: core::listen::Accept<(D::Target, BoxedIo)> + Clone,
+    S: tower::Service<(D::Target, BoxedIo)>,
+    S: Clone,
```
olix0r: nit:
```suggestion
    S: tower::Service<(D::Target, BoxedIo)> + Clone,
```

Why not continue to use the `Accept` alias though? is this not still an `Accept`?

linkerd/app/outbound/src/lib.rs
```diff
@@ -427,13 +428,19 @@ impl<A: OrigDstAddr> Config<A> {
                 http_server.into_inner(),
                 h2_settings,
                 drain.clone(),
-                disable_protocol_detection_for_ports.clone(),
             );
 
+            let svc = svc::stack(tcp_server)
```
olix0r: can we give this a better name than `svc`? There are a lot of services in  this scope... How about something like `tcp_detect`?

linkerd/app/src/env.rs
```diff
@@ -171,6 +174,7 @@ const DEFAULT_INBOUND_CONNECT_BACKOFF: ExponentialBackoff = ExponentialBackoff {
 };
 const DEFAULT_OUTBOUND_DISPATCH_TIMEOUT: Duration = Duration::from_secs(3);
 const DEFAULT_OUTBOUND_CONNECT_TIMEOUT: Duration = Duration::from_secs(1);
+const DEFAULT_PROTOCOL_DETECT_TIMEOUT: Duration = Duration::from_secs(10);
```
olix0r: What do you think about using the DISPATCH_TIMEOUT settings to control the detect timeout. Specifically, I'm fine with them being distinct variables in the config objects, but from the operator's point of view, are these things materially different? They are both effectively "the amount of time I'm willing to spend in the proxy before giving up".

linkerd/proxy/detect/src/lib.rs
```diff
@@ -90,36 +113,45 @@ where
...
+                            let target = detect.detect_peeked_prefix(
+                                target.take().expect("polled after complete"),
+                                io.prefix().as_ref(),
+                            );
+                            let inner_fut = inner_svc.call((target, BoxedIo::new(io)));
+                            *self = DetectProtocolFuture::Detected(inner_fut);
+                        }
+
+                        Err(err) => return Err(Error::from(err)),
+                    },
```
olix0r: It's not clear to me what's changing here exactly... Would you mind reordering the branches as they were previously (peek, and then detected). It looks like there's some unnecessary changes here... For instance, can't we continue to use `try_ready!(` in the peek branch? Can this diff be made smaller? Is it just renames/reorg or are there functional changes?

linkerd/app/outbound/src/lib.rs
```diff
@@ -427,13 +428,19 @@ impl<A: OrigDstAddr> Config<A> {
...
+                    disable_protocol_detection_for_ports.clone(),
+                )))
+                .push_timeout(detect_protocol_timeout)
+                .check_service_response::<(tls::accept::Meta, BoxedIo), ()>();
+
             // The local application does not establish mTLS with the proxy.
             let no_tls: tls::Conditional<identity::Local> =
                 Conditional::None(tls::ReasonForNoPeerName::Loopback.into());
-            let accept = tls::AcceptTls::new(no_tls, tcp_server)
+            let accept = tls::AcceptTls::new(no_tls, svc.into_inner())
```
olix0r: Don't we need to timeout outside of the TLS detection? Or, why don't we encounter this problem with TLS detection?

olix0r: Errr, this won't matter on the outbound side. But it will on the inbound side.

linkerd/app/inbound/src/lib.rs
```diff
@@ -290,11 +292,16 @@ impl<A: OrigDstAddr> Config<A> {
...
 
+            let svc = svc::stack(tcp_server)
+                .push(DetectProtocolLayer::new(ProtocolDetect::new(
+                    disable_protocol_detection_for_ports.clone(),
+                )))
+                .push_timeout(detect_protocol_timeout);
+
             // Terminate inbound mTLS from other outbound proxies.
-            let accept = tls::AcceptTls::new(local_identity, tcp_server)
+            let accept = tls::AcceptTls::new(local_identity, svc.into_inner())
```
olix0r: Don't we need to timeout outside of the TLS detection? Or, why don't we encounter this problem with TLS detection?

linkerd/app/src/env.rs
hawkw: :+1:, I'm not convinced these need to be separately tuneable knobs?

linkerd/proxy/detect/src/lib.rs
```diff
@@ -62,25 +47,63 @@ impl<D, A> Accept<D, A> {
...
+        self.inner.poll_ready().map_err(Into::into)
     }
 
     fn call(&mut self, (target, io): (T, BoxedIo)) -> Self::Future {
         match self.detect.detect_before_peek(target) {
-            Ok(detected) => AcceptFuture::Accept(self.accept.accept((detected, io))),
-            Err(target) => AcceptFuture::Detect {
+            Ok(detected) => {
+                let inner_fut = self.inner.call((detected, io));
+                return DetectProtocolFuture::Detected(inner_fut);
```
hawkw: nit: this isn't in a loop or assignment, so I don't think the `return` is necessary here
```suggestion
                DetectProtocolFuture::Detected(inner_fut)
```

linkerd/proxy/detect/src/lib.rs
```diff
@@ -62,25 +47,63 @@ impl<D, A> Accept<D, A> {
...
+    D: Detect<T>,
+{
+    Detected(S::Future),
+    Detecting {
+        inner_svc: S,
+        detect: D,
+        inner: PeekAndDetect<T, D>,
+    },
+}
+
```
hawkw: very minor nit: elsewhere in Linkerd, we put all public struct definitions at the top of the file, before the implementations, can we do that here as well?

linkerd/app/inbound/src/lib.rs
```diff
@@ -290,11 +292,16 @@ impl<A: OrigDstAddr> Config<A> {
...
                 h2_settings,
                 drain.clone(),
-                disable_protocol_detection_for_ports.clone(),
             );
 
+            let svc = svc::stack(tcp_server)
+                .push(DetectProtocolLayer::new(ProtocolDetect::new(
+                    disable_protocol_detection_for_ports.clone(),
+                )))
+                .push_timeout(detect_protocol_timeout);
```
olix0r: I just realized this is unlikely to work as expected... The accept stack handles processing of the request, so this timeout seems like it will limit the _lifetime_ of all TCP connections...

I think we have two options:

1. Handle the timeout in the detect layer. The advantage to this approach would be that we can return the underlying socket in the error so that it could be rerouted if we want to do that later.
2. If we want to keep the timeout separate; we'd have to model Detect to have a response type that is the detected protocol and the socket; and then we'd have another layer that basically does `detect_or_timeout.call(sock).and_then(self.inner.accept)`.

I think, all things considered, it might be simplest to add the timeout to the detection layer directly.

zaharidichev:
closed in favour of #464

PR #462, opened 2020-03-24
Wire authority override coming from discovery

zaharidichev:
This PR uses the latest changes to the proxy-api and consumes the authority override data, adding to the `Metadata` struct

### DIFF
--- a/linkerd/app/Cargo.toml
+++ b/linkerd/app/Cargo.toml
@@ -31,7 +31,7 @@ http = "0.1"
 http-body = "0.1"
 hyper = "0.12"
 linkerd2-metrics = { path = "../metrics", features = ["test_util"] }
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], branch = "zd/auth_override" }
+linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], tag = "v0.1.12" }
 net2 = "0.2"
 quickcheck = { version = "0.9", default-features = false }
 ring = "0.16"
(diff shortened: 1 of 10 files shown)


linkerd/app/Cargo.toml
```diff
@@ -31,7 +31,7 @@ http = "0.1"
 http-body = "0.1"
 hyper = "0.12"
 linkerd2-metrics = { path = "../metrics", features = ["test_util"] }
-linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", features = ["arbitrary"], tag = "v0.1.11" }
```
kleimkuhler: Do we want to block this getting a linkerd2-proxy-api release out first?

zaharidichev: Yes we do

linkerd/proxy/api-resolve/src/pb.rs
```diff
@@ -56,6 +60,19 @@ fn to_id(pb: TlsIdentity) -> Option<identity::Name> {
...
 }
 
+pub(in crate) fn to_authority(o: AuthorityOverride) -> Option<Authority> {
+    match o.authority_override.parse() {
+        Ok(name) => Some(name),
+        Err(_) => {
+            tracing::warn!(
+                "Ignoring invalid authority override: {}",
+                o.authority_override
+            );
```
olix0r: Do we `warn` elsewhere in this file? Generally, if there's nothing a user can do about it, I prefer to keep this at debug. IMO, if the destination service is providing invalid values, that's an internal error and not a user error.

linkerd/app/Cargo.toml
zaharidichev: @kleimkuhler We are using the released api now.

PR #463, opened 2020-03-25
Add Labels from a path as oc-collector attributes

Pothulapati:
Fixes https://github.com/linkerd/linkerd2/issues/4008

This PR makes the proxy read file present at `/var/run/linkerd/podinfo/labels` and convert it into a `HashMap` and use those labels as attributes for spans/traces. 

https://github.com/linkerd/linkerd2/pull/4199 updates the control-plane helm charts to mount pod labels as file to the proxy coontainer [using downwardAPI](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/)

With these changes, Traces would have pod labels as attributes, Allowing us to directly link jager dashboard links from the Linkerd dashboard (like we do with Grafana), part of https://github.com/linkerd/linkerd2/pull/4177 . This was not possible previously because we didn't have a way to know the distinguish traces based on `workloadKind` and `workloadName`

Screenshot_2020-03-25 Jaeger UI
image

P.S: This is my first take in Rust and the proxy, Sorry if there are any mistakes :) 
Signed-off-by: Tarun Pothulapati <tarunpothulapati@outlook.com>

### DIFF
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -72,7 +72,7 @@ pub const ENV_OUTBOUND_ROUTER_MAX_IDLE_AGE: &str = "LINKERD2_PROXY_OUTBOUND_ROUT
 pub const ENV_INBOUND_MAX_IN_FLIGHT: &str = "LINKERD2_PROXY_INBOUND_MAX_IN_FLIGHT";
 pub const ENV_OUTBOUND_MAX_IN_FLIGHT: &str = "LINKERD2_PROXY_OUTBOUND_MAX_IN_FLIGHT";
 
-pub const ENV_LABELS_FILE_PATH: &str = "LINKERD2_PROXY_LABELS_FILE_PATH";
+pub const ENV_TRACE_ATTRIBUTES_PATH: &str = "LINKERD2_PROXY_TRACE_ATTRIBUTES_PATH";
 
 /// Constrains which destination names are resolved through the destination
 /// service.
@@ -283,7 +283,7 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
 
     let hostname = strings.get(ENV_HOSTNAME);
 
-    let oc_labels_file_path = strings.get(ENV_LABELS_FILE_PATH);
+    let oc_attributes_file_path = strings.get(ENV_TRACE_ATTRIBUTES_PATH);
 
     let trace_collector_addr = if id_disabled {
         parse_control_addr_disable_identity(strings, ENV_TRACE_COLLECTOR_SVC_BASE)
@@ -454,15 +454,15 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
                 outbound.proxy.connect.clone()
             };
 
-            let labels = oc_labels_file_path
+            let attributes = oc_attributes_file_path
                 .map(|path| match path {
-                    Some(path) => oc_trace_labels(path),
+                    Some(path) => oc_trace_attributes(path),
                     None => HashMap::new(),
                 })
                 .unwrap_or_default();
 
             oc_collector::Config::Enabled {
-                labels,
+                attributes,
                 hostname: hostname?,
                 control: ControlConfig {
                     addr,
@@ -514,21 +514,21 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
     })
 }
 
-fn oc_trace_labels(oc_file_path: String) -> HashMap<String, String> {
-    match fs::read_to_string(oc_file_path.clone()) {
-        Ok(label_string) => convert_labels_to_map(label_string),
+fn oc_trace_attributes(oc_attributes_file_path: String) -> HashMap<String, String> {
+    match fs::read_to_string(oc_attributes_file_path.clone()) {
+        Ok(attributes_string) => convert_attributes_string_to_map(attributes_string),
         Err(err) => {
             warn!(
-                "could not read OC trace labels file at path {}: {}",
-                oc_file_path, err
+                "could not read OC trace attributes file at {}: {}",
+                oc_attributes_file_path, err
             );
             HashMap::new()
         }
     }
 }
 
-fn convert_labels_to_map(labels: String) -> HashMap<String, String> {
-    labels
+fn convert_attributes_string_to_map(attributes: String) -> HashMap<String, String> {
+    attributes
         .lines()
         .filter_map(|line| {
             let mut parts = line.splitn(2, "=");
@@ -1040,8 +1040,8 @@ mod tests {
     }
 
     #[test]
-    fn convert_labels_to_map_different_values() {
-        let labels_string = "\
+    fn convert_attributes_string_to_map_different_values() {
+        let attributes_string = "\
             cluster=\"test-cluster1\"\n\
             rack=\"rack-22\"\n\
             zone=us-est-coast\n\
@@ -1078,7 +1078,10 @@ mod tests {
         .cloned()
         .collect();
 
-        assert_eq!(convert_labels_to_map(labels_string), expected);
+        assert_eq!(
+            convert_attributes_string_to_map(attributes_string),
+            expected
+        );
     }
 
     #[test]
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -72,7 +72,7 @@ pub const ENV_OUTBOUND_ROUTER_MAX_IDLE_AGE: &str = "LINKERD2_PROXY_OUTBOUND_ROUT
 pub const ENV_INBOUND_MAX_IN_FLIGHT: &str = "LINKERD2_PROXY_INBOUND_MAX_IN_FLIGHT";
 pub const ENV_OUTBOUND_MAX_IN_FLIGHT: &str = "LINKERD2_PROXY_OUTBOUND_MAX_IN_FLIGHT";
 
-pub const LABELS_FILE_PATH: &str = "/var/run/linkerd/podinfo/labels";
+pub const ENV_LABELS_FILE_PATH: &str = "LINKERD2_PROXY_LABELS_FILE_PATH";
 
 /// Constrains which destination names are resolved through the destination
 /// service.
@@ -283,8 +283,12 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
 
     let hostname = strings.get(ENV_HOSTNAME);
 
-    let f = fs::read_to_string(LABELS_FILE_PATH);
-    let labels = f.ok().map(convert_labels_to_map);
+    let labels = strings.get(ENV_LABELS_FILE_PATH).map(|path_option| {
+        path_option
+            .map(fs::read_to_string)
+            .and_then(|label_string| label_string.ok().map(convert_labels_to_map))
+            .unwrap_or_default()
+    });
 
     let trace_collector_addr = if id_disabled {
         parse_control_addr_disable_identity(strings, ENV_TRACE_COLLECTOR_SVC_BASE)
@@ -455,7 +459,7 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
                 outbound.proxy.connect.clone()
             };
             oc_collector::Config::Enabled {
-                labels: labels.unwrap_or_default(),
+                labels: labels?,
                 hostname: hostname?,
                 control: ControlConfig {
                     addr,
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -457,10 +457,7 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
             let labels = oc_labels_file_path
                 .map(|path| match path {
                     Some(path) => oc_trace_labels(path),
-                    None => {
-                        warn!("env variable {} not present", ENV_LABELS_FILE_PATH);
-                        HashMap::new()
-                    }
+                    None => HashMap::new(),
                 })
                 .unwrap_or_default();
 
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -462,13 +462,7 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
                         HashMap::new()
                     }
                 })
-                .unwrap_or_else(|err| {
-                    warn!(
-                        "could not read env variable {}: {}",
-                        ENV_LABELS_FILE_PATH, err
-                    );
-                    HashMap::new()
-                });
+                .unwrap_or_default();
 
             oc_collector::Config::Enabled {
                 labels,
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -284,7 +284,7 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
     let hostname = strings.get(ENV_HOSTNAME);
 
     let f = fs::read_to_string(LABELS_FILE_PATH);
-    let labels = convert_labels_to_map(f.unwrap_or_default());
+    let labels = f.ok().map(convert_labels_to_map);
 
     let trace_collector_addr = if id_disabled {
         parse_control_addr_disable_identity(strings, ENV_TRACE_COLLECTOR_SVC_BASE)
@@ -455,7 +455,7 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
                 outbound.proxy.connect.clone()
             };
             oc_collector::Config::Enabled {
-                labels: Some(labels),
+                labels: labels.unwrap_or_default(),
                 hostname: hostname?,
                 control: ControlConfig {
                     addr,
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -14,6 +14,7 @@ use std::path::PathBuf;
 use std::str::FromStr;
 use std::time::Duration;
 use std::{fmt, fs};
+use std::collections::HashMap;
 use tracing::{error, warn};
 
 /// The strings used to build a configuration.
@@ -70,6 +71,8 @@ pub const ENV_OUTBOUND_ROUTER_MAX_IDLE_AGE: &str = "LINKERD2_PROXY_OUTBOUND_ROUT
 
 pub const ENV_INBOUND_MAX_IN_FLIGHT: &str = "LINKERD2_PROXY_INBOUND_MAX_IN_FLIGHT";
 pub const ENV_OUTBOUND_MAX_IN_FLIGHT: &str = "LINKERD2_PROXY_OUTBOUND_MAX_IN_FLIGHT";
+
+pub const LABELS_FILE_PATH: &str = "/var/run/linkerd/podinfo/labels";
 
 /// Constrains which destination names are resolved through the destination
 /// service.
@@ -280,6 +283,9 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
 
     let hostname = strings.get(ENV_HOSTNAME);
 
+    let f = fs::read_to_string(LABELS_FILE_PATH);
+    let labels = convert_string_to_map(f.unwrap_or_default());
+
     let trace_collector_addr = if id_disabled {
         parse_control_addr_disable_identity(strings, ENV_TRACE_COLLECTOR_SVC_BASE)
     } else {
@@ -449,6 +455,7 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
                 outbound.proxy.connect.clone()
             };
             oc_collector::Config::Enabled {
+                labels: Some(labels),
                 hostname: hostname?,
                 control: ControlConfig {
                     addr,
@@ -498,6 +505,15 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
         outbound,
         inbound,
     })
+}
+
+fn convert_string_to_map(labels :String) -> HashMap<String, String> {
+    let mut labels_map: HashMap<String,String> = HashMap::new();
+    for line in labels.lines() {
+        let label = line.split("=").collect::<Vec<&str>>();
+        labels_map.insert(label[0].to_string(), label[1][1..label[1].len()-1].to_string());
+    }
+    labels_map
 }
 
 fn default_disable_ports_protocol_detection() -> IndexSet<u16> {
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -283,12 +283,7 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
 
     let hostname = strings.get(ENV_HOSTNAME);
 
-    let labels = strings.get(ENV_LABELS_FILE_PATH).map(|path_option| {
-        path_option
-            .map(fs::read_to_string)
-            .and_then(|label_string| label_string.ok().map(convert_labels_to_map))
-            .unwrap_or_default()
-    });
+    let oc_labels_file_path = strings.get(ENV_LABELS_FILE_PATH);
 
     let trace_collector_addr = if id_disabled {
         parse_control_addr_disable_identity(strings, ENV_TRACE_COLLECTOR_SVC_BASE)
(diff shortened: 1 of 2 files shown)


linkerd/app/src/env.rs
```diff
@@ -500,6 +507,18 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
     })
 }
 
+fn convert_string_to_map(labels: String) -> HashMap<String, String> {
+    let mut labels_map: HashMap<String, String> = HashMap::new();
+    for line in labels.lines() {
+        let label = line.split("=").collect::<Vec<&str>>();
+        labels_map.insert(
+            label[0].to_string(),
+            label[1][1..label[1].len() - 1].to_string(),
```
olix0r: What does this do? Perhaps break this into multiple statements? Or at least comment what this indexing does... Also, is this guaranteed to be in range or can this cause runtime panics?

linkerd/app/src/env.rs
```diff
@@ -500,6 +507,18 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
+        );
+    }
```
olix0r: Perhaps safer as

```suggestion
    label.lines()
        .filter_map(|line| {
           let parts = line.split("=");
           parts.pop().and_then(move |key| {
              parts.pop().map(move |val| (key.to_string(), val.to_string()))
           })
         }).collect::<HashMap<_, _>>()
```

olix0r: It might also be worth putting a small unit test at the bottom of this file...

linkerd/app/src/env.rs
```diff
@@ -71,6 +72,8 @@ pub const ENV_OUTBOUND_ROUTER_MAX_IDLE_AGE: &str = "LINKERD2_PROXY_OUTBOUND_ROUT
 pub const ENV_INBOUND_MAX_IN_FLIGHT: &str = "LINKERD2_PROXY_INBOUND_MAX_IN_FLIGHT";
 pub const ENV_OUTBOUND_MAX_IN_FLIGHT: &str = "LINKERD2_PROXY_OUTBOUND_MAX_IN_FLIGHT";
 
+pub const LABELS_FILE_PATH: &str = "/var/run/linkerd/podinfo/labels";
```
hawkw: Should this be hard-coded or set as an env var?

linkerd/app/src/env.rs
Pothulapati: So, k8s adds double quotes to the label value `cluster="test-cluster1"`. Hence the indexing. 

To make it more generic for cases like mesh expansion, etc. I guess having a check for `"` and performing indexing only when they are present, should be good?

hawkw: What are we trying to do with the string slice `[1..label[1].len() - 1]`? It looks like we are slicing one character off both ends of the string? Is there whitespace or something that needs to be trimmed off?

This code might be easier to understand if there was a brief comment explaining what format we expect the labels to be in.

linkerd/app/src/oc_collector.rs
```diff
@@ -17,6 +17,7 @@ pub enum Config {
     Disabled,
     Enabled {
         control: ControlConfig,
+        labels: Option<HashMap<String, String>>,
```
hawkw: nit, take it or leave it: Since we use `Option::unwrap_or_default` later, if this is `None`, we will eventually create an empty `HashMap`. Seems like we could just pass an empty hash map if there were no labels. Since [`HashMap::new` doesn't allocate](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.new), there isn't a lot of additional overhead (it's a _little_ bigger than an `Option` but this isn't a hot path).

linkerd/app/src/env.rs
```diff
@@ -280,6 +283,9 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
 
     let hostname = strings.get(ENV_HOSTNAME);
 
+    let f = fs::read_to_string(LABELS_FILE_PATH);
+    let labels = convert_string_to_map(f.unwrap_or_default());
```
hawkw: nit: The `unwrap_or_default` here will pass an empty string to `convert_string_to_map` if we couldn't read labels. Why not just skip calling this function? We could do something like
```suggestion
    let labels = f.ok().map(convert_string_to_map);
```
and avoid calling the function if the labels are empty.

linkerd/app/src/env.rs
```diff
@@ -500,6 +507,18 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
     })
 }
 
+fn convert_string_to_map(labels: String) -> HashMap<String, String> {
```
hawkw: nit: maybe the name should suggest that this is specifically converting pod labels for OpenCensus? `convert_string_to_map` could be converting any string to any map...

linkerd/app/src/env.rs
hawkw: What about using [`str::trim_matches`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.trim_matches) instead? Like
```suggestion
            label[1].trim_matches('"').to_string()
```

hawkw: If we `collect` the iterator returned by `split` to a `Vec` and then index it, we might panic if an index is out of bounds (e.g. if the line didn't contain a `=` to split on, and the iterator only had one item in it). 

Instead, I think we should just skip any lines that are invalid. We could do this by just calling `next` on the iterator twice. Since [`Iterator::next`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#tymethod.next) returns an `Option`, we'll know if the expected number of elements is missing. Maybe something like
```rust
        let mut label = line.split("=");
        if let (Some(key), Some(val)) = (labels.next(), labels.next()) {
            // the line contained both a `key` and a `value`, trim them and insert into the map...
         }
         // otherwise, the line was malformed, continue iterating...
```

linkerd/app/src/env.rs
```diff
@@ -1017,6 +1020,24 @@ mod tests {
         assert_eq!(parse_duration("1"), Err(ParseError::NotADuration));
     }
 
+    #[test]
+    fn convert_labels_to_map_different_values() {
```
hawkw: It seems like it might also be good to test some malformed/invalid labels, and ensure that they are ignored rather than panicking?

Pothulapati:
@hawkw @olix0r [K8s has gurantees on what kinds of keys and values it allows as labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set). It allows the `value` to be empty in which case, the pair is written to the file as `new=""`, which we succesfully consider. and we do the same even when `key` is empty (though k8s dosen't allow them)

One concern I have is for future cases (i.e mesh expansion, etc), For Labels without value, Users have to add `=`, as we ignore labels without `=` even though value is empty, Should we allow that?

olix0r:
@Pothulapati 

> For Labels without value, Users have to add =, as we ignore labels without = even though value is empty, Should we allow that?

Can you elaborate on this? Can you give an example of how a user would interact with this? I would imagine that users never have to interact with any of this stuff outside of the normal k8s tooling?

Pothulapati:
@olix0r Yep, no direct interaction as you mentioned but my concern was  for future cases outside k8s (mesh expansion? ), where users would have to add the labels manually into a file, and even for labels without values,  `=` has to be added.

olix0r:
> where users would have to add the labels manually into a file

I don't think we'd ever do this, fwiw.

linkerd/app/src/env.rs
```diff
@@ -283,8 +283,12 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
     let hostname = strings.get(ENV_HOSTNAME);
 
-    let f = fs::read_to_string(LABELS_FILE_PATH);
-    let labels = f.ok().map(convert_labels_to_map);
+    let labels = strings.get(ENV_LABELS_FILE_PATH).map(|path_option| {
+        path_option
+            .map(fs::read_to_string)
+            .and_then(|label_string| label_string.ok().map(convert_labels_to_map))
+            .unwrap_or_default()
+    });
```
hawkw: :+1:

linkerd/app/src/env.rs
```diff
@@ -283,8 +283,12 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
 
     let hostname = strings.get(ENV_HOSTNAME);
 
-    let f = fs::read_to_string(LABELS_FILE_PATH);
-    let labels = f.ok().map(convert_labels_to_map);
+    let labels = strings.get(ENV_LABELS_FILE_PATH).map(|path_option| {
+        path_option
+            .map(fs::read_to_string)
```
hawkw: TIOLI: do you think we should log here if the path doesn't exist? Since the file not existing and the file containing invalid contents are different errors, it could be useful to record which one happened for debugging purposes.

linkerd/app/src/env.rs
```diff
@@ -280,6 +283,13 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
 
     let hostname = strings.get(ENV_HOSTNAME);
 
+    let labels = strings.get(ENV_LABELS_FILE_PATH).map(|path_option| {
```
olix0r: Would you mind either: (1) renaming `labels` to `oc_trace_labels` or (2) moving this into a function and invoking at lines 462...

"Labels" means several things in the context of the proxy, so I want to be very clear about use of this term here... In this case, these labels are only used for tracing exports (which is not used the majority of the time, so it seems generally preferable to only execute this when tracing is enabled).

linkerd/app/src/env.rs
Pothulapati: Yep, but should we err out? I added `warn` for most cases (It's a bit verbose though :p ) as labels should not be necessary.

Pothulapati:
Updated the proxy code to also include `namespace` label. Especially useful for jaeger integration into the Linkerd dashboard.
Screenshot_2020-04-06 Jaeger UI(1)
Screenshot_2020-04-06 Jaeger UI

linkerd/app/src/env.rs
```diff
@@ -998,6 +1063,37 @@ mod tests {
         assert_eq!(parse_duration("1"), Err(ParseError::NotADuration));
     }
 
+    #[test]
+    fn convert_labels_to_map_different_values() {
+        let labels_string = "cluster=\"test-cluster1\"\nrack=\"rack-22\nzone=us-est-coast\nlinkerd.io/control-plane-component=\"controller\"\nlinkerd.io/proxy-deployment=\"linkerd-controller\"\nworkload=\nkind=\"\"\nkey1=\"=\"\nkey2==value2\nkey3\n=key4\n
```
olix0r, line 1068: This would be a bit more readable as a multiline string: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=daa8a3650244073182fc93cde9c7a77d

```suggestion
        let labels_string =  "\
cluster=\"test-cluster1\"
rack=\"rack-22
zone=us-est-coast
linkerd.io/control-plane-component=\"controller\"
linkerd.io/proxy-deployment=\"linkerd-controller\"
workload=
kind=\"\"
key1=\"=\"
key2==value2
key3
=key4
";
```

linkerd/app/src/env.rs
```diff
@@ -448,7 +455,20 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
+                namespace.unwrap_or_default().unwrap_or_default(),
+            )]
+            .iter()
+            .cloned()
+            .collect();
+
+            labels.extend(oc_trace_labels(oc_labels_file_path));
+
             oc_collector::Config::Enabled {
+                labels: labels,
```
olix0r, line 471:
```suggestion
                labels,
```

linkerd/app/src/env.rs
```diff
@@ -448,7 +455,20 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
             };
+
+            // Add any extra Labels, that are not present in the labels file
+            let mut labels: HashMap<String, String> = [(
+                "namespace".to_string(),
+                namespace.unwrap_or_default().unwrap_or_default(),
+            )]
+            .iter()
+            .cloned()
+            .collect();
```
olix0r, line 466: What's this doing, just creating a Map? Isn't this more code than just instantiating it and inserting the value?

```rust
let mut labels = HashMap::new();
labels.insert("namespace".to_string(), namespace.flatten().unwrap_or_default())
```

Keep it simple.

linkerd/app/src/env.rs
```diff
@@ -125,6 +128,7 @@ pub const ENV_IDENTITY_SVC_BASE: &str = "LINKERD2_PROXY_IDENTITY_SVC";
 pub const ENV_DESTINATION_SVC_BASE: &str = "LINKERD2_PROXY_DESTINATION_SVC";
 
 pub const ENV_HOSTNAME: &str = "HOSTNAME";
+pub const ENV_NAMESPACE: &str = "_pod_ns";
```
olix0r, line 131: Strong -1 on this. This couples the proxy to a temp variable used in the injector. The proxy shouldn't know anything about this, or namespaces for that matter. I'd prefer to support an additional env variable that can extend the set of tracing labels.

linkerd/app/src/env.rs
hawkw: Alternatively, we could use the line continuation `\` syntax and also indent each line:

```suggestion
        let labels_string =  "\
            cluster=\"test-cluster1\"\n\
            rack=\"rack-22\"\n\
            zone=us-est-coast\n\
            linkerd.io/control-plane-component=\"controller\"\n\
            linkerd.io/proxy-deployment=\"linkerd-controller\"\n\
            workload=\n\
            kind=\"\"\n\
            key1=\"=\"\n\
            key2==value2\n\
            key3\n\
            =key4\n\
            ";
```
Although I don't know if that's significantly clearer.

linkerd/app/src/env.rs
```diff
@@ -448,7 +455,20 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
+            // Add any extra Labels, that are not present in the labels file
+            let mut labels: HashMap<String, String> = [(
+                "namespace".to_string(),
+                namespace.unwrap_or_default().unwrap_or_default(),
+            )]
+            .iter()
+            .cloned()
+            .collect();
+
+            labels.extend(oc_trace_labels(oc_labels_file_path));
```
hawkw, line 468: IMO, creating a hashmap and then extending it with the labels from the file seems a little weird, although it does let us override the namespace label one from the file, if there is one. What about just doing
```suggestion
            let mut labels = oc_trace_labels(oc_labels_file_path);

            // If the file did not contain a `namespace` label, add one.
            labels.entry("namespace".to_string()).or_insert_with(|| namespace.flatten().unwrap_or_default());
            labels.extend(oc_trace_labels(oc_labels_file_path));
```

linkerd/app/src/env.rs
```diff
@@ -500,6 +520,51 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
     })
 }
 
+fn oc_trace_labels(oc_file_path: Result<Option<String>, EnvError>) -> HashMap<String, String> {
```
hawkw, line 523: I find it somewhat strange that we pass the path in as a `Result`, rather than just not calling this function if we can't get it from the env. What if instead of matching that result here, we made this
```rust
fn oc_trace_labels(oc_file_path: &str) -> HashMap<String, String> {
    // ...
}
```
removed a lot of the error handling logic from this function, and changed the call site to something like
```rust
let mut labels = oc_labels_file_path
    .map(|path| match path {
        Some(path) => oc_trace_labels(path),
        None => {
            warn!("env variable {} not present", ENV_LABELS_FILE_PATH);
            HashMap::new()
        }
    }).unwrap_or_else(|err| {
        warn!(
            "could not read env variable {}: {}",
            ENV_LABELS_FILE_PATH, err
        );
        HashMap::new()
});
``` 
or something like that?

This function feels like it's doing a lot of error handling.

linkerd/app/src/env.rs
```diff
@@ -500,6 +520,51 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
+    let oc_labels = match oc_file_path {
+        Ok(path_option) => match path_option {
+            Some(path) => match fs::read_to_string(path.clone()) {
+                Ok(label_string) => convert_labels_to_map(label_string),
+                Err(err) => {
+                    warn!(
+                        "could not read OC trace labels file at path {}: {}",
+                        path, err
+                    );
+                    HashMap::<String, String>::new()
```
hawkw, line 533: Note that because the the return type of this function is `HashMap<String, String>`, the compiler will infer the type of this for you, and you could just write `HashMap::new()` in this function.

linkerd/app/src/env.rs
```diff
@@ -500,6 +520,51 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
+                }
+            },
+            None => {
+                warn!("env variable {} not present", ENV_LABELS_FILE_PATH);
+                HashMap::<String, String>::new()
+            }
```
hawkw, line 539: Do we expect this env var to always be set? If it's only set when OpenCensus is in use, I am not sure if this should be a warning.

linkerd/app/src/env.rs
```diff
@@ -998,6 +1063,37 @@ mod tests {
...
+                "linkerd-controller".to_string(),
+            ),
+            ("workload".to_string(), "".to_string()),
+            ("kind".to_string(), "".to_string()),
+            ("key1".to_string(), "=".to_string()),
+            ("key2".to_string(), "=value2".to_string()),
+            ("".to_string(), "key4".to_string()),
+        ]
+        .iter()
+        .cloned()
```
hawkw, line 1091: I think we could remove the need to clone here if we changed this to 
```suggestion
        .into_iter()
```
which moves the values out of the array by value, instead of iterating over references.

linkerd/app/src/env.rs
```diff
@@ -500,6 +520,51 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
+        .lines()
+        .filter_map(|line| {
+            let mut parts = line.splitn(2, "=").collect::<Vec<&str>>();
+            parts.pop().and_then(move |val| {
+                parts.pop().map(move |key|
+                // Trim double quotes in value, present by default when attached through k8s downwardAPI
+                (key.to_string(), val.trim_matches('"').to_string()))
+            })
+        })
+        .collect::<HashMap<String, String>>();
```
hawkw, line 564: Nit: Since this function returns a `HashMap<String, String>`, the compiler should be able to infer the type for this `collect` without an explicit type annotation:
```suggestion
        .collect();
```

linkerd/app/src/env.rs
```diff
@@ -500,6 +520,51 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
+fn convert_labels_to_map(labels: String) -> HashMap<String, String> {
+    let map = labels
+        .lines()
+        .filter_map(|line| {
+            let mut parts = line.splitn(2, "=").collect::<Vec<&str>>();
+            parts.pop().and_then(move |val| {
+                parts.pop().map(move |key|
+                // Trim double quotes in value, present by default when attached through k8s downwardAPI
+                (key.to_string(), val.trim_matches('"').to_string()))
+            })
```
hawkw, line 562: Nit: We could avoid collecting a temporary `Vec` if we used the [`Iterator::next`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#tymethod.next) method on [`SplitN`](https://doc.rust-lang.org/std/str/struct.SplitN.html), rather than `Vec::pop`. Like so:
```suggestion
            let mut parts = line.splitn(2, "=");
            parts.next().and_then(move |key| {
                parts.next().map(move |val|
                // Trim double quotes in value, present by default when attached through k8s downwardAPI
                (key.to_string(), val.trim_matches('"').to_string()))
            })
```

linkerd/app/src/env.rs
Pothulapati: Hmm, Interesting. How does this coupling differ from the ENV coupling we did for the `metadata.labels` file injection, They both use env variables added by the injector, right? why is it temporary?

This could be solved easily by making the proxy injector also, add the `namespace` as a label to the pod, Just want to understand why this is a strong negative :)

linkerd/app/src/env.rs
```diff
@@ -500,6 +520,51 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
+        })
+        .collect::<HashMap<String, String>>();
+    map
```
hawkw, line 565: Nit: the `let`-binding for `map` isn't really necessary here, we could just return it directly. Like:
```suggestion
    labels
        .lines()
        .filter_map(|line| {
            let mut parts = line.splitn(2, "=").collect::<Vec<&str>>();
            parts.pop().and_then(move |val| {
                parts.pop().map(move |key|
                // Trim double quotes in value, present by default when attached through k8s downwardAPI
                (key.to_string(), val.trim_matches('"').to_string()))
            })
        })
        .collect::<HashMap<String, String>>()
```

linkerd/app/src/env.rs
olix0r: I'd rather that the proxy not know anything about namespaces at all. The proxy doesn't know about Kubernetes, and namespaces are a kubernetes concept. The only coupling the proxy has with Kubernetes are the default values for the resolution suffix and, now, the path to the metadata labels. I'd prefer that we not default any of these values, to be hoenst.

The proxy exposes a set of `LINKERD2_PROXY_*` environment variables that configure the proxy. The proxy injector sets those, but other things may set them as well, like a test harness, for instance.  The proxy injector uses this environment as an API, basically. The proxy shouldn't have any awareness of how the proxy injector goes about building those values. In this case `_pod_ns` is just part of that implementation detail of how the injector builds identity strings.

I'd prefer that we expose a tracing-specific environment variable that can be used to augment the attributes, rather than specifically exposing a _namespace_ that then goes and does that. The proxy doesn't care about namespaces, it cares about tracing attributes.

Pothulapati: > In this case _pod_ns is just part of that implementation detail of how the injector builds identity strings.
Yep, That makes sense. `namespace` is not an important detail for the proxy and should not be a dependency. I will update the control-plane PR to also include `namespace` as a label, thus making the tracing attributes path, the only way of adding attributes to traces!

linkerd/app/src/env.rs
Pothulapati: Yep, Definitely make sense :+1:

linkerd/app/src/env.rs
Pothulapati: Hmm, But we need a  `std::iter::Iterator<Item=(std::string::String, std::string::String)>`
So the compiler complains as

`value of type "std::collections::HashMap<std::string::String, std::string::String>" cannot be built from "std::iter::Iterator<Item=&(std::string::String, std::string::String)>"`

linkerd/app/src/env.rs
Pothulapati: Hmm, removed the warning! but still working on https://github.com/linkerd/linkerd2/pull/4199, to make the `env`, `volumeMounts` present only when tracing is enabled. (which are present by default before)

linkerd/app/src/env.rs
hawkw: ah, never mind. not a big deal, anyway.

linkerd/app/src/env.rs
```diff
@@ -71,6 +72,8 @@ pub const ENV_OUTBOUND_ROUTER_MAX_IDLE_AGE: &str = "LINKERD2_PROXY_OUTBOUND_ROUT
 pub const ENV_INBOUND_MAX_IN_FLIGHT: &str = "LINKERD2_PROXY_INBOUND_MAX_IN_FLIGHT";
 pub const ENV_OUTBOUND_MAX_IN_FLIGHT: &str = "LINKERD2_PROXY_OUTBOUND_MAX_IN_FLIGHT";
 
+pub const ENV_LABELS_FILE_PATH: &str = "LINKERD2_PROXY_LABELS_FILE_PATH";
```
olix0r, line 75: Sorry, I think this is my last issue... Can we rename the env var/constants to indicate this is for tracing? And, seeing that OC calls these _attributes_, I'm inclined to use that terminology (i.e. to indicate that this is actually not specifically tied to k8s or prometheus labels). Perhaps `LINKERD2_PROXY_TRACE_ATTRIBUTES_PATH`

Pothulapati: No problem, happy to do it. Updated all the terminology from `labels` to `attributes`

PR #464, opened 2020-03-30
Time out protocol detect futures

zaharidichev:
This PR changes the Accept trait to return a Future<Item = Future<()>>. The inner future is the one driving the connection while the outer future represents work such as protocol detect. This allows us to time out the outer future independently of the inner one. Additionally a timeout has been introduced to guard against serve first connections that do not run on a standard for the protocol port hanging and causing memory issues.

Fixes linkerd/linkerd2#4069

### DIFF
--- a/linkerd/app/core/src/accept_error.rs
+++ b/linkerd/app/core/src/accept_error.rs
@@ -17,10 +17,18 @@ impl<A> AcceptError<A> {
 impl<T, A> tower::Service<T> for AcceptError<A>
 where
     A: Accept<T>,
+    A::ConnectionError: 'static,
     A::ConnectionFuture: Send + 'static,
     A::Future: Send + 'static,
 {
-    type Response = Box<dyn Future<Item = (), Error = Never> + Send + 'static>;
+    type Response = future::Either<
+        future::OrElse<
+            A::ConnectionFuture,
+            Result<(), Never>,
+            fn(A::ConnectionError) -> Result<(), Never>,
+        >,
+        future::FutureResult<(), Never>,
+    >;
     type Error = Never;
     type Future = future::Then<
         A::Future,
@@ -37,18 +45,16 @@ where
     }
 
     fn call(&mut self, sock: T) -> Self::Future {
-        self.0.accept(sock).then(|x| match x {
-            Ok(connection_future) => Ok(Box::new(connection_future.then(|v| {
-                if let Err(e) = v {
-                    let error: Error = e.into();
-                    tracing::debug!(%error, "Connection failed");
-                }
+        self.0.accept(sock).then(|f| match f {
+            Ok(connection_future) => Ok(future::Either::A(connection_future.or_else(|e| {
+                let error: Error = e.into();
+                tracing::debug!(%error, "Connection failed");
                 Ok(())
             }))),
             Err(e) => {
                 let error: Error = e.into();
                 tracing::debug!(%error, "Accept failed");
-                Ok(Box::new(future::ok(())))
+                Ok(future::Either::B(future::ok(())))
             }
         })
     }
--- a/linkerd/app/core/src/accept_error.rs
+++ b/linkerd/app/core/src/accept_error.rs
@@ -1,4 +1,4 @@
-use futures::{future, Future, Poll};
+use futures::{Future, Poll};
 use linkerd2_error::{Error, Never};
 use linkerd2_proxy_core::listen::Accept;
 use tracing;
@@ -17,11 +17,12 @@ impl<A> AcceptError<A> {
 impl<T, A> tower::Service<T> for AcceptError<A>
 where
     A: Accept<T>,
+    A::ConnectionFuture: Send + 'static,
+    A::Future: Send + 'static,
 {
-    type Response = ();
-    type Error = Never;
-    type Future =
-        future::Then<A::Future, Result<(), Never>, fn(Result<(), A::Error>) -> Result<(), Never>>;
+    type Response = Box<dyn Future<Item = (), Error = Never> + Send + 'static>;
+    type Error = A::Error;
+    type Future = Box<dyn Future<Item = Self::Response, Error = Self::Error> + Send + 'static>;
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
         Ok(self
@@ -32,12 +33,14 @@ where
     }
 
     fn call(&mut self, sock: T) -> Self::Future {
-        self.0.accept(sock).then(|v| {
-            if let Err(e) = v {
-                let error: Error = e.into();
-                tracing::debug!(%error, "Connection failed");
-            }
-            Ok(())
-        })
+        Box::new(self.0.accept(sock).map(|connection_future| {
+            Box::new(connection_future.then(|v| {
+                if let Err(e) = v {
+                    let error: Error = e.into();
+                    tracing::debug!(%error, "Connection failed");
+                }
+                Ok(())
+            })) as Self::Response
+        })) as Self::Future
     }
 }
--- a/linkerd/app/core/src/accept_error.rs
+++ b/linkerd/app/core/src/accept_error.rs
@@ -1,4 +1,4 @@
-use futures::{Future, Poll};
+use futures::{future, Future, Poll};
 use linkerd2_error::{Error, Never};
 use linkerd2_proxy_core::listen::Accept;
 use tracing;
@@ -21,8 +21,12 @@ where
     A::Future: Send + 'static,
 {
     type Response = Box<dyn Future<Item = (), Error = Never> + Send + 'static>;
-    type Error = A::Error;
-    type Future = Box<dyn Future<Item = Self::Response, Error = Self::Error> + Send + 'static>;
+    type Error = Never;
+    type Future = future::Then<
+        A::Future,
+        Result<Self::Response, Never>,
+        fn(Result<A::ConnectionFuture, A::Error>) -> Result<Self::Response, Never>,
+    >;
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
         Ok(self
@@ -33,14 +37,19 @@ where
     }
 
     fn call(&mut self, sock: T) -> Self::Future {
-        Box::new(self.0.accept(sock).map(|connection_future| {
-            Box::new(connection_future.then(|v| {
+        self.0.accept(sock).then(|x| match x {
+            Ok(connection_future) => Ok(Box::new(connection_future.then(|v| {
                 if let Err(e) = v {
                     let error: Error = e.into();
                     tracing::debug!(%error, "Connection failed");
                 }
                 Ok(())
-            })) as Self::Response
-        })) as Self::Future
+            }))),
+            Err(e) => {
+                let error: Error = e.into();
+                tracing::debug!(%error, "Accept failed");
+                Ok(Box::new(future::ok(())))
+            }
+        })
     }
 }
--- a/linkerd/app/core/src/admin/mod.rs
+++ b/linkerd/app/core/src/admin/mod.rs
@@ -81,7 +81,7 @@ impl<M: FmtMetrics> Service for Admin<M> {
 impl<M: FmtMetrics + Clone + Send + 'static> svc::Service<Connection> for Accept<M> {
     type Response = Box<dyn Future<Item = (), Error = hyper::error::Error> + Send + 'static>;
     type Error = Error;
-    type Future = Box<dyn Future<Item = Self::Response, Error = Self::Error> + Send + 'static>;
+    type Future = future::FutureResult<Self::Response, Self::Error>;
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
         Ok(().into())
@@ -99,7 +99,7 @@ impl<M: FmtMetrics + Clone + Send + 'static> svc::Service<Connection> for Accept
         });
 
         let connection_future: Self::Response = Box::new(self.1.serve_connection(io, svc));
-        Box::new(future::ok(connection_future))
+        future::ok(connection_future)
     }
 }
 
--- a/linkerd/app/core/src/admin/mod.rs
+++ b/linkerd/app/core/src/admin/mod.rs
@@ -98,7 +98,7 @@ impl<M: FmtMetrics + Clone + Send + 'static> svc::Service<Connection> for Accept
             svc.call(req)
         });
 
-        let connection_future: Self::Response = Box::new(self.1.serve_connection(io, svc));
+        let connection_future = Box::new(self.1.serve_connection(io, svc));
         future::ok(connection_future)
     }
 }
--- a/linkerd/app/core/src/admin/mod.rs
+++ b/linkerd/app/core/src/admin/mod.rs
@@ -8,6 +8,7 @@ use futures::{future, Future, Poll};
 use http::StatusCode;
 use hyper::service::{service_fn, Service};
 use hyper::{Body, Request, Response};
+use linkerd2_error::Error;
 use linkerd2_metrics::{self as metrics, FmtMetrics};
 use std::io;
 
@@ -78,8 +79,8 @@ impl<M: FmtMetrics> Service for Admin<M> {
 }
 
 impl<M: FmtMetrics + Clone + Send + 'static> svc::Service<Connection> for Accept<M> {
-    type Response = ();
-    type Error = hyper::error::Error;
+    type Response = Box<dyn Future<Item = (), Error = hyper::error::Error> + Send + 'static>;
+    type Error = Error;
     type Future = Box<dyn Future<Item = Self::Response, Error = Self::Error> + Send + 'static>;
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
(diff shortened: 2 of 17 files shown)


linkerd/proxy/core/Cargo.toml
```diff
@@ -13,3 +13,4 @@ futures = "0.1"
 linkerd2-error = { path = "../../error" }
 tokio = "0.1"
 tower = "0.1"
+tracing = "0.1"
```
olix0r, line 16: This seems to be unused?

linkerd/proxy/core/src/listen.rs
```diff
@@ -118,10 +127,14 @@ where
...
         loop {
-            try_ready!(self.accept.poll_ready());
+            try_ready!(self.accept.poll_ready().map_err(Into::into));
             let conn = try_ready!(self.listen.poll_accept().map_err(Into::into));
-            let accept = self.accept.accept(conn).map_err(|e| match e {});
-            self.executor.spawn(Box::new(accept)).map_err(Error::from)?;
+            let accept = self.accept.accept(conn).map_err(|_| {});
+            self.executor
+                .spawn(Box::new(
+                    accept.and_then(|conn_future| conn_future.map_err(|_| {})),
```
olix0r, line 135: If I understand correctly, this just silently drops errors? I think we need to restore the `Error = Never` constraints (for both error types) and modify accept_error to handle both.

linkerd/app/core/src/accept_error.rs
```diff
@@ -32,12 +33,14 @@ where
...
-        })
+        Box::new(self.0.accept(sock).map(|connection_future| {
+            Box::new(connection_future.then(|v| {
+                if let Err(e) = v {
+                    let error: Error = e.into();
+                    tracing::debug!(%error, "Connection failed");
+                }
+                Ok(())
+            })) as Self::Response
+        })) as Self::Future
```
olix0r, line 44: Are these `as` directives load-bearing?

Regardless, I think in this case, we should prefer concrete future types to Boxed types. Given this is such a fundamental part of the data path, we should avoid unnecessary alloc when it's easy.

linkerd/proxy/core/src/listen.rs
hawkw: Yeah, IIRC, when we were pairing on this, we noted that the `accept_error`  layer should be updated to handle both future's errors?

linkerd/proxy/tcp/src/forward.rs
```diff
@@ -27,23 +27,27 @@ impl<C> Forward<C> {
...
     C: Service<T>,
+    C::Response: Send + 'static,
+    C::Future: Send + 'static,
     C::Error: Into<Error>,
     C::Response: AsyncRead + AsyncWrite,
-    I: AsyncRead + AsyncWrite,
+    I: AsyncRead + AsyncWrite + Send + 'static,
 {
-    type Response = ();
+    type Response = Box<dyn Future<Item = (), Error = Error> + Send + 'static>;
```
hawkw, line 36: This is always the same type, so I don't think we should have to box it:
```suggestion
    type Response = ForwardFuture<I, C::Future>;
```

linkerd/proxy/tcp/src/forward.rs
```diff
@@ -27,23 +27,27 @@ impl<C> Forward<C> {
...
     type Error = Error;
-    type Future = ForwardFuture<I, C::Future>;
+    type Future = Box<dyn Future<Item = Self::Response, Error = Error> + Send + 'static>;
```
hawkw, line 38: And this is always a `future::ok`, so again, it shouldn't need to be boxed
```suggestion
    type Future = futures::future::FutureResult<Self::Response, Error>;
```

linkerd/proxy/tap/src/accept.rs
```diff
@@ -58,9 +58,9 @@ impl AcceptPermittedClients {
 }
 
 impl Service<Connection> for AcceptPermittedClients {
-    type Response = ();
+    type Response = Box<dyn Future<Item = (), Error = Error> + Send + 'static>;
     type Error = Error;
-    type Future = ServeFuture;
+    type Future = Box<dyn Future<Item = Self::Response, Error = Self::Error> + Send + 'static>;
```
hawkw, line 63: This is always a `FutureResult`, so it doesn't need to be boxed:
```suggestion
    type Future = future::FutureResult<Self::Response, Self::Error>>;
```

linkerd/proxy/tap/src/accept.rs
```diff
@@ -58,9 +58,9 @@ impl AcceptPermittedClients {
 }
 
 impl Service<Connection> for AcceptPermittedClients {
-    type Response = ();
+    type Response = Box<dyn Future<Item = (), Error = Error> + Send + 'static>;
```
hawkw, line 61: This looks like it's always a `ServeFuture`, so we don't need to box it:
```suggestion
    type Response = ServeFuture;
```

linkerd/app/core/src/accept_error.rs
hawkw: Yeah, it looks like there is only ever one possible `Future` type returned here, and one possible `Response` type for that future?

linkerd/proxy/core/src/listen.rs
zaharidichev: makes total sense. Just to solidify my understanding a `.map_err(|e| match e {})` when `Error = Never` works and does not complain about non exhaustive matches because the compiler knows that Never cannot be instantiated. Correct? And I assume this further serves as a proove that this future can actually never error. This is quite neat.

linkerd/app/core/src/proxy/server.rs
```diff
@@ -173,41 +183,44 @@ where
...
+                        .serve_connection(io, HyperServerSvc::new(http_svc));
+                    Either::B(
+                        drain
+                            .watch(conn, |conn| conn.graceful_shutdown())
+                            .map(|_| ())
+                            .map_err(Into::into),
+                    )
+                }
+            },
+        )))
```
hawkw, line 223: I noticed that, since we're boxing the future returned by this match expression, I don't think we need to use `Either` here. I think we can simplify this to (although this was already the case before this PR.)
```suggestion
        Box::new(futures::future::ok::<Self::Response, Error>(
            match http_version {
                HttpVersion::Http1 => {
                    // Enable support for HTTP upgrades (CONNECT and websockets).
                    let svc = upgrade::Service::new(http_svc, drain.clone());
                    let exec =
                        tokio::executor::DefaultExecutor::current().instrument(info_span!("http1"));
                    let conn = builder
                        .with_executor(exec)
                        .http1_only(true)
                        .serve_connection(io, HyperServerSvc::new(svc))
                        .with_upgrades();
                    Box::new(
                        drain
                            .watch(conn, |conn| conn.graceful_shutdown())
                            .map(|_| ())
                            .map_err(Into::into),
                    )
                }

                HttpVersion::H2 => {
                    let exec =
                        tokio::executor::DefaultExecutor::current().instrument(info_span!("h2"));
                    let conn = builder
                        .with_executor(exec)
                        .http2_only(true)
                        .http2_initial_stream_window_size(initial_stream_window_size)
                        .http2_initial_connection_window_size(initial_conn_window_size)
                        .serve_connection(io, HyperServerSvc::new(http_svc));
                    Box::new(
                        drain
                            .watch(conn, |conn| conn.graceful_shutdown())
                            .map(|_| ())
                            .map_err(Into::into),
                    )
                }
            },
        ))
```
Note that this was already the case before this PR, but I only just noticed it. :)

zaharidichev: Maybe i am way off here, but does that not introduce one additional allocation in case we hit this codepath, or I am failing to understand how that works?

hawkw: Nope, there are the same number of boxes in the returned future. In the current code, we return a boxed `FutureResponse` with a boxed `Either` in it, while in the suggested code we return a boxed `FutureResponse` with a boxed future for each match arm. If that makes sense? There's the same number of boxes.

zaharidichev: Got it!. Makes sense. Thanks a lot !

linkerd/proxy/core/src/listen.rs
olix0r: That's right. Never is just `enum Never {}` with some implementations for Error, Display, etc.

linkerd/app/core/src/admin/mod.rs
```diff
@@ -96,7 +97,9 @@ impl<M: FmtMetrics + Clone + Send + 'static> svc::Service<Connection> for Accept
             req.extensions_mut().insert(ClientAddr(peer));
             svc.call(req)
         });
-        Box::new(self.1.serve_connection(io, svc))
+
+        let connection_future: Self::Response = Box::new(self.1.serve_connection(io, svc));
+        future::ok(connection_future)
```
olix0r, line 102: Is the type assertion necessary?

linkerd/proxy/tcp/src/forward.rs
```diff
@@ -27,23 +27,25 @@ impl<C> Forward<C> {
...
-    type Future = ForwardFuture<I, C::Future>;
+    type Future = futures::future::FutureResult<Self::Response, Error>;
 
     fn poll_ready(&mut self) -> Poll<(), self::Error> {
         self.connect.poll_ready().map_err(Into::into)
     }
 
     fn call(&mut self, (meta, io): (T, I)) -> Self::Future {
-        ForwardFuture::Connect {
+        futures::future::ok::<Self::Response, Error>(ForwardFuture::Connect {
```
olix0r, line 45: Are these type annotations necessary?

linkerd/proxy/tap/src/accept.rs
```diff
@@ -70,15 +70,19 @@ impl Service<Connection> for AcceptPermittedClients {
...
                 } else {
-                    self.serve_unauthenticated(io, format!("Unauthorized peer: {}", peer))
+                    future::ok::<Self::Response, Self::Error>(
+                        self.serve_unauthenticated(io, format!("Unauthorized peer: {}", peer)),
+                    )
                 }
             }
             Conditional::None(ReasonForNoIdentity::NoPeerName(ReasonForNoPeerName::Loopback)) => {
-                self.serve_authenticated(io)
+                future::ok::<Self::Response, Self::Error>(self.serve_authenticated(io))
```
olix0r, line 81: Similarly

linkerd/proxy/tap/src/accept.rs
```diff
@@ -70,15 +70,19 @@ impl Service<Connection> for AcceptPermittedClients {
         match meta.peer_identity {
             Conditional::Some(ref peer) => {
                 if self.permitted_client_ids.contains(peer) {
-                    self.serve_authenticated(io)
+                    future::ok::<Self::Response, Self::Error>(self.serve_authenticated(io))
```
olix0r, line 73: Similarly

linkerd/proxy/tap/src/accept.rs
```diff
@@ -70,15 +70,19 @@ impl Service<Connection> for AcceptPermittedClients {
...
                 } else {
-                    self.serve_unauthenticated(io, format!("Unauthorized peer: {}", peer))
+                    future::ok::<Self::Response, Self::Error>(
```
olix0r, line 75: etc

linkerd/proxy/tap/src/accept.rs
```diff
@@ -70,15 +70,19 @@ impl Service<Connection> for AcceptPermittedClients {
...
+                        self.serve_unauthenticated(io, format!("Unauthorized peer: {}", peer)),
+                    )
                 }
             }
             Conditional::None(ReasonForNoIdentity::NoPeerName(ReasonForNoPeerName::Loopback)) => {
-                self.serve_authenticated(io)
+                future::ok(self.serve_authenticated(io))
+            }
+            Conditional::None(reason) => {
+                future::ok(self.serve_unauthenticated(io, reason.to_string()))
```
hawkw, line 84: Nit, take it or leave it: it looks like we wrap each match arm in `future::ok`. We should be able to simplify this a little by wrapping the entire match in a `future::ok`, like:
```rust
future::ok(match meta.peer_identity { 
    Conditional::Some(ref peer) => {
        if self.permitted_client_ids.contains(peer) {
            self.serve_authenticated(io)
        } else {
            // ...
        }
    }
    // ...
})
```

linkerd/app/core/src/accept_error.rs
```diff
@@ -32,12 +37,19 @@ where
...
-        self.0.accept(sock).then(|v| {
-            if let Err(e) = v {
+        self.0.accept(sock).then(|x| match x {
+            Ok(connection_future) => Ok(Box::new(connection_future.then(|v| {
+                if let Err(e) = v {
+                    let error: Error = e.into();
+                    tracing::debug!(%error, "Connection failed");
+                }
+                Ok(())
+            }))),
```
hawkw, line 47: Nit, take it or leave it: I think we can get rid of the `if let Err(e) = v` if we change the `then` to `or_else`:
```suggestion
            Ok(connection_future) => Ok(Box::new(connection_future.or_else(|e| {
                let error: Error = e.into();
                tracing::debug!(%error, "Connection failed");
                Ok(())
            }))),
```
`then` closures are called with the result of a future, whether it's `Ok` _or_ `Err`, while `or_else` is called only with an `Err` result but allows returning `Ok` to recover.

linkerd/app/core/src/proxy/server.rs
```diff
@@ -163,8 +166,15 @@ where
...
                     .clone()
                     .into_service()
-                    .oneshot((proto.tls, io));
-                return Box::new(drain.watch(fwd.map_err(Into::into), |_| {}));
+                    .oneshot((proto.tls, io))
+                    .map(|connect_future| {
+                        let connect_future = drain
+                            .watch(connect_future.map_err(Into::into), |_| {})
+                            .map_err(Into::into);
+                        Box::new(connect_future) as Self::Response
```
hawkw, line 174: Little surprised this cast is necessary --- have you tried removing it? Does the compiler complain?

linkerd/app/inbound/src/lib.rs
```diff
@@ -283,15 +285,21 @@ impl Config {
...
-                disable_protocol_detection_for_ports.clone(),
             );
 
-            // Terminate inbound mTLS from other outbound proxies.
-            let accept = tls::AcceptTls::new(local_identity, tcp_server)
-                .with_skip_ports(disable_protocol_detection_for_ports);
+            let tcp_detect = svc::stack(tcp_server)
+                .push(DetectProtocolLayer::new(ProtocolDetect::new(
+                    disable_protocol_detection_for_ports.clone(),
+                )))
```
hawkw, line 293: Nit, take it or leave it: I find it a _little_ weird we have to constuct a `DetectProtocolLayer` _and_ a `ProtocolDetect` here. Does it make sense to have one constructor (maybe for `DetectProtocolLayer`) that also builds the `ProtocolDetect`?

zaharidichev:
@hawkw I will add the test in a follow up PR if thats fine. At least it will be a good educational experience.

linkerd/app/core/src/accept_error.rs
```diff
@@ -17,11 +17,16 @@ impl<A> AcceptError<A> {
 impl<T, A> tower::Service<T> for AcceptError<A>
 where
     A: Accept<T>,
+    A::ConnectionFuture: Send + 'static,
+    A::Future: Send + 'static,
 {
-    type Response = ();
+    type Response = Box<dyn Future<Item = (), Error = Never> + Send + 'static>;
```
olix0r, line 23: https://github.com/linkerd/linkerd2-proxy/pull/471

linkerd/app/core/src/accept_error.rs
olix0r: addressed in https://github.com/linkerd/linkerd2-proxy/pull/471

linkerd/app/inbound/src/lib.rs
olix0r: IIRC `ProtocolDetect` is a concrete strategy and the layer is generic.

zaharidichev: Yes, that is correct.

linkerd/app/core/src/proxy/server.rs
zaharidichev: @hawkw I have, it errors out with: 

```
177 |return Box::new(fwd.map_err(Into::into));
    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected trait `futures::future::Future`, found struct `futures::future::map_err::MapErr`
```

olix0r:
I made just a few superficial changes: https://github.com/linkerd/linkerd2-proxy/pull/464/files/842dd496e4039104a7e1ea26ee24df9e3313d3df..39d3e1d83bc6ba6c7873febbf37cf63fd54c11c2

linkerd/proxy/detect/src/lib.rs
```diff
@@ -128,15 +125,12 @@ where
                             target.take().expect("polled after complete"),
                             io.prefix().as_ref(),
                         );
-                        let inner_fut = accept.accept((target, BoxedIo::new(io)));
```
olix0r: Just noting that there was a bug lurking in this change: we were no longer polling the accept to ready before dispatching. These changes were just superficial anyway, it seems, so I've backed this out to be as it was before.

PR #466, opened 2020-03-30
Add a `mock-orig-dst` feature flag

olix0r:
The proxy code is generic over the strategy it uses for determining the original destination address for a given connection. In integration tests, the strategy is replaced with a configured version.

The `mock-orig-dst` feature modifies the implementation to be swappable at compile-time via feature flag. This removes integration-test specific test code, and makes it generally possible to start the proxy outside of Kubernetes (via the compile-time flag). When this feature is enabled, runtime configuration is required via a pair of new environment variables.

There was an integration test that tested the proxy's behavior when an address could not be determined, but that test has been removed as that use case is not particularly important.

To build a proxy in this mode, one runs:
```
:; (cd linkerd2-proxy && cargo build --release --features=mock-orig-dst)
```


linkerd/proxy/api-resolve/src/metadata.rs
```diff
@@ -48,7 +44,6 @@ impl Metadata {
             protocol_hint: ProtocolHint::Unknown,
             identity: None,
             weight: 10_000,
-            authority_override: None,
```
kleimkuhler: Is removing `authority_override` code related to this change?

linkerd/proxy/transport/src/listen.rs
```diff
@@ -38,8 +38,20 @@ pub struct Addrs {
...
+pub use self::sys::SysOrigDstAddr;
+
+#[cfg(not(feature = "mock-orig-dst"))]
+pub type DefaultOrigDstAddr = SysOrigDstAddr;
+
+#[cfg(feature = "mock-orig-dst")]
+pub use self::mock::MockOrigDstAddr;
+
+#[cfg(feature = "mock-orig-dst")]
+pub type DefaultOrigDstAddr = MockOrigDstAddr;
```
kleimkuhler: nit: how about?

```
#[cfg(not(feature = "mock-orig-dst"))]
pub type DefaultOrigDstAddr = self::sys::SysOrigDstAddr;

#[cfg(feature = "mock-orig-dst")]
pub type DefaultOrigDstAddr = self::mock::MockOrigDstAddr;
```

linkerd/proxy/api-resolve/src/metadata.rs
olix0r: No, thanks. This is a merge-o

linkerd/proxy/transport/src/listen.rs
```diff
@@ -38,8 +38,14 @@ pub struct Addrs {
...
-#[derive(Copy, Clone, Debug)]
-pub struct SysOrigDstAddr(());
+// The mock-orig-dst feature disables use of the syscall-based OrigDstAddr implementation and
+// replaces it with one that must be configured.
+
+#[cfg(not(feature = "mock-orig-dst"))]
+pub use self::sys::SysOrigDstAddr as DefaultOrigDstAddr;
+
+#[cfg(feature = "mock-orig-dst")]
+pub use self::mock::MockOrigDstAddr as DefaultOrigDstAddr;
```
kleimkuhler: 👍

hawkw:
_Technically_, I don't think this should be a feature flag. Cargo features are intended to be strictly additive in nature, since they can be enabled by dependencies. It might be more correct to use `cfg(mock_orig_dst)` rather than `cfg(feature = "mock-orig-dst")`; then, we would set the configurations with `RUSTFLAGS="--cfg mock-orig-dst"` when building and it couldn't be enabled by a dependency.

With that said, I don't think this actually matters for linkerd. Since linkerd is a binary application rather than a library, we don't need to worry about dependent code enabling conflicting features. So, I am not sure if this actually matters at all; I just thought it was worth mentioning.

linkerd/app/outbound/src/lib.rs
```diff
@@ -48,7 +48,7 @@ const EWMA_DEFAULT_RTT: Duration = Duration::from_millis(30);
 const EWMA_DECAY: Duration = Duration::from_secs(10);
 
 #[derive(Clone, Debug)]
-pub struct Config<A: OrigDstAddr = SysOrigDstAddr> {
+pub struct Config<A: OrigDstAddr = DefaultOrigDstAddr> {
     pub proxy: ProxyConfig<A>,
```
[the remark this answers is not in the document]
hawkw: Actually, since `app-core` dependens on `transport`, it seems like we _could_ just remove the generics everywhere, get rid of the `OrigDstAddr` trait, and and put everything in inherent methods on `DefaultOrigDstAddr` (which could maybe lose the default).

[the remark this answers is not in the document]
olix0r: Not quite. We use an alternate implementation when we don't need any original dsts (i.e. admin, tap servers). But I've simplified it.

olix0r:
> Technically, I don't think this should be a feature flag.

I was wondering about this. The benefit to using feature flags is that the integration tests can just depend on the feature rather than requiring compiler settings. I'm going to merge as is but happy to revisit if it can be improved.

hawkw:
> The benefit to using feature flags is that the integration tests can just depend on the feature rather than requiring compiler settings.

Ah, yeah, I hadn't thought of that. Let's do it this way, then — since Linkerd is not a library with dependents, I doubt the feature flag will cause issues.

PR #467, opened 2020-03-31
http-metrics: Make latency export optional

olix0r:
We export raw histograms for all HTTP metrics layers; but in some cases
these values are never consumed. This change modifies the http metrics
`Report` type to support disabling export of latencies.

This feature is not used yet but will be used in a follow-up.

---

We can revisit how to skip recording as well, but in order to address https://github.com/linkerd/linkerd2/issues/4215 expediently, I think addressing this only on the report-side is a good stopgap.


olix0r:
@adleong that's right. as i noted in the description, we can figure out if it make sense to go deeper with this, but in the name of expediency, I think this is the right change.

PR #469, opened 2020-04-02
inbound: Do not cache per-endpoint services

olix0r:
When the inbound proxy serves many unique targets on a single port,
requests may incorrectly timeout (especially, it seems, when the client
is attempting to reconnect).

This change creates an HTTP endpoint stack for each target (i.e.
Authority). This means that requests to distinct targets will always be
transported over distinct  localhost connections.

This problem was reproducible via:

```sh
for _ in $(seq 10) ; do
  for i in $(seq -w 100) ; do
    host="server-${i}.test.example.com"
    curl -sH "Host: $host" 127.0.0.1:4143 -w "$host\t%{http_code}\n"
  done
done
```

### DIFF
--- a/linkerd/app/inbound/src/lib.rs
+++ b/linkerd/app/inbound/src/lib.rs
@@ -100,35 +100,24 @@ impl Config {
                 })
                 .push(svc::layer::mk(tcp::Forward::new));
 
-            // Caches HTTP clients for each inbound port & HTTP settings.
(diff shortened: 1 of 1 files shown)


hawkw:
Somehow this PR is still open even though the branch has merged? Is this GitHub misbehaving?

kleimkuhler:
Yea I was just looking at this.. probably good to close

PR #470, opened 2020-04-10
Add liveness endpoint to admin server

adleong:
Relates to https://github.com/linkerd/linkerd2/issues/3984

Add a `/live` endpoint to the admin server which always returns a 200 success.  This can be used for liveness checking.

### DIFF
--- a/linkerd/app/core/src/admin/mod.rs
+++ b/linkerd/app/core/src/admin/mod.rs
@@ -59,6 +59,13 @@ impl<M: FmtMetrics> Admin<M> {
                 .expect("builder with known status code must not fail")
         }
     }
+
+    fn live_rsp(&self) -> Response<Body> {
+        Response::builder()
+            .status(StatusCode::OK)
+            .body("live\n".into())
+            .expect("builder with known status code must not fail")
+    }
 }
 
 impl<M: FmtMetrics> Service for Admin<M> {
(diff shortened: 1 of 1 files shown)


olix0r:
@adleong Why not use the existing /ready endpoint, which is at least tied to identitiy?

adleong:
See https://github.com/linkerd/linkerd2/issues/3984 for a fuller discussion.

Kubernetes differentiates liveness and readiness. The `/ready` endpoint only returns successfully if the proxy is ready to accept traffic, even if it is alive.  Currently, the `/metrics` endpoint is used to measure liveness.

olix0r:
But we don't want to consider the proxy live until it's received it's initial identity, do we? Afaict, that issue doesn't explain why we shouldn't use /ready, just that we shouldn't use /metrics.

zaharidichev:
@olix0r I do not think it will be considered live until the ready check returns true. I think the readyness checks need to pass and only then the liveliness checks kick in. So no I do not see a risk of k8s considering the proxy ready and alive before the identity is obtained.

grampelberg:
The endpoint will only be added to the service once readiness passes. So, having a liveness check that succeeds before the readiness check is okay. The liveness check is used to restart the container, so we'll want that to be tied to events that require restarting and not readiness itself.

There's probably no real harm having readiness and liveness checks be the same today, as the liveness check itself is mostly just making sure the lights are on. That said, we'd want to split it out in the future if there were reasons to trigger a restart (leaking internal structures, deadlocks, ...).

PR #472, opened 2020-04-11
docker: Use buildkit for caching

olix0r:
Our docker builds do not permit caching of dependencies or intermediate
build artifacts. However, Docker's new (experimental) buildkit features
add this functionality. I've been using this configuration locally for
some time, and it seems generaly useful enough to promote (especially
since this Dockerfile is only intended for development).

This change also untroduces the `PROXY_FEATURES` build-arg so that the
Dockerfile can be used to support profiling builds.

Furthermore, the `update-rust-versions.sh` script has been updated to
check versions and be more permissive about how it replaces versions in
the Dockerfile.

--

@linkerd/proxy-maintainers please verify that you're able to use this workflow, as documented in the Dockerfile.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -21,5 +21,4 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - name: Docker
-        if: github.event.pull_request.head.repo.fork
         run: DOCKER_UNOPTIMIZED=1 make docker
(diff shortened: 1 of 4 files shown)


.github/workflows/docker.yml
```diff
@@ -6,14 +6,14 @@ name: Docker
...
+      - master
     paths:
-    - 'Cargo.lock'
-    - 'Dockerfile'
+      - "Cargo.lock"
+      - "Dockerfile"
   pull_request:
     paths:
-    - 'Cargo.lock'
-    - 'Dockerfile'
```
olix0r, line 16: This was all changed by vscode...

PR #473, opened 2020-04-13
prepare to update the proxy to use std::future

hawkw:
This branch begins a "top-down" update of the proxy to use
`std::future`. I've started by commenting out most of the stack, making
the proxy into a raw TCP forwarder/HTTP server that just always returns
`200 OK <eof>`. Then, I've updated the top-level `serve` function to use
`std::future`. This is pretty mechanical, but we _may_ want to consider
doing some refactoring here in the future.

Currently, most of the integration tests are failing, since they test
functionality that is now commented out. I've added `ignore` attributes
temporarily to disable those tests. However, since we are ignoring the
tests rather than disabling them via conditional compilation, we will
still try to build them, so this ensures we don't accidentally break any
of these tests. There is a feature flag, "nyi", that can be enabled to
run the ignored tests, so that we can test if changes have made any
additional tests pass again. As functionality is re-enabled, we can
re-enable the corresponding tests as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,7 +21,7 @@ version = "0.6.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d6531d44de723825aa81398a6415283229725a00fa30713812ab9323faa82fc4"
 dependencies = [
- "memchr 2.0.1",
+ "memchr 2.3.3",
 ]
 
 [[package]]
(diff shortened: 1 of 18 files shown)


olix0r:
How do you want to handle integration tests on this branch? I think our options are:
* disable workflow in github
* keep integration tests compiling (using feature flags?)

hawkw:
It would be good to keep checking that the tests _compile_ though, even if we don't actually run them. Also, I think we'll want to be able to turn them back on incrementally as the relevant parts of the proxy are ported over. So I'm not sure if just turning off the integration workflow is really ideal? I think the right move would be to replace all the `#[test]` attributes in the integration tests with `#[ignore]` for now. That way, we the integration workflow will still compile them, and we can incrementally re-enable them. What do you think?

hawkw:
@olix0r okay, I think I've settled on a scheme for ignoring tests in 8eb6061. I've added `#[ignore]` attributes for all the tests that are currently failing. This will ensure that they still compile. I've also added a "nyi" feature flag that disables the `#[ignore]` attribute on these tests. That way, as changes are made, we can re-run the tests to see which ones are now passing, without having to change the test code, and we can then remove the `#[ignore]`s as functionality is re-enabled.

What do you think?

olix0r:
@hawkw that sounds reasonable

PR #474, opened 2020-04-13
port `drain` to std::future

hawkw:
This updates the `drain` crate to use std::future.

Note that because Tokio 0.2's `spawn` returns a `JoinHandle` that may
be used to await a task's completion, we can probably simplify the 
"waiting for tasks to drain" logic a bit. I didn't do that because I wanted
to do as close to a strict translation as possible, without adding possible
changes in behavior. Also, we are currently still using the compat runtime's
0.1-style spawning.

### DIFF
--- a/linkerd/drain/src/lib.rs
+++ b/linkerd/drain/src/lib.rs
@@ -1,10 +1,12 @@
 #![deny(warnings, rust_2018_idioms)]
 
-use futures::future::Shared;
-use futures::sync::{mpsc, oneshot};
-use futures::{try_ready, Async, Future, Poll, Stream};
+use futures::{future::Shared, FutureExt, TryFutureExt};
 use linkerd2_error::Never;
-use std::mem;
+use pin_project::pin_project;
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use tokio::sync::{mpsc, oneshot};
 
 /// Creates a drain channel.
 ///
@@ -17,7 +19,7 @@ pub fn channel() -> (Signal, Watch) {
         Signal { drained_rx, tx },
         Watch {
             drained_tx,
-            rx: rx.shared(),
+            rx: rx.map_err((|_| ()) as fn(_) -> _).shared(),
         },
     )
 }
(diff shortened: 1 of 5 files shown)


linkerd/drain/src/lib.rs
```diff
@@ -35,28 +37,23 @@ pub struct Signal {
...
-#[derive(Debug)]
-enum State<F> {
-    Watch(F),
-    Draining,
+    rx: Shared<
+        futures::future::MapErr<
+            oneshot::Receiver<()>,
+            fn(tokio::sync::oneshot::error::RecvError) -> (),
+        >,
+    >,
```
olix0r, line 50: What's the context on dropping the errors? Does this have something to do with Clone requirements? Is the error condition still meaningful to indicate the sender was dropped? If so, should we have a marker type? If not, can we drop the Result-ness of the future entirely?

linkerd/drain/src/lib.rs
```diff
@@ -83,184 +80,162 @@ impl Watch {
...
-        Watching {
-            future,
-            state: State::Watch(on_drain),
-            watch: self,
+        tokio::select! {
+            res = &mut future => res,
+            _ = self.rx => {
+                on_drain(&mut future);
+                (&mut future).await
+            }
```
olix0r, line 93: How do I interpret this? Is this an idiom that we're going to get intimately familiar with? If not, maybe warrants some explanation in comments...

olix0r: Oh, I think I get it. It's saying try to poll the future, and return its result, or receive the rx, and then wait on the future?

linkerd/drain/src/lib.rs
hawkw: The error only ever indicates that all senders have been dropped, and we still handle it the way we did before. IIRC, you're correct; the reason it now needs to be ignored is that it lost its `Clone` impl in tokio 0.2.

I think we could just replace this with a `then` and drop the resultiness of the future entirely, good call.

linkerd/drain/src/lib.rs
```diff
@@ -83,184 +80,162 @@ impl Watch {
...
-        }
+    /// Wrap a future to count it against the completion of the `Drained`
+    /// future that corresponds to this `Watch`.
+    ///
+    /// Unlike `Watch::watch`, this method does not take a callback that is
+    /// triggered on drain; the wrapped future will simply be allowed to
+    /// complete. However, like `Watch::watch`, the `Drained` future returned
+    /// by calling `drain` on the corresponding `Signal` will not complete until
+    /// the wrapped future finishes.
+    pub async fn with<A>(self, future: A) -> A::Output
```
olix0r, line 105: tioli: Perhaps better named `after`? I.e. `drain.after(fut)`

linkerd/drain/src/lib.rs
hawkw: Hmm, apparently this broke the build somehow... https://github.com/linkerd/linkerd2-proxy/pull/474/checks?check_run_id=596534255

hawkw: oh, huh, the breakage is unrelated. what the heck!

kleimkuhler:
@hawkw This can be closed now correct?

hawkw:
@kleimkuhler yup, I thought the `Closes #474` in that PR would do that?

kleimkuhler:
I think that may just be for closing issues

PR #475, opened 2020-04-14
Makefile: Use STRIP variable with strip as default

joakimr-axis:
If we run make in a cross-compile environment where the strip binary we
want to use is something else than strip, it is useful to have a STRIP
variable in the Makefile. If STRIP is not set, the default value is the
common strip.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ ifndef PACKAGE_VERSION
 	PACKAGE_VERSION = $(shell git rev-parse --short HEAD)
 endif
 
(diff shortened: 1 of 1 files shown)


olix0r:
@joakimr-axis Thanks for the submission. Out of curiosity, do you need to run `make package` at all? (As opposed to `make build`?) I ask because I'm in the middle of introducing [additional dependencies](https://github.com/linkerd/linkerd2-proxy/pull/476) to `make package`, as it's primarily intended to be run in the context of an automated release.

PR #476, opened 2020-04-14
Add checksec to the release process

olix0r:
A recent [Twitter thread][mudge] suggested that tools like
[`checksec`][checksec] be used to validate release binaries. Checksec
reports whether modern security features like stack canaries are
employed. Proxy builds appear to do pretty well out-of-the-box.

This change introduces a checksec.sh wrapper that is used by the
Makefile during packaging. A new _package_ github action is introduced
to provide `checksec` and `jq` dependencies at runtime. (Note: the
version of checksec provided by debian does not include JSON output, so
it is instead fetched directly from GitHub).

During an automated release, the generated checksec is compared to an
expected set of values and, if a regression is detected, the release
will fail.

[mudge]: https://twitter.com/dotMudge/status/1249359519471341569
[checksec]: https://github.com/slimm609/checksec.sh

### DIFF
--- a/.github/actions/package/expected-checksec.json
+++ b/.github/actions/package/expected-checksec.json
@@ -1,0 +1,10 @@
+{
+  "canary": "yes",
+  "fortify_source": "yes",
+  "nx": "yes",
+  "pie": "yes",
+  "relro": "full",
+  "rpath": "no",
+  "runpath": "no",
+  "symbols": "no"
+}
--- a/checksec.sh
+++ b/checksec.sh
@@ -9,6 +9,7 @@ require_from() {
     fi
 }
 
+require_from readelf     binutils
 require_from checksec    https://github.com/slimm609/checksec.sh
 require_from jq          https://stedolan.github.io/jq/
 
--- a/checksec.sh
+++ b/checksec.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+set -eu
+
+require_from() {
+    if ! command -v "$1" >/dev/null 2>/dev/null ; then
+        echo "Please acquire $1 from $2" >&2
+        exit 1
+    fi
+}
+
+require_from checksec    https://github.com/slimm609/checksec.sh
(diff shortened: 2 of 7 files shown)


checksec.sh
```diff
@@ -0,0 +1,27 @@
...
+
+require_from() {
+    if ! command -v "$1" >/dev/null 2>/dev/null ; then
+        echo "Please acquire $1 from $2" >&2
+        exit 1
+    fi
+}
+
+require_from checksec    https://github.com/slimm609/checksec.sh
+require_from jq          https://stedolan.github.io/jq/
```
hawkw, line 13: AFAIK `checksec` also requires a `readelf` executable...should we also validate that it exists and prompt the user to install `binutils` if it doesn't?

.github/actions/package/expected-checksec.json
```diff
@@ -0,0 +1,10 @@
+{
+  "canary": "yes",
+  "fortify_source": "yes",
+  "nx": "yes",
+  "pie": "yes",
+  "relro": "full",
+  "rpath": "no",
+  "runpath": "no",
+  "symbols": "no"
```
hawkw, line 9: nit/tioli: it might be nice to have comments explaining why we expect to see "no" for these?

olix0r: I didn't do too much research into all of the various checks checksec does, but i'd welcome additional docs.

I think in the case of _symbols_, particularly, the guidance is intended to make reverse-engineering more difficult. In our case, we mostly care that the binary is small (and we have another build option that includes symbols as a separate file)

olix0r: It appears that rpath/runpath can be used to influence dynamic linking in dangerous ways...

hawkw: yeah, i thought that was the case for symbols. I wasn't super familiar with the rpath/rumpath stuff.

PR #480, opened 2020-04-16
Remove the IntoAbsForm trait

olix0r:
@zaharidichev What do you think of this?

This change modifies the `HasSettings` trait to return a `Settings`
(rather than a reference). The `Settings` type is a simple type with a
few bools, so it's safe to copy freely.

This enables us to implement `HasSettings` for `http::HttpEndpoint` by
overriding the `was_absolute_from` when appropriate without having to
introduce an additional trait


hawkw:
Looks like this might've [broken something](https://github.com/linkerd/linkerd2-proxy/runs/590686577#step:5:565)?

olix0r:
@hawkw afaict that test may be flakey (passes locally)

PR #486, opened 2020-04-21
port `timeout` (and friends) to std::future

hawkw:
This branch ports the `linkerd2-timeout` crate and the
`linkerd2-proxy-http` crate's `timeout` module (which depends on it) to
use `std::future` and `tokio` 0.2's timers. I've also re-enabled the
`push_*` methods for `timeout`, `probe_ready`, `fail_fast`, and `idle`,
as well.

This change should be pretty straightforward to read, as it's fairly
mechanical in nature. I've basically just changed the code to account
for the new APIs, but the implementation should be quite similar.


olix0r:
To be clear... this depends on/follows #482? Should I review that one first?

hawkw:
@olix0r Yeah, that's correct. Sorry #482 is so huge, but it should unblock a _lot_ of small, isolated PRs to update individual crates (like this one).

hawkw:
Rebased onto `master-tokio-0.2`

hawkw:
> So happy to see the tokio::test cleanups!

The `tokio-test` crate also has a lot of really nice utilities for testing futures _outside_ of an actual tokio runtime. I haven't used them here because I also needed timers to work (for obvious reasons), but the `buffer` PR shows them off rather nicely: https://github.com/linkerd/linkerd2-proxy/pull/488/files#diff-7f4858c542866bece52c0a8dfbca74adR44-R116

linkerd/timeout/src/lib.rs
```diff
@@ -1,137 +1,141 @@
...
+    F: Future<Output = Result<T, E>>,
+    E: Into<Error>,
+{
+    type Output = Result<T, Error>;
+    #[project]
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        #[project]
+        match self.project() {
+            TimeoutFuture::Passthru(f) => f.poll(cx).map_err(Into::into),
+            TimeoutFuture::Timeout(f, duration) => {
```
kleimkuhler, line 128: We could consider using select here now too. I don't think we need to change it as part of this PR, but it may be worth considering just selecting over the call on `inner` and a `delay_for(duration)`. If there is no change in performance it may at least simplify the implementation a little.

hawkw: I didn't do that here because if we replaced this with an async function using `select!` we would need to add a layer of boxing in order to return it from a `Service` impl. 

In a follow up after we're on `std::future` on master, I want to look into switching to nightly Rust to allow returning `impl Future` as `Service`'s `Future` type, and rewrite a lot of our manual future implementations with async-await.

kleimkuhler: Sounds good

PR #487, opened 2020-04-21
port stack to std::future

hawkw:
This branch ports the `linkerd2-stack` crate to use Tower 0.3 and
`std::future` and tokio 0.2's timers. I've also re-enabled the
`push_*` methods for layers exported by that crate.

This change should be pretty straightforward to read, as it's fairly
mechanical in nature. I've basically just changed the code to account
for the new APIs, but the implementation should be quite similar.

### DIFF
--- a/linkerd/stack/src/fallback.rs
+++ b/linkerd/stack/src/fallback.rs
@@ -1,9 +1,11 @@
 //! A middleware that may retry a request in a fallback service.
-
-use futures::{Future, Poll};
+use futures::TryFuture;
 use linkerd2_error::Error;
+use pin_project::{pin_project, project};
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
 use tower::util::{Either, Oneshot, ServiceExt};
-
 /// A Layer that augments the underlying service with a fallback service.
 ///
 /// If the future returned by the primary service fails with an error matching a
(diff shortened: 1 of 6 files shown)


linkerd/stack/src/fallback.rs
```diff
@@ -21,13 +23,21 @@ pub struct Fallback<I, F, P = fn(&Error) -> bool> {
     predicate: P,
 }
 
-pub enum MakeFuture<A, B, P> {
+#[pin_project]
+pub struct MakeFuture<A, B, P> {
+    #[pin]
+    state: State<A, B, P>,
```
kleimkuhler, line 29: Any reason for adding the `State` enum here instead of keeping the `MakeFuture` as one?

linkerd/stack/src/make_ready.rs
```diff
@@ -1,18 +1,30 @@
...
 
 #[derive(Debug)]
 pub struct MakeReady<M, Req>(M, PhantomData<fn(Req)>);
 
+#[pin_project]
 #[derive(Debug)]
-pub enum MakeReadyFuture<F, S, Req> {
-    Making(F),
-    Ready(Ready<S, Req>),
+pub struct MakeReadyFuture<F, S, Req> {
```
kleimkuhler, line 17: Similar question here about making this a struct with an enum field, instead of keeping it as an enum.

linkerd/stack/src/fallback.rs
hawkw: Yes, it's so that we can repeatedly pin project the state on multiple iterations of a loop even if we use [`Pin::set`](https://doc.rust-lang.org/std/pin/struct.Pin.html#method.set) to change the value between iterations. Tower does this in a few places as well.

PR #492, opened 2020-04-24
update concurrency-limit to std::future

hawkw:
This branch updates the `concurrency-limit` middleware to std::future.
The new implementation uses the new owned semaphore permit API added in
Tokio 0.2.19 (tokio-rs/tokio#2421). Similarly to #490, the old
implementation could not be directly translated due to `tokio::sync`'s
`Semaphore` losing its `poll`-based API in 0.2.

### DIFF
--- a/linkerd/concurrency-limit/src/lib.rs
+++ b/linkerd/concurrency-limit/src/lib.rs
@@ -6,10 +6,13 @@
 
 #![deny(warnings, rust_2018_idioms)]
 
-use futures::{try_ready, Future, Poll};
-use linkerd2_error::Error;
+use pin_project::pin_project;
+use std::future::Future;
+use std::pin::Pin;
 use std::sync::Arc;
-use tokio_sync::semaphore::{Permit, Semaphore};
+use std::task::{Context, Poll};
+use std::{fmt, mem};
+use tokio::sync::{OwnedSemaphorePermit, Semaphore};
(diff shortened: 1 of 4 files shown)


linkerd/concurrency-limit/src/lib.rs
```diff
@@ -89,45 +94,42 @@ impl<T> ConcurrencyLimit<T> {
...
+                State::Ready(_) => {
+                    trace!(available = %self.semaphore.available_permits(), "permit acquired");
+                    return Poll::Ready(Ok(()));
+                }
+                State::Waiting(ref mut fut) => {
+                    tokio::pin!(fut);
+                    let permit = futures::ready!(fut.poll(cx));
+                    State::Ready(permit)
+                }
+                State::Empty => State::Waiting(Box::pin(self.semaphore.clone().acquire_owned())),
```
olix0r, line 115: I'm curious about this boxing...  Are we unable to name the type returned by `acquire_owned`?

hawkw: It's an `async fn`, so, yes, we are unable to name it except as a `dyn Future` trait object. Also, `Box::pin` ensures that the future is pinned, so that we can poll it while we _ourselves_ are not pinned (which `Service::poll_ready` does not guarantee).

PR #493, opened 2020-04-27
Handle GRPC body errors

zaharidichev:
This branch modifies the `Respond` trait by allowing it to modify the response's body. In errors responder, we wrap the body of the response to allow it to set the GRPC status on trailers in case we get errors when calling `poll_data` on the body. 

Fixes https://github.com/linkerd/linkerd2/issues/4262

### DIFF
--- a/linkerd/app/core/src/errors.rs
+++ b/linkerd/app/core/src/errors.rs
@@ -72,7 +72,7 @@ where
                         let mut error_trailers = http::HeaderMap::new();
                         let code = set_grpc_status(&error, &mut error_trailers);
                         warn!(
-                            "Handing body error: {:?} with grpc status {:?}",
+                            "Handling body error: {:?} with grpc status {:?}",
                             error, code
                         );
                         *trailers = Some(error_trailers);
@@ -107,13 +107,13 @@ impl<B: Default + hyper::body::Payload> Default for ResponseBody<B> {
     }
 }
 
-impl<A, B: Default + hyper::body::Payload> respond::NewRespond<http::Request<A>> for NewRespond<B> {
-    type ResponseIn = http::Response<B>;
-    type ResponseOut = http::Response<ResponseBody<B>>;
+impl<ReqB, RspB: Default + hyper::body::Payload>
+    respond::NewRespond<http::Request<ReqB>, http::Response<RspB>> for NewRespond<RspB>
+{
+    type Response = http::Response<ResponseBody<RspB>>;
+    type Respond = Respond<RspB>;
 
-    type Respond = Respond<B>;
-
-    fn new_respond(&self, req: &http::Request<A>) -> Self::Respond {
+    fn new_respond(&self, req: &http::Request<ReqB>) -> Self::Respond {
         match req.version() {
             http::Version::HTTP_2 => {
                 let is_grpc = req
@@ -134,14 +134,15 @@ impl<B> Clone for NewRespond<B> {
     }
 }
 
-impl<B: Default + hyper::body::Payload> respond::Respond for Respond<B> {
-    type ResponseIn = http::Response<B>;
-    type ResponseOut = http::Response<ResponseBody<B>>;
+impl<RspB: Default + hyper::body::Payload> respond::Respond<http::Response<RspB>>
+    for Respond<RspB>
+{
+    type Response = http::Response<ResponseBody<RspB>>;
 
     fn respond(
         &self,
-        reseponse: Result<Self::ResponseIn, Error>,
-    ) -> Result<Self::ResponseOut, Error> {
+        reseponse: Result<http::Response<RspB>, Error>,
+    ) -> Result<Self::Response, Error> {
         match reseponse {
             Ok(response) => Ok(response.map(|b| match *self {
                 Respond::Http2 { is_grpc } if is_grpc == true => ResponseBody::Grpc {
--- a/linkerd/app/core/src/errors.rs
+++ b/linkerd/app/core/src/errors.rs
@@ -12,8 +12,8 @@ use linkerd2_timeout::{error::ResponseTimeout, FailFastError};
 use tower_grpc::{self as grpc, Code};
 use tracing::{debug, warn};
 
-pub fn layer<B: Default>() -> respond::RespondLayer<NewRespond<B>> {
-    respond::RespondLayer::new(NewRespond(std::marker::PhantomData))
+pub fn layer() -> respond::RespondLayer<NewRespond> {
+    respond::RespondLayer::new(NewRespond(()))
 }
 
 #[derive(Clone, Default)]
@@ -36,12 +36,12 @@ pub enum Reason {
     Unexpected,
 }
 
-#[derive(Debug)]
-pub struct NewRespond<B>(std::marker::PhantomData<fn() -> B>);
+#[derive(Copy, Clone, Debug)]
+pub struct NewRespond(());
 
 #[derive(Copy, Clone, Debug)]
-pub enum Respond<B> {
-    Http1(http::Version, std::marker::PhantomData<fn() -> B>),
+pub enum Respond {
+    Http1(http::Version),
     Http2 { is_grpc: bool },
 }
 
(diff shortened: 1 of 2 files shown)


linkerd/app/core/src/errors.rs
```diff
@@ -159,6 +239,11 @@ fn set_grpc_status(error: &Error, headers: &mut http::HeaderMap) -> grpc::Code {
             headers.insert(GRPC_MESSAGE, msg);
         }
         code
+    } else if error.is::<hyper::error::Error>() {
```
zaharidichev, line 242: That seems super naive... However, most of the methods on hyper::Error are private and I could not figure out a way to determine that this error is `Error(Body, Error { kind: Io(Kind(BrokenPipe)) })`

linkerd/error-respond/src/lib.rs
```diff
@@ -7,17 +7,20 @@ use linkerd2_error::Error;
 
 /// Creates an error responder for a request.
 pub trait NewRespond<Req, E = Error> {
-    type Response;
-    type Respond: Respond<E, Response = Self::Response>;
+    type ResponseIn;
+    type ResponseOut;
```
zaharidichev, line 11: @olix0r Maybe I am missing something here, but do you think there is a way to get rigt of the extra type here and still be able to map the response body only in the case when we are GRPC.

olix0r: I would probably do this as

```rust
pub trait NewRespond<Req, Rsp, E = Error> {
    type Response
    type Respond: Respond<Rsp, E, Response = Self::Response>;
```

The general rule being that the Req/Rsp/E are all passed into the the trait (and not returned by it). We don't want to have to have a concrete in-response type. Implementations should be generic over it.

linkerd/app/core/src/errors.rs
olix0r: I think we can actually use `hyper::Error::is_closed` for this: https://github.com/hyperium/hyper/blob/9a8413d91081ad5a949276f05337e984c455e251/src/error.rs#L114-L129

olix0r: Though, we'd have to check a few other methods besides that one. Perhaps we can use [`Error::into_cause`](https://github.com/hyperium/hyper/blob/9a8413d91081ad5a949276f05337e984c455e251/src/error.rs#L145-L148)?

linkerd/error-respond/src/lib.rs
zaharidichev: Okey, maybe I am way off here, but in essence what is happening here is that we get one response type and map it to another, right. The concrete example is an http:Response<SomePayload> -> http::Response<SomePayloadThatKnowsAboutGrpcErrors>. So don't we need two type params for that? Because at the end in the implementation in errors.rs this all maps to getting an http::Response<HttpBody> and mapping it to http::Response<errors::ResponseBody>.

olix0r: You are correct, except this crate is protocol-agnostic.  So the impl will only be generic over body like:

```rust
impl<ReqB, RspB, E> NewRespond<http::Request<ReqB>, http::Response<RspB>, E> for MyImpl<ReqB, RspB, E>
```

zaharidichev: Got it. Makes sense

linkerd/app/core/src/errors.rs
zaharidichev: Not sure we can. In this particular case we get a kind of `Io`, so right here https://github.com/hyperium/hyper/blob/9a8413d91081ad5a949276f05337e984c455e251/src/error.rs#L317. If we to into_cause, this does not give us much apart from being able to  look at the textual form of this error. Not sure we want to be comparing strings.

olix0r: (As we discussed offline...) It looks like `into_cause` should be able to return the underlying `io::Error`.

zaharidichev: @olix0r So I am a bit confused here. into_cause consumes the error and gives you the optional cause. Do we actually want to use that? Dont we want something that just takes it by ref ()judging by the signature of the `set_grpc_status` func. Alternatively we can use Error::source but that gives me back a `Option<&(dyn StdError + 'static)>` which we cannot pipe recursively into `set_grpc_status` as it is not `Send + Sync` (correct me if I am wrong). Looks like we want to use cause2 which gives back `Option<&(dyn StdError + 'static + Sync + Send)>` but that... is deprecated. Not sure what is the best option here.  Also, what is the reason for `set_grpc_status` to take an `&Error` instead of `Error`. Is it because of the `ServiceError` holding an `Arc` to its inner error ?

olix0r: Y'know... this is all probably overcomplicating it. I think it's fine to just match `hyper::Error` for now. We should leave a comment explaining the situation...

zaharidichev: But still what would be a proper way to introspect that error type according to you ?

linkerd/app/core/src/errors.rs
```diff
@@ -44,11 +45,75 @@ pub enum Respond<B> {
     Http2 { is_grpc: bool },
 }
 
-impl<A, B: Default> respond::NewRespond<http::Request<A>> for NewRespond<B> {
-    type Response = http::Response<B>;
-    type Respond = Respond<B>;
+pub enum ResponseBody<B> {
```
olix0r, line 48: small simplification https://github.com/linkerd/linkerd2-proxy/pull/494

linkerd/app/core/src/errors.rs
```diff
@@ -44,11 +45,75 @@ pub enum Respond<B> {
...
+                assert!(trailers.is_none());
+                match inner.poll_data() {
+                    Err(error) => {
+                        let error = error.into();
+                        let mut error_trailers = http::HeaderMap::new();
+                        let code = set_grpc_status(&error, &mut error_trailers);
+                        warn!(
+                            "Handling body error: {:?} with grpc status {:?}",
+                            error, code
+                        );
```
olix0r, line 77: I think this should probably be at `debug` and not `warn`?  We mostly want to warn to surface proxy errors.

Also, we should use typical tracing metadata instead of formatting like:
```rust
debug!(%error, grpc.status = %code, "Handling gRPC stream failure");

File: linkerd/app/core/src/errors.rs (hunk 1 above)
Comment (olix0r) [reply linkerd/app/core/src/errors.rs:232]: After looking more closely, I believe we can probalby modify `set_grpc_status` to consume the error... We'll have to change some of those `downcast_ref` to `downcast`.

File: linkerd/app/core/src/errors.rs [hunk 5]
```diff
@@ -35,20 +36,81 @@ pub enum Reason {
...
+    fn poll_trailers(&mut self) -> futures::Poll<Option<http::HeaderMap>, Self::Error> {
+        match self {
+            Self::NonGrpc(inner) => inner.poll_trailers(),
+            Self::Grpc { inner, trailers } => match trailers.take() {
+                Some(t) => Ok(Async::Ready(Some(t))),
+                None => inner.poll_trailers(),
+            },
+        }
+    }
+    fn is_end_stream(&self) -> bool {
```
olix0r, line 93: there should be a newline here. i'm surprised rustfmt doesn't complain..

linkerd/app/core/src/errors.rs
zaharidichev: Will give it a shot

linkerd/app/core/src/errors.rs
zaharidichev: yes me as well actually.

linkerd/app/core/src/errors.rs
zaharidichev: The main head scratcher for me is the fact that we are handling a `ServiceError` here, that holds an Arc to its inner error. So we can only take a shared ref of it and cannot consume it, thereby preventing us from calling the method recursively. Can we at that point of the code assume that .`try_unwrap `on the Arc holding the inner error will succeed ?

olix0r: No, that won't work, because the error is shared with the lock/buffer--both of these layers are used to share an inner service, and when its inner service fails, we have to share that error with all outer instances... So try_unwrap won't have exclusive ownership of the error.

As I said, I'm fine with just matching on hyper::Error in this case to get this to master. I'd rather get the fix out and refine it on our own time. I'll take a quick look at the error handler and see if there's a way to simplify the problem, but I'm happy to submit that as a followup. If I can figure it out, that is.

olix0r: Honestly, the best option might be to submit a PR to hyper to add an `Error::cause` like

```rust
    /// Returns a reference to the cause
    pub fn cause(&self) -> Option<&Box<dyn StdError + Send + Sync>> {
        &self.inner.cause
    }
```

We probably won't pick up this change until tokio 0.2 merges, but it would allow us to clean this up later...

PR #494, opened 2020-04-28
Remove unneeded PhantomData

olix0r:
The changes to the `Respond` trait eliminate the need for our
implementation to track the repsonse-body type via `PhantomData`.


linkerd/app/core/src/errors.rs
```diff
@@ -36,12 +36,12 @@ pub enum Reason {
     Unexpected,
 }
 
-#[derive(Debug)]
-pub struct NewRespond<B>(std::marker::PhantomData<fn() -> B>);
+#[derive(Copy, Clone, Debug)]
+pub struct NewRespond(());
```
olix0r, line 40: note: this is done to prevent instantiation `NewRespond` directly from outside of this module. not critical but something we do frequently to encourage going through the constructors.

PR #496, opened 2020-04-30
update http-box to tower 0.3

hawkw:
This does **not** update the payload module to `std::future`, since we
are still on the futures 0.1 version of Hyper. It only updates the
`tower-service` impls.

This will be necessary to bring back the HTTP clients.


hawkw:
ugh why is CI broken?

PR #497, opened 2020-04-30
Adapting makefile for multiarch builds

jaerik:
Make it possible to to build the linkerd2-proxy for other architectures the than the host architecture with make. This by updating the makefile to be aware of environment variables specifying the target architecture.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
-TARGET = target/debug
+CARGO_TARGET ?= $(shell rustup show |sed -n 's/^Default host: \(.*\)/\1/p')
+TARGET = target/$(CARGO_TARGET)/debug
 ifdef CARGO_RELEASE
 	RELEASE = --release
-	TARGET = target/release
+	TARGET = target/$(CARGO_TARGET)/release
 endif
 
 ifndef PACKAGE_VERSION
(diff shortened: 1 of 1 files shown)


olix0r:
Thanks @jaerik! This looks great.

You don't have to necessarily make these changes in this branch: but do you have any thoughts on how we can support these builds in our release process? Our release process is driven by this github action:

https://github.com/linkerd/linkerd2-proxy/blob/c72c5f57da8af60cf89f57c2cc2c9a223cf15bd8/.github/workflows/release.yml

I assume that we need to figure out a way to produce binaries for each of the target platforms. Can we do this easily in Actions? If not, we may be able to provision build hosts for the target platforms... Anyway, this isn't a blocker; but I'd love to figure out a way to support these artifacts as a part of our automated release process. (Until then, it will be hard for us to consider these builds officially supported).

olix0r:
This CI failure is an issue with actions, not this branch.

Makefile
```diff
@@ -112,3 +113,6 @@ docker: Dockerfile Cargo.lock
 
 .PHONY: all
 all: build test
+
+print-%:
```
kleimkuhler, line 117: Is this needed?

jaerik: Sorry, just a forgotten debug rule. Thanks for pointing that out.

jaerik:
> I assume that we need to figure out a way to produce binaries for each of the target platforms. Can we do this easily in Actions? 

@olix0r I think you just can add a new action step for each target platform, e.g:

 - name: package-arm32v7
        env:
          CARGO_RELEASE: "1"
          CARGO_TARGET=armv7-unknown-linux-gnueabihf
          CARGO_TARGET_ARMV7_UNKNOWN_LINUX_GNUEABIHF_LINKER=arm-linux-gnueabihf-gcc
          PACKAGE_VERSION: ${{ steps.release-tag-meta.outputs.name }}
        uses: ./.github/actions/package
        with:
          entrypoint: make
          args: package

And then just change the with-files in the realese step to: 
          files: target/x86_64-unknown-linux-gnu/release/package/* target/arm-linux-gnueabihf/release/package/* target/aarch64-unknown-linux-gnu/release/package/*

I'm not sure how to adapt the checksec step, when there are limitations when testing[ cross compiled systems.](https://github.com/slimm609/checksec.sh#using-with-cross-compiled-systems)

PR #500, opened 2020-05-02
feat: add json log format as an option

naseemkullah:
JSON logs can be easily parsed in logging sinks such as Elasticsearch. 

resolves https://github.com/linkerd/linkerd2/issues/2491

cc @hawkw 

Please bare with me as this is my first PR in Rust, unless you count a message change. :)

Could you please break down the anatomy of the following line:

https://github.com/linkerd/linkerd2-proxy/blob/610309ebc240494426f9169000c46da9d5fe9647/linkerd/app/core/src/trace.rs#L12

### DIFF
--- a/linkerd/app/core/src/trace.rs
+++ b/linkerd/app/core/src/trace.rs
@@ -47,12 +47,11 @@ pub fn with_filter_and_format(
     format: impl AsRef<str>,
 ) -> (Dispatch, LevelHandle) {
     let filter = filter.as_ref();
-    let format = format.as_ref();
 
     // Set up the subscriber
     let start_time = clock::now();
 
-    match format {
+    match format.as_ref().to_uppercase().as_ref() {
         "JSON" => {
             let builder = FmtSubscriber::builder()
                 .json()
--- a/linkerd/app/core/src/trace.rs
+++ b/linkerd/app/core/src/trace.rs
@@ -10,6 +10,9 @@ use tracing_subscriber::{
 const ENV_LOG_LEVEL: &str = "LINKERD2_PROXY_LOG";
 const ENV_LOG_FORMAT: &str = "LINKERD2_PROXY_LOG_FORMAT";
 
+const DEFAULT_LOG_LEVEL: &str = "warn,linkerd2_proxy=info";
+const DEFAULT_LOG_FORMAT: &str = "PLAIN";
+
 type JsonFormatter = Formatter<format::JsonFields, format::Format<format::Json, Uptime>>;
 type PlainFormatter = Formatter<format::DefaultFields, format::Format<format::Full, Uptime>>;
 
@@ -26,8 +29,8 @@ pub enum LevelHandle {
 /// Initialize tracing and logging with the value of the `ENV_LOG`
 /// environment variable as the verbosity-level filter.
 pub fn init() -> Result<LevelHandle, Error> {
-    let log_level = env::var(ENV_LOG_LEVEL).unwrap_or_default();
-    let log_format = env::var(ENV_LOG_FORMAT).unwrap_or_default();
+    let log_level = env::var(ENV_LOG_LEVEL).unwrap_or(DEFAULT_LOG_LEVEL.to_string());
+    let log_format = env::var(ENV_LOG_FORMAT).unwrap_or(DEFAULT_LOG_FORMAT.to_string());
 
     let (dispatch, handle) = with_filter_and_format(log_level, log_format);
 
--- a/linkerd/app/core/src/trace.rs
+++ b/linkerd/app/core/src/trace.rs
@@ -18,12 +18,8 @@ type PlainFormatter = Formatter<format::DefaultFields, format::Format<format::Fu
 
 #[derive(Clone)]
 pub enum LevelHandle {
-    Json {
-        inner: reload::Handle<EnvFilter, JsonFormatter>,
-    },
-    Plain {
-        inner: reload::Handle<EnvFilter, PlainFormatter>,
-    },
+    Json(reload::Handle<EnvFilter, JsonFormatter>),
+    Plain(reload::Handle<EnvFilter, PlainFormatter>),
 }
 
 /// Initialize tracing and logging with the value of the `ENV_LOG`
--- a/linkerd/app/core/src/trace.rs
+++ b/linkerd/app/core/src/trace.rs
@@ -50,7 +50,6 @@ pub fn with_filter_and_format(
     format: impl AsRef<str>,
 ) -> (Dispatch, LevelHandle) {
     let filter = filter.as_ref();
-    let format = format.as_ref();
 
     // Set up the subscriber
     let start_time = clock::now();
--- a/linkerd/app/core/src/trace.rs
+++ b/linkerd/app/core/src/trace.rs
@@ -58,8 +58,7 @@ pub fn with_filter_and_format(
                 .json()
                 .with_timer(Uptime { start_time })
                 .with_env_filter(filter)
-                .with_filter_reloading()
-                .with_ansi(cfg!(test));
+                .with_filter_reloading();
 
             let handle = LevelHandle::Json {
                 inner: builder.reload_handle(),
@@ -71,10 +70,10 @@ pub fn with_filter_and_format(
         }
         "PLAIN" | _ => {
             let builder = FmtSubscriber::builder()
+                .with_ansi(cfg!(test))
                 .with_timer(Uptime { start_time })
                 .with_env_filter(filter)
-                .with_filter_reloading()
-                .with_ansi(cfg!(test));
+                .with_filter_reloading();
 
             let handle = LevelHandle::Plain {
                 inner: builder.reload_handle(),
--- a/linkerd/app/core/src/trace.rs
+++ b/linkerd/app/core/src/trace.rs
@@ -51,35 +51,25 @@ pub fn with_filter_and_format(
     // Set up the subscriber
     let start_time = clock::now();
 
+    let builder = FmtSubscriber::builder()
+        .with_timer(Uptime { start_time })
+        .with_env_filter(filter);
+
     match format.as_ref().to_uppercase().as_ref() {
         "JSON" => {
-            let builder = FmtSubscriber::builder()
-                .json()
-                .with_timer(Uptime { start_time })
-                .with_env_filter(filter)
-                .with_filter_reloading();
-
+            let builder = builder.json().with_filter_reloading();
             let handle = LevelHandle::Json {
                 inner: builder.reload_handle(),
             };
-
             let dispatch = Dispatch::new(builder.finish());
-
             (dispatch, handle)
         }
         "PLAIN" | _ => {
-            let builder = FmtSubscriber::builder()
-                .with_ansi(cfg!(test))
-                .with_timer(Uptime { start_time })
-                .with_env_filter(filter)
-                .with_filter_reloading();
-
+            let builder = builder.with_ansi(cfg!(test)).with_filter_reloading();
             let handle = LevelHandle::Plain {
                 inner: builder.reload_handle(),
             };
-
             let dispatch = Dispatch::new(builder.finish());
-
             (dispatch, handle)
         }
     }
--- a/linkerd/app/core/src/trace.rs
+++ b/linkerd/app/core/src/trace.rs
@@ -50,6 +50,7 @@ pub fn with_filter_and_format(
     format: impl AsRef<str>,
 ) -> (Dispatch, LevelHandle) {
     let filter = filter.as_ref();
+    let format = format.as_ref();
 
     // Set up the subscriber
     let start_time = clock::now();
--- a/linkerd/app/core/src/trace.rs
+++ b/linkerd/app/core/src/trace.rs
@@ -57,17 +57,13 @@ pub fn with_filter_and_format(
     match format.as_ref().to_uppercase().as_ref() {
         "JSON" => {
             let builder = builder.json().with_filter_reloading();
-            let handle = LevelHandle::Json {
-                inner: builder.reload_handle(),
-            };
+            let handle = LevelHandle::Json(builder.reload_handle());
             let dispatch = Dispatch::new(builder.finish());
             (dispatch, handle)
         }
         "PLAIN" | _ => {
             let builder = builder.with_ansi(cfg!(test)).with_filter_reloading();
-            let handle = LevelHandle::Plain {
-                inner: builder.reload_handle(),
-            };
+            let handle = LevelHandle::Plain(builder.reload_handle());
             let dispatch = Dispatch::new(builder.finish());
             (dispatch, handle)
         }
@@ -99,8 +95,8 @@ impl LevelHandle {
         let level = level.as_ref();
         let filter = level.parse::<EnvFilter>()?;
         match self {
-            Self::Json { inner } => inner.reload(filter)?,
-            Self::Plain { inner } => inner.reload(filter)?,
+            Self::Json(level) => level.reload(filter)?,
+            Self::Plain(level) => level.reload(filter)?,
         }
         tracing::info!(%level, "set new log level");
         Ok(())
@@ -108,8 +104,12 @@ impl LevelHandle {
 
     pub fn current(&self) -> Result<String, Error> {
         match self {
-            Self::Json { inner } => inner.with_current(|f| format!("{}", f)).map_err(Into::into),
-            Self::Plain { inner } => inner.with_current(|f| format!("{}", f)).map_err(Into::into),
+            Self::Json(with_current) => with_current
+                .with_current(|f| format!("{}", f))
+                .map_err(Into::into),
+            Self::Plain(with_current) => with_current
+                .with_current(|f| format!("{}", f))
+                .map_err(Into::into),
         }
     }
 }
@@ -117,7 +117,7 @@ impl LevelHandle {
 impl fmt::Debug for LevelHandle {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Self::Json { inner } => inner
+            Self::Json(with_current) => with_current
                 .with_current(|c| {
                     f.debug_struct("LevelHandle")
                         .field("current", &format_args!("{}", c))
@@ -128,7 +128,7 @@ impl fmt::Debug for LevelHandle {
                         .field("current", &format_args!("{}", e))
                         .finish()
                 }),
-            Self::Plain { inner } => inner
+            Self::Plain(with_current) => with_current
                 .with_current(|c| {
                     f.debug_struct("LevelHandle")
                         .field("current", &format_args!("{}", c))
--- a/linkerd/app/core/src/trace.rs
+++ b/linkerd/app/core/src/trace.rs
@@ -104,10 +104,10 @@ impl LevelHandle {
 
     pub fn current(&self) -> Result<String, Error> {
         match self {
-            Self::Json(with_current) => with_current
+            Self::Json(handle) => handle
                 .with_current(|f| format!("{}", f))
                 .map_err(Into::into),
-            Self::Plain(with_current) => with_current
+            Self::Plain(handle) => handle
                 .with_current(|f| format!("{}", f))
                 .map_err(Into::into),
         }
@@ -117,7 +117,7 @@ impl LevelHandle {
 impl fmt::Debug for LevelHandle {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Self::Json(with_current) => with_current
+            Self::Json(handle) => handle
                 .with_current(|c| {
                     f.debug_struct("LevelHandle")
                         .field("current", &format_args!("{}", c))
@@ -128,7 +128,7 @@ impl fmt::Debug for LevelHandle {
                         .field("current", &format_args!("{}", e))
                         .finish()
                 }),
-            Self::Plain(with_current) => with_current
+            Self::Plain(handle) => handle
                 .with_current(|c| {
                     f.debug_struct("LevelHandle")
                         .field("current", &format_args!("{}", c))
--- a/linkerd/app/core/src/trace.rs
+++ b/linkerd/app/core/src/trace.rs
@@ -10,19 +10,16 @@ use tracing_subscriber::{
 const ENV_LOG_LEVEL: &str = "LINKERD2_PROXY_LOG";
 const ENV_LOG_FORMAT: &str = "LINKERD2_PROXY_LOG_FORMAT";
 
+type JsonFormatter = Formatter<format::JsonFields, format::Format<format::Json, Uptime>>;
+type PlainFormatter = Formatter<format::DefaultFields, format::Format<format::Full, Uptime>>;
+
 #[derive(Clone)]
 pub enum LevelHandle {
     Json {
-        inner: reload::Handle<
-            EnvFilter,
-            Formatter<format::JsonFields, format::Format<format::Json, Uptime>>,
-        >,
+        inner: reload::Handle<EnvFilter, JsonFormatter>,
     },
     Plain {
-        inner: reload::Handle<
-            EnvFilter,
-            Formatter<format::DefaultFields, format::Format<format::Full, Uptime>>,
-        >,
+        inner: reload::Handle<EnvFilter, PlainFormatter>,
     },
 }
 
--- a/linkerd/app/core/src/trace.rs
+++ b/linkerd/app/core/src/trace.rs
@@ -7,20 +7,32 @@ use tracing_subscriber::{
     reload, EnvFilter, FmtSubscriber,
 };
 
-const ENV_LOG: &str = "LINKERD2_PROXY_LOG";
-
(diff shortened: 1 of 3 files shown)


hawkw:
> Could you please break down the anatomy of the following line:

@naseemkullah this is a _type alias_ --- we are assigning a shorter name to a long, complicated type. Here's the [section on type aliases](https://doc.rust-lang.org/stable/book/ch19-04-advanced-types.html#creating-type-synonyms-with-type-aliases) in the Rust book.

In this case, the `Formatter` type consists of a formatter for fields (the key-value pairs included on spans and log messages), and a formatter for events that formats each logged event using that field formatter. Here, the event formatter is configured to use the "full" output format (including the fields of all spans in the context where each event occurred) and formats timestamps based on the time since the process started, rather than as date-times (the `Uptime` formatter).

This code will probably have to be changed up a bit if we want to allow conditionally using the `Json` field and event formatters. You may find [this example](https://github.com/tokio-rs/tracing/blob/5a971befacc233612ec6577c9bae6188c79926ad/examples/examples/futures-proxy-server.rs#L142-L155) useful as a starting point.

naseemkullah:
Would it make sense to declare these two type (aliases) up top then?

```rust
type Subscriber = Formatter<format::DefaultFields, format::Format<format::Full, Uptime>>;
type JsonSubscriber = Formatter<format::JsonFields, format::Format<format::Json, Uptime>>;
```

naseemkullah:
Would it make sense to define an enum for Subscriber instead of a type alias then? 

i.e.:

```rust
enum Susbcriber {
  Plain(Formatter<format::DefaultFields, format::Format<format::Full, Uptime>>),
  Json(Formatter<format::JsonFields, format::Format<format::Json, Uptime>>)
}
```

naseemkullah:
Current status @hawkw :

• Replaced the aforementioned type alias with en enum consisting of Json and Plain Subscriber types.

• Now trying to decipher and deal with the following errors introduced by the change regarding *mismatched types* and *unsatisfied trait bounds* (tips, links, references are all most welcome):

```rust
pub fn reload_handle(&self) -> crate::reload::Handle<crate::EnvFilter, Formatter<N, E, W>>
Returns a Handle that may be used to reload the constructed subscriber's filter.

mismatched types

expected enum `trace::Subscriber`, found struct `tracing_subscriber::layer::Layered`

note: expected struct `tracing_subscriber::reload::Handle<_, trace::Subscriber>`
         found struct `tracing_subscriber::reload::Handle<_, tracing_subscriber::layer::Layered<tracing_subscriber::fmt::fmt_layer::Layer<tracing_subscriber::registry::sharded::Registry, tracing_subscriber::fmt::format::DefaultFields, tracing_subscriber::fmt::format::Format<tracing_subscriber::fmt::format::Full, trace::Uptime>>, tracing_subscriber::registry::sharded::Registry>>`rustc(E0308)
trace.rs(53, 16): expected enum `trace::Subscriber`, found struct `tracing_subscriber::layer::Layered`
```

```rust
pub fn reload(&self, new_layer: impl Into<L>) -> Result<(), Error>
Replace the current layer with the provided new_layer.

no method named `reload` found for struct `tracing_subscriber::reload::Handle<tracing_subscriber::filter::env::EnvFilter, trace::Subscriber>` in the current scope

method not found in `tracing_subscriber::reload::Handle<tracing_subscriber::filter::env::EnvFilter, trace::Subscriber>`

note: the method `reload` exists but the following trait bounds were not satisfied:
      `trace::Subscriber : tracing_core::subscriber::Subscriber`
      `tracing_subscriber::filter::env::EnvFilter : tracing_subscriber::layer::Layer<trace::Subscriber>`rustc(E0599)
trace.rs(84, 20): method not found in `tracing_subscriber::reload::Handle<tracing_subscriber::filter::env::EnvFilter, trace::Subscriber>`
```

```rust
pub fn with_current<T>(&self, f: impl FnOnce(&L) -> T) -> Result<T, Error>
Invokes a closure with a borrowed reference to the current layer, returning the result (or an error if the subscriber no longer exists).

no method named `with_current` found for struct `tracing_subscriber::reload::Handle<tracing_subscriber::filter::env::EnvFilter, trace::Subscriber>` in the current scope

method not found in `tracing_subscriber::reload::Handle<tracing_subscriber::filter::env::EnvFilter, trace::Subscriber>`

note: the method `with_current` exists but the following trait bounds were not satisfied:
      `trace::Subscriber : tracing_core::subscriber::Subscriber`
      `tracing_subscriber::filter::env::EnvFilter : tracing_subscriber::layer::Layer<trace::Subscriber>`rustc(E0599)
trace.rs(91, 14): method not found in `tracing_subscriber::reload::Handle<tracing_subscriber::filter::env::EnvFilter, trace::Subscriber>`
```

hawkw:
Hi Naseem, sorry it took me a bit to reply to your comments, I've been quite busy.

The reason your approach of adding an enum isn't working is because the enum you've created doesn't inherit the trait implementations of its two variants. If you wanted to implement the feature this way, you would need to write an implementation of the [`Subscriber` trait](https://docs.rs/tracing/0.1.14/tracing/subscriber/trait.Subscriber.html) that forwards to each variant, like
```rust
impl tracing::Subscriber for Subscriber {
    fn new_span(&self, span: &tracing::span::Attributes<'_>) -> tracing::span::Id {
        match self {
            Subscriber::Json(s) => s.new_span(span),
            Subscriber::Plain(s) => s.new_span(span),
        }
    }

    // ... and do the same thing for all the other trait methods on
    // the `Subscriber` trait ...
}
```

However, I don't know if this is the best approach. Adding the enum introduces a layer of indirection to every subscriber method call, which is not ideal, and it requires writing a bunch of additional code to forward the `Subscriber` implementation. Instead, since the function ultimately returns a `Dispatch`, which erases the type of an underlying `Subscriber`, I would just have two separate branches that return either a JSON or plaintext subscriber, wrapped in a `Dispatch`.

The other error you're running into is due to the [`reload::Handle` type](https://docs.rs/tracing-subscriber/0.2.5/tracing_subscriber/reload/struct.Handle.html) being parameterized over a generic `Subscriber` variable. Since the subscriber may now be either a JSON subscriber _or_ a plaintext subscriber, which are different types, we cannot write out the type of the `reload::Handle` here: 
https://github.com/linkerd/linkerd2-proxy/blob/1d1eabdc213edae756d7248845edff7691b5b037/linkerd/app/core/src/trace.rs#L20

Here, I think changing the `LevelHandle` type to be an enum _does_ make sense. Since reloading the filter is not in the hot path, it's okay to introduce the overhead of having to match on two enum variants, and there are only two places we would need to match on the enum:
https://github.com/linkerd/linkerd2-proxy/blob/1d1eabdc213edae756d7248845edff7691b5b037/linkerd/app/core/src/trace.rs#L81-L93

I think the `Subscriber` type alias will probably now be unnecessary and should just be removed.

Does that all make sense?

naseemkullah:
> Hi Naseem, sorry it took me a bit to reply to your comments, I've been quite busy.
> 

Hi @hawkw ,not at all, I'm sorry for being somewhat of a burden as I am a little out of my depth!

> Does that all make sense? Let me know if you have any additional questions!

Yes, thanks for thoroughly explaining all that! Stay tuned for my attempt at implementing this (or another question 😅 )!

naseemkullah:
Hi @hawkw, there is probably room for stylistic improvement, refactoring and deduplicating so please let me know. But I think this is ready for a first review. 

- Could you please advise as to what would be the most straightforward way to test this locally? I guess I build and run the proxy standalone locally?

- Where would the fallback/default value of `PLAIN` go for `LINKERD2_LOG_FORMAT`?

hawkw:
> Hi @hawkw, there is probably room for stylistic improvement, refactoring and deduplicating so please let me know. But I think this is ready for a first review.
> 
>   * Could you please advise as to what would be the most straightforward way to test this locally? I guess I build and run the proxy standalone locally?

The proxy integration tests use the same tracing configuration as the proxy binary: https://github.com/linkerd/linkerd2-proxy/blob/cb161a13d504b963f10ed386d08352b17b4fe501/linkerd/app/integration/src/lib.rs#L40-L51
(note the call to `app::core::trace::with_filter`)

So, you should be able to run the integration tests with `LINKERD2_PROXY_LOG=debug,LINKERD2_PROXY_LOG_FORMAT=json` for a quick spot check. You can also run a proxy locally with `cargo run`, but you'll need to set all the required environment variables, so running the integration tests is probably quicker and easier.

>   *  Where would the fallback/default value of `PLAIN` go for `LINKERD2_LOG_FORMAT`?

If you look in [`linkerd/app/src/env.rs`](https://github.com/linkerd/linkerd2-proxy/blob/master/linkerd/app/src/env.rs), you'll see that that module is where we define constants for all the environment variables that configure the proxy, and their default values.

>   * Would `--proxy-log-format` CLI flag and `config.linkerd.io/proxy-log-format` annotation be implemented in the linkerd2 repo?

Yup, that's correct. I believe it would be necessary to add it to the proxy-injector control plane component, and to the CLI. Both of those live in the `linkerd2` repo.

I'll take a closer look at the code changes here soon, but it's great that everything is working!

linkerd/app/core/src/trace.rs
```diff
@@ -7,20 +7,32 @@ use tracing_subscriber::{
     reload, EnvFilter, FmtSubscriber,
 };
 
-const ENV_LOG: &str = "LINKERD2_PROXY_LOG";
-
-type Subscriber = Formatter<format::DefaultFields, format::Format<format::Full, Uptime>>;
+const ENV_LOG_LEVEL: &str = "LINKERD2_PROXY_LOG";
+const ENV_LOG_FORMAT: &str = "LINKERD2_PROXY_LOG_FORMAT";
```
hawkw, line 11: These should probably go in [`env.rs`](https://github.com/linkerd/linkerd2-proxy/blob/master/linkerd/app/src/env.rs), where we define all the other environment variables.

linkerd/app/core/src/trace.rs
```diff
@@ -7,20 +7,32 @@ use tracing_subscriber::{
...
 
 #[derive(Clone)]
-pub struct LevelHandle {
-    inner: reload::Handle<EnvFilter, Subscriber>,
+pub enum LevelHandle {
+    Json {
+        inner: reload::Handle<
+            EnvFilter,
+            Formatter<format::JsonFields, format::Format<format::Json, Uptime>>,
```
hawkw, line 18: Take it or leave it: I _might_ consider adding `JsonFormatter` and `PlainFormatter` type aliases for these, because this is a bit wordy; not a blocker though.

linkerd/app/core/src/trace.rs
```diff
@@ -33,25 +45,52 @@ pub fn init_log_compat() -> Result<(), Error> {
...
-        .with_ansi(cfg!(test));
-    let handle = LevelHandle {
-        inner: builder.reload_handle(),
-    };
-    let dispatch = Dispatch::new(builder.finish());
-
-    (dispatch, handle)
+
+    match format {
+        "JSON" => {
```
hawkw, line 59: I think we probably want these to be case-insensitive — `"json"` should probably not get the plain format.

linkerd/app/core/src/trace.rs
```diff
@@ -33,25 +45,52 @@ pub fn init_log_compat() -> Result<(), Error> {
...
+            let builder = FmtSubscriber::builder()
+                .json()
```
hawkw, line 61: I think that if we call the `json` method on the builder _last_, we can factor out most of the shared configuration:

```rust
let builder = FmtSubscriber::builder()
    .with_timer(Uptime { start_time })
    .with_filter(filter)
    .with_filter_reloading();

match format {
    "JSON" => {
        let builder = builder.json();
        // ...
    }
    // ...
}
```

linkerd/app/core/src/trace.rs
```diff
@@ -33,25 +45,52 @@ pub fn init_log_compat() -> Result<(), Error> {
...
+                .with_timer(Uptime { start_time })
+                .with_env_filter(filter)
+                .with_filter_reloading()
+                .with_ansi(cfg!(test));
```
hawkw, line 65: IIRC the `.with_ansi` option doesn't actually do anything with the JSON formatter --- it doesn't support ANSI colors.

linkerd/app/core/src/trace.rs
naseemkullah: 💯 This would increase readability. By one confusion is wether these are Formatters or Subscribers, originally when there was only the Plain version it was type aliased as Subscriber. Going with the term Formatter now but please advise.

linkerd/app/core/src/trace.rs
naseemkullah: For whatever reason, we need to put json();  either first, or at least not last, if not the following error occurs:

```rust
pub fn reload_handle(&self) -> crate::reload::Handle<crate::EnvFilter, Formatter<N, E, W>>
Returns a Handle that may be used to reload the constructed subscriber's filter.

no method named `reload_handle` found for struct `tracing_subscriber::fmt::SubscriberBuilder<tracing_subscriber::fmt::format::json::JsonFields, tracing_subscriber::fmt::format::Format<tracing_subscriber::fmt::format::json::Json, trace::Uptime>, tracing_subscriber::reload::Layer<tracing_subscriber::filter::env::EnvFilter, tracing_subscriber::layer::Layered<tracing_subscriber::fmt::fmt_layer::Layer<tracing_subscriber::registry::sharded::Registry, tracing_subscriber::fmt::format::DefaultFields, tracing_subscriber::fmt::format::Format<tracing_subscriber::fmt::format::Full, trace::Uptime>>, tracing_subscriber::registry::sharded::Registry>>>` in the current scope

method not found in `tracing_subscriber::fmt::SubscriberBuilder<tracing_subscriber::fmt::format::json::JsonFields, tracing_subscriber::fmt::format::Format<tracing_subscriber::fmt::format::json::Json, trace::Uptime>, tracing_subscriber::reload::Layer<tracing_subscriber::filter::env::EnvFilter, tracing_subscriber::layer::Layered<tracing_subscriber::fmt::fmt_layer::Layer<tracing_subscriber::registry::sharded::Registry, tracing_subscriber::fmt::format::DefaultFields, tracing_subscriber::fmt::format::Format<tracing_subscriber::fmt::format::Full, trace::Uptime>>, tracing_subscriber::registry::sharded::Registry>>>`rustc(E0599)
trace.rs(64, 32): method not found in `tracing_subscriber::fmt::SubscriberBuilder<tracing_subscriber::fmt::format::json::JsonFields, tracing_subscriber::fmt::format::Format<tracing_subscriber::fmt::format::json::Json, trace::Uptime>, tracing_subscriber::reload::Layer<tracing_subscriber::filter::env::EnvFilter, tracing_subscriber::layer::Layered<tracing_subscriber::fmt::fmt_layer::Layer<tracing_subscriber::registry::sharded::Registry, tracing_subscriber::fmt::format::DefaultFields, tracing_subscriber::fmt::format::Format<tracing_subscriber::fmt::format::Full, trace::Uptime>>, tracing_subscriber::registry::sharded::Registry>>>`
Peek Problem (⌥F8)
No quick fixes available
```

linkerd/app/core/src/trace.rs
naseemkullah: Thanks!

linkerd/app/core/src/trace.rs
naseemkullah: Thanks I did something that feels hacky `match format.as_ref().to_uppercase().as_ref() {` ... is there a prettier way?

naseemkullah: https://github.com/linkerd/linkerd2-proxy/pull/500/commits/59f85e968e72de4ac11725419143e4072f781554

linkerd/app/core/src/trace.rs
hawkw: Oh, I think `with_filter_reloading` needs to be called after `json`, because `json` changes the type of the formatter. So, you could still factor out the `with_timer` and `with_filter` calls, if you like (but it's not a big deal regardless).

naseemkullah: Perfect, thanks.

naseemkullah:
> > Hi @hawkw, there is probably room for stylistic improvement, refactoring and deduplicating so please let me know. But I think this is ready for a first review.
> > 
> > * Could you please advise as to what would be the most straightforward way to test this locally? I guess I build and run the proxy standalone locally?
> 
> The proxy integration tests use the same tracing configuration as the proxy binary:
> 
> https://github.com/linkerd/linkerd2-proxy/blob/cb161a13d504b963f10ed386d08352b17b4fe501/linkerd/app/integration/src/lib.rs#L40-L51
> 
> 
> (note the call to `app::core::trace::with_filter`)
> So, you should be able to run the integration tests with `LINKERD2_PROXY_LOG=debug,LINKERD2_PROXY_LOG_FORMAT=json` for a quick spot check. You can also run a proxy locally with `cargo run`, but you'll need to set all the required environment variables, so running the integration tests is probably quicker and easier.
> 
> > * Where would the fallback/default value of `PLAIN` go for `LINKERD2_LOG_FORMAT`?
> 
> If you look in [`linkerd/app/src/env.rs`](https://github.com/linkerd/linkerd2-proxy/blob/master/linkerd/app/src/env.rs), you'll see that that module is where we define constants for all the environment variables that configure the proxy, and their default values.
> 
> > * Would `--proxy-log-format` CLI flag and `config.linkerd.io/proxy-log-format` annotation be implemented in the linkerd2 repo?
> 
> Yup, that's correct. I believe it would be necessary to add it to the proxy-injector control plane component, and to the CLI. Both of those live in the `linkerd2` repo.
> 
> I'll take a closer look at the code changes here soon, but it's great that everything is working!

Seems like locally the following integration tests fail with or without the added env vars:

```
test transport::inbound_http_accept ... FAILED
...
test transport::inbound_tcp_accept ... FAILED
...
test transport::outbound_tcp_accept ... FAILED
test transport::outbound_tcp_open_connections ... FAILED
test transport::outbound_http_accept ... FAILED
```

```
did not contain:
tcp_close_total{direction="inbound",peer="src",tls="disabled",errno="ENOTCONN"} 1', linkerd/app/integration/tests/telemetry.rs:762:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
server Listening dropped; addr=127.0.0.1:64482
```

```
did not contain:
tcp_close_total{direction="inbound",peer="src",tls="disabled",errno=""} 1', linkerd/app/integration/tests/telemetry.rs:968:9
server Listening dropped; addr=127.0.0.1:64509
```

```
did not contain:
tcp_close_total{direction="outbound",peer="src",tls="no_identity",no_tls_reason="loopback",errno=""} 1', linkerd/app/integration/tests/telemetry.rs:1123:9
server Listening dropped; addr=127.0.0.1:64571
```

```
did not contain:
tcp_open_connections{direction="outbound",peer="src",tls="no_identity",no_tls_reason="loopback"} 1', linkerd/app/integration/tests/telemetry.rs:1249:9
server Listening dropped; addr=127.0.0.1:64573
```

```
did not contain:
tcp_close_total{direction="outbound",peer="src",tls="no_identity",no_tls_reason="loopback",errno="ENOTCONN"} 1', linkerd/app/integration/tests/telemetry.rs:829:9
server Listening dropped; addr=127.0.0.1:64541
```

linkerd/app/core/src/trace.rs
naseemkullah: I'm having particularly a lot of trouble with this.

After moving these and deifining their `DEFAULT_*` counterparts in env.rs, I started going down a rabbit hole of trying to do something similar to: https://github.com/linkerd/linkerd2-proxy/blob/a7f22f6caae7598855e8ae04d90e8668be6ca61b/linkerd/app/core/src/dns.rs#L6-L26 but it just seems like way too much plumbing.

How can I more simply just access the consts directly from env.rs? I tried importing app to try to get them via `app::env::ENV_LOG_LEVEL` for example but that resulted in a circular dependency error.

naseemkullah: In the meantime I've added the default values here in `trace.rs`  https://github.com/linkerd/linkerd2-proxy/pull/500/commits/337b7e63cc558f3ee8e3861a1e30a252846a2219

naseemkullah:
> Seems like locally the following integration tests fail with or without the added env vars:

These same tests fail when I try `make test-integration` against the master branch 🤷‍♂️

naseemkullah:
> Overall, this looks good to me. I had a few minor suggestions, but nothing blocking.
> 
> We'll need a second approval from someone else in the @linkerd/proxy-maintainers team before we can merge this.
> 
> @naseemkullah, thanks again for working on this, and for your patience waiting for a review!

Thanks so much for your patience and guidance @hawkw , I may have broke it trying to implement code review: https://github.com/linkerd/linkerd2-proxy/pull/500/commits/fd21ce66b9874b3958b01d17288158097564ca7b could please take a look 😅  I did weird things to get it to compile.

linkerd/app/core/src/trace.rs
```diff
@@ -7,20 +7,32 @@ use tracing_subscriber::{
...
 #[derive(Clone)]
-pub struct LevelHandle {
-    inner: reload::Handle<EnvFilter, Subscriber>,
+pub enum LevelHandle {
+    Json {
+        inner: reload::Handle<EnvFilter, JsonFormatter>,
+    },
+    Plain {
+        inner: reload::Handle<EnvFilter, PlainFormatter>,
+    },
```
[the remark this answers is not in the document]
olix0r: @hawkw Do you have any thoughts on this approach over, say, making the enum just the formatter?

i.e.

```rust
enum Formatter {
  Json(JsonFormatter),
  Default(DefaultFormatter),
}
```

and then using `reload::Handle<EnvFilter, Formatter>` throughout?

Is there a reason that we have to vary the whole handle? I don't think we should support changing formats at runtime, for instance...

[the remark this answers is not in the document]
hawkw: @olix0r if the formatter was an enum, we would need an implementation of the [`FormatEvent`](https://docs.rs/tracing-subscriber/0.2.5/tracing_subscriber/fmt/trait.FormatEvent.html) trait that matches the enum and calls the `FormatEvent` impl for each variant. This enum matching would happen every time we log an event (a relatively hot path). By making the reload handle an enum instead, we only need to match on variants when reloading the level filter (which happens very infrequently). 

> Is there a reason that we have to vary the whole handle? I don't think we should support changing formats at runtime, for instance...

The current implementation doesn't allow that --- the reload layer just needs to be parameterized over the inner subscriber type for type system reasons. If we wanted to change the formatter at runtime, it would need a separate reload layer as well.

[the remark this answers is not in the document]
olix0r: makes sense, thanks.

linkerd/app/core/src/trace.rs
```diff
@@ -68,37 +87,58 @@ impl LevelHandle {
...
 
     pub fn current(&self) -> Result<String, Error> {
-        self.inner
-            .with_current(|f| format!("{}", f))
-            .map_err(Into::into)
+        match self {
+            Self::Json(with_current) => with_current
+                .with_current(|f| format!("{}", f))
+                .map_err(Into::into),
+            Self::Plain(with_current) => with_current
```
hawkw, line 110: I would probably have called the inner value `handle` or something, personally
```suggestion
            Self::Json(handle) => handle
                .with_current(|f| format!("{}", f))
                .map_err(Into::into),
            Self::Plain(handle) => handle
```

linkerd/app/core/src/trace.rs
```diff
@@ -68,37 +87,58 @@ impl LevelHandle {
...
+                    f.debug_struct("LevelHandle")
+                        .field("current", &format_args!("{}", c))
+                        .finish()
+                })
+                .unwrap_or_else(|e| {
+                    f.debug_struct("LevelHandle")
+                        .field("current", &format_args!("{}", e))
+                        .finish()
+                }),
+            Self::Plain(with_current) => with_current
```
hawkw, line 131: Similarly, I would probably have said:
```suggestion
            Self::Json(handle) => handle
                .with_current(|c| {
                    f.debug_struct("LevelHandle")
                        .field("current", &format_args!("{}", c))
                        .finish()
                })
                .unwrap_or_else(|e| {
                    f.debug_struct("LevelHandle")
                        .field("current", &format_args!("{}", e))
                        .finish()
                }),
            Self::Plain(handle) => handle
```

linkerd/app/core/src/trace.rs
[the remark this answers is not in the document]
hawkw: In an ideal world, the handle wouldn't need to be parameterized over the inner subscriber, just the type of the layer that's actually being reloaded. IIRC, there was some reason this didn't work when this code was originally written. I'd like to try and fix it in `tracing-subscriber` 0.3...

naseemkullah:
> > Thanks so much for your patience and guidance @hawkw , I may have broke it trying to implement code review: [fd21ce6](https://github.com/linkerd/linkerd2-proxy/commit/fd21ce66b9874b3958b01d17288158097564ca7b) could please take a look sweat_smile I did weird things to get it to compile.
> 
> I don't see anything weird here --- it looks like you're using the tuple-like enum correctly :) This still looks good to me!

Alright thanks, really was not sure (it looks a lot better with the renamed inner values to `handle`)... wow rust really is fool proof!

PR #505, opened 2020-05-05
buffer: Port idleness changes to tokio-0.2

olix0r:
Applies the changes in #502 to the tokio-0.2 branch.


linkerd/app/inbound/src/lib.rs
```diff
@@ -202,14 +191,14 @@ impl Config {
...
-            //                 // If the service has been unavailable for an extend time, eagerly
+            //                 // If the service has been unavailable for an extended time, eagerly
             //                 // fail requests.
             //                 .push_failfast(dispatch_timeout)
             //                 // Shares the service, ensuring discovery errors are propagated.
-            //                 .push_spawn_buffer(buffer_capacity)
+            //                 .push_spawn_buffer_with_idle_timeout(
+            //                     buffer_capacity,
+            //                     cache_max_idle_age,
+            //                 )
```
hawkw, line 201: Thanks for copying over the changes to commented out code as well, that will make my life much easier later :)

linkerd/buffer/src/dispatch.rs
```diff
@@ -84,63 +93,213 @@ where
...
+
+        let (tx, rx) = mpsc::channel(1);
+        let (ready_tx, ready_rx) = watch::channel(Poll::Pending);
+        let (inner, mut handle) = mock::pair::<(), ()>();
+        let mut dispatch = task::spawn(super::Dispatch::new(inner, rx, ready_tx, Some(max_idle)));
+        handle.allow(1);
+
+        // Service ready without requests. Idle counter starts ticking.
+        assert!(dispatch.poll().is_pending());
+        assert!(ready_rx.borrow().is_ready());
```
hawkw, line 207: nit/TIOLI: `tokio_test` also has `assert_ready!` and `assert_pending!` macros...the `assert_ready!` macro has the useful property of also unwrapping the value inside the `Poll::Ready` so you can make additional assertions about it. may or may not be necessary here but they're kinda nice.

linkerd/buffer/src/error.rs
```diff
@@ -36,3 +40,19 @@ impl std::error::Error for ServiceError {
...
+// === impl IdleError ===
+
+impl<'a> std::fmt::Display for IdleError {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let secs = self.0.as_secs();
+        let subsec_ms = self.0.subsec_nanos() as f64 / 1_000_000f64;
+        if secs == 0 {
+            write!(fmt, "Service idled out after {}ms", subsec_ms)
+        } else {
+            write!(fmt, "Service idled out after {}s", secs as f64 + subsec_ms)
```
hawkw, line 53: If I'm reading this correctly, this looks like it will add `subsec_ms` to `secs` as though they were a number of seconds? Don't we need to divide them so they are only added to the fractional part of `secs`?

hawkw: (I think this also applies to the version of this change that merged to master, I just didn't notice it earlier)

linkerd/buffer/Cargo.toml
```diff
@@ -8,12 +8,13 @@ publish = false
...
 linkerd2-error = { path = "../error" }
-tokio = { version = "0.2", features = ["sync", "stream", "macros"] }
+tokio = { version = "0.2", features = ["sync", "stream", "time", "macros"] }
 tower = "0.3"
 tracing = "0.1"
 tracing-futures = { version = "0.1", features = ["std-future"] }
 pin-project = "0.4"
 
 [dev-dependencies]
+tower = { version = "0.3", features = ["util"] }
```
hawkw, line 18: nit: let's also add
```suggestion
tower = { version = "0.3", default_features = false, features = ["util"] }
```
to tower dependencies.

We don't need that to be enabled, since we are collecting all tracing events as traces already.

olix0r:
I took another pass at this to simplify `Dispatch` to be a simpler `async fn`

linkerd/buffer/src/lib.rs
```diff
@@ -17,33 +18,40 @@ struct InFlight<Req, F> {
...
 {
+    use futures::future;
+
     let (tx, rx) = mpsc::channel(capacity);
     let (ready_tx, ready_rx) = watch::channel(Poll::Pending);
-    let dispatch = Dispatch::new(inner, rx, ready_tx);
+    let idle = move || match idle_timeout {
+        Some(t) => future::Either::Left(dispatch::idle(t)),
+        None => future::Either::Right(future::pending()),
+    };
```
kleimkuhler, line 40: Any reason this is a closure and not just pass it in directly to `dispatch::run`? If passed directly it wouldn't need to to evaluate it each iteration

kleimkuhler: Oh is this so `idle` can be repeatedly polled in the loop? This would move the value if we passed in `idle: I` instead of `idle: Fn() -> I`

linkerd/buffer/src/dispatch.rs
```diff
@@ -1,146 +1,169 @@
...
-        });
-
-        // Drive requests from the queue to the inner service.
-        loop {
-            let ready = match this.inner.as_mut() {
-                Some(inner) => inner.poll_ready(cx),
-                None => {
-                    // This is safe because ready.closed() has returned Pending.
-                    return Poll::Pending;
+        select_biased! {
```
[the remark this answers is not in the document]
olix0r: does tokio have a select_biased? we need to be able to bias towards requests.

linkerd/buffer/src/lib.rs
olix0r: Right, this is a factory for idles, so they can be reset every time we start waiting for a request.

linkerd/buffer/src/dispatch.rs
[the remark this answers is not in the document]
kleimkuhler: Ah hm yea I think tokio's randomly chooses one to poll first.

PR #506, opened 2020-05-06
fix concurrency-limit not polling inner service

hawkw:
This fixes a bug where the std::future version of
`linkerd-concurrency-limit` fails to drive the wrapped service to
readiness before calling it. Now, the inner service is polled once a
semaphore permit has been acquired.

Shoutout to @LucioFranco for noticing this!


LucioFranco:
Thanks to @bruceg!

PR #507, opened 2020-05-06
Add middleware that rejects connections with no identity

zaharidichev:
This simple change adds a layer that will fail the accept if no TLS identity is present. This can be toggled by setting the `LINKERD2_PROXY_IDENTITY_REQUIRED` end variable


linkerd/app/src/env.rs
```diff
@@ -1005,6 +1021,7 @@ impl fmt::Display for EnvError {
         match self {
             EnvError::InvalidEnvVar => write!(f, "invalid environment variable"),
             EnvError::NoDestinationAddress => write!(f, "no destination service configured"),
+            EnvError::ConflictingConfig => write!(f, "conflicting configuration"),
```
olix0r, line 1024: FWIW, I believe we just handle this elsewhere as `InvalidEnvVar`. I'd probably be inclined to leave it that way, otherwise we should probably go update these other cases to use the new error variant.

linkerd/app/core/src/reject_with_no_identity.rs
```diff
@@ -0,0 +1,70 @@
...
+
+#[derive(Debug, Clone)]
+pub struct RejectWithNoIdentity<A> {
+    enabled: bool,
+    accept: A,
+}
+
+impl RejectWithNoIdentityLayer {
+    pub fn new(enabled: bool) -> Self {
+        Self { enabled }
```
olix0r, line 32: TIOLI, but we can decouple this middleware from `linkerd/app/core` and potentially make it more generally useful...

If we instead define a
```rust
pub trait Admit<T> {
  type Error: Into<Error>;
  fn admit(&mut self, target: &T) -> Result<(), Self::Error>;
}
```

Then (1) we can totally decouple this from "identity" -- the middleware is just a service and it doesn't even have to know about `Accept`, for that matter...

Then, up in app/core, we define an implementation that knows how to handle `(accept::Meta, BoxedIo)`. Here's a [sketch](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=401eb438c31a626c21f426b996ff7333).

olix0r:
How will this feature interact with Kubernetes healthchecks? Won't we block all unmeshed traffic to the pod?

zaharidichev:
> How will this feature interact with Kubernetes healthchecks? Won't we block all unmeshed traffic to the pod?

@olix0r Yes, unfortunatelly this is correct. The checks will fail in this case. See [this comment](https://github.com/linkerd/linkerd2/pull/4339#issuecomment-624631632) for how I went about it with the gateway specific case

linkerd/app/core/src/reject_with_no_identity.rs
zaharidichev: @olix0r Yes decoupling it makes sense. Seems like I am missing something but The one thing I did not quite get is how is the Admit a service? Or are you suggesting using the Admit in the same fashion we are using the  Detect and just passing it into a service that will either admit it and call the inner svc or not admit it and return the err.

olix0r: @zaharidichev The latter. Basically, modify this middleware to take an `Admit`. I'd then probably recommend moving this into a more generalized (or even dedicated) crate. Also.. we may already have a middleware that works like this. I'm not sure.

PR #510, opened 2020-05-08
update Hyper and tokio::{io, net} dependencies

hawkw:
This branch updates the Hyper dependency to 0.13 (and thus also updates
`http`/`http-body`), and updates all uses of `AsyncRead` and
`AsyncWrite` and `tokio::net` types to use the Tokio 0.2 versions.This
also implies updating `bytes` from 0.4 to 0.5. Unfortunately, this
change is rather massive, but all these dependencies are pretty tightly
coupled. It wasn't really possible to upgrade them incrementally the way
it was for code in Linkerd, at least not without forking and patching the
dependencies, which I wanted to avoid.

## Note on Test Support

This branch also makes a great deal of changes to the test support code
in `linkerd2-app-integration`. This is necessary because this code no
longer compiled after updating Hyper and the IO traits. When practical,
I've tried to make this a direct translation of the existing code. 

I think there's a lot of opportunity to improve this code and reduce
complexity. In particular, the test-support clients work by spawning a
background thread with a Tokio runtime on it, and sending work to that
task and responses back over channels. This is so we can present a
synchronous API to the tests themselves, and write straight-line code in
tests rather than nested futures. In an async/await world, that's kind
of unnecessary. We could switch the integration tests to be async
functions written using `#[tokio::test]`, and remove the complexity of
managing multiple background threads and runtimes for each test client
(and server). We would probably still want to isolate the test _proxy_
in its own runtime as we do currently, but we could run all the test
support code in the main test thread's runtime and not have to do all
this messing around with threads and channels.

However, I didn't do this kind of refactoring in this PR, beyond adding
async _versions_ of some of the existing test support APIs (which still
use channels internally). These were added for use in tests which have
to call into async APIs from other crates and therefore need to be
`async fn`s. I wanted to avoid totally rewriting the test support code
in this branch, however, since it's already a huge change.

## Note on IO Trait Changes

The `BoxedIo` trait used to have a special `shutdown_write` method that
would call `std::net::TcpStream::shutdown(Shutdown::Write)`. This was
added because Tokio 0.1's `<TcpStream as AsyncWrite>::shutdown` did
_not_ call this method; it simply returned `Ok(())`. However, Tokio
0.2's `<TcpStream as AsyncWrite>::poll_shutdown` *does* call
`std::net`'s `TcpStream::shutdown`. Since `BoxedIo` was ported directly
from the 0.1 code and still called both `AsyncWrite::poll_shutdown`
_and_ `Io::shutdown_write`, we would try to shut down the stream twice,
and the second call would fail with a "transport endpoint is not
connected" error. This broke the `tls_accept` tests, which expected
there to be no error.

This commit removes the `Io::shutdown_write` method, since Tokio's
`TcpStream` now does what we want.


hawkw:
I know this branch is kind of massive. However, this should unblock pretty much all the remaining update work. Now that we're on the `std::future` versions of Hyper and `tokio::{io, net}`, we basically don't need any more compat layers between futures 0.1 and `std::future`. It should now be quite easy to turn the proxy back into an actual proxy, and start re-enabling functionality layer-by-layer.

linkerd/app/integration/src/identity.rs
```diff
@@ -217,7 +216,7 @@ impl Controller {
 
 impl pb::server::Identity for Controller {
     type CertifyFuture =
-        Box<dyn Future<Item = grpc::Response<pb::CertifyResponse>, Error = grpc::Status> + Send>;
+        Box<dyn Future01<Item = grpc::Response<pb::CertifyResponse>, Error = grpc::Status> + Send>;
```
olix0r, line 219: Is this because we haven't moved to tonic yet?

hawkw: Yup, that's right.

linkerd/io/src/boxed.rs
```diff
@@ -1,74 +1,87 @@
...
-    fn flush(&mut self) -> Result<()> {
-        self.0.flush()
+    fn poll_read_buf<B: BufMut>(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context,
+        mut buf: &mut B,
+    ) -> Poll<usize> {
+        // A trait object of AsyncWrite would use the default poll_read_buf,
+        // which doesn't allow vectored writes. Going through this method
+        // allows the trait object to call the specialized write_buf method.
```
olix0r, line 29: Should refer to reads instead of writes

linkerd/proxy/http/src/client.rs
```diff
@@ -126,11 +137,8 @@ where
                 wants_h1_upgrade: _,
                 was_absolute_form,
             } => {
-                let exec =
-                    tokio::executor::DefaultExecutor::current().instrument(info_span!("http1"));
-
+                #[allow(deprecated)]
```
olix0r, line 140: Out of curiosity, is there something else we should be doing instead? Or should we open a discussion on hyper about how to satisfy our use case?

linkerd/proxy/http/src/client.rs
```diff
@@ -160,46 +168,51 @@ impl<C: Clone, B> Clone for MakeClient<C, B> {
...
+    type Output = Result<Client<C, T, B>, Error>;
 
-    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-        let svc = match *self {
-            MakeFuture::Http1(ref mut h1) => Client::Http1(h1.take().expect("poll more than once")),
-            MakeFuture::Http2(ref mut h2) => {
-                let svc = try_ready!(h2.poll());
+    #[project]
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        #[project]
```
olix0r, line 183: For my understanding, what does this inner `#[project]` accomplish?

linkerd/proxy/http/src/h2.rs
```diff
@@ -30,17 +33,48 @@ pub struct Connection<B> {
...
+    B::Data: Send,
+    B::Error: std::error::Error + Send + Sync,
+    F::Ok: AsyncRead + AsyncWrite + Send + 'static,
+{
+    Connect(#[pin] F),
+    Handshake(
+        #[pin]
+        Pin<
+            Box<
+                dyn Future<
```
olix0r, line 64: Is this because `handshake` no longer returns a named type? If so, would it better to add a generic for this TryFuture? I don't have a strong opinion... just trying to figure out what the right idiom is.

olix0r: I realized that if it's unnamed we _have to_ box it. Nevermind

hawkw: yeah, that's correct. eventually, it would be nice to replace this hand-written future with an async block future (on nightly, where we can return `impl Future` from tower services), and thus do away with the box; otherwise, there will always be a box somewhere.

linkerd/proxy/http/src/client.rs
hawkw: The `#[project]` attribute [allows matching on projected enum variants](https://docs.rs/pin-project/0.4.16/pin_project/attr.project.html#match-expressions). Since attribute macros on expressions (such as `match`) are not available on stable, we have to also place it on the function so that `pin-project` can pre-process the function body to find the attribute.

linkerd/proxy/http/src/client.rs
hawkw: The `Builder::keep_alive` function was deprecated. After another look at the docs, it turns out it's actually pretty easy to use the new API (you set `pool_max_idle_per_host` to 0 to disable keepalive), so I've removed this in f21809a430b440b3e49d87197c6891e4b36bf04a

linkerd/app/core/src/svc.rs
```diff
@@ -289,13 +289,13 @@ impl<S> Stack<S> {
...
-    {
-        self.push(http::boxed::request::Layer::new())
-    }
+    // pub fn box_http_request<B>(self) -> Stack<http::boxed::BoxRequest<S, B>>
+    // where
+    //     B: hyper::body::HttpBody<Data = http::boxed::Data, Error = Error> + 'static,
+    //     S: tower::Service<http::Request<http::boxed::Payload>>,
+    // {
+    //     self.push(http::boxed::request::Layer::new())
+    // }
```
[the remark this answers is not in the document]
hawkw: we can put it back later.

linkerd/app/integration/src/lib.rs
```diff
@@ -86,7 +91,7 @@ macro_rules! assert_eventually {
                 })
                 .unwrap_or($crate::DEFAULT_TEST_PATIENCE);
             let start_t = Instant::now();
-            for i in 0..($retries + 1) {
+            for i in 0i32..($retries as i32 + 1i32) {
```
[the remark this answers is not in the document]
hawkw: The compiler is not currently great at inferring types inside async blocks. Without the suffixes & cast, it actually errors.

linkerd/app/integration/tests/transparency.rs
```diff
@@ -612,29 +619,32 @@ macro_rules! http1_tests {
...
-        #[test]
+        #[tokio::test]
         #[cfg_attr(not(feature = "nyi"), ignore)]
-        fn http1_request_with_body_chunked() {
+        async fn http1_request_with_body_chunked() {
             let _ = trace_init();
 
             let srv = server::http1()
-                .route_async("/", |req| {
+                .route_async("/", |req| async move {
```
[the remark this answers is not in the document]
hawkw: hopefully the test support stuff will no longer have separate sync and async APIs soon, so that naming scheme will go away except for here.

linkerd/app/integration/src/tcp.rs
```diff
@@ -110,108 +120,97 @@ impl TcpConn {
...
-    pub fn read_timeout(&self, timeout: Duration) -> Vec<u8> {
-        use linkerd2_test_util::BlockOnFor;
-        current_thread::Runtime::new()
-            .unwrap()
-            .block_on_for(timeout, self.read_future())
+    #[tokio::main(basic_scheduler)]
+    pub async fn read_timeout(&self, timeout: Duration) -> Vec<u8> {
+        tokio::time::timeout(timeout, self.read_future())
+            .await
+            .unwrap_or_else(|e| panic!("TcpConn(addr={}) read() error: {:?}", self.addr, e))
```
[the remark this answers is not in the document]
hawkw: It's actually not extra — `timeout` on an `F`-typed future gives us a `Result<F::Output, Elapsed>`, but in this case, `F::Output` is also a `Result`, so we have a `Result<Result<...>, ...>`.

linkerd/app/integration/src/lib.rs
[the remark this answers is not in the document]
kleimkuhler: Hm I had checked before commenting this and ran `cargo test` without issue

PR #511, opened 2020-05-08
Buffer requests while the service is pending

olix0r:
The buffer uses a watch to advertise readiness to consumers. This is
intended to:

1. Prevent the out service from becoming ready before the inner service
   has initialized.
2. Surface errors to outer services' `poll_ready`.

It, however, currently advertises when the inner service moves into
pending, preventing requests from being admitted to the buffer, and then
basically making consumers race to enter the buffer.

This change also relaxes its use of the ready watch channel, especially
to signal that the task should complete. Instead, the dispatch task
completes as soon as all requests have been processed.

### DIFF
--- a/linkerd/buffer/src/dispatch.rs
+++ b/linkerd/buffer/src/dispatch.rs
@@ -88,13 +88,6 @@ where
     type Error = Never;
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-        macro_rules! return_ready {
-            () => {{
-                trace!("Complete");
-                return Ok(Async::Ready(()));
-            }};
-        }
-
         // Drive requests from the queue to the inner service.
         loop {
             match self.inner.poll_ready() {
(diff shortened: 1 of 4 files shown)


olix0r:
@hawkw FWIW this brings the buffer to match the behavior in #505

linkerd/buffer/src/dispatch.rs
```diff
@@ -201,20 +179,11 @@ mod test {
...
-                assert!(
-                    dispatch.inner.is_none(),
-                    "Did not drop inner service after timeout."
-                );
+                assert!(dispatch.poll().unwrap().is_ready());
                 assert!(
                     ready_rx.get_ref().is_err(),
                     "Did not advertise an error to consumers."
                 );
-                drop(ready_rx);
```
kleimkuhler, line 213: Should we still include a test that asserts dropping all senders results in the `Dispatch` being ready?

olix0r:
Here's a comparison between a proxyless baseline, this branch, and master 
image.

This branch seems like a step in the right direction.

olix0r:
At higher concurrency, the benefits of this branch are even more pronounced: 
image

olix0r:
@hawkw 

> in a world where we're admitting requests as long as there's buffer capacity and just having the request future wait on the service, do we really need to have a channel & worker task any longer?

I think so? Something has to be responsible for driving the inner service to ready. I don't think we want every consumer to have to fight over that.

PR #512, opened 2020-05-13
Add an `i/o` error label to http metrics

olix0r:
This change modifies HTTP error-labeling to detect I/O errors
and label them explicitly. Previously all I/O errors were reported as
`unexpected`.

Fixes linkerd/linkerd2#4364

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -641,6 +641,7 @@ dependencies = [
  "linkerd2-dns",
  "linkerd2-drain",
  "linkerd2-duplex",
+ "linkerd2-errno",
  "linkerd2-error",
  "linkerd2-error-metrics",
  "linkerd2-error-respond",
@@ -855,6 +856,10 @@ dependencies = [
  "tokio",
  "tracing",
 ]
+
+[[package]]
+name = "linkerd2-errno"
+version = "0.1.0"
 
 [[package]]
 name = "linkerd2-error"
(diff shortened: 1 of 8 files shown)


adleong:
Is this as granular as we can get here?  Can we distinguish between different IO errors at all?  For example, distinguish between connection refused and connection closed?

hawkw:
> Is this as granular as we can get here? Can we distinguish between different IO errors at all? For example, distinguish between connection refused and connection closed?

We already have some code for formatting labels based on OS error codes (https://github.com/linkerd/linkerd2-proxy/blob/d213c1373f1f5ec479a59a0a388d4373dcb3cde4/linkerd/proxy/transport/src/metrics/errno.rs) that we're already using for TCP metrics:
https://github.com/linkerd/linkerd2-proxy/blob/d213c1373f1f5ec479a59a0a388d4373dcb3cde4/linkerd/proxy/transport/src/metrics/mod.rs#L364-L377

Possibly we could reuse this?

olix0r:
> Is this as granular as we can get here? Can we distinguish between different IO errors at all? For example, distinguish between connection refused and connection closed?

You should never see a connection refused for http errors. We never dispatch requests to endpoints that are disconnected. This really can only indicate that a connection was terminated unexpectedly, as far as I understand.

If someone else would like to propose an alternative that handles more metadata, please do. If we want to get something into this week's edge, I think we should take this approach.

PR #516, opened 2020-05-14
turn inbound proxy back into a proxy

hawkw:
Now that the HTTP connect- and accept-side machinery has all been
updated to use `std::future`, we can start turning the proxy back
into...a proxy. This branch does so on the inbound side. Much of the
actual functionality (approximately everything not directly required to
actually proxy requests) is still disabled, as a lot of the code still
needs to be updated.

I had to make a few additional changes in order to get the proxy to
compile, primarily to deal with trait bounds on `Service` impls that
were no longer satisfied as a result of changes made to update other
parts of the codebase. I also disabled two more of the  `transparency`
integration tests, both of which tested whether the proxy stripped
connection-level headers from proxied requests. Previously, these were
passing "accidentally" — because the proxy was replaced with a HTTP
server that just responded to every request with an empty 200 OK
response, the headers were, of course, not present. Now that we're
actually proxying requests, these tests fail, because the code for
stripping headers has yet to be updated.

### DIFF
--- a/linkerd/buffer/src/service.rs
+++ b/linkerd/buffer/src/service.rs
@@ -61,12 +61,12 @@ impl<Req, F> Buffer<Req, F> {
     }
 }
 
-impl<Req, F> tower::Service<Req> for Buffer<Req, F>
+impl<Req, F, E, T> tower::Service<Req> for Buffer<Req, F>
 where
-    F: TryFuture,
-    F::Error: Into<Error>,
+    F: Future<Output = Result<T, E>>,
+    E: Into<Error>,
 {
-    type Response = F::Ok;
+    type Response = T;
     type Error = Error;
     type Future = ResponseFuture<F>;
 
--- a/linkerd/buffer/src/service.rs
+++ b/linkerd/buffer/src/service.rs
@@ -61,12 +61,12 @@ impl<Req, F> Buffer<Req, F> {
     }
 }
 
-impl<Req, F, E, T> tower::Service<Req> for Buffer<Req, F>
+impl<Req, F> tower::Service<Req> for Buffer<Req, F>
 where
-    F: Future<Output = Result<T, E>>,
-    E: Into<Error>,
+    F: TryFuture,
+    F::Error: Into<Error>,
 {
-    type Response = T;
+    type Response = F::Ok;
     type Error = Error;
     type Future = ResponseFuture<F>;
 
--- a/linkerd/cache/src/layer.rs
+++ b/linkerd/cache/src/layer.rs
@@ -27,7 +27,7 @@ impl<L: Clone, T> Clone for CacheLayer<T, L> {
 
 impl<T, L, N> tower::layer::Layer<N> for CacheLayer<T, L>
 where
-    T: Eq + std::hash::Hash,
+    T: Eq + std::hash::Hash + Send + 'static,
     N: NewService<T> + Clone,
     L: tower::layer::Layer<Track<N>> + Clone,
     L::Service: NewService<T>,
--- a/linkerd/cache/src/lib.rs
+++ b/linkerd/cache/src/lib.rs
@@ -34,7 +34,7 @@ type Services<T, S> = HashMap<T, (S, Weak<()>)>;
 
 impl<T, N> Cache<T, N>
 where
-    T: Eq + Hash,
+    T: Eq + Hash + Send + 'static,
     N: NewService<(T, Handle)>,
 {
     pub fn new(new_service: N) -> Self {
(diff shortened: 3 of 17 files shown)


hawkw:
this compiles but there appear to be some (new!) test failures, which is cool and fun!

linkerd/buffer/src/service.rs
```diff
@@ -61,12 +61,12 @@ impl<Req, F> Buffer<Req, F> {
...
-impl<Req, F> tower::Service<Req> for Buffer<Req, F>
+impl<Req, F, E, T> tower::Service<Req> for Buffer<Req, F>
 where
-    F: TryFuture,
-    F::Error: Into<Error>,
+    F: Future<Output = Result<T, E>>,
+    E: Into<Error>,
 {
-    type Response = F::Ok;
+    type Response = T;
```
olix0r, line 69: why the change?

linkerd/cache/src/layer.rs
```diff
@@ -27,7 +27,7 @@ impl<L: Clone, T> Clone for CacheLayer<T, L> {
 
 impl<T, L, N> tower::layer::Layer<N> for CacheLayer<T, L>
 where
-    T: Eq + std::hash::Hash,
+    T: Eq + std::hash::Hash + Send + 'static,
```
olix0r, line 30: Does the cache actually require this? Or from debugging?

linkerd/cache/src/lib.rs
```diff
@@ -34,7 +34,7 @@ type Services<T, S> = HashMap<T, (S, Weak<()>)>;
 
 impl<T, N> Cache<T, N>
 where
-    T: Eq + Hash,
+    T: Eq + Hash + Send + 'static,
```
olix0r, line 37: Does the cache actually require this? Or from debugging?

linkerd/proxy/http/src/glue.rs
```diff
@@ -274,11 +274,8 @@ where
...
 }
 
-impl<C> hyper_connect::Connection for Connection<C>
-where
-    C: hyper_connect::Connection,
-{
+impl<C> hyper_connect::Connection for Connection<C> {
     fn connected(&self) -> hyper_connect::Connected {
-        self.transport.connected().proxy(self.absolute_form)
+        hyper_connect::Connected::new().proxy(self.absolute_form)
```
olix0r, line 279: What's the deal with `self.transport` now? Unused?

linkerd/cache/src/layer.rs
hawkw: I believe this is necessary due to how the `Lock` layer works now — the type inside the lock must be `'static` so that it can go in a boxed future.

linkerd/proxy/http/src/glue.rs
hawkw: `self.transport` is the underlying transport. When I first wrote the `Connection` impl, I bounded the type of the transport with `C: hyper::client::connect::Connected`, which is implemented for `TcpStream`s. However, in practice, it's a `BoxedIo`, and we don't currently forward that trait.

We may want to revisit this in the future and start forwarding the `Connected` metadata through the whole stack, but for now I think this is fine.

linkerd/cache/src/lib.rs
hawkw: The `Service` impl for `Cache` already has this bound on `master-tokio-0.2`:
https://github.com/linkerd/linkerd2-proxy/blob/1755e2d381a2a341a233a9fe3c13cfbcc78b0170/linkerd/cache/src/lib.rs#L66

It's required because `Lock::poll_acquire` requires that the type inside the lock be `'static` in order to box the `tokio::sync::Mutex::lock` future.

linkerd/buffer/src/service.rs
hawkw: I think this can be rolled back.

linkerd/proxy/http/src/glue.rs
olix0r: Oh, I see... this is just a metadata type and not an io stream.

hawkw: Yeah, that's right. Setting `proxy` on the metadata determines whether absolute URIs are used (should probably add a comment explaining that).

There are additional things we can set on the metadata such as whether H2 has already been negotiated and extensions to automatically add on every response received on that connection. We may wish to use these in the future. I believe that when returned by a `TcpStream`, the metadata is empty, so I don't think it's a problem that we don't forward it.

PR #518, opened 2020-05-18
port `linkerd2-proxy-error::recover` and `linkerd2-proxy-resolve` to std::future

hawkw:
This branch updates the `linkerd2-proxy-resolve` crate to `std::future`.
It also updates the `recover` module in `linkerd2-proxy-error`, which is
a dependency of `resolve`, and `linkerd2-reconnect`, which depends on
`recover` and therefore didn't compile after it was updated.

The `linkerd2-proxy-resolve::recover` code currently only works when the
underlying resolver's resolution and future types are both `Unpin`, and
the backoff type is also `Unpin`. This is because this logic relies a
bit on being able to move these values in and out of `Option`s. I tried
to implement this without requiring these values to be `Unpin`, but the
implementation of the `ResolveFuture`, which drives the resolution until
it's connected and then returns it, was more or less impossible to
translate in its current state. 
 
I suspect that this won't be an issue when updating code that depends on
`recover` — I think the resolutions will probably be `Unpin` without any
intervention from us. If not, we can always either `Box::pin` them, or
try to do a more complete rewrite of this code so that it's refactored
in a way that avoids these issues.

In other cases, we can sometimes just replace code that is difficult to
port to use `async`/`await`. Here, that was not possible, because we are
implementing the `Resolution` trait, rather than a future. We _may_ want
to consider rewriting `Resolution` to be a trait alias for a `Stream`,
so that we can use the `async-stream` crate, but that may have other
downsides, so I decided to not do that just yet.

Hopefully this all makes sense!

### DIFF
--- a/linkerd/proxy/resolve/src/recover.rs
+++ b/linkerd/proxy/resolve/src/recover.rs
@@ -1,12 +1,11 @@
 //! A middleware that recovers a resolution after some failures.
 
-use futures::{ready, Stream};
+use futures::{ready, stream::TryStreamExt};
 use indexmap::IndexMap;
 use linkerd2_error::{Error, Recover};
-use linkerd2_proxy_core::resolve::{self, Resolution as _, Update};
+use linkerd2_proxy_core::resolve::{self, Update};
 use pin_project::{pin_project, project};
 use std::future::Future;
-use std::net::SocketAddr;
 use std::net::SocketAddr;
 use std::pin::Pin;
 use std::task::{Context, Poll};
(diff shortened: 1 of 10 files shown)


linkerd/proxy/resolve/src/recover.rs
```diff
@@ -1,487 +1,516 @@
...
+        ready!(this
+            .inner
+            .as_mut()
+            .expect("polled after complete")
+            .poll_connected(cx))?;
+        let inner = this.inner.take().expect("polled after complete");
+        Poll::Ready(Ok(Resolution {
+            inner,
+            cache: IndexMap::default(),
+            //cache: Cache::default(),
```
kleimkuhler, line 147: Can remove this

PR #520, opened 2020-05-19
update dns to std::future

hawkw:
This branch updates the `linkerd2-proxy-dns` crate to `std::future` and
the latest release of `trust-dns-resolver`, and updates the uses of DNS
in `linkerd2-proxy-app` to track the changes. 

For the most part, this change is pretty straightforward. However, the
`std::future` version of `trust-dns-resolver` changed how the background
task driving DNS resolution is executed. Previously, it was returned by
the `AsyncResolver` constructor, but now, the constructors are `async
fn`s that implicitly spawn the background task on the runtime that
executes them.

Therefore, I had to change how the DNS resolver is constructed. I opted
to build the admin runtime early and have it block on the DNS resolver
construction, and then pass that runtime to the admin thread when it is
spawned. This seemed better than making the DNS resolver lazy by making
the underlying `trust-dns` async resolver an `Option` or something, and
having it panic when the resolver hasn't been constructed yet. Doing
that safely would have required adding a `RwLock` or something, which
didn't seem ideal.

### DIFF
--- a/linkerd/app/src/lib.rs
+++ b/linkerd/app/src/lib.rs
@@ -72,7 +72,7 @@ impl Config {
     ///
     /// It is currently required that this be run on a Tokio runtime, since some
     /// services are created eagerly and must spawn tasks to do so.
-    pub fn build(self, log_level: trace::LevelHandle) -> Result<App, Error> {
+    pub async fn build(self, log_level: trace::LevelHandle) -> Result<App, Error> {
         let Config {
             admin,
             dns,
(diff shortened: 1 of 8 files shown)


linkerd/app/src/lib.rs
```diff
@@ -86,7 +86,22 @@ impl Config {
...
+        let mut admin_rt = tokio_02::runtime::Builder::new()
+            .basic_scheduler()
+            .enable_all()
+            .build()
+            .expect("admin runtime");
+
+        let dns = dns
+            .build(admin_rt.handle().clone())
+            .instrument(info_span!("dns"))
+            .await;
```
olix0r, line 104: WDYT about putting this in `admin::Config::build` and then just giving a handle to the resolver off of that?

Definitely not a blocker.

hawkw: Hmm, that seemed nice at first until I actually tried it, and it turned out to be kinda awkward: we currently construct the identity client and then pass it in to `admin::Config::build`, and constructing the identity client depends on the DNS resolver:
https://github.com/linkerd/linkerd2-proxy/pull/520/files#diff-104ba36cb3e84114b06ebde95af46477R107

so, I _think_ that if we were going to have the `Admin` struct own the runtime, we'd need to make the resolver lazy internally?

olix0r: Nevermind. Let's come back to this later.

PR #525, opened 2020-05-22
outbound: Prevent loops

olix0r:
This change modifies the outbound proxy to fail to build services
targetting localhost:4140 (where 4140 is the outbound port). This
prevents looping and will result in 502s.


olix0r:
@hawkw 

```
[   110.714608758s]  WARN outbound:accept{peer.addr=127.0.0.1:59504}:source{target.addr=127.0.0.1:4140}: linkerd2_app_core::errors: Failed to proxy request: outbound requests must not target localhost:4140
```

PR #526, opened 2020-05-22
turn outbound back into a proxy

hawkw:
This branch turns the outbound side of the proxy back into a real proxy,
similarly to what #516 did for the inbound side. Getting outbound to
work again was somewhat more complex, since the outbound side also
requires the load balancer (and, therefore, discovery). In order to get
this all working, it was also necessary to update the
`linkerd2-request-filter` crate (which was quite trivial), as it's
necessary for the resolver to work, and to make some changes in the
resolver so that all trait bounds were satisfied.

In addition, now that both inbound and outbound work, we're now able
to re-enable a large amount of the integration tests in the 
`transparency` and `discovery` modules. Some of these tests are still
disabled, as they either depend on service profiles, HTTP upgrades and
orig-proto, or header maniuplation, all of which have yet to be 
updated.

### DIFF
--- a/linkerd/app/Cargo.toml
+++ b/linkerd/app/Cargo.toml
@@ -31,7 +31,7 @@ futures_03 = { package = "futures", version = "0.3" }
 tokio-compat = "0.1"
 tokio_02 = {package = "tokio", version = "0.2", features = ["rt-util"] }
 http-body = "0.3"
-
+tower = "0.3"
 
 [dev-dependencies]
 bytes = "0.5"
@@ -48,5 +48,5 @@ tokio-connect = { git = "https://github.com/carllerche/tokio-connect" }
 tokio-io = "0.1.6"
 tokio-current-thread = "0.1.4"
 tokio-rustls = "0.13"
-tower = "0.1"
+# tower = "0.1"
 webpki = "=0.21.0"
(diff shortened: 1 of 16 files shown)


linkerd/app/Cargo.toml
```diff
@@ -48,5 +48,5 @@ tokio-connect = { git = "https://github.com/carllerche/tokio-connect" }
 tokio-io = "0.1.6"
 tokio-current-thread = "0.1.4"
 tokio-rustls = "0.13"
-tower = "0.1"
+# tower = "0.1"
```
olix0r, line 51: rm?

linkerd/proxy/resolve/src/recover.rs
```diff
@@ -75,7 +75,12 @@ enum State<F, R: resolve::Resolution, B> {
 
 // === impl Resolve ===
 
-impl<E, R> Resolve<E, R> {
+impl<E, R> Resolve<E, R> 
+where
+R: Clone,
+E: Recover + Clone,
+E::Backoff: Unpin,
```
olix0r, line 82: surprised this passes rustfmt?

hawkw: agh, i thought i deleted this

PR #527, opened 2020-05-22
Add loop detection to inbound & TCP forwarding

olix0r:
e77fe181 introduced loop detection to the outbound HTTP proxy. This
change extends this behavior to the inbound HTTP proxy and the TCP
proxy for both inbound and outbound. This helps ensure malicious
requests can't consume proxy resources.

### DIFF
--- a/linkerd/app/inbound/src/prevent_loop.rs
+++ b/linkerd/app/inbound/src/prevent_loop.rs
@@ -1,4 +1,4 @@
-use super::endpoint::{HttpEndpoint, Target, TcpEndpoint};
+use super::endpoint::{Target, TcpEndpoint};
 use linkerd2_app_core::admit;
 
 /// A connection policy that drops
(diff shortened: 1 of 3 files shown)


linkerd/app/inbound/src/prevent_loop.rs
```diff
@@ -0,0 +1,57 @@
...
+
+        Ok(())
+    }
+}
+
+impl std::fmt::Display for LoopPrevented {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "inboudn requests must not target localhost:{}",
```
hawkw, line 51: typo:
```suggestion
            "inbound requests must not target localhost:{}",
```

linkerd/app/inbound/src/prevent_loop.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    fn admit(&mut self, ep: &TcpEndpoint) -> Result<(), Self::Error> {
+        tracing::debug!(port = %ep.port, self.port);
+        if ep.port == self.port {
+            return Err(LoopPrevented { port: self.port });
+        }
+
+        Ok(())
+    }
+}
```
hawkw, line 45: nit/tioli: it seems like there _could_ be some kinda trait for "things that have a port" that we could use to generalize this? but, it's not repetitive enough for it to matter all that much, esp. since i can't immediately think of a way to generalize between the inbound and outbound logic...

olix0r: It would take just as much code in endpoints to satisfy it though ;p

PR #529, opened 2020-05-23
Update strip_header to std::future

kleimkuhler:
This branch updates `proxy::app::strip_header` module to `std::future`. It is a fairly straightforward change!

### DIFF
--- a/linkerd/proxy/http/src/strip_header.rs
+++ b/linkerd/proxy/http/src/strip_header.rs
@@ -157,7 +157,7 @@ pub mod response {
             let this = self.project();
             let mut res = ready!(this.inner.try_poll(cx))?;
             res.headers_mut().remove(this.header.clone());
-            Poll::Ready(Ok(res.into()))
+            Poll::Ready(Ok(res))
         }
     }
 }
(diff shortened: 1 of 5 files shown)


linkerd/proxy/http/src/strip_header.rs
```diff
@@ -141,16 +148,16 @@ pub mod response {
...
 
-        fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-            let mut res = try_ready!(self.inner.poll());
-            res.headers_mut().remove(self.header.clone());
-            Ok(res.into())
+        fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+            let this = self.project();
+            let mut res = ready!(this.inner.try_poll(cx))?;
+            res.headers_mut().remove(this.header.clone());
+            Poll::Ready(Ok(res.into()))
```
hawkw, line 160: The `Into` conversion here was to turn the inner value into an `Async::Ready`; it's no longer necessary and we should remove it
```suggestion
            Poll::Ready(Ok(res))
```

hawkw:
I believe this still needs a +1 from someone who isn't me.

PR #534, opened 2020-05-26
fallback: Unwrap errors recursively

olix0r:
This change modifies the fallback layer to inspect error sources
recursively to determine if the given error type is satisfied.

A stack-helper is also added for this case.

### DIFF
--- a/linkerd/stack/src/fallback.rs
+++ b/linkerd/stack/src/fallback.rs
@@ -59,8 +59,19 @@ impl<B> FallbackLayer<B> {
     where
         E: std::error::Error + 'static,
     {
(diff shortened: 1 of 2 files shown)


linkerd/stack/src/fallback.rs
```diff
@@ -59,10 +59,21 @@ impl<B> FallbackLayer<B> {
...
 
+fn is_error<E>(err: &(dyn std::error::Error + 'static)) -> bool
+where
+    E: std::error::Error + 'static,
+{
+    if err.is::<E>() {
+        return true;
+    }
+
+    err.source().map(is_error::<E>).unwrap_or(false)
```
hawkw, line 74: when [`error::Chain`](https://doc.rust-lang.org/nightly/std/error/struct.Chain.html) stabilizes, we might just want to use that, but this looks right for now!

PR #537, opened 2020-05-27
update service profiles to std::future

hawkw:
This branch updates the service profiles controller client and stack
layers to `std::future`, and puts service profiles back in the outbound
and inbound stacks. Retries do not actually happen yet because the retry
code hasn't been updated. 

All of the tests for profile destination overrides can now be
re-enabled. In addition, some of the the tests in `profiles.rs` could
also be re-enabled, with the exception of the ones that actually expect
retries to happen.

Depends on #536.


linkerd/app/outbound/src/lib.rs
```diff
@@ -372,31 +373,23 @@ impl Config {
...
+                        .push_map_target(|inner: Logical<HttpEndpoint>| Concrete {
+                            addr: inner.addr.clone(),
+                            inner,
+                        })
+                        .push_on_response(svc::layers().box_http_response().box_http_request())
+                        .check_service::<Logical<HttpEndpoint>>()
+                        .into_inner(),
+                    is_discovery_rejected,
+                )
+                //     .check_service::<Logical<HttpEndpoint>>()
```
kleimkuhler, line 392: nit: Should be okay to un-comment this

linkerd/app/src/lib.rs
```diff
@@ -160,18 +160,15 @@ impl Config {
...
             let drain = drain_rx.clone();
             info_span!("inbound").in_scope(move || {
                 inbound.build(
-                    identity,
-                    (), // profiles,
-                    // tap,
-                    metrics,
-                    None, // oc,
+                    identity, profiles, // tap,
+                    metrics, None, // oc,
```
kleimkuhler, line 171: This is odd formatting. This should change though as tap and oc are added back in

hawkw: Yup, rustfmt is dead-set on formatting it this way, as far as I can tell, but hopefully it will be a bit less bad when those come back.

linkerd/service-profiles/src/client.rs
```diff
@@ -118,16 +134,19 @@ where
     S: GrpcService<BoxBody> + Clone + Send + 'static,
     S::ResponseBody: Send,
     <S::ResponseBody as Body>::Data: Send,
+    <S::ResponseBody as HttpBody>::Error:
+        Into<Box<dyn std::error::Error + Send + Sync + 'static>> + Send,
```
olix0r, line 138: I think this can just be `Into<Error>`?

hawkw: Ah, you're right — I had a memory of that not working, but I think that was something else, this compiles as you suggested.

PR #538, opened 2020-05-27
fix concurrency-limit layer not being Send

hawkw:
The `linkerd2-concurrency-limit` service was inadvertently not `Send`,
because it failed to bound the wrapped future trait object with `Send`.
This means that we couldn't actually use it in the inbound or outbound
stack, because other layers require the service to be `Send`.

This branch adds the missing `Send` bound, and removes the patch that
pinned Tokio to a Git revision, since `OwnedSemaphorePermit` is now in a
published Tokio release. Additionally, I've actually put back the
concurrency-limit layers now that they don't break everything any more.

### DIFF
--- a/linkerd/app/outbound/src/lib.rs
+++ b/linkerd/app/outbound/src/lib.rs
@@ -420,7 +420,7 @@ impl Config {
                 // Eagerly fail requests when the proxy is out of capacity for a
                 // dispatch_timeout.
                 .push_failfast(dispatch_timeout)
-                .push(metrics.http_errors)
+                // .push(metrics.http_errors)
                 // Synthesizes responses for proxy errors.
                 // .push(errors::layer())
                 // Initiates OpenCensus tracing.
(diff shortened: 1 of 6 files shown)


linkerd/app/outbound/src/lib.rs
```diff
@@ -420,7 +420,7 @@ impl Config {
                 // Eagerly fail requests when the proxy is out of capacity for a
                 // dispatch_timeout.
                 .push_failfast(dispatch_timeout)
-                .push(metrics.http_errors)
+                // .push(metrics.http_errors)
```
kleimkuhler, line 423: Is this another bound possibly missing?

hawkw: This doesn't work for unrelated reasons, and is updated on a separate branch; I had to comment it out here because we were not previously actually pushing the `http_admit_request` stack.

PR #540, opened 2020-05-28
Introduce a gateway between inbound and outbound

olix0r:
When the proxy receives inbound requests without an original dst address
(or with a original dst address matching the inbound listener), the
proxy currently fails these requests.

This change modifies the proxy to attempt to accept these requests and
forward them back through the outbound router.

The gateway requires that all requests are received over an mTLS-secured
connection. It also refines the destination through DNS to determine the
canonical-form name as well as an outbound original dst IP. All
gatewayed destinations must have a suffix as set by the
`LINKERD2_PROXY_GATEWAY_SUFFIXES` environment variable.

All requests that do not meet these criteria are failed with a `403
Forbidden` status.

### DIFF
--- a/linkerd/app/gateway/src/config.rs
+++ b/linkerd/app/gateway/src/config.rs
@@ -23,7 +23,7 @@ impl Config {
         Future = impl Send + 'static,
         Response = impl Send
                        + tower::Service<
-            http::Request<http::glue::HttpBody>,
+            http::Request<http::boxed::Payload>,
             Response = http::Response<http::boxed::Payload>,
             Error = impl Into<Error>,
             Future = impl Send,
@@ -41,7 +41,7 @@ impl Config {
         O::Future: Send + 'static,
         S: Send
             + tower::Service<
-                http::Request<http::glue::HttpBody>,
+                http::Request<http::boxed::Payload>,
                 Response = http::Response<http::boxed::Payload>,
             > + 'static,
         S::Error: Into<Error> + Send + 'static,
--- a/linkerd/app/gateway/src/config.rs
+++ b/linkerd/app/gateway/src/config.rs
@@ -1,4 +1,4 @@
-use super::make::GatewayMake;
+use super::make::MakeGateway;
 use indexmap::IndexSet;
 use linkerd2_app_core::{dns, proxy::http, Error};
 use linkerd2_app_inbound::endpoint as inbound;
@@ -47,6 +47,6 @@ impl Config {
         S::Error: Into<Error> + Send + 'static,
         S::Future: Send,
     {
-        GatewayMake::new(resolve, outbound, self.suffixes.clone())
+        MakeGateway::new(resolve, outbound, self.suffixes.clone())
     }
 }
--- a/linkerd/app/gateway/src/config.rs
+++ b/linkerd/app/gateway/src/config.rs
@@ -1,0 +1,52 @@
+use super::make::GatewayMake;
+use indexmap::IndexSet;
+use linkerd2_app_core::{dns, proxy::http, Error};
+use linkerd2_app_inbound::endpoint as inbound;
+use linkerd2_app_outbound::endpoint as outbound;
+use std::net::IpAddr;
+
+#[derive(Clone, Debug, Default)]
+pub struct Config {
+    pub suffixes: IndexSet<dns::Suffix>,
+}
+
+impl Config {
+    pub fn build<R, O, S>(
+        self,
+        resolve: R,
+        outbound: O,
+    ) -> impl Clone
+           + Send
+           + tower::Service<
+        inbound::Target,
+        Error = impl Into<Error>,
+        Future = impl Send + 'static,
+        Response = impl Send
+                       + tower::Service<
+            http::Request<http::glue::HttpBody>,
+            Response = http::Response<http::boxed::Payload>,
+            Error = impl Into<Error>,
+            Future = impl Send,
+        > + 'static,
+    >
+    where
+        R: tower::Service<dns::Name, Response = (dns::Name, IpAddr)> + Send + Clone,
+        R::Error: Into<Error> + 'static,
+        R::Future: Send + 'static,
+        O: tower::Service<outbound::Logical<outbound::HttpEndpoint>, Response = S>
+            + Send
+            + Clone
+            + 'static,
+        O::Error: Into<Error> + Send + 'static,
+        O::Future: Send + 'static,
+        S: Send
+            + tower::Service<
+                http::Request<http::glue::HttpBody>,
+                Response = http::Response<http::boxed::Payload>,
+            > + 'static,
+        S::Error: Into<Error> + Send + 'static,
+        S::Future: Send,
+    {
+        GatewayMake::new(resolve, outbound, self.suffixes.clone())
+    }
+}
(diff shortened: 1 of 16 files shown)


olix0r:
I've tested this locally on a docker harness to ensure that we get 403s without identity... I'll do some more testing that this actually works as intended in a proper k8s setup.

I've also not yet implemented loop-detection or identity propagation. That will follow in a separate PR.

Note that the gateway isn't particularly stacked/decomposed... I think that's okay for now. Once the gateway takes its shape it may be worth splitting out into components, but for now it's expedient and isolated (none of the gateway types escape the module due to impl trait). We may find that this general approach is more ergonomic: Focus on splitting functionality into modules that may use stacks internally but ultimately expose some opaque type.

linkerd/app/gateway/src/config.rs
```diff
@@ -0,0 +1,52 @@
...
+        Error = impl Into<Error>,
+        Future = impl Send + 'static,
+        Response = impl Send
+                       + tower::Service<
+            http::Request<http::glue::HttpBody>,
+            Response = http::Response<http::boxed::Payload>,
+            Error = impl Into<Error>,
+            Future = impl Send,
+        > + 'static,
+    >
```
hawkw, line 31: is rustfmt okay? does it need help??

linkerd/app/gateway/src/gateway.rs
```diff
@@ -0,0 +1,60 @@
...
+use linkerd2_app_core::proxy::{http, identity};
+use linkerd2_app_core::NameAddr;
+use std::net::SocketAddr;
+
+#[derive(Clone, Debug)]
+pub(crate) enum Gateway<O> {
+    Forbidden,
+    Outbound {
+        // Source-metadata is available via request extensions set by the
+        // inbound, but we
```
hawkw, line 11: looks like this comment was not finished?

linkerd/app/gateway/src/gateway.rs
```diff
@@ -0,0 +1,60 @@
...
+                Box::new(future::ok(rsp))
+            }
+
+            Self::Outbound {
+                outbound,
+                //source_identity,
+                ..
+            } => {
+                // let headers = request.headers_mut();
+                // headers.get_all(http::header::FORWARDED)
```
hawkw, line 55: ?

linkerd/app/gateway/src/make.rs
```diff
@@ -0,0 +1,110 @@
...
+use futures::{future, Future, Poll};
+use linkerd2_app_core::proxy::api_resolve::Metadata;
+use linkerd2_app_core::{dns, transport::tls, Error, NameAddr};
+use linkerd2_app_inbound::endpoint as inbound;
+use linkerd2_app_outbound::endpoint as outbound;
+use std::net::IpAddr;
+use std::sync::Arc;
+
+#[derive(Clone, Debug)]
+pub(crate) struct GatewayMake<R, O> {
```
hawkw, line 11: very nitpicky, feel free to ignore: it seems like we normally call `MakeService` impls `MakeFoo` rather than `FooMake`? should this be made consistent?

linkerd/app/gateway/src/gateway.rs
```diff
@@ -0,0 +1,60 @@
...
+            Self::Outbound { outbound, .. } => outbound.poll_ready(),
+        }
+    }
+
+    fn call(&mut self, request: http::Request<http::glue::HttpBody>) -> Self::Future {
+        match self {
+            Self::Forbidden => {
+                let rsp = http::Response::builder()
+                    .status(http::StatusCode::FORBIDDEN)
+                    .body(Default::default())
```
hawkw, line 44: it _might_ be nice to include a string constant explaining why the request is forbidden, in case the user wasn't looking at the proxy's logs? not sure if it matters.

linkerd/app/outbound/src/lib.rs
```diff
@@ -466,20 +468,25 @@ impl Config {
...
-                .push_on_response(http_admit_request)
-                .push_on_response(metrics.stack.layer(stack_labels("source")))
-                .instrument(
-                    |src: &tls::accept::Meta| {
-                        info_span!("source", target.addr = %src.addrs.target_addr())
-                    },
-                )
-                .check_new_service::<tls::accept::Meta>();
+            // Resolve the application-emitted destination via DNS to determine
+            // it's canonical FQDN to use for routing.
```
hawkw, line 472: nit: "it's" == "it is"; i think this should be
```suggestion
            // its canonical FQDN to use for routing.
```

linkerd/app/outbound/src/lib.rs
```diff
@@ -382,35 +386,33 @@ impl Config {
                 is_discovery_rejected,
             )
             .check_service::<Logical<HttpEndpoint>>()
-            // Sets the canonical-dst header on all outbound requests.
-            .push(http::header_from_target::layer(CANONICAL_DST_HEADER))
```
hawkw, line 386: is there a reason the `header_from_target` layer moved?

linkerd/app/src/env.rs
```diff
@@ -121,6 +121,11 @@ pub const ENV_DESTINATION_GET_NETWORKS: &str = "LINKERD2_PROXY_DESTINATION_GET_N
 /// If unspecified, a default value is used.
 pub const ENV_DESTINATION_PROFILE_SUFFIXES: &str = "LINKERD2_PROXY_DESTINATION_PROFILE_SUFFIXES";
 
+/// Constrains which destination names are permitted.
+///
+/// If unspecified or empty, no inbound gateway s configured.
```
hawkw, line 126: nit:
```suggestion
/// If unspecified or empty, no inbound gateways are configured.
```

linkerd/app/src/lib.rs
```diff
@@ -179,8 +181,11 @@ impl Config {
                     .build_server(
                         outbound_listen,
                         outbound_addr.port(),
+                        svc::stack(refine.clone())
+                            .push_map_response(|(n, _)| n)
```
hawkw, line 185: minor nit, probably not important: it seems like it _could_ be good to avoid doing all the stuff with the vec and so on in cases where we only care about the refined name? but, i don't think it's that costly.

linkerd/app/gateway/src/gateway.rs
olix0r: this is for the followup PR... just a note to myself

hawkw: ah, gotcha, carry on then!

linkerd/app/src/lib.rs
olix0r: Yeah, agreed.. I think it's fine for now.

olix0r:
> it seems like we really ought to have integration tests for this, at least in the simple case without identity. is there a reason this isn't practical to write integration tests in linkerd2-app-integration

No, but the failure case isn't particularly interesting to test with integrations (because we can only exercise the identity-based failure). To do real testing we need to mock identity or stub the transports entirely.

linkerd/app/gateway/src/gateway.rs
olix0r: Because these ports are expected to be exposed to the internet, I don't want to leak any information on responses.

PR #543, opened 2020-05-29
update opencensus to tonic/std::future

hawkw:
This branch updates the proxy's OpenCensus tracing code and collector
client to use Tonic and `std::future`, and puts the OpenCensus stack
layers back in the inbound and outbound proxy.

Of note: the previous implementation constructed a single client
`Service` which was reused when restarting a new RPC if the stream
terminated. Because Tonic's generated RPCs borrow the client service
mutably for the duration of the RPC future, if we want these futures to
be `'static`, we must move the service into the future. Rather than
requiring the service to be `Clone`, I solved this by replacing the
`Service` instance held by the client with a `NewService` instance. I
don't *think* this should be an issue, let me know if I'm wrong?

I'd like to do some additional testing with the OpenCensus integration
tests but I'm opening this PR now to start getting reviews.

Depends on #541.


linkerd/app/src/oc_collector.rs
```diff
@@ -68,7 +69,7 @@ impl Config {
                     }))
                     .push(control::add_origin::Layer::new())
                     .into_new_service()
-                    .new_service(addr.clone());
+                    .with_fixed_target(addr.clone());
```
hawkw, line 72: This seemed like a better approach to me than making `linkerd2-opencensus` depend on `linkerd2-app-core` just to get the `ControlAddr` type, wdyt?

linkerd/opencensus/src/lib.rs
```diff
@@ -1,59 +1,77 @@
...
+enum State {
     Idle,
     Sending {
         sender: mpsc::Sender<ExportTraceServiceRequest>,
         // Node data should only be sent on the first message of a streaming
         // request.
         node: Option<Node>,
         // We hold the response future, but never poll it.
-        _rsp: ResponseFuture<ExportTraceServiceResponse, T::Future>,
+        _rsp: Pin<
```
hawkw, line 49: Lucio Franco mentioned to me that not polling this future _may_ result in the h2 connection window filling up, so I may look into changing this code so that it is polled (even though we don't care about the result).

PR #544, opened 2020-05-30
gateway: Add a Forwarded header

olix0r:
When the gateway forwards requests, it now adds a [`Forwarded`][mdn] header
including the source identity, the local identity, and the destination
authority.

---

I've tested this manually. I'll follow up with tests as we port this to tokio-0.2 (where the test infra is much friendlier).

Also, the gateway, in general, seems to incur a pretty heavy compile-time hit... we'll have to revisit that post-release as well.

[mdn]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded

### DIFF
--- a/linkerd/app/gateway/src/make.rs
+++ b/linkerd/app/gateway/src/make.rs
@@ -1,6 +1,7 @@
 use super::gateway::Gateway;
 use futures::{future, Future, Poll};
 use linkerd2_app_core::proxy::api_resolve::Metadata;
+use linkerd2_app_core::proxy::identity;
 use linkerd2_app_core::{dns, transport::tls, Error, NameAddr};
 use linkerd2_app_inbound::endpoint as inbound;
 use linkerd2_app_outbound::endpoint as outbound;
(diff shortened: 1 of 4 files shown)


linkerd/app/gateway/src/make.rs
```diff
@@ -64,46 +87,65 @@ where
...
+            } => {
+                let mut outbound = outbound.clone();
+                let local_identity = local_identity.clone();
+                let suffixes = suffixes.clone();
+                Box::new(
+                    // First, resolve the original name. This determines both the
+                    // canonical name as well as an IP that can be used as the outbound
+                    // original dst.
+                    resolve
+                        .call(orig_dst.name().clone())
```
hawkw, line 111: should we stick a trace span on this for the original destinaton name, or is that added somewhere else in the stack?

olix0r: Yep, we get that on the stack ;)

https://github.com/linkerd/linkerd2-proxy/blob/9a9fa33718dc3c612ac8108420d55d34e83e3816/linkerd/app/outbound/src/lib.rs#L118

PR #545, opened 2020-05-30
Fail requests that loop through the gateway

olix0r:
This change uses the gateway's `Forwarded` header to detect if the
request has already transited through this gateway. This is
determination is made by comparing ID strings, so this will prevent
gateway daisy-chaining when clusters do not use distinct identity
domains.

---

I've tested this manually: tapping the gateway's client, we see responses like:
```json
{
  "source": {
    "ip": "10.42.0.25",
    "port": 54748,
    "metadata": {
      "control_plane_ns": "linkerd",
      "deployment": "web",
      "namespace": "emojivoto",
      "pod": "web-9c9cc6d7d-9wn9d",
      "pod_template_hash": "9c9cc6d7d",
      "serviceaccount": "web",
      "tls": "loopback"
    }
  },
  "destination": {
    "ip": "172.24.0.4",
    "port": 4143,
    "metadata": {
      "namespace": "emojivoto",
      "remote_cluster": "east",
      "remote_gateway": "linkerd-gateway",
      "remote_gateway_namespace": "linkerd-multicluster",
      "remote_service": "voting-svc",
      "remote_service_namespace": "emojivoto",
      "server_id": "linkerd-gateway.linkerd-multicluster.serviceaccount.identity.linkerd.east.k3d.example.com",
      "service": "voting-svc-east",
      "tls": "true"
    }
  },
  "routeMeta": null,
  "proxyDirection": "OUTBOUND",
  "responseInitEvent": {
    "id": {
      "base": 1,
      "stream": 23
    },
    "sinceRequestInit": {
      "nanos": 1886209
    },
    "httpStatus": 508,
    "headers": [
      {
        "name": ":status",
        "valueStr": "508"
      },
      {
        "name": "date",
        "valueStr": "Sat, 30 May 2020 19:02:03 GMT"
      }
    ]
  }
}
```

### DIFF
--- a/linkerd/app/gateway/src/gateway.rs
+++ b/linkerd/app/gateway/src/gateway.rs
@@ -10,8 +10,12 @@ pub(crate) enum Gateway<O> {
     Outbound {
         outbound: O,
         forwarded_header: http::header::HeaderValue,
+        local_identity: identity::Name,
     },
 }
+
+type ResponseFuture<E> =
+    Box<dyn Future<Item = http::Response<http::boxed::Payload>, Error = E> + Send + 'static>;
 
 impl<O> Gateway<O> {
     pub fn new(
@@ -26,6 +30,7 @@ impl<O> Gateway<O> {
         );
         Gateway::Outbound {
             outbound,
+            local_identity,
             forwarded_header: http::header::HeaderValue::from_str(&fwd)
                 .expect("Forwarded header value must be valid"),
         }
@@ -41,7 +46,7 @@ where
 {
     type Response = O::Response;
     type Error = O::Error;
-    type Future = Box<dyn Future<Item = Self::Response, Error = Self::Error> + Send + 'static>;
+    type Future = ResponseFuture<Self::Error>;
 
     fn poll_ready(&mut self) -> Poll<(), Self::Error> {
         match self {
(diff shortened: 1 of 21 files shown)


linkerd/app/gateway/src/gateway.rs
```diff
@@ -54,11 +59,30 @@ where
...
+                for fwd in request
+                    .headers()
+                    .get_all(http::header::FORWARDED)
+                    .into_iter()
+                    .filter_map(|h| h.to_str().ok())
+                {
+                    if let Some(by) = fwd_by(fwd) {
+                        if by == local_identity.as_ref() {
+                            tracing::info!(fwd, "Loop detected");
+                            return rsp(http::StatusCode::LOOP_DETECTED);
```
hawkw, line 76: my understanding is that 508 Loop Detected is specific to WebDAV ([per MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/508)). the *name* of the status is definitely correct in this case, but the documented semantics of the code seem to mean something entirely different. is it correct to overload the meaning of that status here, or should we use a different error code?

linkerd/app/gateway/src/gateway.rs
```diff
@@ -54,11 +59,30 @@ where
...
+                // transited through this gateway.
+                for fwd in request
+                    .headers()
+                    .get_all(http::header::FORWARDED)
+                    .into_iter()
+                    .filter_map(|h| h.to_str().ok())
+                {
+                    if let Some(by) = fwd_by(fwd) {
+                        if by == local_identity.as_ref() {
+                            tracing::info!(fwd, "Loop detected");
```
hawkw, line 75: should this be a warning? also, do we want to consider eventually adding a metric label or something so that the responses that fail due to looping are counted separately? of course, this can be a follow up; just a thought.

linkerd/app/gateway/src/gateway.rs
olix0r: Good point!

I think that leaves us with 502. I'll do that and upgrade the info to a warn.

linkerd/app/gateway/src/gateway.rs
olix0r: Alternatively, we could throw an error here and let the error-handler take the wheel...

olix0r: https://github.com/linkerd/linkerd2-proxy/compare/ver/error-status?expand=1

linkerd/app/gateway/src/gateway.rs
olix0r: Leaving this as 508 for now, as it's more descriptive to users. I'm less worried about the WebDAV-origin, because it's semantically correct and we're not violating any part of the RFC.

PR #546, opened 2020-05-30
Port gateway changes to tokio-0.2

olix0r:
Rebases master onto master-tokio-0.2


hawkw:
looks like this breaks several of the retry integration tests?

linkerd/app/gateway/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "linkerd2-app-gateway"
+version = "0.1.0"
+authors = ["Linkerd Developers <cncf-linkerd-dev@lists.cncf.io>"]
+edition = "2018"
+publish = false
+
+[dependencies]
+http = "0.1"
```
hawkw, line 9: Why does this depend on version 0.1 of `http`? The proxy should be using `http` 0.2 more or less everywhere else.

linkerd/app/gateway/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+            let target = inbound::Target {
+                addr: SocketAddr::from(([127, 0, 0, 1], 4143)),
+                dst_name: dst_name.map(|n| NameAddr::from_str(n).unwrap()),
+                http_settings: http::Settings::Http2,
+                tls_client_id: peer_id,
+            };
+            assert_ready_ok!(make_gateway.poll_ready());
+            let gateway = make_gateway.call(target).await.unwrap();
+
+            tokio::spawn(async move {
```
hawkw, line 159: if the spawned task panics, we should probably fail the test, right? might be good to hang onto its `JoinHandle` and await & unwrap it after we await the gateway request, just to make sure.

linkerd/app/integration/src/lib.rs
```diff
@@ -1,7 +1,9 @@
 //! Shared infrastructure for integration tests
 
 #![deny(warnings, rust_2018_idioms)]
-#![type_length_limit = "1586225"]
+#![recursion_limit = "256"]
+#![type_length_limit = "15882953"]
```
hawkw, line 5: oh boy!

linkerd/app/outbound/src/lib.rs
```diff
@@ -56,400 +59,481 @@ pub struct Config {
...
+            .push_failfast(dispatch_timeout)
+            //.push(metrics.http_errors.clone())
+            // Synthesizes responses for proxy errors.
+            .push(errors::layer())
+            // Initiates OpenCensus tracing.
+            .push(TraceContextLayer::new(span_sink.clone().map(|span_sink| {
+                SpanConverter::server(span_sink, trace_labels())
+            })));
+        // // Tracks proxy handletime.
+        // //.push(metrics.clone().http_handle_time.layer());
```
hawkw, line 484: Handle time has already been updated to 0.2 (in #542), so I don't thinl this needs to be commented out?

linkerd/app/gateway/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+publish = false
+
+[dependencies]
+http = "0.2"
+futures = "0.3"
+indexmap = "1.0"
+linkerd2-app-core = { path = "../core" }
+linkerd2-app-inbound = { path = "../inbound" }
+linkerd2-app-outbound = { path = "../outbound" }
+tokio = "0.2"
```
hawkw, line 15: Let's make sure that this crate enables all the tokio feature flags it depends on. Otherwise, this crate won't compile if you cd into its directory and run `cargo build` --- it compiles when built from the workspace since other crates in the workspace enable its' features.

At a glance, it looks like the features this needs are:
```suggestion
tokio = { version = "0.2", features = ["rt-core", "macros"] }
```

Also, it looks like the Tokio dependency is only used for tests, so maybe it should be a dev-dep?

PR #547, opened 2020-06-01
gateway: Return errors instead of responses

olix0r:
This ensures that error metrics are recorded and that logging is emitted
uniformly. This also ensures that gRPC requests don't get HTTP error
responses.

### DIFF
--- a/linkerd/app/core/src/errors.rs
+++ b/linkerd/app/core/src/errors.rs
@@ -386,7 +386,7 @@ impl HttpError {
     pub fn identity_required(message: &'static str) -> Self {
         Self {
             message,
-            http: http::StatusCode::PROXY_AUTHENTICATION_REQUIRED,
+            http: http::StatusCode::FORBIDDEN,
             grpc: Code::Unauthenticated,
             reason: Reason::IdentityRequired,
         }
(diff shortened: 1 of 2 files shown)


olix0r:
Note, I've also opted for slightly different error codes here, using PROXY_AUTHENTICATION_REQUIRED when identity isn't present and NOT_FOUND otherwise.

linkerd/app/core/src/errors.rs
```diff
@@ -363,3 +381,31 @@ impl Metrics {
         self.0.clone()
     }
 }
+
+impl HttpError {
+    pub fn identity_required(message: &'static str) -> Self {
+        Self {
+            message,
+            http: http::StatusCode::PROXY_AUTHENTICATION_REQUIRED,
```
hawkw, line 389: MDN suggests that we should return a [`proxy-authenticate` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authenticate) along with this status code, is this something we want to do?

olix0r: We're not actually participating in http authentication, so i'm not going to try to pretend we are. But I think the descriptive error numbers are helpful. With this in mind, I'm inclined to continue using 508 in the other PR (since it's descriptive for users, even if we're repurposing error codes outside their original uses).

olix0r: Hm, actually, the RFC uses a _MUST_ here. I guess we should revert to using FORBIDDEN...

PR #549, opened 2020-06-03
inbound: Do not cache gateway services

olix0r:
When the inbound caches gateway services, it eagerly obtains an
outbound service to cache. If the outbound service employs a traffic
split, this inbound service is pinned to a specific leaf, and requests
will never be routed to the other leaf.

This change moves the gateway fallback to be outside all of the inbound
caches, so that outbound splits work as intended.

### DIFF
--- a/linkerd/app/inbound/src/lib.rs
+++ b/linkerd/app/inbound/src/lib.rs
@@ -276,6 +276,10 @@ impl Config {
                 http_target_cache
                     .push_on_response(svc::layers().box_http_response().box_http_request()),
             )
+            // If the traffic is targeted at the inbound port, send it through
+            // the loopback service (i.e. as a gateway). This is done before
+            // caching so that the loopback stack can determine whether it
+            // should cache or not.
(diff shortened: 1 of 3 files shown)


linkerd/app/inbound/src/lib.rs
```diff
@@ -279,13 +272,16 @@ impl Config {
...
             .push_make_ready()
-            // Don't resolve profiles for inbound-targetted requests. They'll be
-            // resolved on the outbound side if necsesary.
-            .push(admit::AdmitLayer::new(prevent_loop))
             .push_fallback(
                 http_target_cache
                     .push_on_response(svc::layers().box_http_response().box_http_request()),
             )
+            .push(admit::AdmitLayer::new(prevent_loop))
+            .push_fallback_on_error::<prevent_loop::LoopPrevented, _>(
```
hawkw, line 280: May be worth a comment explaining that this has to be outside the cache? Not a blocker.

PR #558, opened 2020-06-11
Add a CODEOWNERS

olix0r:
This CODEOWNERS change will ensure that proxy-maintainers are requested on all PRs.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,1 @@
-# Require a review from at least Eliza or Oliver, hopefully both.
 * @linkerd/proxy-maintainers
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
 # Require a review from at least Eliza or Oliver, hopefully both.
-* @hawkw @olix0r
+* @linkerd/proxy-maintainers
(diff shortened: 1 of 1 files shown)


hawkw:
> This CODEOWNERS change will enforce a requirement that I approve all
> PRs to master.

IIRC, CODEOWNERS, on its own, doesn't actually _require_ a review from code owners to merge a PR; it just adds you as a reviewer automatically. I _think_ there's also a repo setting to "require a review from CODEOWNERS" that needs to be flipped.

> This is already practically the case, but it's not
> codified.

This is basically true, yeah. I feel like I _also_ review pretty much all proxy changes, but I don't feel like my review needs to be blocking, unlike yours.

My only concern about doing this for now is bus factor --- if you're OOO for a week, we can't merge anything during that time. I don't think we would merge any major functional changes even without Github enforcing that your approval is required, and minor changes like spelling fixes or whatever could just wait until you got back, but it could be good to be able to get a hypothetical fix for a major issue through without having to track you down in the woods or whatever?

.github/CODEOWNERS
```diff
@@ -1,2 +1,2 @@
-# All changes require an ack from Oliver for now.
-*	@olix0r
+# Require a review from at least Eliza or Oliver, hopefully both.
+* @hawkw @olix0r
```
hawkw, line 2: :heart:

PR #562, opened 2020-06-13
Remove tokio-compat deps

olix0r:
This change drops (unused, defunct) dependencies on the futures-0.1/compat ecosystem.

It does not yet adopt the new runtime due to issues spotted in testing.


linkerd/app/src/lib.rs
```diff
@@ -339,31 +338,31 @@ impl App {
...
 
                         // Spawn the DNS resolver background task.
-                        tokio_02::spawn(dns.instrument(info_span!("dns")));
+                        tokio::spawn(dns.instrument(info_span!("dns")));
 
                         if let tap::Tap::Enabled { daemon, serve, .. } = tap {
-                            tokio_02::spawn(daemon.instrument(info_span!("tap")));
-                            tokio_02::task::spawn(
+                            tokio::spawn(daemon.instrument(info_span!("tap")));
+                            tokio::task::spawn(
```
hawkw, line 345: nit: should probably be 
```suggestion
                            tokio::spawn(
```
for consistency (I think this was a `spawn_local` at one point or something)

linkerd2-proxy/src/main.rs
```diff
@@ -21,9 +21,9 @@ fn main() {
         }
     };
 
-    tokio_compat::runtime::current_thread::Runtime::new()
+    tokio::runtime::Runtime::new()
         .expect("main runtime")
-        .block_on_std(async move {
+        .block_on(async move {
```
hawkw, line 26: TIOLI: what do you think about just sticking a `#[tokio::main(basic_scheduler)]` on `main` and turning the whole thing into an `async fn`, instead? that's what I was planning on doing...

hawkw:
Looks like there is still a use of `tokio-compat` in `linkerd2-app-core::admin`: https://github.com/linkerd/linkerd2-proxy/pull/562/checks?check_run_id=768417300#step:5:351

linkerd2-proxy/src/main.rs
```diff
@@ -21,9 +21,9 @@ fn main() {
         }
     };
 
-    tokio_compat::runtime::current_thread::Runtime::new()
+    tokio::runtime::Runtime::new()
```
olix0r, line 24: It seems that this introduces a regression that i can catch via the benchmark suite: the proxy thread eventually deadlocks such that no new connections are accepted. the admin thread continues to operate normally.

hawkw: Hmm...that's pretty weird. I can look into it.

olix0r: It turns out that I must have accidentally been testing the threaded runtime. I don't see anything like this when i explicitly set the basic scheduler, so i'll add that here.

hawkw: Ahh, okay, that explains the deadlock...I am sure there are plenty of mutexen in e.g. metrics that only work with the assumption that they're not contended by multiple threads (or that rely on the write lock not being contended or something).

PR #565, opened 2020-06-15
remove & deny warnings; cleanup

hawkw:
This PR removes a few remaining compiler warnings from
`master-tokio-0.2` and re-adds `deny` attributes for warnings. Also,
I've updated all crates to consistently use `tracing-futures` v0.2
(previously, some crates were on 0.2 while others were still on 0.1).


olix0r:
@hawkw looks like your cargo.lock is out of sync

PR #574, opened 2020-06-18
Update Rust to 1.44.0

olix0r:
See https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1440-2020-06-04


hawkw:
Can we go ahead and bump to 1.44.1 instead? It's a patch that [fixes a few tooling regressions](https://github.com/rust-lang/rust/blob/stable/RELEASES.md#version-1441-2020-06-18).

olix0r:
@hawkw It takes a few days for those images to show up: https://hub.docker.com/_/rust

hawkw:
I think the SIGSEGV when finding backtraces on macOS is the only issue here that effects us, and it's just an annoyance for devs running tests on macOS, not a showstopper.

PR #575, opened 2020-06-18
Use async/await to simplify connection-accept task

olix0r:
`app::core::serve` provides a task that drives connections from a
listener to an accept stack until a drain is signaled, at which point
the task completes and no further connections are accepted.

The implementation for this can be simplified by using newer async/await
syntax.


linkerd/app/core/src/proxy/server.rs
```diff
@@ -192,35 +192,48 @@ where
...
+                            .serve_connection(io, HyperServerSvc::new(svc))
+                            .with_upgrades();
+
+                        tokio::select! {
+                            res = &mut conn => {
+                                res?
+                            }
+                            () = drain.into_future() => {
+                                Pin::new(&mut conn).graceful_shutdown()
+                            }
```
hawkw: I don't think this has the same behavior as the way we were using `Drain::watch`  previously. `watch` not only triggers the graceful shutdown but also causes the drained future to wait until the connection's graceful shutdown has _completed_. On the other hand, I think this just triggers the graceful shutdown without counting it against the drain's completion.

I believe that this may result in the proxy exiting before it has performed a graceful shutdown on all connections,

linkerd/app/core/src/serve.rs
```diff
@@ -32,71 +29,22 @@ where
...
-            State::Cancelled => Poll::Ready(Ok(())),
+    // Drive connections from the listener into the accept stack until a drain
+    // is signaled.
+    tokio::select! {
+        res = listen.serve(accept) => {
+            // The server can error but it never completes on its own.
+            match res? {}
+        }
+        () = drain.into_future() => {
+            Ok(())
```
hawkw, line 47: As in my comment above, I don't think this is right.

linkerd/app/core/src/proxy/server.rs
```diff
@@ -223,17 +220,14 @@ where
...
 
                         tokio::select! {
-                            res = &mut conn => {
-                                res?
-                            }
+                            res = &mut conn => { res }
                             () = drain.into_future() => {
-                                Pin::new(&mut conn).graceful_shutdown()
+                                Pin::new(&mut conn).graceful_shutdown();
+                                conn.await
```
hawkw, line 226: I don't think this is gonna cut it, either. We still need the future _triggering_ the graceful shutdown (which is in the runtime's `block_on` future) to wait for all the graceful shutdowns to complete, which is what `drain::Sender::drained()` did. Not waiting in `block_on` will mean the runtime can exit before the drain has completed.

hawkw:
Is this just flaky, or is it a too-early shutdown in some case? https://github.com/linkerd/linkerd2-proxy/pull/575/checks?check_run_id=786204633#step:5:499

olix0r:
@hawkw pretty sure it's flakiness but we'll see...
\

linkerd/drain/src/lib.rs
```diff
@@ -96,6 +96,15 @@ impl Watch {
...
     }
 
+    pub async fn and_drop<A, F>(self, future: A, on_drain: F) -> A::Output
+    where
+        A: Future + Unpin,
+        F: FnOnce() -> A::Output,
+    {
+        self.watch(future, |_| {}).await;
+        on_drain()
+    }
```
olix0r, line 106: Eh, but this is weird because it potentially drops the return value of `future`... I can fix it..

linkerd/drain/src/lib.rs
```diff
@@ -88,14 +92,19 @@ impl Watch {
...
+            res = &mut future => res,
+            _handle = self.handle() => {
                 on_drain(&mut future);
-                (&mut future).await
+                future.await
             }
         }
     }
 
+    pub async fn handle(self) -> Handle {
```
hawkw, line 103: TIOLI: would be nice to have a comment explaining what this does exactly, because we now have three options for consuming the `Watch`.

linkerd/app/core/src/proxy/server.rs
```diff
@@ -177,7 +177,12 @@ where
                     .oneshot((proto.tls, io));
                 let fwd = async move {
                     let conn = accept.await.map_err(Into::into)?;
-                    Ok(Box::pin(drain.after(conn).map_err(Into::into)) as Self::Response)
+                    Ok(Box::pin(
+                        drain
+                            .ignore_signal()
+                            .release_after(conn)
```
hawkw, line 183: :+1: for `release_after`

PR #577, opened 2020-06-22
Simplify protocol detection with async/await

olix0r:
The protocol detection middleware is pretty verbose, including a manual
future implementation that manages a complex set of state types.

This can all be simplified with async/await (at the cost of ~2 boxes per connection).

### DIFF
--- a/linkerd/app/core/src/proxy/server.rs
+++ b/linkerd/app/core/src/proxy/server.rs
@@ -1,18 +1,22 @@
+use crate::proxy::http::{
+    glue::{Body, HyperServerSvc},
+    h2::Settings as H2Settings,
+    trace, upgrade, Version as HttpVersion,
+};
+use crate::transport::{
+    self,
+    io::{self, BoxedIo, Peekable},
+    labels::Key as TransportKey,
+    metrics::TransportLabels,
+    tls,
+};
 use crate::{
     drain,
-    proxy::{
-        core::Accept,
-        detect,
-        http::{
-            glue::{Body, HyperServerSvc},
-            h2::Settings as H2Settings,
-            trace, upgrade, Version as HttpVersion,
-        },
-    },
+    proxy::{core::Accept, detect},
     svc::{NewService, Service, ServiceExt},
-    transport::{self, io::BoxedIo, labels::Key as TransportKey, metrics::TransportLabels, tls},
     Error,
 };
+use async_trait::async_trait;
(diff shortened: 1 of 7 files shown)


linkerd/app/core/src/proxy/server.rs
```diff
@@ -34,35 +38,43 @@ pub type Connection = (Protocol, BoxedIo);
...
 
-        Err(tls)
-    }
-
-    fn detect_peeked_prefix(&self, tls: tls::accept::Meta, prefix: &[u8]) -> Self::Target {
-        Protocol {
-            tls,
-            http: HttpVersion::from_prefix(prefix),
-        }
+        // Otherwise, attempt to peek the client connection to determine the protocol.
```
hawkw, line 72: nit, take it or leave it: it's a _little_ weird to me that this comment starts with "otherwise" when there's no preceding comment. clearly it's intended to be read as "if self.skip_ports.contains(&port), ..., otherwise, ...", but the "otherwise" made me look for another comment before i figured that out. this probably definitely doesn't actually matter.

linkerd/proxy/detect/src/lib.rs
```diff
@@ -83,85 +40,40 @@ impl<D: Clone, A> tower::layer::Layer<A> for DetectProtocolLayer<D> {
...
     type Response = A::ConnectionFuture;
     type Error = Error;
-    type Future = AcceptFuture<T, D, A>;
+    type Future = Pin<Box<dyn Future<Output = Result<A::ConnectionFuture, Error>> + Send>>;
 
-    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        self.accept.poll_ready(cx).map_err(Into::into)
+    fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        // The `accept` is cloned into the response future, so its readiness isn't important.
+        Poll::Ready(Ok(()))
```
hawkw, line 62: this does seem like a slight change in behavior, but i don't think it's one that actually matters. is there a reason we changed this? it seems like we could continue driving it here.

olix0r: I think it was technically a bug before. We would poll the accept to ready, and then clone it, and then not poll the clone to ready (which is technically required).

I think, practically, this stack doesn't use readiness and there's a oneshot inside of it.

PR #580, opened 2020-06-23
update test-support clients and servers to be natively async

hawkw:
Currently, the test clients and servers in `linkerd2-app-integration`
each run their own separate Tokio runtime on a background thread. This
was done in order to allow writing the test code in a synchronous
fashion, rather than using futures. Now, however, a majority of the
tests already use async-await syntax, so we can run all this stuff on
the same test runtime. This should reduce the number of threads in the
tests significantly, which may decrease flakiness and possibly improve
test performance a bit.

This branch updates the test support servers and clients to run as tasks
spawned on the main test runtime. It does not touch the test control
plane (Destination and Identity services) as they still need to be
updated to use Tonic and `std::future`.

This does introduce one potential footgun that is worth noting: if a
test server that makes assertions is spawned in the background, the
`JoinHandle` for its task must be awaited. Otherwise, panics will not be
propagated to the main test task, and the test can spuriously pass if
the assertion fails.


linkerd/app/integration/src/lib.rs
```diff
@@ -330,3 +343,58 @@ impl<T: fmt::Debug, E: fmt::Debug> ResultWaitedExt for Result<T, Waited<E>> {
         }
     }
 }
+
+pub async fn cancelable<E: Send + 'static>(
+    drain: drain::Watch,
+    f: impl Future<Output = Result<(), E>> + Send + 'static,
+) -> Result<(), E> {
+    drain.watch(Cancelable::new(f), |f| f.cancel()).await
```
olix0r, line 351: I think you could eliminate the Cancelable boilerplate with this:

```suggestion
    tokio::select! {
      res = &mut f => { res }
      shutdown = drain.signal() => {
        trace!("canceling...");
        let res = shutdown.release_after(f);
        trace!("canceled");
        res
      }
    };
```

olix0r: And actually, if all you want to do is drop the inner future, it can be even simpler...

hawkw: hmm, yeah, I think so (we _do_ need the `release_after` part here) --- this code predates the latest `drain` additions on main.

olix0r: It looks like Cancelable just drops the inner future at the moment? Unless I'm misreading it.

hawkw: ah, yeah, it was the code in your suggestion that _i_ misread :P

PR #582, opened 2020-06-26
Simplify tap server

kleimkuhler:
## Description

Remove the Tap daemon that is responsible for handling service registrations and
tap subscriptions.

There is now a registry that contains all active taps.

The tap service has access to this registry and it uses it to retrieve an
updated set of active taps for every request. This set is used to determine
whether the request should be tapped.

New taps are added when the gRPC server receives an observe request. If the
request is valid, the server adds the tap to the registry.

Inactive taps are cleared by a background task. This background task is much
simpler than the one before; it loops on an interval and checks whether each tap
still has an actively used response stream, and if it has not reached its limit.

### Changes

The primary change is the addition of the `Registry` in
`linkerd/proxy/tap/src/registry.rs`.

The tap service holds on to a `Registry` and uses the `taps_recv` field to get
an updated list of taps for each request. The tap service is not responsible for
adding or removing any taps from the registry.

The gRPC service also holds on to a `Registry` and uses the `inner` field to
register new taps with the registry. The field is locked giving access to an
`Inner`. With this, the gRPC server can add new taps to the registry and then
send updates with the `taps_send` sender.

Neither the tap service or the gRPC is responsible for clearing out inactive
taps. To solve this, we spawn a background task with the `async fn clean` method
which loops on an interval and checks that each tap still has an actively used
response stream.

### Traits

The `tap::iface` module now only contains the `Tap`, `TapPayload`, and
`TapResponse` traits. These are helpful for traits that are implemented in
`tap::grpc`.

The `Subscribe` and `Register` traits have been removed. Services no longer
*register* with the tap server. Taps still do subscribe (now called `register`)
with the registry, but it was only implemented by one object and did not
simplify the implementation in terms of understanding or save on duplicated
code.


linkerd/proxy/tap/src/registry.rs
```diff
@@ -0,0 +1,58 @@
...
+    pub async fn clean(self) {
+        loop {
+            if let Ok(mut inner) = self.inner.lock() {
+                let count = inner.taps.len();
+                inner.taps.retain(|tap| tap.can_tap_more());
+                trace!("retained {} of {} taps", inner.taps.len(), count);
+            }
+            time::delay_for(Duration::from_millis(100)).await;
+        }
+    }
```
olix0r, line 57: * Let's pass in an `interval: Duration` rather than hardcode it here.
* This task never completes. We'll need to select on `tx.closed()` -- I think this means that the tx should be stored on the registry and not within the mutex...
* The task should probably start by waiting, rather than cleaning immediately, but that's fairly minor.

olix0r: Better yet, you could pass in a `wakeup: impl Stream<Output = ()>` to completely decouple the decision of when to clean from the task.

olix0r: Nevermind re: shutting this task down.. because the registry holds the receiver we can't know when all instances have been dropped.  If we'd like to support graceful cleanup (which isn't at all required, functionally), we'd have to decouple `clean` from `Registry`...

linkerd/proxy/tap/src/registry.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub fn register(&self, tap: T) {
+        if let Ok(mut inner) = self.inner.lock() {
+            inner.taps.push(tap);
+            let _ = inner.taps_send.broadcast(inner.taps.clone());
+        }
+    }
+
+    pub async fn clean(self, wakeup: impl Stream) {
+        futures::pin_mut!(wakeup);
+        while let Some(_) = wakeup.next().await {
```
hawkw, line 48: :+1:

linkerd/app/src/lib.rs
```diff
@@ -343,8 +343,15 @@ impl App {
...
 
-                        if let tap::Tap::Enabled { daemon, serve, .. } = tap {
-                            tokio::spawn(daemon.instrument(info_span!("tap")));
+                        if let tap::Tap::Enabled {
+                            registry, serve, ..
+                        } = tap
+                        {
+                            tokio::spawn(
+                                registry
+                                    .clean(tokio::time::interval(Duration::from_millis(100)))
```
olix0r, line 352: 100ms is a pretty tight poll for something this that should be super rare, and the cost of a stale instance should be pretty light.

```suggestion
                                    .clean(tokio::time::interval(Duration::from_seconds(60)))
```

PR #583, opened 2020-06-26
Print build diagnostics in docker

olix0r:
This change modifies the development dockerfile to wrap build commands
with `/usr/bin/time -v` so that we get diagnostic information (like the
RSS of the build process).


olix0r:
@hawkw seems useful to capture this in CI so we can spot check it.

PR #584, opened 2020-06-26
Eliminate Bind & Listen traits

olix0r:
We have some abstractions, like the `Bind` and `Listen` traits, that
don't really serve a purpose. Each have only a single implementation in
our codebase. And, furthermor, the `Listen` trait is better-represented
as a `Stream` (so that we can use general tools with it).

This change eliminates these traits and modifies the implementation to
use `async-stream` to take advantage of `async/await` syntax. Similarly,
the logic of `listen::Serve` is now directly implemented in the `serve`
task. The `proxy::cor::listen` module has been replaced by an `accept`
module.

All of this helps us cut down on unnecessary types and helps to
consolidate logic (rather than having it spread across many crates).


linkerd/proxy/transport/src/listen.rs
```diff
@@ -79,68 +63,42 @@ impl<A: OrigDstAddr> Bind<A> {
...
+        let keepalive = self.keepalive;
+        let get_orig = self.orig_dst_addr.clone();
+
+        let accept = try_stream! {
+            // The tokio listener is built lazily so that it is initialized on
+            // the proper runtime.
+            let listen = tokio::net::TcpListener::from_std(listen).expect("listener must be valid");
+
+            futures::pin_mut!(listen);
+            futures::pin_mut!(shutdown);
```
hawkw, line 82: nit/TIOLI: i believe the `tokio::pin!` macro's `let` form (second example [here](https://docs.rs/tokio/0.2.21/tokio/macro.pin.html#examples) might let us make this look a little nicer:
```suggestion
            tokio::pin! {
                // The tokio listener is built lazily so that it is initialized on
                // the proper runtime.
                let listen = tokio::net::TcpListener::from_std(listen).expect("listener must be valid");
                let shutdown = shutdown;
            }
```

linkerd/proxy/transport/src/listen.rs
```diff
@@ -79,68 +63,42 @@ impl<A: OrigDstAddr> Bind<A> {
...
+            loop {
+                let accept = tokio::select! {
```
hawkw, line 84: TIOLI: Is this where the `select!`/`async_stream` incompatibility was worked around? It _might_ be worth documenting that for future generations....

linkerd/app/core/src/serve.rs
```diff
@@ -15,36 +16,47 @@ pub trait HasSpan {
...
-    tokio::select! {
-        err = listen.serve(accept) => {
-            // The listener may fail but it may not complete otherwise.
-            match err? {}
+    futures::pin_mut!(listen);
+    loop {
+        let ready = accept.ready_and().await.map_err(Into::<Error>::into)?;
+        match listen.next().await {
+            None => return Ok(()),
+            Some(conn) => {
```
hawkw, line 43: I may be wrong, but is it possible that driving the accept service to readiness _before_ waiting for the next incoming connection could be problematic, along the same lines as how the buffer worker was causing load balancers to miss discovery updates?

Should we do something more like 
https://github.com/linkerd/linkerd2-proxy/blob/1b3e899efad2ad4ee471c5b49482437e7a325bde/linkerd/buffer/src/dispatch.rs#L47-L51
instead, or is that not an issue here?

olix0r: You're probably right that it's safer to drive it once the connection is received.

linkerd/proxy/transport/src/listen.rs
olix0r: I was previously trying to yield out of the select, which is not supported by async-stream. I don't think that should be documented here, though. It wouldn't help to make this code easier to undertand...

linkerd/app/core/src/serve.rs
hawkw: Yeah, I didn't see anything that looked obviously broken as a result of this, but it's bitten us before...

PR #585, opened 2020-06-27
update test controllers to std::future/Tonic; remove threads

hawkw:
This branch updates the test support mock control plane components in
`linkerd2-app-integration` to use `std::future`, Tonic, and Tokio 0.2.
Rather than spawning a separate thread for each control plane componwnr
as we did previously, they are now spawned as tasks on the main test
thread's runtime. As discussed in #580, this _may_ make the tests
slightly less flaky and/or slightly faster on CI.

In order to make this compile I had to increase the type length and
recursion limits for all the test crates. I have no idea why this branch
made this necessary. My guess is that the fact that we have ~12MB type
names now is at least part of the cause of linkerd/linkerd2#4676.

Depends on #580
Closes linkerd/linkerd2#3963

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -931,7 +931,6 @@ dependencies = [
  "ring",
  "rustls",
  "tokio 0.2.21",
- "tokio-compat",
  "tokio-rustls",
  "tonic",
  "tower",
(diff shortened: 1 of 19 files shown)


olix0r:
@hawkw needs a rebase

hawkw:
> @hawkw needs a rebase

yup, working on it right now!

hawkw:
@olix0r okay, this has been rebased onto `main` and fixed up, it should be reviewable for real now.

PR #586, opened 2020-06-29
buffer: Box the inner service's reponse future

olix0r:
The buffer middleware is parameterized on the inner service's repsonse
future, which "leaks" the inner service's type to the outer components
(which slows down compilations).

This changes the signature to instead be parameterized by the response
type, boxing the inner service's response future to avoid exposing its
type signature.

Relates to linkerd/linkerd2#4676

### DIFF
--- a/linkerd/buffer/src/service.rs
+++ b/linkerd/buffer/src/service.rs
@@ -1,13 +1,11 @@
 use crate::error::{Closed, ServiceError};
 use crate::InFlight;
-use futures::{ready, TryFuture};
 use linkerd2_error::Error;
-use pin_project::{pin_project, project};
 use std::task::{Context, Poll};
 use std::{future::Future, pin::Pin};
 use tokio::sync::{mpsc, oneshot, watch};
 
-pub struct Buffer<Req, F> {
+pub struct Buffer<Req, Rsp> {
     /// Updates with the readiness state of the inner service. This allows the buffer to reliably
     /// exert backpressure and propagate errors, especially before the inner service has been
     /// initialized.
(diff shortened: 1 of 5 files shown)


olix0r:
This appears to shave of ~4mins off of the integration test build in CI.

hawkw:
Are we able to run the performance tests with this change? My assumption is that the added `Box` is not going to have a noticeable impact, but it would be nice to have a measurement regardless --- could serve as a useful data point for similar changes in the future.

olix0r:
@hawkw I'll run some tests.

My hunch is that no single Box is going to influence the overall numbers substantially. If so, my general preference would be to bias for our ability to make changes to the proxy than trying to shave of µs until we know where/why they matter.

hawkw:
@olix0r sure, i totally agree --- i'm just thinking it would be nice for that hunch to be validated by numbers. it's not a blocker for merging this!

linkerd/buffer/src/service.rs
```diff
@@ -61,14 +47,13 @@ impl<Req, F> Buffer<Req, F> {
...
 where
-    F: TryFuture,
-    F::Error: Into<Error>,
+    Rsp: Send + 'static,
 {
-    type Response = F::Ok;
+    type Response = Rsp;
     type Error = Error;
-    type Future = ResponseFuture<F>;
+    type Future = Pin<Box<dyn Future<Output = Result<Rsp, Error>> + Send + 'static>>;
```
hawkw, line 56: i'd be curious to see if switching to nightly and returning `impl Future` resulted in the same build time improvement or not --- but we can do that separately. it would be good to know eventually, since that could help guide future changes...

PR #587, opened 2020-06-29
cache: replace Lock with Buffer

hawkw:
The use of a buffer hides some of the type complexity of the inner
service (all of the type complexity, after #586). However, Linkerd
currently only uses buffers at the `Service` level, not at the
`MakeService` level. Since `MakeService`/`NewService`s tend to be
generic over both inner `MakeService` types _and_ over the produced
`Service`s (as well as potentially two future types), they probably have
much longer concrete types than `Service`s in most cases.

As @olix0r suggested, we can replace the `Lock` that's currently used to
ensure exclusive access to `Cache`s (which are at the `MakeService`)s
with `Buffer`s. The `Lock` is currently essentially doing something
quite similar to adding a `Buffer` anyway. Introducing buffers around
all caches erases the inner type for everything layered around the
cache, which should make overall type length much shorter.

This seems to have a fairly noticeable impact on build time and memory
use (see linkerd/linkerd2#4676). On my machine, running `make docker` on
#586 gets SIGKILLed (presumably by the OOM killer) after 7m53s. After
this change, `make docker` completes successfully after 1m44s.

Also, the `linkerd2-lock` crate can be removed, as it was used only by 
`Cache`.

To work around increased tail latencies when the buffer fills up, this
branch also sets the default buffer capacity to be equal to the 
default concurrency limit, rather than 10. This is fine, since the buffer
capacity isn't _actually_ what enforces a bound on proxy memory use.
The spawned tasks waiting on a full buffer are still sitting on the Tokio
executor, and it's actually the in flight limit that stops us from accepting
any more requests when we have too many in flight.

Depends on #586.
Fixes linkerd/linkerd2#4676.


linkerd/app/inbound/src/lib.rs
```diff
@@ -237,10 +237,7 @@ impl Config {
                         .push(metrics.stack.layer(stack_labels("target"))),
                 ),
             )
-            .instrument(|_: &Target| info_span!("target"))
-            // Prevent the cache's lock from being acquired in poll_ready, ensuring this happens
-            // in the response future. This prevents buffers from holding the cache's lock.
-            .push_oneshot()
```
hawkw, line 243: the oneshots here were only necessary due to implementation details of `Cache` (acquiring the lock in `poll_ready`), so they are now removeable.

linkerd/app/inbound/src/lib.rs
```diff
@@ -268,10 +265,9 @@ impl Config {
                         .push(metrics.stack.layer(stack_labels("profile"))),
                 ),
             )
+            .spawn_buffer(buffer_capacity)
```
hawkw, line 269: i wasn't sure if we wanted to put an idle timeout here or not. i _think_ we only want that at the request level?

olix0r: i think that's right -- because the idle timeout is part of cache eviction. i'll give this a closer look though.

PR #588, opened 2020-07-06
opencensus: drive rsp future to establish conn

hawkw:
In the `futures` 0.1 version of the OpenCensus module, the `tower-grpc`
response future for the client-streaming RPC was never polled, because
we did not need the server response for that RPC. At the time, this was
fine, as `tower-grpc` still established the connection and ensured
messages were sent to the server. However, after updating to `tonic`,
this was no longer the case --- the response future is now responsible
for establishing the connection. This means spans are no longer sent to
the OpenCensus collector.

Due to an issue where the cloud integration tests were silently failing
to run, we missed this regression.

This PR fixes the regression by ensuring that we drive the response
future.

I'd like to rewrite this code to use async-await rather than the manual
state machine, but this change fixes the regression, so I thought I'd go
ahead and open a PR, and refactor in a separate branch.

### DIFF
--- a/linkerd/opencensus/src/lib.rs
+++ b/linkerd/opencensus/src/lib.rs
@@ -1,5 +1,5 @@
 #![deny(warnings, rust_2018_idioms)]
-use futures::{ready, Stream};
+use futures::{ready, FutureExt, Stream};
 use http_body::Body as HttpBody;
 use linkerd2_error::Error;
 use linkerd2_stack::NewService;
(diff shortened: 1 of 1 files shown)


linkerd/opencensus/src/lib.rs
```diff
@@ -178,30 +179,47 @@ where
...
+                    if let Some(mut f) = rsp.take() {
+                        match f.poll_unpin(cx) {
+                            Poll::Ready(Ok(_)) => {}
+                            Poll::Pending => *rsp = Some(f),
+                            Poll::Ready(Err(error)) => {
+                                tracing::debug!(%error, "response future failed, sending a new request");
+                                idle = true;
+                            }
+                        }
+                    }
```
olix0r, line 207: I would nitpick this all more re: readability, but I'd rather just reewrite it with async...

hawkw: yup, i didn't spend a whole lot of time making it nice for that reason --- currently working on the much cleaner rewrite

hawkw:
@olix0r should we merge this? the change to replace this with an async-await-based implementation is _almost_ done but stuck on a weird type error, and i think this might be blocking actually running the deep integration tests on CI again?

PR #589, opened 2020-07-07
enable `parking_lot` locks in dependencies

hawkw:
Tracing and Tokio both expose a feature flagged optional dependency on
the `parking_lot` crate's `Mutex` and `RwLock` implementations rather
than `std`'s. Enabling this dependency causes these crates to use
`parking_lot` internally with no externally visible API change.

In microbenchmarks of Tokio's synchronization primitives, we've observed
better performance with `parking_lot` vs `std`, and tend to recommend it
in high-performance applications where the cost of another dependency is
acceptable.

This branch enables `tokio` and `tracing`'s `parking_lot` features. I
also removed a bonus dependency on a vintage Tokio version that we must
have missed during the 0.2 update, as it was pulling an incompatible
`parking_lot` version. 

There appears to be some performance improvement under load with
`parking_lot` vs the main branch in some cases, although benchmark
results do appear mixed in others (potentially due to noise).
Screenshot_20200707_155947

### DIFF
--- a/linkerd/app/core/Cargo.toml
+++ b/linkerd/app/core/Cargo.toml
@@ -63,7 +63,7 @@ linkerd2-stack-tracing = { path = "../../stack/tracing" }
 linkerd2-trace-context = { path = "../../trace-context" }
 rand = { version = "0.7", features = ["small_rng"] }
 regex = "1.0.0"
-tokio = { version = "0.2", features = ["macros", "sync"]}
+tokio = { version = "0.2", features = ["macros", "sync", "parking_lot"]}
 tokio-timer = "0.2"
 tower-request-modifier = { git = "https://github.com/tower-rs/tower-http" }
 tonic = { version = "0.2", default-features = false, features = ["prost"] }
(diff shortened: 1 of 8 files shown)


linkerd/app/core/Cargo.toml
```diff
@@ -63,7 +63,7 @@ linkerd2-stack-tracing = { path = "../../stack/tracing" }
 linkerd2-trace-context = { path = "../../trace-context" }
 rand = { version = "0.7", features = ["small_rng"] }
 regex = "1.0.0"
-tokio = { version = "0.2", features = ["macros", "sync"]}
+tokio = { version = "0.2", features = ["macros", "sync", "parking_lot"]}
```
olix0r, line 66: Do we need to set this is in all places that `sync` is used?

hawkw: Nope --- if one crate enables the feature (and in this case, `linkerd2_app_core` is always in the tree when building the actual or profiling proxies) the feature is enabled globally. This behavior of cargo features is kind of a blessing and a curse. :)

It would only matter if we needed it to be enabled when building individual crates on their own --- in this case, we don't, because those crates will still _compile_ without `parking_lot`, and rust-analyzer is the only real reason we build individual crates separately from the proxy.

PR #592, opened 2020-07-08
Add compile-time support for mimalloc

olix0r:
Microsoft's `mimalloc` is a general purpose allocator with excellent
performance characteristics and, furthermore, it implements defensive
security features to protect against heap vulnerabilities.  See
https://github.com/microsoft/mimalloc.
    
In initial tests, it appears that it may improve latency at the cost of
~20% memory overhead.
    
This change primarily enables further testing, since it doesn't seem
worth taking the memory overhead at this point.

Some other cleanups to `main.rs` are included in a distinct commit.


linkerd2-proxy/src/main.rs
```diff
@@ -8,9 +8,20 @@ use linkerd2_app::{trace, Config};
...
 
+#[cfg(feature = "mimalloc")]
+#[global_allocator]
+static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
+
+
 #[tokio::main(basic_scheduler)]
 async fn main() {
-    let trace = trace::init();
+    let trace = match trace::init() {
```
hawkw, line 18: This looks unrelated?

PR #594, opened 2020-07-10
Loadbalance requests to control plane

zaharidichev:
This PR enables the discovery of control plane components through DNS. This is done by implementing a resolution stream backed by a DNS resolver. The stream will yield `Update`s  that feed the loadbalancer.

The stream is written with `async-stream` to avoid manual state machines. In addition to that a `dns::Resolver` trait is introduced to allow for easily mocking a DNS resolver and testing the stream implementation.

I have tested this and can confirm that it resolves the problems described in https://github.com/linkerd/linkerd2/issues/4674 and https://github.com/linkerd/linkerd2/issues/4624


linkerd/app/src/identity.rs
```diff
@@ -55,7 +69,8 @@ impl Config {
                     )))
                     .push_timeout(control.connect.timeout)
                     .push(control::client::layer())
-                    .push(control::resolve::layer(dns))
+                    .push(discover)
```
zaharidichev, line 72: Can we abstract all of that for identity, dst and the oc collector. If so, what would a method that produces this stack return? `impl MakeService` ? I had some trouble with the typo contraints

linkerd/app/src/lib.rs
```diff
@@ -100,21 +103,31 @@ impl Config {
 
         let tap = info_span!("tap").in_scope(|| tap.build(identity.local(), drain_rx.clone()))?;
         let dst = {
-            use linkerd2_app_core::{classify, control, reconnect, transport::tls};
+            const EWMA_DEFAULT_RTT: Duration = Duration::from_millis(30);
+            const EWMA_DECAY: Duration = Duration::from_secs(10);
```
zaharidichev, line 107: What are sensible values for all these constants and magix numbers when it comes to communicating with the control plane?

linkerd/app/core/src/control.rs
```diff
@@ -301,6 +301,211 @@ pub mod resolve {
...
+        type Endpoint = Target;
+        type Error = Error;
+
+        fn poll(
+            self: Pin<&mut Self>,
+            cx: &mut Context<'_>,
+        ) -> Poll<Result<Update<Self::Endpoint>, Self::Error>> {
+            let mut this = self.project();
+            loop {
+                match this.state.as_mut().project() {
```
hawkw, line 417: It's really a bummer that we have to write this large manual state machine here rather than being able to use `async-stream`, but I suppose this is because we need a `Resolution` implementation rather than a `Stream`...

Here's a thought: what if we added a function `linkerd2_proxy_core::resolve::from_stream` that looked like this:
```rust
pub fn from_stream<S>(stream: S) -> FromStream<S> {
     FromStream(stream)
}

#[pin_project]
pub struct FromStream<S>(#[pin] S);

impl<E, S> Resolution for FromStream<S>
where
    S: TryStream<Ok = Update<E>>,
    S::Error: Into<Error>,
{
    type Endpoint = E;
    type Error = S::Error;

    fn poll(
        self: Pin<&mut Self>,
        cx: &mut Context<'_>,
    ) -> Poll<Result<Update<Self::Endpoint>, Self::Error>> {
        let res = futures::ready!(self.project().0.try_poll_next(cx))
            .expect("resolution stream must be infinite!");
        Poll::Ready(res)
    }
}
```

Then, we could implement `Resolution` here by implementing a `Stream` using `async-stream`, and turning that `Stream` into a `Resolution`. This avoids the manual state machine implementation (and would let us use unboxed futures for the DNS resolution)...

linkerd/dns/src/lib.rs
```diff
@@ -26,15 +28,23 @@ pub trait ConfigureResolver {
...
-    NoAddressesFound,
+    NoAddressesFound(Instant),
     ResolutionFailed(ResolveError),
     TaskLost,
 }
 
 pub type Task = Pin<Box<dyn Future<Output = ()> + Send + 'static>>;
 
+pub type ResolveResponseFuture =
+    Pin<Box<dyn Future<Output = Result<ResolveResponse, Error>> + Send + 'static>>;
```
hawkw, line 39: If the `Resolution` was implemented using `async-stream`, the `resolve_ips` function could just return `impl Future` (or even just be an `async fn`, I think). That would let us avoid the box here.

linkerd/dns/src/lib.rs
```diff
@@ -112,12 +122,32 @@ impl Resolver {
         &self,
         name: &Name,
     ) -> Pin<Box<dyn Future<Output = Result<net::IpAddr, Error>> + Send + 'static>> {
+        Box::pin(
+            self.resolve_ips(name).map(|rsp| {
+                rsp.map(|result| result.ips.first().expect("would error if empty").clone())
+            }),
+        )
```
hawkw, line 129: it's a bummer that we are now double-boxing this future. it might be worth the cost of a little more code duplication to have this method just call `resolver.lookup_ip`, the way that `resolve_ips` does, to avoid having this future be double-boxed.

alternatively, if we are able to change `resolve_ips` to return an `impl Future` as I discussed above, this wouldn't be double boxing...

linkerd/app/core/src/control.rs
```diff
@@ -358,150 +188,148 @@ pub mod dns_resolve {
...
-                                let mut new_updates = VecDeque::default();
+                                let (adds, removes) = diff_targets(&current_targets, &new_targets);
                                 if !adds.is_empty() {
-                                    new_updates.push_back(Update::Add(adds))
+                                    pending_updates.push_back(Update::Add(adds))
                                 }
                                 if !removes.is_empty() {
-                                    new_updates.push_back(Update::Remove(removes))
+                                    pending_updates.push_back(Update::Remove(removes))
                                 }
```
hawkw: I think it _might_ be possible to simplify some of this logic now that we're using `async-stream`, by just `yield`ing these updates _here_ rather than queueing them. Since the stream will then resume _from_ the last `yield` when it's polled again, I think we might be able to avoid the queue entirely. 

It's fine if this ends up not being possible --- just a thought.

linkerd/app/core/src/control.rs
```diff
@@ -358,150 +188,148 @@ pub mod dns_resolve {
...
+        let mut removes = Vec::default();
 
-    impl resolve::Resolution for Resolution {
-        type Endpoint = Target;
-        type Error = Error;
+        for new_target in new_targets.iter() {
+            if !old_targets.contains(new_target) {
+                adds.push((new_target.addr, new_target.clone()))
+            }
+        }
```
hawkw: I know this code was already there, so don't worry about changing it if you don't want to, but I think we could probably do this with [`HashSet::difference`](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.difference).
```suggestion
        for new_target in new_targets.difference(old_targets) {
                adds.push((new_target.addr, new_target.clone()))
        }
```
or even
```suggestion
        let removes = new_targets.difference(old_targets)
            .map(|new_target| (new_target.addr, new_target.clone())
            .collect();
```

linkerd/app/core/src/control.rs
```diff
@@ -358,150 +188,148 @@ pub mod dns_resolve {
...
+    ) -> impl Stream<Item = Result<Update<Target>, Error>> {
+        return match address.clone().addr {
+            Addr::Socket(_) => unimplemented!(),
+            Addr::Name(name_addr) => {
+                let port = address.addr.port().clone();
+                let server_name = address.identity.clone();
+                try_stream! {
+                    let mut resolve_ips = Some(dns.resolve_ips(name_addr.name()));
+                    let mut pending_updates: VecDeque<Update<Target>> = VecDeque::default();
+                    let mut current_targets: HashSet<Target> = HashSet::new();
```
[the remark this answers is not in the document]
zaharidichev: Yes it simply does not compile without them.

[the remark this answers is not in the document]
hawkw: :+1 thanks for confirming.

linkerd/app/core/src/control.rs
```diff
@@ -145,160 +145,375 @@ pub mod add_origin {
...
+            ]);
+
+            let mut stream = task::spawn(
+                resolution_stream(dns_resolver, address("test.com:8888"))
+                    .map_err(|err| err.to_string()),
+            );
+            let result = assert_ready!(stream.poll_next());
+            assert!(result.unwrap().is_err());
+            assert_ready_eq!(stream.poll_next(), Some(Ok(add(vec!["127.0.0.1:8888"]))));
+        }
```
zaharidichev, line 502: This should really keep yielding after error. Need to verify the assumptions async-stream makes.

linkerd/app/core/src/control.rs
```diff
@@ -145,160 +145,376 @@ pub mod add_origin {
...
+        let mut adds = Vec::default();
+        let mut removes = Vec::default();
 
-        fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-            self.inner.poll_ready(cx).map_err(Error::Inner)
+        for new_target in new_targets.iter() {
+            if !old_targets.contains(new_target) {
+                adds.push((new_target.addr, new_target.clone()))
+            }
         }
```
zaharidichev, line 212: Is it worth using a Set here? This messes up the order in which these endpoints are returned and causes problems. Does the .contains in a `Vec` pose a significant problem when considering the number of endpoints should not be so high (I suppose)?

linkerd/app/core/src/control.rs
```diff
@@ -145,160 +145,376 @@ pub mod add_origin {
...
+    pub struct Resolve<R> {
+        dns: R,
     }
 
-    #[derive(Clone, Debug)]
-    pub struct Resolve<M> {
-        dns: dns::Resolver,
-        inner: M,
-    }
+    // === impl ControlPlaneResolve ===
```
kleimkuhler, line 169: nit

```suggestion
// === impl Resolve ===
```

linkerd/dns/src/lib.rs
```diff
@@ -3,45 +3,64 @@
...
-    NoAddressesFound,
+    NoAddressesFound(Instant, bool),
     ResolutionFailed(ResolveError),
     TaskLost,
 }
 
 pub type Task = Pin<Box<dyn Future<Output = ()> + Send + 'static>>;
 
-pub type IpAddrFuture = Pin<Box<dyn Future<Output = Result<net::IpAddr, Error>> + Send + 'static>>;
+pub type ResolveResponseFuture =
```
kleimkuhler, line 58: I don't think this type is used anymore

linkerd/dns/src/lib.rs
```diff
@@ -94,9 +113,11 @@ impl Resolver {
             }
             tracing::debug!("all resolver handles dropped; terminating.");
         });
-        Ok((Resolver { tx }, task))
+        Ok((DnsResolver { tx }, task))
     }
-
+}
+#[async_trait]
```
kleimkuhler, line 119: nit: Add newline above?

linkerd/app/core/src/control.rs
```diff
@@ -145,160 +145,376 @@ pub mod add_origin {
...
-                        let (addr, config) = cfg.take().unwrap();
-                        let state = State::make_inner(addr, &config, svc);
-                        this.state.as_mut().set(state);
+                }
+                Addr::Name(name_addr) => {
+                    let mut current: Vec<Target> = Vec::new();
+                    loop {
+                        let mut resolve_ips = Some(dns.resolve_ips(name_addr.name()));
+                        if let Some(resolve_result) = resolve_ips {
+                            match resolve_result.await {
```
kleimkuhler, line 252: I think this can be simplified my just matching the `resolve_ips` result

```suggestion
                        match dns.resolve_ips(name_addr.name()).await {
```

linkerd/app/core/src/control.rs
```diff
@@ -145,160 +145,376 @@ pub mod add_origin {
...
+                                Err(dns::Error::NoAddressesFound(valid_until, exists)) => {
+                                    resolve_ips = None;
```
kleimkuhler, line 254: Then this can be removed

linkerd/app/core/src/control.rs
```diff
@@ -145,160 +145,376 @@ pub mod add_origin {
...
+                                    current.clear();
+                                    if exists {
+                                        yield Update::Empty;
+                                    } else {
+                                        yield Update::DoesNotExist;
+                                    }
+                                    time::delay_until(valid_until).await;
+                                }
+                                Ok(result) => {
+                                    resolve_ips = None;
```
kleimkuhler, line 265: And this

linkerd/app/core/src/control.rs
```diff
@@ -145,160 +145,376 @@ pub mod add_origin {
...
+                    // this should yield the socket address once
+                    // and keep on yielding NotReady forever.
+                    let mut stream = futures::stream::once(future::ok::<Update<Target>, Error>(
+                        Update::Add(vec![(sa, Target::new(sa, address.identity))]),
+                    ))
+                    .chain(tokio::stream::pending());
+
+                    while let Some(value) = stream.next().await {
+                        debug!(?value, "resolved once");
+                        yield value.expect("no errors in this stream");
```
hawkw, line 244: I think this could be a bit simpler:
```suggestion
                    // this should yield the socket address once
                    // and keep on yielding NotReady forever.
                    debug!(?value, "resolved once");
                    yield Ok(Update::Add(vec![(sa, Target::new(sa, address.identity))]));
                    futures::future::pending().await;
```

linkerd/dns/src/lib.rs
```diff
@@ -3,45 +3,64 @@
...
 
-struct ResolveRequest {
-    name: Name,
-    result_tx: oneshot::Sender<Result<LookupIp, ResolveError>>,
-    span: tracing::Span,
-}
+pub type IpAddrFuture = Pin<Box<dyn Future<Output = Result<net::IpAddr, Error>> + Send + 'static>>;
 
-impl Resolver {
+impl DnsResolver {
```
hawkw, line 63: Was the rename necessary? It makes this diff much larger. Not really a blocker, just a nit.

linkerd/app/core/src/control.rs
zaharidichev: Sorry, I was not clear. This does not break anything. Having consistent order simply makes testing easier.

linkerd/dns/src/lib.rs
zaharidichev: I let the trait be `Resolver` and the dns impl be `DnsResolver`, therefore the raname.

olix0r: FWIW, the rename makes this PR a bit harder to review. Can we call the trait `Resolve` (we tend to use verbs for trait names) to avoid this?

linkerd/dns/src/lib.rs
```diff
@@ -3,45 +3,60 @@
...
 
 pub trait ConfigureResolver {
     fn configure_resolver(&self, _: &mut ResolverOpts);
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub enum Error {
-    NoAddressesFound,
+    NoAddressesFound(Instant, bool),
```
olix0r, line 51: please use named fields here. It's not at all obvious what the intent is from looking at the type.

linkerd/dns/src/lib.rs
```diff
@@ -3,45 +3,60 @@
...
+
+struct ResolveRequest {
+    name: Name,
+    result_tx: oneshot::Sender<Result<LookupIp, ResolveError>>,
+    span: tracing::Span,
+}
+
+#[async_trait]
+pub trait Resolver: Clone {
+    async fn lookup_ip(&self, name: Name, span: Span) -> Result<LookupIp, Error>;
```
olix0r, line 34: Why is `Span` in this API? that seems weird?

linkerd/dns/src/lib.rs
```diff
@@ -3,45 +3,60 @@
...
+}
+
+struct ResolveRequest {
+    name: Name,
+    result_tx: oneshot::Sender<Result<LookupIp, ResolveError>>,
+    span: tracing::Span,
+}
+
+#[async_trait]
+pub trait Resolver: Clone {
```
olix0r, line 33: What is this trait used for? Where are we generic over the DNS impl?

zaharidichev: In the tests

https://github.com/linkerd/linkerd2-proxy/blob/33d4ec8aa2a3e081ea286f24573119fb7fb4fcaf/linkerd/app/core/src/control.rs#L294

olix0r: Hm, it feels like this shouldn't be tied to the DNS behavior bout to resolution behavior in general...

zaharidichev: But we are testing the logic that turns a Dns resolution request/response API into a resolution stream. So not sure how we can abstract that or avoid the trait. Maybe I am missing something

zaharidichev: This set of tests is here to test the logic in this function: https://github.com/linkerd/linkerd2-proxy/pull/594/files/33d4ec8aa2a3e081ea286f24573119fb7fb4fcaf#diff-fe9e02d6254c4455e6c23df861eb03d1R222

olix0r: Maybe another way to approach this... Can we do something similar without having to introduce a new trait? It looks like the tests only mock a single async function. What if, instead of the dns-specific trait, we model this use case as a `Service` that is either implemented by or adaptable from the dns resolver type?

zaharidichev: Got it! Thanks for the tip!

linkerd/app/src/identity.rs
```diff
@@ -45,6 +47,18 @@ impl Config {
...
+                const EWMA_DEFAULT_RTT: Duration = Duration::from_millis(30);
+                const EWMA_DECAY: Duration = Duration::from_secs(10);
+                let discover = {
+                    const BUFFER_CAPACITY: usize = 1_000;
+                    let cache_timeout = Duration::from_secs(60);
+                    discover::Layer::new(
+                        BUFFER_CAPACITY,
+                        cache_timeout,
+                        control::dns_resolve::Resolve::new(dns),
+                    )
```
olix0r, line 59: So, while I think the larger change looks fine, I'm a bit unsure on this detail... specifically how the watchdog timeout (what you call `cache_timeout` here) will interact with these clients.

The watchdog timeout probably isn't even needed anymore, but it was there to detect the case when the client had been orphaned such that discovery updates would back up in the buffer, and then eventually exert backpressure on the client. But in cases like the identity client, we may only send something like one request a day; so in this case we _expect_ the client to be idle (and not dropped) for long periods of time. So I worry that the watchdog timeout is inappropriate here.

linkerd/app/core/src/control.rs
```diff
@@ -145,160 +145,361 @@ pub mod add_origin {
...
+            }
+        }
+
+        (adds, removes)
+    }
+
+    pub fn resolution_stream<S>(
+        mut dns: S,
+        address: ControlAddr,
+    ) -> impl Stream<Item = Result<Update<Target>, Error>>
```
olix0r, line 227: [This comment](https://github.com/linkerd/linkerd2-proxy/pull/594#pullrequestreview-459533777) should have referenced this implementation

linkerd/app/src/identity.rs
olix0r: On further thought, I don't think we even _need_ a buffer around these resolutions?

In the discovery case, resolutions are sent by the server to the proxy and we have to consume them all to prevent the stream/window from backing up. In the DNS case, everything is driven by the act of polling the stream. So I don't think we have anything to buffer, really.

olix0r:
See https://github.com/linkerd/linkerd2-proxy/pull/631 and https://github.com/linkerd/linkerd2-proxy/pull/632

olix0r:
@zaharidichev I believe you said you had this working with SRV records?

zaharidichev:
@olix0r Yes that is correct. I have some local changes that address  some of your points from above, including the SRV records and the ports situation.

linkerd/app/core/src/control.rs
```diff
@@ -145,160 +145,117 @@ pub mod add_origin {
...
-            inner,
-        })
+        // should yield the socket address once
+        // and keep on yielding Pending forever.
+        pub fn resolve_once_stream(&self, sa: SocketAddr, identity: PeerIdentity) -> UpdatesStream {
+            Box::pin(try_stream! {
+                let update = Update::Add(vec![(sa, Target::new(sa, identity))]);
+                tracing::debug!(?update, "resolved once");
+                yield update;
+                future::pending::<Update<Target>>().await;
```
olix0r, line 190: Why is this necessary? Does the balancer blow up if the stream ends?

linkerd/app/core/src/control.rs
```diff
@@ -145,160 +145,117 @@ pub mod add_origin {
...
-            }
+            let this = self.project();
+            let identity = this.identity.clone();
+            let stream = futures::ready!(this.inner.poll(cx))?;
+            Poll::Ready(Ok(Box::pin(stream.map_ok(move |update| {
+                match update {
+                    Update::Add(adds) => Update::Add(
+                        adds.clone()
+                            .into_iter()
+                            .map(|(sa, _)| (sa, Target::new(sa, identity.clone())))
```
olix0r, line 249: Curious if we can avoid having to have this code by using a resolve::MapEndpoint instead

linkerd/proxy/dns-resolve/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+        pin_mut!(input);
+        let mut current: Vec<SocketAddr> = Vec::new();
+        while let Some(result) = input.next().await {
+            let result = result?;
+            if result.is_empty() {
+                current.clear();
+                yield Update::Reset(Vec::new());
+            }
+
+            let (adds, removes) = diff_targets(&current, &result);
```
olix0r, line 83: Should this be in an `} else {`? Or do we want to execute this even when the result is empty?

linkerd/dns/src/lib.rs
```diff
@@ -101,24 +126,79 @@ impl Resolver {
...
+
+        std::str::from_utf8(first_label)
+            .expect("should be utf8")
+            .replace("-", ".")
+            .parse::<std::net::IpAddr>()
+            .map(|ip| net::SocketAddr::new(ip, srv.port()))
+            .map_err(|_| Error::InvalidSRVRecord(srv.clone()))
+    }
+
+    pub fn resolve_service_ips(
```
olix0r, line 169:
```suggestion
    pub fn resolve_service_addrs(
```

This looks up addresses (i.e. with port), not just IPs.

linkerd/proxy/dns-resolve/src/lib.rs
```diff
@@ -0,0 +1,156 @@
...
+        let sa_1 = "127.0.0.1:8080".parse().unwrap();
+        let sa_2 = "127.0.0.2:8080".parse().unwrap();
+
+        tx.send(Ok(vec![sa_1, sa_2])).await.unwrap();
+        assert_ready_eq!(
+            stream.poll_next(),
+            Some(Ok(Update::Add(vec![(sa_1, sa_1), (sa_2, sa_2)])))
+        );
+        tx.send(Ok(vec![])).await.unwrap();
+        assert_ready_eq!(stream.poll_next(), Some(Ok(Update::Empty)));
```
hawkw, line 154: Looks like this [doesn't compile](https://github.com/linkerd/linkerd2-proxy/pull/594/checks?check_run_id=1022713111#step:4:486) --- was it supposed to be `Update::Reset`?

olix0r:
Are any control plane changes needed to test this? I seem to hit errors like:
```
linkerd-web-766dd9bb65-mn7zm linkerd-proxy [     0.12560880s]  WARN ThreadId(13) dst: linkerd2_reconnect::service: Service failed error=load balancer discovery error: Invalid SRV record SRV { priority: 0, weight: 100, port: 8086, target: Name { is_fqdn: true, labels: [linkerd-dst, linkerd, svc, cluster, local] } }
```

Are we going to have to support falling back kube-proxy?

zaharidichev:
@olix0r yes you need to make identitiy and dst headless. Here is the branch: https://github.com/linkerd/linkerd2/commit/0037ad1e334b47ef606ffac5feb3d1a80179352f. 

Do we want to fall back on kube-proxy. If so, how would we do that? Just do an Ip lookup in case we error out?

linkerd/app/core/src/control.rs
```diff
@@ -145,174 +151,74 @@ pub mod add_origin {
...
 
-    #[derive(Clone, Debug)]
-    pub struct Layer {
+    pub fn layer<M>(
         dns: dns::Resolver,
+    ) -> impl svc::Layer<
+        M,
+        Service = discover::MakeEndpoint<
+            discover::FromResolve<map_endpoint::Resolve<IntoTarget, DnsResolve>, Target>,
+            M,
```
hawkw, line 169: not a blocker but could we maybe make this thing a type alias or something?

olix0r: it's only used the once, isn't it? so, yes, but it wouldn't really reduce anything

lieberlois:
Any chance that this is not fixed in non-HA mode? We had this behaviour recently after an AKS node crash, had 2 Linkerd Destination Pods, one of which was on that node 🤔

PR #595, opened 2020-07-15
admin: add an endpoint to dump spawned Tokio tasks

hawkw:
## Motivation

When debugging proxy issues, it can be useful to inspect the list of
currently spawned Tokio tasks and their states. This can be used
similarly to the thread or coroutine dumps provided by other languages'
runtimes.

## Solution

This branch adds a new endpoint to the proxy's admin server, `/tasks`,
that returns a dump of all tasks currently spawned on the Tokio runtime,
using the new Tracing instrumentation added in tokio-rs/tokio#2655, and
a work-in-progress [`tokio-trace`] crate that provides Tokio-specific
Tracing layers. 

Currently, the `/tasks` admin endpoint records the following information
about each task:

* Whether it is a normal, local, or blocking task (not relevant to us 
  currently, since Linkerd does not use local or blocking tasks...but
  we might eventually!)
* Whether the task is active (currently being polled) or idle (waiting
  to be polled)
* The type of the future that was spawned
* The Tracing span context from which the task was spawned
* The total number of times the task has been polled
* Timing statistics about the task, including:
  - The time in nanoseconds between when the task was spawned and when
    it was first polled (essentially, measuring the Tokio scheduler's
    latency)
  - The total time in nanoseconds the task has existed
  - The task's _busy time_ in nanoseconds (time it was actively being
    polled)
  - The tasks _idle time_ in nanoseconds (time it was _not_ being 
    polled)

In the future, Tokio will likely expose additional Tracing information,
which we'll be able to collect as well.

The task dump can be accessed either as an HTML table or as JSON. JSON
is returned if the request has an `Accept: application/json` header, or
whenever the path `/tasks.json` is requested; otherwise, the data is
rendered as an HTML table. Like the `/proxy-log-level` endpoint, access
is denied to requests coming from sources other than localhost, to help
restrict access to authorized users (since a high volume of requests for
task dumps could be used to starve the proxy).

Example JSON output (in Firefox Dev Edition's extremely nice GUI 
JSON viewer):

Screenshot_20200715_121938

Zoomed in on the timing data for a single task:
Screenshot_20200715_122047

And HTML:

Screenshot_20200715_143155

Because the task data is generated from Tracing spans emitted by Tokio,
the task spans must be enabled for it to be used. This can be done by
setting a trace filter that enables the `trace` level for the target
`tokio::task`, e.g.:
```
tokio::task=trace
```
or 
```
tokio=trace
```

## Notes

* This branch depends on unreleased code from upstream, including a
  Tokio change that has merged to master but not been published, and my
  unreleased work-in-progress [`tokio-trace`] crate. Therefore, I've
  pinned these upstreams to fixed Git SHAs, to guard against
  dependencies changing under us unexpectedly.
* I considered requiring a build-time feature flag to enable this
  feature, the way we do for the mock SO_ORIG_DST implementation for
  testing. However, this would make it harder to use task tracking to
  debug issues in proxies not built with the flag. I'm happy to change
  this code to be feature flagged if we think that's the right approach.

[`tokio-trace`]: https://github.com/hawkw/tokio-trace

Closes linkerd/linkerd2#3803


olix0r:
@hawkw re: feature-flagging: What's the "cost" of this? Is there any discernible different in benchmarks, for instance?

hawkw:
> @hawkw re: feature-flagging: What's the "cost" of this? Is there any discernible different in benchmarks, for instance?

It should be pretty minimal when disabled at runtime, but I'll do a benchmark run to make sure, good call.

linkerd/app/core/src/admin/tasks.rs
```diff
@@ -0,0 +1,134 @@
...
+use http::{header, StatusCode};
+use hyper::{service::Service, Body, Request, Response};
+use std::task::{Context, Poll};
+use std::{
+    fmt::{self, Write},
+    io,
+};
+use tokio_trace::tasks::TaskList;
+#[derive(Clone)]
+pub struct Tasks {
```
davidbarsky, line 11: Not a blocker (obviously) but do you think this can live in hawkw/tokio-trace behind a feature flag?

hawkw: PRs_welcome

Dockerfile
```diff
@@ -36,21 +36,21 @@ ARG PROXY_UNOPTIMIZED
 ARG PROXY_FEATURES
 
 RUN --mount=type=cache,target=/var/lib/apt/lists \
-    --mount=type=cache,target=/var/tmp \
-    apt update && apt install -y time cmake
+  --mount=type=cache,target=/var/tmp \
```
kleimkuhler, line 39: Auto formatting here? Not a blocker but just wondering if this was intentional.

olix0r:
> @hawkw re: feature-flagging: What's the "cost" of this? Is there any discernible different in benchmarks, for instance?

@hawkw  Have you run benchmarks since your updates?

Dockerfile
hawkw: yup, i think this change can be removed entirely — this was left over from an attempt at feature-flagging.

hawkw:
> > @hawkw re: feature-flagging: What's the "cost" of this? Is there any discernible different in benchmarks, for instance?
> 
> @hawkw Have you run benchmarks since your updates?

@olix0r here are benchmark results with the most recent commit to this branch. Looks like a little bit of overhead:
Screenshot_20200724_134832

hawkw:
@ver my latest commit 3f404bbb70c27c174a7e64354d2a95677dc1dd91 should decrease the overhead a little bit more:
Screenshot_20200724_152747
(except for in the x100 case, weirdly? maybe that was bench env noise?)

Dockerfile
olix0r: Yeah, would prefer to revert this as it removes some indentation below

linkerd/app/core/src/trace.rs
```diff
@@ -90,24 +117,26 @@ impl LevelHandle {
...
 
+impl LevelHandle {
     pub fn set_level(&self, level: impl AsRef<str>) -> Result<(), Error> {
         let level = level.as_ref();
         let filter = level.parse::<EnvFilter>()?;
         match self {
-            Self::Json(level) => level.reload(filter)?,
-            Self::Plain(level) => level.reload(filter)?,
+            LevelHandle::Json(level) => level.reload(filter)?,
+            LevelHandle::Plain(level) => level.reload(filter)?,
```
olix0r, line 128: Why change this from `Self`?

hawkw: I think these methods briefly moved to a different type and then came back. Will back that out.

olix0r:
@hawkw looks like this merge wasn't clean

hawkw:
> @hawkw looks like this merge wasn't clean

yeah, I'm fixing that up right now!

PR #601, opened 2020-07-22
Access Logging

tustvold:
Access logging is very important functionality for my team as we wish to maintain feature-parity with our existing AWS ALB-based approach. This functionality was requested [here](https://github.com/linkerd/linkerd2/issues/1913) and was marked as help wanted, so thought I'd take a stab at implementing it.

Creating as a draft as still needs more testing and benchmarking, and I'm new to tower and so might have made some rookie errors. However, I wanted to create a draft as an opportunity to get some early feedback.

The basic design consists of an AccessLogLayer that instruments both requests and responses that flow through it, in a similar manner to how handle_time is already computed. I'm not a massive fan of this, but it was the only way I could easily see to get accurate processing time metrics. I've tried to avoid any memory allocation on the hot path, although there are possibly more atomic increments than I would like. The performance impact with the feature disabled, i.e. LINKERD2_PROXY_ACCESS_LOG_FILE not set should be minimal.

The results of this instrumentation are then sent over a mpsc channel to an AccessLogCollector that writes them in a space-delimited format to a file specified as an environment variable. It buffers in memory and flushes on termination and on write if more than FLUSH_TIMEOUT_SECS since the last flush. This makes the access logging best effort much like AWS ALBs.

An example deployment scenario using this functionality might deploy a fluent-bit sidecar to ship the logs, or write to /dev/stderr and use a log shipper deployed as a DaemonSet.


olix0r:
Thanks @tustvold!

@hawkw would you mind looking at this? Can we use `tracing` for any of this? Particularly, the tracker code looks pretty complex to me; and I'm curious if we already have code that does this sort of thing.

hawkw:
I've not looked _too_ closely yet, but yeah, I imagine it should be possible to come up with a solution using existing `tracing` information. If we used the existing request spans to time request duration, I think we could potentially avoid a lot of complexity.

Eventually, it might even be worth implementing a `tracing` layer for access logging as a general purpose library --- but this might require other projects, like `hyper`, to opt into conventional naming for HTTP-specific fields (which would be nice eventually regardless...).

tustvold:
As far as I can tell it isn't possible to directly consume the tracking metrics as not only are they sampled, but they don't record the necessary request metadata for access logging. That being said I've updated this PR to use similar timing logic, doing away with the complex tracker logic.

Unfortunately this does result in a loss of granularity, in particular the tracker based implementation yielded 3 separate time-related metrics:

- request_processing_time: the time the request spends in the proxy
- upstream_processing_time: time between when the request was sent and a response was received
- response_processing_time: the time between when the response was received and dispatched to the downstream

Whereas this approach only yields a single metric that is the sum of the previous implementation's request_processing_time and upstream_processing_time. However, in my (limited) testing response_processing_time is consistently on the order of a couple of microseconds, and the request_processing_time is already being tracked and exported as the handle_time metric, so this is no great loss.

Additionally the previous approach, whilst based on the way the handle_time metric is computed, is potentially misleading as it only really measures the time until the request is handed off to the L4 stack, so I'm not all that attached to it :smile: 

I'll take a look over the weekend at getting some test coverage in, but let me know if there is a better way to implement this functionality.

tustvold:
Appears to be shutdown race in test, will look into tomorrow

tustvold:
Working on separate PR to fix non-determinism in integration test shutdown

hawkw:
> As far as I can tell it isn't possible to directly consume the tracing metrics as not only are they sampled, but they don't record the necessary request metadata for access logging. That being said I've updated this PR to use similar timing logic, doing away with the complex tracker logic.

`tracing` spans (`tracing` in the sense of [the Rust crate](https://docs.rs/tracing/0.1.17/tracing/), not "tracing" in the sense of the Linkerd OpenCensus integration) are not sampled but _filtered_; I imagine that if we added a [`tracing` `Layer`](https://docs.rs/tracing-subscriber/0.2.9/tracing_subscriber/layer/trait.Layer.html) for access logging, it would always enable the spans necessary to generate the access log, regardless of other filter configurations. And, if we decided to implement this using existing `tracing` instrumentation, I think we would want to add a single per-request span that contains all the necessary request metadata as fields on that span (which it looks like we don't _currently_ have, as all that data is split across multiple spans). 

I brought this up primarily because I think a reusable library for access logging with `tracing` would be beneficial to the whole ecosystem. If the current approach here is a quicker and easier way of adding access logging to the Linkerd proxy in particular, I'm not going to block merging this PR on implementing a general-purpose library...it just might be nice to do at some point.

linkerd/access-log/src/layer.rs
```diff
@@ -0,0 +1,122 @@
...
+        };
+
+        let t0 = SystemTime::now();
+
+        let host = request.headers().get("Host").cloned();
+
+        let trace_id = request
+            .headers()
+            .get("X-Amzn-Trace-Id")
+            .or_else(|| request.headers().get("X-Request-ID"))
```
hawkw, line 72: we should also handle B3 trace propagation, especially since it's the only trace propagation header that Linkerd proxies are [actually aware of](https://github.com/linkerd/linkerd2-proxy/blob/1b3e899efad2ad4ee471c5b49482437e7a325bde/linkerd/trace-context/src/propagation.rs)...

linkerd/access-log/src/layer.rs
```diff
@@ -0,0 +1,122 @@
...
+
+        let user_agent = request.headers().get("User-Agent").cloned();
+
+        let log = AccessLog {
+            uri: request.uri().clone(),
+            method: request.method().clone(),
+            status: Default::default(),
+            host,
+            user_agent,
+            trace_id,
```
hawkw, line 83: bit of a shame that we have to clone so many strings here (the URI, `host` header, `user-agent` header, and trace id).

it seems like it might be more efficient if we just formatted all of the fields we know at this point into a single _new_ string, and then appended the remaining fields to it when the request completes. that way, we'd be making 2 allocations (one for the string, and one to carry the trace ID along because it occurs at the end of the line), rather than 5 or so (4 for all these values, and then another for the string that we format when the request completes).

just a thought!

linkerd/app/src/access_log_collector.rs
```diff
@@ -0,0 +1,192 @@
...
+    const FLUSH_TIMEOUT_SECS: u64 = 1;
+
+    /// Transition future to shutting down. Future will complete with result
+    fn shutdown(&mut self) {
+        debug!("shutting down access log collector");
+        self.state = State::Shutdown
+    }
+}
+
+impl Future for CollectorSink {
```
hawkw, line 107: I think this could probably be expressed much more neatly as an `async fn`...unless the manual future implementation is necessary for the shutdown process? We could also just move the flush-on-shutdown behavior to a `Drop` guard that runs when the future is dropped...

linkerd/access-log/src/layer.rs
tustvold: I _think_ that HeaderValue internally uses Bytes, which is internally reference counted, so these _should_ just be atomic increments... But I could be wrong on that front

linkerd/app/src/access_log_collector.rs
tustvold: I'll have another play around, I couldn't work out how to make it work with the shutdown signal  :sweat_smile:

linkerd/access-log/src/layer.rs
tustvold: Haha, yes definitely! In fact I'll just use that header, no need to unnecessarily pollute linkerd2 with Amazon specific headers. Longer term might be nice to have the ability to include arbitrary headers, maybe as part of a [custom log format system](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage), but definitely not something for an MVP implementation such as this.

linkerd/access-log/src/layer.rs
hawkw: Oh, you're right; I didn't look closely enough at the struct definition and assumed they were being accessed as strings. The URI is also represented using `Bytes`, so this is fine.

Never mind!

tustvold:
Thanks for the review, I'll look to get an update done, factoring in your feedback, over the next week time-permitting. Currently trying to fix/workaround the fact that the integration test drops the proxy's thread leading to non-deterministic shutdown behavior.

tustvold:
Well it was a bit more than a week, but I eventually manged to get time to work on this.

I've re-implemented the functionality to use tokio-tracing instead of a custom mpsc queue as requested. In particular the AccessLogLayer now produces tokio::tracing spans. This PR then adds an additional tracing-subscriber layer `AccessLogWriter` which asynchronously writes access_log spans as they are closed. This is currently to stdout but this could easily be extended in future.

As the `AccessLogWriter` is part of the single tracing layer stack, the instrumentation level is controlled by `EnvFilter` and by extension the `LINKER2_PROXY_LOG` environment variable. For example, `LINKERD2_PROXY_LOG=warn,access_log`, will enable access logs. This feels a little bit off, but I wasn't really sure how else to handle this?

Another thing to note is if people have the JSON log format enabled, this will probably not perform very well due to the way on_record is currently implemented and the number of fields.

Finally the way request_bytes and response_bytes is determined won't work for chunked requests, but I wasn't sure how else to easily obtain this information?

tustvold:
Rebased onto current master

tustvold:
Various iterations of this have been running in a test cluster for the last two months without issue. However, I've noticed the recent PRs to reduce the tracing due to performance issues, which makes me somewhat concerned that this may not be the best approach? 

I'll shortly be performing some load-testing on this test cluster and so would appreciate your insight into this, in case the approach in this PR is fundamentally flawed and there would be a better way to provide this functionality.

olix0r:
@tustvold Thanks for your patience and persistence on this. We were pretty focused on getting stable-2.9.0 ready; but now is probably a good time for us to come back to this.

My gut says that tracing is the right approach to this -- access logging is going to be expensive in high-throughput systems and I don't think there's a lot we can do about that. That said, there may be approaches that minimize the impact by more-narrowly targeting the spans/events to be captured for access logging.

I'll try to make some time to look at this more closely in the next ~week. I think we'd be pretty eager to add support to access logging as long as its cost is only incurred when a user opts to enable the feature.

olix0r:
@tustvold Again, please forgive our slowness in getting this merged. I've merged recent changes into this branch pushed them so we can hopefully move this forward in the next ~week.

I think the one change I'd be interested in pursuing before merging would be to only enable this functionality as needed. I.e., for most uses we don't even want to enable the maps/etc and we want the layer to be a simple no-op. When a user has requested access logging (probably via workload annotation), we want to enable this. At the very least, this will let us more confidently ignore this when doing performance analysis/debugging.

@hawkw Perhaps you can help with this?

linkerd/access-log/src/tower.rs
```diff
@@ -0,0 +1,169 @@
+use futures::TryFuture;
+use pin_project::pin_project;
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use std::time::{Duration, Instant};
+use tracing::{field, span, Level, Span};
+
+/// A tower layer that associates a tokio-tracing Span with each request
```
hawkw, line 9: nit: i would probably say
```suggestion
/// A tower layer that associates a `tracing` `Span` with each request
```

linkerd/access-log/src/tower.rs
```diff
@@ -0,0 +1,169 @@
...
+#[derive(Clone)]
+pub struct AccessLogLayer {}
+
+#[derive(Clone)]
+pub struct AccessLogContext<Svc> {
+    inner: Svc,
```
hawkw, line 15: nit/style: generally we would just call this
```suggestion
pub struct AccessLogContext<S> {
    inner: S,
```

linkerd/access-log/src/tower.rs
```diff
@@ -0,0 +1,169 @@
...
+            .and_then(|x| x.to_str().ok())
+            .map(|x| span.record("request_bytes", &x));
+
+        request
+            .headers()
+            .get("x-b3-traceid")
+            .or_else(|| request.headers().get("X-Request-ID"))
+            .or_else(|| request.headers().get("X-Amzn-Trace-Id"))
+            .and_then(|x| x.to_str().ok())
+            .map(|x| span.record("trace_id", &x));
```
hawkw, line 100: This could probably be much simpler without sacrificing performance. My guess is that we're using `field::Empty` here to avoid evaluating the fields if the span is disabled. But, if the span is disabled, the `tracing::span!` macro won't evaluate any field expressions, so i think we could pass most of these (or at least, the simpler ones) as fields in the macro.

Also, in the case where the span _is_ enabled, it's actually more efficient to record as many fields in the `span!` macro as possible, since they will be batched and passed to the `tracing` subscriber in one function call. When we call the `record` method a bunch of times, we pass each individual field on its own.

linkerd/access-log/src/tracing.rs
```diff
@@ -0,0 +1,74 @@
...
+        }
+    }
+}
+
+impl<S, W, F> Layer<S> for AccessLogWriter<W, F>
+where
+    W: MakeWriter + 'static,
+    S: Subscriber + for<'span> LookupSpan<'span>,
+    F: for<'writer> FormatFields<'writer> + 'static,
+{
```
hawkw, line 63: This should probably have `enabled` and `register_callsite` implementations that ensure we enable the `access_log` span. Otherwise, the current filtering configuration might disable it.

hawkw (commented):
Overall, this looks quite good --- I only commented on a few minor nits. 

I agree with @olix0r's suggestion that the access log should be conditionally enabled/disabled, probably by an environment variable; I have a sense for how I would probably implement that. @tustvold, if you like, I can go ahead and update this PR so that we can toggle on and off the access log based on an env var, but if you'd prefer to implement that yourself, I'm also happy to just  give some pointers.

linkerd/access-log/src/tower.rs
tustvold: I wasn't aware that the macro wouldn't evaluate fields if the span is disabled - will fix

tustvold:
@hawkw Thanks for the review, I've pushed a commit that I think addresses your comments, although the tracing layer still doesn't seem to be working without `LINKERD2_PROXY_LOG=access_log` but I'm not entirely sure why... Perhaps you could advise?

> if you like, I can go ahead and update this PR so that we can toggle on and off the access log based on an env var, but if you'd prefer to implement that yourself,

Relates somewhat to the above, but if you are willing to do so that would be amazing, I think I'm probably just misunderstanding something but I remember having a fair bit of difficulty trying to get it to work previously, and I've had no further luck this evening. If you could assist with this that would likely be better than me stumbling about in the dark, and I will be interested to learn from what you come up with :grinning:

hawkw (commented):
> > if you like, I can go ahead and update this PR so that we can toggle on and off the access log based on an env var, but if you'd prefer to implement that yourself,
> 
> Relates somewhat to the above, but if you are willing to do so that would be amazing, I think I'm probably just misunderstanding something but I remember having a fair bit of difficulty trying to get it to work previously, and I've had no further luck this evening. If you could assist with this that would likely be better than me stumbling about in the dark, and I will be interested to learn from what you come up with grinning

Okay, great --- I've opened a PR against your branch (https://github.com/tustvold/linkerd2-proxy/pull/1) that  should fix this. I also went ahead and made a few other additions. Please feel free to merge this into your branch if you like, and I'd be happy to answer any questions about it.

Another thing we might want to think about is what format the access log should be emitted in. Right now we just format the `tracing` fields the same way we do in other proxy log lines. The original feature request for access logging (linkerd/linkerd2#1913) requests "Apache-style" access logs, so we may want to consider changing the format to look more like the Apache common log format? I have some thoughts on how we could potentially do that as well.

Also, I left a couple more style comments. :)

tustvold:
@hawkw I've pushed a version containing your changes, incorporating your review feedback and fixing the merge conflicts.

I made some further changes on top which should be fairly self-explanatory. In order to get the system to behave as I would expect I had to remove the `enabled` and `register_callsite` implementations from the AccessLogWriter layer. Otherwise when access logging was enabled, all other logging would be disabled.

With this everything seems to be working as I would expect. With `LINKERD2_PROXY_ACCESS_LOG` not set no access logs appear, even if `access_log` is included in `LINKERD2_PROXY_LOG`. With `LINKERD2_PROXY_ACCESS_LOG` set logs will be written, even if `LINKERD2_PROXY_LOG` doesn't include `access_log`. I've additionally confirmed the system to work with special files such as `dev/stdout` which is very helpful in containerised environments.

olix0r (changes requested):
This is looking really good but I still have one question from my [prior comment](https://github.com/linkerd/linkerd2-proxy/pull/601#issuecomment-841420845)

> I think the one change I'd be interested in pursuing before merging would be to only enable this functionality as needed. I.e., for most uses we don't even want to create the mpsc/etc and we want the layer to be a simple no-op. When a user has requested access logging (probably via workload annotation), we want to enable this. At the very least, this will let us more confidently ignore this when doing performance analysis/debugging.

If I understand correctly, this logging is going to be enabled by default since it's emitted at the INFO level? How do we make this functionality opt-in? I'm mostly concerned that it's likely to negatively impact performance and costs (logging services are expensive) for people who don't necessarily want it.

linkerd/tracing/Cargo.toml
```diff
@@ -15,15 +15,16 @@ bytes = "1"
 html-escape = "0.2.5"
 http = "0.2"
 hyper = { version = "0.14.2", features = ["http1"] }
+linkerd-access-log = { path = "../access-log" }
```
olix0r, line 18: @hawkw Do you have any thoughts on how we can avoid having `tracing` depend on `access-log`? My concern is that now the tracing dependency pulls in a bunch of HTTP, etc specific dependencies that shouldn't be needed by such a core dependency like this.

olix0r: Perhaps the tracing crate should expose `AccessLogWriter` and we should pull that in from the access-log crate so that the dependencies are inverted?

hawkw:
> If I understand correctly, this logging is going to be enabled by default since it's emitted at the INFO level? How do we make this functionality opt-in? I'm mostly concerned that it's likely to negatively impact performance and costs (logging services are expensive) for people who don't necessarily want it.

By default, the log filter is `error,linkerd=info`, right? The access log spans have the target `access_log` (_not_ `linkerd::access_log`), so they *won't* be enabled by the default log filter --- they're only turned on if `info` is enabled _globally_ with a filter like `info` (or `info,linkerd_app_outbound=trace,tokio=debug`, etc).

linkerd/tracing/Cargo.toml
olix0r: I'm just going to give this a try and push these changes to the branch.

olix0r:
@hawkw

> By default, the log filter is `error,linkerd=info`, right? The access log spans have the target `access_log` (_not_ `linkerd::access_log`), so they *won't* be enabled by the default log filter --- they're only turned on if `info` is enabled _globally_ with a filter like `info` (or `info,linkerd_app_outbound=trace,tokio=debug`, etc).

https://github.com/linkerd/linkerd2-proxy/blob/777e0e55782cac7366bdd1d9ec57edacecb212e6/linkerd/tracing/src/lib.rs#L157-L163

Doesn't this always enable the access logging? We're adding the directive to the env filter. If not, how does this work?

olix0r (approved):

olix0r:
PTAL. I've made a few changes:

- the writer is now in `linkerd_tracing::access_log`. It is no longer generic over the `MakeWriter` since it was never constructed without the `NonBlocking` impl. initialization logic is also moved into this module.
- The target is constified as `_access_log` and used from the access-log crate.
- The unneeded AccessLogLayer type has been removed in favor a function, `NewAccessLog::layer`
- The `chrono` call has been made lazy so that we don't call `SystemTime::now` when the access logging is disabled.

olix0r:
@hawkw When the access log is enabled, won't we output this logs to _both_ the files and standard logging output? Is there any way to avoid this?

olix0r:
Perhaps we don't want to support having a separate log file at all -- I'm not sure how that will even work in Kubernetes, anyway. Perhaps just getting it all sent on the standard output is good enough and that makes opting into it fairly trivial (by  setting the proxy log level).

hawkw:
> @hawkw When the access log is enabled, won't we output this logs to _both_ the files and standard logging output? Is there any way to avoid this?

The access logging is generated based on a _span_, not a trace event. So, the data from the access log span will be included on any events that are enabled within that span (which, with the default filter settings, is just warnings --- other per-request events are all below the info level). But, the actual log lines generated for the access log won't be written to stdout.

I think that when we are already recording the extra request context for the access log, also including it in any events we log to stdout is fine, and probably useful?

jon-depop:
Hi :wave:! Would it be possible to get a status update on this PR @olix0r @hawkw?

jon-depop:
@hawkw what's the latest update on this PR?

olix0r:
I'm going to close this out in favor of #1319, which takes advantage of new tracing features. I'm hopeful we can merge that PR fairly soon!

PR #606, opened 2020-07-23
Remove `Resolution` trait

zaharidichev:
This PR removes the `Resolution` and replaces it with a `TryStream<Ok = Update<Endpoint>, Error>`.  This will allow us to treat any resolution as a normal `Stream` and as a consequence make it easier to  transition state machine based stream implementations to using `async-stream`. 

In addition a ResolutionStreamExt trait is defined that abstracts away the fact that we expect this stream to never end (i.e. should never return `None`)

### DIFF
--- a/linkerd/proxy/api-resolve/src/resolve.rs
+++ b/linkerd/proxy/api-resolve/src/resolve.rs
@@ -149,12 +149,7 @@ impl Stream for Resolution {
 
                     None => {} // continue
                 },
-                None => {
-                    return Poll::Ready(Some(Err(grpc::Status::new(
-                        grpc::Code::Ok,
-                        "end of stream",
-                    ))))
-                }
+                None => return Poll::Ready(None),
             };
         }
     }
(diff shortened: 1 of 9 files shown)


olix0r:
I think we probably aught to just let the resolution end rather than panic. And I don't think we need to lift the end-of-stream to a grpc error... and then I would probably avoid the `Ext` trait.

zaharidichev:
@olix0r Understood. I made the requested changes

zaharidichev:
@kleimkuhler just rebased

hawkw:
@zaharidichev it looks like a bunch of the integration tests are now failing at this `expect`:
https://github.com/linkerd/linkerd2-proxy/blob/1b3e899efad2ad4ee471c5b49482437e7a325bde/linkerd/proxy/discover/src/make_endpoint.rs#L198-L199

which was written under the assumption that `Discover` streams would be terminated by errors rather than `None`s.

PR #607, opened 2020-07-25
Move tls::accept to async/await

olix0r:
The `tls::accept` module can be simplified by moving to async/await.
This presents a few opportunities for improvement:

1. The recently-introduced `Detect` trait is used. This sets up further
   changes to the accept logic, decoupling protocol detection from the
   `Accept` stack/type.
2. The `tls::ReasonForNoIdentity` type has been removed, as it provided
   no value; and the `tls::ReasonForNoPeerName` has been updated with
   additional states.
3. We now avoid having to use `PrefixedIo` in most cases by using
   `TcpStream::peek` with a smaller buffer. The prior behavior is
   retained as a fallback in case peek doesn't return enough data.

This comes at the cost of ~2 additional Boxes per connection; but these
will probably be shaved off in a follow-up as the detection logic
becomes more flexible/dynamic.

### DIFF
--- a/linkerd/proxy/transport/src/tls/accept.rs
+++ b/linkerd/proxy/transport/src/tls/accept.rs
@@ -1,22 +1,18 @@
-use super::{conditional_accept, ReasonForNoPeerName};
+use super::{conditional_accept, Conditional, PeerIdentity, ReasonForNoPeerName};
 use crate::io::{BoxedIo, PrefixedIo};
-use crate::listen::{self, Addrs};
+use crate::listen::Addrs;
 use bytes::BytesMut;
 use indexmap::IndexSet;
-use linkerd2_conditional::Conditional;
 use linkerd2_dns_name as dns;
-use linkerd2_error::Error;
 use linkerd2_identity as identity;
-use linkerd2_proxy_core::Accept;
-use linkerd2_stack::layer;
-use pin_project::pin_project;
+use linkerd2_proxy_detect as detect;
 pub use rustls::ServerConfig as Config;
-use std::future::Future;
-use std::pin::Pin;
 use std::sync::Arc;
-use std::task::{Context, Poll};
-use tokio::net::TcpStream;
-use tracing::{debug, trace};
+use tokio::{
+    io::{self, AsyncReadExt},
+    net::TcpStream,
+};
+use tracing::{debug, trace, warn};
 
 pub trait HasConfig {
     fn tls_server_name(&self) -> identity::Name;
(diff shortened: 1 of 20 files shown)


linkerd/proxy/transport/src/tls/accept.rs
```diff
@@ -32,272 +28,157 @@ pub fn empty_config() -> Arc<Config> {
...
+impl<I: HasConfig> DetectTls<I> {
+    // The initial peek buffer is statically allocated on the stack and is fairly small; but it is
+    // large enough to hold the ~300B ClientHello sent by proxies.
+    const PEEK_CAPACITY: usize = 512;
 
-impl<A: Accept<Connection>, T: HasConfig> AcceptTls<A, T> {
-    const PEEK_CAPACITY: usize = 8192;
+    // A larger fallback buffer is allocated onto the heap if the initial peek buffer is
+    // insufficient. This is the same value used in HTTP detection.
+    const BUFFER_CAPACITY: usize = 8192;
```
hawkw, line 50: I wonder if we could come up with a scheme where there's a single heap buffer that's used for both...but this would probably take a much bigger change to pull off. Just a thought.

linkerd/proxy/transport/src/tls/accept.rs
```diff
@@ -32,272 +28,157 @@ pub fn empty_config() -> Arc<Config> {
...
+            peer_identity: Conditional::None(ReasonForNoPeerName::NoTlsFromRemote),
+            addrs,
+        };
+
+        // First, try to use MSG_PEEK to read the SNI from the TLS ClientHello.
+        // Because peeked data does not need to be retained, we use a static
+        // buffer to prevent needless heap allocation.
+        //
+        // Anecdotally, the ClientHello sent by Linkerd proxies is <300B. So a
+        // ~500B byte buffer is more than enough.
```
hawkw, line 98: It would be interesting to do some actual analysis of this in the wild. We could add a debug metric of the number of times we had to spill over into the fallback buffer or something. Definitely not a blocker, just idle curiosity.

linkerd/proxy/transport/src/tls/accept.rs
olix0r: We can if we handle all connections on a single task, but that seems like a poor design.

linkerd/proxy/transport/src/tls/accept.rs
olix0r: We know what the proxy does for its TLS because we can check it. But I don't think the peek approach is reliable if, for instance, only partial data is sent initially. The larger fallback buffer matches the prior behavior as well as the HTTP behavior. I'm not sure what we'd really do with more data. We know this approach helps in the typical case, but there's nothing we can really do about having to support the fallback; and this is already the existing behavior, which works well.

linkerd/proxy/transport/src/tls/accept.rs
```diff
@@ -32,272 +28,151 @@ pub fn empty_config() -> Arc<Config> {
...
+    type Error = io::Error;
 
-    fn call(&mut self, (addrs, socket): listen::Connection) -> Self::Future {
-        // Protocol detection is disabled for the original port. Return a
-        // new connection without protocol detection.
-        let target_addr = addrs.target_addr();
-
-        match &self.tls {
-            // Tls is disabled. Return a new plaintext connection.
+    async fn detect(&self, addrs: Addrs, mut tcp: TcpStream) -> io::Result<(Meta, BoxedIo)> {
```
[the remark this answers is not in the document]
olix0r: I suspect there are much more dominant performance factors, to the extent that these changes won't be discernible. I.e. there are probably better uses of our time.

linkerd/proxy/transport/src/tls/accept.rs
hawkw: Yeah, we definitely don't want that.

linkerd/proxy/transport/src/tls/accept.rs
hawkw: Yeah, I think this approach is right. If there isn't any more tuning we can do with additional data, it's definitely not really worth the extra effort, I was just curious.

PR #608, opened 2020-07-27
Simplify proxy::Server as ServeHttp

olix0r:
The `proxy::Server` type used to handle all protocol detection, but now
it really only handles initializing HTTP services.

This change extracts server-transport instrumentation into a distinct
stack layer to decouple this logic from the server implementation.

The server also now requires that response bodies are boxed to eliminate
the need for a PhantomData.


olix0r:
We should merge #610 first

linkerd/proxy/transport/src/metrics.rs
```diff
@@ -217,23 +224,67 @@ impl<L: Clone, K: Eq + Hash + FmtLabels, M> tower::layer::Layer<M> for ConnectLa
...
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, (target, io): (T, io::BoxedIo)) -> Self::Future {
+        let labels = self.label.transport_labels(&target);
+        let metrics = self
+            .registry
+            .lock()
+            .expect("metrics registr poisoned")
```
kleimkuhler, line 262:
```suggestion
            .expect("metrics registry poisoned")
```

PR #609, opened 2020-07-27
http2: Use adaptive windows

olix0r:
This change removes default window sizes and instead enables hyper's
adaptive window support by default. If environment settings are
provided, they override this default behavior, disabling adaptive
windows.


olix0r:
appears to negatively impact latency in benchmarks...

hawkw:
> appears to negatively impact latency in benchmarks...

probably worth letting @seanmonstar know? i think he was curious to see the results of using adaptive windows in real applications...

PR #611, opened 2020-07-31
add feature-flagged support for multi-threaded Tokio runtime

hawkw:
This branch enables feature-flagged support for using Tokio's
multithreaded runtime when the proxy is given more than two CPU cores to
run on. The proxy's understanding of the system's number of CPUs
reflects cgroups limits, so docker CPU limits are taken into
consideration. 

The multithreaded runtime offers significantly better performance under
load when multiple CPU cores are available. Here's benchmark results
from a run on my 24-thread Ryzen 3900X system, where the proxy, the load
generator, and the test server were all given 8 cores:

```
eliza on noctis in ~
:; nproc --all
24
``` 
Screenshot_20200730_162805

Note that the locking in multiple metrics layers doesn't really seem
to have a big impact, as shown by the run with metrics disabled. There 
might be some improvement in high tails from further optimizations to
the metrics code, though.

If only a single core is available to run a worker on, we will construct
Tokio's `basic_scheduler`, rather than a `threaded_scheduler` with a
single worker thread. This is because there is some overhead introduced
by the threaded scheduler, which results in worse performance when only
1 core is available:

Screenshot_20200730_152202

I'd like to fix this upstream in Tokio, by having `Runtime::new()` and 
the macros intelligently select the basic scheduler when only 1 core 
is available, but this works for now.

Finally, a bug in the handle time metrics layer results in a deadlock
when used with the multithreaded runtime. Therefore, I've disabled this
metric for now --- my understanding is that nothing consumes it at the
moment. If it's needed, we can fix the bug and put it back.

### DIFF
--- a/linkerd2-proxy/src/rt.rs
+++ b/linkerd2-proxy/src/rt.rs
@@ -1,0 +1,31 @@
+use tokio::runtime::{self, Runtime};
+
+#[cfg(feature = "multicore")]
+pub(crate) fn build() -> Runtime {
+    let builder = runtime::Builder::new()
+        .enable_all()
+        .thread_name("linkerd2-proxy-worker");
+    let num_cpus = num_cpus::get();
+    if num_cpus > 2 {
+        builder
+            .threaded_scheduler()
+            .core_threads(num_cpus - 1) // Save 1 core for the admin runtime.
+            .build()
+            .expect("failed to build multithreaded runtime!")
+    } else {
+        builder
+            .basic_scheduler()
+            .build()
+            .expect("failed to build single-threaded runtime!")
+    }
+}
+
+#[cfg(not(feature = "multicore"))]
+pub(crate) fn build() -> Runtime {
+    runtime::Builder::new()
+        .enable_all()
+        .thread_name("linkerd2-proxy-worker")
+        .basic_scheduler()
+        .build()
+        .expect("failed to build single-threaded runtime!")
+}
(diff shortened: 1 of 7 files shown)


linkerd2-proxy/src/rt.rs
```diff
@@ -0,0 +1,31 @@
...
+#[cfg(feature = "multicore")]
+pub(crate) fn build() -> Runtime {
+    let builder = runtime::Builder::new()
+        .enable_all()
+        .thread_name("linkerd2-proxy-worker");
+    let num_cpus = num_cpus::get();
+    if num_cpus > 2 {
+        builder
+            .threaded_scheduler()
+            .core_threads(num_cpus - 1) // Save 1 core for the admin runtime.
```
olix0r, line 12: I don't think we should dedicate an entire core to the admin thread... That's *a lot* of resources for something we shouldn't be doing much work on. The real goal of the separate admin runtime is to ensure we can still interact with the proxy if, for instance, the main runtime isn't healthy. I think we're best off keeping it simple and treating it as if the admin thread needs no dedicated resources.

hawkw: Mmm, yeah, that makes sense. My thought process was that if forwarding has more workers, load on the admin runtime for e.g. discovery & DNS lookups might increase. I think the long term solution might be to just move that work off of the admin runtime, and have it _only_ run metrics/tap/trace endpoints/etc. Then we can probably feel confident not giving it a whole core.

PR #612, opened 2020-07-31
Add `multicore` to default features

olix0r:
We don't have any blockers to using the threaded runtime by default.
This change adds the `multicore` feature flag to the default feature
list.

It also modifies the threaded runtime initialization to use all
available cores.

The thread names have been changed from `linkerd2-proxy-worker` to just
`proxy`, which is more consistent with the `admin` thread name. This
should be unambiguous in the context of the process.

### DIFF
--- a/linkerd2-proxy/src/rt.rs
+++ b/linkerd2-proxy/src/rt.rs
@@ -18,6 +18,7 @@ pub(crate) fn build() -> Runtime {
             .thread_name("proxy")
             .threaded_scheduler()
             .core_threads(num_cpus)
+            .max_threads(num_cpus)
             .build()
             .expect("failed to build threaded runtime!"),
     }
@@ -30,5 +31,5 @@ pub(crate) fn build() -> Runtime {
         .thread_name("proxy")
         .basic_scheduler()
         .build()
-        .expect("failed to build threaded runtime!")
+        .expect("failed to build basic runtime!")
 }
--- a/linkerd2-proxy/src/rt.rs
+++ b/linkerd2-proxy/src/rt.rs
@@ -1,31 +1,34 @@
-use tokio::runtime::{self, Runtime};
+use tokio::runtime::{Builder, Runtime};
 
 #[cfg(feature = "multicore")]
 pub(crate) fn build() -> Runtime {
-    let builder = runtime::Builder::new()
-        .enable_all()
-        .thread_name("linkerd2-proxy-worker");
-    let num_cpus = num_cpus::get();
-    if num_cpus > 2 {
-        builder
-            .threaded_scheduler()
-            .core_threads(num_cpus - 1) // Save 1 core for the admin runtime.
-            .build()
-            .expect("failed to build multithreaded runtime!")
-    } else {
-        builder
+    // The proxy creates an additional admin thread, but it would be wasteful to allocate a whole
+    // core to it; so we let the main runtime consume all cores the process has. The basic scheduler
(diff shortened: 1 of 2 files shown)


linkerd2-proxy/src/rt.rs
```diff
@@ -1,31 +1,34 @@
-use tokio::runtime::{self, Runtime};
+use tokio::runtime::{Builder, Runtime};
 
 #[cfg(feature = "multicore")]
 pub(crate) fn build() -> Runtime {
-    let builder = runtime::Builder::new()
-        .enable_all()
-        .thread_name("linkerd2-proxy-worker");
```
olix0r, line 7: afaict this didn't actually compile this way

linkerd2-proxy/src/rt.rs
```diff
@@ -1,31 +1,34 @@
...
+            .thread_name("proxy")
             .basic_scheduler()
             .build()
-            .expect("failed to build single-threaded runtime!")
+            .expect("failed to build basic runtime!"),
+        num_cpus => Builder::new()
+            .enable_all()
+            .thread_name("proxy")
+            .threaded_scheduler()
+            .core_threads(num_cpus)
```
olix0r, line 20: and... I think maybe we should be using max_threads here?

hawkw: we should probably set `core_threads` and `max_threads` equal to the same number. tokio will only spawn more than `core_threads` threads when using the blocking pool. AFAICT, we are not enabling Tokio's `blocking` or `fs` features, so no blocking pool is created; however, we may want to ensure that no blocking pool is created if even if a dependency enables those feature flags. 

i *believe* that if `core_threads` == `max_threads` and the `blocking` feature _is_ enabled, calls to `tokio::spawn_blocking` or `tokio::block_in_place` will fail (possibly by panicking? would have to check). this shouldn't be a problem for us.

linkerd2-proxy/src/rt.rs
```diff
@@ -1,31 +1,34 @@
...
-        builder
-            .threaded_scheduler()
-            .core_threads(num_cpus - 1) // Save 1 core for the admin runtime.
-            .build()
-            .expect("failed to build multithreaded runtime!")
-    } else {
-        builder
+    // The proxy creates an additional admin thread, but it would be wasteful to allocate a whole
+    // core to it; so we let the main runtime consume all cores the process has. The basic scheduler
+    // is used when the threaded scheduler would provide no benefit.
```
hawkw, line 7: <3

linkerd2-proxy/src/rt.rs
```diff
@@ -1,31 +1,34 @@
...
 pub(crate) fn build() -> Runtime {
-    runtime::Builder::new()
+    Builder::new()
         .enable_all()
-        .thread_name("linkerd2-proxy-worker")
+        .thread_name("proxy")
         .basic_scheduler()
         .build()
-        .expect("failed to build single-threaded runtime!")
+        .expect("failed to build threaded runtime!")
```
hawkw, line 33: nit: this builds the basic scheduler runtime, we should probably change the panic message:
```suggestion
        .expect("failed to build basic runtime!")
```

linkerd2-proxy/src/rt.rs
hawkw:
```suggestion
            .core_threads(num_cpus)
            .max_threads(num_cpus)
```

olix0r:
image

PR #614, opened 2020-07-31
[WIP] use `tracing-flame` to generate flamegraphs of tracing spans

hawkw:
To get more value from this, we probably need to add more verbose spans at the `debug` and `trace` levels that are normally filtered out. But, it's a cool proof of concept.

Depends on #595 because it would definitely conflict otherwise.


olix0r:
@hawkw Do we want to revive this branch? Or should we close it? Last I checked, it wasn't really practical to run the proxy with tokio's tracing enabled.

olix0r:
This branch is pretty outdated at this point. Closing due to inactivity.

PR #615, opened 2020-07-31
trace: roll `tracing` and `tracing-subscriber` dependencies

hawkw:
This picks up upstream changes tokio-rs/tracing#853,
tokio-rs/tracing#868, and tokio-rs/tracing#869 which improve performance
in some use cases. The overhead removed by these changes may already be
amortized enough in the proxy that it's not a problem, but it seems
worth picking up regardless.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2706,9 +2706,9 @@ dependencies = [
 
 [[package]]
 name = "tracing"
-version = "0.1.15"
+version = "0.1.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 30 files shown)


hawkw:
I'm doing a bench run with this because I'm curious to see if we benefit from it as much as the Rust compiler did --- it may improve tails a bit!

hawkw:
Turns out that the upstream change, although it had about a 30% performance improvement in `rustc`, doesn't really make a difference in the proxy, which is super interesting:
Screenshot_20200731_164111
(worse tails on the branch are probably just me logging into my computer to check if the benchmark was still running).

Oh well. We should probably pick up the dependency update anyway.

PR #619, opened 2020-08-05
Move HTTP detection & server into linkerd2_proxy_http

olix0r:
In preparation for upcoming simplifications, this moves the HTTP
detection/serving logic into the linkerd2_proxy_http crate.

### DIFF
--- a/linkerd/app/core/src/proxy/mod.rs
+++ b/linkerd/app/core/src/proxy/mod.rs
@@ -4,16 +4,8 @@ pub use linkerd2_proxy_api_resolve as api_resolve;
 pub use linkerd2_proxy_core as core;
 pub use linkerd2_proxy_detect as detect;
 pub use linkerd2_proxy_discover as discover;
-pub use linkerd2_proxy_http::{
-    self as http,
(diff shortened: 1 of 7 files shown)


linkerd/app/core/src/proxy/mod.rs
```diff
@@ -4,16 +4,8 @@ pub use linkerd2_proxy_api_resolve as api_resolve;
...
-};
+pub use linkerd2_proxy_http as http;
 pub use linkerd2_proxy_identity as identity;
 pub use linkerd2_proxy_resolve as resolve;
 pub use linkerd2_proxy_tap as tap;
 pub use linkerd2_proxy_tcp as tcp;
-
-pub mod server;
-
-pub use self::server::ServeHttp;
```
hawkw, line 19: Are there dependencies in `linkerd2-app-core`'s `Cargo.toml` that are no longer needed now that this code has moved? If so, can they be removed?

olix0r: More changes coming. We'll clean them up later.

PR #620, opened 2020-08-05
Introduce a SkipDetect layer to preempt detection

olix0r:
This change introduces a new SkipDetect layer that configures whether
protocol detection should be attempted. This module will later be
replaced/augmented by discovery.

Furthermore, this change eliminates the `Accept` trait. Instead of
modeling the accept stack as a simple service whose response is a
future, we now model the stack as, effectively, a
`MakeService<Meta, TcpStream>`. This is intended to support caching of
the service that handles the tcp stream (i.e. to hold discovery
responses).

This change also removes the `Detect` trait. It's no longer useful.

Detection timeouts have been moved from a dedicated layer into the
detection modules.


hawkw:
Looks like the [tap test failed](https://github.com/linkerd/linkerd2-proxy/runs/950455155#step:3:759), but that may be flakiness? Restarting it.

linkerd/app/core/src/proxy/skip_detect.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn call(&mut self, io: T) -> Self::Future {
+        match self {
+            Self::Detect(d) => {
+                let fut = d.call(io).err_into::<Error>();
+                Box::pin(async move { fut.await })
+            }
+            Self::Tcp(f) => {
+                let fut = f.call(io).err_into::<Error>();
+                Box::pin(async move { fut.await })
+            }
```
hawkw, line 106: nit: can't these just be 
```suggestion
                let fut = d.call(io).err_into::<Error>();
                Box::pin(fut.await)
            }
            Self::Tcp(f) => {
                let fut = f.call(io).err_into::<Error>();
                Box::pin(fut)
            }
```

linkerd/app/core/src/serve.rs
```diff
@@ -1,66 +1,63 @@
...
+                            match accept.err_into::<Error>().await {
+                                Err(error) => error!(%error, "Failed to dispatch connection"),
+                                Ok(accept) => match accept.oneshot(io).err_into::<Error>().await {
+                                    Ok(()) => debug!("Connection closed"),
+                                    Err(error) => info!(%error, "Connection closed"),
                                 },
-                                Err(e) => match e {},
                             }
                         }
                         .in_current_span(),
```
hawkw, line 63: We shouldn't use `Span::enter` in cases where the `Entered` guard is held across an await point (see [here](https://docs.rs/tracing/0.1.18/tracing/span/struct.Span.html#in-asynchronous-code) for details).

Instead, we should write something like this:
```rust
                    let span = info_span!(
                        "accept",
                        peer.addr = %addrs.peer(),
                    ); 

                    // Ready the service before dispatching the request to it.
                    //
                    // This allows the service to propagate errors and to exert backpressure on the
                    // listener. It also avoids a `Clone` requirement.
                    let accept = make_accept
                        .ready_and()
                        .err_into::<Error>()
                        .instrument(span.clone())
                        .await?
                        .call(addrs);


                    // Dispatch all of the work for a given connection onto a connection-specific task.
                    tokio::spawn(
                        async move {
                            match accept.err_into::<Error>().await {
                                Err(error) => error!(%error, "Failed to dispatch connection"),
                                Ok(accept) => match accept.oneshot(io).err_into::<Error>().await {
                                    Err(error) => info!(%error, "Connection closed"),
                                },
                            }
                        }
                        .instrument(span),
                    );
```
(not a suggested change because this contains deleted lines & github won't let me suggest changes).

linkerd/app/src/lib.rs
```diff
@@ -155,11 +155,11 @@ impl Config {
         let oc_span_sink = oc_collector.span_sink();
 
         let start_proxy = Box::pin(async move {
+            let refine = outbound.build_dns_refine(resolver, &outbound_metrics.stack);
+
             let outbound_connect =
                 outbound.build_tcp_connect(local_identity.clone(), &outbound_metrics);
 
-            let refine = outbound.build_dns_refine(resolver, &outbound_metrics.stack);
-
```
hawkw, line 162: Is there a reason this moved? Not a blocker, just curious.

linkerd/app/src/tap.rs
```diff
@@ -49,9 +51,20 @@ impl Config {
...
-                    tls::DetectTls::new(identity, Default::default()),
-                    tap::AcceptPermittedClients::new(permitted_peer_identities.into(), server),
+                let service =
+                    tap::AcceptPermittedClients::new(permitted_peer_identities.into(), server);
+                let accept = tls::DetectTls::new(
+                    identity,
+                    service_fn(move |meta: tls::accept::Meta| {
+                        let service = service.clone();
+                        future::ok::<_, Never>(service_fn(move |io: io::BoxedIo| {
+                            let fut = service.clone().oneshot((meta.clone(), io));
```
hawkw, line 61: Nit: do we need to clone `meta` here? isn't it already being moved into this closure from the outer one?

linkerd/proxy/http/src/detect.rs
```diff
@@ -85,45 +41,108 @@ impl detect::Detect<tls::accept::Meta, BoxedIo> for DetectHttp {
...
+        let drain = self.drain.clone();
+        let tcp = self.tcp.clone();
+        let http = self.http.clone();
+        let server = self.server.clone();
+        let timeout = self.timeout;
+
+        Box::pin(async move {
+            let (tcp, http) = futures::try_join!(
+                tcp.oneshot(target.clone()).map_err(Into::<Error>::into),
+                http.oneshot(target).map_err(Into::<Error>::into)
```
hawkw, line 103: nit/TIOLI:
```suggestion
                tcp.oneshot(target.clone()).err_into::<Error>(),
                http.oneshot(target).err_into::<Error>(),
```
maybe slightly more concise. not a blocker though.

linkerd/proxy/transport/src/metrics.rs
```diff
@@ -241,30 +246,53 @@ where
...
     }
 
-    fn call(&mut self, (target, io): (T, io::BoxedIo)) -> Self::Future {
+    fn call(&mut self, target: T) -> Self::Future {
         let labels = self.label.transport_labels(&target);
         let metrics = self
             .registry
             .lock()
-            .expect("metrics registry poisoned")
+            .expect("metrics registr poisoned")
```
hawkw, line 269: nit:
```suggestion
            .expect("metrics registry poisoned")
```

olix0r:
@hawkw tests pass locally...

linkerd/app/src/env.rs
```diff
@@ -169,10 +169,6 @@ const ENV_DNS_MIN_TTL: &str = "LINKERD2_PROXY_DNS_MIN_TTL";
 /// Lookups with TTLs above this value will use this value instead.
 const ENV_DNS_MAX_TTL: &str = "LINKERD2_PROXY_DNS_MAX_TTL";
 
-/// The amount of time to wait for a DNS query to succeed before falling back to
-/// an uncanonicalized address.
-const ENV_DNS_CANONICALIZE_TIMEOUT: &str = "LINKERD2_PROXY_DNS_CANONICALIZE_TIMEOUT";
```
hawkw, line 174: does this mean that refinements will no longer time out?

linkerd/app/src/tap.rs
olix0r: FnOnce vs FnMut

linkerd/proxy/http/src/detect.rs
olix0r: iirc that didn't work

linkerd/app/src/env.rs
olix0r: This snuck in from a followup and should have been reverted

linkerd/app/src/tap.rs
hawkw: oh whoops, yeah. nevermind!

linkerd/app/core/src/proxy/skip_detect.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn call(&mut self, io: T) -> Self::Future {
+        match self {
+            Self::Detect(d) => {
+                let fut = d.call(io).err_into::<Error>();
+                Box::pin(fut)
+            }
+            Self::Tcp(f) => {
+                let fut = f.call(io).err_into::<Error>();
+                Box::pin(fut)
+            }
```
hawkw, line 106: occurs to me that the `let fut` binding could also be removed:
```suggestion
            Self::Detect(d) => Box::pin(d.call(io).err_into::<Error>()),
            Self::Tcp(f) => Box::pin(f.call(io).err_into::<Error>()),
```
but at this point it is definitely just a matter of taste...

linkerd/app/core/src/serve.rs
```diff
@@ -1,69 +1,66 @@
...
 
                     // Ready the service before dispatching the request to it.
                     //
                     // This allows the service to propagate errors and to exert backpressure on the
                     // listener. It also avoids a `Clone` requirement.
-                    let accept = service.ready_and().await?.call(conn);
+                    let accept = make_accept
+                        .ready_and()
+                        .err_into::<Error>()
+                        .instrument(span.clone())
```
hawkw, line 48: :+1:

linkerd/proxy/http/src/detect.rs
```diff
@@ -85,45 +41,108 @@ impl detect::Detect<tls::accept::Meta, BoxedIo> for DetectHttp {
...
+            })
+        })
+    }
+}
+
+// === impl AcceptHttp ===
+
+impl<F, H> AcceptHttp<F, H> {
+    /// Creates a new `AcceptHttp`.
+    pub fn new(server: Server, timeout: Duration, http: H, tcp: F, drain: drain::Watch) -> Self {
```
kleimkuhler, line 121: Do you think this will be used in upcoming changes? It's unused right now.

olix0r: Now used

PR #622, opened 2020-08-05
add richer verbose spans to HTTP clients

hawkw:
This branch adds some DEBUG- and TRACE-level spans to HTTP clients and
client connections. These are hidden by default but may be useful when
debugging, or for generating tracing flamegraphs (as in PR #614).

### DIFF
--- a/linkerd/proxy/http/src/h2.rs
+++ b/linkerd/proxy/http/src/h2.rs
@@ -13,8 +13,8 @@ use std::marker::PhantomData;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 use tokio::io::{AsyncRead, AsyncWrite};
-use tracing::{debug, info_span};
-use tracing_futures::Instrument;
+use tracing::{debug, info_span, trace_span};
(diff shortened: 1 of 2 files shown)


linkerd/proxy/http/src/h2.rs
```diff
@@ -157,15 +166,18 @@ where
                         )
                         .executor(trace::Executor::new())
                         .handshake(io)
-                        .instrument(info_span!("h2"));
+                        .instrument(trace_span!("handshake"))
+                        .instrument(span.clone());
```
olix0r, line 170: Why isn't this necessary? haven't we already entered it?

hawkw: If the `handshake` span is enabled, then the handshake future will be in the outer `h2` span by virtue of `h2` being the parent of `handshake`. However, if the `handshake` span is _not_ enabled (and it's at the TRACE level, so it probably won't be), entering it will do nothing, so the future will not be guaranteed to be in the outer span in that case.

PR #625, opened 2020-08-15
Wait for proxy thread to terminate in integration tests

tustvold:
Whilst implementing #601 I encountered non-determinism due to the way the integration tests don't wait for termination of the proxy thread. The changes in this PR make join_servers explicitly signal shutdown, then wait on a oneshot channel for the thread to signal draining has completed, before finally joining the thread. Simply joining the thread after signaling shutdown is insufficient as this will block the "primary" tokio runtime preventing graceful shutdown of HTTP2 connections, which rely on the clients running in the "primary" tokio runtime to be responsive.

Unfortunately the proxy appears to intentionally wait for TCP upstreams to terminate first, which requires explicit termination of the TCP client in these tests.

### DIFF
--- a/linkerd/app/integration/src/proxy.rs
+++ b/linkerd/app/integration/src/proxy.rs
@@ -41,7 +41,8 @@ pub struct Listening {
     controller: controller::Listening,
     identity: Option<controller::Listening>,
 
-    _shutdown: Shutdown,
+    shutdown: Shutdown,
+    terminated: oneshot::Receiver<()>,
 
     thread: thread::JoinHandle<()>,
 }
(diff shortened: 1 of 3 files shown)


tustvold:
CI seemed to take a rather long time, so ran some tests locally

With the change, and the tests already built

```
Executed in   16.26 secs   fish           external 
   usr time   64.59 secs   64.27 millis   64.52 secs 
   sys time    3.39 secs    9.85 millis    3.38 secs 
```

On main, with the tests already built

```
Executed in   16.75 secs   fish           external 
   usr time   64.57 secs  1010.00 micros   64.57 secs 
   sys time    3.64 secs  363.00 micros    3.64 secs 
```

So at least on my machine it doesn't seem to measurably impact the time the tests take to execute

olix0r:
Thanks @tustvold!

PR #626, opened 2020-08-17
profiles: perform profile resolution for IP addresses

hawkw:
linkerd/linkerd2#3916 added support for resolving service profiles for
IP addresses to the Destination service. This branch updates the proxy's
profiles client to look up profiles for IP addresses, rather than always
rejecting addresses that are IPs rather than DNS names.

Similarly to the Destination service-discovery client, a new
`LINKERD2_PROXY_DESTINATION_PROFILE_NETWORKS` enviroment variable is
used to configure a list of subnets to match which IPs the proxy will
look up profiles for. By default, this is empty. Since the logic for
filtering requests to a service based on IPs and DNS prefixes is now
identical between the profile and destination clients, I factored it out
into a new layer that's used for both.

Fixes linkerd/linkerd2#4877

### DIFF
--- a/linkerd/app/core/src/svc.rs
+++ b/linkerd/app/core/src/svc.rs
@@ -1,9 +1,8 @@
 // Possibly unused, but useful during development.
 
 pub use crate::proxy::http;
-use crate::request_filter;
 use crate::transport::Connect;
-use crate::{cache, Error};
+use crate::{cache, request_filter, Error};
 pub use linkerd2_buffer as buffer;
 use linkerd2_concurrency_limit as concurrency_limit;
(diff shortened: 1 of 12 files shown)


linkerd/app/core/src/svc.rs
```diff
@@ -1,6 +1,7 @@
 // Possibly unused, but useful during development.
 
 pub use crate::proxy::http;
+use crate::request_filter;
 use crate::transport::Connect;
 use crate::{cache, Error};
```
olix0r, line 6: nit
```suggestion
use crate::transport::Connect;
use crate::{cache, request_filter, Error};
```

PR #633, opened 2020-08-22
resolve: Add a Reset state

olix0r:
The recovery layer needs to maintain its own cache for reconcilation (in
addition to the other cache layers we keep). This caching can be
eliminated by introducing a resolve::Update::Reset state that implicitly
remotes all endpoints not present in the endpoint set.

### DIFF
--- a/linkerd/proxy/resolve/src/recover.rs
+++ b/linkerd/proxy/resolve/src/recover.rs
@@ -2,12 +2,10 @@
 
 use futures::stream::TryStream;
 use futures::{prelude::*, ready, FutureExt, Stream};
-use indexmap::IndexMap;
 use linkerd2_error::{Error, Recover};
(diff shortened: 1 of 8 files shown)


zaharidichev (approved):
LGTM

linkerd/proxy/resolve/src/recover.rs
```diff
@@ -316,11 +254,24 @@ where
...
+                        continue;
+                    }
+                    Some(Ok(update)) => {
+                        tracing::debug!("Connected");
+                        // If the first update is an Add; convert it to a Reset, indicating all
+                        // other endpoints should be removed.
+                        let update = match update {
+                            Update::Add(eps) => Update::Reset(eps),
+                            up => up,
+                        };
```
hawkw, line 271: this is much clearer, thanks!

PR #634, opened 2020-08-22
resolve: Eagerly fail resolutions

olix0r:
In 591191c, the initial resolution was moved into the stream. This
prevents the initial future from failing with an error when the control
plane fails eagerly (as is the case with InvalidArgument errors).

To preserve the eager failure, the stream is not returned until an
initial resolution has been completed. Then, we are able to remove
the workaround code from `recover`.


zaharidichev (approved):
LGTM

PR #637, opened 2020-08-25
dns: Run DNS resolutions on the main runtime

olix0r:
DNS resolutions are run on the admin runtime. This requires an
unnecessary layer of indirection around the resolver, including an MPSC.

Now that we allow the main runtime to use more than one thread, it's
preferable to do this discovery on the main runtime and we can simplify
the implementation.


zaharidichev (approved):
LGTM

PR #638, opened 2020-08-26
Unify control plane client construction

olix0r:
We construct three control plane clients. All of the clients are
basically the same (at least with regard to the stack/configuration).
Rather than duplicate this logic throughout the code, we can unify it in
the `control` module (hiding the details of how the client is built).

This change modifies how the OpenCensus collector client is
instantiated. The `SpanExporter` no longer constructs a new client each
time it wishes to initiate a stream, opting to just clone a single
client instead.

This change also fixes reconnect logic for these clients. When the
control client balancer was introduced, the reconnect layer incorrectly
wrapped the balancer (instead of the endpoint). This has been corrected.

### DIFF
--- a/linkerd/app/core/src/control.rs
+++ b/linkerd/app/core/src/control.rs
@@ -1,10 +1,25 @@
-use linkerd2_addr::Addr;
+use crate::{
+    classify, config, control, dns,
+    proxy::http,
+    reconnect,
+    svc::{self, NewService},
+    transport::tls,
+    Addr, ControlHttpMetrics, Error,
+};
+use linkerd2_buffer::Buffer;
 use std::fmt;
+
+#[derive(Clone, Debug)]
+pub struct Config {
+    pub addr: ControlAddr,
+    pub connect: config::ConnectConfig,
+    pub buffer_capacity: usize,
+}
 
 #[derive(Clone, Debug)]
 pub struct ControlAddr {
     pub addr: Addr,
-    pub identity: crate::transport::tls::PeerIdentity,
+    pub identity: tls::PeerIdentity,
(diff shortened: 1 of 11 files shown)


linkerd/app/core/src/control.rs
```diff
@@ -19,8 +34,54 @@ impl fmt::Display for ControlAddr {
...
+    http::Request<B>,
+    http::Response<
+        linkerd2_http_metrics::requests::ResponseBody<
+            http::balance::PendingUntilFirstDataBody<
+                tower::load::peak_ewma::Handle,
+                http::glue::Body,
+            >,
+            classify::Eos,
+        >,
+    >,
```
zaharidichev, line 47: Nit: Is it worth importing `http::balance`, `tower::load` and `http::glue` to potentially make this formatting less weird ?

zaharidichev (approved):
LGTM

linkerd/app/core/src/control.rs
```diff
@@ -19,8 +33,48 @@ impl fmt::Display for ControlAddr {
...
+            }))
+            .push_spawn_ready()
+            .push(self::resolve::layer(dns))
+            .push_on_response(self::control::balance::layer())
+            .push(metrics.into_layer::<classify::Response>())
+            .push(self::add_origin::Layer::new())
+            .into_new_service()
+            .check_new_service()
+            .push_on_response(svc::layers().push_spawn_buffer(self.buffer_capacity))
+            .new_service(self.addr)
```
hawkw, line 72: should we maybe stick an `instrument` in this stack somewhere, or does that happen elsewhere?

PR #639, opened 2020-08-27
Recognize NXDomain Errors

zaharidichev:
Unfrotunatelly due to https://github.com/bluejekyll/trust-dns/issues/1171 we can not see NoRecordsFound errors. There is not a stong type we can rely on and therefore we need to read the actual proto message to determine that we have gotten a negative resolution and return an empty list of IPs

### DIFF
--- a/linkerd/dns/src/lib.rs
+++ b/linkerd/dns/src/lib.rs
@@ -101,9 +101,7 @@ impl Resolver {
                 }
                 ResolveErrorKind::Proto(pe) if self.is_nx_domain(&pe) => Ok((
                     vec![],
-                    time::delay_until(Instant::from_std(
-                        std::time::Instant::now() + Self::DEFAULT_TTL,
-                    )),
+                    time::delay_for(Self::DEFAULT_TTL)),
                 )),
 
                 _ => Err(e),
@@ -112,10 +110,10 @@ impl Resolver {
     }
 
     fn is_nx_domain(&self, pe: &error::ProtoError) -> bool {
-        return match pe.kind() {
+        match pe.kind() {
             error::ProtoErrorKind::Message(msg) => *msg == "Nameserver responded with NXDomain",
             _ => false,
-        };
+        }
     }
 
     async fn resolve_srv(&self, name: &Name) -> Result<(Vec<net::SocketAddr>, time::Delay), Error> {
@@ -138,9 +136,7 @@ impl Resolver {
                 }
                 ResolveErrorKind::Proto(pe) if self.is_nx_domain(&pe) => Ok((
                     vec![],
-                    time::delay_until(Instant::from_std(
-                        std::time::Instant::now() + Self::DEFAULT_TTL,
-                    )),
+                    time::delay_for(Self::DEFAULT_TTL)),
                 )),
                 _ => Err(e.into()),
             },
--- a/linkerd/dns/src/lib.rs
+++ b/linkerd/dns/src/lib.rs
@@ -30,7 +30,10 @@ pub trait ConfigureResolver {
 struct InvalidSrv(rdata::SRV);
 
 impl Resolver {
-    const DEFAULT_TTL: std::time::Duration = std::time::Duration::from_secs(60);
+    // When the DNS library does not return a TTL, we must assume one to prevent
+    // tight-looping. In practice, default kubernetes configs appear to have a
+    // 5s TTL, so we mirror that default here.
+    const DEFAULT_TTL: std::time::Duration = std::time::Duration::from_secs(5);
 
     /// Construct a new `Resolver` from environment variables and system
     /// configuration.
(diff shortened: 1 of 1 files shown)


linkerd/dns/src/lib.rs
```diff
@@ -98,11 +99,25 @@ impl Resolver {
                         .unwrap_or_else(|| std::time::Instant::now() + Self::DEFAULT_TTL);
                     Ok((vec![], time::delay_until(Instant::from_std(expiry))))
                 }
+                ResolveErrorKind::Proto(pe) if self.is_nx_domain(&pe) => Ok((
+                    vec![],
+                    time::delay_until(Instant::from_std(
+                        std::time::Instant::now() + Self::DEFAULT_TTL,
```
zaharidichev, line 105: I think this TTL should be either a lot smaller or be configurable. Seems to cause a pretty long delay when linkerd is firing up and the identitiy pod (for example) has not come up yet. This causes us to get NXDomain once and wait for another 60 seconds before we attempt another resolution.

linkerd/dns/src/lib.rs
```diff
@@ -98,11 +99,25 @@ impl Resolver {
...
+                    )),
```
hawkw, line 106: can't this just be
```suggestion
                    time::delay_for(Self::DEFAULT_TTL)),
```

linkerd/dns/src/lib.rs
```diff
@@ -98,11 +99,25 @@ impl Resolver {
...
                 _ => Err(e),
             },
         }
     }
 
+    fn is_nx_domain(&self, pe: &error::ProtoError) -> bool {
+        return match pe.kind() {
+            error::ProtoErrorKind::Message(msg) => *msg == "Nameserver responded with NXDomain",
+            _ => false,
+        };
```
hawkw, line 118: nit: the `return` is not necessary, this can just be
```suggestion
        match pe.kind() {
            error::ProtoErrorKind::Message(msg) => *msg == "Nameserver responded with NXDomain",
            _ => false,
        }
```

linkerd/dns/src/lib.rs
```diff
@@ -98,11 +99,25 @@ impl Resolver {
...
+                        std::time::Instant::now() + Self::DEFAULT_TTL,
+                    )),
+                )),
+
                 _ => Err(e),
             },
         }
     }
 
+    fn is_nx_domain(&self, pe: &error::ProtoError) -> bool {
```
hawkw, line 114: nit: is there a reason this needs to be a method taking `self`, rather than just
```suggestion
    fn is_nx_domain(pe: &error::ProtoError) -> bool {
```

linkerd/dns/src/lib.rs
```diff
@@ -121,6 +136,12 @@ impl Resolver {
                         .unwrap_or_else(|| std::time::Instant::now() + Self::DEFAULT_TTL);
                     Ok((vec![], time::delay_until(Instant::from_std(expiry))))
                 }
+                ResolveErrorKind::Proto(pe) if self.is_nx_domain(&pe) => Ok((
+                    vec![],
+                    time::delay_until(Instant::from_std(
+                        std::time::Instant::now() + Self::DEFAULT_TTL,
+                    )),
```
hawkw, line 143: again, could be:
```suggestion
                    time::delay_for(Self::DEFAULT_TTL)),
```

linkerd/dns/src/lib.rs
olix0r: Have you tested this and observed this? It's my understanding that the service should always exist...

hawkw: The NXDomain response _should_ include a TTL for the response provided by the DNS server, which `trust-dns-resolver` [would expose on the `NoRecordsFound` error](https://docs.rs/trust-dns-resolver/0.19.5/trust_dns_resolver/error/enum.ResolveErrorKind.html#variant.NoRecordsFound)...but now, that's no longer being returned when it should be. It seems like an ideal solution for this would be to just fix the upstream bug so that `trust-dns-resolver` once again returns the correct error type.

I've worked with Ben (the `trust-dns` maintainer) in the past, and he's a nice guy & very receptive to contributions, so it shouldn't be hard to get upstream fixes for this landed.

zaharidichev: @olix0r 

yes I have tested it . We get NXDomain even though the service exists. It happens when it exists and there are no pods behind it: 

```
[     0.7589800s] DEBUG ThreadId(13) identity: linkerd2_dns: resolve_srv name=linkerd-identity.linkerd.svc.cluster.local
[     0.587119800s] TRACE ThreadId(12) identity: linkerd2_dns: Got no records <-------- The identitiy service pod has not started yet
[    60.588428400s] DEBUG ThreadId(12) identity: linkerd2_dns: resolve_srv name=linkerd-identity.linkerd.svc.cluster.local
[    60.957565500s] DEBUG ThreadId(12) identity: linkerd2_dns: addrs=[V4(10.244.0.2:8080)]
[    90.959252100s] DEBUG ThreadId(12) identity: linkerd2_dns: resolve_srv name=linkerd-identity.linkerd.svc.cluster.local
[    91.309806000s] DEBUG ThreadId(11) identity: linkerd2_dns: addrs=[V4(10.244.0.2:8080)
```]

Comment (olix0r) [reply linkerd/dns/src/lib.rs:105]: OK. makes sense.
```
:; k create service clusterip testsvc --clusterip=None
service/testsvc created
:; k run --image ubuntu:latest -it test
...
root@test:/# dig testsvc.default.svc.cluster.local.

; <<>> DiG 9.16.1-Ubuntu <<>> testsvc.default.svc.cluster.local.
;; global options: +cmd
;; Got answer:
;; WARNING: .local is reserved for Multicast DNS
;; You are currently testing what happens when an mDNS query is leaked to DNS
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 13631
;; flags: qr aa rd; QUERY: 1, ANSWER: 0, AUTHORITY: 1, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
; COOKIE: e1611b1d1cebf509 (echoed)
;; QUESTION SECTION:
;testsvc.default.svc.cluster.local. IN	A

;; AUTHORITY SECTION:
cluster.local.		5	IN	SOA	ns.dns.cluster.local. hostmaster.cluster.local. 1598545810 7200 1800 86400 5

;; Query time: 0 msec
;; SERVER: 10.43.0.10#53(10.43.0.10)
;; WHEN: Thu Aug 27 09:30:3
```

From this, it looks like the default TTL is 5s. so we should use that.

PR #641, opened 2020-08-29
outbound: Cache balancers within profile stack

olix0r:
In an upcoming change, we'd like to do service profile discovery within
the TCP-accept stack (and not the HTTP routing stack). But, today,
service profile discovery is tightly coupled to the HTTP middleware
implementations. This change splits the service profile layers (profile
discovery, http request profiles, and traffic splitting) into several
layers, so that the discovery logic is decoupled from the http-specific
request-routing middleware.

This change removes the balancer cache and the balancer-specific buffer
so that balancers are owned by the split layer. The buffer has been
moved outside of the split layer to drive all balancers in a split and
to make the split cloneable (for the retry middleware). All of this is
cached under the profile cache.

Breaking changes:

This likely breaks support for "external" service profiles, where the
proxy resolves service profilse for domains that it cannot resolve via
the resolver. This feature is not enabled by default and is considered
experimental. We'll have to take attention to at least document this in
upcoming releases.

Side-effects:

* Latency improvements at concurrency image
* Compile time improved by 20-25% (integration tests in CI ~20m => ~15m);


linkerd/app/outbound/src/lib.rs
```diff
@@ -272,27 +268,72 @@ impl Config {
...
+                    // requests.
+                    .push_failfast(dispatch_timeout)
+                    .push(metrics.stack.layer(stack_labels("balance"))),
+            )
             .into_new_service()
+            .instrument(|c: &Concrete<http::Settings>| info_span!("balance", addr = %c.addr))
+            .check_new_service::<Concrete<http::Settings>, http::Request<_>>();
+
+        // Routes `Logical` targets to a cached `Profile` stack, i.e. so that profile
+        // resolutions are shared even as the type of request may vary.
```
hawkw, line 284: again, maybe worth a quick explanation of how the receiver/addr moves through the stack here?

linkerd/app/outbound/src/lib.rs
```diff
@@ -272,27 +268,72 @@ impl Config {
...
+                    // Sets an optional request timeout.
+                    .push(http::MakeTimeoutLayer::default())
+                    // Records per-route metrics.
+                    .push(metrics.http_route.into_layer::<classify::Response>())
+                    // Sets the per-route response classifier as a request
+                    // extension.
+                    .push(classify::Layer::new())
+                    .push_map_target(|(r, l): (profiles::http::Route, Logical<HttpEndpoint>)| {
+                        endpoint::route(r, l)
+                    })
```
hawkw, line 309: nit, again: the `route` fn could maybe take a tuple? or not.

linkerd/service-profiles/src/http/route_request.rs
```diff
@@ -0,0 +1,130 @@
...
+{
+    type Response = R::Response;
+    type Error = Error;
+    type Future = ErrInto<R::Future, Error>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        let mut update = None;
+        while let Poll::Ready(Some(up)) = self.rx.poll_recv_ref(cx) {
+            update = Some(up.clone());
+        }
```
hawkw, line 95: I believe the semantics of `poll_recv_ref` is that it will only return the latest update, so I'm surprised that this logic is necessary; I don't think it should ever be `Some` multiple times here.

Am I wrong about this? I'm fine with leaving this as-is regardless, but I'm curious whether this is just defensive programming or if we've actually seen issues as a result of this behavior.

hawkw: ...oh, I see that this code was here before, and it just moved around in this branch. Let's not mess with it now, then.

linkerd/service-profiles/src/http/route_request.rs
```diff
@@ -0,0 +1,130 @@
...
+use tracing::{debug, trace};
+
+pub fn layer<M, N: Clone, R>(
+    new_route: N,
+) -> impl layer::Layer<M, Service = NewRouteRequest<M, N, R>> {
+    let default = Route::default();
+    layer::mk(move |inner| NewRouteRequest {
+        inner,
+        new_route: new_route.clone(),
+        default: default.clone(),
```
hawkw, line 20: nit: i was about to ask why we clone this around instead of just calling `Route::default`, but then i saw this comment in the removed code:
https://github.com/linkerd/linkerd2-proxy/blob/bf0c222aef3a388d502f6cba5895e8efc48d92df/linkerd/service-profiles/src/http/service.rs#L32-L33

might be worth bringing that back, or something similar?

linkerd/service-profiles/src/split.rs
```diff
@@ -0,0 +1,195 @@
...
+    task::{Context, Poll},
+};
+use tower::ready_cache::ReadyCache;
+use tracing::debug;
+
+pub fn layer<N, S, Req>() -> impl layer::Layer<N, Service = NewSplit<N, S, Req>> + Clone {
+    let rng = SmallRng::from_entropy();
+    layer::mk(move |inner| NewSplit {
+        inner,
+        rng: rng.clone(),
```
hawkw, line 21: should we be doing something like [the second example here](https://github.com/linkerd/linkerd2-proxy/blob/bf0c222aef3a388d502f6cba5895e8efc48d92df/linkerd/service-profiles/src/http/service.rs#L32-L33)? i'll admit i'm not too familiar w/ `rand`'s APIs...

linkerd/service-profiles/src/split.rs
```diff
@@ -0,0 +1,195 @@
...
+{
+    type Response = S::Response;
+    type Error = Error;
+    type Future = Pin<Box<dyn Future<Output = Result<S::Response, Error>> + Send + 'static>>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        let mut update = None;
+        while let Poll::Ready(Some(up)) = self.rx.poll_recv_ref(cx) {
+            update = Some(up.clone());
+        }
```
hawkw, line 95: as above, not sure if we need to do this...

linkerd/service-profiles/src/split.rs
olix0r: @hawkw is that the right link? if so, i'm not sure what you're suggesting...

linkerd/service-profiles/src/split.rs
olix0r: If the ready cache isn't ready, then we want to be notifed when the profile changes, so we have to drive it to not ready.

linkerd/service-profiles/src/split.rs
hawkw: whoops, it is not --- i meant to link to here: https://docs.rs/rand/0.7.3/rand/rngs/struct.SmallRng.html#examples

linkerd/app/inbound/src/lib.rs
```diff
@@ -335,13 +339,9 @@ impl Config {
...
             .push(errors::layer());
 
-        let http_server_observability = svc::layers()
-            .push(TraceContextLayer::new(span_sink.map(|span_sink| {
-                SpanConverter::server(span_sink, trace_labels())
-            })))
-            // // Tracks proxy handletime.
-            // .push(metrics.http_handle_time.layer())
-            ;
+        let http_server_observability = svc::layers().push(TraceContextLayer::new(
```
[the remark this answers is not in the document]
hawkw: :+1:

PR #642, opened 2020-08-29
svc: Update stack diagnostic checks

olix0r:
This change updates and extends the diagnostic stack checks that we use
during development. No functional changes.


zaharidichev (approved):
LGTM

PR #643, opened 2020-08-29
service-profiles: Eliminate the HasDestination trait

olix0r:
The `HasDestination` trait isn't particularly useful, as it's basically
just `AsRef<Addr>`.

This change updates the `GetRoutes` signature to support this; and it
updates the inbound target type to store an `Addr` instead of an
`Option<NameAddr>` (so `Target` will be suitable for this in an upcoming
change).


zaharidichev (approved):
LGTM

PR #644, opened 2020-08-30
outbound: Make discovery error detection generic

olix0r:
The outbound proxy handles discovery rejection error with a fallback
router. The current implementation does not allow for middlewares
to wrap errors in a general fashion.

This change updates the `is_discovery_rejecte` function to examine
aribitrary error types to determine if the underlying error matches the
known types.


zaharidichev (approved):
LGTM

PR #649, opened 2020-09-03
Eliminate the ConnectAddr trait

olix0r:
The connect module can use the more standard `Into<SocketAddr>`. No need
for a special trait.


hawkw (approved):
the trait eliminator has logged on!

PR #650, opened 2020-09-03
profiles: Do not rely on tuples as stack targets

olix0r:
Using tuples for stack targets (particularly, the input targets of the
route_request and split modules) is a bit brittle.  Instead, we can use
`AsRef` on the target type to access the profile receiver.

This change introduces new target types to be used to satisfy these
traits, and generally cleans up stack construction.


hawkw (approved):
This seems much nicer to me. IMO, this is definitely an appropriate use of `AsRef`, and I like that we're now converting the discovery output into a form that's specific to the inbound/outbound profile stacks rather than messing around with tuples. Looks good to me!

PR #651, opened 2020-09-05
proxy-http: Remove unneeded boilerplate

olix0r:
There's a lot of needless boilerplate around the HTTP client, mostly due
to manual future implementations. These have been converted to boxed
futures.

Furthermore, the HasSettings trait has been eliminated, in favor of
AsRef.

### DIFF
--- a/linkerd/app/inbound/src/endpoint.rs
+++ b/linkerd/app/inbound/src/endpoint.rs
@@ -53,9 +53,9 @@ impl Into<SocketAddr> for HttpEndpoint {
     }
 }
 
-impl http::settings::HasSettings for HttpEndpoint {
-    fn http_settings(&self) -> http::Settings {
-        self.settings
+impl AsRef<http::Settings> for HttpEndpoint {
+    fn as_ref(&self) -> &http::Settings {
+        &self.settings
     }
 }
 
@@ -135,9 +135,9 @@ impl http::normalize_uri::ShouldNormalizeUri for Target {
     }
 }
 
-impl http::settings::HasSettings for Target {
-    fn http_settings(&self) -> http::Settings {
-        self.http_settings
+impl AsRef<http::Settings> for Target {
+    fn as_ref(&self) -> &http::Settings {
+        &self.http_settings
     }
 }
 
--- a/linkerd/proxy/http/src/client.rs
+++ b/linkerd/proxy/http/src/client.rs
@@ -207,9 +207,17 @@ where
                         .instrument(span.clone()),
                 )
             }
-            Client::Http2(ref mut h2) => {
-                Box::pin(h2.call(req).err_into::<Error>().instrument(span.clone()))
-            }
+            Client::Http2(ref mut h2) => Box::pin(
+                h2.call(req)
+                    .map_ok(|rsp| {
+                        rsp.map(|b| Body {
+                            body: Some(b),
+                            upgrade: None,
+                        })
+                    })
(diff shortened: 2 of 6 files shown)


linkerd/proxy/http/src/client.rs
```diff
@@ -114,48 +75,59 @@ where
...
     type Response = Client<C, T, B>;
     type Error = Error;
-    type Future = MakeFuture<C, T, B>;
+    type Future =
+        Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send + 'static>>;
 
-    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        self.connect.poll_ready(cx).map_err(Into::into)
+    fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
```
hawkw, line 89: shouldn't we still poll the `Connect` for readiness? IDK if they are ever actually not ready, but this seems like it could be a problem if the impl changed...

linkerd/proxy/http/src/client.rs
```diff
@@ -114,48 +75,59 @@ where
...
-                .set_host(was_absolute_form)
-                .build(HyperConnect::new(connect, target, was_absolute_form));
-                MakeFuture::Http1(Some(h1))
-            }
-            Settings::Http2 => {
-                let h2 = h2::Connect::new(connect, self.h2_settings.clone()).oneshot(target);
-                MakeFuture::Http2(h2)
             }
-        }
+        })
```
hawkw, line 130: it seems like we _could_ do this one pretty easily with `futures::Either` and like, one combinator, since there's only a single `await` in the h2 arm, and no awaits in the h1 arm...may not be worth it, the async fn version is still slightly nicer.

linkerd/app/inbound/src/endpoint.rs
```diff
@@ -53,9 +53,9 @@ impl Into<SocketAddr> for HttpEndpoint {
     }
 }
 
-impl http::settings::HasSettings for HttpEndpoint {
-    fn http_settings(&self) -> http::Settings {
-        self.settings
+impl AsRef<http::Settings> for HttpEndpoint {
```
hawkw, line 56: definitely +1 on this

hawkw (commented):
The new code definitely looks nicer!

As a follow-up, it might be nice to switch to nightly rust and change the boxed futures to `impl Future`, at least to run the benchmarks. I'm not too concerned about the performance implications but it would be nice to do a comparison.

I noticed that we're no longer polling `Connect` readiness. Is that potentially an issue?

linkerd/proxy/http/src/client.rs
olix0r: We clone the connect and oneshot, so any polling we do cannot be used, anyway.

linkerd/proxy/http/src/client.rs
olix0r: Assume, for now, that this is going to change and become more complicated

linkerd/proxy/http/src/client.rs
hawkw: ah, yup, i guess polling it here was never necessary. should maybe have a comment? but, not a big deal.

hawkw (approved):

PR #652, opened 2020-09-07
outbound: TCP discovery and load balancing

olix0r:
The proxy only forward TCP connections to the original destination address, never
using the mesh's mTLS.

This change modifies the outbound TCP forwarding stack to do discovery based on
the original destination address, resolving service IPs to the individual endpoint IPs
(and their associated identities). This enables mTLS for meshed TCP connections.
When the endpoint cannot be discovered (i.e. due to an InvalidArgument response
from the controller), the connection is forwarded as before.

The PeakEWMA balancer is repurposed for this, using connection latency (and
pending connections) as the load metric. In the future, this should be modified
to count active connections towards the load (but this won't Just Work with the
PeakEwma balancer, so this is deferred for now).

This has been tested [manually](https://github.com/olix0r/init-net-test/blob/b3860861c54344667a797ece5b685a0574798fb3/k8s.yml).
In follow-up changes, we'll extend the transparency tests to validate this behavior.


linkerd/proxy/http/src/version.rs
```diff
@@ -34,11 +36,15 @@ impl Version {
...
             Ok(_) | Err(httparse::Error::TooManyHeaders) => {
+                trace!("Detected H1");
                 return Some(Version::Http1);
             }
             _ => {}
         }
 
+        debug!("Not HTTP");
+        trace!(?bytes);
+        trace!(h2.preface = ?Self::H2_PREFACE);
```
hawkw, line 47: nit, take it or leave it: any reason this isn't
```suggestion
        trace!(?bytes, h2.preface = ?Self::H2_PREFACE);
```

linkerd/stack/src/make_thunk.rs
```diff
@@ -0,0 +1,53 @@
+use futures::future;
+use linkerd2_error::Never;
+use std::task::{Context, Poll};
+
+#[derive(Clone, Debug)]
+pub struct MakeThunk<M> {
+    make: M,
+}
```
hawkw, line 8: it might be nice to have a comment explaining what this thing does exactly, for future readers?

i agree that `MakeThunk` describes what it does in a general sense, but it might be worth naming it something that uses `tower`/linkerd-specific language? e.g., when i wrote a version of this for the control plane connector (which we no longer use), i initially named it `Curried`, and then thought that seemed too abstract, and called it `WithTarget`.

sorry for bikeshedding, it fundamentally doesn't matter.

linkerd/app/outbound/src/lib.rs
```diff
@@ -448,22 +458,54 @@ impl Config {
...
+            .push(discover::buffer(1_000, cache_max_idle_age))
+            .push_map_target(Addr::from)
+            .push_on_response(tcp::balance::layer(EWMA_DEFAULT_RTT, EWMA_DECAY))
+            .push_fallback_with_predicate(
+                svc::stack(tcp_connect.clone())
+                    .push_make_thunk()
+                    .push(admit::AdmitLayer::new(prevent_loop))
+                    .push_map_target(TcpEndpoint::from),
+                is_discovery_rejected,
+            )
```
hawkw, line 479: nit/tioli: might be nice to add `trace` or `debug` spans in here indicating whether we are in the tcp balancer or forwarding to the original dst?

hawkw (commented):
pleasantly surprised by how little is going on here! this looks good so far.

olix0r:
Tested via https://github.com/olix0r/init-net-test/blob/b3860861c54344667a797ece5b685a0574798fb3/k8s.yml --  appears to work. We'll want to do some deeper burnin testing, and we'll definitely need to improve metrics, etc; but I don't think any of that needs to block this moving forward.

linkerd/proxy/http/src/version.rs
olix0r: it's easier to compare then when they're not on one long line.

hawkw: ah, fair enough.

linkerd/stack/src/make_thunk.rs
olix0r: I documented it. I don't think `WithTarget` is all that clear, personally (especially when compared with `MapTarget` -- I'd expect `WithTarget` to take a constant target to be reused for all requests, or something...). Anyway, we can easily change this later.

linkerd/app/core/src/svc.rs
```diff
@@ -159,6 +159,14 @@ impl<S> Stack<S> {
         self.push(stack::map_target::MapTargetLayer::new(map_target))
     }
 
+    /// Wraps a `Service<T>` as a `Service<()>`.
+    ///
+    /// Each time the service is called, the `T`-typed request is cloned and
+    /// issued into the inner service.
```
hawkw, line 165: :+1:

linkerd/app/outbound/src/endpoint.rs
```diff
@@ -323,14 +331,35 @@ impl tls::HasPeerIdentity for TcpEndpoint {
...
+//     fn into(self) -> EndpointLabels {
+//         use linkerd2_app_core::metric_labels::{Direction, TlsId};
+//         EndpointLabels {
+//             authority: None,
+//             direction: Direction::Out,
+//             labels: self.labels,
+//             tls_id: self.identity.as_ref().map(|id| TlsId::ServerId(id.clone())),
+//         }
+//     }
+// }
```
hawkw, line 344: should the commented-out code just be removed?

linkerd/app/outbound/src/endpoint.rs
```diff
@@ -302,15 +303,22 @@ impl Into<EndpointLabels> for HttpEndpoint {
...
 
-impl From<listen::Addrs> for TcpEndpoint {
-    fn from(addrs: listen::Addrs) -> Self {
+impl From<SocketAddr> for TcpEndpoint {
+    fn from(addr: SocketAddr) -> Self {
         Self {
-            addr: addrs.target_addr(),
+            addr,
             identity: Conditional::None(tls::ReasonForNoPeerName::NotHttp.into()),
+            //labels: None,
```
hawkw, line 311:
```suggestion
            identity: Conditional::None(tls::ReasonForNoPeerName::NotHttp.into()),
```

linkerd/app/outbound/src/endpoint.rs
```diff
@@ -323,14 +331,35 @@ impl tls::HasPeerIdentity for TcpEndpoint {
...
+            .cloned()
+            .map(Conditional::Some)
+            .unwrap_or_else(|| {
+                Conditional::None(tls::ReasonForNoPeerName::NotProvidedByServiceDiscovery.into())
+            });
+
+        TcpEndpoint {
+            addr,
+            identity,
+            //labels: prefix_labels("dst", metadata.labels().into_iter()),
```
hawkw, line 362:
```suggestion
```

linkerd/proxy/http/src/version.rs
```diff
@@ -34,11 +36,15 @@ impl Version {
...
             Ok(_) | Err(httparse::Error::TooManyHeaders) => {
+                trace!("Detected H1");
                 return Some(Version::Http1);
             }
             _ => {}
         }
 
+        debug!("Not HTTP");
+        trace!(?bytes);
+        //trace!(h2.preface = ?Self::H2_PREFACE);
```
hawkw, line 47:
```suggestion
```

linkerd/app/outbound/src/endpoint.rs
```diff
@@ -58,6 +58,7 @@ pub struct HttpEndpoint {
 pub struct TcpEndpoint {
     pub addr: SocketAddr,
     pub identity: tls::PeerIdentity,
+    //pub labels: Option<String>,
```
hawkw, line 61:
```suggestion
```

hawkw (commented):
overall, this looks good to me, though i had a few nits.

linkerd/app/outbound/src/endpoint.rs
olix0r: There's a followup PR that enables it.

olix0r: https://github.com/linkerd/linkerd2-proxy/pull/654

hawkw: ah okay

hawkw: right, duh. sorry. my brain doesn't work.

hawkw (approved):
lgtm

PR #653, opened 2020-09-08
outbound: Clarify Http target types

olix0r:
The outbound `endpoint::Target` type is generic to the point of being
useless.

This change modifies the stack target types to be either `HttpLogical`,
`HttpConcrete`, or `HttpEndpoint`. Furthermore, the `canonicalize`
middleware no longer uses a bespoke trait, instead opting for
`AsRef` + `AsMut`.

This sets up further changes to the outbound HTTP stack.

### DIFF
--- a/linkerd/app/outbound/src/lib.rs
+++ b/linkerd/app/outbound/src/lib.rs
@@ -339,7 +339,7 @@ impl Config {
                 ),
             )
             .spawn_buffer(buffer_capacity)
-            .instrument(|t: &HttpEndpoint| info_span!("forward", peer.addr = %t.addr))
+            .instrument(|t: &HttpEndpoint| info_span!("forward", peer.addr = %t.addr, peer.id = ?t.identity))
             .check_make_service::<HttpEndpoint, http::Request<_>>();
 
         // Attempts to route route request to a logical services that uses
--- a/linkerd/app/outbound/src/lib.rs
+++ b/linkerd/app/outbound/src/lib.rs
@@ -5,7 +5,7 @@
 
 #![deny(warnings, rust_2018_idioms)]
 
-pub use self::endpoint::{Concrete, HttpEndpoint, Logical, LogicalPerRequest, Target, TcpEndpoint};
+pub use self::endpoint::{HttpConcrete, HttpEndpoint, HttpLogical, LogicalPerRequest, TcpEndpoint};
 use ::http::header::HOST;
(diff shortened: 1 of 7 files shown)


linkerd/app/outbound/src/lib.rs
```diff
@@ -325,31 +323,24 @@ impl Config {
...
             .spawn_buffer(buffer_capacity)
-            .instrument(|endpoint: &Target<HttpEndpoint>| {
-                info_span!("forward", peer.addr = %endpoint.addr, peer.id = ?endpoint.inner.identity)
-            })
-            .push_map_target(|t: Concrete<HttpEndpoint>| Target {
-                addr: t.addr.into(),
-                inner: t.inner.inner,
-            })
-            .check_make_service::<Concrete<HttpEndpoint>, http::Request<_>>();
+            .instrument(|t: &HttpEndpoint| info_span!("forward", peer.addr = %t.addr))
```
hawkw, line 342: looks like this span used to contain an identity, do we still care about this?

linkerd/proxy/http/src/canonicalize.rs
```diff
@@ -146,8 +141,8 @@ where
                     let target = match ready!(future.poll(cx)) {
                         Ok(Ok(refined)) => {
                             let mut target = original.take().expect("illegal state");
-                            let name = NameAddr::new(refined, target.addr().port());
-                            *target.addr_mut() = name.into();
+                            let name = NameAddr::new(refined, target.as_ref().port());
+                            *target.as_mut() = name.into();
```
hawkw, line 145: I guess the one downside of replacing the traits with `AsRef`/`AsMut` is that it's a bit less clear at the callsite what type this is being `AsRef`ed to. 

One thing we _could_ do is keep the trait, but change it to
```rust
pub trait Target {
    fn addr(&self) -> &Addr;
    fn addr_mut(&mut self) -> &mut Addr;
}

impl<T: AsRef<Addr> + AsMut<Addr>> Target for T {
    fn addr(&self) -> &Addr {
        self.as_ref()
    }
    
     fn addr_mut(&mut self) -> &mut Addr {
         self.as_mut()
     }
}
```
and then we could still use methods with semantic names at the callsite, but not require other crates to implement a special trait. 

On the other hand, I'm not sure if this is worth the effort.
```

hawkw (approved):

PR #654, opened 2020-09-09
metrics: Add endpoint labels to outbound TCP metrics

olix0r:
TCP metrics are pretty terse. Now that we're doing discovery for TCP forwards,
it seems appropriate to expose endpoint labels on TCP metrics (at least to
determine if/when mTLS is being applied).


kleimkuhler (approved):
Looks good!

PR #655, opened 2020-09-11
introduce tests for isolated services

hawkw:
This branch introduces a new style of tests to the proxy codebase, and
an initial implementation of new test support code for writing tests in
this style. Rather than running an entire proxy alongside a simulated
control plane, and binding actual network socket, as the present
integration tests do, the new tests run individual proxy components in
isolation, using simpler mock implementations of components like name
resolution and IO. 

This approach has a few advantages. It should reduce flakiness
significantly, since we don't perform any IO and don't need to
synchronize events between multiple threads running test support
servers/controllers (the new mocks all just synchronously provide values
immediately). We can run everything in a single thread using Tokio's
basic scheduler. Additionally, since we are testing individual
components in isolation, these tests can live within the crate for the
part of the proxy being tested. This means we have access to more
internal state to make assertions on, rather than having to make
assertions on side effects like metrics.

I've added a new `linkerd2-app-test` crate that adds some initial mocks
for this kind of testing, including a simple mock resolver and a mock
connector using `tokio-test`'s mock IO. I've also added a simple "hello
world" test for the outbound TCP stack written in this style. The
intention is to use the new test support code for testing the recent
changes adding TCP mTLS and load-balancing.

In order to make this code more testable, it was necessary to move a few
things around. In particular, the outbound TCP balancer stack is now
constructed in a separate method, rather than in `Config::build_server`,
so that it can be tested without requiring all the dependencies of the
full HTTP server stack. A few other similar changes were also necessary.


linkerd/app/outbound/src/lib.rs
```diff
@@ -367,6 +369,83 @@ impl Config {
...
+                        .push_failfast(dispatch_timeout)
+                        .push_spawn_buffer_with_idle_timeout(buffer_capacity, cache_max_idle_age)
+                        .push(metrics.stack.layer(stack_labels("tcp"))),
+                ),
+            )
+            .spawn_buffer(buffer_capacity)
+            .check_make_service::<SocketAddr, ()>()
+            .push(svc::layer::mk(tcp::Forward::new))
+            .instrument(|a: &SocketAddr| info_span!("tcp", dst = %a))
+            .push_map_target(|a: listen::Addrs| a.target_addr())
```
olix0r, line 446: i'd probably have this stack take the SocketAddr and do the mapping in the server side -- listen::Addrs is pure-server detail.

hawkw: mmk, that makes sense. this would also simplify a few things in the tests, i think

linkerd/app/integration/src/lib.rs
```diff
@@ -14,6 +14,8 @@ pub use http::{HeaderMap, Request, Response, StatusCode};
 pub use http_body::Body as HttpBody;
 pub use linkerd2_app as app;
 pub use linkerd2_app_core::drain;
+pub use linkerd2_app_core::Addr;
```
olix0r, line 17: nit: combine core imports?

linkerd/app/outbound/src/tests.rs
```diff
@@ -0,0 +1,97 @@
...
+            .into(),
+            dispatch_timeout: Duration::from_secs(3),
+            max_in_flight_requests: 10_000,
+            detect_protocol_timeout: Duration::from_secs(3),
+        },
+    }
+}
+
+#[tokio::test(core_threads = 1)]
+async fn hello_world() {
```
olix0r, line 57: Can we give this a more descriptive name? What is it actually testing?

olix0r (changes requested):
Tests look great. Some superficial feedback

linkerd/app/test/src/lib.rs
```diff
@@ -0,0 +1,62 @@
+//! Shared infrastructure for integration tests
+
+#![deny(warnings, rust_2018_idioms)]
+#![recursion_limit = "256"]
+#![type_length_limit = "16289823"]
+
+pub use futures::{future, FutureExt, TryFuture, TryFutureExt};
+
+pub use linkerd2_app_core::{self as app_core, Addr, Error};
```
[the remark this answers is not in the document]
hawkw: IIRC it clashed with libcore. will double check if this is the case.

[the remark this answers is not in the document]
hawkw: yup:
```
error[E0659]: `core` is ambiguous (name vs any other name during import resolution)
  --> linkerd/app/outbound/src/tests.rs:12:9
   |
12 |     use core::{
   |         ^^^^ ambiguous name
   |
   = note: `core` could refer to a built-in crate
   = help: use `::core` to refer to this crate unambiguously
note: `core` could also refer to the crate imported here
  --> linkerd/app/outbound/src/tests.rs:3:25
   |
3  | use linkerd2_app_core::{self as core, metrics::Metrics, Addr};
   |                         ^^^^^^^^^^^^
   = help: use `self::core` to refer to this crate unambiguously

```

linkerd/app/outbound/src/lib.rs
```diff
@@ -406,18 +484,23 @@ impl Config {
...
         S::Future: Send,
     {
+        let prevent_loop = PreventLoop::from(listen_addr.port());
+
+        // Load balances TCP streams that cannot be decoded as HTTP.
+        let tcp_balance =
+            svc::stack(self.build_tcp_balance(&tcp_connect, resolve, prevent_loop, &metrics))
+                .push_map_target(|a: listen::Addrs| a.target_addr())
+                .into_inner();
+
```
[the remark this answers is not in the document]
hawkw: it's above the `ProxyConfig` binding because the `ProxyConfig` binding moves out of `self`, and `build_tcp_balance` borrows `self`. we could probably change the `ProxyConfig` binding to be by-ref though.

olix0r (approved):

linkerd/app/test/src/resolver.rs
```diff
@@ -0,0 +1,153 @@
...
+use linkerd2_app_core::Error;
+pub use linkerd2_proxy_api_resolve::{Metadata, ProtocolHint};
+pub use linkerd2_proxy_core::resolve::{Resolve, Update};
+use std::collections::HashMap;
+use std::hash::Hash;
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc, Mutex,
+};
+// use std::future::Future;
```
kleimkuhler, line 11: Can remove this

linkerd/app/test/src/resolver.rs
```diff
@@ -0,0 +1,153 @@
...
+use std::net::SocketAddr;
+use std::task::{Context, Poll};
+use tokio::sync::mpsc;
+// use tower::Service;
```
kleimkuhler, line 15: And this

kleimkuhler (approved):
Looks great!

PR #656, opened 2020-09-15
http: Simplify stacks and target types

olix0r:
Currently, various properties of HTTP requests are exposed into the
stack target type, meaning that stacks (including their resolutions)
cannot be shared across requests that vary by any of these settings and,
worse, that per-request determinations must be made to determine which
stack to use.

In order to bind stacks according to connection-level metadata,
where these request-level parameters cannot be known, we need to remove
this metadata from the stack target.

The HTTP server now applies URI normalization as an internal
implementation detail (because it's a detail of how hyper clients and
servers differ). This eliminates the `ShouldNormalizeUri` trait, as that
determination can be made entirely based on the properties of the
request without any additional stack-specific metadata.

The HTTP client now takes an updated `Settings` type that explicitly
handles the orig-proto upgrade logic (removing this from the client
stack).

Stack targets now hold an `http::Version` on the accept-side, and map
that to a per-endpoint `http::client::Settings` for use by the http
client.


linkerd/proxy/http/src/orig_proto.rs
```diff
@@ -26,93 +27,85 @@ pub struct Downgrade<S> {
...
+                .headers_mut()
+                .remove(L5D_ORIG_PROTO)
+                .and_then(|orig_proto| {
+                    if orig_proto == "HTTP/1.1" {
+                        Some(http::Version::HTTP_11)
+                    } else if orig_proto == "HTTP/1.0" {
+                        Some(http::Version::HTTP_10)
+                    } else {
+                        None
+                    }
```
hawkw, line 102: nit/tioli: could be 
```suggestion
                    match orig_proto {
                        "HTTP/1.1" =>  Some(http::Version::HTTP_11),
                        "HTTP/1.0" => Some(http::Version::HTTP_10),
                        _ => None,
                    }
```
...unless i'm missing something, of course?

linkerd/proxy/transport/src/listen.rs
```diff
@@ -142,6 +142,12 @@ impl Addrs {
     }
 }
 
+impl<'t> Into<SocketAddr> for &'t Addrs {
```
hawkw, line 145: nit/TIOLI: could be
```suggestion
impl Into<SocketAddr> for &'_ Addrs {
```

linkerd/proxy/transport/src/tls/accept.rs
```diff
@@ -286,3 +286,9 @@ impl std::fmt::Display for DetectTimeout {
 }
 
 impl std::error::Error for DetectTimeout {}
+
+impl<'t> Into<std::net::SocketAddr> for &'t Meta {
```
hawkw, line 290: nit/tioli: could be
```suggestion
impl Into<std::net::SocketAddr> for &'_ Meta {
```

linkerd/proxy/http/src/h1.rs
```diff
@@ -1,45 +1,142 @@
...
+            };
+
+            if client.is_none() {
+                debug!("Caching new client");
+                *client = Some(hyper::Client::builder().set_host(absolute_form).build(
+                    HyperConnect::new(self.connect.clone(), self.target.clone(), absolute_form),
+                ));
+            }
+
+            client.as_ref().unwrap().request(req)
```
hawkw, line 109: nit/TIOLI: there's actually an [`Option::get_or_insert_with`] method that basically implements this logic and avoids the unwrap:
```suggestion
            client
                .get_or_insert_with(|| {
                    debug!("Caching new client");
                    hyper::Client::builder().set_host(absolute_form).build(
                        HyperConnect::new(self.connect.clone(), self.target.clone(), absolute_form),
                    )
                })
                .request(req)
```

hawkw (approved):
the new factoring makes a lot of sense to me --- this feels simpler and easier to understand. i had some style nits and suggestions, but didn't notice any problems.

linkerd/proxy/http/src/h1.rs
olix0r: Unfortunately, this doesn't play with the borrow checker and `self` -- because we're holding a mutable ref to the client, we can't pass a reference to self into the closure.

linkerd/proxy/http/src/orig_proto.rs
olix0r: `orig_proto` is a HeaderValue and not a string, so the matching form doesn't work

linkerd/proxy/http/src/normalize_uri.rs
```diff
@@ -104,27 +105,23 @@ where
...
+                if h1::is_absolute_form(req.uri()) {
+                    trace!(uri = ?req.uri(), "Absolute");
+                    req.extensions_mut().insert(h1::WasAbsoluteForm(()));
+                } else if req.uri().authority().is_none() {
+                    let authority =
+                        h1::authority_from_host(&req).unwrap_or_else(|| self.default.clone());
+                    trace!(%authority, "Normalizing URI");
+                    h1::set_authority(req.uri_mut(), authority);
+                }
+            }
```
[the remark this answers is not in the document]
olix0r: used `if let` with http/1 types (didn't realize that syntax worked there.

linkerd/app/gateway/src/lib.rs
```diff
@@ -153,7 +153,7 @@ mod test {
                 dst: dst_name
                     .map(|n| NameAddr::from_str(n).unwrap().into())
                     .unwrap_or_else(|| socket_addr.into()),
-                http_settings: http::Settings::Http2,
+                http_version: http::Version::Http1,
```
kleimkuhler, line 156: Is this now `Http1` because we know it should be upgraded to `H2`? If not, why this change?

kleimkuhler (approved):
This looks good! I have a question about a change in Gateway, but otherwise this makes sense to me.

linkerd/app/gateway/src/lib.rs
olix0r: it's a test, so it's not actually used

olix0r:
So, this branch seems to impact CI time quite a bit.

Locally, I see:

#### `main`

```
#15 240.5     Finished release [optimized] target(s) in 4m 00s
#15 240.6       User time (seconds): 2043.05
#15 240.6       System time (seconds): 31.72
#15 240.6       Percent of CPU this job got: 863%
#15 240.6       Elapsed (wall clock) time (h:mm:ss or m:ss): 4:00.35
#15 240.6       Maximum resident set size (kbytes): 4573996
```

#### `ver/orig-proto`

```
#15 389.1     Finished release [optimized] target(s) in 6m 28s
#15 389.1       User time (seconds): 2143.76
#15 389.1       System time (seconds): 31.86
#15 389.1       Percent of CPU this job got: 559%
#15 389.1       Elapsed (wall clock) time (h:mm:ss or m:ss): 6:28.91
#15 389.1       Maximum resident set size (kbytes): 4328072
```

I'm not sure how this could have made it 30-40% worse... i would have expected the opposite.

PR #657, opened 2020-09-16
Make SkipDetect more generic as stack::MakeSwitch

olix0r:
SkipDetect is needlessly coupled to its `listen::Addrs` target type and
the Accept stack's shape. In reality, it's pretty similar to the
fallback module, except that decisions are made on the target rather
than the response future.

This change replaces the `core::proxy::skip_detct` module with
`stack::switch`.

Above all, this allows us to use other targets with this module, which
is needed to do discovery before protoocl detection (and to use the
results of discovery to inform detection).


linkerd/stack/src/switch.rs
```diff
@@ -0,0 +1,84 @@
...
+use tower::util::ServiceExt;
+
+/// Determines whether the primary stack should be used.
+pub trait Switch<T> {
+    fn use_primary(&self, target: &T) -> bool;
+}
+
+/// Makes either the primary or fallback stack, as determined by an `S`-typed
+/// `Switch`.
+pub struct MakeSwitch<S, P, F> {
```
hawkw, line 16: nit: not sold on calling this `MakeSwitch`, since it would work equally well at the `Service`/`Request` level as well as the `MakeService`/`Target` level...names starting with `Make` kinda suggest to me that this only ever has a `Service` response type. Maybe the service should be `Switch` and the trait `Predicate or something?

hawkw (approved):
this looks good to me, and the change is really quite simple...i had some dumb naming bikesheds, feel free to disregard them

olix0r:
Can address naming in a followup, but want to get this merged

PR #658, opened 2020-09-16
outbound: initial tests for TCP mTLS

hawkw:
This branch uses the new test-support code from #655 to add some initial
tests for TCP mTLS on the outbound router. The new tests use the mock IO
types and service discovery resolver added in #655, and test the
outbound TCP stack in isolation. Therefore, we don't run the rest of the
proxy, open any sockets, or spawn multiple threads.

In order to write these tests it was necessary to do some additional
internal refactoring. In particular, we were previously passing a mock
connect service into the outbound TCP stack. However, the connect stack
is responsible for handling TLS handshakes, so in order to run tests
with TLS, it was necessary to add a new function for building the TCP
connect stack with an arbitrary base service passed in, in order to
build the TCP connect stack around a mock connector.

Additionally, I've added a vendored version of the
`tokio::io::DuplexStream` type, which implements `AsyncRead` and
`AsyncWrite` for an in-memory channel. This exists upstream, but has yet
to be released, and the upstream implementation only exists for Tokio
0.3, so I just copied and pasted it. This was necessary to perform TLS
handshakes on a mock IO. The pre-configured `tokio-test::io` mock IOs
don't work in this case, since the `tokio-rustls` API is proactive
rather than reactive: it provides a future that performs a handshake on
an IO type, rather than an IO type that performs a handshake when
written to.

Finally, it was necessary to add a test feature to the `linkerd2-io`
crate that enables using the test IO `Mock` and `DuplexStream` types as
`BoxedIo`s.

This branch adds tests for the following: 
- the proxy uses mTLS for client-first and server-first TCP connections
  when it receives a TLS identity from service discovery 
- the proxy uses mTLS for client-first and server-first TCP connections
  *only* if it receives a TLS identity from service discovery --- 
  plaintext ports on the same IP are still plaintext 
- if a connection is accepted when an address does not have a TLS hint,
  and then one is received before the upstream connection is opened, it
  will use TLS


linkerd/app/outbound/src/tests.rs
```diff
@@ -89,9 +95,320 @@ async fn plaintext_tcp() {
 
     // Build the outbound TCP balancer stack.
     let outbound_tcp = cfg.build_tcp_balance(&connect, resolver, prevent_loop, &metrics.outbound);
+    // Oneshot the TCP stack with the target and request (client conn).
+    outbound_tcp.make_oneshot(target_addr, client_io).await
```
kleimkuhler, line 99: Does this test need to assert that the client/server read/write what they expect?

kleimkuhler (approved):
This looks great! I have two nonblocking comments, but otherwise the tests seem good to me.

linkerd/app/test/src/io.rs
```diff
@@ -0,0 +1,10 @@
+pub use linkerd2_io::{
+    duplex::{duplex, DuplexStream},
+    BoxedIo,
+};
+pub use std::io::Error;
+pub use tokio_test::io::*;
```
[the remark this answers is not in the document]
hawkw: the `io::Mock` type is also imported through this reexport elsewhere.

linkerd/app/outbound/src/tests.rs
hawkw: the mock IO types do this for us automatically --- they'll panic if they see unexpected `read`/`write` calls, or if they're dropped without completing their "script"s

linkerd/app/outbound/src/lib.rs
```diff
@@ -68,9 +68,47 @@ impl Config {
...
            + Clone
            + Send {
+        let connect = svc::connect(self.proxy.connect.keepalive).into_inner();
         // Establishes connections to remote peers (for both TCP
         // forwarding and HTTP proxying).
-        svc::connect(self.proxy.connect.keepalive)
+        self.build_tcp_connect_with(connect, local_identity, metrics)
+    }
+
+    pub fn build_tcp_connect_with<C>(
```
olix0r, line 77: Is this really necessaary? Can we provide a service and check its target?

In other words, why do we actually care about validating the socket/connection behavior? It comes at the cost of... a lot of verbosity.

olix0r, line 77: If we do need, let's call it build_tcp_endpoint or something... and i think we shuold have a build_tcp_connect that only produces the connect layer...

linkerd/app/outbound/src/lib.rs
```diff
@@ -68,9 +68,47 @@ impl Config {
...
+        C: tower::Service<HttpEndpoint, Error = std::io::Error> + Clone + Send + Unpin,
+        <C as tower::Service<HttpEndpoint>>::Future: Send + 'static,
+        <C as tower::Service<HttpEndpoint>>::Response:
+            tokio::io::AsyncRead + tokio::io::AsyncWrite + Unpin + Send + 'static,
+        transport::BoxedIo: From<<C as tower::Service<HttpEndpoint>>::Response>
+            + From<tokio_rustls::client::TlsStream<<C as tower::Service<HttpEndpoint>>::Response>>,
+        C: tower::Service<TcpEndpoint, Error = std::io::Error> + Clone + Send + Unpin,
+        <C as tower::Service<TcpEndpoint>>::Future: Send + 'static,
+        <C as tower::Service<TcpEndpoint>>::Response:
+            tokio::io::AsyncRead + tokio::io::AsyncWrite + Unpin + Send + 'static,
```
olix0r, line 105: If we do need to keep this, let's please use another intermediary type attr inttead of having to do all of the `as tower::Service` noise

linkerd/io/src/boxed.rs
```diff
@@ -86,6 +87,38 @@ impl Io for BoxedIo {
...
+    }
+}
+
+impl<S: Io + Unpin + 'static> From<tokio_rustls::server::TlsStream<S>> for BoxedIo {
+    fn from(stream: tokio_rustls::server::TlsStream<S>) -> Self {
+        Self::new(stream)
+    }
+}
+
+impl<S: Io + Unpin + 'static> From<tokio_rustls::client::TlsStream<S>> for BoxedIo {
```
olix0r, line 102: Is it possible to implement this as just `From<I>`, since the TLS stream already implements Io? Or does the compiler get grumpy?

hawkw: `BoxedIo` itself implements `Io`, so it conflicts with the existing `From<Self>` impl that exists for everything. that was the first thing I tried.

PR #659, opened 2020-09-18
http: Put normalize_uri back on the stack

olix0r:
In b59172a, we moved the NormalizeUri logic into the HTTP server, but
this misses downgraded requests.

This change extracts NormalizeUri from the server, putting it back on
the HTTP stack so that it can be applied after `orig_proto::Downgrade`
is applied on the inbound proxy.


zaharidichev (approved):
LGTM

kleimkuhler (approved):
👍

hawkw (approved):

PR #660, opened 2020-09-18
inbound: Apply loop detection on the connect stack

olix0r:
a233e1af altered the accept stack so that both TCP and HTTP stacks are
instantiated prior to performing detection on a socket. This will allow
us to cache these stacks for each destination.

However, this eager binding broke the gateway fallback logic, since the
TCP stack's failure invalidated the HTTP stacks' fallback to the
gateway.

This change moves loop detection onto the TCP connection stack so this
eager binding does not fail.

Fixes linkerd/linkerd2#4943


kleimkuhler (approved):
Looks good!

PR #661, opened 2020-09-18
tracing: Elide redundant info in tracing contexts

olix0r:
The tracing context includes redundnant information, particularly about
the traffic target.

This change modifies the accept stack to include both the source peer and
target addresses in the `accept` context; and the target address has
been removed from intermediate contexts, many of which can be moved to
the debug level now.

### DIFF
--- a/linkerd/app/core/src/serve.rs
+++ b/linkerd/app/core/src/serve.rs
@@ -36,6 +36,7 @@ where
                     let span = info_span!(
                         "accept",
                         peer.addr = %addrs.peer(),
+                        target.adr = %addrs.target_addr(),
                     );
 
                     // Ready the service before dispatching the request to it.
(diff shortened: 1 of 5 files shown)


linkerd/app/inbound/src/lib.rs
```diff
@@ -354,13 +354,8 @@ impl Config {
                     .box_http_request()
                     .box_http_response(),
             )
+            .instrument(|_: &_| debug_span!("source",))
```
hawkw, line 357: nit:
```suggestion
            .instrument(|_: &_| debug_span!("source"))
```

kleimkuhler (approved):
👍

linkerd/app/core/src/serve.rs
```diff
@@ -36,6 +36,7 @@ where
                     let span = info_span!(
                         "accept",
                         peer.addr = %addrs.peer(),
+                        target.adr = %addrs.target_addr(),
```
hawkw, line 39: typo:
```suggestion
                        target.addr = %addrs.target_addr(),
```

hawkw (commented):
noticed a couple typos. LGTM otherwise!

hawkw (approved):

PR #662, opened 2020-09-18
outbound: Reorganize outbound stacks

olix0r:
There's no functional change here, just some general cleanup
after recent changes, including:

* The TCP balancer is no longer responsible for fallback
  forwarding. This makes the server responsible for all TCP
  forwarding. The metrics have been removed from the TCP
  balancer stack (for now).
* Loop detection is moved to the connect stack, rather than
  the accept stack.
* The TCP balancer tests have been simplified somewhat.
* `tcp::Forward` no longer has a maker. `OnReponse` does
  the job just fine.

### DIFF
--- a/linkerd/app/outbound/src/lib.rs
+++ b/linkerd/app/outbound/src/lib.rs
@@ -6,7 +6,6 @@
 #![deny(warnings, rust_2018_idioms)]
 
 pub use self::endpoint::{HttpConcrete, HttpEndpoint, HttpLogical, LogicalPerRequest, TcpEndpoint};
-use ::http::header::HOST;
 use futures::{future, prelude::*};
 use linkerd2_app_core::{
     admit, classify,
@@ -21,8 +20,8 @@ use linkerd2_app_core::{
     spans::SpanConverter,
     svc::{self, NewService},
     transport::{self, listen, tls},
-    Addr, Conditional, DiscoveryRejected, Error, ProxyMetrics, StackMetrics, TraceContextLayer,
-    CANONICAL_DST_HEADER, DST_OVERRIDE_HEADER, L5D_REQUIRE_ID,
+    Addr, Conditional, DiscoveryRejected, Error, Never, ProxyMetrics, StackMetrics,
+    TraceContextLayer, CANONICAL_DST_HEADER, DST_OVERRIDE_HEADER, L5D_REQUIRE_ID,
 };
 use std::{
     collections::HashMap,
@@ -53,6 +52,7 @@ pub struct Config {
 impl Config {
     pub fn build_tcp_connect(
         &self,
+        prevent_loop: impl Into<PreventLoop>,
         local_identity: tls::Conditional<identity::Local>,
         metrics: &ProxyMetrics,
     ) -> impl tower::Service<
@@ -76,6 +76,7 @@ impl Config {
             // Limits the time we wait for a connection to be established.
             .push_timeout(self.proxy.connect.timeout)
             .push(metrics.transport.layer_connect(TransportLabels))
+            .push(admit::AdmitLayer::new(prevent_loop.into()))
             .into_inner()
     }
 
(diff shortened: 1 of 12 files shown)


hawkw:
this will probably conflict with #658, so whoever merges second is gonna have to figure that out

linkerd/app/outbound/src/lib.rs
```diff
@@ -76,9 +76,75 @@ impl Config {
             // Limits the time we wait for a connection to be established.
             .push_timeout(self.proxy.connect.timeout)
             .push(metrics.transport.layer_connect(TransportLabels))
+            .push(admit::AdmitLayer::new(prevent_loop.into()))
             .into_inner()
     }
 
+    /// Constructs a TCP load balancer.
+    pub fn build_tcp_balance<C, E, I>(
```
hawkw, line 84: is there any actual change here, or did this just move around?

linkerd/app/outbound/src/lib.rs
```diff
@@ -605,3 +598,7 @@ fn is_discovery_rejected(err: &Error) -> bool {
     tracing::debug!(rejected, %err);
     rejected
 }
+
+fn is_loop(err: &(dyn std::error::Error + 'static)) -> bool {
+    err.is::<prevent_loop::LoopPrevented>() || err.source().map(is_loop).unwrap_or(false)
+}
```
hawkw, line 604: nit/tioli: `is_discovery_rejected` has a `debug!` event in it, should this as well?

linkerd/app/outbound/src/tests.rs
```diff
@@ -1,51 +1,35 @@
...
         canonicalize_timeout: Duration::from_millis(100),
-        proxy: ProxyConfig {
-            server: ServerConfig {
-                bind: listen::Bind::new(SocketAddr::new(LOCALHOST.into(), LISTEN_PORT), None)
+        proxy: config::ProxyConfig {
+            server: config::ServerConfig {
+                bind: listen::Bind::new(SocketAddr::new(LOCALHOST.into(), 0), None)
                     .with_orig_dst_addr(orig_dst.into()),
-                h2_settings,
+                h2_settings: h2::Settings::default(),
```
hawkw, line 17: i think this `Default` impl returns different settings than the actual default values defined in `env.rs`; IIRC it's derived and these are both `None`.

linkerd/app/outbound/src/lib.rs
```diff
@@ -359,82 +427,6 @@ impl Config {
...
-                is_discovery_rejected,
-            )
-            .into_new_service()
-            .check_new_service::<SocketAddr, ()>()
-            .cache(
-                svc::layers().push_on_response(
-                    svc::layers()
-                        .push_failfast(dispatch_timeout)
-                        .push_spawn_buffer_with_idle_timeout(buffer_capacity, cache_max_idle_age)
-                        .push(metrics.stack.layer(stack_labels("tcp"))),
```
hawkw, line 429: what happened to the TCP metrics layer? did that get misplaced, or is it added somewhere else that i'm missing?

olix0r: Mentioned in the PR description

> The metrics have been removed from the TCP balancer stack (for now).

These are just stack metrics (for our own debugging, really). Will probably reintroduce this after the caching stops moving around.

linkerd/app/outbound/src/tests.rs
hawkw: yeah, it's not a big deal, i just thought it was better for this to match the actual defaults everywhere, so that we don't go to write a test that _does_ depend on some setting and get surprising results. not a blocker though.

linkerd/app/outbound/src/lib.rs
hawkw: looks like some of the parameters were also removed?

linkerd/app/outbound/src/lib.rs
hawkw: ah, yup, i can't read. carry on!

hawkw (approved):
lgtm!

linkerd/app/outbound/src/lib.rs
olix0r: Oh, fallback was extracted from tcp_balance https://github.com/linkerd/linkerd2-proxy/pull/662/commits/c18b0aed9c92844a105ebe96e1059892b6bd1518

PR #663, opened 2020-09-18
app: Decouple stacks from listeners

olix0r:
The inbound and outbound configs bind an accept stack to the listener
via serve::serve; but this makes it infeasible to test the accept stacks
without actually binding a socket.

This change moves the server binding into the main app initialization so
that the inbound and outbound stacks can be built without a listener.

### DIFF
--- a/linkerd/app/inbound/src/lib.rs
+++ b/linkerd/app/inbound/src/lib.rs
@@ -8,7 +8,7 @@
 pub use self::endpoint::{HttpEndpoint, ProfileTarget, RequestTarget, Target, TcpEndpoint};
 use self::prevent_loop::PreventLoop;
 use self::require_identity_for_ports::RequireIdentityForPorts;
-use futures::{future, prelude::*};
+use futures::future;
 use linkerd2_app_core::{
     admit, classify,
     config::{ProxyConfig, ServerConfig},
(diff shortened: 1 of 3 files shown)


linkerd/app/inbound/src/lib.rs
```diff
@@ -40,18 +40,29 @@ pub struct Config {
...
+        listen::Addrs,
+        Error = impl Into<Error>,
+        Future = impl Send + 'static,
+        Response = impl tower::Service<
+            tokio::net::TcpStream,
+            Response = (),
+            Error = impl Into<Error>,
+            Future = impl Send + 'static,
+        > + Send
+                       + 'static,
```
hawkw, line 63: :o

linkerd/app/inbound/src/lib.rs
```diff
@@ -275,18 +283,28 @@ impl Config {
...
+    > + Send
+           + 'static
```
hawkw, line 307: IIRC we weren't returning these because there was an impact on build times, is that still the case?

hawkw (approved):
lgtm!

linkerd/app/inbound/src/lib.rs
olix0r: It doesn't appear to be... builds are basically in line with where they are on main (which isn't so speedy, either ;)

olix0r: Eh, it does seem that this, on top of the prior changes, makes integration compilation pretty slow. I'm going to merge it but I think this all in general puts more pressure on us to figure out how to avoid integration tests being soooo costly.

PR #664, opened 2020-09-18
inbound: Split HTTP detection stack from TLS

olix0r:
The DetectTls module only operates on `TcpStream`s (because it uses the
`TcpStream::peek` api); but this complicates writing tests on the HTTP
stack to validate changes like #660.

This change decouples these accept stacks so that they can be tested
more easily.


olix0r:
Depends on https://github.com/linkerd/linkerd2-proxy/pull/663; needs to merge https://github.com/linkerd/linkerd2-proxy/pull/660

hawkw:
FWIW, #658 also changes some stuff to support using mock IO types in place of `TcpStream`, is this related?

hawkw (approved):

PR #665, opened 2020-09-18
integration: Bundle tests in src

olix0r:
Integration tests are split out so each is effectively its own build
target. While this is convenient at times, it also means that compiling
all integration tests is SLOW.

By bundling them all into the integration/src, we avoid having to
recompile/link the proxy many times when building integration tests.


hawkw (approved):

kleimkuhler (approved):
Nice!

PR #667, opened 2020-09-21
fix missing clone impl for MakeSwitch

hawkw:
The cache requires the service type to be `Clone`, but 
`MakeSwitch` doesn't implement `Clone`, even when all
of its constituents services do. We can add a derive to 
fix this, and now everything compiles.


olix0r (approved):
amazing

hawkw:
(tests still don't compile because a test service is being passed the wrong target type, but that was there before i got here and is unrelated to the error)

PR #668, opened 2020-09-22
cache: Implement NewService

olix0r:
The cache used to implement logic in its `poll_ready`; but this was
changed some time ago. Otherwise, the cache is synchronous and
infallible, much like the `NewService` trait.

This change modifies `NewService` to take a mutable reference, as needed
by the cache; and it modifies the cache to implement `NewService`
instead of `Service`.

The stacks then coerce the cache back into a service (via `into_make`)
to satisfy the buffer's requirements. It is expected that this buffering
will be decoupled from the cache in a followup change.

### DIFF
--- a/linkerd/app/core/src/svc.rs
+++ b/linkerd/app/core/src/svc.rs
@@ -44,7 +44,7 @@ pub struct IdentityProxy(());
 
 impl<T> NewService<T> for IdentityProxy {
     type Service = ();
-    fn new_service(&self, _: T) -> Self::Service {
+    fn new_service(&mut self, _: T) -> Self::Service {
         ()
     }
 }
@@ -389,7 +389,7 @@ where
 {
     type Service = N::Service;
 
-    fn new_service(&self, t: T) -> Self::Service {
+    fn new_service(&mut self, t: T) -> Self::Service {
         self.0.new_service(t)
     }
 }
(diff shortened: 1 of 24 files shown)


linkerd/app/core/src/svc.rs
```diff
@@ -180,6 +180,13 @@ impl<S> Stack<S> {
         self.push(stack::new_service::FromMakeServiceLayer::default())
     }
 
+    pub fn into_make<T>(self) -> Stack<stack::new_service::IntoMakeService<S>>
```
hawkw, line 183: nit/TIOLI: the `NewService` conversion layer is `into_new_service`, should this be `into_make_service` rather than `into_make`?

hawkw (approved):
one naming nit, otherwise this lgtm! i'm surprised the change to take a mutable receiver didn't require any other changes!

linkerd/app/core/src/svc.rs
olix0r: There's already `NewService::into_make_service`, and I was hoping to miss that ambiguity (i.e. whether you're getting a `Stack<IntoMakeService<S>>` vs `IntoMakeService<Stack<S>>`.

We could drop the NewService impl on Stack, I guess?

PR #669, opened 2020-09-23
Implement fallback for NewService

olix0r:
This change adds an impementation of `NewService` for `Fallback`.

The returned service holds a primary service and the means to build a
fallback service if the primary service errors before becoming ready.

### DIFF
--- a/linkerd/stack/src/fallback.rs
+++ b/linkerd/stack/src/fallback.rs
@@ -1,4 +1,6 @@
 //! A middleware that may retry a request in a fallback service.
+
+use super::NewService;
 use futures::TryFuture;
 use linkerd2_error::Error;
 use pin_project::pin_project;
@@ -6,6 +8,7 @@ use std::future::Future;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 use tower::util::{Either, Oneshot, ServiceExt};
+
 /// A Layer that augments the underlying service with a fallback service.
 ///
 /// If the future returned by the primary service fails with an error matching a
@@ -21,6 +24,12 @@ pub struct Fallback<I, F, P = fn(&Error) -> bool> {
     inner: I,
     fallback: F,
     predicate: P,
+}
+
+// Falls back to an alternate service if the primary service does not become ready.
+pub struct ReadyFallback<NewB, T, A, B, P> {
+    inner: Either<A, B>,
+    fallback: Option<(NewB, T, P)>,
 }
 
 #[pin_project]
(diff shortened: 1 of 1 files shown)


olix0r:
Sets up https://github.com/linkerd/linkerd2-proxy/pull/670

linkerd/stack/src/fallback.rs
```diff
@@ -1,11 +1,14 @@
...
 use linkerd2_error::Error;
 use pin_project::pin_project;
 use std::future::Future;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 use tower::util::{Either, Oneshot, ServiceExt};
+
 /// A Layer that augments the underlying service with a fallback service.
 ///
 /// If the future returned by the primary service fails with an error matching a
```
hawkw, line 14: it might be worth documenting the difference in behavior between the `NewService` and `MakeService` impls here?

linkerd/stack/src/fallback.rs
```diff
@@ -172,3 +199,57 @@ where
...
+                    "must not be in fallback yet"
+                );
+
+                match futures::ready!(self.inner.poll_ready(cx)) {
+                    Ok(()) => {
+                        // If the primary service becomes ready, drop the
+                        // fallback.
+                        drop(fallback.take().expect("fallback expected"));
+                        Poll::Ready(Ok(()))
+                    }
```
hawkw, line 234: so, we won't use the fallback on errors that occur after the primary becomes ready a single time? i guess that makes sense in the case where the fallback is used for stuff like a discovery lookup returning `InvalidArgument`, or when a port is skipped, or whatever, but it seems surprising in the general case. this is probably right for what we're currently using `Fallback` for, but it might be worth writing this behavior down so we don't use fallback in cases where that doesn't make sense in the future?

linkerd/stack/src/fallback.rs
```diff
@@ -172,3 +199,57 @@ where
...
+                    Err(e) => {
+                        let (mut alt, target, predicate) =
+                            fallback.take().expect("fallback expected");
+                        // If the initial readiness failed in an expected way,
+                        // use the fallback service.
+                        if (predicate)(&e) {
```
hawkw, line 240: nit: the parentheses around a closure/function pointer are only necessary when calling it through a dotted field access (like `self.predicate`). it's not necessary here:
```suggestion
                        if predicate(&e) {
```

hawkw (approved):
this looks right to me --- i had a couple nits, but everything looks correct overall!

PR #670, opened 2020-09-23
Use a single HTTP cache in each direction

olix0r:
Currently, all of the fallback logic is implemented outside of the
cache, and then each stack must be cached individualy.

This change consolidates all caching in each direction to a single
cache instance.

This is accomplished by coercing more of the stack to be implemented as
a `NewService`, rather than `MakeService`.

`ResultService` is introduced so that `Admit` can operate on
`NewService`.

### DIFF
--- a/linkerd/app/inbound/src/lib.rs
+++ b/linkerd/app/inbound/src/lib.rs
@@ -278,7 +278,7 @@ impl Config {
                     svc::layers()
                         .push_failfast(dispatch_timeout)
                         .push_spawn_buffer_with_idle_timeout(buffer_capacity, cache_max_idle_age)
-                        .push(metrics.stack.layer(stack_labels("forward")))
+                        .push(metrics.stack.layer(stack_labels("logical")))
                         .box_http_response(),
                 ),
             )
--- a/linkerd/app/inbound/src/lib.rs
+++ b/linkerd/app/inbound/src/lib.rs
@@ -21,7 +21,7 @@ use linkerd2_app_core::{
     },
     reconnect, router,
     spans::SpanConverter,
-    svc::{self, NewService},
+    svc::{self},
     transport::{self, io, listen, tls},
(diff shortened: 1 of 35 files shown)


linkerd/app/inbound/src/lib.rs
```diff
@@ -248,50 +249,41 @@ impl Config {
...
+                    .check_new_service::<Target, http::Request<_>>()
                     .into_inner(),
             )
+            .check_new_service::<Target, http::Request<http::boxed::Payload>>()
+            .cache(
+                svc::layers().push_on_response(
+                    svc::layers()
+                        .push_failfast(dispatch_timeout)
+                        .push_spawn_buffer_with_idle_timeout(buffer_capacity, cache_max_idle_age)
+                        .push(metrics.stack.layer(stack_labels("logical")))
```
hawkw, line 281: nit: is this still per-logical?

linkerd/stack/src/result.rs
```diff
@@ -0,0 +1,57 @@
...
+use std::task::{Context, Poll};
+
+#[derive(Copy, Clone, Debug)]
+pub struct ResultService<S, E>(Inner<S, E>);
+
+#[derive(Copy, Clone, Debug)]
+enum Inner<S, E> {
+    Ok(S),
+    Err(Option<E>),
+}
```
hawkw, line 12: nit, take it or leave it: isn't this just
```suggestion
pub struct ResultService<S, E>(Result<S, E>);
```

linkerd/app/inbound/src/lib.rs
olix0r: Yeah, the cache holds a service for each logical target (i.e. the apex service in a traffic split).

hawkw (approved):
1️⃣💸👍 (get it? "single cache" -> "single cash" 😉)

olix0r:
Merged via https://github.com/linkerd/linkerd2-proxy/pull/671

PR #671, opened 2020-09-23
discover: Make endpoints with `NewService`

hawkw:
Depends on #567 

This branch changes `linkerd2_proxy_discover::MakeEndpoint` to wrap a
`NewService` rather than a `MakeService`, and construct endpoint
services synchronously rather than asynchronously. This allows removing
a great deal of complex logic for driving and cancelling `MakeService`
futures that was previously necessary.

For now, we just wrap the underlying endpoint stacks in
`into_new_service` to create them lazily. In the future, these stacks
will likely be changed to native `NewService` impls.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -750,6 +750,7 @@ version = "0.1.0"
 dependencies = [
  "futures 0.3.5",
  "linkerd2-error",
+ "linkerd2-stack",
  "tower",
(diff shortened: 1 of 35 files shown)


olix0r (approved):
>  3 files changed, 33 insertions(+), 413 deletions(-)

I like it a lot!

PR #673, opened 2020-09-24
Implement profile discovery as a NewService

olix0r:
The profile discovery layer is always coerced to a NewService. This
change formalizes this to avoid stack boilerplate.

While we're here, the GetProfile trait has ben simplified to avoid
exposing poll_ready in its public API. Instead, we provide an impl of
Service for GetProfile that does this (though it's not currently
needed).


hawkw (approved):
lovely

PR #674, opened 2020-09-24
Make HTTP client stacks synchronous

olix0r:
This change implements `NewService` for `Reconnect` and other
HTTP-client middlewares. HTTP clients are almost always coerced to
`NewService`, so this reduces boilerplate and type complexity.


hawkw (approved):
:+1:

PR #675, opened 2020-09-25
Update the accept stack to use NewService

olix0r:
This change modifies `serve` to take a `NewService` instead of a
`MakeService`. Services specific to the accept stack have been
updated as well

`DetectHttp` has been updated to work as either a `MakeService` or
`NewService` -- the asynchronous version is still needed by the outbound
proxy (until caching is changed).

`DetectTls` is now purely a `NewService`.

### DIFF
--- a/linkerd/app/core/src/serve.rs
+++ b/linkerd/app/core/src/serve.rs
@@ -1,8 +1,9 @@
+use crate::svc;
 use futures::prelude::*;
 use linkerd2_error::Error;
 use linkerd2_proxy_transport::listen::Addrs;
 use tower::util::ServiceExt;
-use tracing::{debug, error, info, info_span};
+use tracing::{debug, info, info_span};
 use tracing_futures::Instrument;
 
 /// Spawns a task that binds an `L`-typed listener with an `A`-typed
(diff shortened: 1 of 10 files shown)


linkerd/app/core/src/serve.rs
```diff
@@ -43,22 +42,14 @@ where
                     //
                     // This allows the service to propagate errors and to exert backpressure on the
                     // listener. It also avoids a `Clone` requirement.
```
olix0r, line 44: comment is outdated

linkerd/app/inbound/src/lib.rs
```diff
@@ -351,15 +343,13 @@ impl Config {
         ));
 
         let http_server = svc::stack(http_router)
-            // Limits the amount of time each request waits to obtain a
-            // ready service.
-            .push_timeout(dispatch_timeout)
```
olix0r, line 356: This should probably become a failfast, but we already have a failfast on the http router...

linkerd/app/outbound/src/lib.rs
```diff
@@ -402,17 +402,15 @@ impl Config {
...
-        Future = impl Send + 'static,
-        Response = impl tower::Service<
+        Service = impl tower::Service<
             I,
             Response = (),
             Error = impl Into<Error>,
             Future = impl Send + 'static,
         > + Send
-                       + 'static,
+                      + 'static,
```
hawkw, line 413: it is so weird that rustfmt decided to remove...one space...from these `'static` bounds???

linkerd/proxy/transport/src/tls/accept.rs
```diff
@@ -130,7 +121,7 @@ where
 
     fn call(&mut self, tcp: TcpStream) -> Self::Future {
         let addrs = self.addrs.clone();
-        let make = self.inner.clone();
+        let mut make = self.inner.clone();
```
hawkw, line 124: nit: might not wanna call this `make` any more? :)

hawkw (approved):

PR #676, opened 2020-09-25
Split prefix buffering from HTTP detection

olix0r:
Prefix buffering is necessarily asynchronous. HTTP detection is not.
Split these concerns by extracing a `transport::Prefix` middleware.

### DIFF
--- a/linkerd/app/outbound/src/lib.rs
+++ b/linkerd/app/outbound/src/lib.rs
@@ -402,17 +402,15 @@ impl Config {
         metrics: ProxyMetrics,
         span_sink: Option<mpsc::Sender<oc::Span>>,
         drain: drain::Watch,
-    ) -> impl tower::Service<
+    ) -> impl svc::NewService<
         listen::Addrs,
-        Error = impl Into<Error>,
-        Future = impl Send + 'static,
-        Response = impl tower::Service<
+        Service = impl tower::Service<
             I,
             Response = (),
             Error = impl Into<Error>,
             Future = impl Send + 'static,
         > + Send
-                       + 'static,
+                      + 'static,
     > + Send
            + 'static
     where
@@ -530,8 +528,8 @@ impl Config {
 
         svc::stack(svc::stack::MakeSwitch::new(
             skip_detect.clone(),
-            http,
-            tcp_forward.push_map_target(TcpEndpoint::from),
+            svc::stack(http).into_new_service().into_inner(),
+            tcp_forward.push_map_target(TcpEndpoint::from).into_inner(),
         ))
         .push(metrics.transport.layer_accept(TransportLabels))
         .into_inner()
--- a/linkerd/app/outbound/src/lib.rs
+++ b/linkerd/app/outbound/src/lib.rs
@@ -518,17 +518,23 @@ impl Config {
             .push_map_target(|a: listen::Addrs| a.target_addr())
             .into_inner();
 
-        let http = http::DetectHttp::new(
+        const DETECT_BUFFER_CAPACITY: usize = 8192;
+        let http = svc::stack(http::DetectHttp::new(
             h2_settings,
-            detect_protocol_timeout,
             http_server,
             tcp_balance,
             drain.clone(),
-        );
+        ))
+        .push_on_response(transport::Prefix::layer(
+            DETECT_BUFFER_CAPACITY,
+            detect_protocol_timeout,
+        ))
+        .into_new_service()
+        .into_inner();
 
         svc::stack(svc::stack::MakeSwitch::new(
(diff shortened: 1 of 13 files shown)


linkerd/app/outbound/src/lib.rs
```diff
@@ -518,17 +518,23 @@ impl Config {
             .push_map_target(|a: listen::Addrs| a.target_addr())
             .into_inner();
 
-        let http = http::DetectHttp::new(
+        const DETECT_BUFFER_CAPACITY: usize = 8192;
```
hawkw, line 521: nit: if this has the same value between inbound and outbound, and is unlikely to change or be configurable, why not just stick the constant in `transport` and have the prefix layer always use that value?

linkerd/app/outbound/src/lib.rs
```diff
@@ -518,17 +518,23 @@ impl Config {
...
-            detect_protocol_timeout,
             http_server,
             tcp_balance,
             drain.clone(),
-        );
+        ))
+        .push_on_response(transport::Prefix::layer(
+            DETECT_BUFFER_CAPACITY,
+            detect_protocol_timeout,
+        ))
```
hawkw, line 531: just a thought: it seems like it might be possible to split the timeout out of the prefix layer and just build a little stack with a prefix layer and then a timeout layer here. i don't know if this is actually worth it, as the code for adding the timeout in the prefix layer is quite simple...so it probably doesn't matter...

linkerd/proxy/transport/src/prefix.rs
```diff
@@ -0,0 +1,75 @@
...
+    fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(().into()))
+    }
+
+    fn call(&mut self, io: I) -> Self::Future {
+        debug!(capacity = %self.capacity, "Buffering prefix");
+        let accept = self.inner.clone();
+        let peek = time::timeout(self.timeout, io.peek(self.capacity)).map_err(|_| ReadTimeout(()));
+        Box::pin(async move {
+            let io = peek.await??;
```
hawkw, line 62: it might be nice to stick a `trace`-level span on the peek future, so we can time the span when debugging. not a blocker though and we can always just add it later if we want to look at this.

linkerd/proxy/transport/src/prefix.rs
```diff
@@ -0,0 +1,75 @@
...
+    type Response = ();
+    type Error = Error;
+    type Future = Pin<Box<dyn Future<Output = Result<(), Error>> + Send + 'static>>;
+
+    fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(().into()))
+    }
+
+    fn call(&mut self, io: I) -> Self::Future {
+        debug!(capacity = %self.capacity, "Buffering prefix");
```
hawkw, line 58: nit: `usize` is a `tracing` primitive `Value` type, so we don't really need to use `fmt::Display` here:
```suggestion
        debug!(capacity = self.capacity, "Buffering prefix");
```

linkerd/proxy/transport/src/prefix.rs
```diff
@@ -0,0 +1,75 @@
...
+            let io = peek.await??;
+            accept.oneshot(io).err_into::<Error>().await?;
+            Ok(())
+        })
+    }
+}
+
+impl std::fmt::Display for ReadTimeout {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Timed out while reading client stream")
```
hawkw, line 71: it might be good if this error message mentioned that we were trying to read the prefix? although, on the other hand...we just fall back on these, so they don't really get exposed to the user, so the error message doesn't _really_ matter...

hawkw (approved):
a few minor notes, but everything looks great overall!

linkerd/proxy/transport/src/prefix.rs
olix0r: We don't actually fall back on these, so these manifest as connection errors. I'm not sure that the prefix part of it matters, though. It's really just a read timeout -- the client hasn't provided any data.

linkerd/app/outbound/src/lib.rs
olix0r: writing that timeout layer is a bit complicated, as it has to be both inside and outside of the inner service (we're timing out the time between the request entering the outer layer and entering the inner layer). i'm not opposed to it, but i'd rather keep it simple.

hawkw (approved):

PR #679, opened 2020-09-28
outbound: Restore spawn-ready

olix0r:
The spawn-ready layer was removed in 1382e32fa; but this means that
endpoint-level services aren't necessarily driven to readiness. This
can manifest in TLS detection timeouts, since a client may not be driven
through the TLS handshake.

This change removes the `push_spawn_ready` helper, because it's really a
`push_make_spawn_ready` and the make layer is totally superfluous.


hawkw (approved):
ah, yup, this makes sense. good catch!

PR #680, opened 2020-09-29
outbound: Explicitly ignore the source address for tap

olix0r:
For outbound requests, the src address is always an ephemeral port on
localhost, which isn't particularly valuable. Futhermore, as we add
caching outside of HTTP detection, this becomes annoying to track.

This changes the outbound stack to always use a `None` src addr.

### DIFF
--- a/linkerd/app/outbound/src/endpoint.rs
+++ b/linkerd/app/outbound/src/endpoint.rs
@@ -172,7 +172,7 @@ impl Into<http::client::Settings> for &'_ HttpEndpoint {
 }
 
 impl tap::Inspect for HttpEndpoint {
-    fn src_addr<B>(&self, req: &http::Request<B>) -> Option<SocketAddr> {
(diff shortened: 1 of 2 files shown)


linkerd/app/outbound/src/endpoint.rs
```diff
@@ -173,7 +173,7 @@ impl Into<http::client::Settings> for &'_ HttpEndpoint {
 
 impl tap::Inspect for HttpEndpoint {
     fn src_addr<B>(&self, req: &http::Request<B>) -> Option<SocketAddr> {
```
hawkw, line 175: [`req` is unused](https://github.com/linkerd/linkerd2-proxy/pull/680/checks?check_run_id=1183684537#step:3:509), changing this to
```suggestion
    fn src_addr<B>(&self, _: &http::Request<B>) -> Option<SocketAddr> {
```
should fix the build.

hawkw (approved):
looks good to me, once we get CI to pass

PR #684, opened 2020-09-30
profiles: Expose the fully_qualified_name

olix0r:
Service profiles may include an optional service name, which is
especially useful if the profile is resolved by IP. This change attempts
to parse this value as a DNS name. If the default (empty) name is
returned, no name is exposed.


kleimkuhler (approved):
👍

PR #685, opened 2020-09-30
request-filter: Allow altering request type

olix0r:
The request filter takes ownership of the request but does not support
changing its type. Furthermore, the trait requires that the error type
be named even though it always coerced to an `Error`.

This change cleans up the request-filter module as follows:

- The `RequestFilter` trait is now named `FilterRequest` (traits
  generally are verbs). This type now has a `Request` type attribute
  instead of an `Error` type attribute.
- The `Service` has been renamed to `RequestFilter` and the
  `RequstFilterLayer` type has been eliminated.
- The manual future implementation can be eliminated with `Either`.
- The stack helper has been removed, as it's only used in one place.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1462,7 +1462,7 @@ version = "0.1.0"
 dependencies = [
  "futures 0.3.5",
  "linkerd2-error",
- "pin-project",
+ "linkerd2-stack",
  "tower",
  "tracing",
 ]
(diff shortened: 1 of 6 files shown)


hawkw (approved):
this all seems good to me!

i had to double check whether or not `RequestFilter` was derived from an upstream Tower thing that we might want to contribute back to, but it looks like the `tower::filter` stuff has a separate heritage.

PR #686, opened 2020-10-01
tracing: Set contexts in new_service/make_service

olix0r:
* Modify stack-tracing to set contexts on the initial new_service/make_service invocation (and not just the resulting future/service)
* Tighten up the HTTP context. `http{version=http/1.x}` was verbose and redundant

### DIFF
--- a/linkerd/stack/tracing/src/lib.rs
+++ b/linkerd/stack/tracing/src/lib.rs
@@ -88,11 +88,10 @@ where
     type Future = Instrument<M::Future>;
 
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        trace!("make ready");
         let ready = self.make.poll_ready(cx);
         match ready {
-            Poll::Pending => trace!(ready = false),
-            Poll::Ready(ref res) => trace!(ready = true, ok = res.is_ok()),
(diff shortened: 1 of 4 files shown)


linkerd/stack/tracing/src/lib.rs
```diff
@@ -88,16 +88,20 @@ where
     type Future = Instrument<M::Future>;
 
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        trace!("poll_ready");
         let ready = self.make.poll_ready(cx);
-        trace!(ready = ready.is_ready());
+        match ready {
+            Poll::Pending => trace!(ready = false, "make"),
+            Poll::Ready(ref res) => trace!(ready = true, ok = res.is_ok(), "make"),
```
hawkw, line 94: TIOLI: if `res` is an error, should we include the error in the event?

hawkw (approved):
lovely!

linkerd/stack/tracing/src/lib.rs
olix0r: adding an error constraint means adding more type constraints in more places, so would prefer to avoid that for now. we tend to log errors otherwise.,

hawkw: fair, it's not a blocker

PR #688, opened 2020-10-01
fix default profile stream hanging forever

hawkw:
`stream::once` returns `None` the second time it's polled, indicating
EOF. However, the proxy's discovery code expects discovery streams to
never return `None`, and either be terminated by errors, or return
`Pending` when there are no remaining updates at the current point in
time. Because `stream::once` returns `None` instead, the
`discovery::buffer` task hangs forever.

This commit fixes that by changing `default_resolve` to return a stream
that returns the initial update and then returns `Poll::Pending`
forever, rather than returning `None`. We do this by chaining the
`stream::once` with a `stream::pending`.

### DIFF
--- a/linkerd/app/src/dst/default_resolve.rs
+++ b/linkerd/app/src/dst/default_resolve.rs
@@ -24,12 +24,17 @@ where
     S::Endpoint: Default + Send + 'static,
     S::Resolution: Send + 'static,
     S::Future: Send + 'static,
-    stream::Once<future::Ready<Result<Update<S::Endpoint>, S::Error>>>:
-        stream::TryStream<Ok = Update<S::Endpoint>, Error = S::Error>,
+    stream::Chain<
+        stream::Once<future::Ready<Result<Update<S::Endpoint>, Error>>>,
+        stream::Pending<Result<Update<S::Endpoint>, Error>>,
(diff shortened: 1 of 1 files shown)


linkerd/app/src/dst/default_resolve.rs
```diff
@@ -24,12 +24,17 @@ where
...
+        stream::Pending<Result<Update<S::Endpoint>, Error>>,
+    >: stream::TryStream<Ok = Update<S::Endpoint>, Error = S::Error>,
 {
     type Response = future::Either<
         S::Resolution,
-        stream::Once<future::Ready<Result<Update<S::Endpoint>, Error>>>,
+        stream::Chain<
+            stream::Once<future::Ready<Result<Update<S::Endpoint>, Error>>>,
+            stream::Pending<Result<Update<S::Endpoint>, Error>>,
+        >,
```
hawkw, line 37: yes, i know this part is kinda terrible, sorry. we could box these if you prefer...

olix0r:
Thanks for tracking this down! I took an alternate approach in 89670dfa, to stop erroring (and busy looping) in these cases.

hawkw:
Cool, I'll just close this one then!

PR #689, opened 2020-10-01
discover: Allow resolution streams to terminate

olix0r:
In 571f4cba we accidentally introduced a busy-loop on resolution stream
termination; and, furthermore, the buffer layer returns an error when
the inner stream terminates.

This change alters both of these to gracefully propagate stream
termination.


hawkw (approved):
🗺️📡🔚👍👍

PR #690, opened 2020-10-02
outbound: Return a default endpoint on reject

olix0r:
When the resolver rejects resolution, we currently propagate that error
so that it can be handled via fallback. And due to recent HTTP router
changes, these resolution errors can propagate up across splits, etc.

This change simplifies this behavior by instead synthesizing a
resolution with a default endpoint.

The `not_http` reason has been removed, as it's no longer useful.

### DIFF
--- a/linkerd/app/outbound/src/lib.rs
+++ b/linkerd/app/outbound/src/lib.rs
@@ -20,8 +20,8 @@ use linkerd2_app_core::{
     spans::SpanConverter,
     svc::{self},
     transport::{self, listen, tls},
-    Conditional, DiscoveryRejected, Error, ProxyMetrics, StackMetrics, TraceContextLayer,
-    CANONICAL_DST_HEADER, DST_OVERRIDE_HEADER, L5D_REQUIRE_ID,
+    Conditional, Error, ProxyMetrics, StackMetrics, TraceContextLayer, CANONICAL_DST_HEADER,
+    DST_OVERRIDE_HEADER, L5D_REQUIRE_ID,
 };
 use std::{collections::HashMap, net, time::Duration};
 use tokio::sync::mpsc;
@@ -500,13 +500,6 @@ impl Config {
 
         // Load balances TCP streams that cannot be decoded as HTTP.
         let tcp_balance = svc::stack(self.build_tcp_balance(tcp_connect, resolve))
-            .push_fallback_with_predicate(
-                tcp_forward
-                    .clone()
-                    .push_map_target(TcpEndpoint::from)
-                    .into_inner(),
-                is_discovery_rejected,
-            )
             .push_on_response(
                 svc::layers()
                     .push_failfast(dispatch_timeout)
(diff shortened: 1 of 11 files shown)


kleimkuhler (approved):
👍

linkerd/app/src/dst/default_resolve.rs
```diff
@@ -0,0 +1,59 @@
...
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+pub fn layer<S>() -> impl svc::Layer<S, Service = RecoverDefaultResolve<S>> + Clone {
+    svc::layer::mk(RecoverDefaultResolve)
+}
+
+#[derive(Clone, Debug)]
+pub struct RecoverDefaultResolve<S>(S);
```
hawkw, line 18: nit: it'd be nice to have a comment explaining what this does --- the name is about as descriptive as i can think of, but i can see the purpose of this thing not being immediately obvious to new contributors?

linkerd/app/src/dst/mod.rs
```diff
@@ -21,6 +23,9 @@ pub struct Config {
     pub initial_profile_timeout: Duration,
 }
 
+#[derive(Clone, Debug)]
+pub struct Rejected(());
```
hawkw, line 27: nit: maybe worth a comment on the semantics of what this error means?

linkerd/app/outbound/src/lib.rs
```diff
@@ -591,9 +584,7 @@ pub fn trace_labels() -> HashMap<String, String> {
 
 fn is_discovery_rejected(err: &Error) -> bool {
```
hawkw, line 585: this is now specifically for profiles, not for discovery --- maybe it needs a new name?

hawkw: also: nit/TIOLI: might be nice to refactor this  (and `is_loop`) into something more like `Rejected::matches` for consistency (definitely not a big deal though)

linkerd/app/src/dst/default_resolve.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    fn call(&mut self, t: T) -> Self::Future {
+        let addr = (&t).into();
+        Box::pin(
+            self.0
+                .resolve(t)
+                .map_ok(future::Either::Left)
+                .or_else(move |error| {
+                    if Rejected::matches(&*error) {
+                        tracing::debug!(%error, %addr, "Synthesizing endpoint");
```
hawkw, line 49: nit/TIOLI: if we added `addr` to this message by `instrument`ing the future instead of moving it in, we wouldn't need this to be a `move` closure and could therefore remove the box around the future (since it's just combinators). not a blocker though; and the concrete type is probably pretty gross...

hawkw (approved):
lgtm, commented on a few minor things that aren't blockers

linkerd/app/outbound/src/lib.rs
hawkw: oh, i see that we're getting rid of this in the next branch, nvm.

linkerd/app/src/dst/default_resolve.rs
olix0r: The `addr` is used on the next line (as part of the returend stream), so the `move` is needed either way. And, really, the `addr` is already on the context, so we could omit it here, but 🤷‍♂️

olix0r:
I'll add comments in https://github.com/linkerd/linkerd2-proxy/pull/691

linkerd/app/src/dst/default_resolve.rs
hawkw: whoops, i overlooked that --- carry on!

PR #691, opened 2020-10-02
Skip endpoint resolution when profile lookup is rejected

olix0r:
When profile lookup is rejected, this error propagates to the stack and
is handled by falling back to an alternate stack.

In order to move profile discovery out of the HTTP stack and onto the
initial TCP accept stack, we want to avoid this sort of fallback
behavior. So this change modifies profile discovery to make the profile
optional. When a profile lookup is rejected/skipped, we now simply
return a `None` profile; and in these cases we avoid performing endpoint
resolution by omitting a concrete address.

### DIFF
--- a/linkerd/app/src/dst/default_profile.rs
+++ b/linkerd/app/src/dst/default_profile.rs
@@ -1,4 +1,4 @@
-use super::InvalidProfileAddr;
+use super::Rejected;
 use futures::prelude::*;
 use linkerd2_app_core::{profiles, svc, Addr, Error};
 use std::{
@@ -35,7 +35,7 @@ where
     fn call(&mut self, dst: T) -> Self::Future {
         Box::pin(self.inner.call(dst).or_else(move |e| {
             let err = e.into();
-            if is_rejected(&*err) {
+            if Rejected::matches(&*err) {
                 debug!("Handling rejected discovery");
                 future::ok(None)
             } else {
@@ -43,16 +43,4 @@ where
             }
         }))
     }
-}
-
-fn is_rejected(err: &(dyn std::error::Error + 'static)) -> bool {
-    if err.is::<InvalidProfileAddr>() {
-        return true;
-    }
-
-    if let Some(status) = err.downcast_ref::<tonic::Status>() {
-        return status.code() == tonic::Code::InvalidArgument;
-    }
-
-    err.source().map(is_rejected).unwrap_or(false)
 }
--- a/linkerd/app/src/dst/default_profile.rs
+++ b/linkerd/app/src/dst/default_profile.rs
@@ -1,0 +1,58 @@
+use super::InvalidProfileAddr;
+use futures::prelude::*;
+use linkerd2_app_core::{profiles, svc, Addr, Error};
+use std::{
+    pin::Pin,
+    task::{Context, Poll},
+};
+use tracing::debug;
+
+pub fn layer<S>() -> impl svc::Layer<S, Service = RecoverDefaultProfile<S>> + Clone {
+    svc::layer::mk(|inner| RecoverDefaultProfile { inner })
+}
+
+#[derive(Clone, Debug)]
+pub struct RecoverDefaultProfile<S> {
+    inner: S,
+}
+
+impl<T, S> tower::Service<T> for RecoverDefaultProfile<S>
+where
+    for<'t> &'t T: Into<Addr>,
+    S: tower::Service<T, Response = Option<profiles::Receiver>>,
+    S::Error: Into<Error>,
+    S::Future: Send + 'static,
+{
+    type Response = Option<profiles::Receiver>;
+    type Error = Error;
+    type Future =
+        Pin<Box<dyn Future<Output = Result<Option<profiles::Receiver>, Error>> + Send + 'static>>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx).map_err(Into::into)
+    }
+
+    fn call(&mut self, dst: T) -> Self::Future {
+        Box::pin(self.inner.call(dst).or_else(move |e| {
+            let err = e.into();
+            if is_rejected(&*err) {
+                debug!("Handling rejected discovery");
+                future::ok(None)
+            } else {
+                future::err(err)
+            }
+        }))
+    }
+}
+
+fn is_rejected(err: &(dyn std::error::Error + 'static)) -> bool {
+    if err.is::<InvalidProfileAddr>() {
+        return true;
+    }
+
+    if let Some(status) = err.downcast_ref::<tonic::Status>() {
+        return status.code() == tonic::Code::InvalidArgument;
+    }
+
+    err.source().map(is_rejected).unwrap_or(false)
+}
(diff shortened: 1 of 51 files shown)


linkerd/app/src/dst/default_profile.rs
```diff
@@ -0,0 +1,45 @@
...
+impl<T, S> tower::Service<T> for RecoverDefaultProfile<S>
+where
+    S: tower::Service<T, Response = Option<profiles::Receiver>>,
+    S::Error: Into<Error>,
+    S::Future: Send + 'static,
+{
+    type Response = Option<profiles::Receiver>;
+    type Error = Error;
+    type Future =
+        Pin<Box<dyn Future<Output = Result<Option<profiles::Receiver>, Error>> + Send + 'static>>;
```
hawkw, line 28: nit/TIOLI: the closure in the `or_else` doesn't capture anything, so i think this _could_ be 
```suggestion
    type Future = future::OrElse<
        S::Future,
        future::Ready<Result<Self::Output, Self::Error>>
        fn(S::Error) -> future::Ready<Result<Self::Output, Self::Error>>,
     >;
```
if we wanted to remove the box. in practice, i don't know if the box here matters at all, but since we're already using combinators with a non-capturing closure, i thought it was worth pointing out.

linkerd/app/src/dst/permit.rs
```diff
@@ -1,79 +1,96 @@
...
+    fn filter(&self, t: T) -> Result<Addr, Error> {
+        let permitted = (&t).into().and_then(|addr| {
+            if self.0.matches(&addr) {
+                Some(addr)
+            } else {
+                None
+            }
+        });
+        tracing::debug!(permitted = %permitted.is_some(), "Resolve");
+        permitted.ok_or_else(|| Rejected(()).into())
```
hawkw, line 79: i was a little surprised at first that there are two separate implementations for resolve and profiles which _look_ equivalent after a quick skim, but i didn't notice the `Option` until i took a second look at it. it does occur to me that we could probably only have one permit implementation if we added `Into<Option<Addr>>` impls that always return `Some` for profile targets. 

i don't think this is like, important --- the small amount of duplication is totally fine with me, i was just surprised until i looked closer.

linkerd/service-profiles/src/split.rs
```diff
@@ -58,31 +61,67 @@ impl<N: Clone, S, Req> Clone for NewSplit<N, S, Req> {
...
-        let rx = target.as_ref().clone();
-        Split {
-            rx,
-            target,
-            new_service: self.inner.clone(),
-            rng: self.rng.clone(),
-            inner: None,
-            services: ReadyCache::default(),
-        }
+        let inner = match Into::<Option<Receiver>>::into(&target) {
```
hawkw, line 75: nit/tioli: personally i might have written this with an early return in the `None` case instead of putting most of the code inside the match, like 
```rust
let rx = match Into::<Option<Receiver>>::into(&target) {
    Some(rx) => rx,
    None => {
        trace!("Building default service");
        return Self { inner: Inner::Default(self.inner.new_service((None, target))) };
    },
};

let mut targets = rx.borrow().targets.clone();
// ...
```
but i'm not actually sure if this is any clearer...

linkerd/service-profiles/src/split.rs
```diff
@@ -93,116 +132,84 @@ where
...
+                        services.evict(&addr);
+                    }
+                }
 
-        // Wait for all target services to be ready. If any services fail, then
-        // the whole service fails.
-        Poll::Ready(ready!(self.services.poll_pending(cx)).map_err(Into::into))
+                // Wait for all target services to be ready. If any services fail, then
+                // the whole service fails.
+                Poll::Ready(ready!(services.poll_pending(cx)).map_err(Into::into))
```
hawkw, line 189: this logic used to all live in `update_inner` and was just moved here as a result of `inner` becoming an enum, right? the diff didn't make that super obvious so i wanted to double check

linkerd/service-profiles/src/split.rs
```diff
@@ -58,31 +61,67 @@ impl<N: Clone, S, Req> Clone for NewSplit<N, S, Req> {
...
+                let mut services = ReadyCache::default();
+                let mut new_service = self.inner.clone();
+                for Target { weight, addr } in targets.into_iter() {
+                    services.push(
+                        addr.clone(),
+                        new_service.new_service((Some(addr.clone()), target.clone())),
+                    );
+                    addrs.insert(addr);
+                    weights.push(weight);
+                }
```
hawkw, line 101: i guess the only real change in this file besides adding the default case is that we now call `new_service` eagerly in `new_service` on splits, instead of doing it lazily in the first `poll_ready`? 

this _seems_ better regardless; less moving parts and `new_service` is expected to be called eagerly. i can't imagine anything relying on it happening in `poll_ready`?

hawkw (commented):
this all seems good to me after reading over it. there were a couple of things that seemed kind of weird but beyond a couple small nits i think this looks right.

linkerd/service-profiles/src/split.rs
olix0r: There's no functional change, but the block has been cleaned up slightly (mostly pushing a default target on to targets when there are none) to reduce duplication.

linkerd/service-profiles/src/split.rs
olix0r: Yeah, that's right

linkerd/app/src/dst/default_profile.rs
```diff
@@ -8,32 +8,29 @@ use std::{
 use tracing::debug;
 
 pub fn layer<S>() -> impl svc::Layer<S, Service = RecoverDefaultProfile<S>> + Clone {
-    svc::layer::mk(|inner| RecoverDefaultProfile { inner })
+    svc::layer::mk(RecoverDefaultProfile)
 }
 
+/// Wraps a `GetProfile to produce no profile when the lookup is rejected.
```
hawkw, line 14: typo: 
```suggestion
/// Wraps a `GetProfile` to produce no profile when the lookup is rejected.
```

linkerd/app/src/dst/permit.rs
olix0r: I had something like that initially but i think it's a bit clearer to have them separated (for now at least).

linkerd/app/src/dst/permit.rs
```diff
@@ -3,9 +3,13 @@ use ipnet::{Contains, IpNet};
 use linkerd2_app_core::{dns::Suffix, request_filter::FilterRequest, Addr, Error};
 use std::{net::IpAddr, sync::Arc};
 
+/// Rejects profile ookups if the destinatino address is outside of the
```
hawkw, line 6: typos:
```suggestion
/// Rejects profile .ookups if the destination address is outside of the
```

linkerd/app/src/dst/permit.rs
```diff
@@ -3,9 +3,13 @@ use ipnet::{Contains, IpNet};
...
+/// configured networks/domains.
 #[derive(Clone, Debug)]
 pub struct PermitProfile(Inner);
 
+/// Rejects endpoint resolutions if the destinatino address is outside of the
+/// configured networks/domains or if there is no resolveable concrete address.
```
hawkw, line 12: typos/spelling
```suggestion
/// Rejects endpoint resolutions if the destination address is outside of the
/// configured networks/domains or if there is no resolvable concrete address.
```

hawkw (approved):
looks great, thanks for adding comments! i noticed a couple copyedits but the change all makes sense

PR #692, opened 2020-10-04
Unify RequestFilter and Admit middlewares

olix0r:
`RequestFilter` and `Admit` are basically exactly the same. This change
moves `RequestFilter` into the `stack` crate and moves all `Admit`
implementations to `RequestFilter`. This change also restores a `Stack`
helper for `RequestFilter`, as there are now many uses.

This change also simplifies the `prevent_loop` modules to be generic over
target types that can be coerced to `SocketAddr`.


kleimkuhler (approved):
Makes sense 👍

PR #693, opened 2020-10-05
outbound: initial tests for TCP mTLS (with fewer moving parts)

hawkw:
This branch introduces a second pass at unit tests for TCP mTLS in the
outbound proxy, without the complexity of actually performing handshakes
on mock IOs (as proposed in #658). The new tests just rely on assertions
that the connect stack receives the expected peer identity metadata. We
can test that the handshake is performed correctly in separate tests for
_just_ the TLS client layer, while avoiding the complexity necessary to
use mock IOs in the existing connect stack. This also means we don't 
have to actually load and parse all the test key material we use in the
integration tests.

If this approach seems better, I'll open further PRs to add more tests
in this style.

Closes #658


olix0r (approved):
looks good!

PR #694, opened 2020-10-06
stack: add SwitchReady service

hawkw:
This is essentially a form of fallback that falls back when a primary
service takes too long to become ready. When the primary service's
`poll_ready` does not return ready for over the specified duration, the
secondary service is used until the primary becomes ready again,
resetting the readiness timeout. Erros from both services are
propagated.

I considered naming this `ReadyFallback`, but that name was already used
for an unrelated type in the `fallback` module. :)


linkerd/stack/src/switch_ready.rs
```diff
@@ -0,0 +1,362 @@
...
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+    time::Duration,
+};
+use tokio::time::{delay_for, Delay, Instant};
+
+#[derive(Debug, Clone)]
+pub struct SwitchReadyLayer<B> {
```
olix0r, line 11: For now, let's skip the impl of SwitchReadyLayer -- this is something we can easily do via layer::mk as needed for now, and start with a `SwitchReady::new` instead.

olix0r (changes requested):
This looks really good! Thanks for tests.

* Let's skip the layer impl for now and instead expose a SwitchReady::new
* Let's base this directly off of `main` so it can merge independently?

linkerd/stack/src/switch_ready.rs
hawkw: yup, i considered using `layer::mk`, and can happily make that change now.

linkerd/stack/src/switch_ready.rs
```diff
@@ -0,0 +1,389 @@
+use super::NewService;
+use futures::future::Either;
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+    time::Duration,
+};
+use tokio::time::{delay_for, Delay, Instant};
+/// A service which falls back to a secondary service if the primary service
```
olix0r, line 10: newline, please

linkerd/stack/src/switch_ready.rs
```diff
@@ -0,0 +1,389 @@
...
+            delay: delay_for(switch_after),
+            state: State::Primary,
+        }
+    }
+}
+
+impl<A, B, R> tower::Service<R> for SwitchReady<A, B>
+where
+    A: tower::Service<R>,
+    B: tower::Service<R, Response = A::Response, Error = A::Error>,
```
olix0r, line 94: note that if you use `tower::util::Either`, you don't need to require that the error types are the same (though they both need to be `Into<Error>`)

olix0r (approved):
hooray, this looks great

linkerd/stack/src/switch_ready.rs
hawkw: whoopsy!

linkerd/stack/src/switch_ready.rs
hawkw: didn't realize `Either` was `Into<Error>` when both arms are `Into<Error>`...mainly i thought i'd avoid adding a `map_err` here that could be added in a separate layer?

hawkw: oh, i see, it has a future impl, that's neat

PR #695, opened 2020-10-06
Only allow name-based profile discovery for inbound requests

olix0r:
The proxy uses a single profile client for both inbound and outbound
resolutions. We now want to perform profile discovery for all outbound
traffic; but this is overkill for inbound traffic. We should only do
profile lookup for traffic that includes a valid service name and not,
for instance, healtchecks and prometheus scrapes.

This changes inbound to only do resolution for valid names, skipping
traffic that is not targetting a valid service name.

In order to do this, the following has changed:

- An `AddrMatch` utility is introduced to `core` so that this logic need
  not be propagated.
- `app::dst` now simply builds a profile client without filtering or
  recovery.
- The profile client is no longer responsible for synthesizing a
  default profile on timeout -- this will be extracted into the stack in
  a followup.
- The `profiles::discovery::layer` utility now applies filtering and
  discovery, so that the inbound and outbound proxies may supply their own
  filters.
* When we reject discovery and handle failed discovery, we always use gRPC
   error codes now


linkerd/app/core/src/addr_match.rs
```diff
@@ -0,0 +1,59 @@
...
+        suffixes: impl IntoIterator<Item = Suffix>,
+        nets: impl IntoIterator<Item = IpNet>,
+    ) -> Self {
+        Self(NameMatch::new(suffixes), IpMatch::new(nets))
+    }
+
+    pub fn matches(&self, addr: &Addr) -> bool {
+        match addr {
+            Addr::Name(ref name) => self.0.matches(name.name()),
+            Addr::Socket(sa) => self.1.matches(sa.ip()),
```
hawkw, line 28: nit/TIOLI: this might read nicer if we named the two fields of `AddrMatch`, but that's not a huge deal...

linkerd/app/inbound/src/lib.rs
```diff
@@ -243,7 +247,14 @@ impl Config {
                     .into_inner(),
             ))
             .push_map_target(endpoint::Logical::from)
-            .push(profiles::discover::layer(profiles_client))
+            // TODO: Profile resolution should be time-bounded so that slowness
+            // cannot cause inbound requests to fail.
```
hawkw, line 251: shall we merge #694 first and use that here, or should we do that as a follow-up? probably fine either way?

linkerd/app/src/dst/permit.rs
```diff
@@ -1,67 +1,19 @@
...
-/// Rejects profile lookups if the destination address is outside of the
-/// configured networks/domains.
-#[derive(Clone, Debug)]
-pub struct PermitProfile(Inner);
+use ipnet::IpNet;
+use linkerd2_app_core::{
+    discovery_rejected, dns::Suffix, svc::stack::FilterRequest, Addr, AddrMatch, Error,
+};
 
 /// Rejects endpoint resolutions if the destinatino address is outside of the
```
hawkw, line 6: typo:
```suggestion
/// Rejects endpoint resolutions if the destination address is outside of the
```

linkerd/app/src/env.rs
```diff
@@ -229,7 +229,7 @@ const DEFAULT_OUTBOUND_MAX_IN_FLIGHT: usize = DEFAULT_BUFFER_CAPACITY;
 
 const DEFAULT_DESTINATION_GET_SUFFIXES: &str = "svc.cluster.local.";
 const DEFAULT_DESTINATION_PROFILE_SUFFIXES: &str = "svc.cluster.local.";
-const DEFAULT_DESTINATION_PROFILE_INITIAL_TIMEOUT: Duration = Duration::from_millis(500);
+//const DEFAULT_DESTINATION_PROFILE_INITIAL_TIMEOUT: Duration = Duration::from_millis(500);
```
hawkw, line 232: i'm assuming that we want to add this back once we start bounding the time it takes to become ready?

linkerd/service-profiles/src/discover.rs
```diff
@@ -1,11 +1,22 @@
...
 use std::{future::Future, pin::Pin};
 
-pub fn layer<G: Clone, M>(
+pub fn layer<T, G: Clone, F: Clone, M: Clone>(
     get_profile: G,
-) -> impl layer::Layer<M, Service = Discover<G, M>> + Clone {
+    filter: F,
+) -> impl layer::Layer<
+    M,
+    Service = Discover<RecoverDefault<RequestFilter<F, GetProfileService<G>>>, M>,
```
hawkw, line 11: nit/tioli: seems like it could be time for a type alias or something?

hawkw (approved):
this all looks good to me, i like how the filtering logic was factored out.

linkerd/app/src/env.rs
olix0r: yeah, probably

linkerd/app/inbound/src/lib.rs
olix0r: i think it's fine either way. we'll need both in before the next release, but ordering doesn't really matter.

PR #696, opened 2020-10-07
Stop honoring DESTINATION_GET_* configuration

olix0r:
In a5e0b296, we extracted profile resolution so that each component
(inbound/outbound/gateway) makes its own filtering decisions. This
change does the same for endpoing resolution, moving the bulk of the
logic out of `app:dst` and into `app::outbound::resolve`.

The `app::dst` module is now only responsible for providing API clients
that transparently retry/reconnect.

Furthermore, the `...DESTINATION_GET_{SUFFIXES,NETWORKS}` environment
variables are no longer honored. Now, endpoints are resolved for all
addresses for which a profile lookup is successful.

The `Resolve` API has been cleaned up slightly: we no longer rely on the
`TryStream` alias (as it causes confusion + boilerplate), instead relying
on the more general `Stream`.


hawkw:
Looks like some tests aren't compiling?

linkerd/app/core/src/addr_match.rs
```diff
@@ -4,7 +4,10 @@ use linkerd2_dns::{Name, Suffix};
 use std::{net::IpAddr, sync::Arc};
 
 #[derive(Clone, Debug)]
-pub struct AddrMatch(NameMatch, IpMatch);
+pub struct AddrMatch {
+    names: NameMatch,
+    nets: IpMatch,
```
hawkw, line 9: <3 thanks

linkerd/app/outbound/src/resolve.rs
```diff
@@ -0,0 +1,118 @@
...
+    endpoint::FromMetadata: map_endpoint::MapEndpoint<T, Metadata>,
+    F: FilterRequest<T>,
+    R: Resolve<F::Request, Endpoint = Metadata>,
+    R::Future: Send + 'static,
+    R::Resolution: Send + 'static,
+{
+    map_endpoint::Resolve::new(
+        endpoint::FromMetadata,
+        RecoverDefault(RequestFilter::new(filter, resolve.into_service())),
+    )
```
hawkw, line 45: just a thought: it seems like, theoretically, we _could_ compose a lot of the resolver stuff using a stack rather than a bunch of nested constructors...maybe a future refactor?

linkerd/proxy/api-resolve/src/resolve.rs
```diff
@@ -20,7 +20,6 @@ use tracing::{debug, info, trace};
 #[derive(Clone)]
 pub struct Resolve<S> {
     service: DestinationClient<S>,
-    scheme: String,
```
hawkw, line 23: does the control plane mind that we're no longer sending schemes?

linkerd/proxy/core/src/resolve.rs
```diff
@@ -8,23 +8,23 @@ use std::task::{Context, Poll};
 pub trait Resolve<T> {
     type Endpoint;
     type Error: Into<Error>;
-    type Resolution: TryStream<Ok = Update<Self::Endpoint>, Error = Self::Error>;
+    type Resolution: Stream<Item = Result<Update<Self::Endpoint>, Self::Error>>;
```
hawkw, line 11: :+1: using `TryStream` in the trait signature was probably never a good idea...

hawkw (approved):
this looks good to me (i'm assuming that the change to not send schemes in `Destination.Get` requests is expected on the controller side). the refactoring all makes sense.

i wonder if we want to rename the `PROFILE_GET_xxx` env variables, since they now configure both profile and destination discovery? but, that would require changing the proxy injector etc as well, so we should probably punt on that for the immediate future.

linkerd/proxy/api-resolve/src/resolve.rs
olix0r: We've been sending empty strings for a long time

hawkw: oh, i see, this just removes it from the struct. cool, carry on!

linkerd/app/outbound/src/resolve.rs
olix0r: I think stacks are preferable when we have to compose many orthogonal concerns. In this case, I think it's simpler/clearer to just construct the services manually.

olix0r:
> i wonder if we want to rename the PROFILE_GET_xxx env variables, since they now configure both profile and destination discovery? but, that would require changing the proxy injector etc as well, so we should probably punt on that for the immediate future.

Punt for now

PR #698, opened 2020-10-08
outbound: Discover profiles for each unique TCP target

olix0r:
The proxy has classically discovered profiles & endpoints by looking at
each HTTP request's headers. This change removes all per-request
discovery. Instead, resolution is now done for each unique outbound TCP
target (regardless of whether the connection serves HTTP or not). This
eager resolution eliminates per-request cache binding; and supports
using `TrafficSplit` with non-HTTP services.

This has a few side effects:

- The `l5d-dst-override` header is no longer honored.
- When the application attempts to connect to a pod IP, the proxy no
  longer load balances these requests among all pods in the service.
  The proxy will now honor session-stickiness as selected by an
  application-level load balancer (see linkerd/linkerd2#4956).
- TrafficSplits are only applied when a client targets a service's IP.
- The proxy no longer performs DNS "canonicalization" to translate
  relative host header names to a fully-qualified form.


olix0r:
Depends on #697

linkerd/proxy/http/src/detect.rs
```diff
@@ -92,6 +92,7 @@ where
     H: NewService<(HttpVersion, T)>,
 {
     pub fn new(target: T, server: Server, new_http: H, new_tcp: F, drain: drain::Watch) -> Self {
+        tracing::trace!("new AcceptHttp");
```
hawkw, line 95: we _probably_ don't need this, i stuck it in while debugging and forgot to take it out
```suggestion
```

linkerd/proxy/http/src/detect.rs
```diff
@@ -136,8 +137,10 @@ where
             Some(HttpVersion::Http1) => {
                 debug!("Handling as HTTP");
                 let http1 = if let Some(svc) = self.http1.clone() {
+                    trace!("HTTP service already exists");
```
hawkw, line 140: these were also from debugging but i thought they might actually be worth having.

linkerd/app/outbound/src/endpoint.rs
```diff
@@ -53,6 +53,12 @@ pub struct TcpLogical {
     pub profile: Option<profiles::Receiver>,
 }
 
+#[derive(Clone, Debug)]
```
hawkw, line 56: nit: while working on the test changes i realized that it might not be great to derive `Debug` for `TcpLogical`, since it has a MPSC receiver in it, which dumps a bunch of internal state that we don't care about...we might want a custom `Debug` impl for `TcpLogical` because of that.

not a blocker, can fix it later; just something i remembered while looking at this.

hawkw (approved):
i was surprised by how simple the actual functional change is; most of this is actually just the test changes. the change itself is very straightforward! this looks good to me :+1:

PR #704, opened 2020-10-10
outbound: Discover profiles for each unique TCP target

olix0r:
The proxy has classically discovered profiles & endpoints by looking at
each HTTP request's headers. This change removes all per-request
discovery. Instead, resolution is now done for each unique outbound TCP
target (regardless of whether the connection serves HTTP or not). This
eager resolution eliminates per-request cache binding; and supports
using `TrafficSplit` with non-HTTP services.

This has a few side effects:

- The `l5d-dst-override` header is no longer honored.
- When the application attempts to connect to a pod IP, the proxy no
  longer load balances these requests among all pods in the service.
  The proxy will now honor session-stickiness as selected by an
  application-level load balancer (see linkerd/linkerd2#4956).
- `TrafficSplits` are only applied when a client targets a service's IP.
- The proxy no longer performs DNS "canonicalization" to translate
  relative host header names to a fully-qualified form.

Co-authored-by: Eliza Weisman <eliza@buoyant.io>


olix0r:
#698 was based against #697, so when we merged that branch it did not merge into main. This PR includes all of the needed changes.

linkerd/app/gateway/src/make.rs
```diff
@@ -1,132 +1,62 @@
...
-        let res = match self {
-            Self::Enabled { resolve, .. } => ready!(resolve.poll_ready(cx)),
-            _ => Ok(()),
-        };
-        Poll::Ready(res.map_err(Into::into))
-    }
+    type Service = Gateway<O::Service>;
 
-    fn call(&mut self, target: inbound::Target) -> Self::Future {
+    fn new_service(&mut self, (profile, target): Target) -> Self::Service {
```
hawkw, line 32: this looks much nicer as a new_service! <3

linkerd/app/outbound/src/endpoint.rs
```diff
@@ -71,49 +67,152 @@ pub struct TcpEndpoint {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("TcpLogical")
+            .field("addr", &self.addr)
+            .field(
+                "profile",
+                if self.profile.is_some() {
+                    &"Some(..)"
+                } else {
+                    &"None"
+                },
```
hawkw, line 125: _very_ minor and probably unimportant nit: since the second argument to `DebugStruct::field` is a `&dyn fmt::Debug`, these strings will be formatted with their `Debug` impl, which includes quotes. we might want to wrap them in `format_args!` or something to format them without quotes
```suggestion
                &format_args!("{}", if self.profile.is_some() {
                    "Some(..)"
                } else {
                    "None"
                })
```

linkerd/app/outbound/src/endpoint.rs
```diff
@@ -71,49 +67,152 @@ pub struct TcpEndpoint {
...
+        f.debug_struct("HttpLogical")
+            .field("version", &self.version)
+            .field("orig_dst", &self.orig_dst)
+            .field(
+                "profile",
+                if self.profile.is_some() {
+                    &"Some(..)"
+                } else {
+                    &"None"
+                },
```
hawkw, line 206: as i mentioned above, maybe consider
```suggestion
                &format_args!("{}", if self.profile.is_some() {
                    "Some(..)"
                } else {
                    "None"
                }),
```

linkerd/app/outbound/src/lib.rs
```diff
@@ -101,67 +100,25 @@ impl Config {
...
-            + Send
-            + 'static,
-        E::Future: Unpin + Send,
-        E::Resolution: Unpin + Send,
+        R: Resolve<Addr, Endpoint = Metadata, Error = Error> + Unpin + Clone + Send + 'static,
+        R::Future: Unpin + Send,
+        R::Resolution: Unpin + Send,
         I: tokio::io::AsyncRead + tokio::io::AsyncWrite + std::fmt::Debug + Unpin + Send + 'static,
     {
-        let watchdog = self.proxy.cache_max_idle_age * 2;
```
hawkw, line 113: nit/tioli: seems like we could just stick a `fn watchdog_timeout` on `Config` or on `Proxy`?

linkerd/app/outbound/src/lib.rs
```diff
@@ -349,39 +286,23 @@ impl Config {
...
             .push(http::header_from_target::layer(CANONICAL_DST_HEADER))
-            // Strips headers that may be set by this proxy.
-            .push_on_response(http::strip_header::request::layer(DST_OVERRIDE_HEADER))
-            .check_make_service_clone::<HttpLogical, http::Request<B>>()
+            .push_on_response(
+                svc::layers()
+                    // Strips headers that may be set by this proxy.
+                    .push(http::strip_header::request::layer(DST_OVERRIDE_HEADER))
+                    .push(svc::layers().box_http_response()),
+            )
```
hawkw, line 296: since we no longer honor `DST_OVERRIDE`, does this still need to be here at all? or can we just stop caring about it?

linkerd/proxy/http/src/detect.rs
```diff
@@ -92,6 +92,7 @@ where
     H: NewService<(HttpVersion, T)>,
 {
     pub fn new(target: T, server: Server, new_http: H, new_tcp: F, drain: drain::Watch) -> Self {
+        tracing::trace!("new AcceptHttp");
```
hawkw, line 95: nit/TIOLI: i put this in here while debugging, but i'm not sure if we really need it any more...
```suggestion
```

hawkw (approved):
Overall, this looks great, although I basically just clicked "viewed" on all the test changes I made --- we may want another set of eyes on that part.

Because we're no longer relying on DNS name canonicalization, I think this closes linkerd/linkerd2#4660 and linkerd/linkerd2#2706. :)

linkerd/app/outbound/src/lib.rs
olix0r: We should continue to strip the header through at least the stable release so that we're not all-of-a-sudden passing through the header if it's set by config.

linkerd/app/outbound/src/lib.rs
olix0r: will address this in a followup cleanup

PR #705, opened 2020-10-11
cache: Delete benchmarks

olix0r:
The `cache` module has benchmarks, but they are not tested in CI and we
don't really care about results -- the cache is basically just a
HashMap. Let's just delete this unused benchmark.


hawkw (approved):
looks good to me — the build errors in rust-analyzer were getting annoying, and i think the cache implementation has gotten significantly simpler since these were written.

PR #706, opened 2020-10-12
metrics: Coerce targets to metric labels by-reference

olix0r:
The `TransportLabels` trait & types serve no real purpose and can be
better-expressed via `Into`.

This change replaces the `TransportLabels` trait with an `Into`
constraint.  Furthermore, HTTP metrics have been updated to coerce the
target by-reference, without cloning the whole target.

### DIFF
--- a/linkerd/app/core/src/metric_labels.rs
+++ b/linkerd/app/core/src/metric_labels.rs
@@ -50,8 +50,8 @@ struct Authority<'a>(&'a http::uri::Authority);
 
 // === impl CtlLabels ===
 
-impl<'c> From<&'c control::ControlAddr> for ControlLabels {
-    fn from(c: &'c control::ControlAddr) -> Self {
+impl From<&'_ control::ControlAddr> for ControlLabels {
+    fn from(c: &'_ control::ControlAddr) -> Self {
         ControlLabels {
             addr: c.addr.clone(),
             tls_status: c.identity.clone().map(|id| TlsId::ServerId(id)).into(),
@@ -70,8 +70,8 @@ impl FmtLabels for ControlLabels {
 
 // === impl RouteLabels ===
 
-impl<'t> From<&'t dst::Route> for RouteLabels {
-    fn from(r: &'t dst::Route) -> Self {
+impl From<&'_ dst::Route> for RouteLabels {
+    fn from(r: &'_ dst::Route) -> Self {
         Self {
             target: r.target.clone(),
             direction: r.direction,
--- a/linkerd/app/core/src/metric_labels.rs
+++ b/linkerd/app/core/src/metric_labels.rs
@@ -50,11 +50,11 @@ struct Authority<'a>(&'a http::uri::Authority);
 
 // === impl CtlLabels ===
 
-impl From<control::ControlAddr> for ControlLabels {
-    fn from(c: control::ControlAddr) -> Self {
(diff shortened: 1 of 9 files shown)


linkerd/app/core/src/metric_labels.rs
```diff
@@ -50,11 +50,11 @@ struct Authority<'a>(&'a http::uri::Authority);
 
 // === impl CtlLabels ===
 
-impl From<control::ControlAddr> for ControlLabels {
-    fn from(c: control::ControlAddr) -> Self {
+impl<'c> From<&'c control::ControlAddr> for ControlLabels {
+    fn from(c: &'c control::ControlAddr) -> Self {
```
hawkw, line 54: nit/TIOLI: we could also use `'_` here if you like:
```suggestion
impl From<&'_ control::ControlAddr> for ControlLabels {
    fn from(c: &control::ControlAddr) -> Self {
```

linkerd/app/core/src/metric_labels.rs
```diff
@@ -70,12 +70,12 @@ impl FmtLabels for ControlLabels {
 
 // === impl RouteLabels ===
 
-impl From<dst::Route> for RouteLabels {
-    fn from(r: dst::Route) -> Self {
+impl<'t> From<&'t dst::Route> for RouteLabels {
+    fn from(r: &'t dst::Route) -> Self {
```
hawkw, line 74: ...and here
```suggestion
impl From<&'_ dst::Route> for RouteLabels {
    fn from(r: &dst::Route) -> Self {
```

linkerd/app/inbound/src/endpoint.rs
```diff
@@ -37,12 +43,44 @@ pub struct TcpEndpoint {
...
+    fn into(self) -> SocketAddr {
+        self.target_addr
+    }
+}
+
+impl Into<transport::labels::Key> for &'_ TcpAccept {
+    fn into(self) -> transport::labels::Key {
+        transport::labels::Key::accept(transport::labels::Direction::In, self.peer_id.clone())
+    }
+}
```
hawkw, line 82: elsewhere i've noticed comments explaining where the various `From`/`Into` impls get used...it might be nice to have those here?

hawkw: ...although, i suppose that when it's labels, it's pretty obvious what it's for...

linkerd/app/inbound/src/endpoint.rs
```diff
@@ -78,20 +116,23 @@ impl tls::HasPeerIdentity for HttpEndpoint {
...
-            port: addrs.target_addr().port(),
-        }
+impl Into<transport::labels::Key> for &'_ HttpEndpoint {
+    fn into(self) -> transport::labels::Key {
+        transport::labels::Key::Connect(transport::labels::EndpointLabels {
+            direction: transport::labels::Direction::In,
+            authority: None,
+            labels: None,
+            tls_id: tls::Conditional::None(tls::ReasonForNoPeerName::Loopback.into()).into(),
+        })
```
hawkw, line 126: ~~nit/tioli: maybe there should be a `Default` impl for `EndpointLabels` so we don't have to write out the `None`s?~~

actually, hmm, that might be a bit weird since it would mean picking a `Direction` to default to, and then always clobber it...never mind, maybe this isn't great?

hawkw (approved):
:train::link::arrow_right::label:

PR #707, opened 2020-10-12
outbound: Unify TCP & HTTP target types

olix0r:
We have target types like `{Http,Tcp}{Logical,Concrete,Endpoint}` that
differ only superficially between the HTTP and TCP variants. This
results in unnecessary, duplicate trait impls.

This change consolidates these targets as `Logical`, `Concrete`, and
`Endpoint`, all of which are generic over the protocol type. We provide
`Tcp*` and `Http*` trait aliases as a shorthand and compatibility.

### DIFF
--- a/linkerd/app/outbound/src/lib.rs
+++ b/linkerd/app/outbound/src/lib.rs
@@ -5,7 +5,7 @@
 
 #![deny(warnings, rust_2018_idioms)]
 
-use self::allow_discovery::{AllowHttpResolve, AllowProfile, AllowTcpResolve};
+use self::allow_discovery::{AllowProfile, AllowResolve};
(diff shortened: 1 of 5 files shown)


linkerd/app/outbound/src/lib.rs
```diff
@@ -178,7 +167,13 @@ impl Config {
...
             }))
             .check_new::<HttpEndpoint>()
-            .push(observability.clone())
+            .push(tap_layer.clone())
+            .push(metrics.http_endpoint.into_layer::<classify::Response>())
+            .push_on_response(TraceContextLayer::new(
+                span_sink
+                    .clone()
+                    .map(|sink| SpanConverter::client(sink, trace_labels())),
+            ))
```
olix0r, line 176: moved so it's easier to add check_new etc

hawkw (approved):
lovely!

PR #708, opened 2020-10-13
outbound: Use profile to inform protocol detection

olix0r:
The outbound proxy skips protocol detection when the target port is in a
configured list. But there's nothing that actually sets this
configuration currently. Now that we lookup profiles before protocol
detection, we can use the profile response to provide a more granular
per-endpoint hint.

This change stops honoring the
`LINKERD2_PROXY_OUTBOUND_PORTS_DISABLE_PROTOCOL_DETECTION` environment
variable. Service profiles are now used exclusively to determine whether
protocol detection should be skipped for outbound connections.

### DIFF
--- a/linkerd/app/core/src/config.rs
+++ b/linkerd/app/core/src/config.rs
@@ -23,7 +23,6 @@ pub struct ProxyConfig {
     pub connect: ConnectConfig,
     pub buffer_capacity: usize,
     pub cache_max_idle_age: Duration,
-    pub disable_protocol_detection_for_ports: crate::SkipByPort,
     pub dispatch_timeout: Duration,
     pub max_in_flight_requests: usize,
     pub detect_protocol_timeout: Duration,
--- a/linkerd/app/core/src/lib.rs
+++ b/linkerd/app/core/src/lib.rs
@@ -113,21 +113,3 @@ pub struct ProxyMetrics {
     pub stack: StackMetrics,
     pub transport: transport::Metrics,
 }
-
-#[derive(Clone, Debug, Default)]
-pub struct SkipByPort(std::sync::Arc<indexmap::IndexSet<u16>>);
-
-impl From<indexmap::IndexSet<u16>> for SkipByPort {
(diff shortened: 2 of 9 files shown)


kleimkuhler (approved):
Looks good 👍

hawkw:
IIRC, at one point we discussed that implementing opaque ports would require the outbound proxy to send some kind of message to the inbound proxy indicating that the connection is an opaque protocol. Is that still the case, and does this branch implement that?

hawkw:
> This change stops honoring the
> `LINKERD2_PROXY_OUTBOUND_PORTS_DISABLE_PROTOCOL_DETECTION` environment
> variable. Service profiles are now used exclusively to determine whether
> protocol detection should be skipped for outbound connections.

This implies that we probably want to hold off on releasing this change until after the control plane actually starts sending the opaque port field? Otherwise, releasing a proxy with this change will break server-first protocols since there won't be any way to configure them.

hawkw (approved):
lgtm modulo a couple of questions in comments i already left

olix0r:
> IIRC, at one point we discussed that implementing opaque ports would require the outbound proxy to send some kind of message to the inbound proxy indicating that the connection is an opaque protocol. Is that still the case, and does this branch implement that?

This is necessary to implement mTLS with opaque ports, since the inbound port will skip TLS detection. This is not necessary to just forward the request to the original dst.

> This implies that we probably want to hold off on releasing this change until after the control plane actually starts sending the opaque port field? Otherwise, releasing a proxy with this change will break server-first protocols since there won't be any way to configure them.

Note:

> > The outbound proxy skips protocol detection when the target port is in a configured list. But there's nothing that actually sets this configuration currently. 

There is no way to configure `LINKERD2_PROXY_OUTBOUND_PORTS_DISABLE_PROTOCOL_DETECTION` -- all outbound port skipping works at the iptables layer, currently. So there's no need to gate this change on control plane changes. We'll continue to treat all of these as detectable by default until the control plane starts setting this field on profile responses.

PR #709, opened 2020-10-13
app: Consolidate metrics types in `core::metrics`

olix0r:
`app::core` spreads its metrics-related types over three modules:
`metrics`, `metric_labels` and the base, `app::core`. This change
consolidates all of these types in `app::core::metrics` to simplify
future changes to metrics formatting.

Also, we were still exporting handle_time metrics even though we do
not record them.


hawkw (approved):
seems much less messy!

PR #710, opened 2020-10-13
trace-context: Simplify implementation with async

olix0r:
TraceContext relies on some older idioms like a manual future
implementaton.

This change replaces the `TraceContextLayer` type with a
`TraceContext::layer` function and the `ResponseFuture` implementation
with an (optionally) boxed response future. The response future is boxed
only when tracing is active.

No functional changes/

### DIFF
--- a/linkerd/trace-context/src/layer.rs
+++ b/linkerd/trace-context/src/layer.rs
@@ -1,7 +1,6 @@
 use crate::{propagation, Span, SpanSink};
-use futures::{prelude::*, ready};
-use linkerd2_error::Error;
-use linkerd2_stack::{layer, Either};
+use futures::{future::Either, prelude::*};
+use linkerd2_stack::layer;
 use std::{
     collections::HashMap,
     future::Future,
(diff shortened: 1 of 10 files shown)


linkerd/trace-context/src/layer.rs
```diff
@@ -1,171 +1,127 @@
...
-    labels.insert("http.path".to_string(), path);
-    if let Some(authority) = req.uri().authority() {
-        labels.insert("http.authority".to_string(), authority.as_str().to_string());
-    }
-    if let Some(host) = req.headers().get("host") {
-        if let Ok(host) = host.to_str() {
-            labels.insert("http.host".to_string(), host.to_string());
-        }
+        // If there's no tracing to be done, just pass on the request to the inner service.
+        Either::Left(self.inner.call(req))
```
hawkw, line 125: :+1: for only boxing the future when actually doing the tracing stuff.

hawkw (approved):

PR #711, opened 2020-10-13
outbound: more TCP tests, test cleanup

hawkw:
This branch adds two new TCP tests: one asserting that resolutions are
cached across multiple connections with the same original destination
IP, and one asserting that outbound TCP load balancing actually
distributes load over multiple endpoints.

In addition, I've done some more cleanup and refactoring of the new
tests to reduce boilerplate, etc. This should make adding new tests in
subsequent PRs easier.


linkerd/app/outbound/src/tests/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+};
+use linkerd2_app_test as test_support;
+use std::{net::SocketAddr, str::FromStr, time::Duration};
+use tower::ServiceExt;
+
+mod tcp;
+
+const LOCALHOST: [u8; 4] = [127, 0, 0, 1];
+
+fn profile() -> profiles::Receiver {
```
olix0r, line 22: nit: `default_profile()`? since this doesn't give you any control over which profile is used

hawkw: this code was all already there, i just moved it from `tests.rs` into `tests/mod.rs`

PR #712, opened 2020-10-13
inbound: Fix source tap annotations

olix0r:
The inbound target type changed recently, but the tap inspector was not
updated to fetch the new target type from the request extensions.

This change adds a `peer_addr` to the `TcpAccept` target type and uses
this to satisfy the `src_addr` tap metadata.


kleimkuhler (approved):
👍

PR #713, opened 2020-10-13
inbound: Fix URI normalization for orig-proto requests

olix0r:
linkerd/linkerd2#5069 reports that proxied requests can incorrectly
end up in the absolute-form. This happens because URI normalization is
applied after protocol downgrade (where the absolute-form is annotated
explicitly).

This change improves the inbound orig-proto test to detect this bug and
fixes it by performing normalization before protocol downgrade so that
the downgrade layer's behavior is not overridden.


hawkw (approved):
looks good to me!

PR #715, opened 2020-10-14
gateway: Ensure proper outbound metadata

olix0r:
The gateway has two deficiencies:

* Using the outbound proxy port for the outbound original dst leads to
  the incorrect port being used in the `l5d-dst-canonical` header. Now,
  we instead use the proper port with an unroutable IP address.
* Because the source cluster's outbound proxy strips `Host` headers,
  there's no `Host` present on gatewayed HTTP/1 requests. Now, the
  `Host` header is updated on all gatewayed HTTP/1 requests.

### DIFF
--- a/linkerd/app/gateway/src/gateway.rs
+++ b/linkerd/app/gateway/src/gateway.rs
@@ -98,7 +98,7 @@ where
                 if let ::http::Version::HTTP_11 | ::http::Version::HTTP_10 = request.version() {
                     request
                         .headers_mut()
-                        .insert(::http::header::HOST, host_header.clone());
+                        .insert(http::header::HOST, host_header.clone());
                 }
 
                 tracing::debug!(
--- a/linkerd/app/gateway/src/gateway.rs
+++ b/linkerd/app/gateway/src/gateway.rs
@@ -12,8 +12,9 @@ pub(crate) enum Gateway<O> {
     BadDomain(dns::Name),
     Outbound {
         outbound: O,
-        forwarded_header: http::header::HeaderValue,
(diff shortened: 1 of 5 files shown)


linkerd/app/gateway/src/gateway.rs
```diff
@@ -86,6 +91,16 @@ where
                     .headers_mut()
                     .append(http::header::FORWARDED, forwarded_header.clone());
 
+                // If we're forwarding HTTP/1 requests, the old `Host` header
+                // was stripped on the peer's outbound proxy. But the request
+                // should have an updated `Host` header now that it's being
+                // routed in the cluster.
```
hawkw, line 97: thanks for documenting why this matters!

linkerd/app/gateway/src/gateway.rs
```diff
@@ -86,6 +91,16 @@ where
...
+                if let ::http::Version::HTTP_11 | ::http::Version::HTTP_10 = request.version() {
+                    request
+                        .headers_mut()
+                        .insert(::http::header::HOST, host_header.clone());
```
hawkw, line 101: nit: elsewhere in this file, we just import from `http::`. is the leading `::` necessary here? if not, i'd prefer to remove it:
```suggestion
                if let http::Version::HTTP_11 | http::Version::HTTP_10 = request.version() {
                    request
                        .headers_mut()
                        .insert(http::header::HOST, host_header.clone());
```

hawkw (approved):
looks good to me --- thanks for leaving comments for the next time i look at this and have forgotten all the context :)

PR #714, opened 2020-10-13
outbound: test profile search nets filtering

hawkw:
This PR adds a new test asserting that service profiles are only
resolved for original destination IPs within the specified search
networks. Also, I updated the existing tests so that the mock connectors
for endpoints which are expected to be plaintext actually assert that no
server identity was provided.

PR #717, opened 2020-10-14
outbound: test load balancer adding/removing TCP endpoints

hawkw:
Depends on #714

This branch adds new tests for updating TCP destinations by adding and
removing endpoints. The first test asserts that when new endpoints are
added to a load balancer, some connections after they are added are
routed to the new endpoint. The other test asserts that when endpoints
are removed, they no longer receive connections.

I've also done some more test cleanup & boilerplate removal.

### DIFF
--- a/linkerd/app/outbound/src/tests/tcp.rs
+++ b/linkerd/app/outbound/src/tests/tcp.rs
@@ -37,7 +37,10 @@ async fn plaintext_tcp() {
     let mut srv_io = test_support::io();
     srv_io.read(b"hello").write(b"world");
     // Build a mock "connector" that returns the upstream "server" IO.
-    let connect = test_support::connect().endpoint_builder(target_addr, srv_io);
+    let connect = test_support::connect().endpoint_fn(target_addr, move |endpoint: TcpEndpoint| {
+        assert!(endpoint.peer_identity().is_none());
+        Ok(srv_io.build())
+    });
 
     // Configure mock IO for the "client".
     let client_io = test_support::io().write(b"hello").read(b"world").build();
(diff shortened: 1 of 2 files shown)


linkerd/app/outbound/src/tests/tcp.rs
```diff
@@ -402,6 +599,48 @@ async fn no_profiles_when_outside_search_nets() {
...
+            identity: tls::Conditional::None(
+                tls::ReasonForNoPeerName::NotProvidedByServiceDiscovery,
+            ),
+            count: Arc::new(AtomicUsize::new(0)),
+            enabled: Arc::new(AtomicBool::new(true)),
+        }
+    }
+}
+
+impl Into<Box<dyn FnMut(TcpEndpoint) -> test_support::connect::ConnectFuture + Send + 'static>>
```
olix0r, line 620: tioli. This strikes me as pretty subtle. It might be a little clearer as an explicitly named fn like `into_connect`? Not a blocker

linkerd/app/test/src/connect.rs
```diff
@@ -46,15 +51,15 @@ impl<E: fmt::Debug> Connect<E> {
         }
     }
 
-    pub fn endpoint_future(
+    pub fn endpoint(
         self,
         endpoint: impl Into<SocketAddr>,
-        f: impl (FnMut(E) -> ConnectFuture) + Send + 'static,
+        f: impl Into<Box<dyn FnMut(E) -> ConnectFuture + Send + 'static>>,
```
olix0r, line 57: It's weird to me that the thing to me that's called `endpoint` is the socketaddr but the the thing that is an `Endpoint` is called `f` -- when I saw `endpoint.into()` i assumed it was converting it to a connect... Can we clarify naming?

linkerd/app/outbound/src/tests/tcp.rs
```diff
@@ -325,6 +311,217 @@ async fn load_balances() {
...
+    dst.add(Some((endpoints[2].0, meta.clone())))
+        .expect("still listening");
+
+    conns().await;
+    assert!(endpoints[0].1.load(Ordering::Acquire) >= 10);
+    assert_ne!(
+        endpoints[1].1.load(Ordering::Acquire),
+        0,
+        "no connections to endpoint 1"
+    );
```
olix0r, line 416: I don't think this can ever fail because of the prior assertion

olix0r (approved):
lgtm. would suggest taking a pass at disambiguating some of the naming, but not a hard blocker

hawkw:
@olix0r i gave the naming a pass, let me know what you think?

PR #718, opened 2020-10-15
outbound: Set source address in Tap metadata

olix0r:
We recently stopped exposing source addresses on outbound Tap metadata,
but this causes some confusion in the Tap UIs.

This change introduces an `io::PeerAddr` trait that is implemented by
all `Io` types and a `proxy::http:SetClientAddr` middleware that inserts
a `ClientAddr` type onto requests. This middleware is now applied by the
`DetectHttp` server to ensure the HTTP stack can access the source
address.

### DIFF
--- a/linkerd/app/core/src/admin/mod.rs
+++ b/linkerd/app/core/src/admin/mod.rs
@@ -4,6 +4,7 @@
 //! * `/ready` -- returns 200 when the proxy is ready to participate in meshed traffic.
 
 use crate::{
+    proxy::http::{ClientAddr, SetClientAddr},
     svc, trace,
     transport::{io, tls},
 };
@@ -17,7 +18,7 @@ use std::{
     pin::Pin,
     task::{Context, Poll},
 };
-use tower::{service_fn, util::ServiceExt, Service};
+use tower::Service;
 
 mod readiness;
 mod tasks;
@@ -39,9 +40,6 @@ pub struct Accept<M: FmtMetrics>(Admin<M>, hyper::server::conn::Http);
 
 #[derive(Debug, Clone)]
 pub struct Serve<M: FmtMetrics>(tls::accept::Meta, Accept<M>);
-
-#[derive(Clone, Debug)]
-pub struct ClientAddr(std::net::SocketAddr);
 
 pub type ResponseFuture =
     Pin<Box<dyn Future<Output = Result<Response<Body>, io::Error>> + Send + 'static>>;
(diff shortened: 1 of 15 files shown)


olix0r:
The other option would be to make it return Option or Result instead of erroring... i'm not really sure that's better though.

linkerd/app/core/src/admin/mod.rs
```diff
@@ -134,22 +132,12 @@ impl<M: FmtMetrics + Clone + Send + 'static> svc::Service<io::BoxedIo> for Serve
         // client's IP address, we wrap the service with a new service
         // that adds the remote IP as a request extension.
         let peer = meta.addrs.peer();
-        let svc = svc.clone();
-        let svc = service_fn(move |mut req: Request<Body>| {
-            req.extensions_mut().insert(ClientAddr(peer));
-            svc.clone().oneshot(req)
-        });
```
hawkw, line 141: <3 for remembering that we were also doing that here :)

linkerd/app/outbound/src/lib.rs
```diff
@@ -66,7 +66,7 @@ impl Config {
         HttpEndpoint,
         Error = Error,
         Future = impl future::Future + Send,
-        Response = impl tokio::io::AsyncRead + tokio::io::AsyncWrite + Unpin + Send + 'static,
+        Response = impl io::AsyncRead + io::AsyncWrite + Unpin + Send + 'static,
```
hawkw, line 69: nit/tioli: for brevity's sake, i wonder if we want to start exposing the `Io` trait or something, so we don't have to write all this out...not a blocker.

linkerd/app/outbound/src/endpoint.rs
```diff
@@ -258,8 +257,10 @@ impl<P> CanOverrideAuthority for Endpoint<P> {
 }
 
 impl tap::Inspect for HttpEndpoint {
-    fn src_addr<B>(&self, _: &http::Request<B>) -> Option<SocketAddr> {
-        None
+    fn src_addr<B>(&self, req: &http::Request<B>) -> Option<SocketAddr> {
+        req.extensions()
+            .get::<ClientAddr>()
+            .map(|c| c.as_ref().clone())
```
hawkw, line 263: nit: `SocketAddr` is `Copy`, so I think this could just be  
```suggestion
            .map(|c| c.as_ref().copied())
```

hawkw: or, maybe we should just have an `Into<SocketAddr>` impl for `&'_ ClientAddr`?

hawkw (approved):
this looks good to me!

PR #720, opened 2020-10-18
admin: Simplify metrics server

olix0r:
The metrics server has a lot of needless boilerplate logic. This change
simplifies the `Service` impl to a single, synchronous `fn serve`,
eliminates unneeded error types, and relaxes type constraints where
possible.


kleimkuhler (approved):
👍

hawkw (approved):

PR #721, opened 2020-10-18
Split tracing init & admin handlers into crate

olix0r:
The `app::core::trace` module is coupled to the handlers in
`app::core::admin`. This change splits this logic into a dedicated
crate, `linkerd2-tracing`.

The only functional change is that requests that do not originate from
localhost now receive 404s when trying to access the tracing admin
endpoints.

This change isolates tracing logic so that tracing overhead can disabled
entirely in performance-critical configurations.


linkerd/app/core/src/admin/mod.rs
```diff
@@ -158,23 +168,11 @@ fn rsp(status: StatusCode, body: impl Into<Body>) -> Response<Body> {
...
-    if let Some(addr) = req.extensions().get::<ClientAddr>() {
-        let addr = addr.as_ref();
-        if addr.ip().is_loopback() {
-            return Ok(());
-        }
-        tracing::warn!(%addr, "denying request from non-loopback IP");
-        Err(rsp(
-            StatusCode::FORBIDDEN,
-            "access to /proxy-log-level and /trace only allowed from loopback interface",
-        ))
```
hawkw, line 171: any reason we no longer return Forbidden? 404 seems much less semantically correct in this particular case. not a big deal though.

hawkw (approved):

PR #722, opened 2020-10-18
tracing: Fix time formatting to ensure alignment

olix0r:
In logs, we emit timestamps as time fractional seconds since the process
started. These timestamps are intended to be formatted at a fixed
length, but we don't properly truncate nanoseconds to fit the 6-digit
field we leave for subsecond time.

This change converts nanoseconds to microseconds when formatting
the uptime so that the value fits into six digits. It also adds a test
to validate this behavior.

### DIFF
--- a/linkerd/tracing/src/lib.rs
+++ b/linkerd/tracing/src/lib.rs
@@ -2,10 +2,11 @@
 
 mod level;
 mod tasks;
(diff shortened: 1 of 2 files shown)


hawkw (approved):
<3 this has been bugging me for ages

PR #723, opened 2020-10-18
tracing: Support disabling tracing entirely

olix0r:
When the proxy log level is set to `off`, the proxy now skips all
tracing initialization and disables the tracing admin endpoints. This
allows us to minimize the overhead of tracing in benchmarks and
performance-critical environments.

This has a noticeable impact on _memory_. In local tests, we see
`main` use around ~24MB with the log level set to `off`. With this
branch, memory comes down to ~16MB.

### DIFF
--- a/linkerd/tracing/src/lib.rs
+++ b/linkerd/tracing/src/lib.rs
@@ -32,8 +32,11 @@ type Formatter<F, E> = Layered<
 /// environment variable as the verbosity-level filter.
 pub fn init() -> Result<Handle, Error> {
     let log_level = env::var(ENV_LOG_LEVEL).unwrap_or(DEFAULT_LOG_LEVEL.to_string());
-    let log_format = env::var(ENV_LOG_FORMAT).unwrap_or(DEFAULT_LOG_FORMAT.to_string());
+    if let "OFF" = log_level.to_uppercase().trim() {
+        return Ok(Handle(Inner::Disabled));
+    }
 
+    let log_format = env::var(ENV_LOG_FORMAT).unwrap_or(DEFAULT_LOG_FORMAT.to_string());
     let (dispatch, handle) = with_filter_and_format(log_level, log_format);
 
     // Set up log compatibility.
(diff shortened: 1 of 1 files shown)


hawkw:
Is there a significant difference in performance between special-casing `LINKERD2_PROXY_LOG=OFF` and not setting a `tracing` subscriber, and just specifying the "off" tracing level, with `LINKERD2_PROXY_LOG=off` (bypassing this code entirely)?

This approach may still be better (it should use a bit less RAM, since no subscriber is ever constructed and the storage it requires won't be allocated), but the `off` level is pretty highly optimized and _should_ have similar performance to no subscriber at all. I'm asking about this less because I think this is the wrong way to do it, and more because if there's a major difference in performance, there could potentially be an upstream issue preventing those optimizations from working correctly.

Also, just using an "off" level filter would permit disabling all tracing by default and then turning it on later via the tracing admin endpoint.

olix0r:
@hawkw 

> Is there a significant difference in performance between special-casing LINKERD2_PROXY_LOG=OFF and not setting a tracing subscriber, and just specifying the "off" tracing level, with LINKERD2_PROXY_LOG=off (bypassing this code entirely)?

Latency is more or less the same between branches when logging is set to `off`; but, in the local test I ran, the proxy uses about %33 less memory with this branch.

> Also, just using an "off" level filter would permit disabling all tracing by default and then turning it on later via the tracing admin endpoint. This might even be a good default configuration for running in prod, especially if we make it easier to dynamically enable tracing via the CLI.

Yeah, though I think it's important to log errors and warnings by default.

We could change the special value from `off` to `disabled` or something, but I think I'm okay with having `off` be special-cased (and immutable).

hawkw (approved):
okay, i'm fine with continuing to special-case `off` for now. in future `tracing` releases, the memory use should be significantly reduced when everything is disabled, so we might want to revisit this, and allow dynamically toggling the log level to and from `off`.

PR #724, opened 2020-10-22
router: Use NewService instead of MakeService

olix0r:
There's no reason for `NewRouter` to operate over `MakeService`. This
changes `NewRouter` to build the inner service synchronously, avoiding
the need for a `ResponseFuture` that joins `MakeService` and the inner
response future.


hawkw (approved):
💕

PR #725, opened 2020-10-22
outbound: Split TCP stack into dedicated modules

olix0r:
The main `outbound/src/lib.rs` is getting large and unwieldy. Before
making additional changes, this reorganizes outbound as follows:

* `endpoint` has been renamed to `target`. It includes target types,
  one of which is `Endpoint`.
* A new `tcp` module includes TCP-specific target aliases.
* The TCP connect stack is built in `tcp::connect`. The `PreventLoop`
  filter is now a private type within this module.
* The TCP balance stack is built in `tcp::balance`.
* The `resolve` module is no longer generic over its target types, now
  relying on the `Concrete` target type.
* `allow_discovery::AllowProfile` is now `tcp::AllowProfile`, as it is
  specific to the `tcp::Accept` target type.
* The `tcp::connect` stack is now generic over the endpoint's protocol
  (instead of hardcoding implementations for TcpEndpoint and
  HttpEndpoint). This means that separate concrete connect stacks must
  be used for each protocol.
* TCP tests have been moved under the `tcp` module.

There are no functional changes. Similar changes will follow for the HTTP
stacks.


linkerd/app/outbound/src/resolve.rs
```diff
@@ -24,59 +24,69 @@ use std::{
...
+    type Request = Addr;
+
+    fn filter(&self, target: Concrete<P>) -> Result<Addr, Error> {
+        target.resolve.ok_or_else(|| discovery_rejected().into())
+    }
+}
+
 // === impl RecoverDefault ===
 
 type Resolution<R> =
```
hawkw, line 92: very minor style nit: i have a slight preference for having the `impl` blocks in the same order as the types are defined, not a big deal.

linkerd/app/outbound/src/target.rs
```diff
@@ -79,22 +90,31 @@ impl Into<transport::labels::Key> for &'_ Accept {
...
-    fn from((protocol, l): (http::Version, TcpLogical)) -> Self {
+impl<P> From<(http::Version, Logical<P>)> for HttpLogical {
+    fn from(
+        (
+            protocol,
+            Logical {
+                orig_dst, profile, ..
+            },
+        ): (http::Version, Logical<P>),
+    ) -> Self {
```
hawkw, line 113: wow...

linkerd/app/outbound/src/tcp/connect.rs
```diff
@@ -0,0 +1,71 @@
...
+        // Initiates mTLS if the target is configured with identity.
+        .push(tls::client::ConnectLayer::new(local_identity))
+        // Limits the time we wait for a connection to be established.
+        .push_timeout(config.timeout)
+        .push(metrics.transport.layer_connect())
+        .push_request_filter(PreventLoop { port: server_port })
+        .into_inner()
+}
+
+/// A connection policy that drops
```
hawkw, line 31: drops what?

linkerd/app/outbound/src/test_util.rs
```diff
@@ -1,31 +1,21 @@
...
 use linkerd2_app_core::{
-    config, exp_backoff, profiles,
-    proxy::http::h2,
-    svc::NewService,
-    transport::{listen, tls},
-    Error, IpMatch,
+    config, exp_backoff, profiles, proxy::http::h2, transport::listen, IpMatch,
 };
-use linkerd2_app_test as test_support;
+pub use linkerd2_app_test as support;
```
hawkw, line 7: nit/TIOLI: might be slightly nicer to just
```suggestion
pub use linkerd2_app_test::*;
```
 and then have the tests just use _this_ module as `support`? not a blocker.

hawkw (approved):
looks good to me; couple of minor nits i don't really care about

linkerd/app/outbound/src/test_util.rs
olix0r: maybe in a followup. glob imports end up being pretty hairy to understand, and in this case it's easy for unintentional overlaps like `io() vs transport::io`

linkerd/app/outbound/src/resolve.rs
olix0r: aren't they already? what needs to change?

hawkw: oh you know what? never mind. i can't read.

PR #728, opened 2020-10-23
outbound: Introduce 'ingress mode'

olix0r:
When the `LINKERD2_PROXY_INGRESS_MODE` environment variable is set to
`true`, the outbound proxy alters its discovery behavior as follows:

* All outbound traffic is assumed to be HTTP traffic or unmeshed opaque
  TCP traffic. Discovery, routing, balancing, and mTLS is only applied
  for outbound THTTP traffic.
* The `l5d-dst-override` request header is used to specify the target
  authority.
* If the header is not present, the original dst address is used.
* Service profiles, traffic split, and balancing are only applied when
  the header is present or when the original dst addr refers to a service
  address (and not a pod address).


linkerd/app/outbound/src/ingress.rs
```diff
@@ -0,0 +1,204 @@
...
+};
+use tokio::sync::mpsc;
+use tracing::info_span;
+
+/// Routes HTTP requests according to the l5d-dst-override header.
+///
+/// Forwards TCP connections without discovery/routing (or mTLS).
+///
+/// This is only intended for Ingress configurations, where we assume all
+/// outbound traffic is either
```
hawkw, line 20: either what?

linkerd/app/src/lib.rs
```diff
@@ -147,29 +153,54 @@ impl Config {
...
-                    outbound_listen,
-                    outbound::server::stack(
-                        &outbound,
-                        dst.profiles.clone(),
-                        dst.resolve,
-                        outbound::tcp::connect::stack(
-                            &outbound.proxy.connect,
-                            outbound_addr.port(),
-                            local_identity.clone(),
+            if ingress_mode {
```
hawkw, line 156: nit: maybe worth an `info` log here indicating whether or not we're in ingress mode?

linkerd/app/outbound/src/ingress.rs
```diff
@@ -0,0 +1,204 @@
...
+                dispatch_timeout,
+                max_in_flight_requests,
+                detect_protocol_timeout,
+                buffer_capacity,
+                cache_max_idle_age,
+                ..
+            },
+    } = config.clone();
+
+    let http = svc::stack(http)
```
hawkw, line 77: I haven't looked closely yet but it feels like there's _probably_ some common code between the ingress-mode and normal mode http stacks that we might want to share. nbd if not, but adding more places that future changes will have to update always worries me a little bit...

olix0r: I agree, but deduping things is easier said than done. If you can get it working, I'll merge it ;)

hawkw: i'm not sure if it makes sense here or not, but i can take a look.

linkerd/app/src/lib.rs
```diff
@@ -146,30 +152,55 @@ impl Config {
...
+                            outbound_http.clone(),
+                            outbound_metrics,
+                            oc_span_sink.clone(),
+                            drain_rx.clone(),
+                        ),
+                        drain_rx.clone().signal(),
+                    )
+                    .map_err(|e| panic!("outbound failed: {}", e))
+                    .instrument(span.clone()),
+                );
```
hawkw, line 202: nit: spent a bit trying to figure out how to reduce repetition here; turns out it's not super easy. but, I think the TCP connect stack could at least be factored out?
```rust
            info!(listen.addr = %outbound_addr, ingress_mode);
            let tcp_connect = outbound::tcp::connect::stack(
                &outbound.proxy.connect,
                outbound_addr.port(),
                local_identity.clone(),
                &outbound_metrics,
            );
            if ingress_mode {
                tokio::spawn(
                    serve::serve(
                        outbound_listen,
                        outbound::ingress::stack(
                            &outbound,
                            dst.profiles.clone(),
                            outbound::tcp::connect::forward(tcp_connect),
                            outbound_http.clone(),
                            &outbound_metrics,
                            oc_span_sink.clone(),
                            drain_rx.clone(),
                        ),
                        drain_rx.clone().signal(),
                    )
                    .map_err(|e| panic!("outbound failed: {}", e))
                    .instrument(span.clone()),
                );
            } else {
                tokio::spawn(
                    serve::serve(
                        outbound_listen,
                        outbound::server::stack(
                            &outbound,
                            dst.profiles.clone(),
                            dst.resolve,
                            tcp_connect,
                            outbound_http.clone(),
                            outbound_metrics,
                            oc_span_sink.clone(),
                            drain_rx.clone(),
                        ),
                        drain_rx.clone().signal(),
                    )
                    .map_err(|e| panic!("outbound failed: {}", e))
                    .instrument(span.clone()),
                );
            }
```

linkerd/app/outbound/src/ingress.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+}
+
+impl<B> svc::stack::RecognizeRoute<http::Request<B>> for TargetPerRequest {
+    type Key = Target;
+
+    fn recognize(&self, req: &http::Request<B>) -> Self::Key {
+        Target {
+            accept: self.0,
+            dst: http_request_l5d_override_dst_addr(req).unwrap_or_else(|_| self.0.orig_dst.into()),
```
hawkw, line 201: nit/TIOLI: it might be worth having a debug event indicating whether or not we saw the override header, or something, since this is kinda load-bearing?

hawkw (approved):
this seems good to me! 

i'd like to figure out if we could factor out some of the repeated code between the ingress and non-ingress stacks but it's definitely not a blocker, and we can come back and clean it up later. other than that, just some minor style nits you can feel free to ignore.

it would also be good to have some tests for the ingress mode behavior, as well, but that can also land in a follow-up.

linkerd/app/outbound/src/ingress.rs
olix0r: It should be pretty obvious based on the resulting target, which is ends up in span metadata

linkerd/app/src/lib.rs
```diff
@@ -40,6 +40,11 @@ pub struct Config {
     pub inbound: inbound::Config,
     pub gateway: gateway::Config,
 
+    // In "ingress mode", we assume we are always routing HTTP requests and do
+    // not perform per-target-address discovery. Non-HTTP connections are
+    // forwarded without discovery/routing/mTLS.
+    pub ingress_mode: bool,
```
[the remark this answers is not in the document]
olix0r: Yeah, this was put here intentionally because we may opt to change inbound behavior in the future; and, more importantly, nothing in the outbound crate needs to be aware of this flag currently.

PR #730, opened 2020-10-26
Reduce tracing spans to the debug level

olix0r:
We currently set almost all span metadata at the info level, meaning
that these spans are recorded/maintained for all connections/requests
even though there are virtually no logs emitted with this metadata at
the info level.

This changes all of these spans to only be recorded at the debug level.
The one exception is that the per-connection `accept` span is maintained
at the info level so that errors are reported with at least
connection-level metadata.

### DIFF
--- a/linkerd/app/gateway/src/config.rs
+++ b/linkerd/app/gateway/src/config.rs
@@ -4,7 +4,7 @@ use linkerd2_app_core::{
 };
 use linkerd2_app_inbound::endpoint as inbound;
 use linkerd2_app_outbound as outbound;
-use tracing::info_span;
+use tracing::debug_span;
 
 #[derive(Clone, Debug, Default)]
 pub struct Config {
@@ -51,7 +51,7 @@ impl Config {
                 Allow(self.allow_discovery),
             ))
             .check_new_service::<inbound::Target, http::Request<http::boxed::Payload>>()
-            .instrument(|_: &inbound::Target| info_span!("gateway"))
+            .instrument(|_: &inbound::Target| debug_span!("gateway"))
             .into_inner()
     }
 }
--- a/linkerd/app/outbound/src/tcp/tests.rs
+++ b/linkerd/app/outbound/src/tcp/tests.rs
@@ -296,7 +296,7 @@ async fn load_balances() {
         .map(|i| {
             tokio::spawn(
                 hello_world_client(svc_addr, &mut server)
-                    .instrument(tracing::debug_span!("conn", i)),
+                    .instrument(tracing::info_span!("conn", i)),
             )
             .err_into::<Error>()
         })
@@ -385,7 +385,7 @@ async fn load_balancer_add_endpoints() {
             .map(|i| {
                 tokio::spawn(
                     hello_world_client(svc_addr, &mut server)
-                        .instrument(tracing::debug_span!("conn", i)),
+                        .instrument(tracing::info_span!("conn", i)),
                 )
                 .err_into::<Error>()
             })
@@ -494,7 +494,7 @@ async fn load_balancer_remove_endpoints() {
             .map(|i| {
                 tokio::spawn(
                     hello_world_client(svc_addr, &mut server)
-                        .instrument(tracing::debug_span!("conn", i)),
+                        .instrument(tracing::info_span!("conn", i)),
                 )
                 .err_into::<Error>()
             })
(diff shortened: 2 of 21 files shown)


linkerd/app/outbound/src/tcp/tests.rs
```diff
@@ -296,7 +296,7 @@ async fn load_balances() {
         .map(|i| {
             tokio::spawn(
                 hello_world_client(svc_addr, &mut server)
-                    .instrument(tracing::info_span!("conn", i)),
+                    .instrument(tracing::debug_span!("conn", i)),
```
hawkw, line 299: IMO it's nice to have the spans in these tests on by default, was the change to the tests intended, or was it just an accident of find-and-replace?

linkerd/app/src/lib.rs
```diff
@@ -86,31 +86,32 @@ impl Config {
...
-            info_span!("opencensus")
+            debug_span!("opencensus")
                 .in_scope(|| oc_collector.build(identity, dns, metrics, client_metrics))
         }?;
 
         let admin = {
             let identity = identity.local();
             let drain = drain_rx.clone();
-            info_span!("admin").in_scope(move || admin.build(identity, report, log_level, drain))?
+            debug_span!("admin")
```
hawkw, line 119: seems potentially worth keeping these top-level spans at INFO? i think it's mainly the ones in the actual data path where there's a noticeable latency impact?

linkerd/app/src/lib.rs
```diff
@@ -144,7 +145,7 @@ impl Config {
                 outbound_metrics.clone(),
             );
 
-            let span = info_span!("outbound");
+            let span = debug_span!("outbound");
```
hawkw, line 154: similarly, I think it's worth keeping the outbound and inbound spans at INFO for errors/warnings, as well as the accept span?

linkerd/app/test/src/connect.rs
```diff
@@ -32,7 +32,7 @@ where
 
     fn call(&mut self, endpoint: E) -> Self::Future {
         let addr = endpoint.clone().into();
-        let span = tracing::info_span!("connect", %addr);
+        let span = tracing::debug_span!("connect", %addr);
```
hawkw, line 35: this is a test, and i think it's useful to have these traces in the test on by default

linkerd/service-profiles/src/client.rs
```diff
@@ -256,7 +256,7 @@ where
         mut self: Pin<&mut Self>,
         cx: &mut Context<'_>,
     ) -> Poll<Result<Profile, Error>> {
-        let span = info_span!("poll_profile");
+        let span = debug_span!("poll_profile");
```
hawkw, line 259: this one honestly seems like it should probably have been at trace from day one, that seems pretty verbose...

linkerd/app/gateway/src/config.rs
```diff
@@ -51,7 +51,7 @@ impl Config {
                 Allow(self.allow_discovery),
             ))
             .check_new_service::<inbound::Target, http::Request<http::boxed::Payload>>()
-            .instrument(|_: &inbound::Target| info_span!("gateway"))
+            .instrument(|_: &inbound::Target| debug_span!("gateway"))
```
hawkw, line 54: IMO we might want to keep this one on at info, so we can tell what errors/warnings come from the gateway...

hawkw (commented):
for the most part, i'm onboard with making the request-path spans disabled with the default filters. however, i noticed a few changes that i'm not convinced about:
- "top-level" spans (e.g. inbound, outbound, gateway, admin tasks) are now debug. i think it's probably worth keeping these at info for errors/warnings, and having just those on by default (+ accept) is probably not a huge impact compared to having a bunch of spans at every stack layer.
- a lot of spans in test code were also demoted to debug. i think it's nice to have these on by default when tests fail, but that's not a huge blocker.

also, might be worth rdoing a perf comparison again, since #726 should have also decreased the latency impact from tracing a bit...

linkerd/app/gateway/src/config.rs
olix0r: I don't this is really practically a concern... You're either looking at a gateway pod or not.

linkerd/app/outbound/src/tcp/tests.rs
olix0r: yeah, global replace. restored test spans.

hawkw (approved):
⛵

PR #731, opened 2020-10-30
buffer: Remove readiness watch

olix0r:
Previously, our `buffer` implementation exposed the inner service's
(initial) readiness via a `watch`. This was primarily present due to the
`fallback` layer which used the service's initial readiness to determine
whether to use the service -- on failures it would fallback to another
service.

Now that fallback is no longer used, the overhead of this readiness
watch can be removed. The fallback middlewares are removed, too,
since they are unused.


hawkw (approved):
lgtm

PR #732, opened 2020-10-31
update the proxy to use Tokio 0.3 

hawkw:
This branch updates the proxy to use Tokio 0.3 and the Tokio 0.3
versions of various ecosystem crates. This includes `tower` 0.4 and
`bytes` 0.6, as well as the Tokio 0.3 versions of `tokio-util`, `hyper`,
`tonic`, etc. Due to API changes in Tokio and in other dependencies, it
was necessary to make some code changes as well as updating
dependencies, but there should be no functional change.

In particular:
* Tokio's support for vectored IO changed significantly in 0.3, so this
  branch updates our use of `AsyncWrite` to participate in the new
  vectored write APIs
* Hyper's HTTP/1.1 upgrade API changed in 0.14, so this branch changes the
  proxy's code for handling CONNECT to use the new API
* Tokio removed support for some socket options, which now need to be
  set using `socket2`
* Tokio removed the `poll_ready` method was removed from the bounded
  MPSC channel, so the proxy's buffers (`linkerd2-buffer` and the
  `buffer` module in `linkerd2-proxy-discover`) had to be switched to
  our own implementation (this merged separately, in PR #759).

Several ecosystem crates have yet to be released, so we depend on them
via Git dependencies for now. The patches in Cargo.toml can be
removed as other dependencies publish their Tokio 0.3 versions.

Performance testing indicates that there may be a bit of an improvement
in tail latencies (especially above p99) on the new Tokio:
Screenshot_20201202_140038


linkerd/dns/Cargo.toml
```diff
@@ -10,8 +10,14 @@ futures = "0.3"
 linkerd2-dns-name = { path = "./name" }
 linkerd2-error = { path = "../error" }
 tracing = "0.1.19"
-tokio = { version = "0.2", features = ["rt-core", "sync"] }
+tokio = { version = "0.3", features = ["rt", "sync"] }
 pin-project = "0.4"
+# trust-dns depends on Tokio 0.2
+# TODO(eliza): when `trust-dns-resolver` supports Tokio 0.3, remove this.
```
olix0r, line 16: It looks like this has been fixed on master https://github.com/bluejekyll/trust-dns/blob/main/crates/resolver/Cargo.toml#L80

linkerd/channel/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "linkerd2-channel"
+version = "0.1.0"
+authors = ["Linkerd Developers <cncf-linkerd-dev@lists.cncf.io>"]
+edition = "2018"
+publish = false
+description = """
+A bounded MPSC channel where senders expose a `poll_ready` method.
```
olix0r, line 8: Why did tokio drop support for poll_ready? Do we understand the thought process behind that?

If we're going to implement our own, we should probably get some tests on it...

Also, would prefer that we do this in its own change (i.e. first do this against main with tokio-0.2, and then port to tokio-0.3 in this branch).

linkerd/dns/src/lib.rs
```diff
@@ -41,18 +41,22 @@ impl Resolver {
...
         let (config, mut opts) = system_conf::read_system_conf()?;
         c.configure_resolver(&mut opts);
         trace!("DNS config: {:?}", &config);
         trace!("DNS opts: {:?}", &opts);
-        Ok(Self::new(config, opts))
+        Ok(Self::new(config, opts).await)
     }
 
-    pub fn new(config: ResolverConfig, mut opts: ResolverOpts) -> Self {
+    pub async fn new(config: ResolverConfig, mut opts: ResolverOpts) -> Self {
```
olix0r, line 54: Why is this async? Nothing awaits inside of it?

linkerd/dns/src/lib.rs
```diff
@@ -41,18 +41,22 @@ impl Resolver {
...
         // Disable Trust-DNS's caching.
         opts.cache_size = 0;
-        let dns = AsyncResolver::tokio(config, opts).expect("Resolver must be valid");
+        // This function is synchronous, but needs to be called within the Tokio
+        // 0.2 runtime context, since it gets a handle.
```
olix0r, line 58: I think this comment is outdated?

linkerd/dns/src/lib.rs
```diff
@@ -41,18 +41,22 @@ impl Resolver {
     /// could not be parsed.
     ///
     /// TODO: This should be infallible like it is in the `domain` crate.
-    pub fn from_system_config_with<C: ConfigureResolver>(c: &C) -> Result<Self, ResolveError> {
+    pub async fn from_system_config_with<C: ConfigureResolver>(
```
olix0r, line 44: I don't believe this needs to be async if the `new` isn't...

linkerd/duplex/src/lib.rs
```diff
@@ -200,18 +200,21 @@ impl Buf for CopyBuf {
...
     fn remaining_mut(&self) -> usize {
         self.buf.len() - self.write_pos
     }
 
-    fn bytes_mut(&mut self) -> &mut [MaybeUninit<u8>] {
+    fn bytes_mut(&mut self) -> &mut bytes::buf::UninitSlice {
         unsafe {
             // this is, in fact, _totally fine and safe_: all the memory is
             // initialized.
             // there's just no way to turn a `&[T]` into a `&[MaybeUninit<T>]`
```
olix0r, line 212: This comment should be updated. We're no longer doing anything with MaybeUninit

linkerd/io/src/boxed.rs
```diff
@@ -53,126 +37,93 @@ impl AsyncWrite for BoxedIo {
...
-                _ => panic!("poll_write_buf should return Poll::Ready(Ok(_))"),
-            }
-            Poll::Ready(Ok(()))
-        })
-        .await
-        .unwrap()
-    }
-}
+// TODO(eliza): Tokio 0.3 doesn't have writev support. Bring all this back when
+// vectored writes work again.
```
olix0r, line 51: Is this still true?

linkerd/io/src/lib.rs
```diff
@@ -50,86 +52,95 @@ impl PeerAddr for tokio::io::DuplexStream {
         Ok(([0, 0, 0, 0], 0).into())
     }
 }
-mod internal {
```
olix0r, line 53: Why do we need to change the visibility as a part of this? If this is necessary, can it be decoupled from the upgrade?

linkerd/io/src/lib.rs
```diff
@@ -50,86 +52,95 @@ impl PeerAddr for tokio::io::DuplexStream {
...
-            cx: &mut Context<'_>,
-            mut buf: &mut dyn BufMut,
-        ) -> Poll<usize> {
-            self.poll_read_buf(cx, &mut buf)
-        }
+// Vendored from upstream Tokio implementation, which is currently private.
+// https://github.com/tokio-rs/tokio/blob/24ed874e81fa27a6505613051955127ba8ddfdf2/tokio-util/src/lib.rs#L72-L100
+//
+// TODO(eliza): if tokio makes this public, it would be nice to not vendor it.
+pub fn poll_read_buf<T: AsyncRead>(
```
olix0r, line 104: I'd be +1 to trying to get the upstream API exposed; but in lieu of taht we should at least document what it does

linkerd/io/src/lib.rs
```diff
@@ -50,86 +52,95 @@ impl PeerAddr for tokio::io::DuplexStream {
...
-        fn poll_read_buf_erased(
-            self: Pin<&mut Self>,
-            cx: &mut Context<'_>,
-            mut buf: &mut dyn BufMut,
-        ) -> Poll<usize> {
-            self.poll_read_buf(cx, &mut buf)
-        }
+    let n = {
+        let dst = buf.bytes_mut();
+        let dst = unsafe { &mut *(dst as *mut _ as *mut [MaybeUninit<u8>]) };
```
olix0r, line 115: Worth a `// Safety: `

linkerd/proxy/transport/src/lib.rs
```diff
@@ -27,7 +27,23 @@ fn set_nodelay_or_warn(socket: &TcpStream) {
 }
 
 fn set_keepalive_or_warn(tcp: &TcpStream, ka: Option<Duration>) {
-    if let Err(e) = tcp.set_keepalive(ka) {
+    // XXX(eliza): we now have to do "socket2 nonsense" here because tokio
+    // doesn't have `set_keepalive` any more...
```
olix0r, line 31: Replace this with a `// Safety:`? Or are we planning to get this upstreamed before merge?

linkerd/service-profiles/src/split.rs
```diff
@@ -225,3 +225,35 @@ where
         }
     }
 }
+
+impl<T, N, S, Req> fmt::Debug for Inner<T, N, S, Req>
```
olix0r, line 229: A lot of verbosity for something we don't really need. tioli; but pretty sure we can just drop the debug impls and everything will compile

olix0r (changes requested):
Looking pretty good.

It seems like there are some bigger structural changes (io, linkerd2-channel, etc) that could be split out of this and changed on main before the upgrade. Would help to reduce the size/surface of this PR.

olix0r:
Also not sure why, but if i do a `git co main -- Cargo.lock && cargo check` I get a pretty big diff -- maybe worth doing this to minimize unnecessary change? Or it may just be non-deterministic

linkerd/proxy/transport/src/lib.rs
hawkw: The hope was that this would be upstreamed, but it seems like that's not the case. However, it _does_ look like we will at least be able to convert tokio sockets into std sockets safely in the future.

linkerd/io/src/lib.rs
hawkw: It's public now, I'll just remove this.

linkerd/io/src/lib.rs
hawkw: This is now depended on from upstream, so the code is now removed.

hawkw:
> Also not sure why, but if i do a `git co main -- Cargo.lock && cargo check` I get a pretty big diff -- maybe worth doing this to minimize unnecessary change? Or it may just be non-deterministic

That's probably my fault --- I manually edited the lockfile a couple times while merging main, and I may have included some deps we don't actually need, etc? I'll fix that.

linkerd/channel/Cargo.toml
hawkw: > Why did tokio drop support for poll_ready? Do we understand the thought process behind that?

@olix0r I'm happy to provide some insight into this. Most of the Tokio synchronization primitives lost their `poll`-based interfaces in 0.3 as part of the move to intrusive lists of wakers for synchronization primitives (see tokio-rs/tokio#2325, tokio-rs/tokio#2509, and tokio-rs/tokio#2861). This change takes advantage of the inherently pinned nature of `async fn` and `async` blocks to avoid needing a separate heap allocation to store the waiter state for a task waiting on a synchronization primitive. However, it means that a synchronization primitive can _only_ be waited on when the future that waits on it is pinned --- otherwise, there is a potential dangling pointer. The `poll`-based APIs allowed waiting on synchronization primitives from unpinned contexts, so they were removed. 

To wait on the synchronization primitives from contexts that may not be pinned, such as `poll_ready`, it's necessary to add a `Pin<Box<...>>` around the future that's waiting on the synchronization primitive. This ensures that the future will not move while it's part of the wait list. It's important to note that this isn't an _additional_ allocation per waiter versus Tokio 0.2; instead, it's the same allocation that would have _always_ happened internally to the synchronization primitive in the 0.2 API. Now, it's moved outside of the `tokio::sync` type so that it can be avoided when used with `async`/`await` syntax, and added by the user when polling the sync primitives.

hawkw: > If we're going to implement our own, we should probably get some tests on it...
> 
> Also, would prefer that we do this in its own change (i.e. first do this against main with tokio-0.2, and then port to tokio-0.3 in this branch).

I'm happy to try and factor this out into a separate PR, and add tests. It's worth noting that I used approximately the same approach in tower-rs/tower#476, for `tower-buffer`; it's necessary to reimplement it in the proxy codebase because we use our own buffer implementation.

Also, it's possible that there will eventually be a wrapper for MPSC senders that adds back `poll_ready`, in `tokio-util`: https://github.com/tokio-rs/tokio/pull/3105. This is moved out of the core MPSC API so that the additional overhead is opt-in. However, I'm not sure when that PR will be ready to merge...

linkerd/proxy/transport/src/lib.rs
hawkw: added a nicer comment in cb9e35c30d7671f4ecf0684c38473296193af86b

olix0r (approved):

PR #733, opened 2020-11-02
Allow specifying the number of available cores via the env

olix0r:
When testing the proxy (i.e. outside of Kubernetes), it can be desirable
to limit the number of cores the proxy uses. This change introduces a
`LINKERD2_PROXY_CORE` environment that overrides the `num-cpus`-based
detection.

### DIFF
--- a/linkerd2-proxy/src/rt.rs
+++ b/linkerd2-proxy/src/rt.rs
@@ -2,10 +2,18 @@ use tokio::runtime::{Builder, Runtime};
 
 #[cfg(feature = "multicore")]
 pub(crate) fn build() -> Runtime {
-    // The proxy creates an additional admin thread, but it would be wasteful to allocate a whole
-    // core to it; so we let the main runtime consume all cores the process has. The basic scheduler
-    // is used when the threaded scheduler would provide no benefit.
-    match num_cpus::get() {
+    // The proxy creates an additional admin thread, but it would be wasteful to
+    // allocate a whole core to it; so we let the main runtime consume all
+    // available cores. The number of available cores is determined by checking
+    // the environment or by inspecting the host or cgroups.
+    //
+    // The basic scheduler is used when the threaded scheduler would provide no
+    // benefit.
+    let cpus = std::env::var("LINKERD2_PROXY_CORES")
+        .ok()
+        .and_then(|v| v.parse::<usize>().ok())
+        .unwrap_or_else(|| num_cpus::get());
+    match cpus {
         // `0` is unexpected, but it's a wild world out there.
         0 | 1 => Builder::new()
             .enable_all()
--- a/linkerd2-proxy/src/rt.rs
+++ b/linkerd2-proxy/src/rt.rs
@@ -1,4 +1,5 @@
 use tokio::runtime::{Builder, Runtime};
+use tracing::{debug, warn};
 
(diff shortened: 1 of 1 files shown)


linkerd2-proxy/src/rt.rs
```diff
@@ -2,10 +2,18 @@ use tokio::runtime::{Builder, Runtime};
...
+    // The proxy creates an additional admin thread, but it would be wasteful to
+    // allocate a whole core to it; so we let the main runtime consume all
+    // available cores. The number of available cores is determined by checking
+    // the environment or by inspecting the host or cgroups.
+    //
+    // The basic scheduler is used when the threaded scheduler would provide no
+    // benefit.
+    let cpus = std::env::var("LINKERD2_PROXY_CORES")
+        .ok()
+        .and_then(|v| v.parse::<usize>().ok())
```
hawkw, line 14: should we maybe print something if the value is not a number?

linkerd2-proxy/src/rt.rs
```diff
@@ -2,10 +2,18 @@ use tokio::runtime::{Builder, Runtime};
...
+        .unwrap_or_else(|| num_cpus::get());
+    match cpus {
```
hawkw, line 16: should we maybe use 
```suggestion
    match std::cmp::max(cpus, num_cpus::get()) {
```
or something? so that we don't spawn 2000 threads on the off chance that someone tries to set this like it's a k8s CPU request.

this may not matter _that_ much if we only expect the env var to be used for benchmarking etc rather than being UI, but it does seem like it might be good to not silently screw up benchmark results?

linkerd2-proxy/src/rt.rs
```diff
@@ -2,10 +2,18 @@ use tokio::runtime::{Builder, Runtime};
...
+    // allocate a whole core to it; so we let the main runtime consume all
+    // available cores. The number of available cores is determined by checking
+    // the environment or by inspecting the host or cgroups.
+    //
+    // The basic scheduler is used when the threaded scheduler would provide no
+    // benefit.
+    let cpus = std::env::var("LINKERD2_PROXY_CORES")
+        .ok()
+        .and_then(|v| v.parse::<usize>().ok())
+        .unwrap_or_else(|| num_cpus::get());
```
hawkw, line 15: nit/TIOLI: could be
```suggestion
        .unwrap_or_else(num_cpus::get);
```

hawkw (approved):
i'm +1 on making this change, but i had a couple of minor nits.

linkerd2-proxy/src/rt.rs
```diff
@@ -1,11 +1,39 @@
...
+    if cores > cpus {
+        warn!(
+            cpus,
+            LINKERD2_PROXY_CORES = cores,
+            "Ignoring configuration due to insufficient resources"
+        );
+        cores = cpus;
+    }
+
+    debug!(cores, "Initializing runtime");
```
hawkw, line 35: TIOLI maybe this should be
```suggestion
    info!(cores, "Initializing runtime");
```
so that it gets logged in the default configuration? since it only gets logged once, there's no way to get the proxy to print it _after_ it's started up, so i feel like it may as well be there.

hawkw (approved):

PR #734, opened 2020-11-06
http: Configure client connection pools

olix0r:
When an HTTP/1.1 client has many concurent requests, it may create an
arbitrary number of TCP connections. These connections are retained for
90s before they are dropped. In HTTP/1.1. applications communicate at
high concurrency, this can lead to elevated memory requirements.

Similar behavior is observed with HTTP/2 clients (i.e. when using mesh
protocol upgrades) if the application communicates with many unique
services, as is the case for Prometheus.

To address this issue, we want to be more aggressive about tearing down
idle/unused clients. This is primarily achieved by reducing the outbound
"max idle age" setting to 5s, which matches the default Kubernetes DNS
TTL. On the inbound side, the max idle age is set to 20s, as the number
of clients is typically on the order of <5.

Additionally, `MAX_IDLE_CONNS_PER_ENDPOINT` configurations are
introduced to bound the number of idle connections that may be retained
in an HTTP/1.1 client's connection pool. These pools remain unbounded by
default, as this setting causes elevated CPU usage in load tests. It is
generally preferable to bound the length of time an idle
client/connection is cached.

### DIFF
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -2,7 +2,7 @@ use crate::core::{
     addr,
     config::*,
     control::{Config as ControlConfig, ControlAddr},
-    proxy::http::h2,
+    proxy::http::{h1, h2},
     transport::{listen, tls},
     Addr, AddrMatch, NameMatch,
 };
@@ -347,6 +347,8 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
             bind: bind.with_orig_dst_addr(outbound_orig_dst),
             h2_settings,
         };
+        let cache_max_idle_age =
+            outbound_cache_max_idle_age?.unwrap_or(DEFAULT_OUTBOUND_ROUTER_MAX_IDLE_AGE);
         let connect = ConnectConfig {
             keepalive: outbound_connect_keepalive?,
             timeout: outbound_connect_timeout?.unwrap_or(DEFAULT_OUTBOUND_CONNECT_TIMEOUT),
@@ -356,6 +358,10 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
                 DEFAULT_OUTBOUND_CONNECT_BACKOFF,
             )?,
             h2_settings,
+            h1_settings: h1::PoolSettings {
+                max_idle: 2,
+                idle_timeout: cache_max_idle_age,
+            },
         };
 
         let dispatch_timeout =
@@ -369,8 +375,7 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
             proxy: ProxyConfig {
                 server,
                 connect,
-                cache_max_idle_age: outbound_cache_max_idle_age?
-                    .unwrap_or(DEFAULT_OUTBOUND_ROUTER_MAX_IDLE_AGE),
+                cache_max_idle_age,
                 buffer_capacity,
                 dispatch_timeout,
                 max_in_flight_requests: outbound_max_in_flight?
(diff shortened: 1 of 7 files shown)


kleimkuhler (approved):
Lgtm 👍

linkerd/app/src/env.rs
```diff
@@ -82,6 +82,9 @@ pub const ENV_BUFFER_CAPACITY: &str = "LINKERD2_PROXY_BUFFER_CAPACITY";
 pub const ENV_INBOUND_ROUTER_MAX_IDLE_AGE: &str = "LINKERD2_PROXY_INBOUND_ROUTER_MAX_IDLE_AGE";
 pub const ENV_OUTBOUND_ROUTER_MAX_IDLE_AGE: &str = "LINKERD2_PROXY_OUTBOUND_ROUTER_MAX_IDLE_AGE";
 
+const ENV_INBOUND_MAX_IDLE_PER_ENDPOINT: &str = "LINKERD2_PROXY_MAX_IDLE_PER_ENDPOINT";
```
hawkw, line 85: not/tioli: should this maybe be "MAX_IDLE_CONNS" or something? we also use the string "MAX_IDLE" to refer to the service cache eviction timeout, which is a pretty different thing...

linkerd/proxy/http/src/h1.rs
```diff
@@ -108,9 +117,17 @@ where
 
             if client.is_none() {
                 debug!(use_absolute_form, "Caching new client");
-                *client = Some(hyper::Client::builder().set_host(use_absolute_form).build(
-                    HyperConnect::new(self.connect.clone(), self.target.clone(), use_absolute_form),
-                ));
+                *client = Some(
+                    hyper::Client::builder()
+                        .pool_max_idle_per_host(self.pool.max_idle)
+                        .pool_idle_timeout(self.pool.idle_timeout)
```
hawkw, line 123: here's a thought, take it or leave it: I wonder if it would make sense to just pass around a builder, rather than a `PoolSettings`? may not be the right approach...

linkerd/app/src/env.rs
```diff
@@ -189,8 +192,11 @@ const DEFAULT_INITIAL_CONNECTION_WINDOW_SIZE: u32 = 1048576; // 1MB ~ 16 streams
 // 10_000 is arbitrarily chosen for now...
 const DEFAULT_BUFFER_CAPACITY: usize = 10_000;
 
-const DEFAULT_INBOUND_ROUTER_MAX_IDLE_AGE: Duration = Duration::from_secs(60);
-const DEFAULT_OUTBOUND_ROUTER_MAX_IDLE_AGE: Duration = Duration::from_secs(60);
+const DEFAULT_INBOUND_ROUTER_MAX_IDLE_AGE: Duration = Duration::from_secs(10);
+const DEFAULT_OUTBOUND_ROUTER_MAX_IDLE_AGE: Duration = Duration::from_secs(3);
+
+const DEFAULT_INBOUND_MAX_IDLE_PER_ENDPOINT: usize = std::usize::MAX;
+const DEFAULT_OUTBOUND_MAX_IDLE_PER_ENDPOINT: usize = std::usize::MAX;
```
hawkw, line 199: The PR description suggests that only two idle connections will be kept in the pool. This suggests that we don't cap the number of idle connections by default. Do we expect the control plane to always set this to 2, or something? Or should we change the default value here?

linkerd/app/src/env.rs
```diff
@@ -189,8 +192,11 @@ const DEFAULT_INITIAL_CONNECTION_WINDOW_SIZE: u32 = 1048576; // 1MB ~ 16 streams
 // 10_000 is arbitrarily chosen for now...
 const DEFAULT_BUFFER_CAPACITY: usize = 10_000;
 
-const DEFAULT_INBOUND_ROUTER_MAX_IDLE_AGE: Duration = Duration::from_secs(60);
-const DEFAULT_OUTBOUND_ROUTER_MAX_IDLE_AGE: Duration = Duration::from_secs(60);
+const DEFAULT_INBOUND_ROUTER_MAX_IDLE_AGE: Duration = Duration::from_secs(10);
+const DEFAULT_OUTBOUND_ROUTER_MAX_IDLE_AGE: Duration = Duration::from_secs(3);
```
hawkw, line 196: nit: *might* be nice to have a comment explaining why these have different timeouts?

hawkw (commented):
overall, this looks good! however, i noticed the default max number of pooled connections in the code doesn't line up with the number you quoted in the PR description — is this correct?

linkerd/app/src/env.rs
hawkw: Update: ah, I see, this was explained in https://github.com/linkerd/linkerd2/issues/5146#issuecomment-723204412. Let's just be sure to update the commit message when merging?

hawkw (approved):

PR #736, opened 2020-11-11
outbound: Skip endpoint resolution on profile hint

olix0r:
Profile resolutions may provide an endpoint response when the
target is a pod or otherwise known endpoint. When these endpoints fail
requests with an I/O error, we assume that the endpoint is defunct and
fail the service so that subsequent connections require a new
resolution.

### DIFF
--- a/linkerd/stack/src/fail_on_error.rs
+++ b/linkerd/stack/src/fail_on_error.rs
@@ -2,15 +2,14 @@ use linkerd2_error::Error;
 use std::{
     future::Future,
     pin::Pin,
-    sync::Arc,
+    sync::{Arc, RwLock},
     task::{Context, Poll},
 };
-use tokio::sync::Mutex;
 
 #[derive(Clone, Debug)]
 pub struct FailOnError<E, S> {
     inner: S,
-    error: Arc<Mutex<Option<SharedError>>>,
+    error: Arc<RwLock<Option<SharedError>>>,
     _marker: std::marker::PhantomData<E>,
 }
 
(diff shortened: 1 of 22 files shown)


linkerd/stack/src/fail_on_error.rs
```diff
@@ -0,0 +1,88 @@
+use linkerd2_error::Error;
+use std::{
+    future::Future,
+    pin::Pin,
+    sync::Arc,
+    task::{Context, Poll},
+};
+use tokio::sync::Mutex;
```
hawkw, line 8: AFAICT this lock is never held across an await, so it's probably more efficient to just use a `std::sync` mutex.

linkerd/stack/src/fail_on_error.rs
```diff
@@ -0,0 +1,88 @@
...
+    type Response = S::Response;
+    type Error = Error;
+    type Future = Pin<Box<dyn Future<Output = Result<S::Response, Error>> + Send + 'static>>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Error>> {
+        futures::ready!(self.inner.poll_ready(cx).map_err(Into::into))?;
+        if let Ok(e) = self.error.try_lock() {
+            if let Some(e) = &*e {
+                return Poll::Ready(Err(e.clone().into()));
+            }
```
hawkw, line 47: It seems like we may want to use a `RwLock`, so that this case is not synchronized?  It seems like it's possible for `poll_ready` to spuriously return `Ok` if another thread is currently accessing the lock. If we used a `RwLock` here we would be able to clone the error even if someone else also is...

hawkw (approved):

PR #737, opened 2020-11-12
h2: enable HTTP/2 keepalive PING frames

hawkw:
This branch enables HTTP/2 PING frames in the proxy's HTTP/2 clients and
servers. The timeout for responding to a PING frame is configured based
on the `{INBOUND, OUTBOUND}_{CONNECT, ACCEPT}_KEEPALIVE` env variables,
and the interval between PING frames is currently 1/4th of the timeout.
I'm happy to change that if anyone has better ideas.

Collecting metrics related to H2 PINGs probably requires support in 
Hyper that doesn't currently exist, so this PR doesn't add that. We 
can implement metrics in a follow-up, as it's lower priority.

Closes linkerd/linkerd2#1580

### DIFF
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -371,28 +371,36 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
     let ingress_mode = parse(strings, ENV_INGRESS_MODE, parse_bool)?.unwrap_or(false);
 
     let outbound = {
+        let keepalive = outbound_accept_keepalive?;
         let bind = listen::Bind::new(
             outbound_listener_addr?
                 .unwrap_or_else(|| parse_socket_addr(DEFAULT_OUTBOUND_LISTEN_ADDR).unwrap()),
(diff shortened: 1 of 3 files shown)


linkerd/app/src/env.rs
```diff
@@ -371,28 +371,36 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
+            keepalive,
             timeout: outbound_connect_timeout?.unwrap_or(DEFAULT_OUTBOUND_CONNECT_TIMEOUT),
             backoff: parse_backoff(
                 strings,
                 OUTBOUND_CONNECT_BASE,
                 DEFAULT_OUTBOUND_CONNECT_BACKOFF,
             )?,
-            h2_settings,
+            h2_settings: h2::Settings {
+                keepalive_timeout: Keepalive,
```
olix0r, line 401:
```suggestion
                keepalive_timeout: keepalive,
```

hawkw: oops

olix0r (approved):

PR #738, opened 2020-11-12
actions: Add timeouts to GitHub workflows

olix0r:
In order to prevent runaway tests from consuming actions resources, we
set reasonable default timeouts for each job. This replaces the default
6 hour(!) timeout.


hawkw (approved):
:+1:

PR #740, opened 2020-11-13
Introduce a connection-header module

olix0r:
Proxies are not able to securely transport server-first protocols,
since the server-side proxy cannot now whether to perform TLS detection
or direct forwarding. Similarly, multicluster gateways are unable to
proxy arbitrary TCP connections, since the gateway has no means to
discover the intended target for a connection in the destination
cluster.

In order to address both of these situations, this change introduces a
"connection header" module that can be used to transport routing
information for arbitrary TCP streams. When instrumented in the stack,
this module will allow inbound proxies to receive mTLS connections on a
dedicated port and route the connection to proper destination port.
Similar logic applies for multicluster gateways.

This protocol includes a prefix, `"proxy.l5d.io/connect\r\n\r\n"` so
that proxies can differentiate these streams from HTTP streams.
Following this, a network-endian 4-byte length prefix is encoded,
indicating the number of bytes comprising the protocol header. Finally,
the protocol header is encoded as a protobuf structure. The protobuf
message currently includes a target port, which must be set to indicate
how the connection should be routed, as well as an optional `name`,
which may be used for discovery in the multicluster gateway use case.

This change does not actually instrument the module into the proxy
stacks.

### DIFF
--- a/linkerd/connection-header/src/lib.rs
+++ b/linkerd/connection-header/src/lib.rs
@@ -1,0 +1,271 @@
+use bytes::{
+    buf::{Buf, BufMut},
+    BytesMut,
+};
+use linkerd2_dns_name::Name;
+use linkerd2_error::Error;
+use linkerd2_io::{self as io, AsyncReadExt};
+use linkerd2_proxy_transport::Detect;
+use prost::Message;
+use std::str::FromStr;
+
+mod proto {
+    include!(concat!(env!("OUT_DIR"), "/header.proxy.l5d.io.rs"));
+}
+
+#[derive(Clone, Debug)]
+pub struct Header {
+    /// The target port.
+    port: u16,
+
+    /// The logical name of the target (service), if one is known.
+    pub name: Option<Name>,
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct DetectHeader(());
+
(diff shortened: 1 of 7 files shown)


linkerd/connection-header/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+            None
+        } else {
+            let n = Name::from_str(&h.name)
+                .map_err(|_| io::Error::new(io::ErrorKind::InvalidData, "Invalid name"))?;
+            Some(n)
+        };
+
+        Ok(Some(Self {
+            name,
+            port: h.port as u16,
```
hawkw, line 157: this will silently truncate the port if it's greater than 65535, should we return an error instead?

if we got a port number greater than `u16::max` and we know the remote peer is using the same code (which will only encode `Header` structs with `u16` port values), seeing a value higher than `u16::max` probably indicates that the message is corrupted or we decoded it wrongly, right?

hawkw (commented):
overall, this seems good. it occurs to me that the length-delimited encoding *could* be implemented using `tokio-util::codec::LengthDelimited`, but the implementation is simple enough that this may not really be worth it. 

i had a few notes but didn't notice any major flaws.

PR #741, opened 2020-11-13
add tests; fix profile endpoint TLS ID being ignored

hawkw:
This branch fixes a bug where TLS would never be used for endpoint hints
provided by service profile discovery, even when the profile endpoint
metadata has an identity. This is because the wrong conversion was used
when mapping a `Logical` target into a `tcp::Endpoint`, which always
ignored the profile endpoint metadata.

Additionally, I've added tests for the new behavior added in this
branch, including one which reproduced the above issue, and one for
connection error propagation.


olix0r (approved):
❤️

PR #742, opened 2020-11-14
outbound: Avoid redundant TCP endpoint resolution

olix0r:
This change extends the changes in linkerd/linkerd2-proxy#736 to the
TCP-forwarding stack so that load balancer resolutions are not created
when the control plane does not indicate that the target is a service.

### DIFF
--- a/linkerd/app/outbound/src/server.rs
+++ b/linkerd/app/outbound/src/server.rs
@@ -101,6 +101,21 @@ where
     ))
     .push_map_target(tcp::Concrete::from)
     .push(profiles::split::layer())
+    .check_new_service::<tcp::Logical, transport::io::PrefixedIo<transport::metrics::SensorIo<I>>>()
+    .push_switch(
+        tcp::Logical::should_resolve,
+        svc::stack(tcp_connect.clone())
+            .push_make_thunk()
+            .check_make_service::<tcp::Endpoint, ()>()
+            .push_on_response(svc::layer::mk(tcp::Forward::new))
+            .into_new_service()
+            .check_new_service::<tcp::Endpoint, transport::io::PrefixedIo<transport::metrics::SensorIo<I>>>()
+            .push_map_target(tcp::Endpoint::from_logical(
+                tls::ReasonForNoPeerName::NotProvidedByServiceDiscovery,
+            ))
+            .check_new_service::<tcp::Logical, transport::io::PrefixedIo<transport::metrics::SensorIo<I>>>()
+            .into_inner(),
+    )
     .push_on_response(
         svc::layers()
             .push_failfast(dispatch_timeout)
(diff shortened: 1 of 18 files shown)


linkerd/app/outbound/src/server.rs
```diff
@@ -101,6 +101,21 @@ where
...
+            .push_make_thunk()
+            .check_make_service::<tcp::Endpoint, ()>()
+            .push_on_response(svc::layer::mk(tcp::Forward::new))
+            .into_new_service()
+            .check_new_service::<tcp::Endpoint, transport::io::PrefixedIo<transport::metrics::SensorIo<I>>>()
+            .push_map_target(tcp::Endpoint::from_logical(
+                tls::ReasonForNoPeerName::NotProvidedByServiceDiscovery,
+            ))
+            .check_new_service::<tcp::Logical, transport::io::PrefixedIo<transport::metrics::SensorIo<I>>>()
+            .into_inner(),
```
hawkw, line 117: nit, take it or leave it: i kind of feel like this code might be a little easier to read if this stack was in a `let`-binding so that we could give it a descriptive name? not a blocker though!

linkerd/app/outbound/src/tcp/tests.rs
```diff
@@ -197,10 +199,20 @@ async fn resolutions_are_reused() {
...
     let resolve_state = resolver.handle();
 
-    let profiles = support::profiles().profile(addr, Default::default());
+    let profiles = support::profiles().profile(
+        addr,
+        profile::Profile {
+            name: Some(svc_name),
+            ..profile::Profile::default()
+        },
+    );
```
hawkw, line 215: nit/tioli: the tests might be a _little_ more concise if we added a
```rust
pub fn named(name: &str) -> Profile {
    let name =  profile::Name::try_from(name.as_bytes()).unwrap();
    Profile {
         name: Some(name),
         ..Profile::default()
    }
}
```
or something in the test support `profiles` module? not a blocker though.

hawkw: ...oh, i guess we also use the DNS name for addrs, so the let binding does make sense. _might_ be nice to have a test helper for `&str` -> `profile::Name` conversion, but not a blocker.

linkerd/app/test/src/connect.rs
```diff
@@ -35,9 +35,14 @@ where
...
             tracing::trace!("connecting...");
-            match self.endpoints.lock().unwrap().get_mut(&addr) {
+            let mut endpoints = self.endpoints.lock().unwrap();
+            match endpoints.get_mut(&addr) {
                 Some(f) => (f)(endpoint),
-                None => panic!("did not expect to connect to the endpoint {:?}", endpoint),
+                None => panic!(
+                    "did not expect to connect to the endpoint {} not in {:?}",
+                    addr,
+                    endpoints.keys().collect::<Vec<_>>()
```
hawkw, line 44: <3

hawkw (approved):
lgtm! couple style nits, feel free to ignore them

PR #744, opened 2020-11-15
detect: Make protocol detection more robust

olix0r:
The existing protocol detection implementation is somewhat naive: we
simply perform a single read on the socket and use that initial buffer
to try to determine the protocol. This means that if the initial read
only returns a few bytes, we won't actually have enough information to
do protocol detection. Furthermore, because we don't actually read until
a `\r\n`, we can incorrectly infer that a non-HTTP protocol is HTTP/1.1.

This change fixes this by replacing the `Prefix` middleware with a
`Detect` service and trait that does its own buffering as it tries to
read from the socket.

When detection times out, an error is thrown to the server. This change
requires modifying test data to include a trailing newline (since it's
not feasible to close these connections given the current test
infrastructure).

This change increases the default dispatch (and therefore detection)
timeouts to 5s.


linkerd/app/inbound/src/lib.rs
```diff
@@ -375,20 +376,22 @@ impl Config {
...
             svc::stack(tcp_forward)
                 .push_map_target(TcpEndpoint::from)
                 .into_inner(),
             drain.clone(),
         ))
-        .push_on_response(svc::layers().push_spawn_buffer(buffer_capacity).push(
-            transport::Prefix::layer(
-                http::Version::DETECT_BUFFER_CAPACITY,
-                detect_protocol_timeout,
+        .cache(
```
olix0r, line 387: This probably shouldn't be a cache?

olix0r: Ah, previously `ServeHttp` did its own caching. This replaces that.

linkerd/app/outbound/src/ingress.rs
```diff
@@ -130,13 +131,10 @@ where
...
     svc::stack(http::NewServeHttp::new(h2_settings, http, tcp, drain))
-        .check_new_service::<tcp::Accept, io::PrefixedIo<transport::metrics::SensorIo<I>>>()
-        .push_on_response(svc::layers().push_spawn_buffer(buffer_capacity).push(
-            transport::Prefix::layer(
-                http::Version::DETECT_BUFFER_CAPACITY,
-                detect_protocol_timeout,
-            ),
-        ))
+        .check_new_service::<(Option<http::Version>, tcp::Accept), io::PrefixedIo<transport::metrics::SensorIo<I>>>()
+        .check_new_clone::<(Option<http::Version>, tcp::Accept)>()
```
hawkw, line 135: nit, take it or leave it: i wonder if the `(Option<http::Version>, tcp::Accept)` tuple ought to be a named struct, or at least a type alias?

also, maybe we could import `transport::metrics::SensorIo` so this line is a little less wordy? up to you, though.

linkerd/proxy/http/src/detect.rs
```diff
@@ -0,0 +1,227 @@
...
+
+                // Not a match. Don't check for an HTTP/2 preface again.
+                maybe_h2 = false;
+            }
+
+            if maybe_h1 {
+                // Scan up to the first line ending to determine whether the
+                // request is HTTP/1.1.
+                for i in scan_idx..(buf.len() - 1) {
+                    if &buf[i..=i + 1] == b"\r\n" {
```
hawkw, line 77: nit, take it or leave it: I think it's possible that this logic could also be accomplished with [`std::slice::windows`](https://doc.rust-lang.org/std/primitive.slice.html#method.windows), which might be a little bit neater:

```rust
for window in &buf[scan_index..].windows() {
    if window == b"\r\n" {
        // ...
    }
}
```

linkerd/proxy/http/src/detect.rs
```diff
@@ -0,0 +1,227 @@
...
+            if !maybe_h1 && !maybe_h2 {
+                trace!("Not HTTP");
+                return Ok(None);
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
hawkw, line 118: <3

linkerd/proxy/http/src/detect.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    const HTTP11_LINE: &'static [u8] = b"GET / HTTP/1.1\r\n";
+    const TIMEOUT: time::Duration = time::Duration::from_secs(10);
+    const H2_AND_GARBAGE: &'static [u8] = b"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\ngarbage";
+    const GARBAGE: &'static [u8] =
+        b"garbage garbage garbage garbage garbage garbage garbage garbage garbage garbage garbage garbage garbage garbage garbage garbage garbage";
+
+    #[tokio::test]
+    async fn h2() {
+        let _ = tracing_subscriber::fmt::try_init();
```
hawkw, line 131: nit/tioli: if these tests all used a single-threaded tokio executor, we could just give each one its own single-threaded default subscriber:
```suggestion
    use tracing_subscriber::prelude::*;
      
    #[tokio::test(flavor = "current_thread")]
    async fn h2() {
        let _subscriber = tracing_subscriber::fmt().set_default();
```
this isn't hugely different but it does keep the tests completely isolated...

hawkw: also, we may want to use 
```rust
tracing_subscriber::fmt().with_test_writer()
```
here so that the traces get captured by libtest and only printed on failures?

linkerd/proxy/http/src/server.rs
```diff
@@ -28,24 +28,15 @@ pub struct NewServeHttp<F, H> {
...
-    new_tcp: F,
-    tcp: Option<F::Service>,
-    new_http: H,
-    http1: Option<H::Service>,
-    h2: Option<H::Service>,
-    server: hyper::server::conn::Http<trace::Executor>,
-    drain: drain::Watch,
+#[derive(Clone, Debug)]
+pub enum ServeHttp<F, H> {
+    Opaque(F, drain::Watch),
```
hawkw, line 33: nit: it seems a little weird to me to have something named `ServeHttp` that might be serving a protocol that isn't HTTP...but i notice that we already did that, so it's not a blocker for _this_ PR...

linkerd/proxy/http/src/server.rs
```diff
@@ -145,120 +124,72 @@ where
...
-                    trace!("HTTP service already exists");
-                    svc
-                } else {
-                    trace!("Building new HTTP service");
-                    let svc = self
-                        .new_http
-                        .new_service((HttpVersion::Http1, self.target.clone()));
-                    self.http1 = Some(svc.clone());
-                    svc
-                };
```
hawkw, line 162: nice to see that this logic is no longer necessary --- this always felt a little messy to me

linkerd/proxy/transport/src/detect.rs
```diff
@@ -0,0 +1,163 @@
...
+                    let bytes = buf.len();
+                    let elapsed = time::Instant::now() - t0;
+                    return Err(DetectTimeout { bytes, elapsed }.into());
+                }
+            };
+            debug!(
+                ?protocol,
+                elapsed = ?(time::Instant::now() - t0),
+                "Detected"
+            );
```
hawkw, line 127: this is really lovely --- it's nice how async-await lets us express the timing code in a way that feels really natural. i feel like this would have been pretty unpleasant to do with manual future impls.

linkerd/app/outbound/src/ingress.rs
olix0r: i'm okay with using tuple targets temporarily -- i think it's better than having a bunch of strict dependencies and traits. For instance, ServeHttp takes `(Option<Version>, T)` and uses `(Version, T)` on its inner service... doing this with named types/traits would be much more verbose for no real benefit, imo

hawkw: ah, yup, that makes sense --- carry on then!

linkerd/app/integration/src/tests/identity.rs
```diff
@@ -25,7 +25,7 @@ async fn nonblocking_identity_detection() {
         .await;
     let proxy = proxy::new().identity(id_svc);
 
-    let msg1 = "custom tcp hello";
+    let msg1 = "custom tcp hello\n";
```
hawkw, line 28: nit/TIOLI: do you think it's worth also having some tests with `\r\n`s in them, just to make sure we're handling that correctly? the logic isn't too complex so i'm not *too* concerned about it, just a thought...

linkerd/proxy/http/src/detect.rs
```diff
@@ -0,0 +1,188 @@
...
+                    // Not a match. Don't check for an HTTP/2 preface again.
+                    maybe_h2 = false;
+                }
+            }
+
+            if maybe_h1 {
+                // Scan up to the first line ending to determine whether the
+                // request is HTTP/1.1. HTTP expects \r\n, so we just look for
+                // any \n to indicate that we can make a determination.
+                if buf[scan_idx..].contains(&b'\n') {
```
hawkw, line 71: :+1: this is much simpler than what you had here yesterday --- and (IIRC) `contains`) should use a fast SIMD `memchr` impl when SIMD instructions are available.

linkerd/proxy/http/src/detect.rs
```diff
@@ -0,0 +1,188 @@
...
+    use tokio_test::io;
+
+    const HTTP11_LINE: &'static [u8] = b"GET / HTTP/1.1\r\n";
+    const H2_AND_GARBAGE: &'static [u8] = b"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\ngarbage";
+    const GARBAGE: &'static [u8] =
+        b"garbage garbage garbage garbage garbage garbage garbage garbage garbage garbage garbage garbage garbage garbage garbage garbage garbage";
+
+    #[tokio::test(flavor = "current_thread")]
+    async fn h2() {
+        let _ = tracing_subscriber::fmt().with_test_writer();
```
hawkw, line 121: this doesn't do anything. you're constructing a subscriber but not setting it as the default, so it doesn't get used. you want:
```suggestion
        let _guard = tracing_subscriber::fmt().with_test_writer().set_default();
```
(which requires the `SubscriberInitExt` trait in scope, imported by `use tracing_subscriber::prelude::*`)
or 
```suggestion
        let _guard = tracing::subscriber::set_default(
            tracing_subscriber::fmt().with_test_writer().finish()
        );
```
(formatting mine; it's possible that rustfmt won't like that)
or 
```suggestion
    let _ = tracing_subscriber::fmt().with_test_writer().try_init();
```
to try to set it as the global default, which might be simplest if isolating these tests doesn't matter.

(sorry for options overload x_x)

linkerd/proxy/http/src/server.rs
```diff
@@ -145,120 +124,72 @@ where
...
+                            shutdown = drain.signal() => {
+                                debug!("The process is shutting down the connection");
+                                Pin::new(&mut conn).graceful_shutdown();
+                                shutdown.release_after(conn).await?;
+                            }
+                            () = closed => {
+                                debug!("The stack is tearing down the connection");
+                                Pin::new(&mut conn).graceful_shutdown();
+                                conn.await?;
+                            }
```
hawkw, line 158: it occurs to me that we probably _could_ stick the `select!` into an `async fn` that takes a drain, closed future, and connection future, so that this isn't repeated...but i dunno if that's worth it here, and it was like that before...

linkerd/proxy/transport/src/detect/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+        let target = self.target.clone();
+        Box::pin(async move {
+            trace!("Starting protocol detection");
+            let t0 = time::Instant::now();
+            let protocol = detect.detect(&mut io, &mut buf).await?;
+            debug!(
+                ?protocol,
+                elapsed = ?(time::Instant::now() - t0),
+                "Detected"
+            );
```
hawkw, line 109: TIOLI: it seems like we _could_ move the timing here into the timeout layer, since it always takes a t0 anyway to generate the error. that way, we could avoid a second `Instant::now()`, which may or may not be a syscall. not sure if it matters here since this is a once-per-connection code path rather than a per-request code path...

hawkw (approved):
This looks good to me overall! 

The way the `tracing` subscriber is set up in the tests is wrong and won't actually do anything, but once that's fixed, this seems good to me.

linkerd/app/integration/src/tests/identity.rs
olix0r: the http detect tests have specific tests for this sort of thing. i think the integration tests should be as simple as possible to test whatever we expect them to tests.

kleimkuhler (approved):
Looks good!

linkerd/proxy/http/src/server.rs
olix0r: let's do this separately -- the type constraints for this function would end up being pretty weighty. And this doesn't trigger the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).

PR #746, opened 2020-11-17
Add a FromStr for dns::Name

olix0r:
We use `TryFrom<&[u8]>` in many cases we can use `FromStr`. This is a
minor cleanup of these type coersions.


hawkw (approved):
<3 <3 <3

PR #748, opened 2020-11-17
outbound: add test for fwding conn errors to HTTP clients (good version)

hawkw:
whoops, i pushed this to the wrong branch --- the test actually
does the correct thing now.


olix0r:
Merged manually. Thanks!

PR #749, opened 2020-11-18
style: fix some random clippy lints

hawkw:
This commit fixes a few clippy lints, primarily the use of `.clone()` on
types which implement `Copy`. I stumbled across this while trying to use
clippy to find instances of passing references to `SocketAddr`s instead
of copying them, which is inefficient...but it turns out that we don't
actually do that anywhere...


linkerd/dns/name/src/name.rs
```diff
@@ -12,7 +12,7 @@ pub struct InvalidName;
 
 impl Name {
     pub fn is_localhost(&self) -> bool {
-        *self == Name::try_from("localhost.".as_bytes()).unwrap()
+        *self == Name::try_from(&b"localhost."[..]).unwrap()
```
olix0r, line 15: could be `from_str` now

olix0r (approved):

linkerd/dns/name/src/name.rs
hawkw: yeah --- the clippy lint was specifically about using byte string literals rather than `.as_bytes()`.

PR #750, opened 2020-11-23
errors: set `content-type` for synthesized grpc errors

hawkw:
The proxy will synthesize responses for some error conditions.
Currently, there's special logic for synthesizing errors for gRPC
requests: we emit `grpc-status` and `grpc-message` trailers rather than
HTTP status codes for gRPC requests. However, we *don't* currently set a
`content-type` header for gRPC error responses. This makes some clients
angry.

This commit adds a `content-type: application/grpc` header to
synthesized error responses for gRPC requests.

Fixes linkerd/linkerd2#5273

### DIFF
--- a/linkerd/app/core/src/errors.rs
+++ b/linkerd/app/core/src/errors.rs
@@ -151,7 +151,7 @@ impl<ReqB, RspB: Default + hyper::body::HttpBody>
                 let is_grpc = req
                     .headers()
                     .get(http::header::CONTENT_TYPE)
-                    .and_then(|v| v.to_str().ok().map(|s| s.starts_with("GRPC_CONTENT_TYPE")))
+                    .and_then(|v| v.to_str().ok().map(|s| s.starts_with(GRPC_CONTENT_TYPE)))
                     .unwrap_or(false);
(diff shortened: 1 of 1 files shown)


linkerd/app/core/src/errors.rs
```diff
@@ -149,7 +151,7 @@ impl<ReqB, RspB: Default + hyper::body::HttpBody>
                 let is_grpc = req
                     .headers()
                     .get(http::header::CONTENT_TYPE)
-                    .and_then(|v| v.to_str().ok().map(|s| s.starts_with("application/grpc")))
+                    .and_then(|v| v.to_str().ok().map(|s| s.starts_with("GRPC_CONTENT_TYPE")))
```
olix0r, line 154: mmm probably don't want to quote it?

hawkw: agh, thanks for catching that
```suggestion
                    .and_then(|v| v.to_str().ok().map(|s| s.starts_with(GRPC_CONTENT_TYPE)))
```

olix0r (approved):

PR #751, opened 2020-11-23
concurrency-limit: Drop permit on readiness

olix0r:
The concurrency-limit module holds permits until the response future is
dropped. This is fine in theory, but we have reports of clients hitting
the concurrency limit unexpectedly.

This change ensures that the permit is released immediately as soon as
the inner future becomes ready, regardless of how long the response
future is held. This also adds trace logging when the permit is
released.

### DIFF
--- a/linkerd/concurrency-limit/src/lib.rs
+++ b/linkerd/concurrency-limit/src/lib.rs
@@ -7,11 +7,14 @@
 #![deny(warnings, rust_2018_idioms)]
 
 use pin_project::pin_project;
-use std::future::Future;
-use std::pin::Pin;
-use std::sync::Arc;
-use std::task::{Context, Poll};
-use std::{fmt, mem};
+use std::{
+    fmt,
+    future::Future,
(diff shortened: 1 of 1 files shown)


hawkw (approved):
this seems fine to me, though i'd be a bit surprised if we actually _are_ holding onto these futures much longer than when they become ready.  seems worth a shot, though!

if this does turn out to make a difference, it might be worth upstreaming this change to Tower --- the current [concurrency limit implementation](https://github.com/tower-rs/tower/tree/master/tower/src/limit/concurrency) is pretty much identical to ours (i don't currently recall why we have a separate impl? maybe we should try to get on tower's...)

olix0r:
@hawkw the tower implementation creates a semaphore per inner service. We want to share a semaphore across all inner services.

hawkw:
> @hawkw the tower implementation creates a semaphore per inner service. We want to share a semaphore across all inner services.

ah, right, never mind!

PR #752, opened 2020-11-23
Increase the default buffer capacity to 100K

olix0r:
We've received reports that the default in-flight limit may be too
small. Now that the proxy can consume multiple cores, it seems wise to
increase the default from 10K to 100K.

This bound should prevent high-traffic workloads from hitting the
in-flight limit in all but extreme cases, while still protecting the
proxy against infinite growth.


hawkw (approved):
seems fine to me! however, it _might_ be worth trying to get the users who are seeing the concurrency limit reached test just #751 first --- i'd like to have some understanding of what's actually going on, ideally....

PR #753, opened 2020-11-24
Change default max-in-flight and buffer-capacity

olix0r:
There are seperate settings for the default in-flight request limit and
the proxy's buffer capacity, but we've chosen a single default.

This change restores the buffer capacity to 10K while raising the
in-flight request limit to 100K. This is intended to allow the proxy to
support services that do not respond promptly that may legitimitately
have <100K pending requests.


kleimkuhler (approved):
Makes sense!

PR #754, opened 2020-11-27
metrics: Expose process_cpu_seconds_total as a float

olix0r:
Prometheus handles all values as `f64`, but we only expose values as
whole integers. This means that the `process_cpu_seconds_total` metric
only exposes whole second values, while Linux exposes process time in
10ms increments.

This change modifies the `Counter` metric type to store an additional
marker that provides a strategy for converting the stored `u64` value to
`f64` for export.  This strategy is employed so that we can continue to
use `AtomicU64` to back counters and only use floats at export-time.  By
default the unit type is used to convert counters as before, but an
alternate `MillisAsSeconds` strategy is used to expose fractional
seconds from a millisecond counter.

This necessitates changing the histogram buckets to floats as well.
While this change doesn't modify the bucket values, this sets up future
changes to latency metrics.


linkerd/metrics/src/counter.rs
```diff
@@ -14,41 +17,56 @@ use std::sync::atomic::{AtomicU64, Ordering};
...
+}
+
+impl<F: Factor> Counter<F> {
+    pub fn new() -> Self {
+        Self(AtomicU64::default(), std::marker::PhantomData)
+    }
+
     pub fn incr(&self) {
-        self.add(1)
+        self.add(1u64)
```
hawkw, line 37: huh, i'm surprised the explicit type is necessary here...

linkerd/metrics/src/lib.rs
```diff
@@ -31,3 +31,28 @@ macro_rules! metrics {
...
+/// Largest `u64` that can fit without loss of precision in `f64` (2^53).
+///
+/// Wrapping is based on the fact that Prometheus models values as f64 (52-bits
+/// mantissa), thus integer values over 2^53 are not guaranteed to be correctly
+/// exposed.
+const MAX_PRECISE_UINT64: u64 = 0x20_0000_0000_0000;
+
+impl Factor for () {
+    fn factor(n: u64) -> f64 {
+        n.wrapping_rem(MAX_PRECISE_UINT64 + 1) as f64
```
hawkw, line 50: i was kind of wondering if the wrapping module by the max precise value could be factored out so that `Factor` impls don't have to be aware of it...i suppose we could factor the max precise value too and do the modulo in `Counter::value`. 

this is probably not worth the effort.

hawkw (approved):
lgtm! i commented on some nits.

linkerd/app/core/src/telemetry/process.rs
```diff
@@ -77,16 +77,28 @@ mod system {
...
+
+            // On Linux, CLK_TCK is ~always `100`, so pure integer division
+            // works. This is probably not suitable if we encounter other
+            // values.
             let clock_ticks_per_sec = Self::sysconf(libc::_SC_CLK_TCK, "clock ticks per second")?;
+            let ms_per_tick = 1_000 / clock_ticks_per_sec;
+            if clock_ticks_per_sec != 100 {
+                warn!(
+                    clock_ticks_per_sec,
+                    ms_per_tick, "Unexpected value; process_cpu_seconds_total may be inaccurate."
```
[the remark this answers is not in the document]
olix0r: yeah. rustfmt be rustfmting

linkerd/metrics/src/counter.rs
olix0r: Cruft from an intermediate state where `add` took `impl Into<u64>` -- no longer the case so removed the type.

linkerd/metrics/src/lib.rs
olix0r: The intent for doing this inside of factor is so that we can adjust the precision based on the scale of the value:

https://github.com/linkerd/linkerd2-proxy/pull/754/files#diff-36a73a9a9de1e34abac85003560b0aea70e9ebb0f1cc29af3dce703bfbef7a89R56

Now that I look at this again, though, it's probably not really correct; since we may lose some of the precision before scaling it.

olix0r: Actually, the current behavior [seems fine](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4004452323dc29389db7772f97f1a325)

PR #756, opened 2020-11-30
metrics: Support the Summary metric type

olix0r:
Prometheus supports two distribution types: histogram and summary.
Histograms track values over a set of fixed buckets. Summaries, on the
other hand, track fixed quantiles over a sliding time window.

This change introduces a Summary metric type that is backed by
a ring buffer of hdrhistograms. These histograms may either auto-resize
or have a fixed upper bound. The histograms are "rotated" so that old
values expire after a maximum lifetime.


hawkw (commented):
Putting the report (read-side) of the histogram inside a mutex, but recording with a mutable borrow seems a bit iffy to me. If recording requires a mutable borrow, then we need to put the whole summary inside a lock as well. It might make more sense to move a lock on writing into the mutex as well?

Also, incidentally, it occurs to me that the hdrhistogram `add` operation _could_ permit us to make the write path uncontended by giving each thread a local copy of the histograms which are all merged when reading. That way, the histogram is only contended during a report, not on every write. This might be worth considering in a follow-up, since writes are in the request path...

linkerd/metrics/src/summary.rs
```diff
@@ -0,0 +1,386 @@
...
+use parking_lot::{MappedRwLockWriteGuard, Mutex, MutexGuard, RwLock, RwLockWriteGuard};
+use std::fmt;
+use tokio::time;
+
+/// Summarizes a distribution of values at fixed quantiles over a sliding window.
+#[derive(Debug)]
+pub struct Summary<F = ()> {
+    rotate_interval: time::Duration,
+
+    windows: RwLock<Windows>,
```
hawkw, line 15: I'm not sure if this needs to be a `RwLock`, rather than just a `Mutex`. As far as I can tell, recording a value acquires a write lock on the `Windows`, while locking the `report` histogram acquires a read guard on the `Windows`. But, when the read guard is acquired, the `Mutex` around the report histogram is already locked. This means that the read lock is only acquired by one thread at a time, since the mutex around the report must be acquired before the read lock is acquired. Therefore, I don't think there are ever any scenarios where the read lock actually allows multiple readers to access this concurrently.

It might be better to just make this also be a mutex, and avoid the additional overhead of a rwlock?

hawkw (approved):
looks good to me!

i have some thoughts on potentially removing locks from the write path, but this seems solid for now --- i dunno if these will be contended enough for it to matter anyway

PR #760, opened 2020-12-05
Rename DetectHttp to NewServeHttp

olix0r:
In preparation for an upcoming change that decouples HTTP detection
logic from server initialization, this chagne renames `DetectHttp` to
`NewServeHttp` and `AcceptHttp` to `ServeHttp`.


hawkw (approved):
just a name change; seems good to me!

PR #761, opened 2020-12-05
http: more consistent names for body types

hawkw:
HTTP bodies are represented by the `Body` trait in the `HttpBody` crate.
Hyper re-exports this trait as `HttpBody`, and provides its own concrete
type named `Body`, implementing this trait. Previously, Hyper provided
its own trait, named `Payload`.

In the proxy, there are several types implementing `Body`. The naming
scheme for these types is inconsistent: some of them are named `Body`,
and others are named `Payload`. This makes it somewhat unclear what
these names refer to, especially when they show up in compiler errors
like "expected type 'Body', but found type 'Body'." 

This branch renames these types to use a more consistent naming scheme:

* The `Body` type in `linkerd2_proxy_http::glue` is now named
  `UpgradeBody`, since it deals with handling HTTP/1.1 upgrades,
* The `Payload` type in `linkerd2_http_box` is now named `BoxBody`,
  since the "payload" terminology isn't really used elsewhere in
  the proxy codebase, and the new name makes it clearer that this
  is the boxed, type-erased body type,
* The re-export of `glue::Body as Payload` from `linkerd2_proxy_http`
  is removed --- it turns out this wasn't being used anywhere, and the
  renaming just makes things more confusing (IMO).

There should be no functional changes in this PR, just renaming.

### DIFF
--- a/linkerd/app/core/src/svc.rs
+++ b/linkerd/app/core/src/svc.rs
@@ -275,7 +275,7 @@ impl<S> Stack<S> {
     // pub fn box_http_request<B>(self) -> Stack<http::boxed::BoxRequest<S, B>>
     // where
     //     B: hyper::body::HttpBody<Data = http::boxed::Data, Error = Error> + 'static,
-    //     S: tower::Service<http::Request<http::boxed::Payload>>,
+    //     S: tower::Service<http::Request<http::boxed::BoxBody>>,
     // {
     //     self.push(http::boxed::request::Layer::new())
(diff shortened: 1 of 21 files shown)


olix0r (approved):
Needs to merge man to pick up the detect rename.

hawkw:
> Needs to merge man to pick up the detect rename.

ugh i _thought_ i branched from the latest main --- should be good now.

PR #763, opened 2020-12-08
cache: Decouple tracker from cache layer

olix0r:
The cache layer takes another layer that is responsible for dropping
each cache entry's handle so that it can be evicted; but now that
there's a single connection-oriented cache on the outbound stack, cache
retention is tied to the time since the last connection and not the
activity of the conncetion.

This change doesn't fix this underlying issue, but it seperates the
tracker layer from the cache layer so that a followup change can modify
the cache retention strategy.


linkerd/cache/src/lib.rs
```diff
@@ -9,53 +12,47 @@ use std::{
...
 type Services<T, S> = Arc<RwLock<HashMap<T, (S, Weak<()>)>>>;
 
 // === impl Cache ===
 
-impl<T, N> Cache<T, N>
+impl<T, N> Cache<T, N, N::Service>
 where
     T: Eq + Hash,
-    N: NewService<(T, Handle)>,
+    N: NewService<(Handle, T)>,
```
olix0r, line 35: This changed because it's (now) idiomatic for services to put their argument on the first side of a tuple.

hawkw (approved):
looks good to me!

PR #764, opened 2020-12-08
WIP: cache: Decouple idle tracking & eviction from the cache layer

olix0r:
The cache relies on the buffer's idle timeout to evict cache entries.
This approach (1) causes us to rely on a specialized Buffer
implementation (rather than using tower's Buffer) and (2) is inflexible,
preventing the outbound connection-oriented cache from `taking into
account connection-level activity.

This change modifies the `cache::NewTrack` middleware to spawn a task
that tracks its service's idleness. When the service hasn't been used
for an idle timeout, the cache handle is dropped so that the cache knows
to clear the appropriate entry.

### DIFF
--- a/linkerd/app/inbound/src/lib.rs
+++ b/linkerd/app/inbound/src/lib.rs
@@ -274,11 +274,11 @@ impl Config {
         // fails, skips that stack to forward to the local endpoint.
         svc::stack(switch_loopback)
             .check_new_service::<Target, http::Request<http::boxed::BoxBody>>()
-            .push_cache_tracker()
+            .push_cache_tracker(cache_max_idle_age)
             .push_on_response(
                 svc::layers()
                     .push_failfast(dispatch_timeout)
-                    .push_spawn_buffer_with_idle_timeout(buffer_capacity, cache_max_idle_age)
+                    .push_spawn_buffer(buffer_capacity)
                     .push(metrics.stack.layer(stack_labels("logical")))
                     .box_http_response(),
             )
--- a/linkerd/cache/src/lib.rs
+++ b/linkerd/cache/src/lib.rs
@@ -10,7 +10,9 @@ use std::{
     hash::Hash,
     sync::{Arc, Weak},
 };
-use tracing::{debug, trace};
+use tokio::sync::Notify;
+use tracing::{debug, debug_span, trace};
+use tracing_futures::Instrument;
 
 #[derive(Clone)]
 pub struct Cache<T, N, S>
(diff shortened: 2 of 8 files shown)


linkerd/cache/src/track.rs
```diff
@@ -1,33 +1,63 @@
...
+        let idle = self.idle;
+        let notify = Arc::new(Notify::new());
+
+        // Create a background task that drops the cache entry's handle when the
+        // service hasn't been notified for an `idle` interval.
+        tokio::spawn({
+            let notify = notify.clone();
+            async move {
+                loop {
+                    tokio::select! {
```
olix0r, line 44: should use select_biased to bias towards resetting

linkerd/cache/src/lib.rs
```diff
@@ -19,32 +21,117 @@ where
...
-        let handle = Arc::new(());
-        let weak = Arc::downgrade(&handle);
-        let svc = new.new_service((handle, target));
+    fn spawn(inner: N) -> (Self, tokio::task::JoinHandle<()>) {
+        let services = Services::default();
+        let evict = Arc::new(Notify::new());
+
+        // Spawn a background task to remove defunct services.
+        let task = tokio::spawn(
+            Self::evict(Arc::downgrade(&evict), services.clone()).instrument(debug_span!("evict")),
```
hawkw, line 71: nit/tioli: could also just stick an `#[instrument(level = "debug", skip(evict, services)]` attribute on `evict`...about the same number of characters i guess...

linkerd/cache/src/lib.rs
```diff
@@ -19,32 +21,117 @@ where
...
+                    trace!("Evicting services");
+                }
+                None => {
+                    debug!("Cache dropped");
+                    return;
+                }
+            }
+
+            // Drop defunct services before inserting the new service into the
+            // cache.
```
hawkw, line 102: Nit: this comment looks like it was copied from a previous impl where eviction happened when inserting a new service.  Now, the eviction is performed in the background, so we might want to update this:
```suggestion
            // Drop defunct services.
```

linkerd/cache/src/track.rs
```diff
@@ -1,33 +1,63 @@
...
+                        }
+                        _ = time::sleep(idle) => {
+                            debug!(idle = idle.as_secs(), "Idled out");
+                            drop(handle);
+                            return;
+                        }
+                    }
+                }
+            }
+            .instrument(debug_span!("tracker"))
```
hawkw, line 56: nit/tioli: _could_ just stick the idle secs in the span:
```suggestion
                            trace!("Resetting");
                        }
                        _ = time::sleep(idle) => {
                            debug!("Idled out");
                            drop(handle);
                            return;
                        }
                    }
                }
            }
            .instrument(debug_span!("tracker", idle = idle.as_secs()))
```

hawkw: nit: incidentally, also, right now this will look like this, supposing the duration is ten seconds:
```
idle=10
```
this doesn't indicate that it's in seconds.

I would either rename the field to `idle_secs` or something, or just use `Duration`'s `fmt::Debug` impl, [which now pretty-prints the duration with a unit](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=22b1d21167f7acd102e7006f942731f1).

linkerd/cache/src/track.rs
```diff
@@ -1,33 +1,63 @@
...
 
-    #[inline]
-    fn new_service(&mut self, (_handle, target): (Handle, T)) -> Self::Service {
+    fn new_service(&mut self, (handle, target): (Handle, T)) -> Self::Service {
+        let idle = self.idle;
+        let notify = Arc::new(Notify::new());
+
+        // Create a background task that drops the cache entry's handle when the
+        // service hasn't been notified for an `idle` interval.
+        tokio::spawn({
```
hawkw, line 40: I was briefly worried about a potential task leak here when the `Track` service is dropped without notifying the background task...but then I realized that in that case, it'll just time out. So this is fine. Might be worth a comment somewhere to remember this in the future?

linkerd/app/inbound/src/lib.rs
```diff
@@ -274,11 +274,11 @@ impl Config {
         // fails, skips that stack to forward to the local endpoint.
         svc::stack(switch_loopback)
             .check_new_service::<Target, http::Request<http::boxed::BoxBody>>()
-            .push_cache_tracker()
+            .push_cache_tracker(cache_max_idle_age)
             .push_on_response(
                 svc::layers()
                     .push_failfast(dispatch_timeout)
-                    .push_spawn_buffer_with_idle_timeout(buffer_capacity, cache_max_idle_age)
+                    .push_spawn_buffer(buffer_capacity)
```
hawkw, line 281: I wonder if removing the idle timeout from buffers means we can just swap this out for `tower-buffer`...might be nice to get rid of the reimplementation. Could be worth doing in a follow-up PR.

hawkw (commented):
this looks like it's on the right track (get it? :D) --- besides your comment about `select_biased!`, i didn't see any major concerns. i commented on a few nits.

hawkw:
it would be really nice to have some tests for this before merging, if it's possible to easily repro the issues with the current eviction behavior? not a blocker if it isn't...

PR #765, opened 2020-12-10
outbound: add stack tests for http

hawkw:
This branch adds some new stack tests for simple HTTP/1 and HTTP/2
meshed and unmeshed outbound requests. These tests also assert that TCP
load balancers are not constructed when HTTP traffic is detected.

The hope was that the panic on constructing TCP balancers would catch a
bug that exists in #744, but that issue appears to be related to the
cache instead. Still, it's probably worth having these tests.

### DIFF
--- a/linkerd/app/outbound/src/http/tests.rs
+++ b/linkerd/app/outbound/src/http/tests.rs
@@ -11,7 +11,13 @@ use linkerd2_app_core::{
     transport::{io, listen},
     Addr, Error,
 };
-use std::{net::SocketAddr, str::FromStr, time::Duration};
+use std::{
+    net::SocketAddr,
+    pin::Pin,
+    str::FromStr,
+    task::{Context, Poll},
+    time::Duration,
+};
 use tokio::time;
 use tower::{Service, ServiceExt};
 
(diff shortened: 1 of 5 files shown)


linkerd/app/outbound/src/http/tests.rs
```diff
@@ -159,3 +200,197 @@ async fn profile_endpoint_propagates_conn_errors() {
...
+
+    // Pretend the upstream is a proxy that supports proto upgrades...
+    let mut server_settings = hyper::server::conn::Http::new();
+    server_settings.http2_only(true);
+    let connect = support::connect().endpoint_fn_boxed(ep1, hello_server(server_settings));
+
+    let profiles = profile::resolver().profile(
+        ep1,
+        profile::Profile {
+            opaque_protocol: false,
```
kleimkuhler, line 252: nit: Just leave this to the `..Default::default()`?

linkerd/io/src/boxed.rs
```diff
@@ -66,7 +66,6 @@ impl std::fmt::Debug for BoxedIo {
         f.debug_struct("BoxedIo").finish()
     }
 }
-
```
kleimkuhler, line 69: Looks like this snuck in here

linkerd/app/outbound/src/http/tests.rs
```diff
@@ -66,6 +81,32 @@ where
...
 }
 
+#[derive(Clone, Debug)]
+struct NoTcpBalancer;
+
+impl svc::NewService<crate::tcp::Concrete> for NoTcpBalancer {
+    type Service = Self;
+    fn new_service(&mut self, target: crate::tcp::Concrete) -> Self::Service {
+        panic!(
+            "no TCP load balancer should be created in this test!\n\ttarget = {:?}",
```
kleimkuhler, line 91: Nice-this seems like a good way to check this 👍

kleimkuhler (approved):
Had some small nits but looks good overall!

olix0r (approved):

alpeb:
image

PR #766, opened 2020-12-10
trace: nicer traces in tests, clean up trace configuration

hawkw:
This branch improves how traces are displayed in tests. In particular,
I've made the following changes:

* When running tests, use a "test writer" that participates in libtest's
  output capturing. Now, traces from tests will be displayed grouped
  together when the test fails, rather than printed to the console as
  soon as they occur. Logs can now be printed for successful tests using
  `cargo test -- --show-output`, and will still be grouped by test.
* Because test traces are now captured, enable a more verbose default
  filter for tests. The logs will no longer be spammed to the console,
  so it's okay to have more verbose logging on by default when tests
  fail.
* Disabled thread IDs in test mode. The tests run single-threaded
  proxies, and the new-style stack tests are completely single threaded.
  Some of the integration tests spawn test clients or servers in
  background threads, but those have their own spans which should make
  it much clearer to figure out where logs came from than just a numeric
  thread ID. Removing IDs makes the log lines a little shorter.

I considered also enabling thread names in the test logs (since the name
of a test thread is the name of the test it's running). I decided not
to, as they end up being quite long since the module path of the test
function is also incldued, and `--show-output` and panic output already
groups the logs by which test output them. However, we could turn thread
names on if we want to.

I also refactored the code for setting up tracing a bit. Hopefully it's
clearer now --- I was able to remove some complexity from how we pass
around the handle for reloading the trace level.

I also bumped some of the more verbose logs from the test support code
down to `trace`, since they get kinda annoying to have them on by
default (particularly the "new service" one).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2896,9 +2896,9 @@ dependencies = [
 
 [[package]]
 name = "tracing-subscriber"
-version = "0.2.14"
+version = "0.2.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2810660b9d5b18895d140caba6401765749a6a162e5d0736cfc44ea50db9d79d"
+checksum = "a1fa8f0c8f4c594e4fc9debc1990deab13238077271ba84dd853d54902ee3401"
 dependencies = [
  "ansi_term",
  "chrono",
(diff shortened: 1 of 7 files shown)


olix0r (approved):
lgtm

PR #767, opened 2020-12-10
Ensure that services are held as long they are being used

olix0r:
Middlewares, especially the cache, may want to use RAII to detect when a
service is idle, but the TCP server and HTTP routers drop services as
soon as the request is dispatched.

This change modifies the TCP server to hold its per-connection service until
all work on that connection is complete. It also introduces a new
`http::Retain` middleware that clones its inner service into repsonse
bodies.

This is necesarry for upcoming cache eviction changes to address
linkerd/linkerd2#5334.

### DIFF
--- a/linkerd/proxy/http/src/retain.rs
+++ b/linkerd/proxy/http/src/retain.rs
@@ -1,5 +1,5 @@
 //! Provides a middleware that holds an inner service as long as responses are
-//! being processed..
+//! being processed.
 
 use linkerd2_stack::layer;
 use pin_project::pin_project;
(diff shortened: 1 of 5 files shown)


linkerd/proxy/http/src/retain.rs
```diff
@@ -0,0 +1,127 @@
+//! Provides a middleware that holds an inner service as long as responses are
+//! being processed..
```
hawkw, line 2: typo:
```suggestion
//! being processed.
```

hawkw (approved):
this change lgtm!

PR #768, opened 2020-12-11
cache: Ensure that actively held services are not evicted

olix0r:
Cache eviction is currently triggered when a service has not processed
new requests for some idle timeout. This is fragile for caches that may
process a few long-lived requests. In practice, we would prefer to only
start tracking idleness when there are no *active* requests on a
service.

This change restructures the cache to return services that are wrapped
with tracking handles, rather than passing a tracking handle into the
inner service. When a returned service is dropped, it spawns a
background task that retains the handle for an idle timeout and, if no
new instances have acquire the handle have that timeout, removes the
service from the cache.

This change reduces latency as well as CPU and memory utilization in
load tests. Furthermore, it ultimately eliminates the need for a specialized
buffer implementation.

Fixes linkerd/linkerd2#5334


linkerd/app/outbound/src/server.rs
```diff
@@ -156,13 +156,12 @@ where
...
-                svc::layers()
-                    .push_failfast(dispatch_timeout)
-                    .push_spawn_buffer_with_idle_timeout(buffer_capacity, cache_max_idle_age),
-            ),
+        .push_on_response(
+            svc::layers()
+                .push_failfast(dispatch_timeout)
+                .push_spawn_buffer(buffer_capacity),
         )
+        .push_cache(cache_max_idle_age)
```
hawkw, line 164: Looks like the retain layer is missing:
```suggestion
        .push_cache(cache_max_idle_age)
        .push_on_response(http::Retain::layer())
```

hawkw (changes requested):

linkerd/app/outbound/src/server.rs
olix0r: This isn't an HTTP stack. the TCP server does the retaining needed to keep the TCP service held until the connection is complete.

olix0r:
Includes the tests from https://github.com/linkerd/linkerd2-proxy/pull/769

linkerd/app/outbound/src/server.rs
```diff
@@ -55,19 +55,62 @@ where
...
-    )
+    );
+    cache(&config.proxy, metrics, accept)
+}
+
+/// Wraps an `N`-typed TCP stack with caching.
+///
+/// Services are cached as long as they are retained by the caller (usually
+/// core::serve) and are evicted from the cache when they have been dropped for
+/// `config.cache_max_idle_age` with no new acquisitions.
```
hawkw, line 75: <3

linkerd/cache/src/lib.rs
```diff
@@ -1,113 +1,258 @@
...
-pub mod layer;
-
-pub use self::layer::CacheLayer;
+use tokio::time;
+use tracing::{debug, debug_span, trace};
+use tracing_futures::Instrument;
 
 #[derive(Clone)]
-pub struct Cache<T, N>
+pub struct Cache<T, N, S>
```
hawkw, line 16: is there any reason for adding a separate type param here rather than keeping it as `N::Service`? i notice that all the impls are for `Cache<T, N, N::Service>`...

hawkw (approved):
after our offline conversation, this definitely seems right to me! :+1:

linkerd/cache/src/lib.rs
olix0r: Probably motivated by an intermediate change. reverted.

PR #769, opened 2020-12-11
outbound: stack tests for idling out services

hawkw:
This branch adds new stack tests in the outbound crate testing behavior
around dropping idle services. I've added a new test utility for
wrapping a `NewService` to track the number of currently alive `Service`
instances created by that `NewService`, using an `Arc`. The tests can
then wrap the HTTP router stack with this wrapper, and make assertions
about how many services are alive.

I had to do a bit of stack surgery to fit the tracking layer in the right place.
Sorry in advance for any merge conflicts.

This branch adds two tests. One asserts that services are dropped after
the idle timeout when they are inactive, and another asserts that
services which are currently streaming response bodies are *not* idled
out.

The second of these tests fails on `main`, which is expected. We expect
PR #768 to fix this.

Depends on #765.

### DIFF
--- a/linkerd/app/outbound/src/server.rs
+++ b/linkerd/app/outbound/src/server.rs
@@ -55,7 +55,7 @@ where
     P::Error: Send,
 {
     let tcp_balance = tcp::balance::stack(&config.proxy, tcp_connect.clone(), resolve);
-    let accept = accept_with_tcp_balancer(
+    let accept = accept_stack(
         config,
         profiles,
         tcp_connect,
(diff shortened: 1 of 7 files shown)


linkerd/app/outbound/src/server.rs
```diff
@@ -55,19 +55,62 @@ where
...
+        metrics.clone(),
         span_sink,
         drain,
-    )
+    );
+    cache_accept(&config.proxy, metrics, accept)
 }
 
-pub fn stack_with_tcp_balancer<P, C, T, H, S, I>(
+pub fn cache_accept<N, S, I>(
```
hawkw, line 71: Open to bikeshed the name of this function (and `accept_with_tcp_balancer`), I had to split these out to sneak in the tracking layer at the right point in the stack.

olix0r: How about just `cache` for this? There's nothing to disambiguate this with, given that it's in the `server` module. And for `accept_with_tcp_balancer`, how about just `stack`?

hawkw:
Closing this as we merged it directly into #768

PR #770, opened 2020-12-12
cache: Only spawn a single task per cache entry

olix0r:
In b0cf2fbb, the proxy moved eviction to be controlled by a background
task; but this scheme spawned a task every time a service was dropped.

This change modifies the cache to only spawn a single task per cached
service and use a notification channel to signal drops. This reduces
memory pressure, especially for HTTP caches.

This change also improves cache logging diagnostics.

### DIFF
--- a/linkerd/cache/src/lib.rs
+++ b/linkerd/cache/src/lib.rs
@@ -116,8 +116,6 @@ where
     type Service = Cached<N::Service>;
 
     fn new_service(&mut self, target: T) -> Cached<N::Service> {
-        let cache = Arc::downgrade(&self.services);
-
         // We expect the item to be available in most cases, so initially obtain
         // only a read lock.
         if let Some((svc, weak)) = self.services.read().get(&target) {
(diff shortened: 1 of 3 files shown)


olix0r:
image

linkerd/cache/src/lib.rs
```diff
@@ -19,72 +19,125 @@ where
...
+        // elapse.
+        loop {
+            futures::select_biased! {
+                // If the reset was notified, restart the timer.
+                _ = reset.notified().fuse() => {
+                    trace!("Reset");
+                }
+                _ = time::sleep(idle).fuse() => match cache.upgrade() {
+                    Some(cache) => match Arc::try_unwrap(reset) {
+                        // If this is the last reference to handle after the
```
hawkw, line 99: nit:
```suggestion
                        // If this is the last reference to the handle after the
```

linkerd/cache/src/lib.rs
```diff
@@ -19,72 +19,125 @@ where
...
+                        // idle timeout, remove the cache entry.
+                        Ok(_) => {
+                            let removed = cache.write().remove(&target).is_some();
+                            debug_assert!(removed, "Cache item must exist: {:?}", target);
```
hawkw, line 103: note that since the `JoinHandle` is dropped when spawning this task, this panic might not be propagated past the task boundary. Since it's a debug assertion, that's probably fine — it will still fail our tests as long as they're run on the single-threaded runtime. Just a note for the future.

hawkw (approved):
looks good to me — i had a couple minor things to point out, but overall, this looks great!

PR #773, opened 2020-12-15
Add a protocol label to stack metrics

olix0r:
Now that TCP stacks are getting more complex, it's worth differentiating
them with a protocol label. This change also renames 'source' stacks to
'server' stacks.


hawkw (approved):
lgtm!

PR #777, opened 2020-12-16
Ensure each crate can be built indpendently

olix0r:
Some subcrates have unneeded or outdated dependencies. This change
updates most of them. The opencensus and trace-context traits should be
updated in a followup.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -28,7 +28,7 @@ jobs:
       - run: make lint
 
   check:
-    timeout-minutes: 5
+    timeout-minutes: 15
     runs-on: ubuntu-18.04
     container:
       image: docker://rust:1.47.0-buster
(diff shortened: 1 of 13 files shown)


hawkw (approved):
<3

i wonder if we want to have some kind of CI step that checks this? like
```bash
for crate in linkerd/*/; do 
    (cd "$crate" && cargo check)
done
```
or something

olix0r:
@hawkw I tried to add that to the Makefile but my make-fu isn't strong enough. Would happily take a PR that does that, though.

hawkw:
> @hawkw I tried to add that to the Makefile but my make-fu isn't strong enough. Would happily take a PR that does that, though.

i'm pretty sure your make-fu is far more powerful than mine...i would solve this by adding a bash script that the Makefile runs :P

PR #779, opened 2020-12-16
Apply clippy to all targets, including tests

olix0r:
There are some lints that are obfuscated by macros, so the integration
tests have some blanket allows.


linkerd/app/integration/src/lib.rs
```diff
@@ -3,6 +3,8 @@
 #![deny(warnings, rust_2018_idioms)]
 #![recursion_limit = "256"]
 #![type_length_limit = "16289823"]
+// It's not clear where this originates.
```
hawkw, line 6: 🙃 someday we really need to at least break up the giant macros into per-test ones, this has gotten pretty unsustainable...

hawkw (approved):

PR #781, opened 2020-12-17
transport: expire idle TCP metrics

hawkw:
The proxy now exposes transport metrics with per-endpoint metadata; but,
unlike HTTP metrics, these metrics are never dropped. So an application
that has communicated with a now-defunct pod will continue to expose
metrics about that pod indefinitely.

This branch factors out the existing code for expiring idle HTTP metrics
so that it can be used for transport metrics as well, and adds eviction
for transport metrics. 

I also fixed a couple issues in the `linkerd2_metrics::summary` module
that were making the crate fail to compile with `--all-features`.

Closes linkerd/linkerd2#5372.


linkerd/metrics/src/store.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    pub fn entry(&mut self, key: K) -> hash_map::Entry<'_, K, Arc<V>> {
+        self.inner.entry(key)
+    }
+
+    pub fn get_or_default(&mut self, k: K) -> &Arc<V>
+    where
+        V: Default,
+    {
+        self.inner.entry(k).or_insert_with(|| Default::default())
```
olix0r, line 55: clippy's unhappy with this

hawkw:
```suggestion
        self.inner.entry(k).or_default()
```

olix0r (approved):
Nice solution! looks great

PR #783, opened 2020-12-17
[WIP] reduce metrics lock contention, faster timers

hawkw:


olix0r:
this branch is stale. we can reopen it if we get better data around this change

PR #784, opened 2020-12-19
Remove weights from endpoint metadata

olix0r:
Endpoint weights are unused. This change removes weights from the
metadata struct.


hawkw (approved):
🚫🏋️‍♂️👍

PR #785, opened 2020-12-19
Transport opaque connections over mTLS

olix0r:
The proxy supports transporting "opaque" TCP streams, but it cannot do
so with mTLS. Because we cannot perform protocol detection (including
mTLS discovery) on server-first or otherwise idle TCP streams, we have
no reliable way to instrument mTLS on these connections.

This change leverages a new discovery API that annotates an
`OpaqueTransport` hint for targets that are marked opaque but have a
Linkerd proxy. This hint includes the inbound port of the target proxy
and, when specified, configures the outbound proxy to:

1. Connect directly to the inbound port (instead of the original target
   port).
2. Write an "opaque transport header"--a special protocol marker and a
   length-delimited protobuf message including the original target port.

This allows the inbound proxy to perform TLS discovery for these
streams; and then the inbound proxy is able to route the connection to
the proper application port as informed by the connection header.

This change, effectively, wraps arbitrary, opaque TCP streams in a
client-first TCP protocol so that inbound proxies can perform mTLS
detection.

This change does **not** extend the proxy's gateway mode to transport
these connections. This change will be done in a followup.


olix0r:
I've tested this manually (against https://github.com/linkerd/linkerd2/pull/5421) and confirmed that it works as expected. This feature definitely needs tests, though...

linkerd/opaque-transport/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+    pub fn encode_prefaced(&self, buf: &mut BytesMut) -> Result<(), Error> {
+        buf.reserve(PREFACE_AND_SIZE_LEN);
+        buf.put(PREFACE);
+
+        debug_assert!(buf.capacity() >= 4);
+        // Safety: These bytes must be initialized below once the message has
+        // been encoded.
+        unsafe {
+            buf.advance_mut(4);
+        }
```
hawkw, line 76: i think i mentioned this on a prior PR, but i'm not convinced that there's an appreciable amount of overhead from just writing `[0, 0, 0, 0]` instead, so idk if this is worth an unsafe block...but, it's not a huge deal.

linkerd/opaque-transport/src/lib.rs
```diff
@@ -0,0 +1,287 @@
...
+        self.encode(buf)?;
+
+        // Once the message length is known, we back-fill the length at the
+        // start of the buffer.
+        let len = buf.len() - PREFACE_AND_SIZE_LEN;
+        assert!(len <= std::u32::MAX as usize);
+        {
+            let mut buf = &mut buf[PREFACE.len()..PREFACE_AND_SIZE_LEN];
+            buf.put_u32(len as u32);
+        }
```
hawkw, line 87: hmm, actually...isn't all this logic just `prost`'s [`Message::encode_length_delimited`](https://docs.rs/prost/0.6.1/prost/trait.Message.html#method.encode_length_delimited) method? is there a reason we're not just using that?

linkerd/opaque-transport/src/lib.rs
hawkw, line 76: I'm not actually sure if the backfilling is necessary --- even if we don't use [`Message::encode_length_delimited`](https://docs.rs/prost/0.6.1/prost/trait.Message.html#method.encode_length_delimited) as I mentioned below, I think we could use [`Message::encoded_len`](https://docs.rs/prost/0.6.1/prost/trait.Message.html#tymethod.encoded_len) to get the length of the message _prior_ to encoding it?

hawkw (commented):
Overall, this looks like the right approach. I noticed a few things that we might want to address --- in particular, I think the code for encoding and decoding the connection header could be a bit simpler.

Also, do we want to add metrics labels for opaque transport connections or something?

linkerd/app/outbound/src/server.rs
```diff
@@ -185,70 +195,68 @@ where
...
     let http = svc::stack(http::NewServeHttp::new(h2_settings, http_server, drain))
-        .push(svc::stack::NewOptional::layer(tcp_balance))
+        .push(svc::stack::NewOptional::layer(
+            // When an HTTP version cannot be detected, we fallback to a logical
+            // TCP stack. This service needs to be buffered so that it can be
+            // cached and cloned per connection.
+            svc::stack(tcp_balance.clone())
+                .push_map_target(tcp::Concrete::from)
+                .push(profiles::split::layer())
+                .push_switch(tcp::Logical::should_resolve, tcp_forward.clone())
```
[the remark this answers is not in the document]
olix0r: Won't compile, though. The stack binds the connection type.

linkerd/opaque-transport/src/lib.rs
olix0r: The length encoded by this is itself [variable-length](https://docs.rs/prost/0.6.1/src/prost/message.rs.html#70); and so it complicates the read-side since we don't know how much data we need to buffer to decode the length. We want our length field to be a fixed size.

linkerd/opaque-transport/src/lib.rs
olix0r: Again, this is uses a varint, which we dont' want to use.

linkerd/opaque-transport/src/lib.rs
hawkw: Ah, yup, I see. Might be worth a comment explaining that.

linkerd/app/outbound/src/ingress.rs
```diff
@@ -125,6 +125,7 @@ where
         .into_inner();
 
     let tcp = svc::stack(tcp)
+        .push_on_response(drain::Retain::layer(drain.clone()))
```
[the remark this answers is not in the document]
olix0r: Yeah, this was missed in the prior change that adds the retain layer.

linkerd/opaque-transport/src/lib.rs
hawkw: `encoded_len` returns a `usize`, so we could just assert it's less than `u32::max`, right? The varint encoding happens in `encode_length_delimited`; AFAICT we should still be able to use `encoded_len`?

linkerd/app/outbound/src/ingress.rs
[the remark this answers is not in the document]
hawkw: :+1: just checking!

linkerd/app/outbound/src/server.rs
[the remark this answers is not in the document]
hawkw: ah, okay, nevermind!

linkerd/opaque-transport/src/lib.rs
olix0r: FWIW, this all comes directly from tonic's message encoding https://github.com/hyperium/tonic/blob/1b03ece2a81cb7e8b1922b3c3c1f496bd402d76c/tonic/src/codec/encode.rs#L54-L67. I'll look at using encoded_len though.

hawkw: oh, huh, i wonder why tonic does that. well, it could be a bit nicer if `encoded_len` works, but if there's some other reason we can't use it, nbd...

hawkw (approved):
looks good to me!

olix0r:
> Also, do we want to add metrics labels for opaque transport connections or something?

Potentially. It seems useful for diagnostics but also necessitates a larger change (especially on inbound target types). Probably reasonable for a followup...

PR #786, opened 2020-12-20
Ensure all forwarded TCP streams keep the proxy running

olix0r:
The proxy is not intended to shutdown until all in-flight responses and
TCP streams are complete. This is enforced by the `ServeHttp`
middleware, which confusingly handles some TCP forwarding in addition to
HTTP server initialization. But there are HTTP streams that are not
instrumented by this middleware and will therefore be abruptly
interrupted when a shutdown signal is received.

This change modifie the `ServeHttp` middleware to only handle HTTP
streams. a `NewOptional` middleware is introduced to conditionally build
a TCP stack when no HTTP type has been detected; and a `drain::Retain`
middleware is added to prevent the drain signal from being released
until TCP streams are completed.


hawkw:
> This change modifie the `ServeHttp` middleware to only handle HTTP
> streams.

<3 <3 <3 <3

linkerd/drain/src/retain.rs
```diff
@@ -0,0 +1,70 @@
+use crate::Watch;
+use linkerd2_stack::{layer, NewService};
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+/// Holds a drain::Watch for as long as a request is pending.
```
hawkw, line 9: Nit: should this be "creates services which ..." or something?

linkerd/drain/src/retain.rs
```diff
@@ -0,0 +1,70 @@
...
+#[derive(Clone, Debug)]
+pub struct NewRetain<N> {
+    inner: N,
+    drain: Watch,
```
hawkw, line 13: since this *also* holds the `Watch`, it will keep the drain from completing as long as a stack with a `NewRetain` in it exists. This may not be an issue if signalling a drain also cancels the tasks which hold these stacks (which I think it will), but this might be worth being aware of for the future...

linkerd/proxy/http/src/server.rs
```diff
@@ -60,137 +55,113 @@ impl<F, H> NewServeHttp<F, H> {
...
+                    tokio::select! {
+                        res = &mut conn => {
+                            debug!(?res, "The client is shutting down the connection");
+                            res?
+                        }
+                        shutdown = drain.signal() => {
+                            debug!("The process is shutting down the connection");
+                            Pin::new(&mut conn).graceful_shutdown();
+                            shutdown.release_after(conn).await?;
+                        }
```
hawkw, line 134: it seems like it *could* be possible to factor this out into a drain layer that's a superset of this logic (actually do a graceful shutdown) and `NewRetain` (just hold the shutdown alive). but, there would basically just be two of these (one for `ServeHttp` and the other for TCP), so it's probably not worth it. it just seems weirdly asymmetrical for the TCP stack to use a separate layer for drains and the HTTP stack to have it tightly coupled.

please feel free to ignore this one, it's not important! :)

linkerd/stack/src/optional.rs
```diff
@@ -0,0 +1,40 @@
...
+#[derive(Clone, Debug, Default)]
+pub struct NewOptional<A, B> {
+    new_some: A,
+    new_none: B,
+}
+
+// === impl NewOptional ===
+
+impl<A, B> NewOptional<A, B> {
+    /// Creates a new `ServeHttp`.
```
hawkw, line 14: nit: i think this comment needs to be updated...

linkerd/stack/src/optional.rs
```diff
@@ -0,0 +1,40 @@
+use crate::{layer, Either, NewService};
+
+/// A stack middleware that takes an optional target type and builds an `A`-typed
+/// stack if the target is set and a `B`-typed stack if it is not.
+#[derive(Clone, Debug, Default)]
+pub struct NewOptional<A, B> {
+    new_some: A,
+    new_none: B,
+}
```
hawkw, line 9: Isn't this just `MakeSwitch` with a predicate like this?
```rust
MakeSwitch::new(new_some, new_none, |&(ref t, _)| t.is_some())
```

It's not a big deal to me, but it seems like we could just use the existing `MakeSwitch` layer for this (it already has a `NewService` impl). If you think it's worth having a separate `NewOptional` type, it could just be a wrapper around `MakeSwitch`. 

The code is simple enough that it's not a *huge* deal either way, just a thought.

hawkw (approved):
this looks good to me! i had some style thoughts.

hawkw:
also: it would be really nice to have a test (using the new stack testing stuff?) that reproduces the bug. not a blocker, though.

linkerd/stack/src/optional.rs
olix0r: no because it unwraps the option -- the inner services are NewService<(T, U)> and NewService<U> instead of both being NewService<(Option<T>, U)>, which is what would be needed for MakeSwitch

hawkw: oh, yup, i missed that --- this does make sense as a separate thing, then

PR #787, opened 2020-12-22
identity: add a metric tracking when the proxy's cert expires

hawkw:
This branch adds a new `identity_cert_expiration_timestamp_seconds` to
the proxy's Prometheus metrics. When identity is enabled, this metric
will report the time at which the proxy's current certificate expires,
in seconds since the Unix epoch.

This is similar to the similar metrics exported by `cert-manager`.

Closes linkerd/linkerd2#5379


kleimkuhler (approved):
Looks good! Tested and works as expected as well 👍

```
..
# HELP certificate_expiration_timestamp_seconds Time when the this proxy's current TLS certificate will expire (in seconds since the UNIX epoch)
# TYPE certificate_expiration_timestamp_seconds gauge
certificate_expiration_timestamp_seconds 1608737171
```

olix0r:
What do you think about using the metric name `identity_expiration_timestamp_seconds`?

* In Linkerd, we don't have a "CA", we have an identity service. I think it's clearer to align these terms within the project.
* It's probably preferable that we differentiate our metrics from those of cert-manager.
* From the [prometheus guide](https://prometheus.io/docs/practices/naming/):
  > [metric names] should have a (single-word) application prefix relevant to the domain the metric belongs to.
  
  "identity" more clearly describes the domain -- there could be multiple types of certificates, in theory

olix0r:
Additionally, it would probably be useful to maintain a counter of identity updates; doesn't necessarily have to be in this PR, but seems like a helpful diagnostic (e.g., it's possible to update the identity and get the same expiration as before)

hawkw:
@olix0r yup, sounds good to me! I'll make the name change now, and can add a counter in a follow-up?

hawkw:
@olix0r so...this is admittedly splitting hairs, and please feel free to ignore me, but it occurs to me that it isn't really  the _identity_ that expires and gets updated. The proxy's identity is immutable, it's the service account of the pod it's injected into. It's the certificate that was _signed_ to certify that the  proxy has that identity.

The reason I bring this up is just that it occurred to me that if we also add an `identity_update_count` metric (or similar), that kind of incorrectly implies that the identity of a proxy _changes_...

olix0r:
@hawkw `identity_cert_...` ?

hawkw:
> @hawkw `identity_cert_...` ?

yup, done!

linkerd/proxy/identity/src/metrics.rs
```diff
@@ -0,0 +1,53 @@
...
+                .map_err(|error| {
+                    tracing::warn!(%error, "an identity would expire before the beginning of the UNIX epoch, something is probably wrong");
+                    fmt::Error
+                })?
+            } else {
+                return Ok(());
+            }
+        } else {
+            return Ok(());
+        };
```
olix0r, line 47: tioli: returning error here will also prevent other reports from being formatted. Would it be better to panic? Or perhaps we should factor this all into a `fn` that returns `Option<Duration>` and then make the format body `if let Some(d) = self.expiry() { ... }`

olix0r (approved):

linkerd/proxy/identity/src/metrics.rs
hawkw: > tioli: returning error here will also prevent other reports from being formatted.

we're returning `Ok` here, so it should continue formatting any other reports?

> Would it be better to panic? Or perhaps we should factor this all into a `fn` that returns `Option<Duration>` and then make the format body `if let Some(d) = self.expiry() { ... }`

yeah, i wasn't sure whether it was better to panic or not. On one hand, this definitely shouldn't ever actually happen, but on the other hand, I dunno if it's a good idea to bring down the whole proxy if one metric can't be formatted...it's not _technically_ a critical error at all. on the *third* hand, if we ever see an expiration timestamp that's before the Unix epoch, something else is probably going wrong...

olix0r: > we're returning Ok here, so it should continue formatting any other reports?

error returned here https://github.com/linkerd/linkerd2-proxy/pull/787/files/b324a9b1693c3af86c9abcf919b277b81d7bd25a#diff-dbf40c2f9158dbeb59c53424a494292110865ee6d84bb197ab3f515afb644812R41

PR #789, opened 2020-12-22
opencensus: rewrite span exporter using async/await

hawkw:
This branch rewrites the `linkerd2_opencensus` crate's `SpanExporter` to
use async/await rather than a manual `Future` implementation.

The new code is a bit clearer and easier to read, and may also be
slightly more efficient, since it can use Tokio 1.0's `mpsc` without
allocating to poll. Additionally, the new code adds a timer after which
any batched spans are flushed, thanks to @olix0r.


linkerd/opencensus/src/lib.rs
```diff
@@ -1,226 +1,134 @@
...
-        node: &mut Option<Node>,
-        metrics: &mut Registry,
+    async fn run(mut self) {
+        'reconnect: loop {
+            let (request_tx, request_rx) = mpsc::channel(1);
+            let mut svc = TraceServiceClient::new(self.client.clone());
+            let req = grpc::Request::new(request_rx);
+            trace!("Establishing new TraceService::export request");
+            self.metrics.start_stream();
+            let mut rsp = Box::pin(svc.export(req));
```
olix0r, line 66: Why is this box necessary?

linkerd/opencensus/src/lib.rs
```diff
@@ -1,226 +1,134 @@
...
+            let mut drive_rsp = true;
+            loop {
+                tokio::select! {
+                    res = &mut rsp, if drive_rsp => match res {
+                        Ok(_) => drive_rsp = false,
```
olix0r, line 71: nit. this is probably just me but I find it easier to read this if the assigment is its own statement

```suggestion
                        Ok(_) => {
                            drive_rsp = false;
                        }
```

olix0r (commented):
This is generally good, but I have to admit that I find the control flow with nested loops to be pretty difficult to follow. I have a feeling that it can be made simpler, but i'd need to give it a try myself to know what I'm actually talking about.

linkerd/opencensus/src/lib.rs
hawkw: we could probably just put an `Unpin` bound on the response future instead.

hawkw:
> I have to admit that I find the control flow with nested loops to be pretty difficult to follow

Yeah, this is probably because I tried to translate the behavior of the previous implementation's state machine futures, which was surprisingly complex. I can add some comments?

hawkw:
Got rid of one of the innermost loops, hopefully this is a bit clearer now?

olix0r:
Took a stab at trying to make things clearer: https://github.com/linkerd/linkerd2-proxy/pull/790/. The primary difference with this approach is that we don't try to process any spans until the client has an open stream from the server. I think this is functionally equivalent because we'd only be able to process one batch while disconnected, anyway.

hawkw:
@olix0r okay, i've merged your changes and rebased onto main, this should be ready for a review!

olix0r (approved):

PR #790, opened 2020-12-23
Simplify opencensus client control flow

olix0r:


linkerd/opencensus/src/lib.rs
```diff
@@ -56,87 +61,111 @@ where
...
+        let Self {
+            client,
+            node,
+            mut spans,
+            mut metrics,
+        } = self;
+
+        // Holds the batch of pending spans. Cleared as the spans are flushed.
+        // Contains no more than MAX_BATCH_SIZE spans.
+        let mut accum = Vec::new();
```
hawkw, line 74: nit/TIOLI: We could just preallocate to `MAX_BATCH_SIZE` here so the vec never gets resized?

linkerd/opencensus/src/lib.rs
```diff
@@ -56,87 +61,111 @@ where
...
-                            tracing::debug!(%error, "response future failed, sending a new request");
-                            continue 'reconnect;
-                        }
-                    },
-                    res = self.send_batch(&request_tx) => match res {
-                        Ok(false) => return, // The span strean has ended --- the proxy is shutting down.
-                        Ok(true) => {} // Continue the inner loop and send a new batch of spans.
-                        Err(()) => continue 'reconnect,
+            let (tx, rx) = mpsc::channel(1);
+            match svc.export(grpc::Request::new(rx)).await {
```
hawkw, line 80: I don't think this is right --- the future returned by `export` finishes when the RPC is *completed*. a lot of the complexity from the previous impl came from the need to drive this future concurrently with exporting spans --- i think if we just `await` this, the stream will have closed by the time we actually start writing spans to the stream, and no spans will ever be exported.

linkerd/opencensus/src/lib.rs
```diff
@@ -56,87 +61,111 @@ where
...
+                res = spans.next().fuse() => match res {
+                    Some(span) => {
+                        trace!(?span, "Adding to batch");
+                        accum.push(span);
+                    }
+                    None => return Err(()),
+                },
+                // Don't hold spans indefinitely. Return if we hit an idle
+                // timeout and spans have been collected.
+                _ = time::sleep(Self::MAX_BATCH_IDLE).fuse() => {
```
hawkw, line 162: :+1:

hawkw (changes requested):
overall, this is definitely a lot clearer, and i think adding an idle flush is a great idea. however, i don't believe this code, as written, will work correctly.

currently, this change would have us wait until the `export` future completes to start sending spans. but that future completes when the RPC has *finished* --- so we're essentially waiting for the channel to have closed before we start sending spans to it. i believe that we'll just constantly start new RPCs in a loop without ever actually sending anything.

i think driving the future concurrently with the streaming of new span batches in a select, as we did previously, was actually pretty load-bearing. we can probably put that back while keeping most of the cleanup in this PR...

linkerd/opencensus/src/lib.rs
olix0r: @hawkw updated; PTAL

hawkw (approved):
okay, i *think* this looks right, thank you! it would be really good to have proper tests for this...

PR #791, opened 2020-12-24
Include scope names in failfast error messages

olix0r:
When a `FailFast` error is triggered, we have no way to know from which stack
this error originates.

This change modifies the `FailFast` module to accept a `scope` name to
be included in error log messages.

Furthermore, this change removes the `FailFastLayer` type in favor of a
simple `FailFast::layer` helper. It also removes the `push_failfast`
stack helper as it doesn't really help us (and the layer type is no
longer nameable).


hawkw (approved):
looks good to me!

this seems like something that, theoretically, could be achieved in a generic way with the [`tracing-error`(https://docs.rs/tracing-error/0.1.2/tracing_error/) crate. but, this seems like a good, simple solution for failfast in particular, so i'm +1 on it!

PR #793, opened 2020-12-24
Add a NewServeHttp::layer helper

olix0r:
Now that `NewServeHttp` takes a single inner stack, we can instrument as
a simple layer.

This change replaces the modules' constructor with a layer and
consolidates the inbound HTTP server stack.


linkerd/app/inbound/src/lib.rs
```diff
@@ -349,20 +349,29 @@ impl Config {
             ..
         } = self.proxy.clone();
 
-        // Handles requests as they are initially received by the proxy.
-        let http_admit_request = svc::layers()
```
hawkw, line 353: nit/tioli: i kind of liked having this factored out, but nbd...

hawkw (approved):
cool, seems good!

linkerd/app/inbound/src/lib.rs
olix0r: that it would split out helped hide the fact that we had redundant failfast modules  -- one in the admit layers and another inside the buffer. It's a bit easier to reason about as one big list...

PR #809, opened 2020-12-26
Simplify tls::client module

olix0r:
The tls::client module uses an old-style manual future implementation,
which makes it difficult to read.

This change replaces the `client::ConnectLayer` type with an anonymous
helper function and replaces the manual future state machine with an
async block.

### DIFF
--- a/linkerd/proxy/transport/src/tls/client.rs
+++ b/linkerd/proxy/transport/src/tls/client.rs
@@ -1,15 +1,18 @@
-use crate::io::BoxedIo;
-use futures::TryFuture;
-use linkerd2_conditional::Conditional;
+use super::Conditional;
+use crate::io::{self, BoxedIo};
+use futures::{
+    future::{Either, MapOk},
+    prelude::*,
+};
 use linkerd2_identity as identity;
-use pin_project::pin_project;
+use linkerd2_stack::layer;
 pub use rustls::ClientConfig as Config;
-use std::future::Future;
-use std::io;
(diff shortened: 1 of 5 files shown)


linkerd/proxy/transport/src/tls/client.rs
```diff
@@ -1,141 +1,84 @@
...
-                ConnectStateProj::Handshake(fut) => {
-                    let io = futures::ready!(fut.poll(cx))?;
-                    trace!("established TLS");
-                    return Poll::Ready(Ok(Connection::new(io)));
-                }
-            };
-        }
+        let tls = match self.local.clone() {
+            Conditional::Some(l) => tokio_rustls::TlsConnector::from(l.tls_client_config()),
+            Conditional::None(reason) => {
```
hawkw, line 62: super minor nit: we don't need to clone this in the case where identity is disabled, this could be
```suggestion
        let tls = match self.local {
            Conditional::Some(ref l) => tokio_rustls::TlsConnector::from(l.clone().tls_client_config()),
            Conditional::None(ref reason) => {
```
but, this probably doesn't make a meaningful difference, so who cares...

hawkw (approved):

linkerd/proxy/transport/src/tls/client.rs
olix0r: Yeah, but in that case cloning the reason is cheap/trivial (it's just a marker type with no data).

PR #810, opened 2020-12-27
Simplify drain module implementation

olix0r:
The drain module can be updated to take advantage of the new `Notify`
primitive, replacing the more costly `watch` channel.

Before doing this, we update the module's tests to avoid manual future
polling. This change also updates method names for clarity,
eliminates an unnecessary manual future implementation, and eliminates
unnecessary crate dependencies.

The `drain::Retain` service is now gated behind a feature flag so the
tower and stack dependencies are optional. This is mostly helpful for
external crates that take a Git dependency on the drain crate.

### DIFF
--- a/linkerd/drain/src/lib.rs
+++ b/linkerd/drain/src/lib.rs
@@ -22,9 +22,6 @@ pub fn channel() -> (Signal, Watch) {
 type Never = std::convert::Infallible;
 
 /// Send a drain command to all watchers.
-///
-/// When a drain is started, this returns a `Drained` future which resolves
-/// when all `Watch`ers have been dropped.
 #[derive(Debug)]
 pub struct Signal {
     drained_rx: mpsc::Receiver<Never>,
@@ -32,8 +29,6 @@ pub struct Signal {
 }
 
 /// Watch for a drain command.
-///
-/// This wraps another future and callback to be called when drain is triggered.
 #[derive(Clone, Debug)]
 pub struct Watch {
     drained_tx: mpsc::Sender<Never>,
--- a/linkerd/drain/src/lib.rs
+++ b/linkerd/drain/src/lib.rs
@@ -53,11 +53,8 @@ pub struct ReleaseShutdown(mpsc::Sender<Never>);
 // === impl Signal ===
 
 impl Signal {
-    /// Start the draining process.
-    ///
-    /// A signal is sent to all futures watching for the signal. A new future
-    /// is returned from this method that resolves when all watchers have
-    /// completed.
+    /// Asynchronously signals all watchers to begin draining and waits for all
+    /// handles to be dropped.
     pub async fn drain(mut self) {
         // Update the state of the signal watch so that all watchers are observe
         // the change.
--- a/linkerd/drain/src/lib.rs
+++ b/linkerd/drain/src/lib.rs
@@ -5,18 +5,28 @@ mod retain;
 
 #[cfg(feature = "tower")]
 pub use crate::retain::Retain;
-use std::future::Future;
-use tokio::sync::{mpsc, watch};
+use std::{
+    future::Future,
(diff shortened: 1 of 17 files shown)


linkerd/drain/src/lib.rs
```diff
@@ -1,95 +1,97 @@
...
     /// Start the draining process.
     ///
     /// A signal is sent to all futures watching for the signal. A new future
     /// is returned from this method that resolves when all watchers have
     /// completed.
-    pub fn drain(self) -> Drained {
-        let _ = self.tx.send(());
-        Drained {
-            drained_rx: self.drained_rx,
+    pub async fn signal(mut self) {
```
hawkw, line 61: bikesheddy, TIOLI: not sure if I like this name change; this signals the drain _and_ waits for it to complete. I think `drain` captured that more clearly IMO?

linkerd/drain/src/lib.rs
```diff
@@ -1,95 +1,97 @@
...
 
-// ===== impl Signal =====
+// === impl Signal ===
 
 impl Signal {
     /// Start the draining process.
     ///
     /// A signal is sent to all futures watching for the signal. A new future
     /// is returned from this method that resolves when all watchers have
     /// completed.
```
hawkw, line 60: This comment is no longer exactly correct; the behavior here changed very slightly. previously, the drain signal was sent _immediately_ when calling the function, and then a future that waits for the drain to complete is returned. Now that this is an `async fn`, that behavior is slightly different: the drain isn't signalled until the future is awaited. We should either update the comment, or change the method to preserve the old behavior. 

If we wanted to keep the old behavior, we could still use async/await syntax, like so:
```rust
pub fn drain(mut self) -> impl Future<Output = ()> + Send + Sync {        
    // Update the state of the signal watch so that all watchers are observe
    let _ = self.signal_tx.send(());

    async move {
        // Wait for all watchers to release their drain handle.
        match self.drained_rx.recv().await {
            None => {}
            Some(n) => match n {},
        }
    }
}
```
This still signals the drain *before* returning a future that can be awaited.

Alternatively, unless this behavior is relied on (and I don't *think* it is) we can just change the comment. :)

linkerd/drain/src/lib.rs
```diff
@@ -1,95 +1,97 @@
...
-// ===== impl Watch =====
+// === impl Watch ===
 
 impl Watch {
     /// Returns a `ReleaseShutdown` handle after the drain has been signaled. The
     /// handle must be dropped when a shutdown action has been completed to
     /// unblock graceful shutdown.
-    pub async fn signal(mut self) -> ReleaseShutdown {
-        let _ = self.rx.changed().await;
+    pub async fn signaled(mut self) -> ReleaseShutdown {
```
hawkw, line 80: +1 for this name change even if the `Drain` method isn't renamed to `signal` --- in Tokio, Carl suggested a style where an async fn that represents some event/condition has names ending with `ed`, like `Notify::notified`, `Watch::changed`, etc...

linkerd/drain/src/lib.rs
```diff
@@ -114,136 +116,147 @@ impl Watch {
 
 impl ReleaseShutdown {
     /// Releases shutdown after `future` completes.
+    #[inline]
```
hawkw, line 119: i'm not sure if this is meaningful on `async fn`s...i don't think they can get inlined? might be worth asking someone who knows about rustc.

hawkw (changes requested):
I had a few minor notes. The only one that's actually important is that a comment on `Drain::signal` is not quite correct anymore — we should fix that. Besides that, feel free to take or leave my style nits & naming preferences.

hawkw:
CI seems broken?

hawkw (approved):
okay, this lgtm once CI passes!

PR #811, opened 2020-12-27
admin: Add a /shutdown endpoint

olix0r:
Some applications, namely batch jobs, may prefer to manually signal that
the proxy should terminate instead of relying on a SIGTERM. There's no
good way for one container to send signals to another, though; so we
need an admin endpoint to support shutdown.

This change adds a `/shutdown` endpoint to the proxy's admin server that
can be used to cause the proxy to initiate shutdown. Shutdown requests
must come on the loopback interface (i.e. from within the same
Kubernetes pod).


hawkw (approved):
looks good to me!

PR #815, opened 2020-12-31
Simplify stack type constraints

olix0r:
We've grown a bunch of type constraints that don't actually influence
compilation. This change removes unnecessary constraints.


olix0r:
Trivial change. Merged.

PR #817, opened 2020-12-31
Avoid boxing I/O types for optional TLS

olix0r:
Our tls client and server currently box the underlying I/O types so that
TcpStreams and TlsStreams can be returned and used interchangeably.

This change introduces a new `EitherIo` type that implements
`AsyncRead`, `AsyncWrite`, and `PeerAddr` to replace uses of `BoxedIo`
in the data path.

### DIFF
--- a/linkerd/app/core/src/admin/mod.rs
+++ b/linkerd/app/core/src/admin/mod.rs
@@ -36,7 +36,7 @@ pub struct Admin<M> {
 pub struct Accept<M>(Admin<M>, hyper::server::conn::Http);
 
 #[derive(Clone)]
-pub struct Serve<M: FmtMetrics>(tls::accept::Meta, Accept<M>);
+pub struct Serve<M>(tls::accept::Meta, Accept<M>);
 
 pub type ResponseFuture =
     Pin<Box<dyn Future<Output = Result<Response<Body>, Never>> + Send + 'static>>;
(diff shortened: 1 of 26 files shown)


kleimkuhler (approved):
This looks good. My main question is if there is noticeable perf difference with this change? If there is not I don't think that'd be a blocker, but just wondering if there is an underlying benefit to this.

Removing the `Box` I was curious if that would affect build times, but running `build` locally on `main` and this branch I saw no difference.

olix0r:
@kleimkuhler These types of changes tend to have pretty negligible runtime perf impacts. I haven't tested but I'm not expecting it to show up. This change really isn't strictly necessary; and we may want to end up using boxedio in the stack if we see compile-time issues (i haven't so far, as you note); but I do think this change helps to make the tls modules a little clearer, especially on the client side. My thought is that it's probably best to avoid boxing if we can, and it was easier to avoid it for the I/O types than i expected.

kleimkuhler:
> @kleimkuhler These types of changes tend to have pretty negligible runtime perf impacts. I haven't tested but I'm not expecting it to show up. This change really isn't strictly necessary; and we may want to end up using boxedio in the stack if we see compile-time issues (i haven't so far, as you note); but I do think this change helps to make the tls modules a little clearer, especially on the client side. My thought is that it's probably best to avoid boxing if we can, and it was easier to avoid it for the I/O types than i expected.

Great, makes sense to me.

hawkw (approved):
lovely, this should also let us simplify some of the stack test code. i was hoping we could get rid of `BoxedIo`!

i think these types are not really the ones that impact our build performance — they're really only ever 3-4 nested types deep at maximum, while the stack types can be significantly more nested. so we don't *need* type erasure here.

PR #818, opened 2020-12-31
Decouple TLS detection from TCP connections

olix0r:
The TLS accept stack--the `DetectTls` type-is currently coupled to the
`TcpStream` type; but the `TcpStream` type requires an actual OS-level
TCP connection, which isn't ideal for testing.

This change introduces a new trait `tls::accept::Detectable` and an
implementation for `TcpStream`. This will permit us to use alternate
implementations (e.g., for `io::DuplexStream`).

This change also updates the `DetectTls` type name to `NewDetectTls`,
and `AcceptTls` to `DetectTls`, to fit our more recent idioms.


linkerd/proxy/transport/src/tls/accept.rs
```diff
@@ -163,71 +176,74 @@ where
...
+        const NO_TLS_META: PeerIdentity = Conditional::None(ReasonForNoPeerName::NoTlsFromRemote);
+
+        // First, try to use MSG_PEEK to read the SNI from the TLS ClientHello.
+        // Because peeked data does not need to be retained, we use a static
+        // buffer to prevent needless heap allocation.
+        //
+        // Anecdotally, the ClientHello sent by Linkerd proxies is <300B. So a
+        // ~500B byte buffer is more than enough.
+        let mut buf = [0u8; PEEK_CAPACITY];
+        let sz = self.peek(&mut buf).await?;
```
olix0r, line 195: This trait is primarily needed so that we can call `TcpStream::peek` which is a specialized API that is (AFAIK) not exposed by `AsyncRead`.

linkerd/proxy/transport/src/tls/accept.rs
```diff
@@ -27,6 +27,17 @@ pub trait HasConfig {
     fn tls_server_config(&self) -> Arc<Config>;
 }
 
+#[async_trait::async_trait]
+pub trait Detectable {
```
hawkw, line 31: nit/tioli: maybe nice to have some kind of doc comment on this?

hawkw (approved):

PR #819, opened 2020-12-31
Rename transport types to ConnectTcp and BindTcp

olix0r:
This change makes it clearer that these are concrete types and not
abstract traits. It also removes the `svc::connect` helper, as its
overly specialized.

This is intended to setup decoupling TCP connections from the
inbound/outbound stacks so that we can setup the entire stack without
interacting with OS sockets.


hawkw (approved):
looks good to me!

PR #820, opened 2021-01-01
tap: Simplify tap initialization

olix0r:
This change updates the tap module to look more like our other modules.
Instead of building a layer that is passed into stack builders, the tap
module now only produces its registry and stacks are instrumented with
the (renamed) `NewTapHttp::layer` helper.

This change removes the unused MakeService implementation and replaces
the manual future implementation with a boxed async closure.


kleimkuhler (approved):
👍

PR #821, opened 2021-01-01
Group metrics, tap, and tracing in a Telemetry type

olix0r:
This change reduces boilerplate passing of telemetry types to stack
constructors.


olix0r:
In retrospect, I don't think this actually helps much

PR #823, opened 2021-01-02
inbound: Decouple inbound stack from TCP connections

olix0r:
In order to make the inbound stack testable without initializing OS
sockets, this change decouples the inbound stack from its TCP client,
passing in an abstract implementation to `inbound::Config::build`.

This change introduces a new `stack::Fail` module that always fails.
This is used to ensure that connections targeting the inbound server
without an opaque port are not forwarded, removing the need for
endpoint-level loop prevention.

### DIFF
--- a/linkerd/app/src/lib.rs
+++ b/linkerd/app/src/lib.rs
@@ -140,6 +140,11 @@ impl Config {
         let oc_span_sink = oc_collector.span_sink();
 
         let start_proxy = Box::pin(async move {
+            let span = info_span!("outbound");
+            let _outbound = span.enter();
+
+            info!(listen.addr = %outbound_addr, ingress_mode);
+
             let outbound_http = outbound::http::logical::stack(
                 &outbound.proxy,
                 outbound::http::endpoint::stack(
@@ -158,9 +163,6 @@ impl Config {
                 outbound_metrics.clone(),
             );
 
-            let span = info_span!("outbound");
-            let _enter = span.enter();
-            info!(listen.addr = %outbound_addr, ingress_mode);
             let connect = outbound::tcp::connect::stack(
                 &outbound.proxy.connect,
                 outbound_addr.port(),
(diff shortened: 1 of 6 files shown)


linkerd/app/src/lib.rs
```diff
@@ -140,6 +140,11 @@ impl Config {
         let oc_span_sink = oc_collector.span_sink();
 
         let start_proxy = Box::pin(async move {
+            let span = info_span!("outbound");
+            let _outbound = span.enter();
```
hawkw, line 144: in general, we say not to use `Span::enter` in async-await. In this case, it's alright, since the `Entered` guard isn't held across an await point, but that wasn't immediately obvious. Maybe there should be a comment about that? 

Alternatively, we could use `Span::in_scope`, which _ensures_ that the span is not entered across an await point, but that would cause some git churn (by indenting all the code inside the span)...

linkerd/app/src/lib.rs
```diff
@@ -205,23 +207,26 @@ impl Config {
                     .instrument(span.clone()),
                 );
             }
-            drop(_enter);
+            drop(_outbound);
+
+            let span = info_span!("inbound");
+            let _inbound = span.enter();
+            info!(listen.addr = %inbound_addr);
```
hawkw, line 214: As above --- it was a little bit surprising to me that this is actually fine.

hawkw (approved):
It seems like there was a bit of unrelated refactoring, so it took a minute to figure out what parts of the diff were actually interesting. This looks good to me, though!

linkerd/app/src/lib.rs
olix0r: fwiw, there's not actually any use of `.await` in this block -- it's just in an `async` to be lazy.

olix0r: going to hold off on these changes as it incurs more boilerplate than i'd like -- we probably need to take a more intentional pass at organizing these builders once the gateway stuff is resolved.

PR #825, opened 2021-01-06
Rename opaque-transport to transport-header

olix0r:
Transport headers may be used for non-opaque traffic, i.e., for
multicluster.

This change renames the module and changes the transport preface to
reflect this.

### DIFF
--- a/linkerd/app/integration/src/tests/transparency.rs
+++ b/linkerd/app/integration/src/tests/transparency.rs
@@ -355,7 +355,7 @@ macro_rules! http1_tests {
         }
 
         #[tokio::test]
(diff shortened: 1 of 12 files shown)


linkerd/app/integration/src/tests/transparency.rs
```diff
@@ -355,7 +355,7 @@ macro_rules! http1_tests {
         }
 
         #[tokio::test]
-        async fn http1_removes_opaque_transports() {
```
olix0r, line 358: this was an incorrect rename to begin with

hawkw (approved):

PR #826, opened 2021-01-09
update to Tokio 1.0

hawkw:
This branch updates the proxy to Tokio 1.0 and its ecosystem, including:

- `tokio` 1.0 (obviously)
- the `tokio` 1.0 versions of `tokio-test` and `tokio-util`
- `bytes` 1.0
- `tower` 0.4
- `hyper` 0.14.2
- `h2` 0.3
- a `tonic` branch with the Tokio update
- `prost` 0.7
- `tokio-rustls` 0.22
- `rustls` 0.19
- `trust-dns-resolver` 0.20

In addition, I've also updated `rand` to 0.8 while I'm updating deps,
to minimize duplicate dependencies.

This branch depends on linkerd/linkerd2-proxy-api#53. Currently, the
`proxy-api` dep is pointed at that branch.

For the most part, the change here is pretty straightforward, just
updating the specified dependency versions. The API changes in `tokio`
were mostly made in 0.3, so we have already tracked most of them.
However, the following small changes were necessary:

- Hyper now has feature flags that must be enabled.
- `bytes` renamed `Buf::bytes` and `Buf::bytes_vectored` and
  `BufMut::bytes_mut` to `chunk`, `chunk_vectored`, and `chunk_mut`,
  respectively.
- `tokio::sync::Semaphore` now has a `close` method, and `acquire`
  returns a `Result`. Updated `linkerd2-channel` to track this. This
  also allowed removing some kind of janky jury-rigged code for closing
  a semaphore.
- The `tokio::Runtime::builder::max_threads` methodc was renamed to
  `max_blocking_threads`.

There were also a couple of larger changes. In particular:

- `tokio` no longer re-exports the `Stream` trait from `Futures`, for
  stability reasons. I added a wrapper API for implementing `Stream` for
  `mpsc`  receivers and for `tokio::time::Interval`.
- `tokio::time::Sleep` is now `!Unpin`, since it no longer contains an
  internal heap allocation. This reduces heap overhead when used with
  async/await syntax. However, for manual `Future` and `tower::Service`
  impls, the `Box` must be added _outside_ the `Sleep` to preserve
  `Unpin`nedness. I've done this in all places where it was necessary. I
  also changed some `Service` impls (`failfast` and `switch_ready`) which
  repeatedly create new `Sleep` futures, to use `Sleep::reset` instead,
  allowing them to avoid reallocations.

  This also means that `tower`'s `Timeout` middleware's future is no
  longer `Unpin`. Since connect futures are required to be `Unpin`, and
  the connect stacks contain timeout layers, I added a new `BoxResponse`
  middleware layer for boxing *just* response futures, while preserving
  the type of the wrapped `Service` (so that connect stacks are still
  `Clone`).
- Removed the `linkerd-app-profiling` crate, which has bit-rotted.

The latest `tower` release also adds upstream implementations of a
number of middlewares we've written our own versions of. As a follow-up,
we probably want to replace some of this stuff with the upstream
versions. However, I thought that we would probably want to do this in a
separate PR, to minimize the size of this change.


olix0r:
Sweet!  A few notes:

* Pushed a fix for `linkerd/dns`, which needed the `time` feature.
* Should we update trust-dns-resolver in `linkerd/dns`? It's pulling in a dependency on tokio v0.3
* `linkerd/concurrency-limit` has a dependency on tokio v0.2
* We still depend on tower-test v0.3 which depends on tokio v0.2
* `opencensus-proto` depends on tonic v0.3 and therefore tokio v0.2
* `linkerd2-tracing` depends on `tokio-trace` v0.1, which depends on tokio v0.3
* `app/core` depends on a git revision of `tower-request-modifier` is that necessary?
* linkerd2-proxy-api (and others) still depend on `tonic v0.3.1 (https://github.com/QuentinPerez/tonic?branch=qperez/upgrade-to-tokio-1#329fe326) `
* Similarly, dependencies still exist on prost v0.6

All of this gleaned by inspecting `cargo tree`

olix0r:
Failures related to app-profiling fixed here https://github.com/linkerd/linkerd2-proxy/pull/827

olix0r:
https://github.com/linkerd/linkerd2-proxy/pull/829 removes tower-request-modifier.

I think after that, the only outdated dependency is tonic. The tonic is upgrade is somehow held up on a warp upgrade, which is silly, since warp is only needed for an example. Regardless, we can probably merge this without tonic...

olix0r:
Correction: i've updated tonic via `cargo update -p tonic` and its old dependencies have been replaced. The only outdated dependency is now `tokio-trace` et al

```
├── tokio-trace v0.1.0 (https://github.com/hawkw/tokio-trace?rev=8bba933f49349cc14665bee7b102f013119c74fa#8bba933f)
│   ├── num_cpus v1.13.0
│   │   └── libc v0.2.76
│   ├── serde v1.0.104 (*)
│   ├── tokio v0.3.5
│   │   ├── pin-project-lite v0.2.0
│   │   └── tracing v0.1.22 (*)
│   │   [build-dependencies]
│   │   └── autocfg v1.0.0
│   ├── tracing-core v0.1.17 (*)
│   └── tracing-subscriber v0.2.15
│       ├── lazy_static v1.4.0
│       ├── matchers v0.0.1
│       │   └── regex-automata v0.1.8
│       │       ├── byteorder v1.3.1
│       │       ├── regex-syntax v0.6.18
│       │       └── utf8-ranges v1.0.0
│       ├── parking_lot v0.11.0
│       │   ├── instant v0.1.8
│       │   │   └── cfg-if v1.0.0
│       │   ├── lock_api v0.4.1
│       │   │   └── scopeguard v1.1.0
│       │   └── parking_lot_core v0.8.0
│       │       ├── cfg-if v0.1.10
│       │       ├── instant v0.1.8 (*)
│       │       ├── libc v0.2.76
│       │       └── smallvec v1.5.0
│       ├── regex v1.3.9
│       │   └── regex-syntax v0.6.18
│       ├── serde v1.0.104 (*)
│       ├── serde_json v1.0.59 (*)
│       ├── sharded-slab v0.1.0
│       │   └── lazy_static v1.4.0
│       ├── smallvec v1.5.0
│       ├── thread_local v1.0.1
│       │   └── lazy_static v1.4.0
│       ├── tracing v0.1.22 (*)
│       ├── tracing-core v0.1.17 (*)
│       ├── tracing-log v0.1.1
│       │   ├── lazy_static v1.4.0
│       │   ├── log v0.4.11 (*)
│       │   └── tracing-core v0.1.17 (*)
│       └── tracing-serde v0.1.2
│           ├── serde v1.0.104 (*)
│           └── tracing-core v0.1.17 (*)
```

olix0r:
What are the impediments to publishing tokio-trace on crates.io?

When updating for tokio v1, can you please mark a license on the project?

hawkw:
> What are the impediments to publishing tokio-trace on crates.io?

The main reason it's unpublished is that it's pretty unfinished at the moment. I can try to get it publishable.

> When updating for tokio v1, can you please mark a license on the project?

Sure, will do!

linkerd/stack/src/box_response.rs
```diff
@@ -0,0 +1,43 @@
...
+    task::{Context, Poll},
+};
+
+/// A middleware that boxes *just* the response future from an inner service,
+/// without erasing the service's type (and its trait impls, such as `Clone`).
+///
+/// This is primarily useful when a service's `Future` type is not `Unpin` and
+/// must be boxed.
+#[derive(Clone, Debug)]
+pub struct BoxResponse<T>(T);
```
olix0r, line 13: WDYT about calling this `BoxFuture`? I would assume that this boxes the response value in some way.

linkerd/timeout/src/failfast.rs
```diff
@@ -51,6 +52,10 @@ impl<S> FailFast<S> {
             scope,
             inner,
             max_unavailable,
+            // the delay is reset whenever the service becomes unavailable; this
+            // initial one will never actually be used, so it's okay to start it
+            // now.
+            wait: Box::pin(time::sleep(max_unavailable)),
```
olix0r, line 58: wdyt about using a max-duration to make this clearer?

olix0r (approved):

linkerd/timeout/src/failfast.rs
hawkw: [`Duration::MAX` is unstable](https://doc.rust-lang.org/std/time/struct.Duration.html#associatedconstant.MAX), so I used `Duration::default()` (which is zero) instead.

PR #829, opened 2021-01-10
Remove dependency on tower-http

olix0r:
tower-http is not actually production ready, and our git depenency now
introduces outdated crate dependencies. This change replaces the use of
`tower-request-modifier` with a simple one-off service.

This ends up being a net reduction in code.

### DIFF
--- a/linkerd/app/core/src/control.rs
+++ b/linkerd/app/core/src/control.rs
@@ -75,59 +75,57 @@ impl Config {
 /// Sets the request's URI from `Config`.
 mod add_origin {
     use super::ControlAddr;
-    use linkerd2_stack::{layer, NewService, ResultService};
-    use std::marker::PhantomData;
-    use tower_request_modifier::{Builder, RequestModifier};
+    use linkerd2_stack::{layer, NewService};
+    use std::task::{Context, Poll};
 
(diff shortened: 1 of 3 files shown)


hawkw (approved):
great! take it or leave it, but it seems like we *could* just merge this to main, and rebase the Tokio 1.0 branch to pick it up?

linkerd/app/core/src/control.rs
```diff
@@ -75,59 +75,59 @@ impl Config {
...
-    // XXX the request_modifier build error does not implement Error...
-    #[derive(Debug)]
-    pub struct BuildError(());
 
-    impl std::error::Error for BuildError {}
-    impl std::fmt::Display for BuildError {
-        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-            write!(f, "failed to build the add-origin request modifier")
+        fn call(&mut self, req: http::Request<B>) -> Self::Future {
+            let (mut parts, body) = req.into_parts();
```
hawkw, line 123: on the other hand, this could probably be implemented using Tower 0.4's `map_request`...

olix0r: thought about that, but we can't name MapRequest, so it gets awkward quick.

PR #833, opened 2021-01-11
Use cargo-deny to audit dependencies

olix0r:
The cargo-deny utility provides a convenient way to audit the proxy's
dependency tree to audit security vulnerabilities & licensing.
Furthermore, it helps detect whether we have multiple versions of a
given dependency.

This change adds a `deny.toml` that configures the project to detect all
such issues and it adds a CI task that can help prevent us from
regressing.

This change introduces a number of dependency updates to eliminate
issues flagged by cargo-deny:

* `futures` has been updated to v0.3.9 to fix safety issues;
* `pin-project` has been updated to v1;
* `regex` has been updated to v1;
* `quickcheck` has been updated to v1, which caught minor issues in
  exp-backoff that have been fixed;

Furthermore, this change updates our github actions to use more more
recent versions of the base container image and updates actions/checkout
to v2.


deny.toml
```diff
@@ -0,0 +1,71 @@
...
+[sources]
+unknown-registry = "deny"
+unknown-git = "deny"
+allow-registry = ["https://github.com/rust-lang/crates.io-index"]
+allow-git = [
+    "https://github.com/hawkw/tokio-trace",
+    "https://github.com/hyperium/tonic",
+]
+
+[sources.allow-org]
```
hawkw, line 68: TIOLI: maybe we should just allow the "tokio", "tower", and "hyperium" orgs globally, to make it less churn  to patch stuff in the future?

linkerd/exp-backoff/src/lib.rs
```diff
@@ -64,6 +63,12 @@ impl ExponentialBackoff {
         if jitter < 0.0 {
             return Err(InvalidBackoff("jitter must not be negative"));
         }
+        if jitter > 100.0 {
+            return Err(InvalidBackoff("jitter must not be greater than 100"));
+        }
+        if !jitter.is_finite() {
+            return Err(InvalidBackoff("jitter must not be finite"));
```
hawkw, line 70: i think this is backwards:
```suggestion
            return Err(InvalidBackoff("jitter must be finite"));
```

linkerd/stack/src/box_response.rs
```diff
@@ -0,0 +1,43 @@
...
+    task::{Context, Poll},
+};
+
+/// A middleware that boxes *just* the response future from an inner service,
+/// without erasing the service's type (and its trait impls, such as `Clone`).
+///
+/// This is primarily useful when a service's `Future` type is not `Unpin` and
+/// must be boxed.
+#[derive(Clone, Debug)]
+pub struct BoxResponse<T>(T);
```
hawkw, line 13: this was renamed to `BoxFuture` before merging #826, so I think this file is now a duplicate and is unused...can we just remove this?

linkerd/stack/src/lib.rs
```diff
@@ -24,6 +25,7 @@ mod unwrap_or;
 pub use self::{
     box_future::BoxFuture,
     box_new_service::BoxNewService,
+    box_response::BoxResponse,
```
hawkw, line 28:
```suggestion
```

opencensus-proto/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 name = "opencensus-proto"
 version = "0.1.0"
 authors = ["The OpenCensus Authors"]
+license = "Apache-2.0"
```
hawkw, line 5: do we know what license the opencensus protobuf API definition is distributed under? does this need to match that?

hawkw (changes requested):
overall, LGTM; I noticed a couple minor issues we should address, and had some additional questions. in particular, let's get rid of `BoxResponse`, which is now unused and i think should have been deleted when merging main, and let's fix the wrong error message in backoff.

linkerd/exp-backoff/src/lib.rs
olix0r: no, the jitter must be finite. this error is thrown when it's not finite.

opencensus-proto/Cargo.toml
olix0r: It matches https://github.com/census-instrumentation/opencensus-proto

deny.toml
olix0r: I'd rather add explicit allowances as we need them. Since we control the linkerd project, though, I'm okay with a blanket allowance for linkerd.

linkerd/exp-backoff/src/lib.rs
olix0r: oh i see yes you're right. duh.

hawkw (approved):
lgtm when CI passes!

PR #834, opened 2021-01-11
Rename crates to reflect directory structure

olix0r:
This change renames most subcrates to be linkerd-* instead of
linkerd2-*, matching the repository layout.

The `linkerd2-proxy` binary crate remains unchanged, as it matches the
repo name; and the `linkerd2-proxy-api` is handled similarly.


hawkw (approved):
seems good to me!

this will change the `tracing` targets, so we may want to update any documentation that refers to setting log filters with targets like `linkerd2-foo`...

linkerd/tracing/src/lib.rs
```diff
@@ -14,7 +14,7 @@ use tracing_subscriber::{fmt::format, prelude::*};
 const ENV_LOG_LEVEL: &str = "LINKERD2_PROXY_LOG";
 const ENV_LOG_FORMAT: &str = "LINKERD2_PROXY_LOG_FORMAT";
 
-const DEFAULT_LOG_LEVEL: &str = "warn,linkerd2_proxy=info";
+const DEFAULT_LOG_LEVEL: &str = "warn,linkerd_proxy=info";
```
olix0r, line 17:
```suggestion
const DEFAULT_LOG_LEVEL: &str = "warn,linkerd=info";
```

hawkw: i forgot this exists

PR #835, opened 2021-01-11
stack: remove `map_response`

hawkw:
The `linkerd::stack::map_response` middleware is currently unused, and
an implementation of a similar middleware was added to `tower::util` in
tower 0.4. Therefore, we can just remove this --- if we need it again,
we can just use the upstream version.


olix0r:
Is this middleware even used? Can we just remove it (and use it from tower::util if we do need it in the future)?

hawkw:
> Is this middleware even used? Can we just remove it (and use it from tower::util if we do need it in the future)?

...you're right, it looks like we never actually use this. i thought we'd already gotten rid of all the unused middlewares. i'll just remove it

hawkw:
okay updated this to just nuke it entirely

PR #837, opened 2021-01-12
Split linkerd-tls from linkerd-proxy-transport

olix0r:
The `linkerd-proxy-transport` crate has too many concerns. This change
splits out the tls module into a dedicated crate, decoupled from the
transport types.

Additionally, the transport crate no longer re-exports the `linkerd-io`
crate.

The linkerd-proxy-transport crate has a lingering dependency on
linkerd-tls for metric labels. This will be addressed in a followup
change.

This change sets up additional simplification of the TLS infrastructure
to support ALPN.


linkerd/proxy/tap/src/accept.rs
```diff
@@ -75,8 +73,8 @@ impl Service<Connection<TcpStream>> for AcceptPermittedClients {
         Poll::Ready(Ok(()))
     }
 
-    fn call(&mut self, (meta, io): Connection<TcpStream>) -> Self::Future {
-        future::ok(match meta.peer_identity {
+    fn call(&mut self, ((peer_id, _), io): Connection<T, TcpStream>) -> Self::Future {
```
hawkw, line 76: this is a lot of tuples to pass around...but, introducing named structs for this might just make it messier, idk...

hawkw (approved):
the pile of tuples feels a little hard to follow, but i'm not sure if there's a nicer solution that decouples the crates in the same way. i'm fine with this change!

linkerd/proxy/tap/src/accept.rs
olix0r: i want to make this all more principled. this is an intermediate step...

PR #838, opened 2021-01-12
Remove tap status being explicitly set in config

kleimkuhler:
This is required as part of linkerd/linkerd2#5326 because the tap env variables
will no longer always be set.

There is a dependency on identity being enabled for tap to work. The
status of tap is determined by the `ENV_TAP_SVC_NAME` env variable being set
or not set.

- If identity is disabled, tap is disabled, but a warning is issued if
  `ENV_TAP_SVC_NAME` is set.
- If identity is enabled, the status of tap is determined by
  `ENV_TAP_SVC_NAME`.

### DIFF
--- a/linkerd/app/integration/src/proxy.rs
+++ b/linkerd/app/integration/src/proxy.rs
@@ -244,17 +244,16 @@ async fn run(proxy: Proxy, mut env: TestEnv, random_ports: bool) -> Listening {
         Some(identity.addr)
     } else {
         env.put(app::env::ENV_IDENTITY_DISABLED, "test".to_owned());
-        env.put(app::env::ENV_TAP_DISABLED, "test".to_owned());
         None
     };
 
     // If identity is enabled but the test is not concerned with tap, ensure
     // there is a tap service name set
-    if !env.contains_key(app::env::ENV_TAP_DISABLED)
-        && !env.contains_key(app::env::ENV_TAP_SVC_NAME)
-    {
-        env.put(app::env::ENV_TAP_SVC_NAME, "test-identity".to_owned())
-    }
+    // if !env.contains_key(app::env::ENV_TAP_DISABLED)
+    //     && !env.contains_key(app::env::ENV_TAP_SVC_NAME)
+    // {
+    //     env.put(app::env::ENV_TAP_SVC_NAME, "test-identity".to_owned())
+    // }
 
     if let Some(ports) = proxy.inbound_disable_ports_protocol_detection {
         let ports = ports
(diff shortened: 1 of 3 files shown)


linkerd/app/src/env.rs
```diff
@@ -730,20 +724,25 @@ fn parse_tap_config(
     match (id_disabled, tap_disabled) {
         (_, true) => Ok(None),
         (true, false) => {
-            error!("{} must be set if identity is disabled", ENV_TAP_DISABLED);
-            Err(EnvError::InvalidEnvVar)
+            let tap_identity = parse(strings, ENV_TAP_SVC_NAME, parse_identity);
+            match tap_identity? {
```
olix0r, line 728: Does the tap injector skip pods on which identity is disabled? Would it be better to warn in this case and continue without tap?

olix0r (commented):
I think we should simplify this even further and remove the TAP_DISABLED case, such that:

* Tap is always disabled by default
* If the tap identity is specified and identity is not disabled, tap is enabled.
* If the tap identity is specified and identity is disabled, we warn and proceed without tap.

kleimkuhler:
Good suggestion! I removed `TAP_DISABLED` now configuring tap is more straightforward. If identity is disabled, tap is disabled—warning if `TAP_SVC_NAME` was set. If identity is enabled, tap status is determined by `TAP_SVC_NAME` being set or not set.

I updated some tests as well.

linkerd/app/integration/src/proxy.rs
```diff
@@ -244,17 +244,16 @@ async fn run(proxy: Proxy, mut env: TestEnv, random_ports: bool) -> Listening {
...
-    if !env.contains_key(app::env::ENV_TAP_DISABLED)
-        && !env.contains_key(app::env::ENV_TAP_SVC_NAME)
-    {
-        env.put(app::env::ENV_TAP_SVC_NAME, "test-identity".to_owned())
-    }
+    // if !env.contains_key(app::env::ENV_TAP_DISABLED)
+    //     && !env.contains_key(app::env::ENV_TAP_SVC_NAME)
+    // {
+    //     env.put(app::env::ENV_TAP_SVC_NAME, "test-identity".to_owned())
+    // }
```
olix0r, line 256: Can we remove this (and the comment above)?

olix0r (approved):

linkerd/app/integration/src/proxy.rs
kleimkuhler: Whoops yep meant to remove that.

PR #839, opened 2021-01-12
replace `FutureService` with Tower's upstream impl

hawkw:
Tower 0.4 included a `FutureService` type which is basically identical
to the version we implemented for converting a `MakeService` into a
`NewService`. This branch replaces our implementation with the upstream
implementation from Tower.

It was necessary to add `MapErr` layers to convert the error types into
boxed errors, since our `FutureService` did that automatically, while
Tower's does not.

Depends on tower-rs/tower#523 from upstream, which is currently
unreleased.


olix0r (commented):
This looks good but I have a slight preference to wait for an actual rather than taking more git deps.

olix0r:
Marking as draft until we have a non-git dep

hawkw:
Tower 0.4.3 has been published, so I've removed the git dep and marked this as ready.

linkerd/app/core/src/svc.rs
```diff
@@ -18,7 +18,11 @@ use tower::{
     make::MakeService,
 };
 pub use tower::{
-    layer::Layer, service_fn as mk, spawn_ready::SpawnReady, util::Either, Service, ServiceExt,
+    layer::Layer,
+    service_fn as mk,
+    spawn_ready::SpawnReady,
+    util::{Either, MapErrLayer},
```
olix0r, line 24:
```suggestion
    util::{Either, MapErr},
```

linkerd/app/outbound/src/server.rs
```diff
@@ -146,6 +146,7 @@ where
     let tcp_forward = svc::stack(tcp_connect)
         .push_make_thunk()
         .push_on_response(tcp::Forward::layer())
+        .push(svc::MapErrLayer::new(Into::into))
```
olix0r, line 149:
```suggestion
        .push(svc::MapErr::layer(Into::into))
```

linkerd/app/outbound/src/ingress.rs
```diff
@@ -73,6 +73,7 @@ where
         .into_inner();
 
     svc::stack(http)
+        .push_on_response(svc::MapErrLayer::new(Into::into))
```
olix0r, line 76:
```suggestion
        .push_on_response(svc::MapErr::layer(Into::into))
```

linkerd/app/outbound/src/http/logical.rs
```diff
@@ -73,6 +73,7 @@ where
                 .push(svc::FailFast::layer("HTTP Balancer", dispatch_timeout))
                 .push(metrics.stack.layer(stack_labels("http", "concrete"))),
         )
+        .push(svc::MapErrLayer::new(Into::into))
```
olix0r, line 76:
```suggestion
        .push(svc::MapErr::layer(Into::into))
```

olix0r (approved):

Cargo.toml
```diff
@@ -66,4 +66,4 @@ debug = false
 [patch.crates-io]
 webpki = { git = "https://github.com/linkerd/webpki", branch = "cert-dns-names-0.21", rev = "b2c3bb3" }
 tonic = { git = "https://github.com/hyperium/tonic" }
-tonic-build = { git = "https://github.com/hyperium/tonic" }
+tonic-build = { git = "https://github.com/hyperium/tonic" }
```
olix0r, line 69: ?

linkerd/app/core/src/svc.rs
hawkw: i actually tried this initially and it didn't compile — `MapErr` is generic over a Service and a fn, while `MapErrLayer` is only generic over the fn type. using the `layer` fn requires a type for the Service type param, which apparently can't be inferred in this usage. :/

Cargo.toml
hawkw: wait, what the heck, is this a whitespace change? i can't see what's different...

PR #841, opened 2021-01-13
Split linkerd-detect from linkerd-proxy-transport

olix0r:
The `linkerd2_proxy_transport::detect` module has no dependencies on
other transport code.

This change simplifies dependencies by moving the protocol detection
infrastructure to a dedicated crate. We also provide a
`NewDetectService::timeout` helper to reduce boilerplate in stack
constructors.


kleimkuhler (approved):
👍

PR #842, opened 2021-01-13
replace `RequestFilter` with Tower's upstream impl

hawkw:
Tower provides a `tower::filter` utility for conditionally rejecting
some requests and accepting others. Rather than using `tower`'s filter,
we have a similar `linkerd_stack::request_filter`, due to two main
issues with the filtering API in tower 0.3 and earlier:

* `tower::filter`'s predicates returned futures, which we don't need
  --- all our filtering predicates are purely sunchronous
* `tower::filter` didn't allow filters to return any error of their
  choosing --- they had to return a specific filter-rejected error.
  This doesn't work with other code where we downcast error types,
  like switch/fallback.

However, Tower 0.4 includes a rewritten `Filter` API that no longer
has these issues (tower-rs/tower#508). Therefore, we can switch to
using the upstream implementation. This branch replaces `RequestFilter`
with `tower::filter::Filter`.

It was necessary to add an impl of `NewService` for `Filter`. This
required an upstream change to expose access to the wrapped service,
so that its `new_service` method can be called when it's a
`NewService`, and to expose access to the `Predicate`. Therefore, this
PR depends on tower-rs/tower#521 and tower-rs/tower#522, and introduces
a Git dependency. The git dep can be removed when Tower 0.4.3 is
released.


hawkw:
Tower 0.4.3 has been published upstream so I'm removing the draft tag.

Cargo.toml
```diff
@@ -66,4 +66,4 @@ debug = false
 [patch.crates-io]
 webpki = { git = "https://github.com/linkerd/webpki", branch = "cert-dns-names-0.21", rev = "b2c3bb3" }
 tonic = { git = "https://github.com/hyperium/tonic" }
-tonic-build = { git = "https://github.com/hyperium/tonic" }
+tonic-build = { git = "https://github.com/hyperium/tonic" }
```
olix0r, line 69: Why do we even need these patches now?

linkerd/stack/Cargo.toml
```diff
@@ -15,7 +15,7 @@ futures = "0.3.9"
 linkerd-error = { path = "../error" }
 pin-project = "1"
 tokio = { version = "1", features = ["time"] }
-tower = { version = "0.4.2", features = ["util"] }
+tower = { version = "0.4.3", features = ["util", "filter"] }
```
olix0r, line 18:
```suggestion
tower = { version = "0.4.3", features = ["filter", "util"] }
```

olix0r (approved):

PR #847, opened 2021-01-15
Correct gRPC 'max-concurrency exhausted' error messages

olix0r:
When the proxy hits a fail-fast error when proxying gRPC messages, it
reports 'max-concurrency exhausted' in the `grpc-message` response
header. This is because failfast was introduced to bound the time a
service may be at-concurrency; but now failfast is used everywhere a
request may be buffered to bound the time a request can stay in the
proxy: if a buffered service fails to become ready within a timeout, the
service goes into 'fail-fast' so that requests are failed eagerly until
the service is ready.

Recently, we've improved failfast errors to indicate which failfast is
being triggered. This change replaces the _max-concurrency exhausted_
`grpc-message` with this more-descriptive error message.

### DIFF
--- a/linkerd/app/core/src/errors.rs
+++ b/linkerd/app/core/src/errors.rs
@@ -265,12 +265,15 @@ fn set_grpc_status(
         headers.insert(GRPC_STATUS, code_header(code));
         headers.insert(GRPC_MESSAGE, HeaderValue::from_static("request timed out"));
         code
-    } else if error.is::<FailFastError>() {
+    } else if let Some(e) = error.downcast_ref::<FailFastError>() {
         let code = Code::Unavailable;
         headers.insert(GRPC_STATUS, code_header(code));
         headers.insert(
             GRPC_MESSAGE,
-            HeaderValue::from_static("proxy max-concurrency exhausted"),
+            HeaderValue::from_str(&e.to_string()).unwrap_or_else(|error| {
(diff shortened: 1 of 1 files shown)


linkerd/app/core/src/errors.rs
```diff
@@ -265,12 +265,15 @@ fn set_grpc_status(
...
         headers.insert(GRPC_MESSAGE, HeaderValue::from_static("request timed out"));
         code
-    } else if error.is::<FailFastError>() {
+    } else if let Some(e) = error.downcast_ref::<FailFastError>() {
         let code = Code::Unavailable;
         headers.insert(GRPC_STATUS, code_header(code));
         headers.insert(
             GRPC_MESSAGE,
-            HeaderValue::from_static("proxy max-concurrency exhausted"),
+            HeaderValue::from_str(&e.to_string()).unwrap_or_else(|error| {
```
hawkw, line 273: since the failfast scopes are static strings, i wonder if we could just add more of the message to them, so that they can be used with `HeaderValue::from_static`...string overhead probably doesn't matter that much so this is probably unnecessary...

hawkw (approved):
:raised_hands: thank you!

linkerd/app/core/src/errors.rs
olix0r: Yeah, probably more we could yakshave around this; but later.

PR #848, opened 2021-01-15
failfast: Improve diagnostic logging

olix0r:
Failfast errors can be difficult to track down. This change updates the
failfast module to emit more descriptive log messages. A `WARN`-level
message is emitted when a service enters failfast.

### DIFF
--- a/linkerd/timeout/src/failfast.rs
+++ b/linkerd/timeout/src/failfast.rs
@@ -11,7 +11,7 @@ use std::{
     task::{Context, Poll},
 };
 use tokio::time::{self, Duration, Instant, Sleep};
-use tracing::{debug, trace};
+use tracing::{debug, trace, warn};
 
 #[derive(Debug)]
 pub struct FailFast<S> {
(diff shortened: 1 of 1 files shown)


linkerd/timeout/src/failfast.rs
```diff
@@ -128,8 +133,8 @@ where
             ret => {
                 match self.state {
                     State::Open => {}
-                    State::Waiting => trace!("Ready"),
-                    State::FailFast => debug!("Recovered"),
+                    State::Waiting => trace!("{} has become ready", self.scope),
+                    State::FailFast => debug!("{} service has recovered", self.scope),
```
hawkw, line 137: should we consider bumping this one up to info? since folks are going to actually see the "entering failfast" message in their logs, it might be worth also letting them know when the service _leaves_ failfast, so they don't assumeit has failed forever?

hawkw (approved):
looks good to me overall, one little nit about log levels

PR #852, opened 2021-01-16
Annotate socket-level errors with a scope

olix0r:
We frequently encounter socket errors like:

    Connection closed error=Transport endpoint is not connected (os error 107)

While the handler is generally scoped with tracing context to tell us
information about the server-side connection, we lack information about
which socket actually failed. For every "logical" TCP stream, we
generally have two concrete connections: One on the server-side of the
proxy and another on the client-side.

This change introduces a new `io::ScopedIo` wrapper type that annotates
all I/O error messages with a `client` or `server` prefix. Then, the
proxy's clients and servers are wrapped with this type so that all
socket-level error messages are disambiguated.

### DIFF
--- a/linkerd/io/src/scoped.rs
+++ b/linkerd/io/src/scoped.rs
@@ -1,0 +1,133 @@
+use crate as io;
+use pin_project::pin_project;
+use std::{pin::Pin, task::Context};
+
+/// An I/O stream where errors are annotated a scope.
+#[pin_project]
+#[derive(Debug)]
+pub struct ScopedIo<I> {
+    scope: Scope,
+
+    #[pin]
+    io: I,
+}
+
+#[derive(Copy, Clone, Debug)]
+enum Scope {
+    Client,
+    Server,
+}
+
+// === impl Scope ===
+
+impl Scope {
+    fn err(&self, err: io::Error) -> io::Error {
+        io::Error::new(err.kind(), format!("{}: {}", self, err.to_string()))
+    }
+}
+
+impl std::fmt::Display for Scope {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Client => write!(f, "client"),
+            Self::Server => write!(f, "server"),
+        }
+    }
+}
+
+// === impl ScopedIo ===
+
+impl<I> ScopedIo<I> {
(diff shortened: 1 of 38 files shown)


linkerd/io/src/scoped.rs
```diff
@@ -0,0 +1,133 @@
...
+enum Scope {
+    Client,
+    Server,
+}
+
+// === impl Scope ===
+
+impl Scope {
+    fn err(&self, err: io::Error) -> io::Error {
+        io::Error::new(err.kind(), format!("{}: {}", self, err.to_string()))
```
hawkw, line 25: nit: the `to_string` _shouldn't_ be necessary --- that's equivalent to `format!("{}", error)`, and we're already making a string, so this is just
```suggestion
        io::Error::new(err.kind(), format!("{}: {}", self, err))
```
but with extra steps

linkerd/io/src/scoped.rs
```diff
@@ -0,0 +1,133 @@
...
+            scope: Scope::Server,
+            io,
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl<I: io::Peek + Send + Sync> io::Peek for ScopedIo<I> {
+    async fn peek(&self, buf: &mut [u8]) -> io::Result<usize> {
+        self.io.peek(buf).await.map_err(|e| self.scope.err(e))
```
hawkw, line 59: nit/tioli: we _could_ change `self.scope.err()` to return a closure, and then write 
```rust
.map_err(self.scope.err())
```
here and everywhere else. maybe slightly more concise. but, that might also just be confusing, so idk.

linkerd/io/src/scoped.rs
```diff
@@ -0,0 +1,133 @@
...
+
+impl<I: io::PeerAddr> io::PeerAddr for ScopedIo<I> {
+    #[inline]
+    fn peer_addr(&self) -> io::Result<std::net::SocketAddr> {
+        self.io.peer_addr().map_err(|e| self.scope.err(e))
+    }
+}
+
+impl<I: io::AsyncRead> io::AsyncRead for ScopedIo<I> {
+    fn poll_read(
```
hawkw, line 71: everything else is inlined, should this one also be?

hawkw (approved):
lgtm modulo a couple nits

PR #853, opened 2021-01-17
Decouple tls::accept from TcpStream

olix0r:
`tls::accept` has a `Detectable` trait which allows TLS detection to use
`TcpStream::peek`; but this is inflexible if we want to wrap `TcpStream`
with any additional behavior and it limits our ability to write tests
for this module.

This change introduces a new `io::Peek` trait to model
`TcpStream::peek` and removes the `tls::accept::Detectable` trait.


hawkw (approved):
looks good to me!

PR #855, opened 2021-01-18
tls: Disambiguate client and server identities

olix0r:
The `tls::PeerIdentity` type is used to describe both remote clients and
servers. This can easily lead to confusion, as it can be ambiguous as to
whether an identity is a client's identity or a target server's
identity.

This change introduces new marker types:

- `identity::LocalId`: The local proxy's ID;
- `tls::server::ClientId`: A remote client ID; and
- `tls::client::ServerId`: A target server ID.

Furthermore, the `tls::ReasonForNoPeerName` has been split into distinct
`tls::server::NoClientId` and `tls::client::NoServerId` types. This
change eliminates the `tls::HasPeerIdentity` and
`tls::{client, server}::HasConfig` types, in favor of simple `Into` coercions.

This change requires changes to the metric labeling, though no functional
changes have been introduced.


linkerd/app/core/src/metrics.rs
```diff
@@ -42,13 +45,24 @@ pub struct Metrics {
...
-    pub tls_id: TlsStatus,
+pub enum EndpointLabels {
+    Inbound(InboundEndpointLabels),
+    Outbound(OutboundEndpointLabels),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct InboundEndpointLabels {
+    pub client_id: tls::server::ConditionalTls,
+    pub authority: Option<http::uri::Authority>,
```
hawkw, line 60: was kind of wondering if this should be on the `EndpointLabels` type, since both directions have an authority, but i realize that just complicates the `Into` impls a bunch so nvm.

linkerd/app/inbound/src/lib.rs
```diff
@@ -5,12 +5,17 @@
 
 #![deny(warnings, rust_2018_idioms)]
 
+mod allow_discovery;
+mod prevent_loop;
+mod require_identity_for_ports;
+pub mod target;
```
hawkw, line 11: nit: any reason for moving this?

linkerd/app/inbound/src/lib.rs
```diff
@@ -171,7 +169,7 @@ impl Config {
                 svc::stack(tcp_forward)
                     .push_map_target(TcpEndpoint::from)
                     .push(metrics.transport.layer_accept())
-                    .push_map_target(TcpAccept::from)
+                    .push_map_target(TcpAccept::port_skipped)
```
hawkw, line 172: :+1:

linkerd/app/test/src/http_util.rs
```diff
@@ -13,7 +8,7 @@ use tracing::Instrument;
 
 pub struct Server {
     settings: hyper::server::conn::Http,
-    identity: Option<PeerIdentity>,
+    expected_id: Option<tls::ConditionalServerId>,
```
hawkw, line 11: :+1: for renaming this

linkerd/identity/src/lib.rs
```diff
@@ -51,6 +51,10 @@ struct CertResolver(rustls::sign::CertifiedKey);
 #[derive(Clone, Debug)]
 pub struct InvalidCrt(rustls::TLSError);
 
+/// A marker type for local server identities.
```
hawkw, line 54: nit: i wouldn't call this a "marker" type --- I think in Rust, "marker" is typically used to refer to traits like `Copy`, `Send`, `Sync` etc which signify something about a type without specifying methods/behaviors.

linkerd/identity/src/lib.rs
```diff
@@ -405,11 +429,37 @@ impl rustls::ResolvesServerCert for CertResolver {
...
+
+impl AsRef<Name> for LocalId {
+    fn as_ref(&self) -> &Name {
+        &self.0
+    }
+}
+
+impl fmt::Display for LocalId {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
```
hawkw, line 454: TIOLI: should this indicate that this is a local identity? i guess the `Debug` impl does that so maybe this shouldn't?

linkerd/proxy/identity/src/certify.rs
```diff
@@ -35,9 +35,9 @@ pub struct Config {
 #[pin_project]
 #[derive(Clone, Debug)]
 pub struct Local {
```
hawkw, line 37: nit/TIOLI: might be worth renaming this now that we have a type called `LocalId`, to destinguish it a little more clearly. Maybe something like `LocalCert`?

linkerd/proxy/identity/src/lib.rs
```diff
@@ -4,4 +4,4 @@ pub mod certify;
 pub mod metrics;
 
 pub use self::certify::{AwaitCrt, CrtKeySender, Local};
-pub use linkerd_identity::{Crt, CrtKey, Csr, InvalidName, Key, Name, TokenSource, TrustAnchors};
+// XXX pub use linkerd_identity::{Crt, CrtKey, Csr, InvalidName, Key, Name, TokenSource, TrustAnchors};
```
hawkw, line 7: should this just be removed? or was there something you were going to change?

linkerd/proxy/tap/src/grpc/server.rs
```diff
@@ -531,8 +531,10 @@ fn base_event<B, I: Inspect>(req: &http::Request<B>, inspect: &I) -> api::TapEve
                 }
                 Conditional::Some(id) => {
                     m.labels.insert("tls".to_owned(), "true".to_owned());
-                    m.labels
-                        .insert("client_id".to_owned(), id.as_ref().to_owned());
+                    m.labels.insert(
+                        "client_id".to_owned(),
+                        id.map(|id| id.to_string()).unwrap_or_default(),
```
hawkw, line 536: nit/TIOLI:
```suggestion
                        id.map(ToString::to_string).unwrap_or_default(),
```

linkerd/tls/src/client.rs
```diff
@@ -1,25 +1,47 @@
...
+    str::FromStr,
     sync::Arc,
     task::{Context, Poll},
 };
 use tokio_rustls::client::TlsStream;
 use tracing::{debug, trace};
 
-pub trait HasConfig {
-    fn tls_client_config(&self) -> Arc<Config>;
+/// A marker type for target server identities.
```
hawkw, line 20: nit: again, I wouldn't call this a "marker" but it's not a big deal.

linkerd/tls/src/client.rs
```diff
@@ -1,25 +1,47 @@
...
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct ServerId(pub id::Name);
```
hawkw, line 22: it _feels_ weird elsewhere to see a type named `ServerId` imported from the `client` module, but after actually thinking about it, it _does_ make sense...and I don't think we should give this a wordier name to clarify that. so :+1:

linkerd/tls/src/server.rs
```diff
@@ -17,19 +19,39 @@ use tokio_rustls::server::TlsStream;
...
 
 /// Produces a server config that fails to handshake all connections.
-pub fn empty_config() -> Arc<Config> {
+pub fn empty_config() -> Config {
     let verifier = rustls::NoClientAuth::new();
-    Arc::new(Config::new(verifier))
+    Arc::new(rustls::ServerConfig::new(verifier))
+}
+
+/// A marker type for remote client idenities..
```
hawkw, line 30: nit: ...and again, I probably wouldn't call this a "marker", since it _does_ contain actual data?

linkerd/tls/src/server.rs
```diff
@@ -17,19 +19,39 @@ use tokio_rustls::server::TlsStream;
...
+
+    /// No TLS is wanted because the connection is a loopback connection which
+    /// doesn't need or support TLS.
+    PortSkipped,
+
+    // TLS not established by the remote client.
+    NoTlsFromRemote,
 }
 
+pub type ConditionalTls = Conditional<Option<ClientId>, NoTls>;
```
hawkw, line 51: nit/TIOLI: it's a little bit weird that the corresponding type in the client module is `ConditionalServerId`, but this is `ConditionalTls`...but, it makes sense because here, on the accept side, we may be doing TLS without a client identity if it's not mutual TLS. I think the names should stay as they are, but it's maybe worth a comment explaining this, though?

hawkw (approved):
this is a big change, but most of it is just renaming. overall, I think this is definitely much clearer! i picked a few nits but didn't see any serious blockers.

linkerd/app/inbound/src/lib.rs
olix0r: It's kind of become the idiom to put modules before imports... From a discoverability point of view: easier to see all the modules as soon as you open the file, and then imports are closer uses (and reference modules that have already been defined). Just a nit really

linkerd/identity/src/lib.rs
olix0r: Correct. This is really for encoding the name into labels, etc. We want to encode the underlying name string and not something more descriptive.

PR #858, opened 2021-01-19
Ensure services in failfast can become ready

olix0r:
When a Service is in failfast, the inner service is only polled as new
requests are processed. This means it's theoretically possible for
certain service tasks to be starved.

This change ensures that these layers are paired with a `SpawnReady`
layer to ensure that the inner service is always driven to readiness.

This could potentially explain behavior as described in
linkerd/linkerd2#5183; though we don't have strong evidence to support
that. This seems like a healthy defensive measure, in any case.

This change also improves stack commentary to favor larger descriptive
comments over layer-level annotations.

While auditing services for readiness, an unnecessary buffer has been
removed from the ingress HTTP stack.

### DIFF
--- a/linkerd/app/outbound/src/http/logical.rs
+++ b/linkerd/app/outbound/src/http/logical.rs
@@ -47,20 +47,24 @@ where
     let watchdog = cache_max_idle_age * 2;
 
     svc::stack(endpoint.clone())
-        .check_new_service::<Endpoint, http::Request<http::BoxBody>>()
         .push_on_response(
             svc::layers()
-                .push(svc::layer::mk(svc::SpawnReady::new))
+                .push(http::BoxRequest::layer())
                 .push(
                     metrics
                         .stack
                         .layer(stack_labels("http", "balance.endpoint")),
                 )
-                .push(http::BoxRequest::layer()),
+                // Ensure individual endpoints are driven to readiness so that
+                // the balancer need not drive them all directly.
+                .push(svc::layer::mk(svc::SpawnReady::new)),
         )
-        .check_new_service::<Endpoint, http::Request<_>>()
+        // Resolve the service to its endponts and balance requests over them.
+        //
+        // If the balancer has been empty/unavailable, eagerly fail requests.
+        // When the balancer is in failfast, spawn the service in a background
+        // task so it becomes ready without new requests.
         .push(resolve::layer(resolve, watchdog))
-        .check_service::<Concrete>()
         .push_on_response(
             svc::layers()
                 .push(http::balance::layer(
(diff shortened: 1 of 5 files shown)


linkerd/app/outbound/src/http/logical.rs
```diff
@@ -47,55 +47,60 @@ where
     let watchdog = cache_max_idle_age * 2;
 
     svc::stack(endpoint.clone())
-        .check_new_service::<Endpoint, http::Request<http::BoxBody>>()
         .push_on_response(
             svc::layers()
-                .push(svc::layer::mk(svc::SpawnReady::new))
+                .push(http::BoxRequest::layer())
```
hawkw, line 52: just out of curiosity, why put the box above metrics now?

linkerd/app/outbound/src/ingress.rs
```diff
@@ -73,42 +73,48 @@ where
...
         .push(profiles::discover::layer(
             profiles,
             AllowHttpProfile(allow_discovery),
         ))
-        .check_new_service::<Target, http::Request<_>>()
         .push_on_response(
             svc::layers()
-                .push(svc::FailFast::layer("Logical", dispatch_timeout))
+                .push(svc::layer::mk(svc::SpawnReady::new))
+                .push(svc::FailFast::layer("HTTP Logical", dispatch_timeout))
```
hawkw, line 97: TIOLI: might be worth having a "push spawn ready and failfast in one method" helper since we now want to pair them in most places. might also help avoid accidentally adding new failfasts without a spawn ready?

linkerd/app/outbound/src/http/logical.rs
```diff
@@ -47,55 +47,60 @@ where
...
                 .push(
                     metrics
                         .stack
                         .layer(stack_labels("http", "balance.endpoint")),
                 )
-                .push(http::BoxRequest::layer()),
+                // Ensure individual endpoints are driven to readiness so that
+                // the balancer need not drive them all directly.
+                .push(svc::layer::mk(svc::SpawnReady::new)),
```
hawkw, line 60: TIOLI: could be nice to add a `SpawnReady::layer`?

hawkw (approved):
this seems right to me, and +1 for the new commentary. 

it would be nice to have more testing around failfast, but since we don't have a tight repro for the issues some people have been seeing, that can probably wait.

linkerd/app/outbound/src/http/logical.rs
olix0r: mostly I want the box as close to the inner service as possible, as it's responsible for satisfying the type signature of the inner service.

linkerd/app/outbound/src/ingress.rs
olix0r: Agreed, but we don't have a nameable spawn ready layer at the moment.

linkerd/app/outbound/src/http/logical.rs
olix0r: https://github.com/tower-rs/tower/pull/536

PR #859, opened 2021-01-19
tests: replace string matching on metrics with parsing

hawkw:
This branch changes the metrics integration tests so that we no longer
make assertions about metrics by searching the entire metrics scrape for
a specific string. Instead, we now have a simple builder API for
constructing a type that tries to parse the metrics scrape to find
individual metrics with certain label combinations and/or values.

Advantages of the new approach include:
- it's not ordering dependent --- changing the label ordering doesn't
  change whether or not the match succeeds. for example, changes
  like #856 won't break the tests.
- the presence of *additional* labels doesn't break the assertions
  (although we may want to add an "exact" matching mode in case we
  *do* want to assert that other labels are not present).
- metrics can easily be matched with and without values
- the failure reporting is *much* better. when the scrape doesn't
  contain the required metric, we can now report the metric we wanted
  in a nicely formatted way, and show only a list of the "similar"
  metrics that were found. with the previous code, panics included the
  entire metrics scrape, which was very hard to read and debug.

This should make it much easier to modify the tests to add new metric
labels or add existing ones, since we no longer depend on specific
hard-coded strings, but are instead making assertions about the
properties we actually want to find.

Example of the new errors:

```
thread 'tests::telemetry::response_classification::inbound_http' panicked at 'did not find a `response_total` metric
  with labels: ["authority=\"tele.test.svc.cluster.local\"", "direction=\"inbound\"", "tls=\"disabled\"", "status_code=\"200 OK\"", "classification=\"success\""]
found metrics: [
    "response_total{authority=\"tele.test.svc.cluster.local\",direction=\"inbound\",tls=\"disabled\",status_code=\"200\",classification=\"success\"} 1",
    "response_total{authority=\"tele.test.svc.cluster.local\",direction=\"inbound\",tls=\"disabled\",status_code=\"304\",classification=\"success\"} 1",
]
  retried 5 times (73.314652ms)', linkerd/app/integration/src/metrics.rs:128:51
stack backtrace:
    ... snip ...
```

These *could* probably be improved even more, for example, by
highlighting the specific label that was missing. However, this felt
like a pretty significant improvement for a first pass.

In a follow-up, I'll also reduce some of the code duplication between
the tests that are repeated for inbound and outbound. While working
on this change, I found that modifying the tests was much harder since
almost all of them were duplicated on the inbound and outbound halves
of the proxy. I'll merge that in a separate PR to reduce the size of
this diff, though.

### DIFF
--- a/linkerd/app/integration/src/lib.rs
+++ b/linkerd/app/integration/src/lib.rs
@@ -136,19 +136,23 @@ macro_rules! assert_contains {
 
 #[macro_export]
 macro_rules! assert_eventually_contains {
-    ($scrape:expr, $contains:expr) => {
+    ($scrape:expr, $contains:expr) => {{
+        let mut res: Result<(), crate::metrics::MatchErr> = Ok(());
         assert_eventually!(
-            $scrape.contains($contains),
-            "metrics scrape:\n{}\ndid not contain:\n{}",
-            $scrape,
-            $contains
+            {
+                res = $contains.is_in($scrape);
+                res.is_ok()
+            },
+            "{}",
+            res.unwrap_err(),
         )
-    };
+    }};
 }
 
 pub mod client;
 pub mod controller;
 pub mod identity;
+pub mod metrics;
 pub mod proxy;
 pub mod server;
 pub mod tap;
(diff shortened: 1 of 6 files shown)


olix0r (approved):
This looks great!

PR #860, opened 2021-01-20
test: reduce repetition in metrics tests

hawkw:
Currently, the metrics tests have a bunch of code that's repeated
between tests in the inbound and outbound directions. This makes
modifying the tests much more painful, since the code needs to be
changed multiple times. Also, the file is hard to navigate.

This commit changes this by factoring out the shared code into functions
which are called by the differing code (e.g. metrics labels) as
parameters. This means that the common behavior is only defined once, so
it should be somewhat easier to edit. This has a significant advantage
over the approach we use elsewhere, where similar tests are generated by
macros: it doesn't mess up the compiler diagnostics for errors in the
tests. We may want to refactor other tests to use a similar style in the
future.

### DIFF
--- a/linkerd/app/integration/src/client.rs
+++ b/linkerd/app/integration/src/client.rs
@@ -95,7 +95,7 @@ pub struct Client {
     tls: Option<TlsConfig>,
 }
 
-pub struct NewClient {
+pub struct Reconnect {
     addr: SocketAddr,
     authority: String,
     run: Run,
@@ -185,7 +185,7 @@ impl Client {
 
     /// Shut down the client, returning a type that can be used to initiate a
     /// new client connection to that target.
-    pub async fn shutdown(self) -> NewClient {
+    pub async fn shutdown(self) -> Reconnect {
         let Self {
             tx,
             task,
@@ -200,7 +200,7 @@ impl Client {
         drop(tx);
         task.await.unwrap();
         running.await;
-        NewClient {
+        Reconnect {
             authority,
             run,
             addr,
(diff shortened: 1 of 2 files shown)


linkerd/app/integration/src/client.rs
```diff
@@ -93,6 +95,13 @@ pub struct Client {
     tls: Option<TlsConfig>,
 }
 
+pub struct NewClient {
```
olix0r, line 98: tioli: Clearer to call this `Reconnect`? having `shutdown(self) -> Reconnect` is a little clearer than `shutdown(self) -> NewClient`

linkerd/app/integration/src/client.rs
```diff
@@ -172,18 +183,39 @@ impl Client {
         self.running.await
     }
 
-    pub async fn shutdown(self) {
+    /// Shut down the client, returning a type that can be used to initiate a
+    /// new client connection to that target.
+    pub async fn shutdown(self) -> NewClient {
```
olix0r, line 188: You could even have this return `impl Fn() -> Client + Clone` -- not a lot of need to have a named type here...

olix0r (approved):
lgtm assuming it passes reliably!

linkerd/app/integration/src/client.rs
hawkw: > You could even have this return `impl Fn() -> Client + Clone` -- not a lot of need to have a named type here...

hmm, I like the named type because it has a named _method_, the test is kinda unclear when you're not calling `.reconnect()` on something? but that's not a bad idea.

linkerd/app/integration/src/client.rs
hawkw: good call, renamed!

PR #861, opened 2021-01-20
metrics: add `target_addr` label for accepted transport metrics

hawkw:
This branch adds a `target_addr` label to the Prometheus metrics
generated for accepted (`peer="src"`) connections, containing the
original destination address of the accepted connection.

When we add the proxy's admin listener to the inbound metrics, 
this will allow distinguishing between admin and inbound proxy
traffic. Additionally, it allows partitioning transport metrics by
the client's target address.

Depends on #860 for test changes.


kleimkuhler (approved):
Nice looks good 👍

linkerd/app/core/src/transport/labels.rs
```diff
@@ -11,7 +11,11 @@ use std::fmt;
 /// Implements `FmtLabels`.
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
 pub enum Key {
-    Accept(Direction, TlsStatus),
+    Accept {
+        direction: Direction,
+        identity: TlsStatus,
+        local_addr: SocketAddr,
```
olix0r, line 17: How does this work on the outbound side? It's not a _local_ addr, is it? but really the _target_ addr? Perhaps we should just call it _target addr_ and encode the whole ip:port so that it makes sense on the outbound side?

olix0r (changes requested):
I suspect this will be odd on the outbound side?

linkerd/app/core/src/transport/labels.rs
hawkw: On the outbound side, it's set from the outbound listener's local address, _not_ the original destination (target) addr:
https://github.com/linkerd/linkerd2-proxy/blob/b0d95e619c3e1afffc4b8e6b19681a004ba70fea/linkerd/app/outbound/src/target.rs#L88

Is that what we want? If we're changing this to `target_addr`, it seems like it should be the original dest addr. In that case, we might want it to be the original dest addr in the inbound direction as well --- if a pod is listening on multiple ports, we would then capture which port the traffic was intended for by the client.

olix0r: Yeah, I think it's generally more useful to use the original dst / target addr. We're missing this data now (on the inbound side) and it's somewhat painful. In the case of the admin server, we'll have no original dst so we should just use the local target addr...

linkerd/app/outbound/src/target.rs
```diff
@@ -13,6 +13,7 @@ pub struct EndpointFromMetadata;
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub struct Accept<P> {
     pub orig_dst: SocketAddr,
+    pub local_addr: SocketAddr,
```
olix0r, line 16: Can this be removed now?

hawkw: whoops, missed that one. should be good now!

olix0r (approved):
lgtm

PR #862, opened 2021-01-20
outbound: Strip endpoint identity when disabled

olix0r:
When the proxy's identity is disabled, we may still mark outbound
endpoints as having a peer identity (even though a TLS connection is
never established.

This change modifies the outbound HTTP and TCP endpoint stacks to
optionally strip server identities before metrics are recorded.


linkerd/app/test/src/http_util.rs
```diff
@@ -75,19 +63,11 @@ impl Server {
...
-        } = self;
+        let Self { f, settings } = self;
         let f = Arc::new(Mutex::new(f));
         move |endpoint| {
             let span = tracing::debug_span!("server::run", ?endpoint);
             let _e = span.enter();
-            if let Some(expected_id) = expected_id.as_ref() {
-                let server_id: tls::ConditionalServerId = (&endpoint).into();
-                assert_eq!(&server_id, expected_id);
-            }
```
olix0r, line 90: Note that we had to drop this check because identity is never present nw...

linkerd/app/integration/src/tests/telemetry.rs
```diff
@@ -1170,8 +1164,7 @@ mod transport {
             metrics::labels()
                 .label("authority", proxy.outbound_server.as_ref().unwrap().addr)
                 .label("direction", "outbound")
-                .label("tls", "no_identity")
-                .label("no_tls_reason", "not_provided_by_service_discovery")
+                .label("tls", "disabled")
```
hawkw, line 1167: Nit/TIOLI: now that we use the same labels in the inbound and outbound directions, rather than `tls="disabled"` on inbound and `tls="no_identity"` on outbound, we could move this label out of the per-direction labels and into the common labels built in the test function,

hawkw (approved):

linkerd/app/integration/src/tests/telemetry.rs
olix0r: do you mean adding this to `Labels::default()`? In any case, I'll save that for a followup.

hawkw: I meant like here: https://github.com/linkerd/linkerd2-proxy/blob/6ea4de5dce10c52d13cc1b1c34928a2036e14236/linkerd/app/integration/src/tests/telemetry.rs#L914

The various `test_whatever` functions add labels that are common between inbound and outbound to the `Labels` struct that's passed in, the idea was that the actual tests would only need to add the labels that _differ_ between inbound and outbound.

PR #863, opened 2021-01-20
inbound: Handle direct connections with a dedicated stack

olix0r:
Currently, the inbound proxy handles direct connections--connections
where the original destination is the inbound proxy port--in two places.
TCP forwarding is handled distinctly from HTTP forwarding (for
gateways).

This change modifies the inbound proxy to dispatch these connections to
a dedicated `direct` stack that does its own protocol detection and
routing.

This sets up to (1) modify the server to configure ALPN on only these
connections and (2) support transport-header-based gateway routing.

### DIFF
--- a/linkerd/app/inbound/src/lib.rs
+++ b/linkerd/app/inbound/src/lib.rs
@@ -132,6 +132,7 @@ impl Config {
         // 1. Protocol detection is always performed;
         // 2. TLS is required;
         // 3. A transport header is expected. It's not strictly required, as
+        //    gateways may need to accept HTTP requests from older proxy versions.
         let direct = {
             let http = svc::stack(http_loopback)
                 .push_on_response(
--- a/linkerd/app/inbound/src/lib.rs
+++ b/linkerd/app/inbound/src/lib.rs
@@ -7,14 +7,16 @@
 
 mod allow_discovery;
 mod prevent_loop;
-mod require_identity_for_ports;
+mod require_identity;
 pub mod target;
 
-use self::allow_discovery::AllowProfile;
-use self::prevent_loop::PreventLoop;
-use self::require_identity_for_ports::RequireIdentityForPorts;
 pub use self::target::{
     HttpEndpoint, Logical, ProfileTarget, RequestTarget, Target, TcpAccept, TcpEndpoint,
+};
+use self::{
+    allow_discovery::AllowProfile,
+    prevent_loop::PreventLoop,
+    require_identity::{RequireIdentityForDirect, RequireIdentityForPorts},
 };
 use linkerd_app_core::{
     classify,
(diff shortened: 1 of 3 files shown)


linkerd/app/inbound/src/lib.rs
```diff
@@ -127,32 +123,88 @@ impl Config {
...
+                .push_map_target(
+                    |(v, (h, mut t)): (
+                        Option<http::Version>,
+                        (Option<TransportHeader>, TcpAccept),
+                    )| {
+                        if let Some(h) = h {
+                            t.target_addr = (t.target_addr.ip(), h.port).into();
+                        }
+                        (v, t)
+                    },
```
hawkw, line 175: yeesh.

(i imagine this is going to change soon though, based on the comment?)

linkerd/app/inbound/src/prevent_loop.rs
```diff
@@ -1,5 +1,10 @@
+use crate::TcpEndpoint;
 //use futures::future;
```
hawkw, line 2: nit: looks like this was already there, but should we delete it?

linkerd/app/inbound/src/require_identity.rs
```diff
@@ -0,0 +1,73 @@
...
+impl Predicate<TcpAccept> for RequireIdentityForDirect {
+    type Request = TcpAccept;
+
+    fn check(&mut self, meta: TcpAccept) -> Result<TcpAccept, Error> {
+        tracing::debug!(client.id = ?meta.client_id);
+        match meta.client_id {
+            Conditional::Some(Some(_)) => {}
+            _ => return Err(IdentityRequired(()).into()),
+        }
+        Ok(meta)
```
hawkw, line 32: nit/tioli: isn't this just 
```suggestion
        match meta.client_id {
            Conditional::Some(Some(_)) => Ok(meta),
            _ =>Err(IdentityRequired(()).into(),
        }
```
or does the borrow checker not like that?

linkerd/app/inbound/src/require_identity.rs
```diff
@@ -0,0 +1,73 @@
...
+
+/// A connection policy that fails connections that don't have a client identity
+/// if they target one of the configured local ports.
+#[derive(Clone, Debug)]
+pub struct RequireIdentityForPorts {
+    ports: Arc<IndexSet<u16>>,
+}
+
+#[derive(Debug)]
+pub struct IdentityRequired(());
```
hawkw, line 19: now that we return this in multiple places, should we add a scope to it? or will it be clear from the logs where it came from?

hawkw (approved):
this looks good to me! i had a few minor nits.

linkerd/app/inbound/src/require_identity.rs
olix0r: it's effectively the same error in both cases. It's as if 4143 is in the required ports list (which i believe it is, anyway).

PR #865, opened 2021-01-20
Update cargo dependencies

olix0r:
Eliminates several duplicate dependencies.


linkerd/service-profiles/src/http/mod.rs
```diff
@@ -26,7 +26,7 @@ pub enum RequestMatch {
     All(Vec<RequestMatch>),
     Any(Vec<RequestMatch>),
     Not(Box<RequestMatch>),
-    Path(Regex),
+    Path(Box<Regex>),
```
olix0r, line 29: I suspect the size of `Regex` changed and now clippy is unhappy about the size of the enum variant

hawkw: Yeah, it probably used to box internally, I guess?

hawkw (approved):

PR #866, opened 2021-01-20
metrics: add `target_addr` labels to HTTP metrics

hawkw:
This branch builds on #861 and adds `target_addr` labels to HTTP metrics
as well as transport metrics. These labels contain a string
representation of the original destination address of the connection on
which that request was sent.

Additionally, I've made some more improvements to the metrics tests:
after the test code failed to catch a missing commma in labels, I
changed the metrics matching code to check that the metric is correctly
formatted, as well as containing the expected labels and values.
Furthermore, after rebasing onto `main` made some labels in tests added
in this PR incorrect, I modified the tests so that a majority of the
expected labels are defined in a single place by the text fuxtures,
since they don't actually differ between tests.

### DIFF
--- a/linkerd/app/core/src/metrics.rs
+++ b/linkerd/app/core/src/metrics.rs
@@ -12,6 +12,7 @@ use linkerd_addr::Addr;
 use linkerd_metrics::FmtLabels;
 pub use linkerd_metrics::*;
 use std::fmt::{self, Write};
+use std::net::SocketAddr;
 use std::time::{Duration, SystemTime};
 
 pub type ControlHttp = http_metrics::Requests<ControlLabels, Class>;
@@ -58,6 +59,7 @@ pub enum EndpointLabels {
 pub struct InboundEndpointLabels {
     pub client_id: tls::server::ConditionalTls,
     pub authority: Option<http::uri::Authority>,
+    pub target_addr: SocketAddr,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
@@ -65,6 +67,7 @@ pub struct OutboundEndpointLabels {
     pub server_id: tls::ConditionalServerId,
     pub authority: Option<http::uri::Authority>,
     pub labels: Option<String>,
+    pub target_addr: SocketAddr,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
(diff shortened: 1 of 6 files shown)


linkerd/app/core/src/metrics.rs
```diff
@@ -266,6 +269,8 @@ impl FmtLabels for InboundEndpointLabels {
             write!(f, ",")?;
         }
 
+        write!(f, "target_addr=\"{}\"", self.target_addr)?;
```
olix0r, line 272: I would think this needs a trailing comma? Surprised our tests pass...

linkerd/app/core/src/metrics.rs
```diff
@@ -279,6 +284,8 @@ impl FmtLabels for OutboundEndpointLabels {
...
```
olix0r, line 287: Similarly...

olix0r (changes requested):

linkerd/app/core/src/metrics.rs
hawkw: Yeah, commit c75b2d5d835615fd77f8a978c2fd5113af1fc3e4 updates the metrics test code to actually check that metric labels are formatted correctly --- the tests now fail with errors like
```
thread 'tests::telemetry::metrics_endpoint_inbound_request_count' panicked at 'malformed metric (multiple '=' in label)
 in label: "target_addr=\"127.0.0.1:34111\"tls=\"disabled\""
   metric: "request_total{direction="inbound",authority="tele.test.svc.cluster.local",target_addr="127.0.0.1:34111"tls="disabled"} 1"
  retried 5 times (73.354973ms)', linkerd/app/integration/src/metrics.rs:194:51
```

Good catch!

linkerd/app/integration/src/tests/telemetry.rs
```diff
@@ -143,31 +143,65 @@ impl TcpFixture {
...
+}
+
+#[tokio::test]
+async fn metrics_endpoint_outbound_response_count() {
+    test_http_count("response_total", Fixture::outbound(), |proxy| {
+        let orig_dst = proxy.outbound_server.as_ref().unwrap().addr;
+        metrics::labels()
+            .label("direction", "outbound")
+            .label("tls", "no_identity")
+            .label("no_tls_reason", "not_provided_by_service_discovery")
```
olix0r, line 193:
```suggestion
            .label("tls", "disabled")
```

olix0r (approved):

PR #868, opened 2021-01-22
admin: stackify admin server

hawkw:
This branch refactors the proxy admin endpoint to build a server stack
using the same primitives used for the proxy servers. This is intended
primarily to allow us to easily add metrics to the admin server by just
slotting in the existing proxy metrics layers.

This change also has some side benefits:

* We now reuse more existing code for building the admin server, rather
  than writing separate versions of our accept & serve HTTP code
  specifically for admin,
* We now support HTTP/2 for the admin server :)

### DIFF
--- a/linkerd/app/src/admin.rs
+++ b/linkerd/app/src/admin.rs
@@ -1,6 +1,7 @@
-use crate::identity::LocalCrtKey;
+use crate::{http, identity::LocalCrtKey, inbound, svc};
 use linkerd_app_core::{
-    admin, config::ServerConfig, drain, metrics::FmtMetrics, serve, tls, trace, Error,
+    admin, config::ServerConfig, detect, drain, metrics::FmtMetrics, serve, tls, trace,
+    transport::listen, Error,
 };
 use std::{net::SocketAddr, pin::Pin, time::Duration};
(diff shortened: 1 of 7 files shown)


linkerd/app/src/admin.rs
```diff
@@ -27,22 +31,45 @@ impl Config {
...
+                    .push(http::BoxResponse::layer())
+                    .push(svc::MapErrLayer::new(Error::from)),
+            )
+            .check_new_clone::<(http::Version, inbound::TcpAccept)>()
+            .push(http::NewServeHttp::layer(Default::default(), drain.clone()))
+            .push(svc::NewUnwrapOr::layer(
+                svc::Fail::<_, AdminHttpOnly>::default(),
+            ))
+            .push(detect::NewDetectService::timeout(
+                Duration::from_secs(1),
```
olix0r, line 52: nit: perhaps consitfy a DETECT_TIMEOUT to be used here and below in TLS detect/

olix0r (approved):
tioli: If we boxed the request body we could avoid having to make tracing/etc generic over the body type.

kleimkuhler (approved):

PR #869, opened 2021-01-22
tls: Report SNI values for non-Linkerd TLS

olix0r:
The proxy's TLS server parses the SNI value for all TLS connections.
This value is immediately discarded; but it's valuable information we
may want to repor to users.

This change modifies the TLS server to extract non-local SNI values and
pass them to the server as a stack parameter. This value is now exposed
as a metric & tap label, paired with a `tls="opaque"` label.

This change sets up extracing additional information on the TLS server
like ALPN values.

--

Note that each commit in this PR addresses a discrete problem. It may be easier
to review the commits individually.


linkerd/app/gateway/src/config.rs
```diff
@@ -59,7 +59,7 @@ impl svc::stack::Predicate<inbound::Target> for Allow {
 
     fn check(&mut self, target: inbound::Target) -> Result<NameAddr, Error> {
         // Skip discovery when the client does not have an identity.
-        if target.client_id.is_some() {
+        if let Conditional::Some(tls::server::Tls::Terminated { client_id: Some(_) }) = target.tls {
```
hawkw, line 62: nit/TIOLI: i might import `Tls` because this is quite long :woman_shrugging:

linkerd/tls/src/server/client_hello.rs
```diff
@@ -1,58 +1,48 @@
...
-/// message is available. TODO: Reject non-matching inputs earlier.
+/// message is available.
+///
+/// TODO: Reject non-matching inputs earlier.
 ///
 /// This assumes that the ClientHello is small and is sent in a single TLS
 /// record, which is what all reasonable implementations do. (If they were not
 /// to, they wouldn't interoperate with picky servers.)
-pub fn match_client_hello(input: &[u8], identity: &identity::Name) -> Match {
+pub fn parse_sni(input: &[u8]) -> Result<Option<ServerId>, Incomplete> {
```
hawkw, line 25: :+1: for making this a `Result`

linkerd/tls/src/server/mod.rs
```diff
@@ -196,20 +200,27 @@ where
...
-            let (peer_id, tls) = handshake(tls_config, PrefixedIo::from(io)).await?;
-            return Ok((peer_id, EitherIo::Right(tls)));
+            let (client_id, io) = handshake(tls_config, PrefixedIo::from(io)).await?;
+            let tls = Conditional::Some(Tls::Terminated { client_id });
+            return Ok((tls, EitherIo::Right(io)));
         }
 
-        conditional_accept::Match::NotMatched => {
+        Ok(Some(sni)) => {
+            trace!("Identified non-matching SNI via peek");
```
hawkw, line 213: nit: should this include the SNI? I guess the `parse_sni` function also logs it, but this module's logs might be enabled while `client_hello`'s might not be...

hawkw (approved):
looks great — i expected this one to be a somewhat bigger change than it ended up being!

PR #871, opened 2021-01-23
admin: add metrics

hawkw:
This PR adds HTTP request/response, transport, and stack metrics for the
admin server to the proxy's inbound metrics. The `target_addr` labels
added in #861 and #866 disambiguate these metrics from other inbound
metrics.

I've also added new tests for admin server metrics.

Since the admin server serves a known set of HTTP paths, I'd like to add
per-path metrics to the admin metrics as well. We could do this by using
the existing service profiles code to add profile metrics to the admin
server with a hard-coded service profile. As this is probably a bigger
change, I'll work on this in a follow-up branch.

### DIFF
--- a/linkerd/app/src/admin.rs
+++ b/linkerd/app/src/admin.rs
@@ -51,11 +51,6 @@ impl Config {
                 svc::layers()
                     .push(metrics.http_errors.clone())
                     .push(errors::layer())
-                    .push(
-                        metrics
-                            .stack
-                            .layer(metrics::StackLabels::inbound("http", "admin")),
-                    )
                     .push(http::BoxResponse::layer()),
             )
             .push(svc::stack::MapTargetLayer::new(|(http_version, tcp)| {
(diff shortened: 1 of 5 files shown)


linkerd/app/src/admin.rs
```diff
@@ -40,12 +46,29 @@ impl Config {
...
                 svc::layers()
-                    .push(http::BoxResponse::layer())
-                    .push(svc::MapErrLayer::new(Error::from)),
+                    .push(metrics.http_errors.clone())
+                    .push(errors::layer())
+                    .push(
+                        metrics
+                            .stack
+                            .layer(metrics::StackLabels::inbound("http", "admin")),
+                    )
```
olix0r, line 58:
```suggestion
```

I don't really think we need these diagnostics.

olix0r (approved):
lgtm

olix0r:
>  I'd like to add per-path metrics to the admin metrics as well.

We should probably hold off on this for now. We don't have any story about consuming them, they come with some cost (i.e. in prometheus), and we probably have some higher priority work to do next.

PR #872, opened 2021-01-23
inbound: Split stack into modules

olix0r:
This change splits up the inbound stack as follows:

commit c32b5cc114325e5506eb0a6ea65c517658356cbd

    inbound: Decouple HTTP server from router
    
    This change moves the RequestTarget router from the http_server stack to
    the http_router and direct stacks.
    
    This sets up decoupling the target types for these stacks, as both
    stacks should share HTTP server logic but may route requests
    differently.

commit c739c29ed0f5ef0f6da01139c818e9d66591fca9

    inbound: Extract stacks into a separate files
    
    This change creates new inbound::http and inbound::direct modules that
    handle stack constructions.
    
    No functional change.

### DIFF
--- a/linkerd/app/inbound/src/direct.rs
+++ b/linkerd/app/inbound/src/direct.rs
@@ -1,0 +1,135 @@
+use crate::{
+    http,
+    require_identity::RequireIdentityForDirect,
+    target::{RequestTarget, Target, TcpAccept, TcpEndpoint},
+};
+use linkerd_app_core::{
+    config::ProxyConfig,
+    detect, drain, io, metrics,
+    opencensus::proto::trace::v1 as oc,
+    proxy::identity::LocalCrtKey,
+    svc, tls,
+    transport::{listen, metrics::SensorIo},
+    transport_header::{DetectHeader, TransportHeader},
+    Error, DST_OVERRIDE_HEADER,
+};
+use std::fmt::Debug;
+use tokio::sync::mpsc;
+
+#[derive(Default)]
+struct NonOpaqueRefused(());
+
+type FwdIo<I> = io::PrefixedIo<SensorIo<tls::server::Io<I>>>;
+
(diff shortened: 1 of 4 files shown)


linkerd/app/inbound/src/http.rs
```diff
@@ -0,0 +1,207 @@
...
+    proxy::{http, tap},
+    reconnect,
+    spans::SpanConverter,
+    svc, Error, NameAddr, TraceContext, DST_OVERRIDE_HEADER,
+};
+use std::net::SocketAddr;
+use tokio::sync::mpsc;
+use tracing::debug_span;
+
+pub fn server<T, I, H, HSvc>(
```
hawkw, line 22: might be nice to have summary comments like on the direct stack?

linkerd/app/inbound/src/direct.rs
```diff
@@ -0,0 +1,135 @@
...
+                .push_spawn_buffer(config.buffer_capacity),
+        )
+        .check_new_service::<Target, http::Request<http::BoxBody>>()
+        // Removes the override header after it has been used to
+        // determine a request target.
+        .push_on_response(http::strip_header::request::layer(DST_OVERRIDE_HEADER))
+        // Routes each request to a target, obtains a service for that
+        // target, and dispatches the request.
+        .instrument_from_target()
+        .push(svc::NewRouter::layer(RequestTarget::from))
```
hawkw, line 75: TIOLI: does it make sense to bundle the router layer, strip header, and instrument layers together somewhere? the direct and http router stacks both have the same configuration.

hawkw (approved):
lgtm!

linkerd/app/inbound/src/direct.rs
olix0r: It's going to change on the direct stack, so no.

In general: [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).

PR #873, opened 2021-01-24
test: Disable tracing-subscriber by default

olix0r:
When running tests, we encounter panics like:

```
tcache_thread_shutdown(): unaligned tcache chunk detected
```

Disabling tracing-subscriber eliminates these panics. This change
disables the tracing subscriber for these tests until the underlying
issue is resolved.


hawkw (approved):
fine w/ me. 

we may want to try switching to just using a global subscriber in tests rather than none at all, but i can experiment with that separately.

PR #874, opened 2021-01-25
Use TLS ALPN to negotiate transport header support

olix0r:
An outbound proxy may not always know a priori whether a destination
supports transport headers. This is especially the case when
communicating with remote multi-cluster gateways, where the target
cluster may not be running a compatible proxy version. We can detect
this at runtime by using TLS's Application-Layer Protocol Negotiation
(ALPN) to determine whether the target proxy supports transport headers.

To accomplish this:

1. The inbound direct stack is modified to add ALPN configuration when
   accepting direct traffic. No ALPN is advertised when accepting
   connections on ports other than the proxy's inbound port.
2. The TLS server includes the negotiated protocol in the `ServerTls`
   stack target.
3. The TLS client's target type is modified to include an optional list
   of ALPN protocols. When this is configured by the stack target, the
   client modifies its TLS configuration to include these ALPN
   protocols.
4. The `tls::HasNegotiatedProtocol` trait is introduced and implemented
   for our IO types. This enables stack modules to inspect a
   connection for an ALPN-negotiated protocol. This will only be set
   when both the client and server include common protocol
   advertisements.
5. The outbound `Endpoint` type is modified to advertise support for
   transport headers when an opaque hint is provided OR an authority
   override is set (i.e., for gateways).
6. The outbound `OpaqueTransport` module is modified to account for
   authority overrides when constructing a `TransportHeader`; and it
   uses the connection's negotiated protocol to determine whether a
   header should be written. This modules tests have been updated to
   exercise mocked ALPN discovery.

This has been tested with server-first protocols. Inbound gateway
support has not yet been implemented and will follow this change.

### DIFF
--- a/linkerd/app/inbound/src/direct.rs
+++ b/linkerd/app/inbound/src/direct.rs
@@ -16,13 +16,15 @@ use linkerd_app_core::{
 use std::fmt::Debug;
 use tokio::sync::mpsc;
 
-#[derive(Debug, Default)]
-struct NonOpaqueRefused(());
-
 #[derive(Clone, Debug)]
 struct WithTransportHeaderAlpn(LocalCrtKey);
 
 type FwdIo<I> = io::PrefixedIo<SensorIo<tls::server::Io<I>>>;
+
+/// Creates I/O errors when a connection cannot be forwarded because no transport
+/// header was present.
+#[derive(Debug, Default)]
+struct RefusedNoHeader(());
 
 /// Builds a stack that handles connections that target the proxy's inbound port
 /// (i.e. without an SO_ORIGINAL_DST setting). This port behaves differently from
@@ -81,7 +83,7 @@ where
     let http_detect = svc::stack(http::server(&config, http, metrics, span_sink, drain))
         .push_cache(config.cache_max_idle_age)
         .push(svc::NewUnwrapOr::layer(
-            svc::Fail::<_, NonOpaqueRefused>::default(),
+            svc::Fail::<_, RefusedNoHeader>::default(),
         ))
         .push(detect::NewDetectService::timeout(
             detect_timeout,
@@ -102,9 +104,16 @@ where
             t.target_addr = (t.target_addr.ip(), h.port).into();
             t
         })
+        // We always try to dect a protocol header. We _can_ know whether it's
+        // expected based on the serverside ALPN (passed via the target), but
+        // it's easier to just do detection and handle the case when it's not
+        // present as an exception.
+        //
         // HTTP detection must *only* be performed when a transport
         // header is absent. When the header is present, we must
         // assume the protocol is opaque.
+        //
+        // TODO Stop supporting headerless connections after stable-2.10.
         .push(svc::NewUnwrapOr::layer(http_detect))
         .push(detect::NewDetectService::timeout(
             detect_timeout,
@@ -130,7 +139,9 @@ impl Into<tls::server::Config> for &'_ WithTransportHeaderAlpn {
     fn into(self) -> tls::server::Config {
         // Copy the underlying TLS config and set an ALPN value.
         //
-        // TODO: Avoid cloning the server config for every connection.
+        // TODO: Avoid cloning the server config for every connection. It would
+        // be preferable if rustls::ServerConfig wrapped individual fields in an
+        // Arc so they could be overridden independently.
         let c: tls::server::Config = (&self.0).into();
         let mut config = c.as_ref().clone();
         config
@@ -146,9 +157,9 @@ impl Into<tls::LocalId> for &'_ WithTransportHeaderAlpn {
     }
 }
 
-// === impl NonOpaqueRefused ===
+// === impl RefusedNoHeader ===
 
-impl Into<Error> for NonOpaqueRefused {
+impl Into<Error> for RefusedNoHeader {
     fn into(self) -> Error {
         Error::from(io::Error::new(
             io::ErrorKind::ConnectionRefused,
--- a/linkerd/app/inbound/src/direct.rs
+++ b/linkerd/app/inbound/src/direct.rs
@@ -16,8 +16,11 @@ use linkerd_app_core::{
 use std::fmt::Debug;
 use tokio::sync::mpsc;
 
-#[derive(Default)]
+#[derive(Debug, Default)]
 struct NonOpaqueRefused(());
+
+#[derive(Clone, Debug)]
+struct WithTransportHeaderAlpn(LocalCrtKey);
 
 type FwdIo<I> = io::PrefixedIo<SensorIo<tls::server::Io<I>>>;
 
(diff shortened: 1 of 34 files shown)


linkerd/app/inbound/src/direct.rs
```diff
@@ -99,9 +104,16 @@ where
             t.target_addr = (t.target_addr.ip(), h.port).into();
             t
         })
+        // We always try to dect a protocol header. We _can_ know whether it's
```
hawkw, line 107: typo:
```suggestion
        // We always try to detect a protocol header. We _can_ know whether it's
```

linkerd/app/inbound/src/direct.rs
```diff
@@ -99,9 +104,16 @@ where
...
+        // expected based on the serverside ALPN (passed via the target), but
+        // it's easier to just do detection and handle the case when it's not
+        // present as an exception.
+        //
         // HTTP detection must *only* be performed when a transport
         // header is absent. When the header is present, we must
         // assume the protocol is opaque.
+        //
+        // TODO Stop supporting headerless connections after stable-2.10.
```
hawkw, line 116: maybe worth a tracking ticket?

linkerd/app/inbound/src/direct.rs
```diff
@@ -114,18 +126,40 @@ where
...
-// === impl NonOpaqueRefused ===
+// === impl WithTransportHeaderAlpn ===
+
+impl Into<tls::server::Config> for &'_ WithTransportHeaderAlpn {
+    fn into(self) -> tls::server::Config {
+        // Copy the underlying TLS config and set an ALPN value.
+        //
+        // TODO: Avoid cloning the server config for every connection. It would
+        // be preferable if rustls::ServerConfig wrapped individual fields in an
+        // Arc so they could be overridden independently.
```
hawkw, line 144: does this need an upstream issue?

linkerd/app/outbound/src/tcp/opaque_transport.rs
```diff
@@ -20,16 +18,27 @@ pub struct OpaqueTransport<S> {
...
-    pub fn layer() -> impl layer::Layer<S, Service = Self> + Copy {
-        layer::mk(|inner| OpaqueTransport { inner })
+    pub fn layer() -> impl svc::Layer<S, Service = Self> + Copy {
+        svc::layer::mk(|inner| OpaqueTransport { inner })
+    }
+
+    /// Determines whether the connection has negotiated support for the
+    /// transport header.
+    #[inline]
+    fn negotiated_header<I: tls::HasNegotiatedProtocol>(io: &I) -> bool {
```
hawkw, line 28: nit/TIOLI: since this returns a bool i might prefer `has_negotiated_header` or something?

linkerd/app/outbound/src/tcp/opaque_transport.rs
```diff
@@ -42,54 +51,51 @@ where
...
-                                name: Some(n),
-                            })
-                    })
-                    .unwrap_or(TransportHeader {
-                        port: orig_port,
-                        name: None,
-                    });
-                debug!(?header, override_port, "Using opaque transport");
-                Some(header)
+        // Configure the target port from the endpoing. In opaque cases, this is
```
hawkw, line 54: typo:
```suggestion
        // Configure the target port from the endpoint. In opaque cases, this is
```

linkerd/tls/src/client.rs
```diff
@@ -38,11 +46,16 @@ pub enum NoClientTls {
...
-pub type ConditionalClientTls = Conditional<ServerId, NoClientTls>;
+/// A stack paramater that indicates whether the target server endpoint has a
+/// known TLS identity.
+pub type ConditionalClientTls = Conditional<ClientTls, NoClientTls>;
 
 pub type Config = Arc<rustls::ClientConfig>;
 
+/// A stack param that configures the available ALPN protocols.
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct AlpnProtocols(pub Vec<Vec<u8>>);
```
hawkw, line 57: I guess the `Vec` representation is locked in due to the `rustls` API? It would be nice to not have to clone the bytes around constantly. But, it's probably not big enough to be terribly important.

linkerd/tls/src/client.rs
```diff
@@ -38,11 +46,16 @@ pub enum NoClientTls {
...
-/// Indicates whether the target server endpoint has a known TLS identity.
-pub type ConditionalClientTls = Conditional<ServerId, NoClientTls>;
+/// A stack paramater that indicates whether the target server endpoint has a
+/// known TLS identity.
+pub type ConditionalClientTls = Conditional<ClientTls, NoClientTls>;
 
 pub type Config = Arc<rustls::ClientConfig>;
 
+/// A stack param that configures the available ALPN protocols.
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
```
hawkw, line 56: Does it make sense for this to have a manual `Debug` impl instead, so that each protocol can be formatted using `NegotiatedProtocolRef` or something, rather than as a pile of bytes?

linkerd/app/outbound/src/target.rs
```diff
@@ -283,13 +277,38 @@ impl<P> Into<metrics::EndpointLabels> for &'_ Endpoint<P> {
...
+        // header.
+        let alpn = if metadata.opaque_transport_port().is_some()
+            || metadata.authority_override().is_some()
+        {
+            Some(tls::client::AlpnProtocols(vec![
+                transport_header::PROTOCOL.into()
+            ]))
+        } else {
+            None
+        };
```
hawkw, line 300: nit: it looks like we construct this vec regardless of whether or not the metadata has an identity, but if it doesn't have one, we always return `Conditional::None`. Maybe we should refactor this so that the `AlpnProtocols` is only constructed inside the map closure so we don't always build the vec?

probably not a huge deal.

linkerd/tls/src/client.rs
```diff
@@ -85,26 +109,45 @@ where
...
+                // configuration and set the protocols. Otherwise, just use the
+                // TLS configuration without modification.
+                //
+                // TODO it would be better to avoid cloning the whole TLS config
+                // per-connection.
+                match alpn {
+                    None => tokio_rustls::TlsConnector::from(local.into()),
+                    Some(AlpnProtocols(protocols)) => {
+                        let mut config: rustls::ClientConfig = local.into().as_ref().clone();
+                        config.alpn_protocols = protocols;
```
hawkw, line 132: take it or leave it: might be worth a comment or something making it explicit that this is where the ALPN protocols are actually handed over to `rustls`? I feel like I'm beginning to see the value of making the places where some configuration is handed over to a library really explicit for debugging purposes later on.

hawkw (approved):
:+1: this is great! i had a few minor nitpicks.

linkerd/app/inbound/src/direct.rs
olix0r: https://github.com/ctz/rustls/issues/431 was closed the other day, so there may not be interest in doing this. We can probably improve the situation internally but it's too big of a change to worry bout for now. Per-connection cloning in special cases isn't great but it won't kill us.

hawkw: :/ yeah, i don't think it's a *huge* deal. it might be possible to make the issue a little clearer to the rustls folks?

linkerd/app/outbound/src/target.rs
olix0r: note that both of these cases practically require identity to be set (gateway and opaque transport) anyway; but i've fixed this for clarity

PR #876, opened 2021-01-26
stack: Introduce the Param trait

olix0r:
When configuring a stack module, we are usually generic over the target
type, yet we need static assurances about the target's ability to
produce configuration values. Our current approach is to use awkward
type constraints like:

    for<'t> &'t T: Into<Config>

which is invoked as:

    (&target).into()

This is cumbersome and doesn't clearly express intent, especially in
implementations.

This change formalizes this pattern under a dedicated trait,
`stack::Param`. Stack modules are updated to express constraints like:

    T: Param<Config>

and is invoked as:

    target.param()

This change includes one minor change in functionality: the identity
client's TLS configuration is now obtained the module's `LocalCrtKey`
instead of its `TrustAnchors`. When using `TrustAnchors`, the client
would never initiate a client-authentcated mTLS session to the identity
service. When using the `LocalCrtKey`, the local identity may be used to
establish mTLS sessions when refreshing existing certificates.


linkerd/stack/src/lib.rs
```diff
@@ -38,3 +38,20 @@ pub use self::{
 };
 pub use tower::util::Either;
 pub use tower::util::{future_service, FutureService};
+
+/// Describes a stack target that can produce `T` typed parameters.
+///
+/// Stacks (usually layered `NewService` implementations) frequently need to be
+/// able to obtain configuration from the stack target, but stack modules are
+/// decoupled from any concrate target types. The `Param` trait
```
hawkw, line 46: the param trait _what_? :)

linkerd/stack/src/lib.rs
```diff
@@ -38,3 +38,20 @@ pub use self::{
...
+pub trait Param<T> {
+    /// Produces a `T`-typed stack paramter.
+    fn param(&self) -> T;
+}
+
+/// The identity `Param`.
+impl<T: Clone> Param<T> for T {
```
hawkw, line 53: wonder if it's also worth having `Param` impls for types implementing `ToOwned`, like
```rust
impl<T: ToOwned> Param<T::Owned> for T
```
but that might get conflicty?

linkerd/stack/src/lib.rs
hawkw, line 46: the param trait _what_? :)

linkerd/stack/src/lib.rs
```diff
@@ -38,3 +38,20 @@ pub use self::{
...
+
+/// Describes a stack target that can produce `T` typed parameters.
+///
+/// Stacks (usually layered `NewService` implementations) frequently need to be
+/// able to obtain configuration from the stack target, but stack modules are
+/// decoupled from any concrate target types. The `Param` trait
+pub trait Param<T> {
+    /// Produces a `T`-typed stack paramter.
+    fn param(&self) -> T;
+}
```
hawkw, line 50: Do we care about being able to have `Param` impls where `T` is borrowed from `Self`? In order to support that we'd need to change this to something like
```rust
pub trait Param<'a, T> {
   fn param(&'a self) -> T;
}
```
so you can write impls like 
```rust
pub struct Foo {
    bar: Bar,
}

impl<'a> Param<&'a Bar> for Foo {
    fn param(&'a self) -> &'a Bar {
       &self.bar
    }
}
```

On the other hand that complicates both the implementation and the usage --- the `for<'a>` probably has to come back, and getting rid of that was part of the motivation for this change...

linkerd/stack/src/lib.rs
```diff
@@ -38,3 +38,20 @@ pub use self::{
...
+
+/// The identity `Param`.
+impl<T: Clone> Param<T> for T {
+    fn param(&self) -> T {
+        self.clone()
+    }
+}
```
hawkw, line 57: Should there also maybe be:
```rust
impl<T: ToOwned> Param<T::Owned> for T
```
provided it's not conflicty (which it might be).

hawkw (approved):
:clap:it's:clap:happening:clap:
image

linkerd/stack/src/lib.rs
olix0r: It's conflicty, but I think it's probably better to implement the identity param on ToOwned, which itself has a blanket impl for Clone types!

linkerd/stack/src/lib.rs
olix0r: I don't think it's worth the complication. If we find a need for it, we can introduce a `ParamRef` trait; but, generally, we accept that target types are cloned freely, so it's up to the types to make that cheap (i.e. by using Arc where appropriate).

PR #877, opened 2021-01-26
transport-header: Encode session protocol

olix0r:
In order to transport both opaque and non-opaque protocols with the
transport header, especially for multicluster gateways, the inbound
proxy needs to be able to discover whether the session protocol is known
by the client.

This change adds a field and enum type to the transport header protobuf
message and wires it through so the outbound proxy parameterizes its
target type on the session protocol.a

Inbound handling of this field is not yet implemented and will be added
in a followup change.


linkerd/app/outbound/src/tcp/opaque_transport.rs
```diff
@@ -80,6 +83,8 @@ where
             }
         }
 
+        let protocol: Option<SessionProtocol> = ep.param();
```
hawkw, line 86: this makes me wonder how possible it would be to add a `Parse`-style handshake trait for `Param`, so you can write
```rust
let protocol = ep.get_param::<Option<SessionProtocol>>();
```
instead. not a blocker, but it might make some code easier to write such as when a param is part of a method chain

linkerd/transport-header/proto/header.proto
```diff
@@ -8,4 +8,18 @@ message Header {
...
   string name = 2;
+
+  // The session protocol, if one is known. When no protocol is specified, the
+  // connection is handled opaquely.
+  SessionProtocol session_protocol = 3;
+}
+
+message SessionProtocol {
+  message Http1 {}
+  message H2 {}
```
hawkw, line 19: nit/TIOLI: it bothers me slightly that we have "http 1" and "h2", can we make these consistent? not a big deal...

linkerd/transport-header/src/lib.rs
```diff
@@ -14,13 +14,21 @@ mod proto {
...
     /// The logical name of the target (service), if one is known.
     pub name: Option<Name>,
+
+    pub protocol: Option<SessionProtocol>,
+}
+
+#[derive(Clone, Debug, PartialEq, Hash)]
+pub enum SessionProtocol {
+    Http1,
+    H2,
```
hawkw, line 31: nit/TIOLI: is there any particular reason we add a new `SessionProtocol` enum, versus just converting the protobuf session protocol enum to/from `http::Version` directly? that seems like it could be simpler.

i suppose we might want to add non-http session protocols in the future...but it might be simpler to just separate them _then_? i could go either way on this, it's just a thought.

linkerd/transport-header/proto/header.proto
hawkw: ...i guess we already use this naming scheme with `http::Version`, so it's whatever i guess.

hawkw (approved):
lgtm overall!

linkerd/transport-header/src/lib.rs
olix0r: I don't think the transport-header module should have a dependency on proxy-http. i'd prefer to keep these things domain-specificon the assumption this will grow to include non-HTTP things.

linkerd/app/outbound/src/tcp/opaque_transport.rs
olix0r: Seems like basically the same amount of typing, so i'm -0 on it for now.

linkerd/transport-header/src/lib.rs
hawkw: ah, yeah, that makes sense. carry on 👍

PR #878, opened 2021-01-27
profiles: Use a LogicalAddr param type

olix0r:
The profiles stack expects targets to satisfy `Param<Addr>`, which is
ambiguous. This change introduces a `profiles::LogicalAddr` type and updates
the target types to provide `Param<profiles::LogicalAddr>`.

### DIFF
--- a/linkerd/app/inbound/src/target.rs
+++ b/linkerd/app/inbound/src/target.rs
@@ -153,9 +153,9 @@ pub(super) fn route((route, logical): (profiles::http::Route, Logical)) -> dst::
 
 // === impl Target ===
 
-impl Param<profiles::Logical> for Target {
-    fn param(&self) -> profiles::Logical {
-        profiles::Logical(self.dst.clone())
+impl Param<profiles::LogicalAddr> for Target {
+    fn param(&self) -> profiles::LogicalAddr {
+        profiles::LogicalAddr(self.dst.clone())
     }
 }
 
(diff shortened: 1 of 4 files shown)


linkerd/service-profiles/src/split.rs
```diff
@@ -79,10 +79,8 @@ where
             Some(rx) => {
                 let mut targets = rx.borrow().targets.clone();
                 if targets.is_empty() {
-                    targets.push(Target {
-                        addr: target.param(),
-                        weight: 1,
-                    })
+                    let LogicalAddr(addr) = target.param();
```
hawkw, line 82: image

linkerd/app/inbound/src/target.rs
```diff
@@ -153,23 +153,9 @@ pub(super) fn route((route, logical): (profiles::http::Route, Logical)) -> dst::
...
-/// Used for profile discovery.
-impl Param<SocketAddr> for Target {
-    fn param(&self) -> SocketAddr {
-        self.target_addr
-    }
-}
-
-impl Param<tls::ConditionalClientTls> for Target {
-    fn param(&self) -> tls::ConditionalClientTls {
-        Conditional::None(tls::NoClientTls::Loopback)
```
hawkw, line 172: was this one just totally unused?

hawkw (commented):
:+1:

linkerd/app/inbound/src/target.rs
olix0r: yep

hawkw (approved):

PR #879, opened 2021-01-27
transport: Add a connect::Addr parameter type

olix0r:
The `connect` module currently depends on a `SocketAddr` parameter, but
`SocketAddr` is domain agnostic (i.e., it can refer to source addresses,
local addresses, or target addresses).

This change introduces a `connect::Addr` type to make target
implementations clearer.


linkerd/app/inbound/src/lib.rs
```diff
@@ -54,7 +54,7 @@ pub fn tcp_connect<T: Param<u16>>(
     // Establishes connections to remote peers (for both TCP
     // forwarding and HTTP proxying).
     svc::stack(transport::ConnectTcp::new(config.keepalive))
-        .push_map_target(|t: T| SocketAddr::from(([127, 0, 0, 1], t.param())))
+        .push_map_target(|t: T| transport::ConnectAddr(([127, 0, 0, 1], t.param()).into()))
```
hawkw, line 57: TIOLI: we could consider adding a 
```rust
impl ConnectAddr {
    pub fn local<T: Param<u16>(port: &T> -> Self {
        Self(([127, 0, 0, 1], port.param()).into())
    }
}
```
or something, and just using that here; might be a little clearer:
```suggestion
        .push_map_target(transport::ConnectAddr::local)
```

hawkw: also --- and this is a totally unrelated change, so it might be better off in a separate PR, but we could maybe consider newtyping ports so we can write `Param<Port>` rather than `Param<u16>` --- might also convey intent better. nbd though.

hawkw (approved):

linkerd/app/inbound/src/lib.rs
olix0r: the localhost special-case is specific to inbound, so i'd rather not have that in transport. The `Port` param type is a good idea but I'll probably do that as part of inbound changes.

PR #880, opened 2021-01-27
stack: Replace switch with Filter and NewEither

olix0r:
When switching between stacks based on target values, we frequently want
to alter the target type so that the primary and secondary stacks may be
parameterized on different target types. This is the case with the
`unwrap_or` stack helper, for instance.

This change removes the `Switch` trait and module, replacing uses with a
mix of `Fitler`, which modifies target types, and a new `NewEither` helper,
which takes `Either`-typed targets.

This change renames `NewUnwrapOr` as `UnwrapOr` and simplifies its
implementation to use `Filter` and `NewEither`

This change also removes various unused `MakeService` implementations.


linkerd/app/outbound/src/target.rs
```diff
@@ -37,6 +37,9 @@ pub struct Endpoint<P> {
     pub concrete: Concrete<P>,
 }
 
+#[derive(Clone, Copy, Debug)]
+pub struct ShouldResolve;
```
hawkw, line 41: nit/TIOLI: i think this could have stayed a free fn and still worked with the filter interface?

hawkw (approved):
This looks good to me.

One potential concern is that passing target types in/out of the predicates rather than passing in a pointer and passing out a bool could result in copying more data around on the stack than we did previously, but I doubt this is noticeable unless target types are _extremely_ large. And this code is not that hot anyway.

PR #881, opened 2021-01-27
inbound: normalize URIs after downgrading to HTTP/1

hawkw:
Somewhere between stable-2.8.1 and stable-2.9.2, we introduced a bug
when downgrading HTTP/1 traffic that originated in origin form. See
linkerd/linkerd2#5298 for details.

Hyper's `Client` type requires all URIs to be in absolute form when
constructing the request, because the URI is used as a key for the
connection pool and to discover connect targets, but will write the
request out on the wire in origin-form unless explicitly configured not
to. Therefore, we normalize all HTTP/1 URIs to absolute-form before
handing them over to Hyper's `Client`, but only configure it to actually
send the request in absolute form if the request was received in
absolute-form. The `NormalizeUri` layer performs the absolute form
normalization and sets a request extension of the request was in
absolute-form originally.

The bug is that, when an orig-proto downgrade request is received in
origin form, it no longer gets normalized. It turns out that this is
because the `NormalizeUri` layer is in the inbound stack _above_ the
`orig_proto::Downgrade` layer --- so the request is still HTTP/2 when it
hits `NormalizeUri`, and is therefore not normalized.

This branch fixes the issue by separating the normalization from setting
the `WasAbsoluteForm` request extension, and moving the `NormalizeUri`
layer _after_ downgrade. The request extension is now set by a new
`DetectAbsoluteForm` layer.

I've also added new stack tests for inbound orig-proto requests in both
absolute- and origin-form. The origin-form test would have found the
issue, since it hits a debug assertion that requests are in
absolute-form, which none of the existing tests hit. Also, since inbound
didn't previously add any stack tests, I also added a simple unmeshed
http1 hello world test as well. This was mainly added as part of the
process for writing other stack tests: I wanted to have a working basic
request test before trying to write a test to reproduce the specific
issue.

Finally, I generalized a bunch of the utilities that were written for
the outbound stack tests so that they can also be used in inbound.

Fixes linkerd/linkerd2#5298.


hawkw:
I'd also like to clean up some of the test utils a bit now that they're used in more palces, but will probably do that separately.

linkerd/proxy/http/src/normalize_uri.rs
```diff
@@ -13,7 +13,7 @@
 //!   from modified requests;
 //! * Otherwise, if the request has a `Host` header, it is used as the authority;
 //! * Otherwise, the target's address is used (as provided by the target).
-
+#![allow(dead_code)]
```
olix0r, line 16: ?

linkerd/proxy/http/src/normalize_uri.rs
```diff
@@ -36,6 +36,13 @@ pub struct NormalizeUri<S> {
     default: http::uri::Authority,
 }
 
+/// Detects the original form of a request URI and inserts a `WasAbsoluteForm`
+/// extension.
+#[derive(Clone, Debug)]
+pub struct DetectAbsoluteForm<S> {
```
olix0r, line 42: prefer something like `MarkAbsoluteForm`... We tend to use `Detect...` for connection-oriented things.

linkerd/proxy/http/src/normalize_uri.rs
```diff
@@ -112,10 +119,7 @@ where
 
     fn call(&mut self, mut req: http::Request<B>) -> Self::Future {
         if let http::Version::HTTP_10 | http::Version::HTTP_11 = req.version() {
-            if h1::is_absolute_form(req.uri()) {
-                trace!(uri = ?req.uri(), "Absolute");
-                req.extensions_mut().insert(h1::WasAbsoluteForm(()));
-            } else if req.uri().authority().is_none() {
+            if !h1::is_absolute_form(req.uri()) && req.uri().authority().is_none() {
```
olix0r, line 122: Should this check the `WasAbsoluteForm` marker instead?

linkerd/proxy/http/src/normalize_uri.rs
```diff
@@ -126,3 +130,41 @@ where
...
+
+impl<S, B> tower::Service<http::Request<B>> for DetectAbsoluteForm<S>
+where
+    S: tower::Service<http::Request<B>>,
+{
+    type Response = S::Response;
+    type Error = S::Error;
+    type Future = S::Future;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), S::Error>> {
```
olix0r, line 154: inline?

linkerd/app/inbound/src/http/tests.rs
```diff
@@ -0,0 +1,222 @@
+#![allow(unused_imports)]
```
olix0r, line 1: rm?

olix0r (commented):
a few notes. i'm going to push fixes for the easy stuff

linkerd/proxy/http/src/normalize_uri.rs
hawkw: naming this thing was probably one of the harder parts of the PR honestly :)

olix0r (approved):

PR #883, opened 2021-01-28
Use transport header protocol hint for HTTP gateway

olix0r:
122d263d introduced session protocol hints to the transport header, but
the inbound direct stack was not updated to use this hint when deciding
whether to pass connections to the TCP gateway.

This change:

1. Updates the inbound direct stack to use the header-provided hint so
   that, when set, HTTP detection can simply be skipped -- the value is
   used to inform the HTTP server directly.
2. The outbound proxy could incorrectly set the protocol hint to the
   original protocol, ignoring HTTP/2 upgrade hinting. This has been
   fixed.

These changes have been tested against Linkerd2's multicluster
integration tests.

### DIFF
--- a/linkerd/app/inbound/src/direct.rs
+++ b/linkerd/app/inbound/src/direct.rs
@@ -26,6 +26,9 @@ type FwdIo<I> = io::PrefixedIo<SensorIo<tls::server::Io<I>>>;
 #[derive(Debug, Default)]
 struct RefusedNoHeader(());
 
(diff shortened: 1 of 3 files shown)


olix0r:
The inbound gateway logic is in a transitional state, so I plan to continue improving it for clarity/correctness; but this updates the proxy to be releasable for an edge release.

kleimkuhler (approved):
Looks good 👍

PR #886, opened 2021-01-29
http: Parameterize normalize_uri::DefaultAuthority

olix0r:
The NormalizeUri module needs is configured with a fallback authority in
case we're dealing with poorly-formed HTTP requests (i.e., HTTP/1.0).
Currently, the module is parameterized on a `SocketAddr`, because on the
outbound side this makes sense as a default value. But on the inbound
side, especially in a gateway configuration, it doesn't make sense to
support a default value. It's better to just fail the request early.

This change updates the NormalizeUri type to require its targets to be
parameterized on a DefaultAuthority type. This type provides an
_optional_ `http::uri::Authority` so the target may provide a named
value (i.e., from a service profile response) when appropriate. It will
also enable us to omit a default when appropriate.

### DIFF
--- a/linkerd/proxy/http/src/normalize_uri.rs
+++ b/linkerd/proxy/http/src/normalize_uri.rs
@@ -15,12 +15,11 @@
 //! * Otherwise, the target's address is used (as provided by the target).
 
 use super::h1;
+use futures::{future, TryFutureExt};
+use http::uri::Authority;
+use linkerd_error::Error;
 use linkerd_stack::{layer, NewService, Param};
-use std::{
-    net::SocketAddr,
-    str::FromStr,
-    task::{Context, Poll},
-};
+use std::task::{Context, Poll};
 use tracing::trace;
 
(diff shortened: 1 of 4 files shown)


linkerd/proxy/http/src/normalize_uri.rs
```diff
@@ -55,57 +61,76 @@ impl<N> NewNormalizeUri<N> {
...
     }
 }
 
+// === impl NoAuthority ===
+
+impl std::fmt::Display for NoAuthority {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "failed to normalize URI because no authority could be determined"
```
hawkw, line 127: it might be worth including the bad URI in this message? but, that would mean carrying it in the error type

hawkw (approved):
looks good to me!

kleimkuhler (approved):

PR #887, opened 2021-01-29
http: Parameterize the HTTP server

olix0r:
The HTTP server requires that targets are in the form `(Version, T)`.
This is inflexible, though.

This change updates the HTTP server to instead require that targets
satisfy `Param<Version>` so that tuple types are not required. This
requires the introduction of a new HttpAccept type in the inbound proxy.


hawkw (approved):
no complaints! this feels a lot neater than passing tuples around.

PR #888, opened 2021-01-30
stack: Make the router fallible

olix0r:
The `Router` service looks at each request to produce a routing key; but
it may not always be possible to satisfy a key for al requests. In these
cases, we currently need to encode the error case into the key type and
fail request on its inner stack. This is needlessly cumbersome.

This change modifies the `router::RecognizeRoute` trait to return a
`Result` so that requests can be failed eagerly by the router.


kleimkuhler (approved):
👍

PR #889, opened 2021-01-30
Update http::Insert to use `Param`

olix0r:
The `http::insert` module provides a utility to insert the target value
into each request's HTTP extensions. But it's not always desirable to
insert the entire target. We may want only insert a single parameter
from the target.

This change replaces the `insert::target` module with an
`insert::NewInsert` type that uses `Param` to extract a value from the
target.

In doing this, I also noticed that 58c78596 changed the extension type
needed by tap::Inspect. This change fixes this and sets up further
improvements to use narrower-scoped param types so that this can't break
as easily moving forward.


kleimkuhler (approved):
Looks good!

PR #891, opened 2021-01-31
Track stack metrics within failfast

olix0r:
Our stack metrics are generally applied outside of failfast modules
so we have no insight into how long services are actually stuck in
pending.

This PR addresses this as follows:

1. The stack metrics `Service` impl no longer implements `Clone`. It
   doesn't really make sense to track idleness across clones. Stacks
   that require clone are no longer tracked by metrics.
2. `poll_total_ms` is now updated on each pending poll so that we need
   not wait until a state change to update the value; and
3. All applicable uses of failfast have been updated to include stack
   metrics. This gives us a better sense of the behavior of buffered
   services.

### DIFF
--- a/linkerd/app/inbound/src/direct.rs
+++ b/linkerd/app/inbound/src/direct.rs
@@ -70,6 +70,8 @@ where
     let http = svc::stack(http_loopback)
         .push_on_response(
             svc::layers()
+                .push(svc::layer::mk(svc::SpawnReady::new))
+                .push(metrics.stack.layer(crate::stack_labels("http", "gateway")))
                 .push(svc::FailFast::layer("Gateway", dispatch_timeout))
                 .push_spawn_buffer(config.buffer_capacity),
         )
--- a/linkerd/stack/metrics/src/service.rs
+++ b/linkerd/stack/metrics/src/service.rs
@@ -1,15 +1,15 @@
 use crate::Metrics;
-use std::sync::Arc;
-use std::task::{Context, Poll};
-use std::time::Instant;
+use std::{
+    sync::Arc,
+    task::{Context, Poll},
+};
+use tokio::time::Instant;
 
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct TrackService<S> {
     inner: S,
     metrics: Arc<Metrics>,
     blocked_since: Option<Instant>,
-    // Helps determine when all instances are dropped.
-    _tracker: Arc<()>,
 }
 
 impl<S> TrackService<S> {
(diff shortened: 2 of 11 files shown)


linkerd/app/inbound/src/direct.rs
```diff
@@ -70,6 +70,8 @@ where
     let http = svc::stack(http_loopback)
         .push_on_response(
             svc::layers()
+                .push(svc::layer::mk(svc::SpawnReady::new))
+                .push(metrics.stack.layer(crate::stack_labels("http", "gateway")))
```
hawkw, line 74: it occurs to me that since the direct stack uses the same TCP forward stack as the normal inbound router, the TCP stack metrics won't have any special labels indicating that they're the TCP gateway rather than in-cluster TCP inbound. It might be a significantly bigger change to allow passing in the tcp forward stack but modifying its labels, though, so maybe that's best saved for a follow-up...

linkerd/app/inbound/src/direct.rs
```diff
@@ -70,6 +70,8 @@ where
     let http = svc::stack(http_loopback)
         .push_on_response(
             svc::layers()
+                .push(svc::layer::mk(svc::SpawnReady::new))
```
hawkw, line 73: could adding a `SpawnReady` here change behavior meaningfully?

linkerd/stack/metrics/src/service.rs
```diff
@@ -35,9 +34,15 @@ where
...
                 self.metrics.not_ready_total.incr();
-                if self.blocked_since.is_none() {
-                    self.blocked_since = Some(Instant::now());
+                // If the service was already pending, then add the time we
+                // waited and reset blocked_since. This allows the value to be
+                // updated even when we're "stuck" in pending.
+                let now = Instant::now();
+                if let Some(t0) = self.blocked_since.take() {
+                    let not_ready = now - t0;
+                    self.metrics.poll_millis.add(not_ready.as_millis() as u64);
```
hawkw, line 43: nit: should we check this cast? i guess if a stack has been blocked for > 18446744073709551615 milliseconds, that's probably pretty bad...

hawkw (approved):
this seems very helpful!

linkerd/app/inbound/src/direct.rs
olix0r: There is no TCP gateway yet. It's all TCP inbound. Will change soon.

linkerd/app/inbound/src/direct.rs
olix0r: Not really. It's just slightly more robust, as we have a buffer/failfast here already

PR #893, opened 2021-02-01
inbound: Cache HTTP gateways per destination

olix0r:
When handling the HTTP gateway connections, the inbound proxy includes
the client's ID in the gateway's cache key so that a gateway service is
created for each (client, target) pair. Furthermore, the HTTP server
cache includes client _addresses_ in its key, so that a service is
cached for each _connection_ (which makes no sense).

In preparation for TCP gateway support, this change simplifies
inbound gateway handling by creating distinct target types used only by
the direct stack. Connections that lack identity or a target are failed
as early as possible, simplifying the gateway implementation. Client
identity is provided to gateway by setting the identity as a request
extension before the service is cached.

### DIFF
--- a/linkerd/app/gateway/src/config.rs
+++ b/linkerd/app/gateway/src/config.rs
@@ -49,7 +49,7 @@ impl Config {
                 Allow(self.allow_discovery),
             ))
             .instrument(
-                |g: &inbound::HttpGatewayTarget| debug_span!("gateway", target=%g.target, v=%g.version),
+                |g: &inbound::HttpGatewayTarget| debug_span!("gateway", target = %g.target, v = %g.version),
             )
             .into_inner()
     }
--- a/linkerd/app/gateway/src/config.rs
+++ b/linkerd/app/gateway/src/config.rs
@@ -1,8 +1,8 @@
 use super::make::MakeGateway;
 use linkerd_app_core::{
-    discovery_rejected, profiles, proxy::http, svc, tls, Conditional, Error, NameAddr, NameMatch,
+    discovery_rejected, profiles, proxy::http, svc, tls, Error, NameAddr, NameMatch,
 };
-use linkerd_app_inbound::target as inbound;
+use linkerd_app_inbound as inbound;
 use linkerd_app_outbound as outbound;
 use tracing::debug_span;
 
@@ -21,7 +21,7 @@ impl Config {
         profiles: P,
         local_id: Option<tls::LocalId>,
     ) -> impl svc::NewService<
-        inbound::Target,
+        inbound::HttpGatewayTarget,
         Service = impl tower::Service<
             http::Request<http::BoxBody>,
             Response = http::Response<http::BoxBody>,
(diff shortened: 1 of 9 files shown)


linkerd/app/inbound/src/direct.rs
```diff
@@ -56,90 +86,136 @@ where
...
+                Conditional::Some(tls::ServerTls::Established {
+                    client_id: Some(client_id),
+                    ..
+                }) => Ok(ClientInfo {
+                    client_id,
+                    client_addr: addrs.peer(),
+                    local_addr: addrs.target_addr(),
+                }),
+                _ => Err(RefusedNoIdentity),
+            },
```
hawkw, line 217: nit/TIOLI: does it make sense for this to be a `TryFrom` impl or something?

linkerd/app/inbound/src/direct.rs
```diff
@@ -148,29 +224,61 @@ where
...
+        )
+    }
+}
+
+// === impl HttpClientInfo ===
+
+impl Param<http::normalize_uri::DefaultAuthority> for HttpClientInfo {
+    fn param(&self) -> http::normalize_uri::DefaultAuthority {
+        http::normalize_uri::DefaultAuthority(self.target.clone().map(|t| {
+            http::uri::Authority::from_str(&t.to_string()).expect("Address must be a valid URI")
```
hawkw, line 247: should this use `NameAddr::as_http_authority`? https://github.com/linkerd/linkerd2-proxy/blob/73fb0ff5e109b37c80498192a92adf7541edd1f3/linkerd/addr/src/lib.rs#L210-L217

linkerd/app/inbound/src/direct.rs
```diff
@@ -56,90 +86,136 @@ where
...
+                        SessionProtocol::Http1 => http::Version::Http1,
+                        SessionProtocol::Http2 => http::Version::H2,
+                    },
+                })),
+                TransportHeader {
+                    name: None,
+                    protocol: Some(_),
+                    ..
+                } => Err(RefusedNoTarget),
+            },
```
hawkw, line 161: yeesh

linkerd/app/gateway/src/config.rs
```diff
@@ -48,30 +48,22 @@ impl Config {
                 profiles,
                 Allow(self.allow_discovery),
             ))
-            .check_new_service::<inbound::Target, http::Request<http::BoxBody>>()
-            .instrument(|_: &inbound::Target| debug_span!("gateway"))
+            .instrument(
+                |g: &inbound::HttpGatewayTarget| debug_span!("gateway", target=%g.target, v=%g.version),
```
hawkw, line 52: nit: in general we like to put whitespace around the `=` i think:
```suggestion
                |g: &inbound::HttpGatewayTarget| debug_span!("gateway", target = %g.target, v = %g.version),
```

linkerd/app/inbound/src/direct.rs
```diff
@@ -148,29 +224,61 @@ where
...
+
+    fn recognize(&self, req: &http::Request<B>) -> Result<Self::Key, Error> {
+        let version = req.version().try_into()?;
+
+        if let Some(target) = self.0.target.clone() {
+            return Ok(HttpGatewayTarget { target, version });
+        }
+
+        if let Some(a) = req.uri().authority() {
+            let target = NameAddr::from_str(&a.to_string())?;
```
hawkw, line 277: `NameAddr` has constructors taking an `http::Authority`, can we use those here?
https://github.com/linkerd/linkerd2-proxy/blob/73fb0ff5e109b37c80498192a92adf7541edd1f3/linkerd/addr/src/lib.rs#L181-L196

hawkw (commented):
> Furthermore, the HTTP server
> cache includes client _addresses_ in its key, so that a service is
> cached for each _connection_ (which makes no sense).

I definitely remember there being a reason for this at the time, though I haven't been able to dig it up so far...

hawkw:
per offline conversation, we determined that there isn't any reason client addrs need to be in cache keys, just client _identities_.

hawkw (approved):

PR #894, opened 2021-02-02
Bias HTTP detection towards availability

olix0r:
1b97e57 changed HTTP detection to bias towards correctness. Instead of
performing a single read, the HTTP detection module reads data until a
newline is encountered so that a dispositive determination can be made.
Practically, this causes many non-HTTP protocols to fail with a
connection timeout because a newline is never transmitted.

This change reverts the HTTP detection logic to perform a single read
and to use that data to make a determination. A minimum of 14B is
required to handle a connection as HTTP.

### DIFF
--- a/linkerd/proxy/http/src/detect.rs
+++ b/linkerd/proxy/http/src/detect.rs
@@ -5,8 +5,22 @@ use linkerd_error::Error;
 use linkerd_io::{self as io, AsyncReadExt};
 use tracing::{debug, trace};
 
-const H2_PREFACE: &[u8] = b"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n";
+// Coincidentally, both our abbreviated H2 preface and our smallest possible
+// HTTP/1 message are 14 bytes.
+const H2_PREFACE: &[u8] = b"PRI * HTTP/2.0";
+const SMALLEST_POSSIBLE_HTTP1_REQ: &str = "GET / HTTP/1.1";
 
+/// Attempts to detect the HTTP version of a stream.
+///
+/// This module biases towards availability instead of correctness. I.e. instead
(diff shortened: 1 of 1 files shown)


linkerd/proxy/http/src/detect.rs
```diff
@@ -19,162 +33,114 @@ impl Detect for DetectHttp {
...
 
-    #[tokio::test(flavor = "current_thread")]
+    #[tokio::test]
     async fn http1() {
         let _ = tracing_subscriber::fmt().with_test_writer().try_init();
 
-        for i in 1..HTTP11_LINE.len() {
-            debug!(read0 = ?std::str::from_utf8(&HTTP11_LINE[..i]).unwrap());
-            debug!(read1 = ?std::str::from_utf8(&HTTP11_LINE[i..]).unwrap());
+        // If we don't read enough to know
```
hawkw, line 101: is this comment incomplete?

hawkw (approved):
> This change reverts the HTTP detection logic to perform a single read
> and to use that data to make a determination. A minimum of 14B is
> required to handle a connection as HTTP.

As we discussed offline, it seems like a potential compromise (which would require a somewhat bigger code change) would be to make multiple reads, but fall back to transparent forwarding when detection times out, rather than failing the connection. This would let us avoid false negatives in the case where a single read doesn't return enough data to make a protocol determination, but would still bias for availability.  I'm not sure whether or not this is _necessary_ though. It seems pretty unlikely that we'll need multiple reads to detect HTTP. 

We may want to do this in a follow-up regardless, but this change seems good to me for now.

linkerd/proxy/http/src/detect.rs
olix0r, line 101:
```suggestion
```

PR #895, opened 2021-02-02
inbound: Use ALPN to determine transport header

olix0r:
The inbound proxy does not need to transparently detect transport
headers. The header is always indicated by TLS protocol negotiation.

This change decouples the `transport-header` module from the `Detect`
trait. The inbound direct stack is modified to require a transport
header when ALPN has determined that a header should be expected. Legacy
HTTP gateway support is attempted when no header is negotiated.

This sets up changes to the `Detect` server implementation.


linkerd/app/inbound/src/direct.rs
```diff
@@ -170,12 +171,22 @@ where
...
+        // Use ALPN to determine whether a transport header should be read.
+        //
         // When the transport header is not present, perform HTTP detection to
         // support legacy gateway clients.
         //
-        // TODO: Stop supporting headerless connections after we have at least
-        // one stable release out with transport header support.
-        .push(svc::UnwrapOr::layer(
+        // TODO: Remove this after we have at least one stable release out
+        // with transport header support.
```
hawkw, line 180: shouldn't the comment on removing the legacy suppport for HTTP detection be on the `push_switch` rather than on the `NewTransportHeaderServer`?

linkerd/app/inbound/src/direct.rs
```diff
@@ -238,13 +242,22 @@ impl TryFrom<(tls::ConditionalServerTls, listen::Addrs)> for ClientInfo {
     }
 }
 
+impl ClientInfo {
+    fn header_negotiated(&self) -> bool {
```
hawkw, line 246: nit/TIOLI: i might prefer 
```suggestion
    fn is_header_negotiated(&self) -> bool {
```

linkerd/transport-header/src/server.rs
```diff
@@ -0,0 +1,96 @@
...
+                    }
+                },
+                _ = time::sleep(timeout).fuse() => {
+                    let e = io::Error::new(
+                        io::ErrorKind::TimedOut,
+                        "Reading a transport header timed out"
+                    );
+                    return Err(e.into());
+                }
+            };
```
hawkw, line 87: nit, take it or leave it: isn't this `select_biased` basically just [`tokio::time::timeout`](https://docs.rs/tokio/1.1.1/tokio/time/fn.timeout.html) with a maperr on it to make `io::Error`s? IMO something like this might read a little nicer:
```suggestion
            let hdr = TransportHeader::read_prefaced(&mut io, &mut buf);
            let hdr = tokio::time::timeout(timeout, header)
                .await
                .map_err(|_|  io::Error::new(
                      io::ErrorKind::TimedOut,
                      "Reading a transport header timed out"
                ))?
                .ok_or_else(|| io::Error::new(
                    io::ErrorKind::InvalidData,
                    "Connection did not include a transport header"
                ))?;
```

hawkw (approved):
lgtm overall! commented on a couple of nits

PR #896, opened 2021-02-02
detect: Return unknown protocol on detection timeout

olix0r:
When performing HTTP detection, we always instrument a detection
timeout; and, if the timeout expires, we fail the connection. This leads
to confusion and complexity, especially when getting started with
Linkerd.

This change modifies the detect module to handle these timeouts
gracefully: instead of failing the connection with an error, we indicate
that no protocol could be detected and continue processing the
connection as a non-HTTP protocol. Furthermore, this change unifies the
timeout logic into the `DetectService` implementation to avoid
unnecessary boxing (due to async-trait).


linkerd/detect/src/lib.rs
```diff
@@ -115,27 +104,38 @@ where
...
+            let protocol = match time::timeout(timeout, detect.detect(&mut io, &mut buf)).await {
+                Ok(Ok(protocol)) => {
+                    debug!(
+                        ?protocol,
+                        elapsed = ?(time::Instant::now() - t0),
+                        "Detected"
+                    );
+                    protocol
+                }
+                Ok(Err(e)) => return Err(e),
```
hawkw, line 121: nit/TIOLI: could avoid the separate `Ok(Ok(_)` and `Ok(Err(_))` cases, like this:
```suggestion
                Ok(x) => {
                    let protocol = x?;
                    debug!(
                        ?protocol,
                        elapsed = ?(time::Instant::now() - t0),
                        "Detected"
                    );
                    protocol
                }
```
but it doesn't really matter

hawkw (approved):
looks good to me!

PR #898, opened 2021-02-02
Extract protocol detection into the gateway crate

olix0r:
The inbound proxy's direct stack has been responsible for performing
HTTP detection to dispatch requests through an HTTP gateway. In
anticipating of supporting a TCP gateway, the gateway module needs to
change to accept TCP streams to be routed.

This change modifies the gateway stack to accept TCP streams and perform
protocol detection as needed to satisfy HTTP gateway requests.

No functional changes are included. TCP gateway support will follow in a
separate change.


linkerd/app/gateway/src/lib.rs
```diff
@@ -20,49 +33,144 @@ pub struct Config {
...
+            .instrument(|h: &HttpTarget| debug_span!("gateway", target = %h.target, v = %h.version))
+            .push_on_response(
+                svc::layers()
+                    .push(svc::layer::mk(svc::SpawnReady::new))
+                    .push(
+                        metrics
+                            .stack
+                            .layer(metrics::StackLabels::inbound("http", "gateway")),
+                    )
+                    .push(svc::FailFast::layer("Gateway", inbound.dispatch_timeout))
```
hawkw, line 108: nit/tioli: does it make sense to have a stack helper for "`SpawnReady`. stack metrics, and `FailFast`"? seems like they almost always occur together

hawkw (approved):

linkerd/app/gateway/src/lib.rs
olix0r: we need to create a spawn ready layer that does the right thing or wait for tower 0.5

PR #899, opened 2021-02-03
outbound: Split outbound::server into multiple modules

olix0r:
In order to support TCP gateway, the outbound stack will need to be
constructed without its HTTP server, profile discovery, etc.

This change splits the `outbound::server` module into several distinct
modules:

* The `outbound::server::cache` helper is now `outbound::discover`. It
  is now responsible for caching profile discovery;
* The HTTP server logic is moved into `outbound::http::server`;
* `outbound::server` now includes only protocol detection and TCP
  forwarding; and
* `outbound::stack` composes these stacks for non-gateway outbound
  proxying.

--
No functional changes.


hawkw (approved):
this is basically just a series of straight-up moves --- looks good to me!

PR #900, opened 2021-02-04
Simplify opencensus stack composition

olix0r:
The tracing infrastructure exposes its implementation details (via type
signatures) to stack layers.

This change renames `app::core::spans` to `app::core::http_tracing`
which provides helpers that bundles the `TraceContext` layer with a
`SpanConverter`. It also provides a `SpanSink` type alias so that proxy
stack builders can be ignorant of its actual type signature.

The `SpanSink` trait is updated to expose `is_enabled` which must be
called before constructing a span so that the `TraceContext` type isn't
required to hold an `Option<SpanSink>`.


linkerd/app/core/src/http_tracing.rs
```diff
@@ -1,23 +1,22 @@
...
-use std::collections::HashMap;
-use std::{error, fmt};
+use linkerd_stack::layer;
+use linkerd_trace_context::{self as trace_context, TraceContext};
+use std::{collections::HashMap, error, fmt, sync::Arc};
 use tokio::sync::mpsc;
 
-const SPAN_KIND_SERVER: i32 = 1;
-const SPAN_KIND_CLIENT: i32 = 2;
+pub type SpanSink = Option<mpsc::Sender<oc::Span>>;
```
hawkw, line 8: nit: imo having a type alias with the _same_ name as the trait is a little bit unclear, but maybe that's just me.

hawkw (approved):

PR #901, opened 2021-02-04
Simplify outbound stack composition with builders

olix0r:
Outbound stack construction is instrumented from a mess of
free-functions that take a large number of arguments. Many stack
functions accept the same arguments. This leads to a lot of boilerplate
and makes it inconvenient to reconstruct stacks, as many values are
needed.

This change simplifies outbound stack construction by introducing two
new types:

- `core::ProxyRuntime`, which holds all common runtime configuration
  (metrics, tap, tracing, drain, etc); and
- `outbound::Outbound`, a helper that holds the outbound config, the
   proxy runtime, along with outbound-specific stack constructors.

All outbound stack constructors are now implemented in terms of these
types. This makes it simple to construct stacks without passing around
the same set of values ad-nauseum.

Following this, we should do the same with the inbound proxy.


linkerd/app/outbound/src/discover.rs
```diff
@@ -1,64 +1,73 @@
...
-        )
-        .check_new_service::<tcp::Accept, SensorIo<I>>()
-        .push(metrics.transport.layer_accept())
-        .push_cache(config.proxy.cache_max_idle_age)
-        .check_new_service::<tcp::Accept, I>()
-        .push_map_target(tcp::Accept::from)
-        .check_new_service::<listen::Addrs, I>()
-        .into_inner()
+impl<N> Outbound<N> {
+    /// Discovers the profile for a TCP endpoint.
```
hawkw, line 10: nit/TIOLI: should this say something like "pushes a stack that discovers..."?

linkerd/app/outbound/src/http/endpoint.rs
```diff
@@ -1,75 +1,95 @@
...
-    B::Data: Send + 'static,
-    C: svc::Service<Endpoint, Error = Error> + Clone + Send + Sync + Unpin + 'static,
+    C: svc::Service<Endpoint> + Clone + Send + Sync + Unpin + 'static,
     C::Response: io::AsyncRead + io::AsyncWrite + Send + Unpin,
+    C::Error: Into<Error>,
     C::Future: Send + Unpin,
 {
-    let identity_disabled = local_id.is_none();
-    svc::stack(tcp_connect)
+    pub fn push_http_endpoint<B>(
```
hawkw, line 19: nit/TIOLI: now that the various stacks are constructed by methods whose names are things like `push_http_endpoint`, instead of free fns that are basically all named `stack`, i'm not _super_ convinced each part of the HTTP stack needs to be in a separate module --- the names are pretty clear without modules to namespace them. We could probably put them all in the `http` module.

on the other hand, maybe you'd prefer to have more, smaller files for organizing the code? definitely not a blocker regardless.

linkerd/app/outbound/src/http/tests.rs
```diff
@@ -29,74 +29,53 @@ use tracing::Instrument;
...
-        resolver.clone(),
-        metrics.outbound.clone(),
-    );
-    let http = crate::http::server::stack(&cfg.proxy, &metrics.outbound, None, router);
-    let accept = crate::server::stack(&cfg.proxy, &metrics.outbound, drain, NoTcpBalancer, http);
-    (accept, drain_tx)
+    let http = Outbound::new(cfg.clone(), rt.clone(), connect)
+        .push_http_endpoint()
+        .push_http_logical(resolver)
+        .push_http_server()
```
hawkw, line 76: <3 <3 <3 gorgeous

linkerd/app/src/lib.rs
```diff
@@ -155,67 +155,46 @@ impl Config {
...
+                    .push_http_endpoint()
+                    .push_http_logical(dst.resolve.clone())
+                    .into_inner();
+                let server = outbound::ingress::new_service(
+                    outbound.clone(),
+                    outbound_rt.clone(),
+                    dst.profiles.clone(),
+                    tcp,
+                    http,
+                );
```
hawkw, line 181: this is really lovely! <3

linkerd/app/outbound/src/tcp/connect.rs
```diff
@@ -1,77 +1,107 @@
...
-        })
-        .push_request_filter(PreventLoop { port: server_port })
-        .into_inner()
+pub fn stack(config: crate::Config, runtime: ProxyRuntime) -> Outbound<ConnectTcp> {
+    let stack = svc::stack(ConnectTcp::new(config.proxy.connect.keepalive));
+    Outbound {
+        config,
+        runtime,
+        stack,
+    }
```
hawkw, line 14: nit/TIOLI: seems like this could alternatively be a function on the `Outbound` type, like `Outbound::new_tcp_connect`? not sure if that's better or worse.

hawkw (approved):
:raised_hands: this is much neater, i've wanted to do something like this for a while! we should probably make a similar change on the inbound side...

PR #902, opened 2021-02-04
Simplify inbound stack composition

olix0r:
18e568a changed how outbound stacks are built--replacing free functions
with an `Outbound` builder type.

This change applies this pattern to the inbound proxy, reducing
boilerplate.

### DIFF
--- a/linkerd/app/inbound/src/lib.rs
+++ b/linkerd/app/inbound/src/lib.rs
@@ -22,14 +22,14 @@ use self::{
 };
 use linkerd_app_core::{
     config::{ConnectConfig, ProxyConfig},
-    detect, drain, http_tracing, io, metrics, profiles,
-    proxy::{identity::LocalCrtKey, tap, tcp},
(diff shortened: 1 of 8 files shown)


linkerd/app/inbound/src/lib.rs
```diff
@@ -44,50 +44,145 @@ pub struct Config {
...
+
+    pub fn into_inner(self) -> S {
+        self.stack.into_inner()
+    }
+}
+
+impl Inbound<()> {
+    pub fn new_without_stack(config: Config, runtime: ProxyRuntime) -> Self {
+        Self::new(config, runtime, ())
+    }
```
hawkw, line 81: nit/TIOLI: should this just be `new` (or `Default::default`)?

hawkw (approved):
assuming the actual contents of the stacks didn't change (and i _think_ they didn't, but github's diff viewer makes it hard to tell), this lgtm!

hawkw:
looks like rustfmt needs to be run?

PR #903, opened 2021-02-05
Prevent outbound connections on the loopback interface

olix0r:
The outbound proxy attempts to prevent loops by preventing connections
to the proxy's outbound server port; but the outbound proxy **only**
listens on the loopback interface, so port-based matching is
ineffective.

We *should* prevent all outbound connections on the loopback
interface--it indicates an iptables misconfiguration; but this would
unfortunately break many of our integration tests

This change removes outbound loop prevention.


hawkw (approved):
lgtm...it would be nice to figure out a way to have proper loop detection on outbound without breaking the tests though.

PR #906, opened 2021-02-06
Improve stack builder ergonomics

olix0r:
Some stack builders were not converted to be functions on the `Inbound`
or `Outbound` types because they had no obvious underlying stack type.

This change modifies `Inbound` and `Outbound` to be built with a
unit-typed stack by default. They each now have a set of common utility
functions used to bind the unit-typed stack to a service-stack.

No functional changes.


linkerd/app/outbound/src/http/detect.rs
```diff
@@ -0,0 +1,127 @@
+use crate::{http, stack_labels, target::ShouldResolve, tcp, Outbound};
+use linkerd_app_core::{
+    config::{ProxyConfig, ServerConfig},
+    detect, io, profiles, svc, Addr, Error,
+};
+use tracing::debug_span;
+
+impl<T> Outbound<T> {
+    pub fn push_detect_http<TSvc, H, HSvc, I>(
```
olix0r, line 9: renamed from `outbound::server`

hawkw (approved):

PR #908, opened 2021-02-08
Parameterize resolution targets

olix0r:
The proxy's destination API clients need to turn target types into a
resolveable address. The implementations currently use only `ToString`
and we provide wrappers that handle dispatching the proper value -- but
this is fragile.

This change parameterizes both the api-resolve and service-profile
clients with newtypes: `Concreteddr` and` LogicalAddr`, respectively.
The test infrastructure has been simplified to no longer be generic over
the target type (instead relying on `Addr` which ultimately backs both
of these types).

This change sets up future stack changes. No functional changes.


linkerd/app/gateway/src/lib.rs
```diff
@@ -122,7 +126,7 @@ where
             let allow = allow_discovery.clone();
             move |addr: NameAddr| {
                 if allow.matches(addr.name()) {
-                    Ok(addr)
+                    Ok(profiles::LogicalAddr(addr.into()))
```
hawkw, line 129: TIOLI: seems like there could be a `From` impl for `LogicalAddr`? nbd.

hawkw (approved):
looks good to me — sorry I didn't get to this one sooner! definitely feels a bit clearer.

linkerd/app/gateway/src/lib.rs
olix0r: I want this to go away in favor of a Param, anyway.

PR #909, opened 2021-02-08
gateway: Transport TCP connections     

olix0r:
The proxy's gateway mode can only transport HTTP requests; but, with the
introduction of the transport header (informed by TLS ALPN), the gateway
now has enough context to route TCP connections through the gateway.
    
This PR modifies the gateway stack to handle transported connections
differently from legacy HTTP connections. When clients connect to a
gateway with a transport header, the header's metadata is used for all
routing/discovery so, in the HTTP case, a request router is not
necessary. A TCP gateway is also introduced when a transport header does
not include a protocol hint. The same discovery is performed and
connections are dispatched through the TCP logical stack to balance and
forward connections within the gateway cluster.
    
More testing is needed on this, but I plan to address this in followup
changes that make the outbound stack more flexible and amenable to
testing.


linkerd/app/gateway/src/lib.rs
```diff
@@ -85,161 +93,246 @@ where
...
+        .push(profiles::discover::layer(profiles.clone(), {
+            let allow = allow_discovery.clone();
+            move |addr: NameAddr| {
+                if allow.matches(addr.name()) {
+                    Ok(addr)
+                } else {
+                    Err(RefusedNotResolved(addr))
+                }
+            }
+        }))
```
hawkw, line 129: nit/TIOLI: seems like there _could_ maybe be some kind of
```rust
impl NameMatch {
    // ...
    pub fn into_predicate<E: Error + Default>(&self) -> NamePredicate<E> {
        NamePredicate {
            m: self.clone(), _e: PhantomData,
        }
    }
}

struct NamePredicate<E: Default> {
    m: NameMatch,
    _e: PhantomData<fn(E)>,
}

impl<E, P> Predicate<E, P> for NamePredicate<E>
where
    E: Error + Default,
    P: Param<NameAddr>,
{
    type Request = P;
    fn check(&mut self, target: P) -> Result<Self::Request, Box<dyn Error + Send + Sync + 'static>> {
        if self.m.matches(target.param()) {
            Ok(target)
        } else {
            Err(E::default().into())
        }
    }
}
```
that could be used both here and in the http stack below?

Not sure if this is worth the effort.

linkerd/app/gateway/src/lib.rs
```diff
@@ -85,161 +93,246 @@ where
...
-                ))
-                .push(detect::NewDetectService::layer(
-                    detect_protocol_timeout,
-                    http::DetectHttp::default(),
-                ))
-                .into_inner(),
+            tcp.into_inner(),
+        )
+        .push_switch(
+            |gw: GatewayConnection| match gw {
```
hawkw, line 237: huh, surprised the type can't be inferred from the match? not a big deal if it can't

linkerd/app/gateway/src/lib.rs
```diff
@@ -85,161 +93,246 @@ where
...
-    // detection (for legacy clients).
+    // When handling gateway connections from older clients that do not
+    // support the transport header, do protocol detection and route requests
+    // based on each request's URI.
     //
-    // TODO: Handle TCP gateway traffic when a target is set without a protocol.
-    svc::stack(http_server.clone())
-        .push_on_response(svc::MapTargetLayer::new(io::EitherIo::Left))
+    // Non-HTTP connections are refused.
+    let legacy_http = inbound
```
hawkw, line 190: should the legacy http stack have its own stack metrics or something? it could be useful to have a way to know how much legacy gateway traffic exists in the mesh.

linkerd/app/inbound/src/direct.rs
```diff
@@ -25,8 +25,14 @@ pub struct RefusedNoIdentity(());
...
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
-pub struct GatewayConnection {
-    pub target: Option<NameAddr>,
+pub enum GatewayConnection {
+    Transported(Transported),
+    Legacy(ClientInfo),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Transported {
```
hawkw, line 34: not _totally_ sold on this naming...all traffic is "transported". maybe `WithHeader`? not gonna block on bikeshedding this though.

linkerd/app/gateway/src/lib.rs
```diff
@@ -85,161 +93,246 @@ where
...
+            }
+        }))
+        .push_on_response(
+            svc::layers()
+                .push(
+                    inbound
+                        .runtime()
+                        .metrics
+                        .stack
+                        .layer(metrics::StackLabels::inbound("tcp", "gateway")),
```
hawkw, line 137: nit/TIOLI: i wonder if we're writing stuff like this in enough places that it would be good for the `Inbound` (and `Outbound`) types to have a method, so we could just say
```rust
    .push(inbound.stack_metrics("tcp", "gateway"))
```

not a big deal though.

hawkw (approved):
I gave this a pretty thorough reading and didn't notice anything worrying --- looks good to me! I had some minor comments.

PR #911, opened 2021-02-10
update Tower to v0.4.5

hawkw:
The recently released v0.4.5 of Tower includes a change to
`tower::buffer` to use `tokio-util`'s `PollSemaphore`, which should
reduce allocations when polling `Buffer` services. Additionally, it
includes a fix for `tracing` spans not being propagated to the tasks
spawned by the `SpawnReady` middleware.

No functional changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2268,9 +2268,9 @@ dependencies = [
 
 [[package]]
 name = "tower"
-version = "0.4.4"
+version = "0.4.5"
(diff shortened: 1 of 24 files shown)


olix0r:
>  a change to tower::buffer to use tokio-util's PollSemaphore, which should reduce allocations when polling Buffer services. 

We don't actually use this, though, right? We'd have to update our channel to use PollSemaphore?

olix0r (approved):

PR #913, opened 2021-02-11
Log warnings when controller components do not resolve

olix0r:
The `dns` module currently swallows resolution errors so that these
errors never hit the error handler.

This change modifies the `dns` module to propagate these errors so they
are handled by the resolver recovery module. The control plane's
recovery implementation is updated to log these errors and to use the
DNS TTL for backoffs, when present.

We now see errors like:

    WARN ThreadId(01) identity: linkerd_app_core::control: Failed to resolve control-plane component error=no record found for name: linkerd-identity-headless.linkerd.svc.cluster.local. type: SRV class: IN

### DIFF
--- a/linkerd/app/core/src/control.rs
+++ b/linkerd/app/core/src/control.rs
@@ -7,7 +7,10 @@ use crate::{
     transport::ConnectTcp,
     Addr, Error,
 };
+use futures::{future::Either, FutureExt};
 use std::fmt;
+use tokio::time;
+use tracing::warn;
 
 #[derive(Clone, Debug)]
 pub struct Config {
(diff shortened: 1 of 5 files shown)


linkerd/app/core/src/control.rs
```diff
@@ -54,16 +58,46 @@ impl Config {
...
+                        ..
+                    } = e.kind()
+                    {
+                        let ttl = time::Duration::from_secs(*ttl_secs as u64);
+                        let stream = stream! {
+                            loop {
+                                time::sleep(ttl).await;
+                                yield;
+                            }
+                        };
```
hawkw, line 84: this is equivalent to taking a [`tokio::time::Interval`](https://docs.rs/tokio/1.2.0/tokio/time/fn.interval.html) and wrapping it as a `Stream`, either via [`tokio-stream`](https://docs.rs/tokio-stream/0.1.3/tokio_stream/wrappers/struct.IntervalStream.html) or our `IntoStream` utility in `linkerd-channel`. might be a little nicer to just use `Interval` (`Interval` is also a little more efficient but this use-case isn't really performance-critical).

linkerd/app/core/src/control.rs
```diff
@@ -54,16 +58,46 @@ impl Config {
...
+        // is one, to drive re-resolution attempts.
+        let resolve_backoff = {
+            let backoff = self.connect.backoff;
+            move |error: Error| {
+                warn!(%error, "Failed to resolve control-plane component");
+                if let Some(e) = error.downcast_ref::<dns::ResolveError>() {
+                    if let dns::ResolveErrorKind::NoRecordsFound {
+                        negative_ttl: Some(ttl_secs),
+                        ..
+                    } = e.kind()
```
hawkw, line 76: should we also log the TTL in this case?

hawkw (approved):

linkerd/app/core/src/control.rs
olix0r: i don't think it's particularly relevant for the warning scope; and it'll generally be pretty obvious based on the interval from the debug perspective

PR #914, opened 2021-02-13
transport: Replace async-stream with tokio-stream

olix0r:
The TCP listener creates an async-stream of accepted connections; but
the tokio-stream crate provides a stream utilitity for TCP listeners.

This change swaps the implementation of BindTcp. The tokio-stream
implementation should be lighter-weight to compile; and this sets up for
refactoring the listener so that TCP binding is not required for tests.


kleimkuhler (approved):
Looks good!

PR #916, opened 2021-02-14
proxy/transport: use into_std in set_keepalive_or_warn

19h:
Follows up on an older comment in the set_keepalive_or_warn function of proxy/transport waiting for https://github.com/tokio-rs/tokio/pull/3189 to be merged, allowing linkerd to use a non-unsafe interface for obtaining a socket2:Socker via a std TcpStream.


19h:
Ignore this. It seems there is no way to do this using the officially provided facilities by reference safely.

PR #917, opened 2021-02-15
Re-export stack::Param as svc::Param

olix0r:
This is a purely superificial change that shortens a very common import.


kleimkuhler (approved):
🤙

PR #920, opened 2021-02-16
update tracing-futures, rm old pin-project

hawkw:
This PR updates the dependency on `tracing-futures` to 0.2.5, which
updates that crate to depend on `pin-project` 1.0. After removing one
last `pin-project` 0.4 dependency, we can now remove `pin-project` from
the duplicate dependency versions allowlist!


olix0r (approved):
:tada: thanks @hawkw

PR #921, opened 2021-02-17
transport: Introduce address new-types

olix0r:
Throughout the codebase, we handle many types of `SocketAddr`s. It's
easy to use these instances in an incorrect context.

In preparation for upcoming changes to listen module and target types,
this change introduces a family of new-types to disambiguate these uses:

- `ClientAddr`  -- a client's (ephemeral) address
- `ServerAddr`  -- a server's address
- `OrigDstAddr` -- an address set by the `SO_ORIGINAL_DST` socket option

Additionally, `Local` and `Remote` address types are introduced to
further disambiguate the above types:

- `Remote<ClientAddr>` -- a server connection's peer address
- `Local<ServerAddr>`  -- a server connection's local address
- `Remote<ServerAddr>` -- a client connection's peer address
- `Local<ClientAddr>`  -- a client connection's local address

A special `ListenAddr` type is introduced to indicate an unbound address
to be used to bind a server. The `OrigDstAddr` trait is renamed to
`GetOrigDstAddr`.

No functional change.

### DIFF
--- a/linkerd/app/inbound/src/http/tests.rs
+++ b/linkerd/app/inbound/src/http/tests.rs
@@ -12,7 +12,7 @@ use linkerd_app_core::{
     proxy,
     svc::{self, NewService, Param},
     tls,
-    transport::{self, ConnectAddr},
+    transport::{ClientAddr, Remote, ServerAddr},
     Conditional, Error, NameAddr, ProxyRuntime,
 };
 use tracing::Instrument;
@@ -21,7 +21,7 @@ fn build_server<I>(
     cfg: Config,
     rt: ProxyRuntime,
     profiles: resolver::Profiles,
-    connect: Connect<ConnectAddr>,
+    connect: Connect<Remote<ServerAddr>>,
 ) -> impl svc::NewService<
     HttpAccept,
     Service = impl tower::Service<
@@ -36,9 +36,7 @@ where
     I: io::AsyncRead + io::AsyncWrite + io::PeerAddr + Send + Unpin + 'static,
 {
     let connect = svc::stack(connect)
-        .push_map_target(|t: TcpEndpoint| {
-            transport::ConnectAddr(([127, 0, 0, 1], t.param()).into())
-        })
+        .push_map_target(|t: TcpEndpoint| Remote(ServerAddr(([127, 0, 0, 1], t.param()).into())))
         .into_inner();
     Inbound::new(cfg, rt)
         .with_stack(connect)
@@ -58,7 +56,7 @@ async fn unmeshed_http1_hello_world() {
         version: proxy::http::Version::Http1,
         tcp: TcpAccept {
             target_addr: ([127, 0, 0, 1], 5550).into(),
-            client_addr: ([10, 0, 0, 41], 6894).into(),
+            client_addr: Remote(ClientAddr(([10, 0, 0, 41], 6894).into())),
             tls: Conditional::None(tls::server::NoServerTls::NoClientHello),
         },
     };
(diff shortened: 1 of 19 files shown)


linkerd/app/inbound/src/http/tests.rs
```diff
@@ -58,7 +56,7 @@ async fn unmeshed_http1_hello_world() {
         version: proxy::http::Version::Http1,
         tcp: TcpAccept {
             target_addr: ([127, 0, 0, 1], 5550).into(),
-            client_addr: ([10, 0, 0, 41], 6894).into(),
+            client_addr: Remote(ClientAddr(([10, 0, 0, 41], 6894).into())),
```
hawkw, line 59: nit/TIOLI: seems like the `FooAddr` newtypes could have
```rust
impl<T> From<T> for FooAddr where T: Into<SocketAddr> { ... }
```
or something? but, this might make it more explicit, so :woman_shrugging:

linkerd/app/inbound/src/target.rs
```diff
@@ -16,7 +16,7 @@ use tracing::debug;
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct TcpAccept {
     pub target_addr: SocketAddr,
```
hawkw, line 18: should this also be newtyped?

linkerd/app/outbound/src/http/tests.rs
```diff
@@ -445,19 +422,21 @@ async fn active_stacks_dont_idle_out() {
     proxy_bg.await.unwrap();
 }
 
+pub fn addrs(od: SocketAddr) -> listen::Addrs {
+    listen::Addrs::new(
+        Local(ServerAddr(([127, 0, 0, 1], 4140).into())),
+        Remote(ClientAddr(([127, 0, 0, 1], 666).into())),
+        Some(OrigDstAddr(od)),
+    )
+}
```
hawkw, line 431: good call!

linkerd/app/outbound/src/tcp/tests.rs
```diff
@@ -721,9 +725,9 @@ async fn profile_endpoint_propagates_conn_errors() {
     let mut server = build_server(cfg, profiles, resolver, connect);
 
     let svc = server.new_service(listen::Addrs::new(
-        ([127, 0, 0, 1], 4140).into(),
-        ([127, 0, 0, 1], 666).into(),
-        Some(ep1),
+        Local(ServerAddr(([127, 0, 0, 1], 4140).into())),
+        Remote(ClientAddr(([127, 0, 0, 1], 666).into())),
+        Some(OrigDstAddr(ep1)),
     ));
```
hawkw, line 731: nit/TIOLI: seems like the `addrs` helper function in the HTTP tests could be pulled out into the top-level test support module and we could use it here as well?

linkerd/proxy/transport/src/addrs.rs
```diff
@@ -0,0 +1,157 @@
...
+    fn as_ref(&self) -> &SocketAddr {
+        &self.0
+    }
+}
+
+impl Into<SocketAddr> for ClientAddr {
+    fn into(self) -> SocketAddr {
+        self.0
+    }
+}
```
hawkw, line 43: Should these types also be `From<SocketAddr>`?

hawkw (approved):
looks good to me — i commented on some non-blocking nits.

looks like there's an import of `SocketAddr` in `transport/listen.rs` that's now unused, and breaking CI.

linkerd/app/inbound/src/target.rs
olix0r: It can't be, currently, because it can be either the local address or the origdst. We'll change this, eventually, to only be the origdst.

linkerd/app/inbound/src/http/tests.rs
olix0r: Yeah; I chose not to support From converters so that we actually have to write down what type it is. The converters let us fall back on implicit behavior. We can add these later if we get frustrated, but for now want to be explicit.

hawkw: ah, yup, that's valid, carry on!

linkerd/app/outbound/src/tcp/tests.rs
olix0r: yyyeah i like that idea but let's do that in a followup.

PR #922, opened 2021-02-17
buffer: replace `linkerd-buffer` with `tower::buffer` from upstream

hawkw:
The proxy currently has its own implementation of a `tower` `Service`
that makes an inner service `Clone`able by driving it in a spawned task
and buffering requests on a channel. This also exists upstream, as
`tower::buffer`.

We implemented our own version for a couple of reasons: to avoid an
upstream issue where memory was leaked when a buffered request was
cancelled, and to implement an idle timeout when the buffered service
has been unready for too long. However, it's no longer necessary to
reimplement our own buffer service for these reasons: the upstream bug
was fixed in `tower` 0.4 (see tower-rs/tower#476, tower-rs/tower#480,
and tower-rs/tower#556); and we no longer actually use the buffer idle
timeout (instead, we idle out unresponsive services with the separate
`Failfast` middleware, note that `push_spawn_buffer_with_idle_timeout`
is never actually used).

Therefore, we can remove our _sui generis_ implementation in favour of
`tower::buffer` from upstream. This eliminates dead code for the idle
timeout, which we never actually use, and reduces duplication (since
`tonic` uses `tower::buffer` internally, its code is already compiled
into the proxy). It also reduces the amount of code I'm personally
responsible for maintaining in two separate places ;)

Since the `linkerd-buffer` crate erases the type of the buffered
service, while `tower::buffer` does not, I've changed the
`push_spawn_buffer`/`spawn_buffer` helpers to also include a
`BoxService` layer. This required adding a `BoxServiceLayer` type, since
`BoxService::layer` returns a `LayerFn` with an unnameable type.

Also, this change ran into issues due to a compiler bug where generators
(async blocks) sometimes forget concrete lifetimes,
rust-lang/rust#64552. In order to resolve this, I had to remove the
outermost `async` blocks from the OpenCensus and identity daemon tasks.
These async blocks were used only for emitting a tracing event when the
task is started, so it wasn't a big deal to remove them; I moved the
trace events into the actual daemon task functions, and used a `tracing`
span to propagate the remote addresses which aren't known inside the
daemon task functions.


linkerd/opencensus/src/lib.rs
```diff
@@ -28,6 +28,7 @@ where
     T::ResponseBody: 'static,
     S: Stream<Item = Span> + Unpin,
 {
+    debug!("span exporter running");
```
olix0r, line 31: nit: prefer to capitalize log messages

```suggestion
    debug!("Span exporter running");
```

linkerd/proxy/identity/src/certify.rs
```diff
@@ -97,6 +97,7 @@ impl Daemon {
             config,
         } = self;
 
+        debug!("identity daemon running");
```
olix0r, line 100: nit:
```suggestion
        debug!("Identity daemon running");
```

linkerd/stack/src/box_service.rs
```diff
@@ -0,0 +1,36 @@
+use std::marker::PhantomData;
+pub use tower::util::BoxService;
+#[derive(Copy, Debug)]
```
olix0r, line 3:
```suggestion
pub use tower::util::BoxService;

#[derive(Copy, Debug)]
```

olix0r (commented):
this looks great! a few nits...

linkerd/app/core/src/control.rs
```diff
@@ -43,7 +43,10 @@ type BalanceBody =
 
 type RspBody = linkerd_http_metrics::requests::ResponseBody<BalanceBody, classify::Eos>;
 
-pub type Client<B> = linkerd_buffer::Buffer<http::Request<B>, http::Response<RspBody>>;
+pub type Client<B> = svc::buffer::Buffer<
+    svc::BoxService<http::Request<B>, http::Response<RspBody>, Error>,
```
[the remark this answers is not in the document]
hawkw: yeah, good call

olix0r (approved):

PR #923, opened 2021-02-17
transport: Introduce a Keepalive type

olix0r:
When handling `Duration` values, it can be ambiguous as to the value's
intended use.

This change introduces a `Keepalive` new-type that wraps an optional
`Duration`. This will allow keepalive values to be parameterized without
ambiguity (in future changes).

No functional changes.


hawkw (approved):
sure, seems good to me! is the intention to also newtype timeouts in the future?

olix0r:
@hawkw yeah, that's the idea. for now i'm only really tackling the connect/accept impl that depend on these.  Basically, anything that needs parameterizing should get a newtype.

PR #924, opened 2021-02-18
outbound: Prevent connections on the loopback interface

olix0r:
linkerd/linkerd2#5764 reports proxy OOMs, apparently caused by traffic
looping from the outbound proxy back to the inbound proxy. While we
don't understand the underlying circumstances, we should *never* proxy
outbound connections on the loopback interface in normal operation.

This change modifies the outbound TCP connection stack to prevent
connections on loopback addresses. A feature flag, `allow-loopback`, is
provided to disable this functionality. This is necessary so that our
local integration tests can continue to pass.

### DIFF
--- a/linkerd/app/Cargo.toml
+++ b/linkerd/app/Cargo.toml
@@ -12,6 +12,7 @@ This is used by tests and the executable.
 """
 
 [features]
+allow-loopback = ["linkerd-app-outbound/allow-loopback"]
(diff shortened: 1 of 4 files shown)


hawkw (approved):
this looks good to me; the feature flagging feels slightly unfortunate, as it would be nice to also be able to test this behavior, but i can't really think of any way for the integration tests to use real sockets without running into problems here...

PR #925, opened 2021-02-20
make proxy builds work on windows

Pothulapati:
Currently, The proxy builds fail on windows because of the absence
of `future::preludes`. That does not seem to be required anymore
and the windows ctrl_c signal can be used similar with that of
the unix signals.

This PR updates the `shutdown` fn in imp for windows to be just
like the unix ones. I'm aware that the proxy is not intended to be ran
on windows machines but this was a easy fix to make builds work.

Signed-off-by: Tarun Pothulapati <tarunpothulapati@outlook.com>

### DIFF
--- a/linkerd/signal/src/lib.rs
+++ b/linkerd/signal/src/lib.rs
@@ -39,7 +39,6 @@ mod imp {
 
 #[cfg(not(unix))]
 mod imp {
-    use futures::prelude::*;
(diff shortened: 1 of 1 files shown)


olix0r (approved):
Thanks, Tarun!

Type1J:
Is there a tracker for LinkerD Windows support?

PR #926, opened 2021-02-22
Add support for TLS v1.3

olix0r:
TLS 1.3 has been available in rustls for a while now, giving us some
confidence that it's stable.

This change adds TLS 1.3 to the list of supported TLS versions. With
this change, proxies will prefer TLS 1.3 when both peers include this
change and will downgrade to TLS 1.2 when communicating with an older
proxy version.

I've tested that this behaves as expected by upgrading just the
emojivoto app's _web_ and _voting_ services with debug logging. Older
proxies communicate with newer proxies via TLS 1.2 and newer proxies
communicate via TLS 1.3.


hawkw (approved):
:lock:1️⃣⏺️3️⃣:+1:

PR #929, opened 2021-02-23
detect: Surface timeouts to inner stack

olix0r:
In 95498395, the detect service was modified to handle detect timeouts
gracefully; but, we can't differentiate this situration from other
detection mismatches.

In order to use the detect trait more generally--especially for TLS
detection, this change modifies the `Detect` trait to pass an error into
the inner stack when detection fails. Uses are updated to either fail on
or log when an error is countered, as appropriate. This will be relied
on in followup changes.

Furthermore, the `Detect` trait's I/O type is moved to be a
trait-generic (rather than a method-generic) so that implementations may
impose further type constraints.

### DIFF
--- a/linkerd/app/src/admin.rs
+++ b/linkerd/app/src/admin.rs
@@ -13,7 +13,7 @@ use crate::{
     inbound::target::{HttpAccept, Target, TcpAccept},
     svc,
 };
-use std::{fmt, net::SocketAddr, pin::Pin, time::Duration};
+use std::{convert::TryFrom, net::SocketAddr, pin::Pin, time::Duration};
 use tokio::sync::mpsc;
 
 #[derive(Clone, Debug)]
@@ -27,9 +27,6 @@ pub struct Admin {
     pub latch: admin::Latch,
     pub serve: Pin<Box<dyn std::future::Future<Output = Result<(), Error>> + Send + 'static>>,
 }
-
-#[derive(Debug, Default)]
-pub struct AdminHttpOnly(());
 
 impl Config {
     pub fn build<R>(
@@ -60,10 +57,7 @@ impl Config {
             )
             .push_map_target(Target::from)
             .push(http::NewServeHttp::layer(Default::default(), drain.clone()))
-            .push_map_target(HttpAccept::from)
-            .push(svc::UnwrapOr::layer(
-                svc::Fail::<_, AdminHttpOnly>::default(),
-            ))
+            .push(svc::Filter::<TcpAccept, _>::layer(HttpAccept::try_from))
             .push(detect::NewDetectService::layer(
                 DETECT_TIMEOUT,
                 http::DetectHttp::default(),
@@ -82,11 +76,3 @@ impl Config {
         })
     }
 }
-
-impl fmt::Display for AdminHttpOnly {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.pad("proxy admin server is HTTP-only")
-    }
-}
-
-impl std::error::Error for AdminHttpOnly {}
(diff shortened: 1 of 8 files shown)


linkerd/detect/src/lib.rs
```diff
@@ -43,6 +46,16 @@ pub struct DetectService<T, D, N> {
 
 const BUFFER_CAPACITY: usize = 1024;
 
+pub fn allow_timeout<P, T>((p, t): (Detected<P>, T)) -> (Option<P>, T) {
+    match p {
+        Ok(p) => (p, t),
+        Err(DetectTimeout(timeout)) => {
+            info!(?timeout, "Protocol detection timeout");
```
hawkw, line 53: very minor, just a potential nice-to-have: if the intention is eventually to use this for both HTTP and TLS detection, it could be useful to have this include _which_ protocol is being detected...maybe something like:
```suggestion
            info!(
                ?timeout,
                protocol = %std::any::type_name::<P>(),
                "Protocol detection timeout"
            );
```

linkerd/detect/src/lib.rs
```diff
@@ -142,3 +152,13 @@ where
         })
     }
 }
+
+// === impl DetectTimeout
+
+impl fmt::Display for DetectTimeout {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Detection timed out after {:?}", self.0)
```
hawkw, line 160: very minor nit: in general, errors (including other errors added in this PR) tend to begin with lower case (so that you can say things like "foo error description: bar error description" without it looking weird):
```suggestion
        write!(f, "detection timed out after {:?}", self.0)
```

linkerd/app/src/admin.rs
```diff
@@ -60,10 +57,7 @@ impl Config {
             )
             .push_map_target(Target::from)
             .push(http::NewServeHttp::layer(Default::default(), drain.clone()))
-            .push_map_target(HttpAccept::from)
-            .push(svc::UnwrapOr::layer(
-                svc::Fail::<_, AdminHttpOnly>::default(),
-            ))
+            .push(svc::Filter::<TcpAccept, _>::layer(HttpAccept::try_from))
```
hawkw, line 60: nit/TIOLI, not a blocker for this PR: wonder if a `push_filter` would help reduce some boilerplate...there could even be a `push_try_from` adapter that just pushes `Filter<R::TryFrom>`...but that might not be inferrable...

hawkw (approved):
lgtm, couple very tiny nits.

olix0r:
@hawkw made some changes to the DetectTimeout type. Anything look weird to you?

linkerd/detect/src/lib.rs
```diff
@@ -5,25 +5,27 @@ use linkerd_error::Error;
...
-    async fn detect<I: io::AsyncRead + Send + Unpin + 'static>(
-        &self,
-        io: &mut I,
-        buf: &mut BytesMut,
-    ) -> Result<Option<Self::Protocol>, Error>;
+    async fn detect(&self, io: &mut I, buf: &mut BytesMut)
+        -> Result<Option<Self::Protocol>, Error>;
 }
 
+pub type Detected<P> = Result<Option<P>, DetectTimeout<P>>;
```
hawkw, line 25: nit/TIOLI: maybe `DetectResult` makes it clearer that this is a `Result` type alias? But that feels wordier so maybe this is better...I don't really care.

linkerd/detect/src/lib.rs
```diff
@@ -142,3 +151,26 @@ where
         })
     }
 }
+
+// === impl DetectTimeout ===
+
+impl<P> fmt::Debug for DetectTimeout<P> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple(std::any::type_name::<Self>())
```
hawkw, line 159: nice :)

hawkw (approved):
looks great!

PR #930, opened 2021-02-25
Remove the tcp_connection_duration_ms histogram

olix0r:
The TCP connection histogram is pretty useless in its current form: its
granularity in milliseconds is generally unsuited to long-lived
connections; and it ends up being extremely high cardinality.  And on
top of all that, the value isn't currently consumed in our UIs.

This change removes the `tcp_connection_duration_ms` histogram from the
proxy.

Relates to linkerd/linkerd2#5818

### DIFF
--- a/linkerd/app/integration/src/tests/telemetry.rs
+++ b/linkerd/app/integration/src/tests/telemetry.rs
@@ -1260,51 +1260,6 @@ mod transport {
             .await;
     }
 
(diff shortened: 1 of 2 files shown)


hawkw (approved):
in theory it definitely seems like some kind of connection duration data could be useful, but yeah, the current metric doesn't seem great. looks good to me!

PR #931, opened 2021-03-01
server: Log connection closed messages at DEBUG

olix0r:
Per linkerd/linkerd2#5331, it seems that "Connection closed" log
messages, which in many cases are compltely innocuous, cause concern
when people see them in production.

This change moves these messages from the INFO to DEBUG log level so
that they are only observed when explicilty configured.

### DIFF
--- a/linkerd/app/core/src/serve.rs
+++ b/linkerd/app/core/src/serve.rs
@@ -5,7 +5,7 @@ use linkerd_error::Error;
 use linkerd_proxy_transport::listen::Addrs;
 use tower::util::ServiceExt;
 use tracing::instrument::Instrument;
-use tracing::{debug, info, info_span, warn};
+use tracing::{debug, info_span, warn};
 
 /// Spawns a task that binds an `L`-typed listener with an `A`-typed
 /// connection-accepting service.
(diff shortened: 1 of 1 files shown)


linkerd/app/core/src/serve.rs
```diff
@@ -52,7 +52,7 @@ where
                                         .await
                                     {
                                         Ok(()) => debug!("Connection closed"),
-                                        Err(error) => info!(%error, "Connection closed"),
+                                        Err(reason) => debug!(%reason, "Connection closed"),
```
hawkw, line 55: something that might be worth doing (but we could do later) is looking at the error's reason and seeing if it was a normal connection close or some kind of local error, which might be an actual bug (e.g. if we get, like, EBADF, that's actually something bad).

then, we could keep the log event at debug, but make the message something like "Connection closed normally" if it was a not connected/connection reset/refused. that way, if people turn on the DEBUG level to try and diagnose some problem they're seeing, they see messages like "Connection closed normally" and it's less scary?

just a thought — we should merge this as-is in any case.

hawkw (approved):
:+1:

PR #932, opened 2021-03-01
server: Log non-i/o errors at INFO

olix0r:
8c87c5de updated the server to log errors at DEBUG; but this will hide
proxy-generated errors like detection timeouts.

This change logs all i/o errors at DEBUG and restores other errors to
the INFO level.

### DIFF
--- a/linkerd/app/core/src/serve.rs
+++ b/linkerd/app/core/src/serve.rs
@@ -5,7 +5,7 @@ use linkerd_error::Error;
 use linkerd_proxy_transport::listen::Addrs;
 use tower::util::ServiceExt;
 use tracing::instrument::Instrument;
-use tracing::{debug, info_span, warn};
+use tracing::{debug, info, info_span, warn};
 
 /// Spawns a task that binds an `L`-typed listener with an `A`-typed
 /// connection-accepting service.
@@ -52,7 +52,13 @@ where
                                         .await
                                     {
                                         Ok(()) => debug!("Connection closed"),
-                                        Err(reason) => debug!(%reason, "Connection closed"),
+                                        Err(reason) => {
+                                            if is_io(&*reason) {
+                                                debug!(%reason, "Connection closed")
+                                            } else {
+                                                info!(%reason, "Connection closed")
+                                            }
+                                        }
                                     }
                                     // Hold the service until the connection is
                                     // complete. This helps tie any inner cache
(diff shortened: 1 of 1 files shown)


linkerd/app/core/src/serve.rs
```diff
@@ -52,7 +52,13 @@ where
                                         .await
                                     {
                                         Ok(()) => debug!("Connection closed"),
-                                        Err(reason) => debug!(%reason, "Connection closed"),
+                                        Err(reason) => {
+                                            if is_io(&*reason) {
+                                                debug!(%reason, "Connection closed")
+                                            } else {
+                                                info!(%reason, "Connection closed")
+                                            }
```
hawkw, line 60: TIOLI, might be a little bit more concise:
```suggestion
                                        Err(reason) if is_io(&*reason) => debug!(%reason, "Connection closed"),
                                        Err(reason) => info!(%reason, "Connection closed"),
```

linkerd/app/core/src/serve.rs
```diff
@@ -79,3 +85,7 @@ where
         _ = shutdown => { Ok(()) }
     }
 }
+
+fn is_io(e: &(dyn std::error::Error + 'static)) -> bool {
+    e.is::<io::Error>() || e.source().map(is_io).unwrap_or(false)
+}
```
hawkw, line 91: TIOLI: this is recursive, it might be a little more efficient to write a loop, like
```suggestion
fn is_io(e: &(dyn std::error::Error + 'static)) -> bool {
    let mut error = Some(e);
    while let Some(e) = error {
        if e.is::<io::Error>() {
            return true;
        }
        error = e.source();
    }
    false
}
```
but, that's significantly less concise, and i imagine that error source chains aren't really deep enough for stack depth to matter here? your call.

linkerd/app/core/src/serve.rs
```diff
@@ -79,3 +85,7 @@ where
         _ = shutdown => { Ok(()) }
     }
 }
+
+fn is_io(e: &(dyn std::error::Error + 'static)) -> bool {
```
hawkw, line 89: nit/TIOLI: i wonder if we'll want this anywhere else (e.g., how are we logging connection errors on the client side currently?). this could maybe go in `linkerd_error`? we should probably hold off on that until we actually need to use this elsewhere, though...

hawkw (approved):
lgtm!

PR #933, opened 2021-03-03
ingress: Instrument outbound TCP metrics

olix0r:
In ingress mode, the HTTP endpoint stack is not configured with the TCP
endpoint layers, so outbound connections are not properly tracked in
metrics.

This change restores this functionality.


kleimkuhler (approved):
👍

PR #934, opened 2021-03-04
inbound: Remove unnecessary connection stack cache

olix0r:
The inbound proxy caches each HTTP server stack, but this stack's cache
key includes client address information, so this cache ends up requiring
an entry per connection and stacks are not reused.

This change removes this unnecessary cache.


hawkw (approved):
:+1: nice catch!

PR #936, opened 2021-03-04
outbound: Use the OrigDstAddr type in stack targets

olix0r:
The outbound stack is ambiguous about the type of orig_dst/target
address it uses.

This change modifies the outbound stack types to use the `OrigDstAddr`
type. The outbound stacks now error when an `OrigDstAddr` type is not
present.

### DIFF
--- a/linkerd/app/outbound/src/ingress.rs
+++ b/linkerd/app/outbound/src/ingress.rs
@@ -133,7 +133,6 @@ impl Outbound<()> {
             ))
             .check_new_service::<tcp::Accept, transport::metrics::SensorIo<I>>()
             .push(self.runtime.metrics.transport.layer_accept())
-            .push_map_target(tcp::Accept::from)
             .push_request_filter(tcp::Accept::try_from)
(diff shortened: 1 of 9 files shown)


linkerd/app/outbound/src/ingress.rs
```diff
@@ -133,6 +134,7 @@ impl Outbound<()> {
             .check_new_service::<tcp::Accept, transport::metrics::SensorIo<I>>()
             .push(self.runtime.metrics.transport.layer_accept())
             .push_map_target(tcp::Accept::from)
```
hawkw, line 136: i think this should have been removed --- it still compiles because it's now just doing the identity conversion (`tcp::Accept` -> `tcp::Accept`):
```suggestion
```

hawkw (approved):
lgtm!

linkerd/app/outbound/src/ingress.rs
olix0r: whoops. good catch.

PR #937, opened 2021-03-04
duplex: include src/dst peer in logs

hawkw:
The `linkerd_duplex` crate emits trace events for every read and write
operation performed, but these events don't make it clear which peer
performed the read or write. This branch adds names to the trace event
to make it possible to determine whether a read/write was performed by
the source or destination peer.

This does make the `Duplex` struct two words larger. Not sure if we
care. It might theoretically be possible to encode this using type
parameters to avoid the `&'static str`s; I'm not sure if this is worth
adding some wacky code to avoid two words of string constant pointer.

### DIFF
--- a/linkerd/proxy/http/src/upgrade.rs
+++ b/linkerd/proxy/http/src/upgrade.rs
@@ -139,7 +139,7 @@ impl Drop for Inner {
             let both_upgrades = async move {
                 let (server_conn, client_conn) = tokio::try_join!(server_upgrade, client_upgrade)?;
                 trace!("HTTP upgrade successful");
-                if let Err(e) = Duplex::new(server_conn, client_conn).await {
+                if let Err(e) = Duplex::new(client_conn, server_conn).await {
                     info!("tcp duplex error: {}", e)
                 }
                 Ok::<(), ()>(())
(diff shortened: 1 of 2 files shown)


olix0r (commented):
This isn't actually a "peer" though, is it? duplex is necessarily _both_ peers. It's really a question of direction: src->dst or dst->src. Or perhaps server->client vs client->server

hawkw:
> This isn't actually a "peer" though, is it? duplex is necessarily _both_ peers. It's really a question of direction: src->dst or dst->src. Or perhaps server->client vs client->server

Hmm, yeah, calling it "direction" probably makes more sense :thumbsup:

olix0r (approved):

linkerd/proxy/http/src/upgrade.rs
```diff
@@ -139,7 +139,7 @@ impl Drop for Inner {
             let both_upgrades = async move {
                 let (server_conn, client_conn) = tokio::try_join!(server_upgrade, client_upgrade)?;
                 trace!("HTTP upgrade successful");
-                if let Err(e) = Duplex::new(server_conn, client_conn).await {
+                if let Err(e) = Duplex::new(client_conn, server_conn).await {
```
hawkw, line 142: these had to be reversed to get the correct direction labels, which, admittedly, is not the best. I also considered passing each label in as an argument (which feels a little confusing since there's now 4 positionals), or making `HalfDuplex` public and making the constructor take two `HalfDuplex`es so that the user adds the direction labels. but, those are both a bit more complicated and since `Duplex::new` is only called in two places, i dunno how much this matters, really...

PR #938, opened 2021-03-05
outbound: Decouple the resolver from concrete target types

olix0r:
The endpoint resolver module currently has a fixed dependency on the
`Concrete` and `Endpoint` target types.

This change extracts endpoint-type construction so that the HTTP logical
stack is agnostic of its endpoint target type. It also updates the
`EndpointFromMetadata` type to be responsible for marking identity as
disabled when appropriate. Finally, we are able to re-enable the
outbound TLS hinting test.

### DIFF
--- a/linkerd/app/outbound/src/http/logical.rs
+++ b/linkerd/app/outbound/src/http/logical.rs
@@ -1,9 +1,9 @@
-use super::{Concrete, Endpoint, Logical};
+use super::{Concrete, Logical};
 use crate::{resolve, stack_labels, Outbound};
 use linkerd_app_core::{
     classify, config, profiles,
-    proxy::{api_resolve::Metadata, core::Resolve, http},
-    retry, svc, tls, Error, CANONICAL_DST_HEADER, DST_OVERRIDE_HEADER,
+    proxy::{core::Resolve, http},
+    retry, svc, tls, Error, Never, CANONICAL_DST_HEADER, DST_OVERRIDE_HEADER,
 };
 use tracing::debug_span;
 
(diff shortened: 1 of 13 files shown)


linkerd/app/outbound/src/http/logical.rs
```diff
@@ -147,7 +154,24 @@ impl<E> Outbound<E> {
...
             .push_switch(
-                Logical::or_endpoint(tls::NoClientTls::NotProvidedByServiceDiscovery),
+                |logical: Logical| {
+                    let should_resolve = match logical.profile.as_ref() {
+                        Some(p) => {
+                            let p = p.borrow();
+                            p.endpoint.is_none() && (p.name.is_some() || !p.targets.is_empty())
+                        }
+                        None => false,
+                    };
```
hawkw, line 164: seems like this could still be abstracted over by a method on `Logical`?

linkerd/app/outbound/src/tcp/logical.rs
```diff
@@ -90,9 +86,10 @@ where
             .push(svc::UnwrapOr::layer(
                 endpoint
                     .clone()
-                    .push_map_target(Endpoint::from_logical(
-                        tls::NoClientTls::NotProvidedByServiceDiscovery,
-                    ))
+                    .push_map_target(|logical: Logical| {
+                        debug!("No profile resolved");
```
hawkw, line 90: nit: should this include the logical endpoint we didn't resolve a profile for?

hawkw (approved):

linkerd/app/outbound/src/tcp/logical.rs
olix0r: the address is already in a parent span

linkerd/app/outbound/src/http/logical.rs
olix0r: let's see where it goes -- I'm trying to decouple all of this from `Logical`.

PR #940, opened 2021-03-06
Make stack trace initialization lazy

olix0r:
The stack-tracing module eagerly initiates its spans once per service;
but when the service is cached, this locks the service to its original
parent span so that log messages from subsequent accessors appear to
have the wrong parent span information.

This change modifies the stack tracing helper to (1) be
NewService-specific and (2) to initiate unique spans for each call. The
instrumented `Service` constructs a new span the first time `poll_ready`
is called, and consumes that span in `call`, so the `poll_ready`s that
drive a service to readiness and the subsequent `call` that consumes
that readiness are both considered one individual span for that request.

### DIFF
--- a/linkerd/app/outbound/src/tcp/logical.rs
+++ b/linkerd/app/outbound/src/tcp/logical.rs
@@ -46,14 +46,13 @@ where
         let endpoint = connect
             .clone()
             .push_make_thunk()
-            .push(svc::MapErrLayer::new(Into::into))
             .instrument(|t: &Endpoint| debug_span!("tcp.forward", server.addr = %t.addr))
             .push_on_response(
                 svc::layers()
+                    .push(svc::MapErrLayer::new(Into::into))
                     .push(tcp::Forward::layer())
                     .push(drain::Retain::layer(rt.drain.clone())),
-            )
-            .into_new_service();
+            );
 
         let stack = connect
             .push_make_thunk()
--- a/linkerd/stack/tracing/src/lib.rs
+++ b/linkerd/stack/tracing/src/lib.rs
@@ -1,8 +1,5 @@
-use futures::TryFuture;
-use linkerd_stack::{NewService, Proxy};
+use linkerd_stack::{layer, NewService, Proxy};
 use pin_project::pin_project;
-use std::future::Future;
-use std::pin::Pin;
 use std::task::{Context, Poll};
 use tracing::instrument::{Instrument as _, Instrumented};
 use tracing::{trace, Span};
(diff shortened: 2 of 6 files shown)


linkerd/app/outbound/src/tcp/logical.rs
```diff
@@ -46,14 +46,13 @@ where
         let endpoint = connect
             .clone()
             .push_make_thunk()
-            .push(svc::MapErrLayer::new(Into::into))
```
hawkw, line 49: just out of curiosity: why'd this have to move?

linkerd/stack/tracing/src/lib.rs
```diff
@@ -148,23 +108,26 @@ where
     type Future = Instrumented<P::Future>;
 
     fn proxy(&self, svc: &mut S, request: Req) -> Self::Future {
-        let _enter = self.span.enter();
+        let span = self.get_span();
+        let _enter = span.enter();
         trace!(?request, "proxy");
-        self.inner.proxy(svc, request).instrument(self.span.clone())
+        self.inner.proxy(svc, request).instrument(span.clone())
```
hawkw, line 114: nit/TIOLI: we could avoid the span clone (atomic ref count bump) if we did:
```rust
let span = self.get_span();
let future = span.in_scope(|| {
    trace!(?request, "proxy");
    self.inner.proxy(svc, request)
});
future.instrument(span.clone())
```
or this, which IMO is the nicest (but requires updating to the latest `tracing` release to get the new [`Span::entered` method](https://docs.rs/tracing/0.1.25/tracing/span/struct.Span.html#method.entered)):
```rust
let span = self.get_span().entered();
trace!(?request, "proxy");
self.inner.proxy(svc, request).instrument(span.exit())
```

linkerd/stack/tracing/src/lib.rs
```diff
@@ -175,10 +138,11 @@ where
     }
 
     fn call(&mut self, request: Req) -> Self::Future {
-        let _enter = self.span.enter();
+        let span = self.get_span();
+        let _enter = span.enter();
 
         trace!(?request, "service");
-        self.inner.call(request).instrument(self.span.clone())
+        self.inner.call(request).instrument(span.clone())
```
hawkw, line 145: similarly, if we update our `tracing` dep to v0.1.25, we could write
```rust
let span = self.get_span().entered();

trace!(?request, "service");
self.inner.call(request).instrument(span.exit())
```
avoiding the span ref bump/dec from cloning and dropping the span, and saving one line :)

linkerd/stack/tracing/src/lib.rs
```diff
@@ -148,23 +108,26 @@ where
...
+    G: GetSpan<T>,
     S: tower::Service<Req>,
 {
     type Response = S::Response;
     type Error = S::Error;
     type Future = Instrumented<S::Future>;
 
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        let _enter = self.span.enter();
+        let span = self.get_span();
```
hawkw, line 129: just a thinky: creating a new separate span for `poll_ready` kiiiiinda makes me wonder if we want that span to _indicate_ that it's a `poll_ready`...but that would be a bigger change to make the `GetSpan` stuff aware of `poll_ready` calls somehow, or to call `get_span` and then immediately create a new child of that span for the `poll_ready`. so maybe let's punt on that for now...

hawkw: yet another idle thinky is that _maybe_ the most semantically correct thing is to have one span that covers poll_ready and the next call after that poll_ready, like
```rust
struct InstrumentedService<S, G> {
    span: Option<Span>,
    get_span: G,
    // ...
}

impl<S, G, R> Service<R> for InstrumentedService<S, G> 
where
    // ...
{
    // ...
    
         let _enter = self.span
            .get_or_insert_with(|| self.get_span(self.target))
            .enter();
            
        // ...
    }

    fn call(&mut self, req: Request) -> Self::Future {
        let span = self.span
            .take()
            .unwrap_or_else(|| self.get_span(self.target))
            .entered();

       trace!(?request, "service");
       self.inner.call(request).instrument(span.exit())
    }
}
```
that way, you have one span whose total active time is the sum of the time spent polling the service, calling it, and polling its future, and whose total idle time includes the time between when the service was driven to readiness and when it was called. but, this maybe only matters if we are actually timing spans, rather than just using them for log contexts.

also, this means that the future's span's parent is the current span when the service was _polled_, which may not always be the same span the service was _called_ in...

anyway, just thinking out loud here.

hawkw (approved):
this looks good to me! 

i had a couple of thoughts: we can make this a little more efficient (and a bit more concise) if we use the new [`Span::entered` method](https://docs.rs/tracing/0.1.25/tracing/span/struct.Span.html#method.entered) in the latest `tracing` release; and, an idle thought about how to get the most theoretically correct timing data from spans which we probably don't actually care about. but, neither of those are blockers (although the `entered` change is a pretty small tweak that might be worth it).

linkerd/stack/tracing/src/lib.rs
hawkw: I made this change in ad87c97.

olix0r:
lgtm. I think the docker build is failing due to changes in Equinix (we killed the build host...) fine to ignore for now

PR #944, opened 2021-03-09
duplex: Ensure written data is flushed

olix0r:
In testing Postgres dumps through Linkerd, we see some situations where
data is written to rustls but does not become visible to the client.
Rustls documents that written data may need to be flushed.

This change updates the `HalfDuplex::copy_into` function to flush data
when all available data has been read. This change also increases the
copy buffer from 4KB to 64KB to reduce processing overhead.

---

Relates to linkerd/linkerd2#5539

Image available at `ghcr.io/olix0r/l2-proxy:flush.863b537b`

### DIFF
--- a/linkerd/duplex/src/lib.rs
+++ b/linkerd/duplex/src/lib.rs
@@ -116,10 +116,8 @@ where
         // shutdown, we finished in a previous poll, so don't even enter into
         // the copy loop.
         if dst.is_shutdown {
-            // A prior shutdown may not have completed, so continue driving the
-            // shutdown.
             trace!(direction = %self.direction, "already shutdown");
-            return Pin::new(&mut dst.io).poll_shutdown(cx);
+            return Poll::Ready(Ok(()));
         }
 
         // If the last invocation returned pending while flushing, resume
@@ -128,7 +126,10 @@ where
             ready!(self.poll_flush(dst, cx))?;
         }
 
+        // `needs_flush` is set to true if the buffer is written so that, if a
+        // read returns pending, that data may be flushed.
         let mut needs_flush = false;
+
         loop {
             // As long as the underlying socket is alive, ensure we've read data
             // from it into the local buffer.
@@ -139,7 +140,9 @@ where
                     if needs_flush {
                         // The poll status of the flush isn't relevant, as we
                         // have registered interest in the read (and maybe the
-                        // write as well).
+                        // write as well). If the flush did not complete
+                        // `self.flushing` is true so that it may be resumed on
+                        // the next poll.
                         let _ = self.poll_flush(dst, cx)?;
                     }
                     return Poll::Pending;
@@ -168,7 +171,10 @@ where
                             needs_flush = false;
                         }
                         Drained::BufferEmpty => {
-                            error!("Invalid state: attempted to write from an empty buffer");
+                            error!(
+                                direction = self.direction,
+                                "Invalid state: attempted to write from an empty buffer"
+                            );
                             debug_assert!(false, "The write buffer should never be empty");
                             return Poll::Ready(Ok(()));
                         }
@@ -197,7 +203,7 @@ where
         // To do this, we'd have to get more complex about handling EOF.
         if let Some(buf) = self.buf.as_mut() {
             if buf.has_remaining() {
-                // Data was already buffered, so just return 0 immediately.
+                // Data was already buffered, so just return immediately.
                 trace!(direction = %self.direction, remaining = buf.remaining(), "skipping read");
                 return Poll::Ready(Ok(Buffered::NotEmpty));
             }
@@ -219,6 +225,8 @@ where
         Poll::Ready(Ok(Buffered::Eof))
     }
 
+    /// Attempts to flush the destination. `self.flushing` is set to true iff the
+    /// flush operation did not complete.
     fn poll_flush<U: AsyncWrite + Unpin>(
         &mut self,
         dst: &mut HalfDuplex<U>,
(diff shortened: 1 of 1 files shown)


linkerd/duplex/src/lib.rs
```diff
@@ -85,81 +97,170 @@ where
...
+            // A prior shutdown may not have completed, so continue driving the
+            // shutdown.
             trace!(direction = %self.direction, "already shutdown");
-            return Poll::Ready(Ok(()));
+            return Pin::new(&mut dst.io).poll_shutdown(cx);
+        }
+
+        // If the last invocation returned pending while flushing, resume
+        // flushing and only proceed when the flush is complete.
+        if self.flushing {
```
hawkw, line 127: nit: we log when continuing a shutdown; should we log when continuing a flush as well?

linkerd/duplex/src/lib.rs
```diff
@@ -24,6 +23,7 @@ struct HalfDuplex<T> {
     #[pin]
     io: T,
     direction: &'static str,
+    flushing: bool,
```
hawkw, line 26: nit/TIOLI: we can't, AFAICT, be both flushing and already shutdown, we *could* consider stuffing these into an

```rust
enum State {
   Reading,
   Flushing,
   ShuttingDown,
}
```
not sure if that matters

linkerd/duplex/src/lib.rs
```diff
@@ -85,81 +97,170 @@ where
...
+            // As long as the underlying socket is alive, ensure we've read data
+            // from it into the local buffer.
+            match self.poll_buffer(cx)? {
+                Poll::Pending => {
+                    // If there's no data to be read and we've written data, try
+                    // flushing before returning pending.
+                    if needs_flush {
+                        // The poll status of the flush isn't relevant, as we
+                        // have registered interest in the read (and maybe the
+                        // write as well).
```
hawkw, line 142: might be worth also mentioning that `self.poll_flush` sets the flushing flag so we'll continue the flush if necessary

linkerd/duplex/src/lib.rs
```diff
@@ -85,81 +97,170 @@ where
...
+            return Pin::new(&mut dst.io).poll_shutdown(cx);
+        }
+
+        // If the last invocation returned pending while flushing, resume
+        // flushing and only proceed when the flush is complete.
+        if self.flushing {
+            ready!(self.poll_flush(dst, cx))?;
         }
+
+        let mut needs_flush = false;
```
hawkw, line 131: nit: might be worth a comment on this explaining the relationship between `needs_flush` (we should start a flush) and `self.flushing` (we are continuing a previous flush). i was able to figure it out but it's not _immediately_ obvious imo...

linkerd/duplex/src/lib.rs
```diff
@@ -85,81 +97,170 @@ where
...
         // return value, we may be polled again after returning Ready, if the
         // other half isn't ready. In that case, if the destination has
         // shutdown, we finished in a previous poll, so don't even enter into
         // the copy loop.
         if dst.is_shutdown {
+            // A prior shutdown may not have completed, so continue driving the
+            // shutdown.
             trace!(direction = %self.direction, "already shutdown");
-            return Poll::Ready(Ok(()));
+            return Pin::new(&mut dst.io).poll_shutdown(cx);
```
olix0r, line 122: Actually, `dst.is_shutdown` can't be true unless shutdown completed. this should go back to how it was.

linkerd/duplex/src/lib.rs
```diff
@@ -85,81 +97,170 @@ where
...
+                        // Only some of the buffered data could be written
+                        // before the destination became pending. Try to flush
+                        // the written data to get capacity.
+                        Drained::Partial(_) => {
+                            ready!(self.poll_flush(dst, cx))?;
+                            // If the flush completed, try writing again to
+                            // ensure that we have a notification registered. If
+                            // all of the buffered data still cannot be written,
+                            // return pending. Otherwise, continue.
+                            if let Drained::Partial(_) = self.drain_into(dst, cx)? {
```
hawkw, line 165: do we ever care about the amount drained in this state?

linkerd/duplex/src/lib.rs
```diff
@@ -85,81 +97,170 @@ where
...
-        if let Some(ref mut buf) = self.buf {
+    ) -> io::Result<Drained> {
+        let mut sz = 0;
+
+        if let Some(buf) = self.buf.as_mut() {
             while buf.has_remaining() {
                 trace!(direction = %self.direction, "writing {}B", buf.remaining());
-                let n = ready!(io::poll_write_buf(Pin::new(&mut dst.io), cx, buf))?;
+                let n = match io::poll_write_buf(Pin::new(&mut dst.io), cx, buf)? {
+                    Poll::Pending => return Ok(Drained::Partial(sz)),
```
hawkw, line 248: AFAICT `sz` is never used?

hawkw (approved):
lgtm overall, couple of nits to pick

linkerd/duplex/src/lib.rs
olix0r: poll_flush has its own logging

linkerd/duplex/src/lib.rs
olix0r: similarly, not used but useful from the api pov.

linkerd/duplex/src/lib.rs
```diff
@@ -39,6 +39,18 @@ struct CopyBuf {
     write_pos: usize,
 }
 
+enum Buffered {
+    NotEmpty,
+    Read(usize),
```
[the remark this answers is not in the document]
olix0r: correct, but useful from api pov; and not really high-cost

PR #945, opened 2021-03-11
Prevent fixed-address resolutions from ending

olix0r:
When a controller client is configured with a fixed IP address and not a
resolveable name, we build that client with a resolution that terminates
immediately after the client is configured with the fixed address. This
causes warnings to be emmited even though the client is configured
properly.

This change updates theses resolutions to avoid termination so these
warnings are not logged.


hawkw (approved):
lgtm!

PR #946, opened 2021-03-12
http: Parameterize authority overriding

olix0r:
The authority override module still uses a dedicated trait,
`CanOverrideAuthority`. This change replaces this trait with a newtype,
`AuthorityOverride`, which is provided via `Param`.


hawkw (approved):
:+1:

PR #947, opened 2021-03-15
tracing: Avoid high-cardinality client in INFO spans

olix0r:
Our default TCP server sets an info-level tracing span that includes the
client IP and port. This client-level metadata is potentially
high-cardinality, it's not helpful on the outbound side, and it is only
rendered in rare situations.

This change modifies the server to set its `accept` span on the debug
level. The inbound and outbound proxies handle setting info-level spans
with the relevant metadata -- the inbound proxy includes only a server
port, and the outbound includes the original dst address. The error
responder has been modified to include the client address in warning
messages so that it need not be derived from span metadata.


hawkw (approved):
lgtm!

PR #949, opened 2021-03-17
outbound: Configure endpoint construction in logical stack

olix0r:
ebf67d6b moved endpoint construction logic out of the logical stacks. This
created more boilerplate than hoped. Realistically, the logical stack needs to
be responsible for its inner target types.

This change moves endpoint construction into each logical stack so that
it need not be configured on each instantiation, simplifying resolver type
constraints.

### DIFF
--- a/linkerd/app/outbound/src/http/logical.rs
+++ b/linkerd/app/outbound/src/http/logical.rs
@@ -54,6 +54,11 @@ impl<E> Outbound<E> {
         let watchdog = cache_max_idle_age * 2;
 
         let identity_disabled = rt.identity.is_none();
+        let no_tls_reason = if identity_disabled {
+            tls::NoClientTls::Disabled
+        } else {
+            tls::NoClientTls::NotProvidedByServiceDiscovery
+        };
         let resolve = svc::stack(resolve.into_service())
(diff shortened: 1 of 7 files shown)


linkerd/app/outbound/src/http/logical.rs
```diff
@@ -100,11 +122,9 @@ impl<E> Outbound<E> {
                             .push(http::BoxRequest::layer())
                             .push(http::BoxResponse::layer()),
                     )
-                    .push_map_target(|logical: Logical| {
-                        R::Endpoint::from((
-                            tls::NoClientTls::NotProvidedByServiceDiscovery,
-                            logical,
-                        ))
+                    .push_map_target({
+                        let no_tls_reason = no_tls_reason;
```
hawkw, line 126: what is this doing, exactly? is this a copy?

hawkw (approved):
looks good to me; one thing that looked kind of weird but it's not a blocker.

linkerd/app/outbound/src/http/logical.rs
olix0r: forgot we didn't need to rebind `Copy`

PR #951, opened 2021-03-18
outbound: Decouple the TCP connect stack from the target type

olix0r:
In order to make the server and logical stack more flexible, we need to
make stacks generic over their input target type.

This change modifies the innermost `tcp::connect` stack to not depend on
a concrete target type, instead relying on `Param` constraints. A new
`tcp::Connect` target type is introduced to be constructed by the TCP
connect stack.

### DIFF
--- a/linkerd/app/outbound/src/target.rs
+++ b/linkerd/app/outbound/src/target.rs
@@ -1,7 +1,9 @@
+use crate::tcp::opaque_transport;
 use linkerd_app_core::{
     metrics, profiles,
     proxy::{
         api_resolve::{ConcreteAddr, Metadata},
+        http,
         resolve::map_endpoint::MapEndpoint,
     },
     svc::{self, Param},
@@ -11,11 +13,6 @@ use linkerd_app_core::{
 };
 use std::net::SocketAddr;
 use tracing::debug;
-
-#[derive(Copy, Clone)]
-pub struct EndpointFromMetadata {
-    pub identity_disabled: bool,
-}
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub struct Accept<P> {
@@ -43,6 +40,11 @@ pub struct Endpoint<P> {
     pub metadata: Metadata,
     pub logical_addr: Addr,
     pub protocol: P,
+}
+
+#[derive(Copy, Clone)]
+pub struct EndpointFromMetadata {
+    pub identity_disabled: bool,
 }
 
 // === impl Accept ===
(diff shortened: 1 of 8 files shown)


linkerd/app/outbound/src/target.rs
```diff
@@ -45,6 +42,11 @@ pub struct Endpoint<P> {
     pub protocol: P,
 }
 
+#[derive(Copy, Clone)]
+pub struct EndpointFromMetadata {
+    pub identity_disabled: bool,
+}
+
```
hawkw, line 49: is there a reason this moved around? not a big deal.

linkerd/app/outbound/src/tcp/connect.rs
```diff
@@ -25,19 +33,24 @@ impl Outbound<()> {
...
     >
     where
-        Endpoint<P>: svc::Param<Option<SessionProtocol>>,
-        C: svc::Service<Endpoint<P>, Error = io::Error> + Clone + Send + 'static,
+        T: svc::Param<Remote<ServerAddr>>
+            + svc::Param<tls::ConditionalClientTls>
+            + svc::Param<Option<opaque_transport::PortOverride>>
+            + svc::Param<Option<http::AuthorityOverride>>
+            + svc::Param<Option<SessionProtocol>>
+            + svc::Param<transport::labels::Key>,
```
hawkw, line 52: whoa, hmm. part of me wonders whether it might be worth making a trait alias for "all these params", but that might just be unnecessary complexity...

linkerd/app/outbound/src/tcp/opaque_transport.rs
```diff
@@ -131,7 +158,17 @@ mod test {
...
+                .identity()
+                .map(|id| {
+                    Conditional::Some(tls::ClientTls {
+                        server_id: id.clone(),
+                        alpn: Some(tls::client::AlpnProtocols(vec![PROTOCOL.into()])),
+                    })
+                })
+                .unwrap_or(Conditional::None(
+                    tls::NoClientTls::NotProvidedByServiceDiscovery,
+                )),
```
hawkw, line 171: wondering how much boilerplate could be saved by some kind of "`Conditional::from_option`-type helper that just takes a reason for the `None` branch. not a blocker for this PR...

linkerd/app/outbound/src/tcp/connect.rs
olix0r: Would a trait alias actually reduce the complexity? We generally don't have to copy them everywhere. So the trait alias would be at least as much code...

I think this probably good/fine as-is--for now, at least. I think we want each of these stack modules to basically take a generic target that provides a set of params; and then it builds its inner target type as necessary (either wrapping or replaces the initial target). Once we've wired this approach we may find a way to simplify it further.

linkerd/app/outbound/src/target.rs
olix0r: just putting near `Endpoint`

linkerd/app/outbound/src/tcp/connect.rs
hawkw: yeah, we're on the same page. i was just thinking about it, but i think this is actually better than wrapping it into one trait.

PR #953, opened 2021-03-18
outbound: Make the HTTP server stack generic

olix0r:
The HTTP server stack is unnecessarily bound to a target type.

In order to support more flexible stack composition, this change makes
the outbound HTTP server stack generic over its target type.


hawkw (approved):
:+1: lgtm

PR #954, opened 2021-03-18
Update profile response to include a logical address

olix0r:
Profile responses optionally include a service name, but we only ever
use it in the context of an address (i.e. with a port).

This change introduces a `profiles::LookupAddr` param type that is used
as input into the profile client -- the lookup address may hold either a
named or numbered address. Service profile responses now include a
`LogicalAddr` that must be a named address.

This sets up for requiring that concrete addresses are named.

### DIFF
--- a/linkerd/app/gateway/src/lib.rs
+++ b/linkerd/app/gateway/src/lib.rs
@@ -75,15 +75,11 @@ pub fn stack<I, O, P, R>(
        + Send
 where
     I: io::AsyncRead + io::AsyncWrite + io::PeerAddr + fmt::Debug + Send + Sync + Unpin + 'static,
-    O: svc::Service<outbound::http::Endpoint, Error = io::Error>
-        + svc::Service<outbound::tcp::Endpoint, Error = io::Error>,
     O: Clone + Send + Sync + Unpin + 'static,
-    <O as svc::Service<outbound::http::Endpoint>>::Response:
+    O: svc::Service<outbound::tcp::Connect, Error = io::Error>,
+    O::Response:
         io::AsyncRead + io::AsyncWrite + tls::HasNegotiatedProtocol + Send + Unpin + 'static,
-    <O as svc::Service<outbound::http::Endpoint>>::Future: Send + Unpin + 'static,
-    <O as svc::Service<outbound::tcp::Endpoint>>::Response:
-        io::AsyncRead + io::AsyncWrite + tls::HasNegotiatedProtocol + Send + Unpin + 'static,
-    <O as svc::Service<outbound::tcp::Endpoint>>::Future: Send + Unpin + 'static,
+    O::Future: Send + Unpin + 'static,
     P: profiles::GetProfile<profiles::LogicalAddr> + Clone + Send + Sync + Unpin + 'static,
     P::Future: Send + 'static,
     P::Error: Send,
--- a/linkerd/app/gateway/src/lib.rs
+++ b/linkerd/app/gateway/src/lib.rs
@@ -84,7 +84,7 @@ where
     O::Response:
         io::AsyncRead + io::AsyncWrite + tls::HasNegotiatedProtocol + Send + Unpin + 'static,
     O::Future: Send + Unpin + 'static,
-    P: profiles::GetProfile<profiles::LogicalAddr> + Clone + Send + Sync + Unpin + 'static,
+    P: profiles::GetProfile<profiles::LookupAddr> + Clone + Send + Sync + Unpin + 'static,
     P::Future: Send + 'static,
     P::Error: Send,
     R: Clone + Send + Sync + Unpin + 'static,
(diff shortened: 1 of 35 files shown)


linkerd/app/gateway/src/lib.rs
```diff
@@ -84,7 +84,7 @@ where
     O::Response:
         io::AsyncRead + io::AsyncWrite + tls::HasNegotiatedProtocol + Send + Unpin + 'static,
     O::Future: Send + Unpin + 'static,
-    P: profiles::GetProfile<profiles::LogicalAddr> + Clone + Send + Sync + Unpin + 'static,
+    P: profiles::GetProfile<profiles::LookupAddr> + Clone + Send + Sync + Unpin + 'static,
```
hawkw, line 87: AFAICT, there are now no impls of `GetProfile` for anything other than a `LookupAddr`:
```bash
:# eliza on butterfly in linkerd2-proxy on  ver/profile-addr [$?] via ⚙️ v1.49.0 
:; rg "GetProfile<"
linkerd/service-profiles/src/default.rs
25:    S: GetProfile<T>,

linkerd/service-profiles/src/discover.rs
14:    G: GetProfile<F::Request> + Clone,
34:    G: GetProfile<T>,

linkerd/service-profiles/src/lib.rs
54:pub trait GetProfile<T> {
68:impl<T, S> GetProfile<T> for S
84:    P: GetProfile<T>,

linkerd/app/outbound/src/discover.rs
28:        P: profiles::GetProfile<profiles::LookupAddr> + Clone + Send + 'static,

linkerd/app/outbound/src/lib.rs
113:        P: profiles::GetProfile<profiles::LookupAddr> + Clone + Send + 'static,
141:        P: profiles::GetProfile<profiles::LookupAddr> + Clone + Send + Sync + Unpin + 'static,

linkerd/app/outbound/src/ingress.rs
44:        P: profiles::GetProfile<profiles::LookupAddr> + Clone + Send + Sync + Unpin + 'static,

linkerd/app/gateway/src/lib.rs
87:    P: profiles::GetProfile<profiles::LookupAddr> + Clone + Send + Sync + Unpin + 'static,

linkerd/app/inbound/src/lib.rs
138:        P: profiles::GetProfile<profiles::LookupAddr> + Clone + Send + Sync + 'static,
228:        P: profiles::GetProfile<profiles::LookupAddr> + Clone + Send + Sync + 'static,

linkerd/app/inbound/src/http/mod.rs
118:        P: profiles::GetProfile<profiles::LookupAddr> + Clone + Send + Sync + 'static,

```

Should we maybe just remove the type parameter and change `GetProfile` to
```rust
pub trait GetProfile {
    type Error: Into<Error>;
    type Future: Future<Output = Result<Option<Receiver>, Self::Error>>;

    fn get_profile(&mut self, target: LookupAddr) -> Self::Future;
}
```
or am I missing something?

linkerd/service-profiles/src/lib.rs
```diff
@@ -105,3 +112,59 @@ fn stream_profile(mut rx: Receiver) -> Pin<Box<dyn Stream<Item = Profile> + Send
...
+    fn from(na: NameAddr) -> Self {
+        Self(na)
+    }
+}
+
+impl Into<NameAddr> for LogicalAddr {
+    fn into(self) -> NameAddr {
+        self.0
+    }
+}
```
hawkw, line 170: some of these trivial impls can probably be `#[inline]`, though I'm not sure if it it actually matters; i just noticed we added the attribute on other code in this PR...

linkerd/service-profiles/src/lib.rs
```diff
@@ -25,16 +26,20 @@ pub type Receiver = tokio::sync::watch::Receiver<Profile>;
...
     pub endpoint: Option<(SocketAddr, Metadata)>,
 }
 
 /// A profile lookup target.
-#[derive(Clone, Debug)]
-pub struct LogicalAddr(pub Addr);
+#[derive(Clone, Debug, Hash, Eq, PartialEq)]
+pub struct LookupAddr(pub Addr);
+
+/// A bound logical service address
```
hawkw, line 40: nit: should the comment maybe make it clearer that this is part of the profile _returned_ by a lookup, in contrast to `LookupAddr`, which is the input?

linkerd/service-profiles/src/lib.rs
hawkw, line 170: a lot of these are pretty trivial, maybe they should be marked as `#[inline]`able? not sure if this matters at all, just noticed that we added the attribute elsewhere in this branch...

hawkw (approved):
overall, this looks good to me! the change is actually much more straightforward than it seems from the changed lines count; it just touches a lot of files.

i did wonder if it makes sense to get rid of the type parameter on `GetProfile`, but that could also happen separately.

linkerd/app/gateway/src/lib.rs
olix0r: seems like a fine change but it will touch more than this pr needs to.

hawkw: 👍 i can throw together a PR after this lands

PR #955, opened 2021-03-19
inbound, outbound: `Param`-ify `listen::Addrs`

hawkw:
This PR changes the `Service`s constructed by the inbound and outbound
stacks from always taking the `listen::Addrs` type as a target, to being
parameterized over a target type that implements `Param` for various
address types.

This required changing some middleware layers that were previously
hardcoded to `listen::Addrs`, and adding `Param` impls to
`listen::Addrs` for various address types. I also added a new
`TcpAccept::from_local_addr` constructor, which sets the target address
based on the server's listen address, rather than from the
SO_ORIGINAL_DST address. This is used by the admin server, which doesn't
configure its listening socket to include original dst addresses.
Previously, the `From` implementation for `TcpAccept` just always tried
the orig dst addr and fell back to the listen addr if it was not
present. Now, we have separate behavior for the admin server and the
proxy servers --- admin always uses the local addr, and the proxies only
try SO_ORIGINAL_DST or fail.

There's no functional change here, but this will set up for a future
change to make stacks generic over SO_ORIGINAL_DST detection, allowing
us to remove the currently quite awkward feature flagging.

### DIFF
--- a/linkerd/app/inbound/src/direct.rs
+++ b/linkerd/app/inbound/src/direct.rs
@@ -52,32 +52,32 @@ pub struct ClientInfo {
 type FwdIo<I> = io::PrefixedIo<SensorIo<tls::server::Io<I>>>;
 pub type GatewayIo<I> = io::EitherIo<FwdIo<I>, SensorIo<tls::server::Io<I>>>;
 
-impl<T> Inbound<T> {
+impl<N> Inbound<N> {
     /// Builds a stack that handles connections that target the proxy's inbound port
     /// (i.e. without an SO_ORIGINAL_DST setting). This port behaves differently from
     /// the main proxy stack:
     ///
     /// 1. Protocol detection is always performed;
     /// 2. TLS is required;
-    /// 3. A transport header is expected. It's not strictly required, as
+    /// 3. T transport header is expected. It's not strictly required, as
     ///    gateways may need to accept HTTP requests from older proxy versions
-    pub fn push_direct<A, I, TSvc, G, GSvc>(
+    pub fn push_direct<T, I, NSvc, G, GSvc>(
         self,
         gateway: G,
     ) -> Inbound<
         impl svc::NewService<
-                A,
+                T,
                 Service = impl svc::Service<I, Response = (), Error = Error, Future = impl Send>,
             > + Clone,
     >
     where
-        A: Param<Remote<ClientAddr>> + Param<TargetAddr> + Clone + Send + 'static,
+        T: Param<Remote<ClientAddr>> + Param<TargetAddr> + Clone + Send + 'static,
         I: io::AsyncRead + io::AsyncWrite + io::Peek + io::PeerAddr,
         I: Debug + Send + Sync + Unpin + 'static,
-        T: svc::NewService<TcpEndpoint, Service = TSvc> + Clone + Send + Sync + Unpin + 'static,
-        TSvc: svc::Service<FwdIo<I>, Response = ()> + Clone + Send + Sync + Unpin + 'static,
-        TSvc::Error: Into<Error>,
-        TSvc::Future: Send + Unpin,
+        N: svc::NewService<TcpEndpoint, Service = NSvc> + Clone + Send + Sync + Unpin + 'static,
+        NSvc: svc::Service<FwdIo<I>, Response = ()> + Clone + Send + Sync + Unpin + 'static,
+        NSvc::Error: Into<Error>,
+        NSvc::Future: Send + Unpin,
         G: svc::NewService<GatewayConnection, Service = GSvc>
             + Clone
             + Send
@@ -159,7 +159,7 @@ impl<T> Inbound<T> {
                 rt.identity.clone().map(WithTransportHeaderAlpn),
                 detect_timeout,
             ))
-            .check_new_service::<A, I>();
+            .check_new_service::<T, I>();
 
         Inbound {
             config,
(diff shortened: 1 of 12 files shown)


linkerd/app/inbound/src/direct.rs
```diff
@@ -61,16 +61,17 @@ impl<T> Inbound<T> {
     /// 2. TLS is required;
     /// 3. A transport header is expected. It's not strictly required, as
     ///    gateways may need to accept HTTP requests from older proxy versions
-    pub fn push_direct<I, TSvc, G, GSvc>(
+    pub fn push_direct<A, I, TSvc, G, GSvc>(
```
olix0r, line 64: nit: let's use `T` and change `T/TSvc` to `N/NSvc`. I've found being consistent about this helps a lot

hawkw: Done!

linkerd/app/inbound/src/direct.rs
```diff
@@ -261,7 +267,7 @@ impl std::error::Error for RefusedNoIdentity {}
 
 impl std::fmt::Display for RefusedNoTarget {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "A named target must be provided on gateway connections")
+        write!(f, "T named target must be provided on gateway connections")
```
olix0r, line 270: oops
```suggestion
        write!(f, "A named target must be provided on gateway connections")
```

linkerd/app/inbound/src/direct.rs
```diff
@@ -170,22 +171,27 @@ impl<T> Inbound<T> {
 
 // === impl ClientInfo ===
 
-impl TryFrom<(tls::ConditionalServerTls, listen::Addrs)> for ClientInfo {
+impl<N> TryFrom<(tls::ConditionalServerTls, N)> for ClientInfo
+where
+    N: Param<TargetAddr>,
+    N: Param<Remote<ClientAddr>>,
```
olix0r, line 177:
```suggestion
impl<T> TryFrom<(tls::ConditionalServerTls, T)> for ClientInfo
where
    T: Param<Local<ServerAddr>>,
    T: Param<Remote<ClientAddr>>,
```

linkerd/app/inbound/src/direct.rs
```diff
@@ -170,22 +171,27 @@ impl<T> Inbound<T> {
...
+{
     type Error = RefusedNoIdentity;
 
-    fn try_from(
-        (tls, addrs): (tls::ConditionalServerTls, listen::Addrs),
-    ) -> Result<Self, Self::Error> {
+    fn try_from((tls, addrs): (tls::ConditionalServerTls, N)) -> Result<Self, Self::Error> {
```
olix0r, line 181:
```suggestion
    fn try_from((tls, addrs): (tls::ConditionalServerTls, T)) -> Result<Self, Self::Error> {
```

linkerd/proxy/transport/src/addrs.rs
```diff
@@ -15,6 +15,12 @@ pub struct ListenAddr(pub SocketAddr);
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub struct ServerAddr(pub SocketAddr);
 
+/// The target address of a connection. This may be the `SO_ORIGINAL_DST` target
+/// address, if present, or the server listener's address if there is no
+/// original destination address.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub struct TargetAddr(pub SocketAddr);
```
olix0r, line 22: I don't think we want this... I think we want:

1. `OrigDstAddr` represents the SO_ORIGINAL_DST.
2. `Local<ServerAddr>` represents the server's address.
3. Proxy stacks should. for now take a `Param<Option<OrigDstAddr>>` and return an error when the `OrigDstAddr` is not present -- i.e. we should never do proxy routing based on the `Local<ServerAddr>`.

olix0r (changes requested):

linkerd/app/inbound/src/direct.rs
hawkw: hahaha, i should probably not have used find-and-replace for this!

linkerd/proxy/transport/src/addrs.rs
hawkw: Okay, sounds good --- I thought this PR should preserve the current behavior, but we can change that! :+1:

linkerd/app/inbound/src/lib.rs
```diff
@@ -266,21 +267,28 @@ where
...
+                    .push_request_filter(TcpAccept::port_skipped)
+                    .check_new_service::<T, _>()
+                    .instrument(|_: &T| debug_span!("forward"))
                     .into_inner(),
             )
-            .check_new_service::<listen::Addrs, I>()
+            .check_new_service::<T, I>()
             .push_switch(
-                PreventLoop::from(server_port),
+                PreventLoop::from(server_port).to_switch(),
```
olix0r, line 277: Can we consolidate this switch to actually build a target type that holds the origdst? It looks like we have to handle this case with an error in at least 3 places... Not a blocker, though, as I understand followups may ease this...

I also find switches harder to understand with the impulse decoupled from the stack -- it'd probably be easier to follow by embedding the closure here directly...

olix0r (approved):

linkerd/app/inbound/src/lib.rs
hawkw: > Can we consolidate this switch to actually build a target type that holds the origdst? It looks like we have to handle this case with an error in at least 3 places... Not a blocker, though, as I understand followups may ease this...

the intent was to follow up with a branch that changes the whole stack to be based on `Param<OrigDstAddr>` rather than `Param<Option<OrigDstAddr>>`, so this is kind of an intermediate state; if you'd prefer to avoid that, we could try to land the whole change in one bigger branch when it's ready?

PR #956, opened 2021-03-22
Initial work for introducing openssl/boringssl as a TLS provider

arnarpall:
In order be able to support running linkerd in fips compliant mode we need replace rustls and ring with openssl
since *ring* is not fips validated/compliant.

* Added support for openssl
* Testing out Boringssl
* Compilation features flags (openssl-tls, rustls-tls, boring-tls)


linkerd/tls/src/imp/boring.rs
```diff
@@ -0,0 +1,243 @@
...
+    where
+        IO: AsyncRead + AsyncWrite + Unpin,
+    {
+        let conf = self.0.configure().unwrap();
+        match tokio_boring::connect(conf, domain.as_ref(), stream).await {
+            Ok(ss) => Ok(ss.into()),
+            Err(err) => {
+                let _ = err.ssl();
+                Err(Error::new(ErrorKind::Other, "Ble"))
+            }
```
arnarpall, line 39: This needs some cleaning up

linkerd/tls/src/imp/boring.rs
```diff
@@ -0,0 +1,243 @@
...
+    pub async fn accept<IO>(&self, stream: IO) -> Result<server::TlsStream<IO>>
+    where
+        IO: AsyncRead + AsyncWrite + Unpin,
+    {
+        match tokio_boring::accept(&self.0, stream).await {
+            Ok(ss) => Ok(ss.into()),
+            Err(err) => {
+                let _ = err.ssl();
+                Err(Error::new(ErrorKind::Other, "Ble"))
+            }
```
arnarpall, line 110: Same as above, error handling cleanup.

linkerd/tls/src/imp/boring.rs
arnarpall: Error handling in the boring crate is being revamped here: https://github.com/cloudflare/boring/pull/23

nox: I'll revisit that PR at a later time because I have some other stuff on my plate and I misunderstood some critical parts of AsyncRead hah.

Dockerfile
```diff
@@ -36,18 +36,26 @@ ARG PROXY_FEATURES
 
 RUN --mount=type=cache,target=/var/lib/apt/lists \
   --mount=type=cache,target=/var/tmp \
-  apt update && apt install -y time cmake
+  apt update && \
+  apt install -y \
+    time \
+    cmake \
+    golang \
```
olix0r, line 43: Do we really need golang here?

.gitignore
```diff
@@ -2,3 +2,6 @@ target
 **/target
 **/corpus
 **/artifacts
+
+.idea/
+.vscode/
```
olix0r, line 7: generally prefer to keep these in your personal ~/.gitignore. Not a huge deal, but in interest of keeping the PR small and focused...
```suggestion
```

Cargo.lock
```diff
@@ -195,6 +294,19 @@ dependencies = [
  "syn",
 ]
 
+[[package]]
+name = "env_logger"
+version = "0.8.3"
```
olix0r, line 299: We shouldn't nee to pull in env-logger... Is this coming from one of the new dependencies?

env-logger should only be needed by binary entry points and not by libraries. And for libraries, the `tracing` crate is definitely preferable.

Cargo.lock
```diff
@@ -21,6 +21,15 @@ dependencies = [
  "memchr",
 ]
 
+[[package]]
+name = "ansi_term"
+version = "0.11.0"
```
olix0r, line 26: Similarly, this pulls in a second (older) version of `ansi_term`. It would be best if we could avoid these duplicate deps.

Cargo.lock
```diff
@@ -119,12 +183,47 @@ version = "1.0.66"
...
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "853eda514c284c2287f4bf20ae614f8781f40a81d32ecda6e91449304dfe077c"
+dependencies = [
+ "glob",
+ "libc",
+ "libloading",
+]
+
+[[package]]
+name = "clap"
```
olix0r, line 213: This is a command-line parsing crate. Why is this getting pulled in?

Cargo.lock
```diff
@@ -89,12 +109,56 @@ version = "0.13.0"
...
+ "lazy_static",
+ "libc",
+]
+
+[[package]]
+name = "boring-sys"
+version = "1.1.1"
+source = "git+https://github.com/netapp/boring?branch=feat/fips-support#85d02f09741053f7af55a1074d5c272d0a79b27c"
+dependencies = [
+ "bindgen",
```
olix0r, line 158: It looks like boring-sys should use bindgen without default features (https://github.com/rust-lang/rust-bindgen/blob/b60339ece3994868a55889b7f1f6043ab29ba30e/Cargo.toml#L74-L80) to avoid pulling in unnecessary dependencies.

olix0r (changes requested):
I'll continue looking at the feature flagging scheme to see what changes we can make to our APIs to support this better. In the meantime it would be great to minimize the dependencies for boring-sys.

olix0r:
Also note that we'll need some DCO sign-off for this change. See https://github.com/linkerd/linkerd2/blob/main/CONTRIBUTING.md#developer-certificate-of-origin. Note that it's not necessary to squash/amend all commits with the sign-off if long as you make a public statement in the PR.

Cargo.lock
olix0r: Specifically, pulling in `log` is okay; but pulling in `env_logger` is undesirable. It seems, though, that bindgen doesn't expose distinct dependencies.

Is bindgen actually a runtime dependency or is it only a build-time dependency? If the latter, it should probably be in `dev-dependencies` and not `dependencies`.

Dockerfile
arnarpall: Ah yeah that's crazy, the boring-sys crate uses go as part of the whole build process.

We probably have to figure out a better way to handle this docker file.

.gitignore
arnarpall: Right good point.

arnarpall:
I agree to the DCO for all the commits in this PR.

olix0r:
It looks like the `linkerd/identity` crate's dependencies need some fixing:

```
:; (cd linkerd/identity && cargo check)
...
error: aborting due to 25 previous errors

Some errors have detailed explanations: E0412, E0433.
For more information about an error, try `rustc --explain E0412`.
error: could not compile `linkerd-identity`
```

I suspect related to feature-flagging and conditional compilation.

Cargo.lock
olix0r: Ah yeah, I see now that it is indeed a build-only dependency in boring-sys. It's probably less of an issue if these aren't actually pulled into the runtime, then.

linkerd/identity/src/imp/openssl.rs
```diff
@@ -0,0 +1,345 @@
...
+
+impl error::Error for InvalidCrt {
+    fn source(&self) -> Option<&(dyn error::Error + 'static)> {
+        match self {
+            InvalidCrt::Verify(err) => err.source(),
+            InvalidCrt::General(err) => err.source(),
+            _ => None,
+        }
+    }
+}
```
hawkw, line 180: nit: again, I think a lot of this boilerplate could be generated using `thiserror`, like so
```suggestion
#[derive(Clone, Debug. Error)]
pub enum InvalidCrt {
    #[error("subject alt name incorrect ({0})"]
    SubjectAltName(String),
    #[error("{0}")]
    Verify(#[source] X509VerifyResult),
    #[error(transparent)]
    General(#[from] Error),
}

impl From<ErrorStack> for InvalidCrt {
    fn from(err: ErrorStack) -> Self {
        InvalidCrt::General(err.into())
    }
}
```

linkerd/identity/src/imp/openssl.rs
```diff
@@ -0,0 +1,345 @@
...
+            .finish()
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct Crt {
+    pub(crate) id: LocalId,
+    expiry: SystemTime,
+    pub cert: X509,
+    pub chain: Vec<X509>,
```
hawkw, line 227: again, if the intention is to abstract over all the openssl-specific implementation details, we should ideally find a way to make these not be `pub`...

linkerd/identity/src/imp/openssl.rs
```diff
@@ -0,0 +1,345 @@
...
+        }
+    }
+    pub fn empty() -> Self {
+        ClientConfig::new(
+            Vec::new(),
+            Arc::new(X509StoreBuilder::new().unwrap().build()),
+            None,
+            None,
+        )
+    }
```
hawkw, line 293: nit, take it or leave it: it might be slightly more idiomatic if this was an implementation of the `Default` trait? but, `empty()` is also fine...

linkerd/identity/src/imp/openssl.rs
```diff
@@ -0,0 +1,345 @@
...
+    pub root_certs: Arc<X509Store>,
+    pub key: Option<Arc<Key>>,
+    pub cert: Option<Arc<Crt>>,
+    pub protocols: Arc<Vec<Vec<u8>>>,
+}
+
+impl fmt::Debug for ClientConfig {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        fmt::Debug::fmt(&self.protocols, f)
+    }
```
hawkw, line 269: it might be a little more idiomatic to make this
```suggestion
    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
        f.debug_struct("ClientConfig").field("protocols", &self.protocols).finish()
    }
```
so that it's clearer which type this is?

linkerd/identity/src/imp/openssl.rs
```diff
@@ -0,0 +1,345 @@
...
+}
+
+impl fmt::Debug for ServerConfig {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "alpn_protocols: {:?}, key: {:?}",
+            self.alpn_protocols, self.key
+        )
```
hawkw, line 343: nit: prefer
```suggestion

        f.debug_struct("ServerConfig")
            .field("alpn_protocols", &self.alpn_protocols)
            .field("key", &self.key)
            .finish()
```

linkerd/identity/src/imp/rustls.rs
```diff
@@ -0,0 +1,415 @@
...
+
+    pub fn client_config(&self) -> Arc<ClientConfig> {
+        self.0.clone()
+    }
+}
+
+impl fmt::Debug for TrustAnchors {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("TrustAnchors")
+            .field(self.0.as_ref())
```
hawkw, line 157: currently, the rustls version of this type does not include the inner value:
https://github.com/linkerd/linkerd2-proxy/blob/c188444e805787a6e271782bba2442f2a8ed0387/linkerd/identity/src/lib.rs#L306-L310

what's the rationale behind changing this code? do we need to print this for debugging?

linkerd/identity/src/imp/rustls.rs
```diff
@@ -0,0 +1,415 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl error::Error for InvalidCrt {
+    fn source(&self) -> Option<&(dyn error::Error + 'static)> {
+        self.0.source()
+    }
+}
```
hawkw, line 175: nit: could use `thiserror` here:
```suggestion
#[derive(Clone, Debug, Error)]
#[error(transparent)]
pub struct InvalidCrt(#[from] rustls::TLSError);
```

linkerd/identity/src/imp/rustls.rs
```diff
@@ -0,0 +1,415 @@
...
+
+impl AsRef<rustls::ClientConfig> for ClientConfig {
+    fn as_ref(&self) -> &rustls::ClientConfig {
+        &self.0
+    }
+}
+
+impl fmt::Debug for ClientConfig {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("")
```
hawkw, line 368: i think this should be
```suggestion
        f.debug_struct("ClientConfig")
```

linkerd/identity/src/imp/rustls.rs
```diff
@@ -0,0 +1,415 @@
...
+
+impl AsRef<rustls::ServerConfig> for ServerConfig {
+    fn as_ref(&self) -> &rustls::ServerConfig {
+        &self.0
+    }
+}
+
+impl fmt::Debug for ServerConfig {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("")
```
hawkw, line 410: and this should be 
```suggestion
        f.debug_struct("ServerConfig")
```

linkerd/identity/src/lib.rs
```diff
@@ -18,59 +24,56 @@ pub use linkerd_dns_name::InvalidName;
...
+    }
+}
 
-#[derive(Clone, Debug)]
-pub struct TokenSource(Arc<String>);
+impl From<imp::Error> for Error {
+    fn from(err: imp::Error) -> Error {
+        Error(err)
+    }
+}
```
hawkw, line 52: these impls could also be generated using `thiserror`

```rust
/// An error returned from the TLS implementation.
#[derive(Error)]
#[error(transparent)]
pub struct Error(#[from] imp::Error);
```

linkerd/identity/src/lib.rs
```diff
@@ -18,59 +24,56 @@ pub use linkerd_dns_name::InvalidName;
...
 #[derive(Clone, Debug, Error)]
 #[error(transparent)]
-pub struct InvalidCrt(rustls::TLSError);
+pub struct InvalidCrt(imp::InvalidCrt);
+
+impl From<imp::InvalidCrt> for InvalidCrt {
+    fn from(err: imp::InvalidCrt) -> Self {
+        InvalidCrt(err)
+    }
+}
```
hawkw, line 71: nit: the `From` impl could be derived by `thiserror`:
```suggestion
pub struct InvalidCrt(#[from] imp::InvalidCrt);

impl From<imp::InvalidCrt> for InvalidCrt {
    fn from(err: imp::InvalidCrt) -> Self {
        InvalidCrt(err)
    }
}
```

linkerd/identity/src/lib.rs
```diff
@@ -318,139 +229,120 @@ impl Crt {
...
+//     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+//         self.0.fmt(f)
+//     }
+// }
+//
+// impl stdErr::Error for InvalidCrt {
+//     fn source(&self) -> Option<&(dyn stdErr::Error + 'static)> {
+//         self.0.source()
+//     }
+// }
```
hawkw, line 310: let's remove the commented-out code?

linkerd/identity/src/lib.rs
```diff
@@ -318,139 +229,120 @@ impl Crt {
...
+
+#[derive(Clone)]
+pub struct ClientConfig(pub imp::ClientConfig);
 
-        self.resolve_(hello.sigschemes())
+impl ClientConfig {
+    pub fn empty() -> Self {
+        Self(imp::ClientConfig::empty())
+    }
+    pub fn set_protocols(&mut self, protocols: Vec<Vec<u8>>) {
```
hawkw, line 319: nit: prefer whitespace here
```suggestion
    }
    
    pub fn set_protocols(&mut self, protocols: Vec<Vec<u8>>) {
```

linkerd/identity/src/imp/openssl.rs
```diff
@@ -0,0 +1,345 @@
...
+}
+
+impl fmt::Debug for ClientConfig {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        fmt::Debug::fmt(&self.protocols, f)
+    }
+}
+
+impl ClientConfig {
+    pub fn new(
```
hawkw, line 273: it looks like the `new` constructor is never called outside this module, can we just change this to
```suggestion
    fn new(
```
since this constructor exposes an implementation detail (the `X509Store` type from `openssl`).

linkerd/identity/src/imp/openssl.rs
```diff
@@ -0,0 +1,345 @@
...
+        self.id.as_ref()
+    }
+}
+
+#[derive(Clone)]
+pub struct ClientConfig {
+    pub root_certs: Arc<X509Store>,
+    pub key: Option<Arc<Key>>,
+    pub cert: Option<Arc<Crt>>,
+    pub protocols: Arc<Vec<Vec<u8>>>,
```
hawkw, line 263: these are never accessed outside this module, they shouldn't need to be `pub`:
```suggestion
    root_certs: Arc<X509Store>,
    key: Option<Arc<Key>>,
    cert: Option<Arc<Crt>>,
    protocols: Arc<Vec<Vec<u8>>>,
```

linkerd/identity/src/imp/openssl.rs
```diff
@@ -0,0 +1,345 @@
...
+#[derive(Clone)]
+pub struct ServerConfig {
+    pub root_certs: Arc<X509Store>,
+    pub key: Option<Arc<Key>>,
+    pub cert: Option<Arc<Crt>>,
+    pub alpn_protocols: Arc<Vec<Vec<u8>>>,
+}
+
+impl ServerConfig {
+    pub fn new(
```
hawkw, line 309: this constructor isn't called outside this module (afaict), it shouldn't need to be `pub`:
```suggestion
    fn new(
```

linkerd/identity/src/imp/openssl.rs
```diff
@@ -0,0 +1,345 @@
...
+        self.protocols = Arc::new(protocols)
+    }
+}
+
+#[derive(Clone)]
+pub struct ServerConfig {
+    pub root_certs: Arc<X509Store>,
+    pub key: Option<Arc<Key>>,
+    pub cert: Option<Arc<Crt>>,
+    pub alpn_protocols: Arc<Vec<Vec<u8>>>,
```
hawkw, line 305: these shouldn't need `pub`
```suggestion
    root_certs: Arc<X509Store>,
    key: Option<Arc<Key>>,
    cert: Option<Arc<Crt>>,
    alpn_protocols: Arc<Vec<Vec<u8>>>,
```

linkerd/tls/src/imp/openssl.rs
```diff
@@ -0,0 +1,295 @@
...
+    pub async fn connect<IO>(&self, domain: Name, stream: IO) -> Result<client::TlsStream<IO>>
+    where
+        IO: AsyncRead + AsyncWrite + Unpin,
+    {
+        let ssl = self
+            .0
+            .configure()
+            .unwrap()
+            .into_ssl(domain.as_ref())
+            .unwrap();
```
hawkw, line 49: in what cases will these `unwrap`s fail? if they should never fail in normal operation, these should probably be `expect`s with messages explaining why; otherwise, we should return an error in this case.

linkerd/tls/src/imp/openssl.rs
```diff
@@ -0,0 +1,295 @@
...
+    pub async fn connect<IO>(&self, domain: Name, stream: IO) -> Result<client::TlsStream<IO>>
+    where
+        IO: AsyncRead + AsyncWrite + Unpin,
+    {
+        let conf = self.0.configure().unwrap();
+        match tokio_boring::connect(conf, domain.as_ref(), stream).await {
+            Ok(ss) => Ok(ss.into()),
+            Err(err) => {
+                let _ = err.ssl();
+                Err(io::Error::new(ErrorKind::Other, "Ble"))
```
hawkw, line 66: what is this code doing? what does `err.ssl()` do, and why do we ignore the return value?

linkerd/tls/src/imp/openssl.rs
```diff
@@ -0,0 +1,295 @@
...
+
+impl From<SslConnectorBuilder> for TlsConnector {
+    fn from(builder: SslConnectorBuilder) -> Self {
+        builder.build().into()
+    }
+}
+
+impl From<Arc<ClientConfig>> for TlsConnector {
+    fn from(conf: Arc<ClientConfig>) -> Self {
+        println!("Setting up ssl connector");
```
hawkw, line 86: let's either remove the `println!` or replace it with a `debug!` or something, please :)

linkerd/tls/src/imp/openssl.rs
```diff
@@ -0,0 +1,295 @@
...
+        }
+
+        match conf.0.cert.clone() {
+            None => debug!("No certificate provided"),
+            Some(cert) => {
+                println!("Setting certificate {:?}", cert);
+                builder.set_certificate(cert.cert.as_ref()).unwrap();
+
+                for cert in &cert.chain {
+                    println!("Adding extra chain certificate {:?}", cert);
```
hawkw, line 104: we should probably remove the `println!`s here as well (or replace them with logging)

linkerd/tls/src/imp/openssl.rs
```diff
@@ -0,0 +1,295 @@
...
+        }
+
+        builder.set_cert_store(X509StoreBuilder::new().unwrap().build());
+        conf.0
+            .root_certs
+            .objects()
+            .iter()
+            .map(|x| x.x509().unwrap().to_owned())
+            .for_each(|cert| {
+                println!("Adding Root certificate {:?}", cert);
```
hawkw, line 117: ...and here

linkerd/tls/src/imp/openssl.rs
```diff
@@ -0,0 +1,295 @@
...
+        let key = conf.0.key.as_ref().unwrap().as_ref().0.as_ref();
+        println!("Setting private key {:?}", key);
+        builder.set_private_key(key).unwrap();
+
+        let cert = conf.0.cert.as_ref().unwrap().as_ref();
+        println!("Setting certificate {:?}", cert);
+        builder.set_certificate(cert.cert.as_ref()).unwrap();
+
+        for cert in &cert.chain {
+            println!("Adding extra chain certificate {:?}", cert);
```
hawkw, line 183: let's remove the `println!`s or replace them with `debug!` or `trace!` logging?

linkerd/tls/src/imp/openssl.rs
```diff
@@ -0,0 +1,295 @@
...
+
+impl From<SslAcceptor> for TlsAcceptor {
+    fn from(acceptor: SslAcceptor) -> Self {
+        Self(acceptor)
+    }
+}
+
+impl From<Arc<ServerConfig>> for TlsAcceptor {
+    fn from(conf: Arc<ServerConfig>) -> Self {
+        println!("SSL Version {}", version::version());
```
hawkw, line 165: let's remove the `println!` or replace it with logging here?

linkerd/tls/src/imp/openssl.rs
```diff
@@ -0,0 +1,295 @@
...
+        }
+
+        builder.set_cert_store(X509StoreBuilder::new().unwrap().build());
+        conf.0
+            .root_certs
+            .objects()
+            .iter()
+            .map(|x| x.x509().unwrap().to_owned())
+            .for_each(|cert| {
+                println!("Adding Root certificate {:?}", cert);
```
hawkw, line 194: another `println!`

linkerd/tls/src/imp/openssl.rs
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    pub fn client_identity(&self) -> Option<ClientId> {
+        let cert = self.0.ssl().peer_certificate()?;
+        let peer = cert.subject_alt_names()?.pop()?;
+
+        match ClientId::from_str(peer.dnsname().unwrap()) {
+            Ok(client) => Some(client),
+            Err(_) => None,
+        }
```
hawkw, line 229: nit: this `match` could be simplified to
```suggestion
        ClientId::from_str(peer.dnsname().unwrap()).ok()
```

hawkw (commented):
i gave this a quick review focusing mostly on style and idiomatic Rust --- I'd like to look closer at the feature-flagging strategy here in the future. the overall strategy --- incapsulating `rustls`/`openssl`/`boringssl`-specific implementation details in private modules that are wrapped by backend-agnostic types to expose them to the rest of the codebase --- definitely looks correct!

some general notes:

- i noticed there's a very large number of `unwrap()`s in this PR. `unwrap` will cause the process to panic (crash instantly). 

  in general, we should try to avoid using `unwrap()`, even for errors that are unrecoverable and _should_ cause the proxy to panic --- we should prefer `expect` for fatal errors, because `expect` includes a message. in general, we should try to include messages for these errors explaining why the error is fatal and/or why we expect it to not occur in normal operation. 

- also, i noticed that there are a lot of `println!`s, presumably for locally debugging while working on this change. before we merge this, we should try to remove those, or replace them with logging if that information is valuable to record in production.

- some error-handling boilerplate could probably be simplified a lot by using the [`?` operator](https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/the-question-mark-operator-for-easier-error-handling.html) in more places

linkerd/identity/Cargo.toml
```diff
@@ -8,13 +8,17 @@ edition = "2018"
...
 test-util = []
+rustls-tls = ["ring", "rustls"]
+openssl-tls = ["openssl"]
+boring-tls = ["boring"]
 
 [dependencies]
 linkerd-dns-name = { path = "../dns/name" }
-ring = "0.16.19"
-rustls = "0.19"
+ring = { version = "0.16.19", optional = true, features = ["std"] }
```
arnarpall, line 17: I added the `features = ["std"]` in order to satisfy `thiserror::Error` not sure if we want this around but I'll circle back to it.

linkerd/identity/src/imp/openssl.rs
```diff
@@ -0,0 +1,345 @@
...
+        }
+
+        Some(Self(Arc::new(store.build())))
+    }
+
+    pub fn certify(&self, key: Key, crt: Crt) -> Result<CrtKey, InvalidCrt> {
+        let cert = crt.cert.clone();
+        if cert
+            .subject_alt_names()
+            .unwrap()
```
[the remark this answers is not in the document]
arnarpall: Yeah this was not really intentional but rather just inexperience. I've applied your suggestions 👍

linkerd/identity/Cargo.toml
hawkw: this is fine; linkerd is always built with the Rust standard library, so enabling `ring`'s dependency on `std` is fine.

arnarpall:
I just want to thank you for your review @hawkw it's immensely helpful.

I've been working my way through your comments and hopefully I should be able to get something together today.

* Usages of `println` have been chanaged to  either `debug` or `trace`. This was really just an oversight on my part.
* `Error` trait boilerplate removed and will now be generated by `thiserror:Error`

linkerd/identity/src/imp/openssl.rs
```diff
@@ -0,0 +1,345 @@
...
+                    Ok(_) => Ok(CrtKey {
+                        id: crt.id.clone(),
+                        expiry: crt.expiry.clone(),
+                        client_config: Arc::new(client_config),
+                        server_config: Arc::new(server_config),
+                    }),
+                    Err(err) => Err(err),
+                }
+            }
+            Err(err) => Err(err.into()),
```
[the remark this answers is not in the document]
arnarpall: Ah yeah thanks for this. I really really disliked this block of code when I wrote it.

linkerd/identity/src/imp/openssl.rs
arnarpall: Yeah I'm going to circle back to this.
The main thing here is the coupling between the Identity and the Tls modules respectively.
Ideally these should be completely hidden from the consumer. I'll figure something out in this regard :D

hawkw:
> I just want to thank you for your review @hawkw it's immensely helpful.

glad to help --- let me know (or click the "re-request review" button) whenever it's ready for another look?

linkerd/identity/src/imp/openssl.rs
```diff
@@ -0,0 +1,345 @@
...
+    },
+};
+
+use tracing::{debug, warn};
+
+use crate::{LocalId, Name};
+use std::fmt::Formatter;
+
+#[derive(Clone, Debug)]
+pub struct Key(pub Arc<PKey<Private>>);
```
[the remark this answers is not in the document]
arnarpall: Yeah I totally agree.
I'm looking into what is the best solution here. The reason why this is public at this time is because of the identity vs tls modules.

When constructing a TlsConnector I need access to the primitive type Pkey<Private> from within the tls module.
I'm pretty sure this is something that can be solved by implementing `From` and `Into` traits but I'm not seeing the solution at this time.

linkerd/identity/src/imp/openssl.rs
arnarpall: Absolutely.

linkerd/identity/src/imp/rustls.rs
arnarpall: Very good point. I added this when I was debugging something myself and forgot to remove it.

linkerd/tls/src/imp/openssl.rs
arnarpall: I'll change this tomorrow morning. I was battling a little bit with the returned error from tokio_boring

arnarpall:
@hawkw @olix0r  could you guys give the pull request another look over ?

I believe I've fixed most of your concerns @hawkw aside from the public fields in a few places. It would be awesome if we could perhaps try to sync up at some point and figure out a good way to hide these fields.

I have not fixed the build/dependency issues that you mentioned @olix0r I will circle back to that.

olix0r:
Last week I spent a few days trying to rework the TLS and identity modules. That work is incomplete but shared [here](https://github.com/linkerd/linkerd2-proxy/pull/new/ver/tls-generis). The summary is;

* The TLS module is generic over the actual TLS implementation (which holds the config types -- these should not be exposed generally).
* The identity module holds only the core structs & traits needed to implement identity. These traits model clients and servers for TLS initiation and termination.
* the proxy-identity module communicates with the API but delegates validation to the underlying identity implementation.
* at startup, the proxy is configured with an identity mode that determines which implementation is used from the implementations compiled with the proxy. This lets us play well with the Rust feature system so that features are purely additive and do not conflict.

I had to put this down to focus on inbound policy changes, but I should be able to revive this work once our access policy work lands.

olix0r:
Closing this out in favor of #1351, which takes advantage of recent API changes to support compilation with `--all-features` (i.e. `rustls` and `boring` are not mutually exclusive).

There are some issues we'll need to resolve before merging that. Notably: ALPN support & trust anchor parsing

PR #957, opened 2021-03-25
transport: introduce Bind trait and move orig dst to the type level

hawkw:
This branch introduces a new `transport::listen::Bind` trait that
abstracts over binding and listening on a socket. Now, rather than
owning an instance of  the `BindTcp` type, `ServerConfig`s are passed as
a parameter to a type implementing the `Bind` trait. This will
eventually allow end-to-end testing without creating actual TCP sockets,
as `Bind` produces a `Stream` of an IO type and an `Addrs` type.

PR #955 changed the proxy stacks to be a target type which implements
the `Param` trait for various types of address, rather than a fixed
type. This PR also changes the inbound and outbound proxies to take
targets which implement `Param<OrigDstAddr>`, rather than requiring
`Param<Option<OrigDstAddr>>` and punting fallibility to every point
where original dst addresses are used.

The `BindTcp` type is now generic over a more general `GetAddrs` trait,
rather than a `SO_ORIGINAL_DST`-specific `OrigDstAddr` trait. Rather
than always producing a `listen::Addrs` with an `Option<OrigDstAddr>`,
we now produce a variable addresses type, which may or may not incldue
the original destination address (the admin and tap servers don't
require SO_ORIGINAL_DST). This allows whether or not original
destination addresses are included to be determined at compile time
rather than at runtime.

Finally, the incorrect feature flagging of mock original destination
addresses has been fixed, The "mock-orig-dst" feature has been removed,
and the integration tests now simply construct their own mock `GetAddrs`
function that's used to configure the test proxy. In the future, we
should be able to refactor the tests to avoid this by simply passing in
their own mock `Bind` types which produce streams of in-memory mock
connections, rather than actual TCP connections.


olix0r (commented):
I need to give this a deeper read through, but my first impression is that we really should not need a `GetAddrs` trait to be exposed to all of the stacks... I.e., we should keep the same general signature for the stacks that they are constructed with a parameterized target and we shouldn't have to have all of the address filtering business in every stack.

linkerd/app/inbound/src/lib.rs
```diff
@@ -128,11 +146,16 @@ impl Inbound<()> {
         gateway: G,
     ) -> (SocketAddr, impl Future<Output = ()> + Send)
     where
+        B: Bind + Clone + Send + Sync + 'static,
+        B::Addrs: Param<Remote<ClientAddr>>
+            + Param<Local<ServerAddr>>
+            + Param<OrigDstAddr>
+            + Clone
+            + Send,
+        B::Io: io::Peek + io::PeerAddr + Debug + Unpin,
```
hawkw, line 155: we could avoid having to write these bounds out in as many places by just making them into bounds on the definition of the `Bind` trait, since it turns out everything requires these traits as well. I wasn't sure if it was worth making the trait definition more restrictive or not...

hawkw: also, we *could* try to figure out a way to get rid of the `PeerAddr` trait and just get client addresses from the target always. IDK if this matters or is worth doing...

olix0r:
What does the `mock-orig-dst` feature do now? Is it necessary at all?

hawkw:
> What does the `mock-orig-dst` feature do now? Is it necessary at all?

It just controls whether or not the `MockOrigDstAddr` type exists. Since it's only used in tests, I thought it was better to have that type only be enabled by the test crates when building tests, and not in the app build.

olix0r:
> It just controls whether or not the MockOrigDstAddr type exists. Since it's only used in tests, I thought it was better to have that type only be enabled by the test crates when building tests, and not in the app build.

Can the type live in the test crate?

hawkw:
> > It just controls whether or not the MockOrigDstAddr type exists. Since it's only used in tests, I thought it was better to have that type only be enabled by the test crates when building tests, and not in the app build.
> 
> Can the type live in the test crate?

...yes, now i feel a bit silly for not thinking of that!

olix0r:
@hawkw I tried hacking together https://github.com/linkerd/linkerd2-proxy/pull/962 on top of this to remove the orig-dst constraints from the config types, which hopefully simplifies things a bit.

Dockerfile
```diff
@@ -31,9 +31,7 @@ FROM $RUST_IMAGE as build
 # When set, causes the proxy to be compiled in development mode.
 ARG PROXY_UNOPTIMIZED
 
-# Controls what features are enabled in the proxy. This is typically empty but
-# may be set to `mock-orig-dst` for profiling builds, or to `multithreaded` to
-# enable the multithreaded Tokio runtime.
```
hawkw, line 36: I removed this part of the comment since the `mock-orig-dst` feature is now removed and the `multithreaded` feature is enabled by default.

hawkw:
> > > It just controls whether or not the MockOrigDstAddr type exists. Since it's only used in tests, I thought it was better to have that type only be enabled by the test crates when building tests, and not in the app build.
> > 
> > 
> > Can the type live in the test crate?
> 
> ...yes, now i feel a bit silly for not thinking of that!

@olix0r removed the feature flag in 97b95906c4af81290b90aa623dfffe9330eff6d8, good call!

linkerd/app/inbound/src/lib.rs
olix0r: we need peeraddr for now at least, as the http server marks extensions with the client addr, but we can't have the http server created per-client (since we may need to do cached resolutions before this) -- so we'd either need to rework caching or have an Io type that supports extensions...

hawkw: yeah, i took a look at that on Monday and i think that we're better off keeping it, and if we are going to get rid of it, that's out of scope for this PR.

hawkw: @olix0r 7b16586 moves all the bounds on the `Bind::Io` type to the `Bind` trait so they're not repeated everywhere else (and removes some of the `Bind::Addrs` bounds that the `Bind` trait already requires).

We _could_ also put the `Param<Local<ServerAddr>>` and `Param<Remote<ClientAddr>>` bounds on `Bind::Addrs` so they aren't repeated, but it felt a _little_ weird to see it require only `Param<OrigDstAddr>` at use-sites; i could go either way on this...

linkerd/proxy/transport/src/orig_dst.rs
```diff
@@ -0,0 +1,191 @@
...
+        let (addr, incoming) = self.inner.bind(t)?;
+
+        let incoming = incoming.map(|res| {
+            let (inner, tcp) = res?;
+            let orig_dst = orig_dst_addr(&tcp)?;
+            let addrs = Addrs { inner, orig_dst };
+            Ok((addrs, tcp))
+        });
+
+        Ok((addr, Box::pin(incoming)))
```
olix0r, line 75: btw, I tried to get rid of the box, but this closure captures `B` (types -- not values) so we get a _expected fn pointer found closure_ compiler error. if you know of a way to avoid the box, by all means...

PR #958, opened 2021-03-26
tls: Fix inbound I/O when TLS detection fails

olix0r:
When the proxy performs inbound TLS detection, it may break TLS streams
for application-terminated TLS connections.

Typically, inbound TLS detection does not actually need to buffer data
from the socket to determine the SNI value for a connection. However,
non-proxy clients may send enough ClientHello extensions such that
detection cannot be completed from a single peek of 512B. In this
situation, the proxy buffers data from the socket to determine the
SNI value, but it unfortunately does not preserve this buffered data as
it forward the connection.

This change updates the TLS detection logic to properly preserve any
data buffered from the socket in forwarded connections.

### DIFF
--- a/linkerd/tls/src/server/mod.rs
+++ b/linkerd/tls/src/server/mod.rs
@@ -247,12 +247,14 @@ where
             Ok(Some(sni)) => {
                 trace!(%sni, "Identified non-matching SNI via peek");
                 let tls = Conditional::Some(ServerTls::Passthru { sni });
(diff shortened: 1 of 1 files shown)


hawkw (approved):
lgtm! nice catch!

olix0r:
Trying to write some tests for this... It should be easier with some larger changes to the TLS detect logic, so I'll target those in a followup.

PR #959, opened 2021-03-27
 tls: Test SNI detection

olix0r:
7d2bdbb fixed an issue with TLS detection. This change adds a test for
this behavior.

In order to implement a test, this change modifies the `detect` helper
to not be responsible for SNI matching & TLS termination. The `detect`
helper is replaced with a `detect_sni` function that only reads the
ClientHello without terminating TLS. A test has been added for the
`detect_sni` helper.


linkerd/tls/src/server/mod.rs
```diff
@@ -159,17 +160,38 @@ where
...
                 Box::pin(async move {
-                    let (peer, io) = tokio::select! {
-                        res = detect(io, config, local_id) => { res? }
+                    // Detect the SNI from a ClientHello (or timeout).
+                    let (sni, io) = tokio::select! {
+                        res = detect_sni(io) => { res? }
                         () = timeout => {
                             return Err(DetectTimeout(()).into());
                         }
                     };
```
hawkw, line 174: nit/tioli (and i realize this code was already there): i wonder if this is expressed more concisely with [`tokio::time::timeout`](https://docs.rs/tokio/1.4.0/tokio/time/fn.timeout.html)? but, maybe we would still have to unpack a nested `Result` so it might be the same number of lines...

linkerd/tls/src/server/mod.rs
```diff
@@ -181,22 +203,20 @@ where
...
         }
     }
 }
 
-async fn detect<I>(
-    mut io: I,
-    tls_config: Config,
-    LocalId(local_id): LocalId,
-) -> io::Result<(ConditionalServerTls, Io<I>)>
+// Peek or buffer the provided stream to determine an SNI value.
```
hawkw, line 215: *very* nitpicky (sorry, this is dumb): 
```suggestion
/// Peek or buffer the provided stream to determine an SNI value.
```

linkerd/tls/src/server/mod.rs
```diff
@@ -236,25 +245,8 @@ where
     while io.read_buf(&mut buf).await? != 0 {
         debug!(buf.len = %buf.len(), "Read bytes from TCP stream");
         match client_hello::parse_sni(buf.as_ref()) {
-            Ok(Some(ServerId(sni))) if sni == local_id => {
-                trace!(%sni, "Identified matching SNI via buffered read");
```
hawkw, line 240: nit, not important: looks like we no longer log the SNI (or whether it matches or not); is this still potentially worth having?

linkerd/tls/src/server/mod.rs
```diff
@@ -373,3 +365,38 @@ impl fmt::Display for NoServerTls {
...
+#[cfg(test)]
+mod tests {
+    use io::AsyncWriteExt;
+
+    use super::*;
+    use std::str::FromStr;
+
+    #[tokio::test]
+    async fn detect_buffered() {
+        let _ = tracing_subscriber::fmt::try_init();
```
hawkw, line 378: nit: for tests, we should prefer
```suggestion
        let _ = tracing_subscriber::fmt().with_test_writer().try_init();
```
which will enable libtest to capture the test's output and only print it when the test fails (perhaps this should just be the default behavior in tracing...).

linkerd/tls/src/server/mod.rs
```diff
@@ -373,3 +365,38 @@ impl fmt::Display for NoServerTls {
...
+
+        let (mut client_io, server_io) = tokio::io::duplex(1024);
+        let input = include_bytes!("testdata/curl-example-com-client-hello.bin");
+        let len = input.len();
+        tokio::spawn(async move {
+            client_io
+                .write_all(&*input)
+                .await
+                .expect("Write must suceed");
```
hawkw, line 387: if this *does* fail, the panic won't be propagated, since we `spawn` this, and never await the returned `JoinHandle`. i dunno if that's a big deal, since this should actually never fail, but still, may be worth awaiting the `JoinHandle` at the end of the test to be sure this didn't panic?

hawkw (approved):
overall, lgtm; I commented on a few nits

linkerd/tls/src/server/mod.rs
olix0r: I tried this and it didn't work. I didn't have time to dig into why...

PR #961, opened 2021-03-30
Initial fuzzer integration

DavidKorczynski:
Adds a set of fuzzers for various of the `linkerd` crates. 

The goal of this PR is to create an initial fuzzer infrastructure as well as fuzzers that can be integrated with OSS-Fuzz for continuous fuzzing. 

The way I set it up was to place the fuzzers for a given crate within the crate folder itself. I then put in a `fuzz_logic` module in the source code that is to be fuzzed, thus following the low of how the existing tests are written. The majority of additions in this PR are independent of the linkerd2-proxy source code, but there are a few changes.


Cargo.toml
```diff
@@ -63,4 +63,4 @@ debug = false
 debug = false
 
 [patch.crates-io]
-webpki = { git = "https://github.com/linkerd/webpki", branch = "cert-dns-names-0.21", rev = "b2c3bb3" }
+webpki = { git = "https://github.com/linkerd/webpki", branch = "cert-dns-names-0.21" }
```
olix0r, line 66: did the pinned revision cause problems?

linkerd/transport-header/fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -0,0 +1,8 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &[u8]| {
+    if let Ok(s) = std::str::from_utf8(data) {
+        linkerd_transport_header::fuzz_logic::fuzz_entry(s, data);
```
olix0r, line 6: We have a CI task that runs `cargo check` in every directory that has a `Cargo.toml` file and this seems to cause some issue:

```
error[E0433]: failed to resolve: could not find `fuzz_logic` in `linkerd_transport_header`
 --> fuzz_targets/fuzz_target_1.rs:6:35
  |
6 |         linkerd_transport_header::fuzz_logic::fuzz_entry(s, data);
  |                                   ^^^^^^^^^^ could not find `fuzz_logic` in `linkerd_transport_header`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `linkerd-transport-header-fuzz`

To learn more, run the command again with --verbose.
```

Perhaps the whole file needs a `#![cfg(fuzzing)]` at the top so it's only compiled in this mode? Either that, or we should instead use a feature-gate on the transport-header crate and have this fuzzing crate depend on that feature.

Cargo.toml
DavidKorczynski:
```linkerd2-proxy/linkerd/proxy/http$ cargo +nightly fuzz build
error: failed to parse manifest at `/linkerd2-proxy-6/linkerd2-proxy/Cargo.toml`

Caused by:
  dependency (webpki) specification is ambiguous. Only one of `branch`, `tag` or `rev` is allowed.
Error: failed to build fuzz script: "cargo" "build" "--manifest-path" "linkerd2-proxy/linkerd/proxy/http/fuzz/Cargo.toml" "--target" "x86_64-unknown-linux-gnu" "--release" "--bins"```

File: linkerd/transport-header/fuzz/fuzz_targets/fuzz_target_1.rs (hunk 2 above)
Comment (DavidKorczynski) [reply linkerd/transport-header/fuzz/fuzz_targets/fuzz_target_1.rs:6]: I solved this by putting a feature hurdle in the fuzzing `Cargo.toml`s

File: Cargo.toml (hunk 1 above)
Comment (hawkw) [reply Cargo.toml:66]: hmm, maybe we should prefer the pinned revision over the pinned branch, since the rev is more specific? but i don't have a strong opinion on this one.

Comment (DavidKorczynski) [reply Cargo.toml:66]: out of curiosity, is there a reason we don't use the latest of the branch?

Comment (hawkw) [reply Cargo.toml:66]: oh, huh, are we not? :)

File: linkerd/addr/fuzz/.gitignore [hunk 3]
```diff
@@ -0,0 +1,4 @@
+
+target
+corpus
+artifacts
```
hawkw, line 4: nit, take it or leave it: could we perhaps add something like
```gitignore
**/fuzz/target
**/fuzz/corpus
**/fuzz/artifacts
```
to the repository's root gitignore, instead? that way, if we add additional fuzzers in the future, we don't have duplicate this gitignore into all of them.

Not a huge deal, just a thought.

linkerd/addr/fuzz/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+
+[dependencies.linkerd-addr]
+path = ".."
+
+# Prevent this from interfering with workspaces
+[workspace]
+members = ["."]
+
+[features]
+fuzzing = []
```
hawkw, line 22: the fuzz target is a separate crate that's only compiled when fuzzing, correct? why do we also need a feature flag to enable it?

linkerd/proxy/http/src/detect.rs
```diff
@@ -147,3 +147,20 @@ mod tests {
...
+
+#[cfg(fuzzing)]
+pub mod fuzz_logic {
+    use super::*;
+
+    pub fn fuzz_entry(input: &[u8]) {
+        let a1 = async {
+            use tokio_test::io;
+
+            debug!(read = ?std::str::from_utf8(input).unwrap());
```
hawkw, line 159: is this ever actually logged within the fuzz test? I notice we don't set up a `tracing` subscriber or `log` logger, so as far as I can tell, this macro never does anything?

hawkw (commented):
I had some largely procedural questions about how the fuzz tests are organized.

linkerd/addr/fuzz/Cargo.toml
DavidKorczynski: The reason we do this is because otherwise `cargo check` fails due to the use of `cfg(fuzzing)` wrapped around the fuzz modules. See here: https://github.com/linkerd/linkerd2-proxy/pull/961#discussion_r605010415

linkerd/addr/fuzz/.gitignore
DavidKorczynski: When creating the fuzzers the initial set up for a given fuzz crate is created by way of `cargo fuzz init`, which creates the folder structure in each of the fuzzer crates. As such, although I agree that it would be annoying if we had to duplicate the files - in this case the files are auto-generated. Because they are auto-generated I probably lean towards keeping them so we won't have to remove auto-generated code when creating new fuzzers in the future.

linkerd/proxy/http/src/detect.rs
DavidKorczynski: You'r right! Fixed.

linkerd/addr/fuzz/.gitignore
hawkw: Ah, okay, I hadn't realized that these were automatically generated...that's cool! In that case, never mind!

It might be nice if `cargo fuzz init` included a comment indicating that the file was generated by `cargo fuzz init`...maybe I should open an issue upstream?

linkerd/addr/fuzz/Cargo.toml
hawkw: Ah, okay, I see, that makes sense. I think the use of the feature flag is probably appropriate in this case.

linkerd/addr/fuzz/.gitignore
DavidKorczynski: I think that's a great idea! I can do a PR in the near future on their repo with an implementation of it!

linkerd/addr/src/lib.rs
```diff
@@ -258,3 +258,22 @@ mod tests {
...
+#[cfg(fuzzing)]
+pub mod fuzz_logic {
+    use super::*;
+
+    pub fn fuzz_addr_1(fuzz_data: &str) {
+        if let Ok(addr) = Addr::from_str(fuzz_data) {
+            addr.is_loopback();
+            addr.to_http_authority();
+            addr.is_loopback();
+            addr.socket_addr();
```
hawkw, line 271: i'm not very familiar with `cargo fuzz` idioms and practices, so this could be a stupid question, but would it make sense to test each of these methods in a separate fuzz target so that it's clearer where failures occurred?

hawkw (commented):
Gave this a closer read and I had some additional suggestions. Let me know if any of these don't or you have follow-up questions?

linkerd/dns/fuzz/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+authors = ["Linkerd Developers <cncf-linkerd-dev@lists.cncf.io>"]
+publish = false
+edition = "2018"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = "0.4"
+tokio = { version = "1", features = ["rt", "sync", "time"] }
```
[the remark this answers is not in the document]
DavidKorczynski: I don't think Tokio has an "io" flag? Do you perhaps mean `io-util`?

linkerd/dns/src/lib.rs
```diff
@@ -246,3 +246,31 @@ mod tests {
...
+            let a1 = async {
+                let _w = _resolver.resolve_a(&_name).await;
+            };
+            futures::executor::block_on(a1);
+            let a2 = async {
+                let _w2 = _resolver.resolve_srv(&_name).await;
+            };
+            futures::executor::block_on(a2);
+        }
+    }
```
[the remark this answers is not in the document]
DavidKorczynski: I think you're right entirely - it also seems the fuzzer is running much better with your suggestions. Please check the new set up!

linkerd/dns/src/lib.rs
```diff
@@ -246,3 +246,31 @@ mod tests {
...
+            let _fcon = FuzzConfig {};
+            let _resolver = Resolver::from_system_config_with(&_fcon).unwrap();
+            let a1 = async {
+                let _w = _resolver.resolve_a(&_name).await;
+            };
+            futures::executor::block_on(a1);
+            let a2 = async {
+                let _w2 = _resolver.resolve_srv(&_name).await;
+            };
+            futures::executor::block_on(a2);
```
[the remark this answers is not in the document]
DavidKorczynski: Fixed, thanks!

linkerd/proxy/http/src/detect.rs
```diff
@@ -147,3 +147,19 @@ mod tests {
...
+
+    pub fn fuzz_entry(input: &[u8]) {
+        let a1 = async {
+            use tokio_test::io;
+
+            let mut buf = BytesMut::with_capacity(1024);
+            let mut io = io::Builder::new().read(&input).build();
+            let _kind = DetectHttp(()).detect(&mut io, &mut buf).await.unwrap();
+        };
+        futures::executor::block_on(a1);
```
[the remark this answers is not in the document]
DavidKorczynski: Fixed!

linkerd/proxy/http/fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -0,0 +1,10 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &[u8]| {
+    let rt = tokio::runtime::Runtime::new().unwrap();
+    rt.block_on(async {
+        linkerd_proxy_http::detect::fuzz_logic::fuzz_entry(data);
+    })
+});
```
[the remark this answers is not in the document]
DavidKorczynski: Fixed!

linkerd/transport-header/src/lib.rs
```diff
@@ -269,3 +269,31 @@ mod tests {
...
+                buf.put_slice(ran_data);
+                _header.encode_prefaced(&mut buf).expect("must encode");
+                buf.put_slice(ran_data);
+                buf.put_slice(ran_data);
+                std::io::Cursor::new(buf.freeze())
+            };
+            let mut buf = BytesMut::new();
+            let a1 = async {
+                let _h = TransportHeader::read_prefaced(&mut _rx, &mut buf).await;
+            };
```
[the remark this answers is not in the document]
DavidKorczynski: Fixed!

linkerd/transport-header/src/lib.rs
```diff
@@ -269,3 +269,31 @@ mod tests {
...
+            futures::executor::block_on(a1);
```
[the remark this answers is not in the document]
DavidKorczynski: I thought it would be nice to have consistency across the fuzzers so I set it up similarly to the other ones (by way of `pub async fn`).

linkerd/addr/src/lib.rs
DavidKorczynski: We will get nice stack traces out that shows us where the bug occurs - so it will be easy to debug issues in this way.

In general we don't want to have too small fuzzers as well - especially once we put them up on clusterfuzz (oss-fuzz) which will throw a bit of computing power on it. We want to make sure each of the fuzzers (or those that we enable for OSS-Fuzz) have a fair amount of coverage.

linkerd/transport-header/src/lib.rs
```diff
@@ -269,3 +269,31 @@ mod tests {
...
+
+#[cfg(fuzzing)]
+pub mod fuzz_logic {
+    use super::*;
+    pub fn fuzz_entry(fuzz_data: &str, ran_data: &[u8]) {
+        if let Ok(name) = Name::from_str(fuzz_data) {
+            let _header = TransportHeader {
+                port: 4040,
+                name: Some(name),
+                protocol: Some(SessionProtocol::Http2),
```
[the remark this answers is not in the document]
DavidKorczynski: I agree - this is one my todo list, namely improving the fuzzers to have randomised data in essentially all inputs. However, can we set this as a todo rather than fix in this PR?

linkerd/transport-header/src/lib.rs
```diff
@@ -269,3 +269,31 @@ mod tests {
...
 }
+
+#[cfg(fuzzing)]
+pub mod fuzz_logic {
+    use super::*;
+    pub fn fuzz_entry(fuzz_data: &str, ran_data: &[u8]) {
+        if let Ok(name) = Name::from_str(fuzz_data) {
+            let _header = TransportHeader {
+                port: 4040,
+                name: Some(name),
```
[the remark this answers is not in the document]
DavidKorczynski: Following the same as above, I agree. This is one my todo list, namely improving the fuzzers to have randomised data in essentially all inputs. However, can we set this as a todo rather than fix in this PR?

Cargo.toml
DavidKorczynski: https://github.com/linkerd/webpki/commits/cert-dns-names-0.21 - just a single commit behind which changes the version.

DavidKorczynski:
@hawkw please check my replies to your review.

linkerd/transport-header/fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -0,0 +1,22 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+use libfuzzer_sys::arbitrary::Arbitrary;
+
+#[derive(Debug, Arbitrary)]
+pub struct AllocatorMethod {
```
hawkw, line 6: I'm not sure if I understand the naming "AllocatorMethod"?

linkerd/transport-header/fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -0,0 +1,22 @@
...
+    pub data: Vec<u8>,
+    pub port: u16,
+    pub protocol: bool,
+}
```
hawkw, line 10: nit, not important: I don't think any of this needs to be `pub`?
```suggestion
struct AllocatorMethod {
    data: Vec<u8>,
    port: u16,
    protocol: bool,
}
```

linkerd/transport-header/src/lib.rs
```diff
@@ -269,3 +269,22 @@ mod tests {
...
+            name: Name::from_str(fuzz_name).ok(),
+            protocol: Some(fuzz_proto),
+        };
+        let mut rx = {
+            let mut buf = BytesMut::new();
+            header.encode_prefaced(&mut buf).expect("must encode");
+            std::io::Cursor::new(buf.freeze())
+        };
+        let mut buf = BytesMut::new();
+        let _h = TransportHeader::read_prefaced(&mut rx, &mut buf).await;
```
hawkw, line 288: I think I mentioned it before, but does it also make sense to have a fuzz target where we call `read_prefaced` on a reader that contains totally arbitrary bytes from the fuzzer? That way, we are also testing the behavior in the face of totally invalid inputs.

hawkw (commented):
Thanks for making the changes we discussed, this looks great. I had a few more comments, but none of them are critical --- I think this is close to ready to merge!

linkerd/transport-header/src/lib.rs
```diff
@@ -287,4 +287,14 @@ pub mod fuzz_logic {
         let mut buf = BytesMut::new();
         let _h = TransportHeader::read_prefaced(&mut rx, &mut buf).await;
     }
+
+    pub async fn fuzz_entry_raw(fuzz_data: &[u8]) {
```
hawkw, line 291: <3 thanks this is great!

hawkw (approved):
Thanks for adding the new fuzzer for invalid input to `TransportHeader::read_prefaced`! I think this PR looks great now --- it looks like `cargo fmt` needs to be run before it'll pass CI, but then I'm happy to merge it!

linkerd/transport-header/src/lib.rs
DavidKorczynski: Wohoo - awesome!

olix0r (approved):

PR #962, opened 2021-03-31
Decouple server config from OrigDstAddr 

olix0r:
Making the config types generic over the orig-dst strategy promulgates
generics across all of the stacks. But the binding logic is only
necessary at the very "top" of the stack where the accept loop is
driven.

This change eliminates the `GetAddrs` trait and modifies the `Bind`
trait to take params that allow the binding implementation to be
configured (i.e., from the environment). `OrigDstAddr` binding is
provided as a `Bind` implementation that wraps the `BindTcp` strategy.

This lets us eliminate the `mock-orig-dst` feature by moving the
`MockOrigDst` helpers into the integration test suite.


linkerd2-proxy/src/main.rs
```diff
@@ -28,8 +28,15 @@ fn main() {
...
     rt::build().block_on(async move {
+        let bind = BindTcp::with_orig_dst();
         let (shutdown_tx, mut shutdown_rx) = mpsc::unbounded_channel();
-        let app = match async move { config.build(shutdown_tx, trace?).await }.await {
+        let app = match async move {
+            config
+                .build(bind, bind, BindTcp::default(), shutdown_tx, trace?)
+                .await
+        }
+        .await
```
hawkw, line 38: oh, wow, huh, this sure is some formatting...

linkerd/proxy/transport/src/orig_dst.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+impl<A> Param<Local<ServerAddr>> for Addrs<A>
+where
+    A: Param<Local<ServerAddr>>,
+{
+    fn param(&self) -> Local<ServerAddr> {
+        self.inner.param()
+    }
+}
```
hawkw, line 46: sad that these can't just be a `impl<A, T> Param<T> for Addrs<A> where A: Param<T>`, but i guess there's no way to convince rustc that that wouldn't conflict with `Param<OrigDstAddrs>`...

linkerd/proxy/transport/src/listen.rs
```diff
@@ -12,339 +12,75 @@ use tokio_stream::wrappers::TcpListenerStream;
...
-    pub fn client(&self) -> Remote<ClientAddr> {
-        self.client
-    }
+        let server = Local(ServerAddr(listen.local_addr()?));
+        let Keepalive(keepalive) = params.param();
+        let accept = TcpListenerStream::new(listen).map(move |res| {
+            let tcp = res?;
+            super::set_nodelay_or_warn(&tcp);
+            super::set_keepalive_or_warn(&tcp, keepalive);
+            let client = Remote(ClientAddr(tcp.peer_addr()?));
```
hawkw, line 64: small nitpick: I put the "accepted" trace event here once we have the addrs, and we don't have that any more...nbd but it might be worth having

linkerd/app/outbound/src/tcp/tests.rs
```diff
@@ -697,11 +697,13 @@ async fn profile_endpoint_propagates_conn_errors() {
...
-        Remote(ClientAddr(([127, 0, 0, 1], 666).into())),
-        OrigDstAddr(ep1),
-    ));
+    let svc = server.new_service(orig_dst::Addrs {
+        orig_dst: OrigDstAddr(ep1),
+        inner: listen::Addrs {
+            server: Local(ServerAddr(([127, 0, 0, 1], 4140).into())),
+            client: Remote(ClientAddr(([127, 0, 0, 1], 666).into())),
+        },
+    });
```
hawkw, line 706: nit/TIOLI: maybe there should be an `orig_dst::Addrs::new(OrigDstAddr, Local<ServerAddr>, Remote<ClientAddr>)` constructor to make this slightly less wordy? not a big deal.

hawkw (approved):
this looks great, I think this factoring is much better than what I ended up with. I had some minor non-blocking nits, and also noticed one accidental change that i think we definitely need to fix, but besides that, I'm happy to merge this into my branch.

linkerd/proxy/transport/src/orig_dst.rs
```diff
@@ -0,0 +1,191 @@
...
+        let (addr, incoming) = self.inner.bind(t)?;
+
+        let incoming = incoming.map(|res| {
+            let (inner, tcp) = res?;
+            let orig_dst = orig_dst_addr(&tcp)?;
+            let addrs = Addrs { inner, orig_dst };
+            Ok((addrs, tcp))
+        });
+
+        Ok((addr, Box::pin(incoming)))
```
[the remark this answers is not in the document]
olix0r: I tried this and it doesn't work as you'd hope because the closure captures part of the type signature... if you can shave off the box, I'm open to it, though

PR #963, opened 2021-04-05
outbound: skip logical stacks when no profile is discovered

hawkw:
Currently, the outbound proxy's `Logical` target type contains an
`Option<profiles::Receiver>`. When no profile is discovered, this field
is set to `None`; the per-profile layers in the logical stack do nothing
when this field is `None`. However, we still *construct* this whole
stack for all destinations, even those without logical addresses.

This PR refactors the outbound proxy so that the `Logical` targets
*always* have discovered profiles, and the logical stack is skipped when
no profile is discovered. Profile discovery still returns an
`Option<profiles::Receiver>`, but we unwrap it in one place immediately
after the profile discovery layer, and skip directly to the endpoint
stack when it is `None`.

This means that we can now avoid constructing layers like traffic split
etc when there is no profile for a target, rather than building a
traffic split that does nothing.

Note that some of the layers in the logical stacks currently *do* still
have code for handling the case where a `Logical` target has no profile.
These layers currently take targets that are bounded with
`Param<Option<profiles::Receiver>>`. The `Logical` type still has its
`Param<Option<profiles::Receiver>>` impl, it just always returns `Some`
now. I chose to leave this code as-is since this branch is already a
fairly large refactor. We can simplify a lot of this code in a follow-up
(although some layers may also require a similar change on the inbound
side).

### DIFF
--- a/linkerd/app/outbound/src/endpoint.rs
+++ b/linkerd/app/outbound/src/endpoint.rs
@@ -1,0 +1,33 @@
+use crate::{target::Endpoint, Outbound};
+use linkerd_app_core::{svc, tls};
+
+impl<E> Outbound<E> {
+    pub fn push_into_endpoint<P, T>(
+        self,
+    ) -> Outbound<
+        impl svc::NewService<T, Service = <E as svc::NewService<Endpoint<P>>>::Service> + Clone,
+    >
+    where
+        Endpoint<P>: From<(tls::NoClientTls, T)>,
+        E: svc::NewService<Endpoint<P>> + Clone,
+    {
+        let Self {
+            config,
+            runtime,
+            stack: endpoint,
+        } = self;
+        let identity_disabled = runtime.identity.is_none();
+        let no_tls_reason = if identity_disabled {
+            tls::NoClientTls::Disabled
+        } else {
+            tls::NoClientTls::NotProvidedByServiceDiscovery
+        };
+        let stack =
+            svc::stack(endpoint).push_map_target(move |t| Endpoint::<P>::from((no_tls_reason, t)));
+        Outbound {
+            config,
+            runtime,
+            stack,
+        }
+    }
+}
--- a/linkerd/app/outbound/src/endpoint.rs
+++ b/linkerd/app/outbound/src/endpoint.rs
@@ -4,9 +4,7 @@ use linkerd_app_core::{svc, tls};
 impl<E> Outbound<E> {
     pub fn push_into_endpoint<P, T>(
         self,
-    ) -> Outbound<
-        impl svc::NewService<T, Service = <E as svc::NewService<Endpoint<P>>>::Service> + Clone,
-    >
+    ) -> Outbound<impl svc::NewService<T, Service = E::Service> + Clone>
     where
         Endpoint<P>: From<(tls::NoClientTls, T)>,
         E: svc::NewService<Endpoint<P>> + Clone,
--- a/linkerd/app/outbound/src/http/detect.rs
+++ b/linkerd/app/outbound/src/http/detect.rs
@@ -1,77 +1,39 @@
-use crate::{http, tcp, Outbound};
+use crate::{http, Outbound};
 use linkerd_app_core::{
     config::{ProxyConfig, ServerConfig},
-    detect, io, svc, Error,
+    detect, io,
+    svc::{self, Param},
+    Error,
 };
 use tracing::debug_span;
 
-impl<T> Outbound<T> {
-    /// Pushes a stack that detects HTTP, unless the service profile indicates
-    /// that this target uses an opaque protocol.
-    pub fn push_detect_with_skip<TSvc, H, HSvc, I>(
-        self,
-        http: H,
-    ) -> Outbound<
-        impl svc::NewService<
-                tcp::Logical,
-                Service = impl svc::Service<I, Response = (), Error = Error, Future = impl Send>,
-            > + Clone,
-    >
-    where
-        I: io::AsyncRead + io::AsyncWrite + io::PeerAddr + std::fmt::Debug + Send + Unpin + 'static,
-        T: svc::NewService<tcp::Logical, Service = TSvc> + Clone + Send + 'static,
-        TSvc: svc::Service<io::EitherIo<I, io::PrefixedIo<I>>, Response = ()> + Send + 'static,
-        TSvc::Error: Into<Error>,
-        TSvc::Future: Send,
-        H: svc::NewService<http::Logical, Service = HSvc> + Clone + Send + 'static,
-        HSvc: svc::Service<http::Request<http::BoxBody>, Response = http::Response<http::BoxBody>>,
-        HSvc: Clone + Send + Sync + Unpin + 'static,
-        HSvc::Error: Into<Error>,
-        HSvc::Future: Send,
-    {
-        let skipped = self
-            .clone()
-            .into_stack()
-            .push_on_response(svc::MapTargetLayer::new(io::EitherIo::Left))
-            .into_inner();
-        self.push_detect_http(http)
-            // When the profile marks the target as opaque, we skip HTTP
-            // detection and just use the TCP logical stack directly.
-            .push_switch(
-                |l: tcp::Logical| -> Result<_, Error> {
-                    if l.profile.borrow().opaque_protocol {
-                        Ok(svc::Either::B(l))
-                    } else {
-                        Ok(svc::Either::A(l))
-                    }
-                },
-                skipped,
-            )
-    }
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct SkipHttpDetection(pub bool);
 
-    pub fn push_detect_http<R, TSvc, TTgt, H, HSvc, HTgt, I>(
+impl<N> Outbound<N> {
+    pub fn push_detect_http<T, NSvc, NTgt, H, HSvc, HTgt, I>(
         self,
         http: H,
     ) -> Outbound<
         impl svc::NewService<
-                R,
+                T,
                 Service = impl svc::Service<I, Response = (), Error = Error, Future = impl Send>,
             > + Clone,
     >
     where
         I: io::AsyncRead + io::AsyncWrite + io::PeerAddr + std::fmt::Debug + Send + Unpin + 'static,
-        T: svc::NewService<TTgt, Service = TSvc> + Clone + Send + 'static,
-        TSvc: svc::Service<io::EitherIo<I, io::PrefixedIo<I>>, Response = ()> + Send + 'static,
-        TSvc::Error: Into<Error>,
-        TSvc::Future: Send,
-        TTgt: From<R> + 'static,
+        N: svc::NewService<NTgt, Service = NSvc> + Clone + Send + 'static,
+        NSvc: svc::Service<io::EitherIo<I, io::PrefixedIo<I>>, Response = ()> + Send + 'static,
+        NSvc::Error: Into<Error>,
+        NSvc::Future: Send,
+        NTgt: From<T> + 'static,
         H: svc::NewService<HTgt, Service = HSvc> + Clone + Send + 'static,
         HSvc: svc::Service<http::Request<http::BoxBody>, Response = http::Response<http::BoxBody>>,
         HSvc: Clone + Send + Sync + Unpin + 'static,
         HSvc::Error: Into<Error>,
         HSvc::Future: Send,
-        HTgt: From<(http::Version, R)> + svc::Param<http::Version> + 'static,
-        R: Clone + Send + Sync + 'static,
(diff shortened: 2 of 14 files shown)


linkerd/app/outbound/src/lib.rs
```diff
@@ -119,7 +136,27 @@ impl<S> Outbound<S> {
...
+            .into_inner();
+
+        // HTTP and TCP per-endpoint stack used when a profile is not discovered.
+        let endpoint = self
+            .clone()
+            .push_tcp_endpoint()
+            .push_tcp_forward()
+            .push_into_endpoint::<(), tcp::Accept>()
+            // If HTTP is detected, use the `http_endpoint` stack
+            .push_detect_http(http_endpoint)
```
kleimkuhler, line 155: If HTTP is detected, then we'd use the `http_endpoint` stack. That already has `push_into_endpoint` and `push_tcp_endpoint` on it though (lines 144 and 142). Can those be removed or changed in a way that we don't need to have those additional calls in this `endpoint` stack?

hawkw: we could call `push_tcp_endpoint` once and then clone it, but we can't reuse the `push_into_endpoint` call, because it's _before_ the TCP forward and HTTP endpoint stacks, which differ.

hawkw: sorry, never mind --- turns out this breaks type inference

kleimkuhler (approved):
Looks good! Sounds good about a follow-up to remove more of the `Option<profiles::Receiver>`s.

hawkw:
> Looks good! Sounds good about a follow-up to remove more of the `Option<profiles::Receiver>`s.

FWIW already started on this in #964

linkerd/app/outbound/src/logical.rs
```diff
@@ -0,0 +1,41 @@
...
+    where
+        tcp::Logical: From<(profiles::Receiver, T)>,
+        L: svc::NewService<tcp::Logical, Service = LSvc> + Clone,
+        LSvc: svc::Service<I, Response = (), Error = Error>,
+        LSvc::Future: Send,
+        E: svc::NewService<T, Service = ESvc> + Clone,
+        ESvc: svc::Service<I, Response = (), Error = Error>,
+        ESvc::Future: Send,
+    {
+        let Self {
```
olix0r, line 26: If I understand correctly, this just drops all of the endpoint-level configuration that may be present on the profile response? Is that correct? If so, this change won't work--we need to account for endpoint configuration that may be provided by the profile response.

linkerd/app/outbound/src/logical.rs
```diff
@@ -0,0 +1,41 @@
...
+            .push_map_target(tcp::Logical::from)
+            .push(svc::UnwrapOr::layer(endpoint))
+            .check_new_service::<(Option<profiles::Receiver>, T), _>();
+        Outbound {
+            config,
+            runtime,
+            stack,
+        }
+    }
+}
```
olix0r, line 41: Can be simplified somewhat, since we actually know the concrete service type:

```suggestion
use crate::{tcp, Outbound};
use linkerd_app_core::{profiles, svc};

impl<L> Outbound<L> {
    /// Pushes a layer that unwraps the [`Logical`] address of a given target
    /// from its profile resolution, or else falls back to the provided
    /// per-endpoint service if there was no profile resolution for that target.
    pub fn push_unwrap_logical<T, E>(
        self,
        endpoint: E,
    ) -> Outbound<
        impl svc::NewService<
                (Option<profiles::Receiver>, T),
                Service = svc::stack::ResultService<svc::Either<L::Service, E::Service>>,
            > + Clone,
    >
    where
        tcp::Logical: From<(profiles::Receiver, T)>,
        L: svc::NewService<tcp::Logical> + Clone,
        E: svc::NewService<T> + Clone,
    {
        let Self {
            config,
            runtime,
            stack: logical,
        } = self;
        let stack = logical
            .push_map_target(tcp::Logical::from)
            .push(svc::UnwrapOr::layer(endpoint));
        Outbound {
            config,
            runtime,
            stack,
        }
    }
}
```

linkerd/app/outbound/src/ingress.rs
```diff
@@ -63,19 +76,31 @@ impl Outbound<()> {
...
+            .push_into_endpoint()
+            .into_stack()
             .push_on_response(
                 svc::layers()
                     .push(http::BoxRequest::layer())
                     .push(svc::MapErrLayer::new(Into::into)),
             )
+            .into_inner();
+        let http_logical = http.push_http_logical(resolve).into_inner();
+        svc::stack(http_logical)
```
olix0r, line 98:
```suggestion

        http
            .push_http_logical(resolve)
            .into_stack()
```

linkerd/app/outbound/src/ingress.rs
```diff
@@ -85,6 +110,8 @@ impl Outbound<()> {
             // inner service to readiness even if new requests aren't
             // received.
             .push_map_target(http::Logical::from)
+            .push(svc::UnwrapOr::layer(http_endpoint))
```
olix0r, line 113: Mmmm.. I don't think we want to support this? We've already dropped the origdstaddr from the target, right? So what endpoint can this possibly route to?

linkerd/app/outbound/src/ingress.rs
```diff
@@ -17,11 +21,12 @@ impl Outbound<()> {
...
     /// outbound traffic is either HTTP or TLS'd by the ingress proxy.
-    pub fn to_ingress<T, I, N, NSvc, H, HSvc, P>(
+    pub fn to_ingress<T, I, N, NSvc, H, HSvc, P, R>(
         &self,
         profiles: P,
-        tcp: N,
-        http: H,
+        tcp: Outbound<N>,
+        http: Outbound<H>,
+        resolve: R,
```
olix0r, line 29: I'm skeptical about why this needs to change?

linkerd/app/outbound/src/http/detect.rs
```diff
@@ -6,7 +6,8 @@ use linkerd_app_core::{
 use tracing::debug_span;
 
 impl<T> Outbound<T> {
-    pub fn push_detect_http<TSvc, H, HSvc, I>(
+    /// Pushes a `DetectHttp` stack with a `SkipByProfile` layer.
```
olix0r, line 9: "The `SkipByProfile` layer" is an implementation detail... what does this mean functionally?

linkerd/app/outbound/src/endpoint.rs
```diff
@@ -0,0 +1,33 @@
+use crate::{target::Endpoint, Outbound};
+use linkerd_app_core::{svc, tls};
+
+impl<E> Outbound<E> {
+    pub fn push_into_endpoint<P, T>(
+        self,
+    ) -> Outbound<
+        impl svc::NewService<T, Service = <E as svc::NewService<Endpoint<P>>>::Service> + Clone,
+    >
```
olix0r, line 9:
```suggestion
    ) -> Outbound<impl svc::NewService<T, Service = E::Service> + Clone>
```

linkerd/app/outbound/src/target.rs
```diff
@@ -84,9 +84,16 @@ impl<P> From<(Option<profiles::Receiver>, Accept<P>)> for Logical<P> {
 }
 
 /// Used for traffic split
+impl<P> Param<profiles::Receiver> for Logical<P> {
+    fn param(&self) -> profiles::Receiver {
+        self.profile.clone()
+    }
+}
+
+// TODO(eliza): eventually we won't need this
```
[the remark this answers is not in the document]
hawkw: it's removed in #967

linkerd/app/outbound/src/logical.rs
hawkw: Oh, yeah, nice catch!

linkerd/app/outbound/src/ingress.rs
hawkw: Ah, yeah, I guess we should change this to a filter that rejects connections when there was no profile discovered?

hawkw:
@olix0r okay, I've addressed all of your feedback & I think this is ready for another look when you have a chance!

linkerd/app/outbound/src/http/detect.rs
```diff
@@ -6,7 +6,9 @@ use linkerd_app_core::{
 use tracing::debug_span;
 
 impl<T> Outbound<T> {
-    pub fn push_detect_http<TSvc, H, HSvc, I>(
+    /// Pushes a stack that detects HTTP, unless the service profile indicates
+    /// that this target uses an opaque protocol.
+    pub fn push_detect_with_skip<TSvc, H, HSvc, I>(
```
olix0r, line 11: nit: should probably be called `push_detect_http_with_skip` to be clear that's what we're detecting.

Also, I don't really think this should be separate from the main `push_detect_http`, ideally... I.e. could we have a marker param like `SkipHttpDetect` and require that targets of this stack impl `Param<SkipHttpDetect>`? this would help to decouple this from `tcp::Logical` and make it so that we don't need separate  stack builders (and duplicate type constraints)

linkerd/app/outbound/src/http/detect.rs
```diff
@@ -26,10 +28,54 @@ impl<T> Outbound<T> {
...
+        HSvc: Clone + Send + Sync + Unpin + 'static,
+        HSvc::Error: Into<Error>,
+        HSvc::Future: Send,
+        HTgt: From<(http::Version, R)> + svc::Param<http::Version> + 'static,
+        R: Clone + Send + Sync + 'static,
     {
         let Self {
             config,
-            runtime: rt,
+            runtime,
```
olix0r, line 78: why this change? we use the `rt` shorthand everywhere else, and it will help with some formatting issues below.

linkerd/app/outbound/src/http/detect.rs
```diff
@@ -44,52 +90,32 @@ impl<T> Outbound<T> {
...
             .push(svc::UnwrapOr::layer(
-                tcp.clone()
-                    .push_on_response(svc::MapTargetLayer::new(io::EitherIo::Right))
+                tcp.push_on_response(svc::MapTargetLayer::new(io::EitherIo::Right))
+                    .check_new_service::<TTgt, _>()
+                    .push_map_target(TTgt::from)
+                    .check_new_service::<R, _>()
                     .into_inner(),
             ))
+            .check_new_service::<(Option<http::Version>, R), _>()
```
olix0r, line 107:
```suggestion
                    .push_map_target(TTgt::from)
                    .into_inner(),
            ))
```

linkerd/app/outbound/src/http/detect.rs
```diff
@@ -26,10 +28,54 @@ impl<T> Outbound<T> {
...
+                Service = impl svc::Service<I, Response = (), Error = Error, Future = impl Send>,
+            > + Clone,
+    >
+    where
+        I: io::AsyncRead + io::AsyncWrite + io::PeerAddr + std::fmt::Debug + Send + Unpin + 'static,
+        T: svc::NewService<TTgt, Service = TSvc> + Clone + Send + 'static,
+        TSvc: svc::Service<io::EitherIo<I, io::PrefixedIo<I>>, Response = ()> + Send + 'static,
+        TSvc::Error: Into<Error>,
+        TSvc::Future: Send,
+        TTgt: From<R> + 'static,
```
olix0r, line 67: nit: It would probably be clearest (i.e. what we do elsewhere) to make `T` the target type (replacing `R`) and to use `N`/`NSvc`

olix0r (approved):
This is looking good though there's a bit of cleanup we should address. Doesn't necessarily have to be in this PR...

linkerd/app/outbound/src/http/detect.rs
hawkw: > Also, I don't really think this should be separate from the main `push_detect_http`, ideally... I.e. could we have a marker param like `SkipHttpDetect` and require that targets of this stack impl `Param<SkipHttpDetect>`? this would help to decouple this from `tcp::Logical` and make it so that we don't need separate stack builders (and duplicate type constraints)

Yeah, I didn't like that we ended up with two functions here, but I couldn't figure out a nice way to do it...the `Param` idea is a good one!

linkerd/app/outbound/src/http/detect.rs
hawkw: fixed in 8f9d9efdf1813afe370187c8c6990b414c8e93d3

linkerd/app/outbound/src/http/detect.rs
hawkw: fixed in 8f9d9efdf1813afe370187c8c6990b414c8e93d3

linkerd/app/outbound/src/http/detect.rs
hawkw: update, got rid of this in 8f9d9efdf1813afe370187c8c6990b414c8e93d3

PR #964, opened 2021-04-05
split: change traffic splits to require a profile

hawkw:
This branch builds on #963 by changing the traffic split service to
require its target type implement `Param<profiles::Receiver>` rather
than requiring `Param<Option<profiles::Receiver>>`. This lets us
simplify the implementation significantly by removing the "default" case
that's built when no profile was discovered.

Depends on #963


kleimkuhler (approved):
Cool this cleans up a lot 👍

olix0r (approved):
excellent!

PR #966, opened 2021-04-06
inbound: only build profile route stacks when a profile is resolved

hawkw:
This is similar to PR #963 but on the inbound side. Since the inbound
proxy only requires logical targets in the HTTP profile route layers,
this change is much simpler --- protocol detection etc are all above the
profile resolution layer. We can simply add an `UnwrapOr` layer that
skips the profile route layers.

This change is much less important than the corresponding outbound
change, but the two changes together will permit us to simplify the
profile route layer by taking a `Param<profiles::Receiver>` rather than
a `Param<Option<profiles::Receiver>>`.

This does also require adding an additional `BoxResponse` layer to erase
differing response body types based on whether or not profile route
metrics are recorded.

### DIFF
--- a/linkerd/app/inbound/src/http/mod.rs
+++ b/linkerd/app/inbound/src/http/mod.rs
@@ -154,7 +154,6 @@ where
             .push_on_response(http::BoxResponse::layer())
             .check_new_service::<Target, http::Request<_>>()
             .into_inner();
-            
         // Attempts to discover a service profile for each logical target (as
         // informed by the request's headers). The stack is cached until a
         // request has not been received for `cache_max_idle_age`.
--- a/linkerd/app/inbound/src/http/mod.rs
+++ b/linkerd/app/inbound/src/http/mod.rs
@@ -148,11 +148,13 @@ where
             .push_on_response(http_tracing::client(rt.span_sink.clone(), trace_labels()))
             .push_on_response(http::BoxResponse::layer())
             .check_new_service::<Target, http::Request<_>>();
+
         let no_profile = target
             .clone()
             .push_on_response(http::BoxResponse::layer())
             .check_new_service::<Target, http::Request<_>>()
             .into_inner();
+            
         // Attempts to discover a service profile for each logical target (as
         // informed by the request's headers). The stack is cached until a
         // request has not been received for `cache_max_idle_age`.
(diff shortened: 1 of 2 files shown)


kleimkuhler (approved):
Makes sense 👍

linkerd/app/inbound/src/http/mod.rs
```diff
@@ -148,7 +148,11 @@ where
             .push_on_response(http_tracing::client(rt.span_sink.clone(), trace_labels()))
             .push_on_response(http::BoxResponse::layer())
             .check_new_service::<Target, http::Request<_>>();
-
+        let no_profile = target
+            .clone()
+            .push_on_response(http::BoxResponse::layer())
+            .check_new_service::<Target, http::Request<_>>()
+            .into_inner();
```
olix0r, line 155: nit: vertical whitespace is easier on the eyes

```suggestion

        let no_profile = target
            .clone()
            .push_on_response(http::BoxResponse::layer())
            .check_new_service::<Target, http::Request<_>>()
            .into_inner();
            
```

linkerd/app/inbound/src/target.rs
```diff
@@ -351,14 +351,14 @@ impl<A> svc::stack::RecognizeRoute<http::Request<A>> for RequestTarget {
...
+impl From<(profiles::Receiver, Target)> for Logical {
+    fn from((profiles, target): (profiles::Receiver, Target)) -> Self {
         Self { profiles, target }
     }
 }
 
 impl Param<Option<profiles::Receiver>> for Logical {
     fn param(&self) -> Option<profiles::Receiver> {
-        self.profiles.clone()
+        Some(self.profiles.clone())
```
olix0r, line 362: Seems like we could update the router to require a profile now?

olix0r (approved):

linkerd/app/inbound/src/target.rs
hawkw: That's what ##967 does :) It's in a separate branch because that also depends on the outbound changes.

linkerd/app/inbound/src/http/mod.rs
```diff
@@ -149,6 +149,12 @@ where
             .push_on_response(http::BoxResponse::layer())
             .check_new_service::<Target, http::Request<_>>();
 
+        let no_profile = target
+            .clone()
+            .push_on_response(http::BoxResponse::layer())
+            .check_new_service::<Target, http::Request<_>>()
+            .into_inner();
+
```
olix0r, line 157: doh

```suggestion

```

PR #967, opened 2021-04-06
profiles: make receiver param in `route_request` non-optional 

hawkw:
This branch changes the `linkerd_service_profiles::http::route_request`
layer to require a `Param<profiles::Receiver>` rather than a
`Param<Option<profiles::Receiver>>` target type. This is possible
because the inbound and outbound logical destination stacks are now only
built when a profile is resolved (as of PRs #963 and #966).

Depends on #964
Depends on #966


kleimkuhler (approved):

olix0r (approved):
👍👍👍👍👍👍👍👍👍

PR #968, opened 2021-04-07
concurrency-limit: use `tokio-util`'s `PollSemaphore`

hawkw:
Currently, our concurrency limit middleware uses
`tokio::sync::Semaphore` to implement the concurrency limit. Because
acquiring permits from the `Semaphore` type is only possible via the
`acquire` and `acquire_owned` methods, which are `async fn`s returning
unameable futures that are `!Unpin`, we cannot simply poll the semaphore
in `poll_ready`. Instead, to allow polling the concurrency limit service
to acquire a permit, we box the `acquire_owned` future in `poll_ready`.
This means that every time we try to acquire a new concurrency limit
permit, we must allocate a new box.

The `tokio-util` crate has a [`PollSemaphore`][1] type which allows
wrapping a `tokio::sync::Semaphore` to add a `poll_acquire` method to
poll the semaphore to acquire a permit. Unlike our implementation,
`PollSemaphore` uses the [`tokio_util::sync::ReusableBoxFuture`][2] type
for boxing the `acquire_owned` future returned by the `Semaphore`.
`ReusableBoxFuture` is a safe abstraction around unsafe code that allows
*one* allocation to be reused for storing multiple type-erased
`Box::pin`ned futures. This means creating a single allocation for each
clone of a `ConcurrencyLimit` service, rather than having each clone of
the concurrency limit service allocate once *every* time a new permit is
acquired and drop that allocation once the permit is acquired.

This should reduce the overhead of the concurrency limit layer a bit. It
also makes the code much simpler! :)

[1]: https://docs.rs/tokio-util/0.6.5/tokio_util/sync/struct.PollSemaphore.html
[2]: https://docs.rs/tokio-util/0.6.5/tokio_util/sync/struct.ReusableBoxFuture.html

### DIFF
--- a/linkerd/concurrency-limit/src/lib.rs
+++ b/linkerd/concurrency-limit/src/lib.rs
@@ -9,14 +9,13 @@
 use linkerd_stack::layer;
 use pin_project::pin_project;
 use std::{
-    fmt,
     future::Future,
-    mem,
     pin::Pin,
     sync::Arc,
     task::{Context, Poll},
 };
-use tokio::sync::{OwnedSemaphorePermit, Semaphore};
+use tokio::sync::{OwnedSemaphorePermit as Permit, Semaphore};
+use tokio_util::sync::PollSemaphore;
 use tower::Service;
 use tracing::trace;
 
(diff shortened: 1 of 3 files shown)


linkerd/concurrency-limit/src/lib.rs
```diff
@@ -69,39 +62,30 @@ where
     type Future = ResponseFuture<S::Future>;
 
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        trace!(available = %self.semaphore.available_permits(), "acquiring permit");
+        trace!(
+            // available = %self.semaphore.available_permits(),
```
hawkw, line 66: logging the number of available permits here depends on upstream PR tokio-rs/tokio#3683

kleimkuhler (approved):
Nice! Thanks for the helpful description.

olix0r (approved):

PR #969, opened 2021-04-07
discover: replace `linkerd-channel` with `tokio-util` `PollSender`

hawkw:
This commit replaces the use of `linkerd-channel`'s pollable MPSC in the
`linkerd_discover` crate's `buffer` module with `tokio-util`'s
[`sync::PollSender`][1] type, which also implements a pollable sender.
The `PollSender` type is likely more efficient than our channel
implementation, since it uses the [`ReusableBoxFuture`][2] type from
`tokio-util`, which avoids allocating a `Box` for _every_ message sent
to the channel, and instead reuses a single allocation. It also wraps
the `tokio::sync::mpsc` `Sender`, rather than reimplementing the
capacity limit on top of the `UnboundedSender` type. This reduces
complexity and should hopefully be less error-prone.

[1]: https://docs.rs/tokio-util/0.6.5/tokio_util/sync/struct.PollSender.html
[2]: https://docs.rs/tokio-util/0.6.5/tokio_util/sync/struct.ReusableBoxFuture.html

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1070,12 +1070,12 @@ version = "0.1.0"
 dependencies = [
  "async-stream",
  "futures",
- "linkerd-channel",
  "linkerd-error",
  "linkerd-proxy-core",
  "linkerd-stack",
  "pin-project",
  "tokio",
+ "tokio-util",
  "tower",
  "tracing",
 ]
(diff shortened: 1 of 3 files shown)


kleimkuhler (approved):
👍

olix0r (approved):

PR #970, opened 2021-04-07
replace `linkerd-channel` with `tokio-stream`

hawkw:
Depends on #969.

It turns out that the trait impls for `linkerd_channel::Sender` in
`linkerd-trace-context` were just never being used at all. If we remove
them, we can remove the custom pollable channel implementation in
`linkerd_channel` entirely. Then, this crate will provide only stream
wrappers. Similar wrappers are also provided by `tokio-stream`. If we
replace the use of `linkerd-channel`'s stream wrappers with
`tokio-stream`, we can remove this crate entirely.

This branch completely removes the `linkerd-channel` crate and converts
all uses of its stream wrappers to use `tokio-stream` instead.


linkerd/trace-context/Cargo.toml
```diff
@@ -12,9 +12,8 @@ bytes = "1"
 futures = "0.3.9"
 hex = "0.3.2"
 http = "0.2"
-linkerd-channel = { path = "../channel" }
 linkerd-error = { path = "../error" }
 linkerd-stack = { path = "../stack" }
 rand = "0.8"
 tower = { version = "0.4.5", default-features = false, features = ["util"] }
-tracing = "0.1.2"
+tracing = "0.1.2"
```
olix0r, line 19:
```suggestion
tracing = "0.1.2"
```

olix0r (approved):

linkerd/trace-context/Cargo.toml
hawkw: Screenshot_20210408_115250
:thinking:

kleimkuhler (approved):
Nice!

PR #971, opened 2021-04-09
outbound: move target types into stack modules

hawkw:
Depends on #967

This branch refactors the `linkerd-app-outbound` crate so that each
stack target type is now defined in the module for the stack that uses
it, rather than in one big `target` module. For example, `Logical` is
now defined in `logical.rs` and `Endpoint` is defined in `endpoint.rs`.

There should be no functional change in this PR --- it just moves code
around.


kleimkuhler (approved):
Looks good!

olix0r (approved):

PR #972, opened 2021-04-09
outbound: only build logical stacks for profiles with logical addrs

hawkw:
Depends on #971

Currently, the proxy should only perform `Destination.Get` resolutions
for logical (named) addresses provided by `Destination.GetProfile`
lookups, and never for IP addresses. However, this is not *guaranteed*;
as the destination resolver still accepts a `LookupAddr` type which may
be either a logical named address *or* an IP address.

In order to remove support for IP lookups from the control plane's
`Destination.Get` implementation, we should ensure that the proxy will
*never* attempt to resolve an IP (see linkerd/linkerd2#5246). Now that
the `Logical` target type and the logical stacks are only constructed
when a profile is resolved (as of PR #963), we can now change the proxy
to ensure statically that this never happens.

This branch makes the following changes:

* Currently (as of #963), the `Logical` target type has an
  `Option<LogicalAddr>`. We attempt to unwrap the `Option<Receiver>`
  returned by a profile lookup, and build a logical stack when there is
  a profile. The `Logical` target is with `Option<LogicalAddr>` from
  that profile.

  However, this is not actually the ideal behavior: ideally, we would
  only contruct the logical stack when there is a resolved profile *and*
  the profile includes a logical address. Therefore, this branch changes
  the `push_unwrap_logical` function added in #963 to use a custom
  filter predicate that only constructs the `Logical` target when the
  profile resolution is `Some` _and_ the profile's `addr` field is
  `Some`. Now, `Logical` targets are only built when there is a known
  logical address.

* The `Destination.Get` resolver requires a target type that implements
  `Param<ConcreteAddr>`. Currently, a `ConcreteAddr` is a newtype around
  `Addr`, which may be either a named address or an IP address. This
  branch changes `ConcreteAddr` to a newtype around `NameAddr`. This
  means that the resolver now ensures `Destination.Get` queries are only
  performed for named addresses at the type level.

* Similarly, the `Target` type in `linkerd-service-profiles` is changed
  to store a `NameAddr` rather than an `Addr`.

* Several of the outbound stack tests now test invalid behavior that
  should no longer happen --- building a logical stack and calling it
  with a target that doesn't include a logical address. I rewrote these
  tests so that the tests without logical addresses don't build a
  logical stack, and added a new test for calling the logical stack
  with a logical address. In the process, I did some cleanup and
  refactoring in the outbound TCP stack tests.

* Finally, while debugging test failures, I noticed that the
  `fmt::Debug` output for some of the target types that include
  `NameAddr`s was a little unnecesarily verbose. I added new custom
  `Debug` impls that should make these a bit more concise.

This is the proxy component of linkerd/linkerd2#5246.


kleimkuhler (approved):
Few small comments but otherwise looks good.

linkerd/app/outbound/src/tcp/tests.rs
```diff
@@ -39,8 +39,63 @@ async fn plaintext_tcp() {
     // ports or anything. These will just be used so that the proxy has a socket
     // address to resolve, etc.
     let target_addr = SocketAddr::new([0, 0, 0, 0].into(), 666);
-    let profile_recv = profile::only_default();
-    let logical = build_logical(target_addr, profile_recv);
+
+    // Configure mock IO for the upstream "server". It will read "hello" and
+    // then write "world".
+    let mut srv_io = support::io();
+    srv_io.write(b"hello\n").read(b"world");
```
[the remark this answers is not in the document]
hawkw: So, it's slightly more complex: before this change, these tests were actually *not* doing what the comment describes. I had misinterpreted the mock IO builder API: `.write(...)` configures the mock IO to *expect* its `write` method to be called (the client will write "hello") while `.read(...)`  configures the mock IO to expect a read (the server will  write "world"). So the comment was actually wrong before...

linkerd/app/outbound/src/logical.rs
```diff
@@ -53,12 +59,15 @@ impl<P> svc::Param<profiles::LookupAddr> for Logical<P> {
...
+
 impl<P> Logical<P> {
     pub fn addr(&self) -> Addr {
-        self.logical_addr
-            .as_ref()
-            .map(|LogicalAddr(a)| Addr::from(a.clone()))
-            .unwrap_or_else(|| self.orig_dst.0.into())
+        Addr::from(self.logical_addr.clone().0)
     }
 }
```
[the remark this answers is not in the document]
hawkw: This _could_ be a `Param` impl. However, it's only used in places where the type *is* concretely known to be `Logical`, so there's no *reason* to make it a `Param` --- `Param` is used primarily in situations where we want to accept targets of multiple types, and only require that they can provide some particular parameter.

We could potentially make the code that uses this method more generic, but that would be a much larger change.

linkerd/app/gateway/src/gateway.rs
```diff
@@ -65,9 +65,9 @@ where
             None => return Gateway::BadDomain(http.target.name().clone()),
         };
 
-        let addr = match profile.borrow().addr.clone() {
+        let logical_addr = match profile.borrow().addr.clone() {
             Some(addr) => addr,
-            None => return Gateway::BadDomain(http.target.name().clone()),
+            _ => return Gateway::BadDomain(http.target.name().clone()),
```
olix0r, line 70:
```suggestion
            None => return Gateway::BadDomain(http.target.name().clone()),
```

linkerd/app/gateway/src/lib.rs
```diff
@@ -117,7 +117,11 @@ where
         .into_stack()
         .push_request_filter(|(p, _): (Option<profiles::Receiver>, _)| match p {
             Some(profile) if profile.borrow().addr.is_some() => {
-                let logical_addr = profile.borrow().addr.clone();
+                let logical_addr = profile
+                    .borrow()
+                    .addr
+                    .clone()
+                    .expect("if is_some returns true, this must be `Some`");
```
olix0r, line 124: this is probably fine but slightly better:

```rust
.push_request_filter(|(p, _): (Option<profiles::Receiver>, _)| {
    let profile = p.ok_or_else(discovery_rejected)?;
    let logical_addr = profile.borrow().addr.clone().ok_or_else(discovery_rejected)?;
    Ok(outbound::tcp::Logical {
        profile,
        logical_addr,
        orig_dst: OrigDstAddr(std::net::SocketAddr::from(([0, 0, 0, 0], 0))),
        protocol: (),
    })
})
```

linkerd/app/inbound/src/http/mod.rs
```diff
@@ -148,7 +148,6 @@ where
             .push_on_response(http_tracing::client(rt.span_sink.clone(), trace_labels()))
             .push_on_response(http::BoxResponse::layer())
             .check_new_service::<Target, http::Request<_>>();
-
```
olix0r, line 151: prefer empty lines

linkerd/app/outbound/src/logical.rs
```diff
@@ -64,10 +76,7 @@ impl svc::Param<SkipHttpDetection> for Logical<()> {
 
 impl<P> Logical<P> {
     pub fn addr(&self) -> Addr {
-        self.logical_addr
-            .as_ref()
-            .map(|LogicalAddr(a)| Addr::from(a.clone()))
-            .unwrap_or_else(|| self.orig_dst.0.into())
+        Addr::from(self.logical_addr.clone().0)
```
olix0r, line 79: this is only used in one place now, probably best to just inline rather than have a dedicated helper

linkerd/app/outbound/src/logical.rs
```diff
@@ -134,34 +143,66 @@ impl<P> svc::Param<ConcreteAddr> for Concrete<P> {
...
         } = self;
         let stack = logical
-            .push_map_target(tcp::Logical::from)
-            .push(svc::UnwrapOr::layer(endpoint));
+            .push(svc::layer::mk(move |primary| {
+                svc::Filter::new(
+                    stack::NewEither::new(primary, endpoint.clone()),
+                    unwrap_logical,
+                )
+            }))
```
olix0r, line 204: Is this different than `push_switch`? This looks like a `push_switch`.

I'd generally prefer to keep the predicate inlined so the logic for consistency.

olix0r (commented):
Looking pretty good. A few comments

linkerd/app/outbound/src/logical.rs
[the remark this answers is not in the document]
kleimkuhler: That makes sense. Since we know the type here I think keeping as-is is good.

linkerd/app/outbound/src/logical.rs
hawkw: Hmm, I don't think that's true, this is used in a bunch of places?
Screenshot_20210414_093823

We definitely could remove it, though, specially since it's much simpler now, but it would be a larger diff.

olix0r (approved):

PR #974, opened 2021-04-11
compile: fix signature mismatching which causes compiling failure for non-linux target

ihciah:
fix signature mismatching which causes compiling failure for non-linux target


hawkw (approved):
whoops, thanks for fixing this!

olix0r (approved):
Thank you!

PR #975, opened 2021-04-11
Update trust-dns to fix possible panic

olix0r:
Our fuzz tests trigger a panic in trust-dns when resolving names through
the search path (i.e., in `/etc/hosts`) as described in
bluejekyll/trust-dns#1447.

This change patches our trust-dns dependencies to use include
bluejekyll/trust-dns@6dfc671 (which has not yet been released to
crates.io)

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33123

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2344,9 +2344,8 @@ dependencies = [
 
 [[package]]
 name = "trust-dns-proto"
-version = "0.20.0"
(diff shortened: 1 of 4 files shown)


hawkw (approved):
looks good to me!

kleimkuhler (approved):

PR #977, opened 2021-04-12
app: inbound: add fuzzer

DavidKorczynski:
The focus of this PR is adding a n end-to-end style fuzzer. 

Notice I had to disable denying of warnings in two places in order to compile this. 

The fuzzer is largely inspired by `unmeshed_http1_hello_world`
https://github.com/linkerd/linkerd2-proxy/blob/863ef23976cdf3821eea83b78ee8466bdfa46772/linkerd/app/inbound/src/http/tests.rs#L49

### DIFF
--- a/linkerd/app/core/src/lib.rs
+++ b/linkerd/app/core/src/lib.rs
@@ -7,7 +7,7 @@
 //! - Tap
 //! - Metric labeling
 
-//#![deny(warnings, rust_2018_idioms)]
+#![deny(warnings, rust_2018_idioms)]
 
 pub use linkerd_addr::{self as addr, Addr, NameAddr};
 pub use linkerd_cache as cache;
--- a/linkerd/app/core/src/lib.rs
+++ b/linkerd/app/core/src/lib.rs
@@ -7,7 +7,7 @@
 //! - Tap
 //! - Metric labeling
 
-#![deny(warnings, rust_2018_idioms)]
+//#![deny(warnings, rust_2018_idioms)]
 
 pub use linkerd_addr::{self as addr, Addr, NameAddr};
 pub use linkerd_cache as cache;
--- a/linkerd/app/inbound/fuzz/fuzz_targets/fuzz_target_1.rs
+++ b/linkerd/app/inbound/fuzz/fuzz_targets/fuzz_target_1.rs
@@ -1,22 +1,26 @@
 #![no_main]
-use libfuzzer_sys::fuzz_target;
 use libfuzzer_sys::arbitrary::Arbitrary;
-extern crate hyper;
+use libfuzzer_sys::fuzz_target;
 
 #[derive(Debug, Arbitrary)]
 struct TransportHeaderSpec {
-        uri: Vec<u8>,
-        header1: Vec<u8>,
-        header2: Vec<u8>,
-        http_method: bool,
+    uri: Vec<u8>,
+    header_name: Vec<u8>,
+    header_value: Vec<u8>,
+    http_method: bool,
 }
 
 fuzz_target!(|inp: TransportHeaderSpec| {
     if let Ok(uri) = std::str::from_utf8(&inp.uri[..]) {
-        if let Ok(header1_s) = std::str::from_utf8(&inp.header1[..]) {
-            if let Ok(header2_s) = std::str::from_utf8(&inp.header2[..]) {
+        if let Ok(header_name) = std::str::from_utf8(&inp.header_name[..]) {
+            if let Ok(header_value) = std::str::from_utf8(&inp.header_value[..]) {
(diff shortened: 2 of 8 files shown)


linkerd/app/inbound/fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -0,0 +1,24 @@
...
+struct TransportHeaderSpec {
+        data: Vec<u8>,
+        h1: Vec<u8>,
+        h2: Vec<u8>,
+        protocol: bool,
+}
+
+//fuzz_target!(|data: &[u8]| {
+fuzz_target!(|inp: TransportHeaderSpec| {
+    if let Ok(s) = std::str::from_utf8(&inp.data[..]) {
```
olix0r, line 16: Since this is used as a URI, do we want to validate that the string is a valid URI? Or is the intent to test invalid URIs?

DavidKorczynski: I would first go for invalid URIs too, then, if we observe any issues with it e.g. false positives we can fix it. In my test this seems to work well.

linkerd/app/core/src/lib.rs
```diff
@@ -7,7 +7,7 @@
 //! - Tap
 //! - Metric labeling
 
-#![deny(warnings, rust_2018_idioms)]
+//#![deny(warnings, rust_2018_idioms)]
```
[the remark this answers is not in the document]
DavidKorczynski: Solution is to simply handle the lint level from the command line when compiling the fuzzers.

linkerd/app/inbound/src/http/mod.rs
```diff
@@ -238,3 +238,127 @@ fn trace_labels() -> std::collections::HashMap<String, String> {
...
+        proxy,
+        svc::{self, NewService, Param},
+        tls,
+        transport::{ClientAddr, Remote, ServerAddr},
+        Conditional, Error, NameAddr, ProxyRuntime,
+    };
+    pub use linkerd_app_test as support;
+    use linkerd_app_test::*;
+
+    pub async fn fuzz_entry_raw(uri: &str, header_data_1: &str, header_data_2: &str, is_get: bool) {
```
hawkw, line 265: Hmm, it looks like `header_data_1` is actually the header _name_, while `header_data_2` is the _header value_. What if we called these
```suggestion
    pub async fn fuzz_entry_raw(uri: &str, header_name: &str, header_value: &str, is_get: bool) {
```

linkerd/app/inbound/fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -0,0 +1,23 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+use libfuzzer_sys::arbitrary::Arbitrary;
+extern crate hyper;
+
+#[derive(Debug, Arbitrary)]
+struct TransportHeaderSpec {
+        uri: Vec<u8>,
+        header1: Vec<u8>,
+        header2: Vec<u8>,
```
hawkw, line 10: This naming suggests to me that this is two separate headers, but actually it's a header name and a header value. Can we change the naming to reflect this?

linkerd/app/inbound/src/http/mod.rs
```diff
@@ -238,3 +238,127 @@ fn trace_labels() -> std::collections::HashMap<String, String> {
...
+        {
+            let rsp = http_util::http_request(&mut client, req).await;
+            let _body = http_util::body_to_string(rsp.into_body()).await;
+
+            // Let's send one with a correct URL
+            let req2 = Request::builder()
+                .method(http::Method::GET)
+                .uri("http://foo.svc.cluster.local:5550")
+                .body(Body::default())
+                .unwrap();
```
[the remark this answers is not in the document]
DavidKorczynski: just any traffic at this point. I think this fuzzer would benefit from creating a sequence of requests rather than just a single random request (as it is now), however, I would like to test the whole end-to-end concept with a single random request at first.

olix0r (approved):

hawkw (approved):
This looks like a good start to me!

I'd love to see fuzzing for more variables in the future --- it would be nice to be able to produce more potentially-malicious requests. But, we can always keep adding more fuzzers, and this looks like a great start!

DavidKorczynski:
> This looks like a good start to me!
> 
> I'd love to see fuzzing for more variables in the future --- it would be nice to be able to produce more potentially-malicious requests. But, we can always keep adding more fuzzers, and this looks like a great start!

Great stuff! I agree, and I will get this up in the near future. It would be nice to do it a bit incrementally though so we can see how things work when clusterfuzz starts putting a lot of compute power on it.

PR #978, opened 2021-04-16
app: inbound: fuzzer: generalise fuzzers and solve coverage build

DavidKorczynski:
This PR is an effort to generalise the inbound fuzzer by allowing it to sending an arbitrary number of random packets. 

This PR also moves the structs used by the structure-aware to be defined outside the fuzz_*.rs files. This solves the coverage compilation problems we currently have - as such, following this PR we will be able to get nice coverage reports shown by OSS-Fuzz.

Signed-off-by: davkor <david@adalogics.com>

### DIFF
--- a/linkerd/app/inbound/fuzz/fuzz_targets/fuzz_target_1.rs
+++ b/linkerd/app/inbound/fuzz/fuzz_targets/fuzz_target_1.rs
@@ -1,27 +1,11 @@
 #![no_main]
-use libfuzzer_sys::arbitrary::Arbitrary;
 use libfuzzer_sys::fuzz_target;
 
-#[derive(Debug, Arbitrary)]
-struct TransportHeaderSpec {
-    uri: Vec<u8>,
-    header_name: Vec<u8>,
-    header_value: Vec<u8>,
-    http_method: bool,
-}
-
-fuzz_target!(|inp: TransportHeaderSpec| {
-    if let Ok(uri) = std::str::from_utf8(&inp.uri[..]) {
-        if let Ok(header_name) = std::str::from_utf8(&inp.header_name[..]) {
-            if let Ok(header_value) = std::str::from_utf8(&inp.header_value[..]) {
-                let rt = tokio::runtime::Runtime::new().unwrap();
-                rt.block_on(linkerd_app_inbound::http::fuzz_logic::fuzz_entry_raw(
-                    uri,
-                    header_name,
-                    header_value,
-                    inp.http_method,
-                ));
-            }
-        }
+fuzz_target!(|packets: Vec<linkerd_app_inbound::http::fuzz_logic::TransportHeaderSpec>| {
+    if packets.len() == 0 {
+        return;
     }
+
+    let rt = tokio::runtime::Runtime::new().unwrap();
+    rt.block_on(linkerd_app_inbound::http::fuzz_logic::fuzz_entry_raw(packets));
 });
--- a/linkerd/app/inbound/fuzz/fuzz_targets/fuzz_target_1.rs
+++ b/linkerd/app/inbound/fuzz/fuzz_targets/fuzz_target_1.rs
@@ -1,11 +1,12 @@
 #![no_main]
 use libfuzzer_sys::fuzz_target;
+use linkerd_app_inbound::http::fuzz_logic::*;
 
-fuzz_target!(|packets: Vec<linkerd_app_inbound::http::fuzz_logic::TransportHeaderSpec>| {
-    if packets.len() == 0 {
+fuzz_target!(|requests: Vec<HttpRequestSpec>| {
+    if requests.len() == 0 {
         return;
     }
 
     let rt = tokio::runtime::Runtime::new().unwrap();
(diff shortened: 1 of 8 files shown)


DavidKorczynski:
@hawkw you suggested also passing non utf8 strings, we could do this by creating our requests by way of the fuzz data as follows `header_value = std::str::from_utf8_unchecked(&inp.header_value[..])`, but this would mean we should wrap it in unsafe. Is this what you had in mind?

DavidKorczynski:
The CI looks to be complaining about something unrelated to the PR? https://github.com/linkerd/linkerd2-proxy/pull/978/checks?check_run_id=2363287950

linkerd/app/inbound/fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -1,27 +1,11 @@
...
-                let rt = tokio::runtime::Runtime::new().unwrap();
-                rt.block_on(linkerd_app_inbound::http::fuzz_logic::fuzz_entry_raw(
-                    uri,
-                    header_name,
-                    header_value,
-                    inp.http_method,
-                ));
-            }
-        }
+fuzz_target!(|packets: Vec<linkerd_app_inbound::http::fuzz_logic::TransportHeaderSpec>| {
```
olix0r, line 4: nit: You could
```rust
use linkerd_app_inbound::http::fuzz_logic::*;
```
at the top to get rid of the long paths in the implementation.

But not a blocker from my point of view.

linkerd/transport-header/fuzz/fuzz_targets/fuzz_target_structured.rs
```diff
@@ -1,22 +1,11 @@
...
-struct TransportHeaderSpec {
-    data: Vec<u8>,
-    port: u16,
-    protocol: bool,
-}
-
-fuzz_target!(|inp: TransportHeaderSpec| {
-    if let Ok(s) = std::str::from_utf8(&inp.data[..]) {
+fuzz_target!(
+    |inp: linkerd_transport_header::fuzz_logic::TransportHeaderSpec| {
```
olix0r, line 5: And similarly here...

olix0r (approved):
lgtm

hawkw:
> @hawkw you suggested also passing non utf8 strings, we could do this by creating our requests by way of the fuzz data as follows `header_value = std::str::from_utf8_unchecked(&inp.header_value[..])`, but this would mean we should wrap it in unsafe. Is this what you had in mind?

Yes, or using [`HeaderValue::from_maybe_shared_unchecked`](https://docs.rs/hyper/0.14.5/hyper/header/struct.HeaderValue.html#method.from_maybe_shared_unchecked). The one concern with doing this is that we might end up testing the behavior of the hyper client used in the fuzz test when handling invalid header values, rather than the proxy's behavior --- we would want to make sure we were actually sending the invalid bytes to the test proxy, rather than just hitting an assertion in the test support code or something...

So, I think we can save this for a future change?

linkerd/app/inbound/src/http/mod.rs
```diff
@@ -270,7 +271,15 @@ pub mod fuzz_logic {
     pub use linkerd_app_test as support;
     use linkerd_app_test::*;
 
-    pub async fn fuzz_entry_raw(uri: &str, header_name: &str, header_value: &str, is_get: bool) {
+    #[derive(Debug, Arbitrary)]
+    pub struct TransportHeaderSpec {
```
hawkw, line 275: this probably should not be called `TransportHeaderSpec`, this is for testing the inbound HTTP proxy, not the linkerd transport header parser...

linkerd/app/inbound/src/http/mod.rs
```diff
@@ -295,29 +304,29 @@ pub mod fuzz_logic {
...
-        if let Ok(req) = Request::builder()
-            .method(http_method)
-            .uri(uri)
-            .header(header_name, header_value)
-            .body(Body::default())
-        {
-            let rsp = http_util::http_request(&mut client, req).await;
-            let _body = http_util::body_to_string(rsp.into_body()).await;
+        // Now send all of the packets
+        for inp in packets.iter() {
```
hawkw, line 308: this is nitpicky, but I would probably not refer to the HTTP requests as "packets" --- i would expect a value named `packets` to be a vector of individual IP-layer packets, rather than HTTP requests.

maybe call the variable `requests`, instead?

hawkw (approved):
Overall, this looks good to me, although I had a couple of style nits that it would be good to address.

linkerd/app/inbound/fuzz/fuzz_targets/fuzz_target_1.rs
DavidKorczynski: Thanks - fixed!

linkerd/app/inbound/src/http/mod.rs
DavidKorczynski: Ah yes - thanks for catching it!

linkerd/app/inbound/src/http/mod.rs
DavidKorczynski: Agreed!

PR #979, opened 2021-04-19
admin: Fail connections when HTTP detection fails

olix0r:
fce8c4ae updated the admin server to try to handle connections that fail
HTTP detection. The intent of this was to handle connections that
don't emit any HTTP request within the first few seconds of the
connection. However, we also try to handle connections on which we have
read data and affirmatively indicated that the connection is not HTTP.
These connections manifest with errors about an invalid HTTP method,
which can be ambiguous to users.

This change handles this latter case explicitly with an error that
describes the client so that users have better diagnostics without
having to enable debug logging.


kleimkuhler (approved):
👍

PR #980, opened 2021-04-19
reduce error boilerplate with `thiserror`

hawkw:
The [`thiserror` crate][1] provides a `#[derive(Error)]` macro that can be
used to generate `std::error::Error` implementations. Using this results
in significantly less boilerplate required by most of the error types in
the proxy, replacing almost all of the manual `Error` implementations.

There were a few cases where the `Error` impl either required complex
formatting logic (it has a conditional in it), or complex `source` logic
(the `Arc`ed `SharedError` type). I left these alone, but was able to
replace every other manual implementation of `std::error::Error`.

No functional change. :)

[1]: https://github.com/dtolnay/thiserror#deriveerror


linkerd/addr/Cargo.toml
```diff
@@ -9,3 +9,4 @@ publish = false
 [dependencies]
 http = "0.2"
 linkerd-dns-name = { path = "../dns/name" }
+thiserror = "1.0"
```
olix0r, line 12: nit

```suggestion
thiserror = "1.0"
```

linkerd/dns/name/Cargo.toml
```diff
@@ -9,3 +9,4 @@ publish = false
 [dependencies]
 webpki = "0.21"
 untrusted = "0.7"
+thiserror = "1.0"
```
olix0r, line 12: nit: sort?

linkerd/dns/src/lib.rs
```diff
@@ -22,7 +23,8 @@ pub trait ConfigureResolver {
     fn configure_resolver(&self, _: &mut ResolverOpts);
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Error)]
+#[error("Invalid SRV record {:?}", self.0)]
```
olix0r, line 27: we've used lowercase elsewhere, mostly
```suggestion
#[error("invalid SRV record {:?}", self.0)]
```

olix0r (approved):

kleimkuhler (approved):
👍

PR #984, opened 2021-04-21
Commit lock files for fuzzers

olix0r:
This change adds fuzz crate lock files to the repo so that
fuzz tests can't regress based on changes in crates.io. It
also removes the per-crate gitignore files by including all
ignored targets in the top-level gitignore.

Finally, we add a CI task to try to build all fuzzers. As this
depends on rust nightly, it will not be a required check.


hawkw:
> It also removes the per-crate gitignore files by including all
> ignored targets in the top-level gitignore.

FWIW, the per-crate gitignores are automatically generated by `cargo fuzz init` per https://github.com/linkerd/linkerd2-proxy/pull/961#discussion_r605507742, so if we generate more fuzz crates that way, we'll have to remember to remove them. Not a big deal.

.github/workflows/rust.yml
```diff
@@ -53,6 +53,19 @@ jobs:
...
+    runs-on: ubuntu-latest
+    container:
+      image: docker://rust:1.49.0-buster
+    steps:
+      - uses: actions/checkout@v2
+      - run: rustup toolchain add nightly
+      - run: cargo install cargo-fuzz
+      # Iterate through all fuzz crates to ensure each compiles indpendently.
+      - run: for d in $(find . -name fuzz | sort -r) ; do echo "# $d" ; (cd $d ; cargo +nightly fuzz build) ; done
+      - run: git diff-index --quiet HEAD # Error if the repo isn't clean (i.e. because lockfiles were modified).
```
hawkw, line 67: could we also use `cargo build --locked here?

.github/workflows/rust.yml
```diff
@@ -53,6 +53,19 @@ jobs:
...
+    timeout-minutes: 20
+    runs-on: ubuntu-latest
+    container:
+      image: docker://rust:1.49.0-buster
+    steps:
+      - uses: actions/checkout@v2
+      - run: rustup toolchain add nightly
+      - run: cargo install cargo-fuzz
+      # Iterate through all fuzz crates to ensure each compiles indpendently.
+      - run: for d in $(find . -name fuzz | sort -r) ; do echo "# $d" ; (cd $d ; cargo +nightly fuzz build) ; done
```
hawkw, line 66: does this need `--features fuzzing` to work?

linkerd/dns/Cargo.toml
```diff
@@ -16,6 +16,6 @@ tokio = { version = "1", features = ["rt", "sync", "time"] }
 pin-project = "1"
 
 [dependencies.trust-dns-resolver]
-version = "0.20.0"
+version = "=0.20.1"
```
hawkw, line 19: Would be nice to be able to remove this...maybe worth a comment about what upstream changes we are waiting on before we can remove the `=`?

hawkw (approved):
lgtm, thanks for the lockfile fix! i had a couple comments.

.github/workflows/rust.yml
olix0r: On closer inspection, we don't actually need this feature at all. I've removed the features and we now rely on `cfg(fuzzing)` gates instead (which also prevents fuzzing deps from being pulled in when running `cargo check`

hawkw: Great --- I wasn't entirely sure why there was both a feature flag and a cfg!

.github/workflows/rust.yml
olix0r: I'm not sure but I don't think so, due to cfg gates on dependencies.

PR #986, opened 2021-04-21
fuzz: don't panic when the proxy closes the conn

hawkw:
Currently, the inbound HTTP fuzz tests are failing. This is because they
reuse the test-support code for making an HTTP request to a proxy stack
and running the futures necessary to drive that request in the
background. This code currently unwraps both the `JoinHandle`s of the
spawned tasks (which would be `Err` if the task panicked) _and_ the
returned `Result` from those `JoinHandle`s (which is an `Err` if the
`Service::call` future returned an error, or if the client returned an
error). If the future completes with an error, then the proxy simply
tears down the connection.

In the integration tests, we currently `expect` _both_ of these
`Results` --- since the inputs are valid, we want to assert the proxy
doesn't return an error incorrectly. However, the fuzz tests can and
will generate malformed HTTP requests, and in this case, the proxy will
reject those requests by returning an error and closing the connection.
This is *not* incorrect behavior. Instead, we want to ensure that the
proxy doesn't panic in the face of potentially malformed requests.

This branch changes the test-support HTTP code to return the `Result` of
serving a request, and unwrap it in the integration tests rather than in
the support code. The fuzz logic is updated to simply log errors
returned here, since returning an error is expected behavior when we
receive invalid inputs.

If we wanted to be _really_ fancy, a potential follow-up would be to
extend the fuzz logic to determine whether or not a fuzz spec should
result in an error, and assert that errors are only returned for invalid
requests...but, doing this without using any of the code that's being
exercised in the fuzz test (e.g. all of `hyper`'s request parsing etc)
would be a _lot_ of work...

Depends on #985


linkerd/app/test/src/http_util.rs
```diff
@@ -93,11 +93,15 @@ where
...
+        let proxy = async move { proxy.await.expect("proxy background task panicked") };
+        let client_bg = async move { client_bg.await.expect("client background task panicked") };
+        let res = tokio::join! {
             proxy,
             client_bg,
         };
-        res.unwrap();
+        res.0?;
+        res.1?;
+        Ok(())
```
olix0r, line 104: what's going on here? given that these are join handles, why do we need to stash these in async closures and use join? I think this can be:

```suggestion
         proxy
             .await
             .expect("proxy task panicked")
             .await?;
         client_bg
             .await
             .expect("client task panicked")
             .await?;
         Ok(())
```

olix0r (approved):

linkerd/app/test/src/http_util.rs
hawkw: IIRC, I used `join` ages ago because I wasn't sure which task would complete first, and I didn't want it to hang...awaiting them in sequence is probably fine. They're just `JoinHandle<Result<...>>`, though, not `JoinHandle<impl Future<...>>`, so the second `await` in your example isn't necessary.

hawkw: ...i guess they're spawned so it doesn't actually matter.

olix0r: Updated the suggestion to have a single await

hawkw: yeah, i checked and it turns out i was totally overthinking it whenever it was that i wrote this originally, the join is completely unnecessary.

PR #987, opened 2021-04-22
[WIP] let the gateway crate define its own stack targets

hawkw:
This is a draft because it needs a lot of cleanup, but it works!


hawkw:
Closing this as it has gotten quite stale.

PR #988, opened 2021-04-23
outbound: fix profile endpoint overrides

hawkw:
PR #963 introduced a bug where, when a service profile contains an
endpoint override but no logical address, the proxy would ignore the
overridden endpoint. This is because the `unwrap_logical` layer checked
for the existence of the profile's logical address, as the `Logical`
target type has been changed to require a logical address. This means
that we would skip the logical stack correctly for endpoint overrides,
but would forward to the original destination address *without* honoring
the overridden address or any metadata provided by the profile. This bug
was detected by the control plane integration tests in the
linkerd/linkerd2 repo.

This branch fixes this by introducing a new stack for endpoint
overrides, which is now pushed *before* the `unwrap_logical` layer. Now,
we will first check if a profile with an endpoint override was
discovered, and if one was, we will build a stack for forwarding to that
endpoint, bypassing the logical stack. If there wasn't an endpoint
override, we will check if there is a profile, and build either a
logical stack or a direct endpoint stack. This required some
restructuring of the existing stacks.

I've also added new tests for profile endpoint overrides, with and
without logical addresses in the profile. This includes HTTP *and* TCP
stack tests, and the tests without logical overrides all fail prior to
this change. While I was working on wiring up these tests, I noticed
that the current stack tests all build their own custom server stacks,
rather than using the `Outbound::into_server` function. This means any
changes to the overall structure of the outbound stack *also* requires
updating the tests to match...so even after fixing the bug, the tests
continued to fail until the test stacks were changed to match the actual
ones. Rather than continuing to manually keep these in sync, I did some
additional refactoring to allow the tests to just call
`Outbound::into_server`, passing in HTTP and TCP endpoint stacks. Now,
the overall stack structure that is actually built in a real proxy is
always used in the tests without having to keep them in sync manually.

I think in a follow-up we may be able to remove additional logic for
handling endpoint overrides from the logical stack. However, I didn't
remove this code in this branch because I believe it's used by the
gateway and ingress code. I'll do some additional refactors in a
follow-up PR.

### DIFF
--- a/linkerd/app/outbound/src/endpoint.rs
+++ b/linkerd/app/outbound/src/endpoint.rs
@@ -1,15 +1,16 @@
 use crate::{
-    logical::{Concrete, Logical},
+    http::SkipHttpDetection,
+    logical::{Concrete, Logical, LogicalAddr},
     tcp::opaque_transport,
     Accept, Outbound,
 };
 use linkerd_app_core::{
-    metrics,
+    metrics, profiles,
     proxy::{api_resolve::Metadata, http, resolve::map_endpoint::MapEndpoint},
     svc::{self, Param},
     tls,
     transport::{self, Remote, ServerAddr},
-    transport_header, Addr, Conditional,
+    transport_header, Addr, Conditional, Error,
 };
 use std::net::SocketAddr;
 
@@ -22,10 +23,22 @@ pub struct Endpoint<P> {
     pub protocol: P,
 }
 
+/// An endpoint from a profile endpoint override.
+///
+/// This has to be a type, rather than a tuple, so that we can implement
+/// `Param<SkipHttpDetection>` for it.
+#[derive(Clone, Debug)]
+pub struct ProfileOverride<P> {
+    endpoint: Endpoint<P>,
+    profile: profiles::Receiver,
+}
+
 #[derive(Copy, Clone)]
 pub struct FromMetadata {
     pub identity_disabled: bool,
 }
+
+pub type OrOverride<S, E> = svc::stack::ResultService<svc::Either<S, E>>;
 
 impl<E> Outbound<E> {
     pub fn push_into_endpoint<P, T>(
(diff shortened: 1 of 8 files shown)


linkerd/app/outbound/src/endpoint.rs
```diff
@@ -54,6 +66,72 @@ impl<E> Outbound<E> {
...
+                        addr: Remote(ServerAddr(addr)),
+                        tls,
+                        metadata,
+                        logical_addr,
+                        protocol: (),
+                    };
+                    let opaque_protocol = profile.opaque_protocol;
+                    return Ok(svc::Either::B(ProfileOverride { endpoint, opaque_protocol }));
+                }
+            };
```
kleimkuhler, line 124: It looks like formatting for this is a little off. Also I don't think the last bracket here needs the `;` after it.

linkerd/app/outbound/src/http/tests.rs
```diff
@@ -448,6 +459,78 @@ async fn active_stacks_dont_idle_out() {
         .expect("proxy background task failed");
 }
 
+mod profile_endpoint_override {
+    use super::*;
+
+    async fn test_endpoint_override(profile: profile::Profile) {
+        // Tests that when a profile is resolved with no logical name, but an
+        // endpoint override, the endpoint override is honored.`s`
```
kleimkuhler, line 467: If if there is a logical name, as long as `endpoint` is populated we honor that right?

kleimkuhler (commented):
Aside from formatting looked a little odd in one place this looks good.

linkerd/app/outbound/src/endpoint.rs
hawkw, line 124: @olix0r as a follow-up, i'm thinking about also rolling this logic into this layer: https://github.com/linkerd/linkerd2-proxy/blob/94c7796351de6a7dc3ec6740e60a355719c37ecd/linkerd/app/outbound/src/logical.rs#L111-L114

wdyt? that way, all the "skip logical and go directly to the endpoint" logic is encapsulated in two places (this, and the "no profile" layer).

linkerd/app/outbound/src/lib.rs
```diff
@@ -129,49 +129,111 @@ impl<S> Outbound<S> {
         P::Error: Send,
         I: io::AsyncRead + io::AsyncWrite + io::PeerAddr + std::fmt::Debug + Send + Unpin + 'static,
     {
+        let tcp_endpoint = self.clone().push_tcp_endpoint().into_inner();
+        let http_endpoint = self.clone().push_tcp_endpoint().into_inner();
```
olix0r, line 133: the `http_endpoint` doesn't actually `push_http_endpoint`? Shouldn't it? If not, do we really need to pass two of these to `into_server_with`?

Why is this necessary? There's a _lot_ of type boilerplate for this split. Why?

hawkw: This is for tests — it's in order to allow us to configure the proxy with two separate endpoint stacks, one which is used for HTTP and another which is used for TCP forwards. This allows us to e.g. fail the HTTP tests if a connection is treated as raw TCP when it shouldn't be. 

Previously, the tests did this by just duplicating the code from `Outbound::into_server`. I wanted to change this so that there wasn't this duplication that has to be kept in sync — in particular, it was very easy to add new layers or change the structure of the stack built in `into_server`, but neglect to change the tests. Some of the test code had already silently become out of sync and wasn't testing a representative proxy, so rather than update the duplicated stacks, I thought it was better to just change `into_server` so that the tests could use it.

PR #991, opened 2021-04-28
outbound: Make the Endpoint::logical_addr field optional

olix0r:
The `outbound::Endpoint::logical_addr` field must currently _always_ be
set, even when it just duplicates the endpoint's target address and
there is no "logical" address.

This change makes this field optional and changes its type to require a
`LogicalAddr` (as returned by a profile lookup).

As a result of this, the `authority` endpoint metrics label is now only
set when a logical address is present.


hawkw (approved):
lgtm!

PR #992, opened 2021-04-28
ingress: Require the l5d-dst-override header

olix0r:
The outbound ingress-mode proxy generally relies on the
`l5d-dst-override` header to be set, but it still tries to handle other
cases (mostly to satisfy type constraints we needed previously). This is
unnecessary.

This change modifies the ingress-mode outbound proxy to require that the
`l5d-dst-override` header be set to a named address (i.e., not an IP
address). When this header isn't set, requests are failed a descriptive
error is emitted.

Furthermore, the inbound proxy no longer honors the `l5d-dst-override`
header. It never should have supported this header, but this was likely
copied from the outbound router. The header is only intended for
ingresses.

### DIFF
--- a/linkerd/app/outbound/src/ingress.rs
+++ b/linkerd/app/outbound/src/ingress.rs
@@ -6,7 +6,7 @@ use linkerd_app_core::{
     proxy::api_resolve::Metadata,
     svc::{self, stack::Param},
     tls,
-    transport::{self, ClientAddr, OrigDstAddr, Remote, ServerAddr},
+    transport::{self, OrigDstAddr, Remote, ServerAddr},
     AddrMatch, Conditional, Error, NameAddr,
 };
 use std::convert::TryFrom;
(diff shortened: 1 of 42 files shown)


linkerd/app/outbound/src/ingress.rs
```diff
@@ -248,3 +239,7 @@ impl<B> svc::stack::RecognizeRoute<http::Request<B>> for TargetPerRequest {
 #[derive(Debug, Error)]
 #[error("ingress routing requires a service profile")]
 struct ProfileRequired;
+
+#[derive(Debug, Error)]
+#[error("ingress routing requires the l5d-dst-override header")]
+struct DstOverrideRequired;
```
hawkw, line 245: could maybe be worth having the error indicate whether the header was missing or malformed? but, that would take a somewhat larger change, so it's not a blocker.

hawkw (approved):

PR #993, opened 2021-04-28
Polish profile-based endpoint stack

olix0r:
This change renames `endpoint_override` to be `profile_endpoint`:
"override" implies some behavior is being superceded, but in this case
we're simply getting endpoint metadata in place of logical/concrete
metadata.

### DIFF
--- a/linkerd/app/outbound/src/endpoint.rs
+++ b/linkerd/app/outbound/src/endpoint.rs
@@ -1,6 +1,6 @@
 use crate::{
     http::SkipHttpDetection,
-    logical::{Concrete, Logical, LogicalAddr},
+    logical::{Concrete, Logical},
     tcp::opaque_transport,
     Accept, Outbound,
 };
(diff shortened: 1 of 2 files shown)


linkerd/app/outbound/src/endpoint.rs
```diff
@@ -68,64 +67,52 @@ impl<E> Outbound<E> {
...
-        O::Service:
-            svc::Service<R, Response = <E::Service as svc::Service<R>>::Response, Error = Error>,
-        <O::Service as svc::Service<R>>::Future: Send,
-        T: std::fmt::Debug,
+        N: svc::NewService<(Option<profiles::Receiver>, T), Service = NSvc> + Clone,
+        NSvc: svc::Service<Req, Error = Error>,
+        NSvc::Future: Send,
+        E: svc::NewService<ProfileEndpoint, Service = ESvc> + Clone,
+        ESvc: svc::Service<Req, Response = NSvc::Response, Error = Error>,
+        ESvc::Future: Send,
```
hawkw, line 87: FWIW, I didn't use separate type parameters for the service types because I noticed that sometimes `push_xxx` methods' target types can't be inferred, and reducing the number of type params means fewer `_, _, _,`s have to be written out at the callsite. This is absolutely not a big deal, so I'm happy to use this style if we prefer it — the `where` clause certainly looks nicer.

linkerd/app/outbound/src/endpoint.rs
```diff
@@ -68,64 +67,52 @@ impl<E> Outbound<E> {
...
+    /// stack) if one exists.
+    pub fn push_profile_endpoint<T, E, ESvc, NSvc, Req>(
         self,
-        ep_override: O,
+        endpoint: E,
     ) -> Outbound<
         impl svc::NewService<
                 (Option<profiles::Receiver>, T),
-                Service = OrOverride<E::Service, O::Service>,
+                Service = svc::stack::ResultService<svc::Either<NSvc, ESvc>>,
```
hawkw, line 78: unrelated thought: might be nice to export a `Switch` type alias for `ResultService<Either<A, B>>` in `svc`, which is a bit more concise than writing out the whole type but doesn't require each stack layer to define a custom alias...not a big deal.

linkerd/app/outbound/src/endpoint.rs
```diff
@@ -68,64 +67,52 @@ impl<E> Outbound<E> {
     }
 
     /// Pushes a layer that checks if a discovered service profile contains an
-    /// endpoint override, and forwards directly to that endpoint (bypassing the
-    /// current stack) if one exists.
-    pub fn push_endpoint_override<T, O, R>(
+    /// endpoint, and forwards directly to that endpoint (bypassing the current
+    /// stack) if one exists.
+    pub fn push_profile_endpoint<T, E, ESvc, NSvc, Req>(
```
hawkw, line 72: :+1: the "override" terminology was just my first thought when trying to describe this — it's probably not necessary. 

I do think it _might_ be nice to have a consistent language for layer functions that introduce a "branch", like this one and `push_unwrap_logical`, like `or_profile_endpoint` or something? but we can do that later.

hawkw: I think the "override" language might also show up in a test someplace — but I can fix that.

hawkw (approved):

PR #994, opened 2021-04-28
trace: explicitly disable ANSI terminal colors

hawkw:
Currently, the use of ANSI terminal color codes by `tracing-subscriber`
is controlled by a feature flag, which is only eenabled by
`dev-dependencies` when building tests. However, despite this, the
feature flag appears to still be enabled in release builds --- despite
`cargo tree -e features -e no-dev -i tracing-subscriber` indicating that
it isn't enabled by any non-dev dependency.

Since this results in mangled logs in Kubernetes (fluentd doesn't handle
ANSI escape codes), this branch changes the proxy's tracing setup to
explicitly disable them outside of tests.


olix0r (approved):
weird

PR #995, opened 2021-04-29
outbound: Do not support TCP-forwarding in ingress-mode

olix0r:
The ingress-mode proxy has some simplistic support for TCP forwarding;
but this is somewhat of a misfeature. Specifically, we don't do any
mTLS/discovery for these connections, so detection failures can manifest
as unexpected insecure connections.

For now, we should remove TCP-forwarding support from ingress-mode
proxies. This will help support further simplifications to the
ingress-mode proxy.

### DIFF
--- a/linkerd/app/outbound/src/ingress.rs
+++ b/linkerd/app/outbound/src/ingress.rs
@@ -129,10 +129,10 @@ impl<H> Outbound<H> {
         )
         .instrument(|a: &http::Accept| debug_span!("http", v = %a.protocol))
         .push(http::NewServeHttp::layer(h2_settings, rt.drain))
-        .push_map_target(http::Accept::from)
-        .push(svc::UnwrapOr::layer(
-            svc::Fail::<_, IngressHttpOnly>::default(),
-        ))
+        .push_request_filter(|(http, accept): (Option<http::Version>, _)| {
+            http.map(|h| http::Accept::from((h, accept)))
+                .ok_or(IngressHttpOnly)
+        })
         .push_cache(cache_max_idle_age)
         .push_map_target(detect::allow_timeout)
         .push(detect::NewDetectService::layer(
--- a/linkerd/app/outbound/src/ingress.rs
+++ b/linkerd/app/outbound/src/ingress.rs
@@ -42,10 +42,8 @@ struct IngressHttpOnly;
 impl<H> Outbound<H> {
     /// Routes HTTP requests according to the l5d-dst-override header.
     ///
-    /// Forwards TCP connections without discovery/routing (or mTLS).
-    ///
     /// This is only intended for Ingress configurations, where we assume all
-    /// outbound traffic is either HTTP or TLS'd by the ingress proxy.
+    /// outbound traffic is HTTP.
     pub fn into_ingress<T, I, HSvc, P>(
         self,
         profiles: P,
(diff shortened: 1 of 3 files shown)


linkerd/app/outbound/src/ingress.rs
```diff
@@ -1,45 +1,61 @@
...
+#[error("ingress-mode routing requires a service profile")]
+struct ProfileRequired;
+
+#[derive(Debug, Error)]
+#[error("ingress-mode routing requires the l5d-dst-override header")]
+struct DstOverrideRequired;
+
+#[derive(Debug, Default, Error)]
+#[error("ingress-mode routing is HTTP-only")]
+struct IngressHttpOnly;
```
hawkw, line 38: this stuff was just all moved from the bottom of the file, right? seems good to me

linkerd/app/outbound/src/ingress.rs
```diff
@@ -63,106 +84,77 @@ impl Outbound<()> {
...
+                .push(errors::layer())
+                .push(http_tracing::server(rt.span_sink, trace_labels()))
+                .push(http::BoxResponse::layer()),
+        )
+        .instrument(|a: &http::Accept| debug_span!("http", v = %a.protocol))
+        .push(http::NewServeHttp::layer(h2_settings, rt.drain))
+        .push_map_target(http::Accept::from)
+        .push(svc::UnwrapOr::layer(
+            svc::Fail::<_, IngressHttpOnly>::default(),
+        ))
```
hawkw, line 137: nit/TIOLI: is it necessary to use an `UnwrapOr` that produces a failing service here? seems like we could just use `push_request_filter`:
```suggestion
        .push_request_filter(|(version, accept)| { 
            Ok(version.ok_or(IngressHttpOnly)?, accept)
         })
```
not a big deal, just maybe slightly fewer moving parts.

linkerd/app/outbound/src/ingress.rs
```diff
@@ -1,45 +1,61 @@
...
+impl<H> Outbound<H> {
     /// Routes HTTP requests according to the l5d-dst-override header.
     ///
     /// Forwards TCP connections without discovery/routing (or mTLS).
     ///
     /// This is only intended for Ingress configurations, where we assume all
     /// outbound traffic is either HTTP or TLS'd by the ingress proxy.
-    pub fn to_ingress<T, I, N, NSvc, H, HSvc, P>(
-        &self,
+    pub fn into_ingress<T, I, HSvc, P>(
```
hawkw, line 49: nit/TIOLI: maybe this should be called `into_http_ingress`, or something? just to make it clearer that it is now http only.

linkerd/app/outbound/src/ingress.rs
```diff
@@ -1,45 +1,61 @@
...
+#[derive(Debug, Default, Error)]
+#[error("ingress-mode routing is HTTP-only")]
+struct IngressHttpOnly;
+
+// === impl Outbound ===
+
+impl<H> Outbound<H> {
     /// Routes HTTP requests according to the l5d-dst-override header.
     ///
     /// Forwards TCP connections without discovery/routing (or mTLS).
```
hawkw, line 45: This should be removed :)
```suggestion
```

hawkw (approved):
looks good to me overall, couple very minor nits. is there a corresponding docs change we ought to make to remove indications that ingress mode can route TCP traffic?

PR #996, opened 2021-04-29
Decouple tcp forward stack from Endpoint target

olix0r:
the TCP forward stack is target-agnostic. This change relaxes the hard
dependency on the `Endpoint` target type.


hawkw (approved):
lgtm!

PR #1002, opened 2021-05-10
outbound: Decouple endpoint & logical stacks

olix0r:
In #963 we began to split the outbound stack so that logical stack was
only built when a profile resolution was successful. However, this
incorrectly handled the case when a profile resolution included an
endpoint in its response.

This change introduces a `outbound::switch_logical` module that takes an
optional profile and builds either an endpoint stack or a logical stack.
Logical stacks always have a profile resolution and (named) logical
address. Endpoint stacks are built otherwise, optionally with metadata
from the profile resolution.

In order to test this--and to simplify/focus tests, in general--I've
moved tests from `outbound/src/{tcp,http}/tests.rs` into the modules
they test. Now each test more narrowly targets the stacks whose logic
they test, rather than composing an entire outbound stack. It was
becoming cumbersome to support mocking each component of the larger
stack. Now, we more intentionaly test targeted logic in each stack
(including the newly fixed endpoint/logical stack switching).

Supersedes #988


linkerd/app/outbound/src/http/detect.rs
```diff
@@ -22,18 +22,17 @@ impl<N> Outbound<N> {
...
-        NTgt: From<T> + 'static,
-        H: svc::NewService<HTgt, Service = HSvc> + Clone + Send + 'static,
+        H: svc::NewService<U, Service = HSvc> + Clone + Send + 'static,
         HSvc: svc::Service<http::Request<http::BoxBody>, Response = http::Response<http::BoxBody>>,
         HSvc: Clone + Send + Sync + Unpin + 'static,
         HSvc::Error: Into<Error>,
         HSvc::Future: Send,
-        HTgt: From<(http::Version, T)> + svc::Param<http::Version> + 'static,
-        T: Param<SkipHttpDetection> + Clone + Send + Sync + 'static,
+        T: Param<Option<Skip>> + Clone + Send + Sync + 'static,
```
olix0r, line 34: This change isn't strictly necessary. Can back it out if you'd prefer the other way. @hawkw

linkerd/app/outbound/src/tcp/logical.rs
```diff
@@ -127,3 +110,223 @@ where
...
+                }
+                msg => unreachable!("unexpected read: {}", msg),
+            }
+            assert!(resolved.only_configured(), "Resolution must be reused");
+            if seen0 && seen1 {
+                tracing::info!("Both endpoints observed after {} iters", i);
+                break;
+            }
+            if i % 10 == 0 {
+                tracing::debug!("iters={} ep0={} ep1={}", i, seen0, seen1);
```
hawkw, line 280: nit, doesn't matter:
```suggestion
                tracing::debug!(iters = i, ep0 = seen0, ep1 = seen1);
```

linkerd/app/outbound/src/tcp/logical.rs
```diff
@@ -127,3 +110,223 @@ where
         }
     }
 }
+
+#[allow(unused_imports)]
```
hawkw, line 114: can we clean up imports here before merging? presumably this was just a temporary development thing?

hawkw (commented):
overall this looks good, and i'm onboard with this direction, although i still think it could be good to have tests that exercise the entire outbound stack in addition.

i left some comments, mostly on minor nits. i tried to read through all the tests, but there's a lot of code here, so it would be nice to get more eyes on them...seems right as far as i can tell!

linkerd/app/outbound/src/http/detect.rs
```diff
@@ -8,10 +8,10 @@ use linkerd_app_core::{
 use tracing::debug_span;
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub struct SkipHttpDetection(pub bool);
+pub struct Skip;
```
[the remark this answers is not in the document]
olix0r: Not really. This is what i meant by the change not being needed. But, yeah, generally it's a slightly more compact encoding: we can use the presence of the value to signal the same meaning. I think I started changing this as I renamed `SkipHttpDetection` to `http::detect::Skip`

linkerd/app/outbound/src/discover.rs
```diff
@@ -71,19 +85,273 @@ impl<N> Outbound<N> {
...
+        assert_eq!(
+            handle.tracked_services(),
+            0,
+            "no services have been created yet"
+        );
+
+        // Instantiate a service from the stack so that it instantiates the tracked inner service.
+        //
+        // The discover stack's buffer does not drive profile resolution (or the inner service)
+        // until the service is called?! So we drive this all on a background ask that gets canceled
```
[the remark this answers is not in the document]
olix0r: Yeah... it's a little surprising. It would at least be good to understand it in more detail.

linkerd/app/outbound/src/discover.rs
```diff
@@ -316,8 +312,8 @@ mod tests {
             svc::mk(move |_: SensorIo<io::DuplexStream>| future::ok::<(), Error>(()))
         };
 
-        // Create a profile stack that uses the tracked inner stack, configured to drop cached
-        // service after `idle_timeout`.
+        // Create a profile stack that uses the tracked inner stack, configured to only do profile
```
hawkw, line 313: nit: configured to only do profile _what_?

hawkw (approved):
okay, lgtm!

linkerd/app/outbound/src/switch_logical.rs
```diff
@@ -0,0 +1,191 @@
+use crate::{endpoint::Endpoint, logical::Logical, tcp, transport::OrigDstAddr, Outbound};
+use linkerd_app_core::{io, profiles, svc, Error, Never};
+use std::fmt;
+
+impl<S> Outbound<S> {
+    /// Wraps an endpoint stack to switch to an alternate logical stack when an appropriate profile
+    /// is provided:
+    ///
+    /// - When a profile include endpoint information, it is used to build an endpoint stack;
```
kleimkuhler, line 9:
```suggested
    /// - When a profile includes endpoint information, it is used to build an endpoint stack;
```

kleimkuhler (approved):
This looks good 👍

PR #1003, opened 2021-05-12
prepare for Rust 1.50+ by boxing large futures

hawkw:
Version 1.50 of the Rust compiler introduced a regression
(https://github.com/rust-lang/rust/issues/84873) that results in the
compiler using extremely large amounts of memory (and eventually getting
OOM killed) when compiling code involving very large nested types. This
regression is triggered by a number of future types in the proxy.

This branch introduces several `BoxService` layers, primarily around
`Switch` layers, to reduce the size of future types, so that the proxy
can successfully be compiled on Rust 1.50+.

This branch does *not* update the toolchain version. The Rust 1.51
toolchain also introduces some new clippy lints, which trigger on some
code patterns that are very common in the current proxy. Therefore, the
diff from a compiler update will be much larger than just adding
additional boxing. I'll land the compiler update in a separate branch
after this merges.

### DIFF
--- a/linkerd/app/gateway/src/lib.rs
+++ b/linkerd/app/gateway/src/lib.rs
@@ -117,6 +117,7 @@ where
         .push_tcp_endpoint()
         .push_tcp_logical(resolve.clone())
         .into_stack()
+        // .push_on_response(svc::BoxService::layer())
         .push_request_filter(
             |(p, _): (Option<profiles::Receiver>, _)| -> Result<_, Error> {
                 let profile = p.ok_or_else(discovery_rejected)?;
@@ -229,6 +230,8 @@ where
         )
         .push_http_server()
         .into_stack()
+        .push_on_response(svc::BoxService::layer())
+        .push(svc::BoxNewService::layer())
         .push_switch(
             |GatewayTransportHeader {
                  target,
@@ -254,6 +257,8 @@ where
             },
             legacy_http,
         )
+        .push_on_response(svc::BoxService::layer())
+        .push(svc::BoxNewService::layer())
         .into_inner()
 }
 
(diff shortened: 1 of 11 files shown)


linkerd/app/gateway/src/lib.rs
```diff
@@ -117,6 +117,7 @@ where
         .push_tcp_endpoint()
         .push_tcp_logical(resolve.clone())
         .into_stack()
+        // .push_on_response(svc::BoxService::layer())
```
hawkw, line 120:
```suggestion
```

linkerd/app/outbound/src/http/endpoint.rs
```diff
@@ -70,7 +73,9 @@ impl<C> Outbound<C> {
                 "host",
                 CANONICAL_DST_HEADER,
             ]))
-            .push_on_response(http::BoxResponse::layer());
+            .push_on_response(http::BoxResponse::layer())
+            // .push_on_response(svc::BoxService::layer())
+            ;
```
hawkw, line 78:
```suggestion
            .push_on_response(http::BoxResponse::layer());
```

linkerd/app/inbound/src/lib.rs
```diff
@@ -242,13 +243,16 @@ where
...
+                    .push_on_response(svc::BoxService::layer())
+                    .push(svc::BoxNewService::layer())
                     .into_inner(),
             ))
             .push_map_target(detect::allow_timeout)
             .push(detect::NewDetectService::layer(
                 config.detect_protocol_timeout,
                 http::DetectHttp::default(),
             ))
+            // .push_on_response(svc::BoxService::layer())
```
hawkw, line 255:
```suggestion
```

olix0r (approved):

PR #1004, opened 2021-05-12
http: Remove `identity_from_header` helper

olix0r:
The `proxy::http::identity_from_header` helper is only used in one place
and doesn't need to be exposed by the proxy::http module. Similarly the
`L5D_REQUIRE_ID` constant does not need to be used outside of the
outbound HTTP endpoint stack, though it is defined in `app::core`.

This change renames the `require_identity_on_endpoint` to
`require_id_header` and consolidates all related logic & constants into
the module. The module is now responsible for clearing the header as
well (rather than pairing the module with a generic strip-header layer).

### DIFF
--- a/linkerd/app/outbound/src/http/require_id_header.rs
+++ b/linkerd/app/outbound/src/http/require_id_header.rs
@@ -1,13 +1,9 @@
-use futures::{
-    future::{self, Either},
-    TryFutureExt,
-};
-use linkerd_app_core::{
-    errors::IdentityRequired, proxy::http::identity_from_header, svc, tls, Conditional, Error,
-    L5D_REQUIRE_ID,
-};
+use futures::{future, TryFutureExt};
+use linkerd_app_core::{errors::IdentityRequired, identity, svc, tls, Conditional, Error};
 use std::task::{Context, Poll};
-use tracing::debug;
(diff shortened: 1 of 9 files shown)


linkerd/app/outbound/src/http/require_id_header.rs
```diff
@@ -65,32 +69,41 @@ where
...
                             found: Some(server_id.clone()),
                         };
-                        return Either::Left(future::err(e.into()));
+                        return future::Either::Left(future::err(e.into()));
+                    } else {
+                        trace!(id = %require_id, "Identity required by header");
                     }
                 }
                 Conditional::None(_) => {
+                    debug!(id = %require_id, "Identity required by header not satisfied");
```
hawkw, line 97: nit: should these be
```suggestion
                    } else {
                        trace!(required = %require_id, "Identity required by header");
                    }
                }
                Conditional::None(_) => {
                    debug!(required = %require_id, "Identity required by header not satisfied");
```
for consistency with the other debug event?

hawkw, line 97: also, it could be nice if this event indicated that there was no server identity (and maybe included the reason)? nbd though

hawkw (approved):
lgtm!

PR #1007, opened 2021-05-13
ci: Use a matrix of fuzzer targets

olix0r:
The fuzzer build CI takes >2x as long as any other CI task.

This change moves the fuzzer building CI task to use a matrix of targets
so that they may be built in parallel. Furthermore, fuzzers are now
excluded from the check task.


hawkw (approved):
cool!

PR #1010, opened 2021-05-13
stack: make `Either` services always return boxed futures

hawkw:
In PR #1005, we added `BoxService` layers before most `push_switch`
layers, to erase large nested future types to reduce rustc memory use.
This helped, but it requires manually adding the `BoxService` layers
around switches.

This branch changes the `Either` implementation from
`tower::util::Either`, whose `Service` impl returns an enum response
future, to a `linkerd_stack::Either` type, whose service impl just
always returns a boxed response future. This way, we don't need to
remember to manually add the boxing before the switch.

Hopefully when the compiler regression that causes it to get OOM killed
gets sorted out, we can replace this with the upstream implementation
again...

### DIFF
--- a/linkerd/app/core/src/svc.rs
+++ b/linkerd/app/core/src/svc.rs
@@ -4,8 +4,8 @@ pub use crate::proxy::http;
 use crate::{cache, Error};
 pub use linkerd_concurrency_limit::ConcurrencyLimit;
 pub use linkerd_stack::{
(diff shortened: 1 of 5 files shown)


olix0r:
With this change, test builds are 2x as long and get OOM killed ;(

hawkw:
> With this change, test builds are 2x as long and get OOM killed ;(

hmm, i guess erasing the service types is also load-bearing...that is, at least, good to know!

PR #1011, opened 2021-05-14
Box all stack modules

olix0r:
Building a release-mode proxy currently consumes significant resources:

        Elapsed (wall clock) time (h:mm:ss or m:ss): 4:23.84
        Maximum resident set size (kbytes): 23751132

With this, change we reduce memory consumption by >50%:

        Elapsed (wall clock) time (h:mm:ss or m:ss): 2:18.08
        Maximum resident set size (kbytes): 10144624

This is accomplished by boxing each top-level module. This is made
explicit by modifying the signature of these stack builders to
explicitly reference boxed types (instead of using `impl`).

Furthermore, we add a `BoxNewService` wherever the inner new service is
cloned into a returned `Service`. This most commonly happens with layers
like `http::NewDetectService`` `NewDetectTls`, and `NewRouter`.


olix0r:
Supersedes #1009 -- this check is useful adding with this change (because it passes!); but I've split out several non-critical checks into a separate workflow.

.github/workflows/advisory.yml
```diff
@@ -0,0 +1,54 @@
...
+  # needed for release builds. This job builds the proxy in release-mode to ensure the build isn't
+  # OOM-killed.
+  release-binary:
+    timeout-minutes: 15
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: make build
+        env:
+          CARGO_RELEASE: true
```
hawkw, line 54: <3

linkerd/app/outbound/src/switch_logical.rs
```diff
@@ -37,43 +32,46 @@ impl<S> Outbound<S> {
...
+                    // endpoint from the original destination address.
+                    Ok(svc::Either::A(Endpoint::forward(
+                        target.param(),
+                        no_tls_reason,
+                    )))
+                },
+                logical,
+            )
+            .push_on_response(svc::BoxService::layer())
+            .push(svc::BoxNewService::layer());
```
hawkw, line 74: weird that adding chained method calls here caused rustfmt to reformat all the code in the closure, which hasn't chained...but whatever.

hawkw, line 74: also, i was wondering if it would make more sense to put the boxing _before_ the switch here, because boxing after the switch would mean that the `Either` future type still contains both branches of the switch, even if it's immediately erased. but, since this is a method on `Outbound`, and _all_ the other such methods now box, i guess we can guarantee that both stacks are already boxed before calling this...

hawkw (approved):
lgtm :package:

hawkw:
This should probably also close #1010.

PR #1013, opened 2021-05-19
Make span name without query string

aatarasoff:
Query string in Span name makes additional noise to Jaeger operation filter as on picture below.

Screenshot 2021-05-19 at 16 15 26

Suggest to use path only for span name but still use path and query in `http.path` property.


aatarasoff:
fix formatting

olix0r (approved):
thank you!

PR #1015, opened 2021-05-20
so it turns out that everything is always Bytes

hawkw:
Currently, the `http_box::BoxBody` type erases both the type of the
`http_body::Body` _and_ the `Body::Data` associated type (to
`Box<dyn Buf + Send>`). However, it turns out that erasing the `Data`
type isn't actually necessary: _all_ `Body` types in Linkerd use
`Bytes` as the `Data` type.

This branch changes `BoxBody` to just require that `B::Data` is `Bytes`,
and always produce `Bytes` as its data type. This will enable a much
more performant implementation of request body buffering for POST
retries (linkerd/linkerd2#6130). Rather than copying all the bytes from
a `dyn Buf` `Data` type into an additional buffer, we can just clone the
`Bytes`, increasing the reference count for that buffer, and return the
original `Bytes`. This means we don't need to perform additional memcpys
to buffer bodies for retries; we just hold onto the already-allocated
`Bytes` buffers.

The cloned `Bytes` can then be stored in a vector, and if we have to
write them out again for a retry, we can do one big vectored write :)

image

### DIFF
--- a/linkerd/app/outbound/src/http/logical.rs
+++ b/linkerd/app/outbound/src/http/logical.rs
@@ -23,8 +23,8 @@ impl<E> Outbound<E> {
         >,
     >
     where
-        B: http::HttpBody<Error = Error> + std::fmt::Debug + Default + Send + 'static,
-        B::Data: Send + 'static,
+        B: http::HttpBody<Data = bytes::Bytes, Error = Error>,
+        B: std::fmt::Debug + Default + Send + 'static,
         E: svc::NewService<Endpoint, Service = ESvc> + Clone + Send + Sync + 'static,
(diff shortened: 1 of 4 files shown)


hawkw:
Actually, upon further consideration, I think we _don't_ want to do this. Making all the buffers `Bytes` means we _can't_ do the "one big vectored write thing", and would instead have to stream the body chunks out one `Bytes` at a time.

Instead we should use [`Buf::copy_to_bytes`](https://docs.rs/bytes/1.0.1/bytes/buf/trait.Buf.html#method.copy_to_bytes), which i had forgotten does a shallow clone ref bump when the `Buf` is a `Bytes`. 

nevermind lol

PR #1016, opened 2021-05-22
ingress: Restore original dst address routing

olix0r:
In f8cc918e, we started requiring the `l5d-dst-override` header. As
described in linkerd/linkerd2#6157, this breaks some ingress
configurations, especially those that may route to out-of-cluster
resources.

This change restores original-dst-addr routing for requests that do not
include the `l5d-dst-override` header.

Internally, this change centralizes the `DST_OVERRIDE_HEADER` and
related utilities to the ingress module, as no other modules should have
to know anything about it.

### DIFF
--- a/linkerd/app/core/src/lib.rs
+++ b/linkerd/app/core/src/lib.rs
@@ -52,7 +52,6 @@ pub mod transport;
 pub use self::addr_match::{AddrMatch, IpMatch, NameMatch};
 
 pub const CANONICAL_DST_HEADER: &str = "l5d-dst-canonical";
-pub const DST_OVERRIDE_HEADER: &str = "l5d-dst-override";
 
 const DEFAULT_PORT: u16 = 80;
 
@@ -63,17 +62,6 @@ pub struct ProxyRuntime {
     pub tap: proxy::tap::Registry,
     pub span_sink: http_tracing::OpenCensusSink,
     pub drain: drain::Watch,
-}
-
-pub fn http_request_l5d_override_dst_name_addr<B>(
-    req: &http::Request<B>,
-) -> Result<NameAddr, addr::Error> {
-    proxy::http::authority_from_header(req, DST_OVERRIDE_HEADER)
-        .ok_or_else(|| {
-            tracing::trace!("{} not in request headers", DST_OVERRIDE_HEADER);
-            addr::Error::InvalidHost
-        })
-        .and_then(|a| NameAddr::from_authority_with_default_port(&a, DEFAULT_PORT))
 }
 
 pub fn http_request_authority_addr<B>(req: &http::Request<B>) -> Result<Addr, addr::Error> {
--- a/linkerd/app/outbound/src/ingress.rs
+++ b/linkerd/app/outbound/src/ingress.rs
@@ -172,7 +172,10 @@ impl Outbound<svc::BoxNewHttp<http::Endpoint>> {
                     .into_inner(),
             )
             .push(svc::BoxNewService::layer())
-            // Obtain a new inner service for each request (fom the above cache).
+            // Obtain a new inner service for each request. Override stacks are cached, as they
+            // depend on discovery that should not be performed many times. Forwarding stacks are
+            // not cached explicitly, as there are no real resources we need to share across
+            // connections. This allows us to avoid buffering requests to these endpoints.
             .push(svc::NewRouter::layer(
                 |http::Accept { orig_dst, protocol }| {
                     move |req: &http::Request<_>| {
--- a/linkerd/app/outbound/src/ingress.rs
+++ b/linkerd/app/outbound/src/ingress.rs
@@ -1,14 +1,16 @@
 use crate::{http, stack_labels, tcp, trace_labels, Config, Outbound};
 use linkerd_app_core::{
     config::{ProxyConfig, ServerConfig},
-    detect, errors, http_request_l5d_override_dst_name_addr, http_tracing, io, profiles,
-    proxy::api_resolve::Metadata,
+    detect, errors, http_tracing, io, profiles,
+    proxy::{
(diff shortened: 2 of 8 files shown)


linkerd/app/outbound/src/ingress.rs
```diff
@@ -1,70 +1,79 @@
...
     version: http::Version,
 }
 
-#[derive(Clone)]
-struct TargetPerRequest(http::Accept);
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+enum Target {
+    Forward(OrigDstAddr),
+    Override(NameAddr),
+}
```
hawkw, line 30: :+1:

linkerd/app/outbound/src/ingress.rs
```diff
@@ -79,142 +88,143 @@ impl<H> Outbound<H> {
...
-impl TargetPerRequest {
-    fn accept(a: http::Accept) -> Self {
-        Self(a)
-    }
-}
-
-impl<B> svc::stack::RecognizeRoute<http::Request<B>> for TargetPerRequest {
-    type Key = Target;
+        // Route requests with destinations that can be discovered via the `l5d-dst-override` header
+        // through a logical stack and route requests without the `l5d-dst-override` header through
```
hawkw, line 94: unimportant nit
```suggestion
        // through a logical stack, and routes requests without the `l5d-dst-override` header through
```

linkerd/app/outbound/src/ingress.rs
```diff
@@ -79,142 +88,143 @@ impl<H> Outbound<H> {
...
-    fn recognize(&self, req: &http::Request<B>) -> Result<Self::Key, Error> {
-        let dst = http_request_l5d_override_dst_name_addr(req).map_err(|_| DstOverrideRequired)?;
-        Ok(Target {
-            dst,
-            version: self.0.protocol,
-        })
+                tracing::debug!(
+                    domains = %profile_domains,
+                    "Address not in a configured domain",
+                );
```
hawkw, line 144: it looks like this will be logged even if the target is a `Target::Forward`, which i don't think is correct --- should this be moved inside the `if let`? 

we could add a separate log line when it's a forward, like:
```rust
            if let Target::Override(dst) = h.target {
                if profile_domains.matches(dst.name()) {
                    return Ok(profiles::LookupAddr(dst.into()));
                }

                tracing::debug!(
                    domains = %profile_domains,
                    "Address not in a configured domain",
                );
            } else {
                tracing::trace!("No `{}` address; skipping discovery", DST_OVERRIDE_HEADER);
            }
```

linkerd/app/outbound/src/ingress.rs
```diff
@@ -79,142 +88,143 @@ impl<H> Outbound<H> {
...
+                Err(profiles::DiscoveryRejected::new(
+                    "not in configured ingress search addresses",
```
hawkw, line 146: nit: similarly, it looks like we will produce the same error message for skipping discovery on forwards without override-dst headers and skipping discovery on addresses out of our search domains? if this gets logged someplace that might be confusing --- maybe we should return errors with different messages here?

linkerd/app/outbound/src/ingress.rs
```diff
@@ -79,142 +88,143 @@ impl<H> Outbound<H> {
...
+            }))
+            // This service is buffered because it needs to initialize the profile resolution and a
+            // fail-fast is instrumented in case it becomes unavailable. When this service is in
+            // fail-fast, ensure that we drive the inner service to readiness even if new requests
+            // aren't received.
+            .push_on_response(
+                svc::layers()
+                    .push(rt.metrics.stack.layer(stack_labels("http", "logical")))
+                    .push(svc::layer::mk(svc::SpawnReady::new))
+                    .push(svc::FailFast::layer("HTTP Logical", dispatch_timeout))
```
hawkw, line 157: nit: if i'm reading this correctly it looks like we will produce stack metrics for "HTTP logical" even on forwards --- shouldn't these stack metrics be on the logical stack inside of the switch, instead?

linkerd/app/outbound/src/ingress.rs
```diff
@@ -79,142 +88,143 @@ impl<H> Outbound<H> {
...
+            .push_cache(cache_max_idle_age)
+            .push_on_response(
+                svc::layers()
+                    .push(http::strip_header::request::layer(DST_OVERRIDE_HEADER))
+                    .push(http::Retain::layer()),
+            )
+            .instrument(|h: &Http| match h.target {
+                Target::Forward(_) => info_span!("forward"),
+                Target::Override(ref dst) => info_span!("override", %dst),
+            })
```
hawkw, line 169: :+1:

linkerd/app/core/src/lib.rs
```diff
@@ -52,7 +52,6 @@ pub mod transport;
 pub use self::addr_match::{AddrMatch, IpMatch, NameMatch};
 
 pub const CANONICAL_DST_HEADER: &str = "l5d-dst-canonical";
-pub const DST_OVERRIDE_HEADER: &str = "l5d-dst-override";
 
 const DEFAULT_PORT: u16 = 80;
```
hawkw, line 56: looks like this constant is no longer used --- should we remove it?
```suggestion
```

hawkw (approved):
overall, lgtm --- i commented on some nits mostly related to logging, but no real blockers!

linkerd/app/core/src/lib.rs
olix0r: it's used below

olix0r:
@hawkw PTAL, moved some things around

linkerd/app/outbound/src/ingress.rs
olix0r: that conjugation doesn't work here. In any case, updated the comment.

PR #1017, opened 2021-05-24
retry: retry requests with <=64KB bodies

hawkw:
Linkerd's proxy does not currently support retries on requests with
payloads. This precludes Linkerd from retrying gRPC requests, which is a
substantial limitation.

This PR adds support for retries on requests with bodies, if and only if
the request has a `Content-length` header and the content length is <=
64 KB.

In order to retry requests with payloads, we need to buffer the body
data for that request so that it can be sent again if a retry is
necessary. This is implemented by wrapping profile requests in a new
`Body` type which lazily buffers each chunk of data polled from the
inner `Body`. The buffered data is shared with a clone of the request,
and when the original body is dropped, ownership of the buffered data is
transferred to the clone. When the cloned request is sent, polling its
body will yield the buffered data.

If the server returns an error _before_ an entire streaming body has
been read, the replay body will continue reading from the initial
request body after playing back the buffered portion.

Data is buffered by calling `Buf::copy_to_bytes` on each chunk of data.
Although we call this method on an arbitrary `Buf` type, all data chunks
in the proxy are actually `Bytes`, and their `copy_to_bytes` methods are
therefore a cheap reference-count bump on the `Bytes`. After calling
`copy_to_bytes`, we can clone the returned `Bytes` and store it in a
vector. This allows us to buffer the body without actually copying the
bytes --- we just increase the reference count on the original buffer.

This buffering strategy also has the advantage of allowing us to write
out the entire buffered body in one big `writev` call. Because we store
the buffered body as a list of distinct buffers for each chunk, we can
expand the buffered body to a large number of scatter-gather buffers in
`Buf::bytes_vectored`. This should make replaying the body more
efficient, as we don't have to make a separate `write` call for each
chunk.

I've also added several tests for the new buffering body. In particular,
there are tests for a number of potential edge cases, including:

- when a retry is started before the entire initial body has been read
  (i.e. if the server returns an error before the request completes),
- when a retry body is cloned multiple times, including if the client's
  body has not yet completed,
- dropping clones prior to completion

Closes linkerd/linkerd2#6130.


linkerd/app/core/src/retry.rs
```diff
@@ -128,10 +132,60 @@ impl<B: Default + HttpBody> CloneRequest<http::Request<B>> for () {
...
+        // Requests without bodies can always be retried, as we will not need to
+        // buffer the body. If the request *does* have a body, retry it if and
+        // only if:
+        // - the request is a POST
+        // - the request contains a `content-length` header,
+        // - the content length is >= 64 kb,
+        let has_body = !req.body().is_end_stream();
+        let method = req.method();
+        if has_body
+            && (method != http::Method::POST
```
olix0r, line 165: Do we actually care about the method? Why not support PUTs etc as well? We should really just let this configuration come from profiles

hawkw: linkerd/linkerd2#6130 mentions POSTs by name, so i (incorrectly) assumed that we didn't want to retry any other methods. i'll change it.

hawkw: incidentally, we could certainly also make the maximum content-length to buffer be configurable as well, but i thought that seemed like a follow-up --- we could either configure it on the profile or globally with an annotation?

olix0r: > incidentally, we could certainly also make the maximum content-length to buffer be configurable as well, but i thought that seemed like a follow-up --- we could either configure it on the profile or globally with an annotation?
 
Yeah, I think it can be configurable. It seems like a process-specific thing more than a service-specific setting, so I'd probably start with an environment configuration.

hawkw: removed the POST requirement in 25e996e95caba4ac277da61ac8cc1f310c986f4b; we can make the max buffer size configurable in a follow-up?

linkerd/retry/src/buf_body.rs
```diff
@@ -0,0 +1,689 @@
...
+
+    /// If this is the initial body, returns a `Some` with a replay body.
+    pub fn try_clone(&self) -> Option<Self> {
+        match self.inner {
+            Inner::Initial(InitialBody { ref shared, .. }) => Some(Self {
+                inner: Inner::Replay(ReplayBody::Waiting(shared.clone())),
+            }),
+            _ => None,
+        }
+    }
```
olix0r, line 110: I don't think I've spotted anything that explicitly addresses this in the above comments: does this only support a single retry for a request with a body? If so, why does this limitation exist? What's required to support an arbitrary number of retries?

linkerd/retry/Cargo.toml
```diff
@@ -12,3 +12,11 @@ linkerd-stack = { path = "../stack" }
 tower = { version = "0.4.7", default-features = false, features = ["retry", "util"] }
 tracing = "0.1.23"
 pin-project = "1"
+bytes = "1"
+http-body = "0.4"
+http = "0.2"
+futures = { version = "0.3", default-features = false }
```
olix0r, line 18: preferably alpha-sort dependencies

linkerd/retry/src/buf_body.rs
hawkw: ah, hmm, the main thing is just that the `VecDeque` of buffers is moved by value into and out of the replay body, so that we don't have to clone it. if we wanted to support cloning the body multiple times, we would need to clone it so that it can be (potentially) used by a subsequent retry...

hawkw: i'm going to go ahead and change this so that the body can be cloned again

linkerd/retry/src/replay.rs
```diff
@@ -0,0 +1,768 @@
+#![allow(dead_code)]
```
olix0r, line 1: ?

linkerd/app/core/src/retry.rs
```diff
@@ -128,10 +132,54 @@ impl<B: Default + HttpBody> CloneRequest<http::Request<B>> for () {
         *clone.headers_mut() = req.headers().clone();
         *clone.version_mut() = req.version();
 
-        // // Count retries toward the request's total handle time.
-        // if let Some(ext) = req.extensions().get::<handle_time::Tracker>() {
-        //     clone.extensions_mut().insert(ext.clone());
-        // }
+        Some(clone)
+    }
+}
```
olix0r, line 137: Is this old impl still used anywhere?

linkerd/retry/src/replay.rs
```diff
@@ -0,0 +1,766 @@
...
+}
+
+/// A `Service`/`Proxy` that wraps an HTTP request's `Body` type in a `ReplayBody`,
+/// allowing it to be cloned.
+///
+// TODO(eliza): it would be nice if this could just be `MapRequest`, but Tower
+// won't let us implement `Proxy` for it (because it doesn't expose access to
+// the closure). Maybe we can change that upstream eventually...
+#[derive(Debug, Default, Clone)]
+pub struct WrapBody<S> {
```
olix0r, line 61: Can we use MapTarget, though? I don't think there's a proxy impl but would be easy to add...

hawkw: we could probably change it to use `MapTarget`... but this is removed in the follow-up PR #1020. happy to change it here first though if you think it's important?

linkerd/retry/src/replay.rs
```diff
@@ -0,0 +1,766 @@
...
+
+        Poll::Ready(Ok(None))
+    }
+
+    fn is_end_stream(&self) -> bool {
+        self.state
+            .as_ref()
+            .and_then(|state| state.rest.as_ref().map(Body::is_end_stream))
+            .unwrap_or(false)
+    }
```
olix0r, line 195: If I understand correctly this will always return `false` unless `poll_data` is called first, and so checks like that in clone_request may not work. What if the underlying body is empty? I suppose it's not a _bug_ for this return false by default; but it also seems like we could preserve the initial value... not sure.

In any case, maybe worth a comment explaining the thought process and/or test.

linkerd/retry/src/replay.rs
```diff
@@ -0,0 +1,766 @@
...
+                is_completed: false,
+            }),
+            shared: Arc::new(Mutex::new(None)),
+            // The initial `ReplayBody` has nothing to replay
+            replay_body: false,
+            replay_trailers: false,
+        }
+    }
+
+    fn state<'a>(
```
olix0r, line 92: perhaps `acquire_state` would be a little more descriptive? or `get_or_insert_from`? my concern is that `state` doesn't necessarily make it obvious that this takes a value out of the shared lock.

linkerd/retry/src/replay.rs
olix0r: Oh! It seems like we always start with a `state` when we construct the first body, and only lazily acquire it in clones;  It's probably fine, then.

linkerd/retry/src/replay.rs
```diff
@@ -0,0 +1,766 @@
...
+    }
+
+    fn is_end_stream(&self) -> bool {
+        self.state
+            .as_ref()
+            .and_then(|state| state.rest.as_ref().map(Body::is_end_stream))
+            .unwrap_or(false)
+    }
+
+    fn size_hint(&self) -> http_body::SizeHint {
```
olix0r, line 197: maybe worth tests? low priority, though.

linkerd/retry/src/replay.rs
```diff
@@ -0,0 +1,766 @@
...
+        if state.is_completed {
+            return Poll::Ready(None);
+        }
+
+        // If there's more data in the initial body, poll that...
+        if let Some(rest) = state.rest.as_mut() {
+            tracing::trace!("Polling initial body");
+            let opt = futures::ready!(Pin::new(rest).poll_data(cx));
+
+            // If the body
```
olix0r, line 141: ... ?

linkerd/retry/src/replay.rs
olix0r: nah that's fine

linkerd/retry/src/replay.rs
hawkw: > Oh! It seems like we always start with a `state` when we construct the first body, and only lazily acquire it in clones; It's probably fine, then.

Yeah, that's right...I can add comments explaining this a little better?

hawkw: Ah, I wrote a test for this and it is, actually, wrong --- since shared state isn't _acquired_ here, we do return `false` until the clone is polled for the first time. I'll fix that.

hawkw: ...although, actually, this logic is _quite_ wrong, since it will also return `true` if the actual body has been polled to completion but a given clone hasn't been polled yet. oops!

linkerd/retry/src/replay.rs
```diff
@@ -192,10 +208,23 @@ impl<B: Body + Unpin> Body for ReplayBody<B> {
...
+            return true;
+        }
+
+        // if this body has data or trailers remaining to play back, it
+        // is not EOS
+        !self.replay_body
+            && !self.replay_trailers
+            // or, if we have replayed everything, the initial body may
+            // still have data remaining, so ask it
+            && self
```
olix0r, line 223: the comment says "or" but the condition is `&&`... won't this always check the initial body? is that necessary/correct?

if so, perhaps extract the lower inner-body check into a temp variable so the condition reads more simply as

```rust
let inner_is_end = ...;

!self.replay_body && !self.replay_trailers && inner_is_end
```

olix0r (approved):

linkerd/retry/src/replay.rs
hawkw: > won't this always check the initial body? is that necessary/correct?

yes, we may have replayed all of the buffered data from a body that still has data remaining, and we can't know for certain unless we have checked the initial body

> if so, perhaps extract the lower inner-body check into a temp variable so the condition reads more simply as

sure, can do, i didn't do that because i thought it was better to allow it to short-circuit and avoid checking the inner body at all if we haven't replayed any data yet? this probably isn't hot enough that it's important for performance reasons, but i thought it was slightly better that way...

PR #1018, opened 2021-05-24
Forbid unsafe code in most module

olix0r:
This change adds the `forbid(unsafe_code)` directive to most modules,
ensuring that these modules are "safe". There are three modules this
cannot apply to:

- `duplex` -- due to the implementation of `BufMut` for `CopyBuf`; and
- `proxy::transport` -- due to socket option interactions.
- `system` -- for system-level counter access

Furthermore, this change adds `deny(warnings)` directives to a few
modules that were missing it.


hawkw (approved):
lgtm!

it _may_ be worth a top-level comment in the crates that require unsafe code explaining why they can't forbid unsafe? but, not a hard blocker.

PR #1020, opened 2021-05-27
retry: only buffer request bodies when retries are enabled

hawkw:
PR #1017 added support for retries on requests with bodies. This
requires buffering the body data so that it can be sent as part of a
retry request.

In #1017, all requests in the profile HTTP route stack have their bodies
wrapped in a `ReplayBody` type that buffers the request data. Although
the buffering strategy is lazy (the request can be sent before the
client finishes sending the body, so we don't introduce the latency of
waiting to buffer the entire body) and fairly lightweight (the buffered
body data is not copied an additional time, and is cloned only by
increasing the reference count on the buffer it was read into from the
wire), there is still some additional overhead that could be avoided
when retries are not configured on a route.

This branch updates the `Retry` service so that buffering is only added
on routes which have a retry policy. If there is no retry policy, we
avoid wrapping the request's body in a `ReplayBody`.

Initially, I tried to do this in a more modular way, by passing an
additional `Proxy` into the retry service that wraps the inner stack on
retries, but getting the types to line up correctly made this
impractical. Instead, I punted on doing this in a more modular way and
just added the body wrapping to the `Retry` proxy directly. This _does_
mean that it is now specific to HTTP requests and responses, which it
wasn't previously...but we only do retries for HTTP traffic anyway. We
can solve the problem of making it more modular if and when we ever need
to implement retries for other request-response protocols.

Depends on #1017


linkerd/retry/src/replay.rs
```diff
@@ -36,6 +36,12 @@ pub struct ReplayBody<B> {
     replay_trailers: bool,
 }
 
+#[derive(Debug)]
+pub enum MaybeReplay<B, R = ReplayBody<B>> {
+    NoReplay(B),
+    Replay(R),
+}
```
olix0r, line 43: Seems like this is really just a `EitherBody`?

linkerd/retry/src/lib.rs
```diff
@@ -93,35 +105,42 @@ where
...
+            );
             let retry = tower::retry::Retry::new(policy.clone(), inner);
-            return ResponseFuture::Retry(retry.oneshot(req));
+            return ResponseFuture::Retry(retry.oneshot(req.map(ReplayBody::new)));
         }
 
-        ResponseFuture::Disabled(self.inner.proxy(svc, req))
+        ResponseFuture::Disabled(
+            self.inner
+                .proxy(svc, req.map(replay::MaybeReplay::NoReplay)),
```
olix0r, line 135: And if this really just an `EitherBody`, can we just use `BoxBody` instead? Or do the `Data` types not line up?

olix0r (approved):
No blockers, just some questions about if it's possible to simply the body types.

Also, do we want to move this stuff into proxy/http now?

linkerd/retry/src/replay.rs
hawkw: ...i forgot we had that. we can probably get rid of this!

hawkw: > ...i forgot we had that. we can probably get rid of this!

Oh, I misinterpreted your comment; you were saying that this could just be a more generic `EitherBody` type, not that we already had one that we could use here. Yes, it could be --- the only reason it's not generic is that there's a default type param for the replay case, rather than hardcoding `ReplayBody<B>` in the replay case to allow using the same type for the data.

But...I'm going to see if this could just be a `BoxBody`---probably can.

hawkw:
@olix0r I've changed this to just use `BoxBody`, and renamed the `linkerd-retry` crate to `linkerd-http-retry` --- any last notes or shall I merge this when CI passes?

PR #1021, opened 2021-05-29
control: Build the identity client each time it's used

olix0r:
We currently build the identity client at startup and hold it to be used
fairly infrequently: the client is used once per day by default, and
even in extremely aggressive scenarios is unlikely to be used more
frequently than once every few minutes. As such, there's very little
benefit to holding the client (and buffers, DNS resolutions, etc) open
continually. Instead, it seems preferable to instantiate the identity
client only as it's needed.

Practically, we see issues like linkerd/linkerd2#6184 where the identity
client may try to reconnect to stale endpoints when the identity
deployment is rescheduled (because there aren't a steady stream of
requests on this client).

This change makes the controller stack a `NewService<()>` so that
clients can be instantiated lazily. The identity module now creates a
new connection for each identity request. Other controller clients are
unaffacted, continuing to use long-live clients.

### DIFF
--- a/linkerd/app/core/src/control.rs
+++ b/linkerd/app/core/src/control.rs
@@ -1,11 +1,6 @@
 use crate::{
-    classify, config, control, dns, metrics,
-    proxy::http,
-    reconnect,
-    svc::{self, NewService, Param},
-    tls,
-    transport::ConnectTcp,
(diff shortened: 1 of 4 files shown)


hawkw (approved):
looks good to me!

PR #1023, opened 2021-06-02
Fix MacOS conditional build in telemetry::process

rahulmutt:
On MacOS, the module `telemetry::process` in `linkerd-app-core` fails
to build because the `info!` macro is used on non-Linux systems, but
it is not imported.


hawkw (approved):

olix0r:
@rahulmutt Thanks for the fix! I pushed a small change to just inline the module reference since it's only used once. We'll merge this once CI passes :)

PR #1024, opened 2021-06-02
outbound: add env var for configuring retry length limit

hawkw:
PR #1017 added support for buffering request bodies for retries, with a
limit on the maximum content-length that can be buffered (to avoid
allocating a potentially unbounded amount of memory). However, that PR
hard-coded the maximum content-length to buffer at 64kb.

In some cases, users may wish to change the maximum content-length to
buffer for retries, either because they wish to retry longer bodies (and
have hopefully ensured that the proxy has enough memory for this), or
because they want to lower the limit to decrease memory use. Therefore,
this PR adds a new environment variable,
`LINKERD2_PROXY_OUTBOUND_MAX_RETRY_LENGTH`, to configure the maximum
content-length to buffer for retries.

To implement this, I also added a function in `linkerd_app::env` for
parsing data sizes given as KB, MB, or GB. This allows users to specify
values like `64kb`, rather than having to give all values in bytes.
While I was making this change, I also updated the config parser to
allow giving the HTTP/2 window size settings
(`LINKERD2_PROXY_HTTP2_INITIAL_{CONNECTION, STREAM}_WINDOW_SIZE`) with a
unit (rather than as a number of bytes), as well. This change should be
backwards-compatible, since when no unit is present, the value is still
parsed as a number of bytes, just like it was before.


linkerd/app/src/env.rs
```diff
@@ -108,6 +117,13 @@ const ENV_OUTBOUND_MAX_IDLE_CONNS_PER_ENDPOINT: &str =
 pub const ENV_INBOUND_MAX_IN_FLIGHT: &str = "LINKERD2_PROXY_INBOUND_MAX_IN_FLIGHT";
 pub const ENV_OUTBOUND_MAX_IN_FLIGHT: &str = "LINKERD2_PROXY_OUTBOUND_MAX_IN_FLIGHT";
 
+/// The maximum content length (in bytes) of request bodies that will be
+/// buffered for retries.
+///
+/// If a request body's `content-length` header is over this length, the request
+/// will not be retried. If unspecified, a default value (64k) is used.
+pub const ENV_OUTBOUND_MAX_RETRY_LENGTH: &str = "LINKERD2_PROXY_OUTBOUND_MAX_RETRY_LENGTH";
```
olix0r, line 125: wdyt about calling this `..._SIZE` to be consistent

hawkw: i went back and forth on this --- started with `length` because of the content-length header, but we can change it to `size`!

hawkw:
Do we still care about having this? I'm happy to rebase and update it if we do, but if we don't want to make this configurable, we should probably just close the PR...

olix0r:
I'm going to close this for now and we can revisit it if we find a use case for making it configurable.

PR #1025, opened 2021-06-02
tracing: Split HTML-formatting into admin module

olix0r:
The tracing crate is fairly core to all of our functionality, and so
it's preferable to avoid pulling in HTML-specific dependencies here. In
order to keep the tracing crate limited to only tracing-related
functionality, this change moves the HTTP handlers to the
`app::core::admin`, which includes the rest of the HTTP serving logic
used by the admin server.

There is no functional change.


hawkw (approved):
good call! :+1:

PR #1026, opened 2021-06-02
tracing: Simplify initialization

olix0r:
The `Settings::build` method is fairly large and complex. This change
splits into a few smaller units and reduces boilerplate--especially
around reading the settings from the environment.

This change is being made in service of #601 to help avoid making the
build method even more complicated.


hawkw (approved):
lgtm --- modularizing the JSON/plaintext formatter construction into separate methods seems like a good call!

PR #1027, opened 2021-06-03
Repace linkerd-drain with drain from crates.io

olix0r:
The linkerd-drain crate has been split out into its on repo at
https://github.com/linkerd/drain-rs and published to crates.io. This
change removes the `linkerd/drain` directory and replaces dependencies
with `drain v0.0.1`.


hawkw (approved):
nice, i had been thinking about this on and off for a while!

PR #1028, opened 2021-06-03
app: Move the admin server into a subcrate

olix0r:
The proxy's admin server is split across `linkerd/app` and
`linkerd/app/core/` -- with the former implementing the stack and the
latter implementing the HTTP server.

This change unifies these modules into `linkerd/app/admin`, which is no
longer needed to compile the inbound and outbound proxies. The admin
crate has a dependency on the inbound crate, as they currently share
target types.

Follow-up from #1025


linkerd/app/admin/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+The linkerd proxy's admin server.
+"""
+
+[dependencies]
+html-escape = "0.2"
+http = "0.2"
+hyper = { version = "0.14", features = ["http1", "http2"] }
+futures = "0.3"
+linkerd-app-core = { path = "../core" }
+linkerd-app-inbound = { path = "../inbound" }
```
hawkw, line 18: out of curiosity, do we want to consider factoring out the shared target types or adding new ones in the admin crate to remove this dependency? definitely makes sense to save that for a future PR; just wondering if it was planned...

hawkw (approved):
mostly just trivial moving code around, lgtm!

linkerd/app/admin/Cargo.toml
olix0r: For now I think this is okay. I _don't_ want to factor out target types -- I generally don't like having "general" target types, they should be pretty narrowly scoped. Treating the admin as another type of inbound is fine today but I think we'll eventually want separate target types (and metrics, etc). But I don't think those changes buy us much in the short term.

PR #1029, opened 2021-06-03
proxy-api: Update proxy-api to use the main branch

olix0r:
While we prepare for upcoming policy work, we should move the proxy to
use the updated crate from the main branch of the repo. This updated
version exposes a feature for each API--and for server- and
client-generation--so that we can more narrowly express dependencies in
the proxy.


hawkw (approved):
lgtm!

PR #1030, opened 2021-06-03
fuzz: Pull fuzzers into the top-level workspace

olix0r:
This avoids having to maintain many Cargo.lock files.


olix0r:
never mind. doesn't actually compile tests.

PR #1032, opened 2021-06-05
Implement reconnect as a NewService

olix0r:
In order to make the reconnect module more flexible--so that it can more
easily be used in different parts of the stack--this change modifies the
`reconnect` crate to be implemented in terms of `NewService` rather than
a `MakeService`.

Furthermore, tests are added to exercise reconnect.

This change sets up changes to reconnect behavior so reconnection is
only used when service discovery is responsible for evicting endpoints
from a balancer.


linkerd/app/core/src/control.rs
```diff
@@ -89,7 +84,12 @@ impl Config {
             .push(tls::Client::layer(identity))
             .push_timeout(self.connect.timeout)
             .push(self::client::layer())
-            .push(reconnect::layer(connect_backoff))
+            .push_on_response(svc::MapErrLayer::new(Into::into))
+            .into_new_service()
+            .push(svc::NewReconnect::layer({
+                let backoff = self.connect.backoff;
+                move |_| Ok(backoff.stream())
+            }))
```
hawkw, line 92: nit/TIOLI: it looks like every `NewReconnect` is called with a closure that takes the `backoff` and calls `stream()` on it...should the `layer` fn just take something implementing `Backoff`?

not a blocker either way.

linkerd/reconnect/src/lib.rs
```diff
@@ -3,34 +3,190 @@
...
+                        let backoff = self.recover.recover(error)?;
+                        debug!("Recovering");
+                        State::Disconnected {
+                            backoff: Some(backoff),
+                        }
+                    }
+                },
+
+                // If the service has not yet become ready, check its readiness and, if that fails,
+                // retain the prior backoff. Otherwise, mark the service as connected. We'll avoid
```
hawkw, line 121: it looks like this is a behavioral change from the previous version, which didn't reset the backoff when `poll_ready` errored? but, this seems right, since the service itself is probably a  `Pending`?

hawkw (approved):
lgtm

linkerd/reconnect/src/lib.rs
olix0r: Previously only `MakeService::poll_ready` wasn't handled with recovery. But now that has been eliminated by using `NewService`. In practice, that will be folded into service's `poll_ready` and handled with recovery. I don't think this matters for us.

PR #1033, opened 2021-06-06
Modify response to have `via` and `server` headers

aryan9600:
This PR addresses https://github.com/linkerd/linkerd2/issues/2295.
I am not very confident about this PR, hence the draft PR.
Also, a rather peculiar thing, checking the response via the viz dashboard doesn't show the `via` or `server` header, but the debug statements do confirm that the response has been modified. It'd be great if I could get some feedback on how to improve this.


olix0r:
@aryan9600 What do you think about omitting the `via` header for now? Users may not necessarily want to expose this information on requests so I'd like to be a bit careful about introducing that.

I think adding a `server` header to Linkerd-terminated requests seems like a good compromise, though; as this information is necessary to effectively debug 5xx responses.

linkerd/app/core/src/errors.rs
```diff
@@ -218,11 +234,13 @@ impl<RspB: Default + hyper::body::HttpBody> respond::Respond<http::Response<RspB
                 }
 
                 let status = http_status(&*error);
+                const SERVER_HEADER: &'static str = concat!("linkerd-proxy/", env!("CARGO_PKG_VERSION"));
```
hawkw, line 237: i don't think it's really correct to use the cargo package version as the proxy's version. this is going to be the version of the `linkerd-app-core` crate, which isn't the actual proxy version: https://github.com/aryan9600/linkerd2-proxy/blob/6438d0d19dafc22d6645d198b566085c7c5d0e73/linkerd/app/core/Cargo.toml#L3

instead, we should probably use the `GIT_VERSION` env var, which is the current proxy version git tag, set by the build script: https://github.com/linkerd/linkerd2-proxy/blob/122344732ec1ca60f81cc89ffe31e08a146cd8c9/linkerd/app/core/build.rs#L24-L27

```suggestion
                const SERVER_HEADER: &'static str = concat!("linkerd-proxy/", env!("GIT_VERSION"));
```

hawkw: the linting checks on CI [are failing](https://github.com/linkerd/linkerd2-proxy/pull/1033/checks?check_run_id=2766075149#step:5:363) because the `clippy` linter doesn't like the explicit `'static` lifetime here (constants always have the `'static` lifetime). to get CI passing, we should probably remove it:
```suggestion
                const SERVER_HEADER: &str = concat!("linkerd-proxy/", env!("CARGO_PKG_VERSION"));
```

hawkw (changes requested):
thanks for working on this! i agree with @olix0r's comments about the `via` header. also, i left a few suggestions about the server header inline.

linkerd/app/core/src/errors.rs
aryan9600: oops, forgot to consult clippy before pushing 😅

aryan9600:
> Users may not necessarily want to expose this information on requests so I'd like to be a bit careful about introducing that.

What if we keep it disabled by default and add an opt-in annotation to enable this? @olix0r @hawkw

PR #1034, opened 2021-06-07
Introduce the tonic-watch crate

olix0r:
The service-profiles client includes some complicated manual-polling
machinery for publish a `tokio::sync::watch::Receiver` from a streaming
gRPC endpoint.

This change moves this logic into a `linkerd-tonic-watch` crate that
is generic over the watch type and the underlying gRPC lookup
functionality. Furthermore, this is code is ported to use async/await
syntax to dramatically simplify state management.

This is being done in anticipation of upcoming inbound-discovery
infrastructure, where a similar watcher pattern will be used.


linkerd/tonic-watch/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+        let (tx1, rx1) = mpsc::channel(3);
+        let send_req = handle.next_request().map(move |req| {
+            let ((), rsp) = req.unwrap();
+            rsp.send_response(tonic::Response::new(Box::pin(ReceiverStream::new(rx1))))
+        });
+        let (_, _) = tokio::join!(tx1.send(Ok(345u16)), send_req);
+
+        // We need to give the background task an opportunity to process the update so this yields
+        // control. The actual sleep duration is unimportant (and time is mocked, anyway).
+        time::sleep(time::Duration::from_secs(1)).await;
```
hawkw, line 265: nit/tioli: if we don't actually need to advance the mocked timer, this could be 
```suggestion
        // Yield to give the background task an opportunity to process the update.
        tokio::task::yield_now().await;
```

linkerd/tonic-watch/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+    async fn block_on_initial_failure() {
+        let _trace = linkerd_tracing::test::trace_init();
+
+        time::pause();
+
+        let (mock, mut handle) = mock::pair::<(), InnerRsp<u16>>();
+        let map_err = MapErrLayer::new(|e: Error| match e.downcast_ref::<tonic::Status>() {
+            Some(status) => tonic::Status::new(status.code(), status.to_string()),
+            None => tonic::Status::internal(e.to_string()),
+        });
```
hawkw, line 280: nit/tioli: this is used in both tests, it could probably be stuck in a function? i don't really care though.

linkerd/tonic-watch/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+                // If there's a already a backoff, wait for it; but if the stream ends, then the
+                // newly-obtained backoff is used.
+                if b.next().await.is_none() {
+                    tracing::trace!("Old backoff exhausted; using new backoff");
+                    backoff = new_backoff.next().await.map(move |()| new_backoff);
+                }
+            } else {
+                tracing::trace!("Using new backoff");
+                backoff = new_backoff.next().await.map(move |()| new_backoff);
+            }
```
hawkw, line 117: i feel like this could maybe be simplified a bit? but it's not a big deal.

linkerd/tonic-watch/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+    where
+        T: Clone + Send + Sync + 'static,
+        U: Send + Sync + 'static,
+        S: Service<T, Response = InnerRsp<U>, Error = tonic::Status>,
+        S::Future: Send,
+    {
+        // Get an update and stream or return None.
+        let (init, rsp) = self.init(&target, None).await?;
+
+        // XXX Response::into_parts isn't public.
```
hawkw, line 54: is there an upstream issue for this?

hawkw (approved):
overall, this seems good --- i did some nit-picking in the comments.

linkerd/tonic-watch/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+        tx: watch::Sender<U>,
+        mut stream: InnerStream<U>,
+    ) where
+        T: Clone + Send + Sync + 'static,
+        U: Send + Sync + 'static,
+        S: Service<T, Response = InnerRsp<U>, Error = tonic::Status>,
+        S::Future: Send,
+    {
+        loop {
+            select_biased! {
```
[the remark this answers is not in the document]
olix0r: oh sweet!

linkerd/tonic-watch/src/lib.rs
```diff
@@ -0,0 +1,329 @@
...
+                },
+            }
+        }
+    }
+
+    /// Gets the next profile from the stream
+    ///
+    /// If the stream or lookup fails in a recoverable way, back-offs are applied and `stream` is
+    /// updated to point at the updated stream..
+    #[inline]
```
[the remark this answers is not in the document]
hawkw: oh, huh, looks like [`#[inline(always)], at least, might do something](https://github.com/rust-lang/rust/issues/63647)?

PR #1035, opened 2021-06-07
ingress: Add failfast to the forwarder

olix0r:
The ingress-mode proxy's forwarding stack--used when a request does not
set the `l5d-dst-override` header--has no failfast implemetation. This
means that when a connection can't be obtained for the endpoint,
requests are buffered indefinitely.

This change adds a failfast layer so that these requests are failed
eagerly after 3s of unavailability, causing the serverside connection to
be dropped (so that the application client may re-resolve the endpoint).

This is really a temporary solution. We should probably avoid
implementing reconnection at all in this case so that connection errors
can be used in place of failfast errors.

Related to linkerd/linkerd2#6184

### DIFF
--- a/linkerd/app/outbound/src/ingress.rs
+++ b/linkerd/app/outbound/src/ingress.rs
@@ -168,6 +168,11 @@ impl Outbound<svc::BoxNewHttp<http::Endpoint>> {
                     })),
                 },
                 http_endpoint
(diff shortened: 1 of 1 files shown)


hawkw (approved):
:+1: let's get this in as a stopgap solution!

PR #1036, opened 2021-06-07
outbound: don't double-wrap replay bodies

hawkw:
In PR #1020, the retry code was changed so that bodies are only wrapped
in a `ReplayBody` when the request will be retried. The intention was to
remove the `replay::layer` pushed to the stack before the retry layer,
so that bodies are only wrapped when retries are enabled for the route,
and not for all routes in a profile.

However, I somehow managed to not remove the `replay::layer` from the
stack (possibly due to a messed up rebase). Oops. This meant that all
bodies for endpoints with service profiles were wrapped, and when the
route has retries enabled, it's wrapped twice. When wrapped twice, this
means that we potentially buffer the body data *twice* --- which may mean
actually copying bytes, since replaying from the buffer would now yield
a `Data` type that isn't itself `Bytes`, breaking the `copy_to_bytes`
optimization. 

It also means that we would buffer body data for _all_ routes in
a `ServiceProfile`.

### DIFF
--- a/linkerd/app/outbound/src/http/logical.rs
+++ b/linkerd/app/outbound/src/http/logical.rs
@@ -122,7 +122,6 @@ impl<E> Outbound<E> {
                     )
                     // Sets an optional retry policy.
                     .push(retry::layer(rt.metrics.http_route_retry.clone()))
-                    .push_on_response(retry::replay::layer())
                     // Sets an optional request timeout.
                     .push(http::MakeTimeoutLayer::default())
                     // Records per-route metrics.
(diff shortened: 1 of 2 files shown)


olix0r:
Should we remove this helper altogether?

https://github.com/linkerd/linkerd2-proxy/blob/0e6df7c4c0b908399c3c9d4b0f4cab6409a3e591/linkerd/http-retry/src/replay.rs#L74-L77

Is there any use case for constructing replay bodies in a distinct layer?

hawkw:
> Should we remove this helper altogether?
> 
> https://github.com/linkerd/linkerd2-proxy/blob/0e6df7c4c0b908399c3c9d4b0f4cab6409a3e591/linkerd/http-retry/src/replay.rs#L74-L77
> 
> Is there any use case for constructing replay bodies in a distinct layer?

yup, forgot to get rid of that --- it's gone now!

olix0r (approved):

PR #1037, opened 2021-06-07
Update tokio, hyper, and socket2

olix0r:
This change updates `tokio`, `hyper`, and `socket2` (from v0.3 to v0.4).
This has a few implications:

* `tokio::select!` can now be used in place of
  `futures::select_biased!`;
* We now only depend on the core `futures` functionality with all
  features disabled.
* `socket2` had several small API changes that have been addressed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -356,9 +356,9 @@ dependencies = [
 
 [[package]]
 name = "h2"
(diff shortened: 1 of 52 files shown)


hawkw (approved):
lgtm --- hadn't realized our tokio version was that old!

PR #1038, opened 2021-06-09
service-profiles: Wrap receiver types

olix0r:
We should be able to cache profile resolutions independently of the
stack; but the fact that the the profile modue's `Receiver` type is just
an alias for `watch::Receiver<Profile>` means that we can't attach
the additional metadata needed for cache entry retention.

In anticipation of caching profiles by lookup address, this change
creates an explicit `Receiver` type that hides the underlying
`watch::Receiver`, exposing only methods used by other crates/modules.

This change also replaces the boxed watch stream with a
`tokio_stream::wrappers::WatchStream`--also wrapped by a
profile-specific type--to avoid needless allocation and support
retaining additional resources.

### DIFF
--- a/linkerd/service-profiles/src/lib.rs
+++ b/linkerd/service-profiles/src/lib.rs
@@ -2,7 +2,7 @@
 #![forbid(unsafe_code)]
 #![allow(clippy::inconsistent_struct_constructor)]
 
-use futures::stream::Stream;
+use futures::Stream;
 use linkerd_addr::{Addr, NameAddr};
 use linkerd_error::Error;
 use linkerd_proxy_api_resolve::Metadata;
@@ -15,6 +15,7 @@ use std::{
     task::{Context, Poll},
 };
 use thiserror::Error;
+use tokio::sync::watch;
 use tower::util::{Oneshot, ServiceExt};
 
 mod client;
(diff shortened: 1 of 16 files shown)


kleimkuhler (approved):
Looks good 👍

linkerd/service-profiles/src/lib.rs
```diff
@@ -26,7 +27,15 @@ pub mod split;
...
-pub type Receiver = tokio::sync::watch::Receiver<Profile>;
+#[derive(Clone, Debug)]
+pub struct Receiver {
+    inner: tokio::sync::watch::Receiver<Profile>,
+}
+
+#[derive(Debug)]
+struct ReceiverStream {
+    inner: tokio_stream::wrappers::WatchStream<Profile>,
+}
```
hawkw, line 38: does this wrapper do anything? AFAICT it could just be a type alias for `WatchStream<Profile>`...is the intent to hang more functionality off the stream type in the future?

not a blocker.

linkerd/service-profiles/src/lib.rs
```diff
@@ -113,20 +122,47 @@ where
...
+    fn from(Receiver { inner }: Receiver) -> Self {
+        let inner = tokio_stream::wrappers::WatchStream::new(inner);
+        ReceiverStream { inner }
+    }
+}
+
+impl Stream for ReceiverStream {
+    type Item = Profile;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
```
hawkw, line 163: nit/TIOLI: since this just trivially forwards to the inner stream's method, should maybe be
```suggestion
    #[inline]
    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
```

hawkw (approved):
couple tiny nits, none of which are particularly important

linkerd/service-profiles/src/lib.rs
olix0r: Yeah. For the same reason we have the `Receiver` wrapper -- we'll need to be able to hold a handle as long as the stream is active.

PR #1040, opened 2021-06-09
retry: Remove generic types from `Retry` policy

olix0r:
The core retry policy doesn't actually need to be generic: it's only
ever constructed with the `WithBody` type. This change simplifies the
retry types by removing unneeded generics.


hawkw (approved):
yeah, i was also thinking about getting rid of this, good call!

PR #1041, opened 2021-06-10
Add a 'rustfmt' feature to crates that use gRPC

olix0r:
Sometimes we need to be able to read generated code. The
linkerd2-proxy-api's 'rustfmt' feature flag makes this behavior
optional. This change adds this feature flag to the linkerd crates that
depend on linkerd2-proxy-api.


hawkw:
lovely

hawkw (approved):
whoops, i forgot to click "approve"

PR #1043, opened 2021-06-10
retry: enforce a cap on max replay buffer size

hawkw:
This branch adds a limit to the maximum number of bytes a `ReplayBody`
will buffer, as a defense-in-depth against unbounded buffering.
Currently, the retry policy should ensure that requests whose
`content-length` header is over the limit will not be wrapped with
`ReplayBody`. However, if there's a bug in the retry policy, or if a
request claims to have an acceptable `content-length` but keeps sending
data past that length, we might buffer a potentially unbounded amount of
data. This limit is intended to _ensure_ we don't buffer data over the
limit despite the `content-length`-based retry policy.

The limit is checked whenever new data is added to the buffer, and if it
is reached, the buffer is discarded and the `ReplayBody` will cease
buffering additional data. However, the body will be allowed to continue
streaming so that the _initial_ request may still complete --- but
attempting to retry it will fail, because the buffer was discarded. This
way, in the event of a request that somehow passes the limit, we won't
fail that request immediately, but instead will fall back to not
retrying it.

In order to make this change nicely, the `ReplayBody` has to be
constructed by the retry policy, rather than by the `Retry` service.
This is because the policy knows what the size limits for retries is. To
allow this, I did some refactoring. In particular, I introduced a new
trait for retry policies that's an extension of `tower::retry::Policy`,
but with the addition of a method that's called to modify the request
for a retry. This method returns an `Either`, with `Either::A`
indicating that the request should be retried, and `Either::B`
indicating that it should not; and in the retry case, the policy may
change the request's type. This allows the retry policy to be
responsible for wrapping the request in a `ReplayBody`, rather than
doing that in the `Retry` service. A nice side-effect of this change is
that the `linkerd-retry` crate no longer needs to be HTTP-specific. All
the HTTP-specific code now lives in `linkerd_app_core::retry`, with the
exception of the `ReplayBody` type, which I put in its own crate (mainly
just to allow running its tests without having to compiel all of
`app-core`).


linkerd/http-retry/src/lib.rs
```diff
@@ -3,160 +3,939 @@
...
+    }
+
+    fn copy_to_bytes(&mut self, len: usize) -> Bytes {
+        // If the length of the requested `Bytes` is <= the length of the front
+        // buffer, we can just use its `copy_to_bytes` implementation (which is
+        // just a reference count bump).
+        match self.bufs.front_mut() {
+            Some(first) if len <= first.remaining() => {
+                let buf = first.copy_to_bytes(len);
+                // if we consumed the first buffer, also advance our "cursor" by
```
olix0r, line 453:
```suggestion
                // If we consumed the first buffer, also advance our "cursor" by
```

linkerd/http-box/src/erase_request.rs
```diff
@@ -0,0 +1,77 @@
...
+use std::task::{Context, Poll};
+
+/// A middleware that boxes HTTP request bodies.
+///
+/// This is *very* similar to the [`BoxRequest`] middleware. However, that
+/// middleware is generic over a specific body type that is erased. A given
+/// instance of `EraseRequest` can only erase the type of one particular `Body`
+/// type, while this middleware will erase
+/// bodies of *any* type.
+///
```
olix0r, line 14:
```suggestion
/// instance of `EraseRequest` can only erase the type of one particular `Body`
/// type, while this middleware will erase bodies of *any* type.
///
```

linkerd/retry/src/lib.rs
```diff
@@ -0,0 +1,162 @@
+#![deny(warnings, rust_2018_idioms)]
+#![forbid(unsafe_code)]
+#![allow(clippy::inconsistent_struct_constructor)]
+#![allow(clippy::type_complexity)]
```
olix0r, line 4: Appears to be unneeded

```suggestion
#![allow(clippy::type_complexity)]
```

linkerd/http-box/src/erase_request.rs
```diff
@@ -0,0 +1,76 @@
+//! A middleware that boxes HTTP request bodies.
+use crate::BoxBody;
```
olix0r, line 2:
```suggestion

use crate::BoxBody;
```

linkerd/retry/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+use tracing::trace;
+
+/// A strategy for obtaining per-target retry polices.
+pub trait NewPolicy<T> {
+    type Policy;
+
+    fn new_policy(&self, target: &T) -> Option<Self::Policy>;
+}
+
+/// A layer that applies per-target retry polcies.
```
olix0r, line 21:
```suggestion
/// A layer that applies per-target retry policies.
```

olix0r:
Shared some suggestions via #1045

olix0r (approved):

hawkw:
yay!

PR #1045, opened 2021-06-14
RetryPolicy refactor suggestions

olix0r:
A few suggestions on #1043 

- Expose `EraseRequest` via `BoxRequest::erased` -- we don't actually care about the different types--the behavior is basically the same -- there's just some minor differences with regards to preserving marker types. It reads a little bit more naturally if we just hang it all off BoxRequest.
- Renamed the `linkerd_retry::RetryPolicy` trait to `linkerd_retry::PrepareRequest` (to match its only method). It's a bit... _malkovich_ to have both `tower::retry::Policy` and `linkerd_retry::RetryPolicy` which are closely related but different.
- `linkerd_retry` now re-exports relevant `tower::retry` types.
- Renames in `app::core::retry` to reflect actual types. For instance, `NewRetry` is now `NewRetryPolicy` -- since it actually implements `NewPolicy` and has nothing to do with the `retry::NewRetry` type.
- `NewRetryLayer` eliminated in favor of an anonymous type.
- Various `inlines` on pass-through service/proxy methods

### DIFF
--- a/linkerd/http-box/src/request.rs
+++ b/linkerd/http-box/src/request.rs
@@ -34,10 +34,12 @@ where
     type Error = S::Error;
     type Future = S::Future;
 
+    #[inline]
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
         self.0.poll_ready(cx)
     }
 
+    #[inline]
     fn call(&mut self, req: http::Request<B>) -> Self::Future {
         self.0.call(req.map(BoxBody::new))
     }
--- a/linkerd/http-box/src/request.rs
+++ b/linkerd/http-box/src/request.rs
@@ -1,6 +1,6 @@
 //! A middleware that boxes HTTP request bodies.
 
-use crate::BoxBody;
+use crate::{erase_request::EraseRequest, BoxBody};
 use linkerd_error::Error;
 use linkerd_stack::layer;
 use std::{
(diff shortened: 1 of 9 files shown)


linkerd/http-box/src/request.rs
```diff
@@ -17,6 +17,13 @@ impl<S, B> BoxRequest<S, B> {
     }
 }
 
+impl<S> BoxRequest<S, ()> {
+    /// Constructs a boxing layer that erases the inner request type with [`EraseRequest`].
+    pub fn erased() -> impl layer::Layer<S, Service = EraseRequest<S>> + Clone + Copy {
+        EraseRequest::layer()
+    }
```
hawkw, line 24: :+1:

linkerd/retry/src/lib.rs
```diff
@@ -18,29 +20,9 @@ pub trait NewPolicy<T> {
...
-#[derive(Clone, Debug)]
-pub struct Retry<P, S> {
-    policy: Option<P>,
-    inner: S,
-}
-
 /// An extension to [`tower::retry::Policy`] that adds a method to prepare a
 /// request to be retried, possibly changing its type.
-pub trait RetryPolicy<Req, Res, E>: tower::retry::Policy<Self::RetryRequest, Res, E> {
+pub trait PrepareRequest<Req, Res, E>: tower::retry::Policy<Self::RetryRequest, Res, E> {
```
hawkw, line 25: my original motivation for naming this `RetryPolicy` --- which i agree is a bit confusing --- was because it requires `tower`'s `RetryPolicy` trait, so when it's used, we only bound the type with _this_ trait. 

i also considered the name `PrepareRequest`, but it seemed kind of weird to me to require that the policy implement `PrepareRequest` and not mention that it is also a retry policy? maybe that's not actually worth worrying about, and i'm fine with this naming scheme if you think it's clearer...

hawkw (approved):
lgtm, i'm on board with all these suggestions! i did leave a comment on the trait naming thing explaining my original thought process, but i'm not attached to it at all --- this is probably better. other than that, nothing stood out, thanks!

linkerd/retry/src/lib.rs
olix0r: could call it `PreparePolicy` or something like that -- but RetryPolicy vs Policy is a bit too subtle imo

PR #1048, opened 2021-06-16
Remove unneeded uses of indexmap

olix0r:
We initially adopted `indexmap` in many places where it's not strictly
necessary: we don't actually rely on the data being insert-ordered and
don't need to mutate the structure while iterating over it. In fact, the
traffic split module is the only place in the codebase where we access
the data structure by index (rather than by key).

In all other cases, we do not care about insertion-order. In the vast
majority of cases we can simply use a `HashMap`. Endpoint and route
labels need to be reliably ordered so that they can be compared (i.e.
implementing `Eq`), in which case the standard `BTreeMap` works
sufficiently.

This change eliminates unnecessary uses of `indexmap` to reduce
dependency and conceptual overhead.


hawkw:
I was also thinking about doing this yesterday when I was looking at metrics stuff...I suspect that `HashMap` has actually outperformed `IndexMap` in a lot of cases since the switch to `hashbrown`.

linkerd/proxy/api-resolve/src/metadata.rs
```diff
@@ -1,12 +1,14 @@
 use http::uri::Authority;
-use indexmap::IndexMap;
 use linkerd_tls::client::ServerId;
+use std::collections::BTreeMap;
+
+pub type Labels = BTreeMap<String, String>;
```
hawkw, line 5: maybe worth a comment here that `BTreeMap` is used because we want these to be sorted?

linkerd/proxy/tap/src/lib.rs
```diff
@@ -26,6 +25,8 @@ pub fn new() -> (Registry, grpc::Server) {
     (registry, server)
 }
 
+pub type Labels = std::collections::BTreeMap<String, String>;
```
hawkw, line 28: similarly, maybe worth a comment about sorting here?

linkerd/service-profiles/src/http/mod.rs
```diff
@@ -67,7 +65,7 @@ impl Route {
         let labels = {
             let mut pairs = label_iter.collect::<Vec<_>>();
             pairs.sort_by(|(k0, _), (k1, _)| k0.cmp(k1));
-            Labels(Arc::new(IndexMap::from_iter(pairs)))
+            Labels(Arc::new(pairs.into_iter().collect()))
         };
```
hawkw, line 69: because we're now using `BTreeMap`, which is inherently sorted, we can skip the intermediary `Vec` here --- this can just be 
```rust
let labels = Arc::new(label_iter.collect());
```
(github won't let me leave a suggestion here because the diff contains deleted lines)

hawkw (changes requested):
overall, this looks good, but i noticed one place where we are still collecting an iterator of labels into a `Vec`, sorting it, and then collecting the `Vec` into a `BTreeMap`...because `BTreeMap`s are always sorted, we shouldn't have to do that.

once that's fixed, this should be good to merge!

hawkw (approved):
okay, i made all the minor changes i mentioned (and opened a separate follow-up PR for https://github.com/linkerd/linkerd2-proxy/pull/1048/files#r653855672, PR #1106). this looks good to merge now!

linkerd/app/inbound/src/require_identity.rs
```diff
@@ -1,14 +1,13 @@
...
-use std::sync::Arc;
+use std::{collections::HashSet, sync::Arc};
 use thiserror::Error;
 
 /// A connection policy that fails connections that don't have a client identity
 /// if they target one of the configured local ports.
 #[derive(Clone, Debug)]
 pub struct RequireIdentityForPorts {
-    ports: Arc<IndexSet<u16>>,
+    ports: Arc<HashSet<u16>>,
```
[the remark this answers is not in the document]
hawkw: see #1106

PR #1080, opened 2021-06-16
Update tonic

olix0r:
We updated tonic-build without updating tonic. There was a hidden API
breakage such that the old tonic version could not be used.

See https://github.com/hyperium/tonic/issues/683

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2274,9 +2274,9 @@ dependencies = [
 
 [[package]]
 name = "tonic"
-version = "0.4.0"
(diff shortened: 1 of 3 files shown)


hawkw (approved):
hmm, i wonder if there's a way to get dependabot to update them together (but also, there _shouldn't_ be breakage like this in minor versions, so it's probably not a big deal)

hawkw:
oh looks like a fuzz lockfile needs to be bumped (for http)

PR #1088, opened 2021-06-16
build(deps): Omit fuzz lockfiles from version control

olix0r:
Each fuzz directory has its own Cargo.lock file because these crates
cannot be a part of the the top-level workspace while cargo-fuzz requies
nightly features. But these lockfiles are frequently invalidated by
non-fuzzing dependency changes and are slow and painful to regenerate.

This change deletes these lockfiles from this repository and adds them
to .gitignore. The trade-off is that fuzzers may run against more recent
dependencies but, now that we're using dependabot, we don't expect much
version skew here.

We continue to validate that fuzzers build in CI; and fuzzer projects
remain under dependabot control so the Cargo.toml files are updated as
appropriate.


hawkw (approved):
:tada::tada::raised_hands:

PR #1106, opened 2021-06-17
config: add a `PortSet` type

hawkw:
This branch adds a `PortSet` type that's a simple wrapper around
`std::collections::HashSet` but specialized for storing ports (`u16`
values). The `PortSet` type overrides the default hasher so that we
don't have to actually hash ports, and just use their numeric value.

See here for details:
https://github.com/linkerd/linkerd2-proxy/pull/1048#discussion_r653855672

Depends on #1048

### DIFF
--- a/linkerd/app/core/src/config.rs
+++ b/linkerd/app/core/src/config.rs
@@ -7,7 +7,6 @@ use crate::{
 use std::{
     collections::HashSet,
     hash::{BuildHasherDefault, Hasher},
-    iter::FromIterator,
     time::Duration,
 };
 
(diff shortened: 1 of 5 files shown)


linkerd/app/core/src/config.rs
```diff
@@ -46,3 +63,21 @@ impl Param<Keepalive> for ServerConfig {
...
+
+impl Hasher for PortHasher {
+    fn write(&mut self, _: &[u8]) {
+        unreachable!("hashing a `u16` calls `write_u16`");
+    }
+
+    #[inline]
+    fn write_u16(&mut self, port: u16) {
+        self.0 = port as u64;
+    }
```
kleimkuhler, line 77: Any reason `PortHasher` has a `u64` meaning we have to cast for every write, instead of it having a `u16` and only casting to `u64` in `finish`?

kleimkuhler (approved):
Looks good! Have a question about something that really shouldn't be an issue but just wondering.

olix0r (commented):
wdyt about adding a quickcheck that roundtrips values through portset?

linkerd/app/core/src/config.rs
hawkw: an individual `PortHasher` will only ever see one `write_u16` call; it should be equivalent either way? but can do.

hawkw:
> wdyt about adding a quickcheck that roundtrips values through portset?

done!

olix0r (approved):

PR #1107, opened 2021-06-17
replace `linkerd-concurrency-limit` with Tower

hawkw:
tower-rs/tower#574 added a `GlobalConcurrencyLimitLayer` which has the
same behavior as our `linkerd-concurrency-limit` crate (sharing a single
semaphore across all instances of a service). This was published in
Tower 0.4.8, and now that we've updated to that version, we can remove
our implementation in favour of upstream.

This has the nice advantage of "improving" code coverage by laundering
the responsibility for testing this code to upstream. And unlike us,
upstream [actually has tests][1]!

[1]: https://github.com/tower-rs/tower/blob/74f9047f309d16fa25978e34ea86d639383c02af/tower/tests/limit/concurrency.rs


olix0r (approved):

hawkw:
why the heck did this decrease coverage, i'm very confused

PR #1114, opened 2021-06-23
inbound: add `inbound_tls_detect_error_total` metric

hawkw:
Currently, the proxy's metrics don't indicate why TLS detection may have
failed.

This branch implements the metric described in linkerd/linkerd2#6268 ---
it adds an `inbound_tls_detect_failure_total` metric which is labeled
with the reason TLS detection failed (either `error="timeout"` or
`error="io`). When IO errors occur, they are additionally labeled with
an errno, if one is available.

I wasn't totally sure where to put some of the code for this metric.
Since we only record this on the inbound side currently, it didn't make
sense to add it to the `ProxyMetrics` struct. Any suggestions on
rearranging the code would be welcome. :) 

Fixes linkerd/linkerd2#6268


linkerd/app/inbound/src/direct.rs
```diff
@@ -90,6 +90,7 @@ impl<N> Inbound<N> {
             config,
             runtime: rt,
             stack: tcp,
+            tls_detect_metrics,
```
olix0r, line 93: I think we actually need to support tls detect metrics on the direct stack...

linkerd/app/integration/src/tests/telemetry/tls_detect.rs
```diff
@@ -0,0 +1,124 @@
...
+    let identity::Identity {
+        mut env,
+        mut certify_rsp,
+        ..
+    } = identity::Identity::new(
+        "foo-ns1",
+        "foo.ns1.serviceaccount.identity.linkerd.cluster.local".to_string(),
+    );
+
+    // The identity service is needed for the proxy to start.
```
hawkw, line 18: hmm, in this case, it's needed specifically because we want to test behavior with TLS enabled --- so i think this should maybe be something more like
```rust
// The identity service is required for the proxy to start because identity is
// enabled in these tests
```

but, not a big deal.

linkerd/app/inbound/src/lib.rs
```diff
@@ -42,11 +42,12 @@ pub struct Config {
     pub profile_idle_timeout: Duration,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct Inbound<S> {
     config: Config,
     runtime: ProxyRuntime,
     stack: svc::Stack<S>,
+    tls_detect_metrics: metrics::tls_detect::ErrorRegistry,
```
hawkw, line 50: i thought about doing this initially but didn't want to make the diff huge from touching every place we destructure this struct...i guess since it's also needed in the direct stack, this was necessary after all.

linkerd/app/inbound/src/lib.rs
```diff
@@ -42,11 +42,12 @@ pub struct Config {
     pub profile_idle_timeout: Duration,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone)]
```
hawkw, line 45: i thought i added a `Debug` impl for the `ErrorRegistry` type? was there a reason we had to remove this?

linkerd/app/inbound/src/direct.rs
hawkw: mmkay, i can add that (plus a test?)

linkerd/app/inbound/src/lib.rs
olix0r: hm, it didn't work for me, but it's also never used afaict

linkerd/app/inbound/src/lib.rs
olix0r: Yeah, actually it might be best to revert this as it breaks some other tests -- unless this impls `Default`?

linkerd/app/integration/src/tests/telemetry/tls_detect.rs
olix0r: Yeah, please feel free to correct these comments, I was just taking a stab at explaining it

linkerd/app/integration/src/tests/telemetry/tls_detect.rs
```diff
@@ -0,0 +1,124 @@
...
+            .run()
+            .await
+    };
+
+    // We provide a mocked TCP server that always immediately drops accepted socket. This should
+    // trigger errors.
+    let srv = tcp::server()
+        .accept_fut(move |sock| async { drop(sock) })
+        .run()
+        .await;
```
olix0r, line 32: I don't actually understand how this works for triggering timeouts...

hawkw: it doesn't, we don't actually talk to the server. it's just there so that there's an ORIG_DST port.

kleimkuhler (approved):
I assume Oliver will probably want another look at this with some of the conversation already, but this looks good to me.

linkerd/app/inbound/src/lib.rs
```diff
@@ -42,12 +42,17 @@ pub struct Config {
     pub profile_idle_timeout: Duration,
 }
 
+#[derive(Clone)]
+struct Runtime {
+    tls_detect_metrics: metrics::tls_detect::ErrorRegistry,
+    proxy: ProxyRuntime,
+}
```
hawkw, line 49: hmm, wrapping this was probably a good call :+1:

olix0r:
Digging into this lead to a slightly bigger yakshave: our existing error metrics are pretty poor (lacking any descriptive labels about the target).

I've put up https://github.com/linkerd/linkerd2-proxy/pull/1133 as a setup PR which does the error-metrics change and changes the HTTP error metric. Once that's merged, I'll pull this into this branch.

I think where we probably want to go here is to introduce `{in,out}bound_tcp_accept_errors_total` metric (or something like that) with a `tls_detect_timeout` reason on the inbound side... This will simplify a bunch of the special casing if we can just surface errors on the tcp stacks (which may include discovery failures on the outbound side). I know this isn't what was in the initial issue description, but I needed to actually see the diff to make sense of this :)

olix0r (approved):

PR #1117, opened 2021-06-28
metrics: factor out `NewService` for target-scoped metrics

hawkw:
While working on adding a target port label to the TLS detection failure
metrics added in #1114, I realized that there was some boilerplate code
for scoping metrics based on a label from a target that was basically
repeated between `linkerd-http-metrics` and `linkerd-transport`'s
`metrics` module, and that this boilerplate code would have to be
rewritten a third time for adding target port labels to the TLS
detection metrics.

Instead of rewriting this code a third time, I factored it out into a
`NewMetrics` type in the `linkerd-metrics` crate, which given a registry
of scopes, a `NewService`, and a metrics service type that can be
constructed from the `NewService`'s service and the registry's metrics,
implements `NewService` by wrapping the inner service with a metrics
service scoped for the target's labels. The `linkerd-http-metrics`
crate's `NewHttpMetrics` and the `linkerd-transport` crate's
`metrics::MakeAccept` can both be replaced with the new `NewMetrics`
type.


kleimkuhler (approved):
Nice! I like the switch to `parking_lot` for the metrics code, and using `NewMetrics` to reduce some of the boilerplate looks good to me.

hawkw:
@olix0r rebased onto `main` after merging #1142

olix0r (approved):

PR #1118, opened 2021-06-28
metrics: add target_addr label to TCP accept error metrics

hawkw:
This branch adds a `target_addr` label to the
`inbound_tcp_accept_errors_total` and
`outbound_tcp_accept_errors_total` metrics.

### DIFF
--- a/linkerd/app/core/src/control.rs
+++ b/linkerd/app/core/src/control.rs
@@ -33,6 +33,13 @@ impl fmt::Display for ControlAddr {
     }
 }
 
+// Needed for metrics labels, admittedly kind of weird...hm.
+impl From<&'_ ControlAddr> for ControlAddr {
+    fn from(addr: &Self) -> Self {
+        addr.clone()
+    }
+}
+
 type BalanceBody =
     http::balance::PendingUntilFirstDataBody<tower::load::peak_ewma::Handle, hyper::Body>;
 
--- a/linkerd/app/core/src/control.rs
+++ b/linkerd/app/core/src/control.rs
@@ -33,13 +33,6 @@ impl fmt::Display for ControlAddr {
     }
 }
 
-// Needed for metrics labels, admittedly kind of weird...hm.
-impl From<&'_ ControlAddr> for ControlAddr {
-    fn from(addr: &Self) -> Self {
-        addr.clone()
-    }
-}
-
 type BalanceBody =
     http::balance::PendingUntilFirstDataBody<tower::load::peak_ewma::Handle, hyper::Body>;
 
--- a/linkerd/app/core/src/metrics/tcp_accept_errors.rs
+++ b/linkerd/app/core/src/metrics/tcp_accept_errors.rs
@@ -27,6 +27,9 @@ pub struct Registry {
 
 type Scope = Mutex<HashMap<AcceptErrors, metrics::Counter>>;
 
+type NewErrorMetrics<N, S> =
+    metrics::NewMetrics<N, OrigDstAddr, Scope, RecordError<LabelAcceptErrors, AcceptErrors, S>>;
+
 #[derive(Clone, Copy, Debug, Default)]
 pub struct LabelAcceptErrors(());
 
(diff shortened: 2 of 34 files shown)


linkerd/app/core/src/metrics/tcp_accept_errors.rs
```diff
@@ -27,32 +36,56 @@ pub enum AcceptErrors {
...
     pub fn outbound() -> Self {
-        Self(linkerd_error_metrics::Registry::new(
-            outbound_tcp_accept_errors_total,
-        ))
+        Self::with_metric(outbound_tcp_accept_errors_total)
     }
 
-    pub fn layer(&self) -> Layer {
-        self.0.layer(LabelAcceptErrors(()))
+    fn with_metric(metric: linkerd_error_metrics::Metric) -> Self {
```
olix0r, line 51: This is just `new`, right?

linkerd/proxy/transport/src/addrs.rs
```diff
@@ -28,6 +28,9 @@ pub struct Local<T>(pub T);
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub struct Remote<T>(pub T);
 
+#[derive(Eq, PartialEq, Debug, Copy, Clone, Hash)]
+pub struct TargetPort(u16);
+
```
olix0r, line 33: Why do we need a new type for this? Isn't this already covered by `Remote<ServerAddr>`?

linkerd/proxy/transport/src/addrs.rs
```diff
@@ -155,3 +158,23 @@ impl<T: fmt::Display> fmt::Display for Remote<T> {
...
+    fn from(addr: SocketAddr) -> Self {
+        Self(addr.port())
+    }
+}
+
+impl linkerd_metrics::FmtLabels for TargetPort {
+    fn fmt_labels(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "target_port=\"{}\"", self.0)
+    }
+}
```
olix0r, line 180: I don't think we should have the target types tightly bound to the metric labels like this... I.e. we don't want to expect that the same target types will always be rendered as the same labels, I think.

olix0r (changes requested):
AFAICT this now applies to both inbound and outbound stacks (the description only references inbound)...  And in light of this (and other upcoming changes to the inbound stack), we're probably better off encoding this as `target_addr` in general (for both inbound and outbound). While we only refer to localhost IPs on the inbound side for now, this will change in the near future.

hawkw:
> we're probably better off encoding this as `target_addr` in general (for both inbound and outbound). 

Yup...that will actually make the implementation much simpler, anyway, since we can remove the special-cased target port param type. I only initially implemented it with target ports because that was what the original issue linkerd/linkerd2#6268 said. :)

hawkw:
@olix0r okay, pretty sure I've addressed all the review feedback now. I've added a new `TargetAddr` metric label type, and it's converted from `Param` types rather than having the targets themselves be aware of metrics labels.

The change took slightly longer than I expected because it turns out there was a bug in the test code that meant the metrics tests wouldn't actually catch a lot of potential bugs. #1146 fixes that (and this branch now depends on it).

hawkw:
Okay, now rebased onto main.

linkerd/app/core/src/control.rs
```diff
@@ -33,6 +33,13 @@ impl fmt::Display for ControlAddr {
     }
 }
 
+// Needed for metrics labels, admittedly kind of weird...hm.
+impl From<&'_ ControlAddr> for ControlAddr {
+    fn from(addr: &Self) -> Self {
+        addr.clone()
+    }
+}
+
```
olix0r, line 42: Is this comment correct? It seems to compile if I remove this impl..

hawkw: oh, yup --- this is an artifact of an earlier implementation. will remove it, thanks for noticing!

linkerd/metrics/src/new_metrics.rs
```diff
@@ -33,28 +33,29 @@ where
...
-impl<N, K, M, S, T> svc::NewService<T> for NewMetrics<N, K, M, S>
+impl<N, K, M, S, T, P> svc::NewService<T> for NewMetrics<N, K, M, S, P>
 where
-    T: svc::Param<K>,
+    T: svc::Param<P>,
     N: svc::NewService<T>,
     S: From<(N::Service, Arc<M>)>,
     M: Default,
-    K: Hash + Eq,
+    K: Hash + Eq + From<P>,
```
olix0r, line 47: More of a question than a suggestion: but what would it mean to avoid the `P` type? I.e. if we just required that `T: P: svc::Param<K>`, how would that change things?

This is a bit heady, so it would be nice to simplify this; but I don't have a great understanding of the tradeoff.

olix0r: I wonder if we can eliminate the TargetAddr type by always using `OrigDstAddr`? The TCP accept errors will _always_ be associated with the orig dst addr.

olix0r: @hawkw Some proposed simplifications here https://github.com/linkerd/linkerd2-proxy/pull/1147

olix0r (approved):

PR #1119, opened 2021-06-29
app: Set the `l5d-proxy-error` header on synthesized responses

kleimkuhler:
This change adds an `l5d-proxy-error` header to responses that fail from
the proxy to application communication. This occurs for situations like fail
fast errors, connections failing because of connect timeouts or the app not
listening, or more generally responses that take too long to start.

When this occurs and it's the proxy that returns a non-200 status code—not the
application—the `l5d-proxy-error` is added which contains the reason for
this non-200 response status code.

Currently this is not used, but the intention is that outbound proxies can start
reading this header and determine if a connection should be torn down when the
response contains this header.


linkerd/app/inbound/src/http/tests.rs
```diff
@@ -37,6 +39,8 @@ where
 {
     let connect = svc::stack(connect)
         .push_map_target(|t: TcpEndpoint| Remote(ServerAddr(([127, 0, 0, 1], t.param()).into())))
+        .push_timeout(cfg.proxy.connect.timeout)
+        .push(svc::stack::BoxFuture::layer())
```
kleimkuhler, line 43: I opted to add the timeout layer to the connect stack for all tests. If wanted, I could change `build_server` to take a connect stack instead of just a `Connect` struct so that the test is responsible for creating the connect stack used, and we don't have a timeout when we don't need it.

hawkw: hmm, rather than pushing _just_ the timeout stack, could we push the entire endpoint stack instead? i'm a little leery of having tests that test stacks that we don't actually build in a real proxy; i'd prefer to build the whole thing if it's possible.

@olix0r may disagree with me here though!

linkerd/app/inbound/src/http/tests.rs
```diff
@@ -187,6 +191,101 @@ async fn downgrade_absolute_form() {
...
+        version: proxy::http::Version::Http1,
+        tcp: TcpAccept {
+            target_addr: ([127, 0, 0, 1], 5550).into(),
+            client_addr: Remote(ClientAddr(([10, 0, 0, 41], 6894).into())),
+            tls: Conditional::None(tls::server::NoServerTls::NoClientHello),
+        },
+    };
+    let connect = support::connect().endpoint_fn_boxed(accept.tcp.target_addr, connect_error());
+
+    // Build a client uses the connect that always errors so that responses
```
hawkw, line 209: typo:
```suggestion
    // Build a client using the connect that always errors, so that responses
```

linkerd/app/inbound/src/http/tests.rs
hawkw: also, if we're including timeouts in all the test stacks now we should _probably_ stick a `tokio::time::pause` in all the tests, so that they don't fail due to spurious timeouts? that probably won't ever happen, but if we don't want to exercise timeout behavior, we should probably turn them off i think...

linkerd/app/inbound/src/http/tests.rs
```diff
@@ -207,3 +306,41 @@ fn hello_server(
...
+#[tracing::instrument]
+fn timeout_server(
+    http: hyper::server::conn::Http,
+) -> Box<dyn FnMut(Remote<ServerAddr>) -> ConnectFuture + Send> {
+    Box::new(move |endpoint| {
+        let http = http.clone();
+        let span = tracing::info_span!("timeout_server", ?endpoint);
+        Box::pin(
+            async move {
+                tracing::info!("sleeping so that the proxy hits a dispatch timeout");
```
hawkw, line 329: i'm surprised this hits the dispatch timeout rather than the connect timeout, but i guess it does?

linkerd/app/inbound/src/http/tests.rs
```diff
@@ -207,3 +306,41 @@ fn hello_server(
...
+                tokio::time::sleep(std::time::Duration::from_secs(3)).await;
+                let (client_io, server_io) = support::io::duplex(4096);
+                let timeout_service =
+                    hyper::service::service_fn(|request: Request<Body>| async move {
+                        tracing::info!(?request);
+                        Ok::<_, io::Error>(Response::new(Body::from("timeout server")))
```
hawkw, line 335: TIOLI: it would probably be fine to make the `service_fn` just be `unreachable!()`, because we _should_ never send the server a request. up to you though.

linkerd/app/core/src/errors.rs
```diff
@@ -217,31 +219,60 @@ impl<RspB: Default + hyper::body::HttpBody> respond::Respond<http::Response<RspB
...
-                Ok(http::Response::builder()
+                let mut rsp = http::Response::builder()
                     .version(self.version)
-                    .status(status)
                     .header(http::header::CONTENT_LENGTH, "0")
                     .body(ResponseBody::default())
-                    .expect("error response must be valid"))
+                    .expect("error response must be valid");
+                let status = set_http_status(&*error, rsp.headers_mut());
+                *rsp.status_mut() = status;
```
hawkw, line 228: hmm, it's a little weird/inconsistent to me that `set_http_status` _mutates_ the headers, but _returns_ a status which we then set as the response's status. what if, instead of taking a `&mut HeaderMap`, `set_http_status` took an `&mut http::response::Builder` instead, and set the headers _and_ the status code on the response in one place, before we completed building the response?

linkerd/app/core/src/errors.rs
```diff
@@ -217,31 +219,60 @@ impl<RspB: Default + hyper::body::HttpBody> respond::Respond<http::Response<RspB
...
+            HeaderValue::from_str(&e.to_string()).unwrap_or_else(|error| {
+                warn!(%error, "Failed to encode fail-fast error message");
+                HeaderValue::from_static("service in fail-fast")
+            }),
+        );
+        http::StatusCode::SERVICE_UNAVAILABLE
+    } else if error.is::<tower::timeout::error::Elapsed>() {
+        headers.insert(
+            L5D_HTTP_ERROR_MESSAGE,
+            HeaderValue::from_static("proxy dispatch timed out"),
```
hawkw, line 261: hmm, is this actually correct? the `tower::timeout` layer is used to implement both the dispatch timeout (which bounds the length of time requests spend in the proxy) _and_ the connect timeout, which is a different error. i'm not sure if all `tower::timeout::error::Elapsed` we could see here will actually be from the dispatch timeout.

i noticed that the metrics code _also_ assumes all `Elapsed` errors are from the dispatch timeout: https://github.com/linkerd/linkerd2-proxy/blob/00c3394ab36eeff14192e6ae52c41238376daabd/linkerd/app/core/src/errors.rs#L365-L366 i'm not sure if this is correct either --- @olix0r, do you happen to know if it's possible for the errors layer to see `Elapsed` errors generated by connect timeouts? if it is, i wonder if we should do something like wrapping the `tower` error type in a new error type to indicate whether an `Elapsed` error was due to a dispatch or connection timeout...

if we see connect timeouts here, i think they should *probably* get 504 Gateway Timeout instead of 503 Service Unavailable?

linkerd/app/core/src/errors.rs
kleimkuhler: Yep that makes sense. `set_http_status` now makes the `http:response::Builder` and returns it after setting the header and status that correspond to the error the proxy encountered.

linkerd/app/inbound/src/http/tests.rs
kleimkuhler: The reason I went with pushing just the timeout instead of the endpoint is right now `push_endpoint` is only a method for `Outbound` and we're dealing with `Inbound` here. By pushing just a timeout here we don't need to add a method to `Inbound` that would only be used by test code I think? Let me know if you had something else in mind.

linkerd/app/inbound/src/http/tests.rs
kleimkuhler: I replaced everything under the sleep with `unreachable!()`.

linkerd/app/inbound/src/http/tests.rs
kleimkuhler: I've addressed this by adding the `ConnectTimeout` and changing the test code to use `to_tcp_connect` which ensures the tower timeouts are wrapped with this new error.

linkerd/app/core/src/errors.rs
kleimkuhler: By changing the test code to use `to_tcp_connect`, and changing it's signature to take a `connect_tcp`, the proxy code and test code can now both use this method to pass in the connect service that the stack should build off of.

In the test code, this `connect_tcp` is the [connect_timeout](https://github.com/linkerd/linkerd2-proxy/blob/911a4f03a6f73b854dac1c1a78a8560928495e1e/linkerd/app/inbound/src/http/tests.rs#L317) function that ensures the test proxy hits a connect timeout before a service is returned.

Additionally, `to_tcp_connect` now wraps timeout errors with a new `ConnectTimeout` which we can check for in [set_http_status](https://github.com/linkerd/linkerd2-proxy/blob/b41e85adb15fab891ef32c0330e44aa61656ff33/linkerd/app/core/src/errors.rs#L259-L264) and return a 504 Gateway Timeout instead of a 503 Service Unavailable.

olix0r (changes requested):
AFAICT this doesn't work with gRPC requests. Probably worth a test

kleimkuhler:
In adding tests for gRPC I realized this change is not working for H2 either so there are a few fixes coming for that as well.

linkerd/app/core/src/errors.rs
```diff
@@ -9,12 +9,15 @@ use linkerd_proxy_http::{ClientHandle, HasH2Reason};
...
 use linkerd_tls as tls;
 use pin_project::pin_project;
+use std::fmt;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 use thiserror::Error;
 use tonic::{self as grpc, Code};
 use tracing::{debug, warn};
 
+pub const L5D_HTTP_ERROR_MESSAGE: &str = "l5d-http-error-message";
```
olix0r, line 19: Let's call this `l5d-proxy-error` -- I don't think _http_ is particularly meaningful (this is set on HTTP requests) and _message_ is superfluous. This header indicates that there was a proxy error.

linkerd/app/core/src/errors.rs
```diff
@@ -229,32 +235,81 @@ impl<RspB: Default + hyper::body::HttpBody> respond::Respond<http::Response<RspB
...
+                debug!(%status, version = ?self.version, "Handling error with HTTP response");
+                Ok(rsp)
             }
         }
     }
 }
 
-fn http_status(error: &(dyn std::error::Error + 'static)) -> StatusCode {
+fn set_l5d_error_header(error: &(dyn std::error::Error + 'static)) -> http::response::Builder {
+    let mut builder = http::Response::builder();
```
olix0r, line 252: nit: it would be a bit more consistent if this took the builder as an argument, like `set_http_status` does. and its name (`set_...`) indicates that it sets a header, not that it creates a new response builder.

linkerd/app/core/src/errors.rs
```diff
@@ -1,4 +1,4 @@
-use http::{header::HeaderValue, StatusCode};
+use http::header::HeaderValue;
```
olix0r, line 1: Why this change? Why not keep `StatusCode` in scope?

linkerd/app/core/src/errors.rs
```diff
@@ -229,32 +235,81 @@ impl<RspB: Default + hyper::body::HttpBody> respond::Respond<http::Response<RspB
...
+    error: &(dyn std::error::Error + 'static),
+) -> http::response::Builder {
     if let Some(HttpError { http, .. }) = error.downcast_ref::<HttpError>() {
-        *http
+        builder.status(*http)
     } else if error.is::<ResponseTimeout>() {
-        http::StatusCode::GATEWAY_TIMEOUT
-    } else if error.is::<FailFastError>() || error.is::<tower::timeout::error::Elapsed>() {
-        http::StatusCode::SERVICE_UNAVAILABLE
+        builder.status(http::StatusCode::GATEWAY_TIMEOUT)
```
olix0r, line 300: i.e. and then shorten this to just `StatusCode::...`

not a big deal either way, i guess

linkerd/app/core/src/errors.rs
kleimkuhler: We are primarily using `http::StatusCode` everywhere else in the file and I had made an intermediate change where this became unused. I'll change all the other references to it and add the import back.

linkerd/app/inbound/Cargo.toml
```diff
@@ -14,6 +14,7 @@ bytes = "1"
 http = "0.2"
 futures = { version = "0.3", default-features = false }
 linkerd-app-core = { path = "../core" }
+linkerd-app-test = {path = "../test"}
```
olix0r, line 17: Why do we need this as a non-dev dependency?

olix0r:
```suggestion
```

It can safely be removed

olix0r (approved):

PR #1122, opened 2021-07-01
tls: always use `tokio-rustls` re-export of `rustls`

hawkw:
The `tokio-rustls` crate includes a public re-export of `rustls`. This
is intended to allow user code to use the *same* version of `rustls`
that `tokio-rustls` depends on, without having to manage a separate
`rustls` dependency.

This commit removes all explicit `rustls` dependencies from
`Cargo.toml`, and replaces all imports from `rustls` with
`tokio_rustls::rustls`. The advantage of this is that it is now much
easier to keep the `tokio-rustls` and `rustls` dependencies in sync; we
can now change only the `tokio-rustls` dependency, and we'll always have
the `rustls` version that `tokio-rustls` depends on.

This change was originally intended to make testing with the upcoming
`rustls` release easier (linkerd/linkerd2#6379), but I factored it out
into its own PR since we can land this change *now*, without
waiting for the new `rustls` release to be published.


olix0r (approved):

olix0r:
The `tokio-rustls` crate includes a public re-export of `rustls`. This
is intended to allow user code to use the *same* version of `rustls`
that `tokio-rustls` depends on, without having to manage a separate
`rustls` dependency.

This commit removes all explicit `rustls` dependencies from
`Cargo.toml`, and replaces all imports from `rustls` with
`tokio_rustls::rustls`. The advantage of this is that it is now much
easier to keep the `tokio-rustls` and `rustls` dependencies in sync; we
can now change only the `tokio-rustls` dependency, and we'll always have
the `rustls` version that `tokio-rustls` depends on.

Furthermore, cargo-deny will now only permit dependencies on `rustls` via
`tokio-rustls`.

Co-authored-by: Oliver Gould <ver@buoyant.io>

PR #1138, opened 2021-07-09
Move tls_detect_errors_total to *_tcp_accept_errors_total

olix0r:
* Replace tls_detect module with tcp_accept_errors
* Remove labeling from the error-metrics--can be embedded with
  tcp_accept_errors
* Remove errno codes from labels so that labels are consistent for all
  instances of the metric
* Add metrics to the outbound side, too.


linkerd/app/core/src/metrics/tcp_accept_errors.rs
```diff
@@ -0,0 +1,90 @@
...
+impl LabelError<Error> for LabelAcceptErrors {
+    type Labels = AcceptErrors;
+
+    fn label_error(&self, err: &Error) -> Self::Labels {
+        let mut curr: Option<&dyn std::error::Error> = Some(err.as_ref());
+        while let Some(err) = curr {
+            if err.is::<TlsDetectTimeout>() {
+                return AcceptErrors::TlsDetectTimeout;
+            } else if err.is::<std::io::Error>() {
+                // We ignore the error code because we want all labels to be consistent.
```
hawkw, line 70: nit/TIOLI: we could alternatively just put an `errno="none"` label or something in the other and detect timeout cases? not sure how important including the errno is though?

linkerd/app/integration/src/tests/telemetry/tls_detect.rs
```diff
@@ -3,6 +3,8 @@ use std::time::SystemTime;
 
 const TIMEOUT: Duration = Duration::from_millis(640); // 640ms ought to be enough for anybody.
 
+const METRIC: &str = "inbound_tcp_accept_errors_total";
```
hawkw, line 6: nit: we should probably also rename this module from `tls_detect.rs` to `tcp_accept.rs`?

linkerd/app/core/src/metrics/tcp_accept_errors.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// === impl AcceptErrors ===
+
+impl FmtLabels for AcceptErrors {
+    fn fmt_labels(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::TlsDetectTimeout => f.pad("error=\"tls_detect_timeout\""),
+            Self::Io => write!(f, "error=\"io\""),
+            Self::Other => write!(f, "error=\"other\""),
```
hawkw, line 87: nit: we should probably format these consistently rather than using `Formatter::pad` for one and `write!` for the other. also, we probably shouldn't use `pad`, as it will add whitespace when the formatter is configured to pad things:
```suggestion
            Self::TlsDetectTimeout => fmt::Display::fmt("error=\"tls_detect_timeout\"", f),
            Self::Io => fmt::Display::fmt("error=\"io\"", f),
            Self::Other => fmt::Display::fmt("error=\"other\"", f),
```

hawkw (approved):
lgtm modulo a couple nits!

PR #1145, opened 2021-07-15
Add Stack::push_connect_timeout

olix0r:
The connect-timeout logic isn't really inbound-specific and may be
useful elsewhere, like on the outbound stack.

This change moves this logic from `Inbound::to_tcp_connect` to
`Stack::push_connect_timeout` and restores `Inbound::to_tcp_connect` to
be responsible for building the TCP connection layer. Our tests go back
to mocking the `to_tcp_connect` layering (as it's quite simple).

### DIFF
--- a/linkerd/app/core/src/errors.rs
+++ b/linkerd/app/core/src/errors.rs
@@ -512,11 +512,11 @@ impl HttpError {
 }
 
 #[derive(Debug)]
-pub struct ConnectTimeout();
+pub(crate) struct ConnectTimeout(pub std::time::Duration);
 
 impl fmt::Display for ConnectTimeout {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.pad("connect timed out")
+        write!(f, "connect timed out after {:?}", self.0)
     }
 }
 
--- a/linkerd/app/inbound/src/http/mod.rs
+++ b/linkerd/app/inbound/src/http/mod.rs
@@ -104,7 +104,7 @@ where
     C: svc::Service<TcpEndpoint> + Clone + Send + Sync + Unpin + 'static,
     C::Response: io::AsyncRead + io::AsyncWrite + Send + Unpin + 'static,
     C::Error: Into<Error>,
-    C::Future: Send + Unpin,
+    C::Future: Send,
 {
     pub fn push_http_router<P>(
         self,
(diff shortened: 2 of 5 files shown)


linkerd/app/inbound/src/http/mod.rs
```diff
@@ -133,6 +133,7 @@ where
 
         // Creates HTTP clients for each inbound port & HTTP settings.
         let endpoint = connect
+            .push(svc::stack::BoxFuture::layer()) // Makes the response future `Unpin`.
```
olix0r, line 136: I moved this from `to_tcp_connect` because it's really the http_router that needs this behavior -- but no other users of the tcp connect stack needs this. This simplifies type constraints.

linkerd/app/inbound/src/http/tests.rs
```diff
@@ -37,9 +37,14 @@ fn build_server<I>(
 where
     I: io::AsyncRead + io::AsyncWrite + io::PeerAddr + Send + Unpin + 'static,
 {
-    let Config { proxy, .. } = cfg.clone();
+    // Mocks to_tcp_connect.
+    let connect = svc::stack(connect)
+        .push_map_target(|t: TcpEndpoint| Remote(ServerAddr(([127, 0, 0, 1], t.param()).into())))
+        .push_connect_timeout(cfg.proxy.connect.timeout)
+        .into_inner();
```
olix0r, line 44: This is simple enough that I'm comfortable duplicating what's in `to_tcp_connect` -- I don't think it's worth having this all factored out for the `push_connect_timeout` (the map_target should go away in an upcoming change).

linkerd/app/core/src/errors.rs
```diff
@@ -512,11 +512,11 @@ impl HttpError {
 }
 
 #[derive(Debug)]
-pub struct ConnectTimeout();
+pub(crate) struct ConnectTimeout(pub std::time::Duration);
 
 impl fmt::Display for ConnectTimeout {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.pad("connect timed out")
+        write!(f, "connect timed out after {:?}", self.0)
```
olix0r, line 519: I don't think `pad` does what we want; and it seems useful to include the timeout.

linkerd/app/inbound/src/http/mod.rs
kleimkuhler: Talked off GitHub about this but makes sense 👍

kleimkuhler (approved):

PR #1147, opened 2021-07-15
Seal the TargetAddr label type in app::core

olix0r:
Simplifies the error metrics signature

### DIFF
--- a/linkerd/app/core/src/metrics/tcp_accept_errors.rs
+++ b/linkerd/app/core/src/metrics/tcp_accept_errors.rs
@@ -80,8 +80,8 @@ impl FmtMetrics for Registry {
         let errors = self.scopes.lock();
 
         self.metric.fmt_help(f)?;
-        for (oda, ms) in errors.iter() {
-            let ta = labels::TargetAddr((*oda).into());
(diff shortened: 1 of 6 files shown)


linkerd/app/core/src/metrics/tcp_accept_errors.rs
```diff
@@ -21,7 +21,7 @@ metrics::metrics! {
 
 #[derive(Clone, Debug)]
 pub struct Registry {
-    scopes: metrics::SharedStore<TargetAddr, Scope>,
+    scopes: metrics::SharedStore<OrigDstAddr, Scope>,
```
hawkw, line 24: Not wanting to just key the registry on `SocketAddr`s was the original reason I didn't do this and just wrap the values in the type that knows how to format labels when actually formatting them. But, at the time, we were only including the target port, that was more important. Now that the label includes the entire address, this is probably fine.

Personally, I thought it was conceptually nicer to have the metrics registry be indexed by the label type rather than by a non-metrics-specific label type...but this definitely reduces the complexity of all the other code!

hawkw (approved):

PR #1148, opened 2021-07-16
Reduce stack-builder boilerplate

olix0r:
We have 100+ lines of boilerplate from the inbound and outbound stack
builders. This change introduces helpers: `Inbound::map_stack` and
`Outbound::map_stack`. These helpers provide a way to map over the inner
stack type, which eliminates the boilerplate of de/re-structruring the
builder type.

No functional changes.

### DIFF
--- a/linkerd/app/inbound/src/direct.rs
+++ b/linkerd/app/inbound/src/direct.rs
@@ -86,87 +86,79 @@ impl<N> Inbound<N> {
         GSvc::Error: Into<Error>,
         GSvc::Future: Send,
     {
-        let Self {
-            config,
-            runtime: rt,
-            stack: tcp,
-        } = self;
-        let detect_timeout = config.proxy.detect_protocol_timeout;
+        self.map_stack(|config, rt, tcp| {
+            let detect_timeout = config.proxy.detect_protocol_timeout;
 
-        let stack = tcp
-            .instrument(|_: &TcpEndpoint| debug_span!("opaque"))
-            // When the transport header is present, it may be used for either local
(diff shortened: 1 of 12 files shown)


linkerd/app/inbound/src/direct.rs
```diff
@@ -86,87 +86,79 @@ impl<N> Inbound<N> {
...
+                // connection if it doesn't include an mTLS identity.
+                .push_request_filter(ClientInfo::try_from)
+                .push(svc::BoxNewService::layer())
+                .push(tls::NewDetectTls::layer(
+                    rt.identity.clone().map(WithTransportHeaderAlpn),
+                    detect_timeout,
+                ))
+                .check_new_service::<T, I>()
+                .push_on_response(svc::BoxService::layer())
+                .push(svc::BoxNewService::layer())
```
hawkw, line 160: since we always add `BoxService` and `BoxNewService` at the end of every stack builder method (and the type signatures require us to do this), I wonder if it makes sense to just make `map_stack` always push `BoxService` and `BoxNewService` methods? that way, it's impossible to forget.

not a blocker, we can add this later.

hawkw (approved):
i had been thinking about adding something like this on and off for a while, lgtm!

linkerd/app/inbound/src/direct.rs
olix0r: that requires adding type constraints to ensure that stack is a service/newservice, etc. not necessarily a bad idea for a followup, but will make this less flexible (e.g., it couldn't be used for the connect stack).

PR #1156, opened 2021-07-19
inbound: Reorganize server into smaller stacks

olix0r:
The inbound server stack is fairly large. While sketching out changes to
support inbound policy, it's helpful to split this larger stack into a
few smaller components (using the recently-introduced `map_stack`
helper).

This change:

* Modifies `Inbound::push_http_server` to return a boxed service. This
  service was being boxed by `Inbound::into_server`. This change helps
  reduce boilerplate in our test utilities, too. (In fact, this change
  requires that we eliminate some unnecessary type constraints).
* Changes `Inbound::to_tcp_connect` to `Inbound::into_tcp_connect`. The
  internal cloning is not necessary.
* Changes `Inbound::into_server` to `Inbound::push_server`. Internally,
  this function now uses several calls to `map_stack` to build the
  inbound server stack.

### DIFF
--- a/linkerd/app/inbound/src/lib.rs
+++ b/linkerd/app/inbound/src/lib.rs
@@ -95,8 +95,8 @@ impl Inbound<()> {
         self.map_stack(move |_, _, _| svc::stack(stack))
     }
 
-    pub fn to_tcp_connect<T: svc::Param<u16>>(
-        &self,
+    pub fn into_tcp_connect<T: svc::Param<u16>>(
+        self,
     ) -> Inbound<
         impl svc::Service<
                 T,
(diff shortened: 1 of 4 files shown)


linkerd/app/inbound/src/lib.rs
```diff
@@ -217,86 +220,93 @@ where
...
+                        move |t: T| {
+                            let OrigDstAddr(addr) = t.param();
+                            if !disable_detect.contains(&addr.port()) {
+                                Ok::<_, Never>(svc::Either::A(t))
+                            } else {
+                                Ok(svc::Either::B(TcpAccept::port_skipped(t)))
+                            }
+                        },
+                        opaque
+                            .instrument(|_: &TcpAccept| debug_span!("forward"))
```
hawkw, line 289: very unimportant nit: it seems a little inconsistent that we add the span to `opaque` here, but the span in the direct stack is added in the let binding above. very minor preference for making these consistent, but this absolutely does not actually matter at all :)

linkerd/app/test/src/http_util.rs
```diff
@@ -36,18 +39,11 @@ impl Default for Server {
...
 }
 
-pub async fn run_proxy<S>(mut server: S) -> (io::DuplexStream, JoinHandle<Result<(), Error>>)
-where
-    S: tower::Service<io::DuplexStream> + Send + Sync + 'static,
-    S::Error: Into<Error>,
-    S::Response: std::fmt::Debug + Send + Sync + 'static,
-    S::Future: Send,
-{
+pub async fn run_proxy(mut server: BoxServer) -> (io::DuplexStream, JoinHandle<Result<(), Error>>) {
```
hawkw, line 42: lovely :)

hawkw (approved):
looks nice, no real concerns here!

linkerd/app/inbound/src/lib.rs
olix0r: agreed, will sort this out with the followups.

kleimkuhler (approved):

PR #1158, opened 2021-07-21
error: Replace `Never` with `std::convert::Infallible`

olix0r:
We define an `error::Never` type that is effectively identical to the
`std::convert::Infallible` type--and `Infallible` has a clearer name and
provides implementations for more common traits. Most importantly, it's
less surprising for new contributors.

This change removes our `Never` type and re-exports `Infallible` from
the `linkerd_error` crate.


kleimkuhler (approved):

hawkw (approved):
:+1:

PR #1160, opened 2021-07-21
metrics: rework `LabelError` for composability

hawkw:
Currently, the `LabelError` trait works by taking a reference to an
error (of a generic type) and returning a `Labels` type. When the error
being labeled is a `dyn Error`, the type implementing `LabelError` must
loop over the error and its sources, and attempt to downcast the error
to various types.

The current design works fine, but it's difficult to compose multiple
`LabelError` types together. In particular, the looping over the source
chain currently has to be repeated in every `LabelError` type that's
composed together. This is kind of a pain and is potentially
inefficient.

This change moves the loop out of each individual implementation of
`LabelError`, and into the `RecordError` middleware. The `LabelError`
trait is changed to return an `Option<Self::Labels>`, and is called by
the loop for each source in the error's source chain. If `LabelError`
returns `Some`, the loop stops. This allows us to compose together
multiple `LabelError` implemetations in such a way that we break out of
the loop if either one wants to stop iterating. The `Labels` type is now
required to implement `Default`, and the default value is returned if
the whole source chain has been traversed without the `LabelError` type
returning any labels.

On top of this, I've added an `or_else` combinator for combining two
`LabelError` implementations with the same `Labels` type. This isn't
currently unused, but it will make sense in context with PR #1161.

### DIFF
--- a/linkerd/app/core/src/errors.rs
+++ b/linkerd/app/core/src/errors.rs
@@ -424,32 +424,30 @@ impl fmt::Display for IdentityRequired {
 impl std::error::Error for IdentityRequired {}
 
 impl LabelError {
-    fn reason(err: &(dyn std::error::Error + 'static)) -> Reason {
+    fn reason(err: &(dyn std::error::Error + 'static)) -> Option<Reason> {
         if let Some(HttpError { reason, .. }) = err.downcast_ref::<HttpError>() {
-            *reason
+            return Some(*reason);
         } else if err.is::<ResponseTimeout>() {
-            Reason::ResponseTimeout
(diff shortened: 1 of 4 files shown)


hawkw:
Per https://github.com/linkerd/linkerd2-proxy/pull/1161#discussion_r674971498, we've decided that this isn't worth the extra effort and we'll just add the new error type in #1161 to the list of errors that `linkerd_app_core`'s `tcp_accept_errors` module knows about. So it's not really necessary to be able to compose together error labeling logic, and we can close this for now.

It might be worth revisiting in the future if we ever need to do this for other reasons.

PR #1161, opened 2021-07-21
inbound: don't forward to addresses not in `LINKERD2_PROXY_INBOUND_IPS`

hawkw:
The injector now sets an environment variable
`LINKERD2_PROXY_INBOUND_IPS`. This environment variable is intended to
limit which target IP addresses the proxy will forward inbound traffic
to.

This branch adds a request filter layer in the inbound accept stack that rejects
connections with `SO_ORIGINAL_DST` addresses outside
of the configured list. If the environment variable is empty/unset, all
connections are allowed.

When connections with target IPs not in the allowlist are rejected, we
increment the `inbound_tcp_accept_error_total` metric with a
`error="unexpected"` label.  I've also added a test for the error
metric.

Furthermore, the inbound proxy now forwards connections to their
original target IP address, rather than to the original target port on
localhost, as described in linkerd/linkerd2#6495. The relevant
target types now carry a `Remote<ServerAddr>` rather than just a
port.

Some of the tests failed after changing the inbound proxy to route to
the original destination IP address rather than to localhost. This is
because of the `Proxy::inbound_fuzz_addr` method in the test-support
proxy builder. Previously, this was used to replace the  mocked
SO_ORIGINAL_DST's IP address with a totally bogus one. We could do this
without breaking stuff because the inbound proxy would still always
forward on localhost.

It doesn't do that anymore, so the IP address part of the mock
SO_ORIGINAL_DST is actually load-bearing now. This branch removes
`inbound_fuzz_addr` from the integration tests. Hopefully those tests
are still testing the right thing...

Closes linkerd/linkerd2#6495

PR #1164, opened 2021-07-23
inbound: allow netmasks in `LINKERD2_PROXY_INBOUND_IPS`

hawkw:
This branch extends #1161 to support specifying netmasks in CIDR
notation in the `LINKERD2_PROXY_INBOUND_IPS` environment variable. This
will cause the proxy to accept connections targeting any IPs that match
that netmask. This isn't strictly necessary, since the proxy injector
only ever sets single IPs in that env variable currently, but I thought
that since we accept CIDRs in other env vars, we may as well accept them
here as well.

In order to continue accepting single IPs without prefix lengths, the
`parse_nets` function was extended to try parsing as an IP address when
parsing as a CIDR (the `ipnet::IpNet` type) fails. If this succeeds, we
construct an `IpNet` from the parsed IP address, with the prefix length
equal to the number of bits in an address of that type (which will match
only that address exactly). The `AllowIps` request filter was rewritten
to just use `IpMatch` internally.

In some ways, this is actually simpler, since it uses more of our
existing code for parsing CIDRs and matching IPs.

Depends on #1161


hawkw:
(this is a draft because it depends on #1161, which is also a draft)

olix0r:
I don't think we really need to support this.

PR #1169, opened 2021-07-27
stack: Introduce ExtractParam and InsertParam

olix0r:
Stack modules that take configuration take one of two approaches: they
either use the `Param` trait to extract a configuration from the target
or they take configuration at construction-time. For example, the TLS
server module takes its detection timeout as an argument when
constructing its layer. But we want to modify the inbound stack to
configure detection timeouts dynamically, as informed by discovery; so
it will no longer be appropriate to take this configuration when
constructing the layer.

This change introduces the `stack::ExtactParam` trait which lets us
provide a _strategy_ to the TLS server layer. For now, strategy
implementations ignore the target type, continuing to use a fixed
timeout set during stack construction; but in an upcoming change we'll
be able to modify some of these stacks to use the target to configure
this value.

The `ExtractParam` trait decouples this decision from the stack module:
The TLS server module no longer cares whether the timeout is configured
statically or dynamically.

Furthermore, some stack modules may want to insert new paramters into
the target. For instance, the TLS server inserts the connections TLS
status into the target. We've generally managed this with tuples, but
this leads to some britleness around the shape of the target.

The `InsertParam` trait allows stacks to construct new target types
other than tuples, though the existing implementations continue to
construct tuples as before. This will support more complicated target
type construction.

This change updates the TLS server to use these new traits without
altering its prior functionality. While this incurs some param boilerplate,
this sets up followup changes that will use this flexbility to configure the
inbound stack based on per-port policies.

### DIFF
--- a/linkerd/app/admin/src/stack.rs
+++ b/linkerd/app/admin/src/stack.rs
@@ -5,7 +5,7 @@ use linkerd_app_core::{
     metrics::{self, FmtMetrics},
     proxy::{http, identity::LocalCrtKey},
     serve,
-    svc::{self, Param},
+    svc::{self, ExtractParam, InsertParam, Param},
     tls, trace,
     transport::{listen::Bind, ClientAddr, Local, Remote, ServerAddr},
     Error,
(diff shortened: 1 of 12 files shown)


olix0r:
buzzer compilation failures are due to https://github.com/rust-lang/rust/issues/87455

kleimkuhler (approved):
Thanks for the thorough description this helped the review. Changes look good and makes sense.

PR #1170, opened 2021-07-28
inbound: Consolidate port-based switching

olix0r:
The `inbound::prevent_loop` module implements predicates for switching
based on the target port. But there's no reason for this control flow to
be decoupled from our stack consruction.

In preparation for further changes to inbound-port-based policy, this
change eliminates the `prevent_loop` module. The tcp connection stack is
updated to handle loop detection (instead of the TCP forward stack) so
that we are totally unable to initiate looping connections (i.e. if some
higher part of the stack were to do something unexpected).


linkerd/app/core/src/svc.rs
```diff
@@ -178,14 +178,10 @@ impl<S> Stack<S> {
...
     /// Note that any timeouts errors from the inner service will be wrapped as well.
-    pub fn push_connect_timeout<T>(
+    pub fn push_connect_timeout(
         self,
         timeout: Duration,
-    ) -> Stack<stack::MapErr<tower::timeout::Timeout<S>, impl FnOnce(Error) -> Error + Clone>>
-    where
-        S: Service<T>,
-        S::Error: Into<Error>,
-    {
```
olix0r, line 188: These constraints are unnecessary and caused us to have to document `T`. Easier to just remove the constraints.

linkerd/app/inbound/src/lib.rs
```diff
@@ -119,11 +118,21 @@ impl Inbound<()> {
...
 
+            #[derive(Debug, thiserror::Error)]
+            #[error("inbound connection must not target port {0}")]
+            struct Loop(u16);
+
             svc::stack(transport::ConnectTcp::new(*keepalive))
-                .push_map_target(|t: T| Remote(ServerAddr(([127, 0, 0, 1], t.param()).into())))
                 // Limits the time we wait for a connection to be established.
                 .push_connect_timeout(*timeout)
-                .push(svc::stack::BoxFuture::layer())
```
olix0r, line 126: This `BoxFuture` layer is redundant -- there's already a `BoxFuture` in the HTTP router where it's actually needed.

kleimkuhler (approved):
👍

mateiidavid (approved):
Looks great!

linkerd/app/inbound/src/lib.rs
```diff
@@ -285,31 +290,37 @@ where
...
-                    .check_new_service::<T, I>()
-                    .push_on_response(svc::BoxService::layer())
-                    .push(svc::BoxNewService::layer())
-            })
-            .map_stack(|_, rt, accept| {
-                accept
                     .push_switch(
-                        PreventLoop::from(server_port).to_switch(),
+                        // If the connection targets the inbound proxy port, the connection is most
+                        // likely using opaque transport to target an alternate port, or possible an
```
olix0r, line 314:
```suggestion
                        // likely using opaque transport to target an alternate port, or possibly an
```

linkerd/app/inbound/src/lib.rs
```diff
@@ -285,31 +290,37 @@ where
                         identity: rt.identity.clone(),
                     }))
             })
-            .map_stack(|cfg, _, detect| {
+            .map_stack(|cfg, rt, detect| {
```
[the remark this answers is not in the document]
olix0r: It's used below the diff (I combined two `map_stack`s into one). Our compilation settings (`#[deny(warning)]`) won't actually let us compile with unused variables :)

linkerd/app/inbound/src/lib.rs
```diff
@@ -285,31 +290,37 @@ where
...
+                            if disable_detect.contains(&addr.port()) {
+                                return Ok(svc::Either::B(TcpAccept::port_skipped(t)));
                             }
+                            Ok(svc::Either::A(t))
                         },
                         opaque
                             .instrument(|_: &TcpAccept| debug_span!("forward"))
                             .into_inner(),
                     )
-                    .check_new_service::<T, I>()
```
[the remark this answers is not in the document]
olix0r: These checks are really just compile-time assertions. They can be helpful for debugging when a stack doesn't compile, but there's no inherent benefit to having them aside from clarity. We could probably restore this without any issue, though....

PR #1174, opened 2021-07-30
http: Use ExtractParam to support dynamic detect configuration

olix0r:
Following #1169, this change modifies the `detect` middleware to use
`ExtractParam`. In doing this, it makes sense to modify the default
implementations of `ExtractParam` so that we can provide a static
configuration to the module that is cloned for all targets without
modification.

This change does not alter any functionality, but sets up to make the
HTTP detection layer configured per-target.


linkerd/detect/src/lib.rs
```diff
@@ -27,28 +27,30 @@ pub trait Detect<I>: Clone + Send + Sync + 'static {
...
+}
+
+#[derive(Copy, Clone, Debug)]
+pub struct NewDetectService<P, D, N> {
     inner: N,
-    detect: D,
-    capacity: usize,
-    timeout: time::Duration,
+    params: P,
+    _detect: std::marker::PhantomData<fn() -> D>,
```
mateiidavid, line 44: This is the only small question that I have: are we using `PhantomData` here to satisfy the compiler? Is this a way of saying that the service will live at least as long as the config itself, or am I misreading it?

I looked through the [docs](https://doc.rust-lang.org/std/marker/struct.PhantomData.html) to see what `PhantomData` does, and also searched the webs. Most people seem to point that it's used to express lifetime dependencies so just wanted to see if I get this right.

mateiidavid (approved):

linkerd/detect/src/lib.rs
olix0r: > This is the only small question that I have: are we using PhantomData here to satisfy the compiler? Is this a way of saying that the service will live at least as long as the config itself, or am I misreading it?

It's not really about lifetime in the ownership/borrowing sense and more about our ability to know the `D` type. Specifically here:

https://github.com/linkerd/linkerd2-proxy/blob/fa5c53becfa93995d7107ada229421bc37bc6bb0/linkerd/detect/src/lib.rs#L97-L111

The `NewService` implementation needs to know about the `D` type (to build the service type) -- but if we tried to be generic over `D` without it being a generic type on the _new_ type, then we'd get a compile error saying that `D` is unconstrained. `PhantomData` lets us bind a `D` type on `NewDetect` so that we can use it for the `NewService` implementation. It also means that we have to _know_ what the `D` type is when we build this layer and if it's ambiguous, we may need to document it explicitly.

It's generally best to avoid marker types like this if at all possible, but they're a necessary evil in situations like this.

Also, we use a marker type like `PhatnomData<fn() -> D>` (instead of just `PhantomData<D>` because this impacts the `Send`/etc requirements on `D` -- `NewDetect` can be `Send` (etc) without `D` having to be documented as such. I think about it as "we are able to produce a `D`" -- because that's what `extract_param` does here.

linkerd/detect/src/lib.rs
```diff
@@ -59,52 +61,67 @@ pub fn allow_timeout<P, T>((p, t): (DetectResult<P>, T)) -> (Option<P>, T) {
     }
 }
 
+// === impl Config ===
+
+impl<D: Default> Config<D> {
+    const DEFAULT_CAPACITY: usize = 1024;
+
+    pub fn from_timeout(timeout: time::Duration) -> Self {
```
hawkw, line 69: very minor nit: probably would have called this `with_timeout` since i would think of it as more of a 'constructor' than a 'conversion' :woman_shrugging:

linkerd/detect/src/lib.rs
```diff
@@ -116,15 +133,18 @@ where
...
+        let Config {
+            detect,
+            capacity,
+            timeout,
+        } = self.config.clone();
         let target = self.target.clone();
-        let timeout = self.timeout;
+        let mut inner = self.inner.clone();
         Box::pin(async move {
             trace!("Starting protocol detection");
```
hawkw, line 144: nit, take it or leave it: now that the config parameters can be dynamic, maybe this event should include them:
```suggestion
            trace!(
                config.capacity = capacity,
                config.timeout = ?timeout,
                "Starting protocol detection"
            );
```

hawkw (approved):
looks good to me, commented on a couple of extremely minor nits.

PR #1180, opened 2021-08-02
Tear down client connections when responses contain `l5d-proxy-error` header

kleimkuhler:
Currently outbound proxy connections are not torn down when communicating with an inbound proxy that encounters a connection error with its application.

In order to support this behavior, a two-part change is required:
1. Add the `L5D-PROXY-ERROR` header. It is added to responses by an inbound proxy when it encounters a connection error with its application (closed by #1119)
2. Tear down outbound connections when responses contain the `L5D-PROXY-ERROR` header

A new `NewHandleProxyError` layer is introduced which is added to the outbound stack; it looks for responses that container the header. When a response does container the header, it uses the `ClientHandle` that is added to the request extensions to close the client connection.

A test has been added which exercises this behavior and asserts that the client connection is closed after receiving a response that contains the header.


linkerd/app/outbound/src/http/handle_proxy_error.rs
```diff
@@ -0,0 +1,195 @@
...
+        let client = req.extensions().get::<ClientHandle>().cloned();
+        debug_assert!(client.is_some(), "Missing client handle");
+        let response = self.inner.call(req);
+        Box::pin(async move {
+            match response.await {
+                Ok(rsp) => {
+                    if let Some(message) =
+                        rsp.headers().get(linkerd_app_core::errors::L5D_PROXY_ERROR)
+                    {
+                        tracing::info!(?message, "response contained L5D_PROXY_ERROR header");
```
hawkw, line 72: very minor nit: what if we interpolate the actual value of the header constant into this message?
```suggestion
                        tracing::info!(
                            ?message,
                            "response contained `{}` header",
                            linkerd_app_core::errors::L5D_PROXY_ERROR,
                        );
```

linkerd/app/outbound/src/http/handle_proxy_error.rs
```diff
@@ -0,0 +1,195 @@
...
+                        // Gracefully teardown the accepted connection.
+                        if let Some(ClientHandle { close, .. }) = client {
+                            tracing::info!("connection closed");
```
hawkw, line 75: not sure if we should have _two_ log lines at the info level for this, since those will be disabled by default. i would probably add something like "response contained ... header, closing connection" to the other log line, and move this one to the debug or trace level just to indicate that the connection was successfully closed for debugging purposes (or remove it entirely?).

linkerd/app/outbound/src/http/handle_proxy_error.rs
```diff
@@ -0,0 +1,195 @@
...
+where
+    S: Service<Request<A>, Response = Response<B>>,
+    S::Response: Send,
+    S::Error: Into<Error> + Send,
+    S::Future: Send + 'static,
+    A: Send + 'static,
+    B: Send + 'static,
+{
+    type Response = S::Response;
+    type Error = Error;
```
hawkw, line 55: I think if we just changed our `Error` type to `S::Error`, we could:
- remove the `S::Error: Into<Error>` bound, 
- remove the `.map_err(Into::into)` in `poll_ready`
- remove the `e.into()` in `call`

which would make this code a bit simpler. as far as i can tell, we never actually return our own errors in this middleware, so it's fine to just return whatever the inner service's error type is.

linkerd/app/outbound/src/http/handle_proxy_error.rs
```diff
@@ -0,0 +1,195 @@
...
+                        // Gracefully teardown the accepted connection.
+                        if let Some(ClientHandle { close, .. }) = client {
+                            tracing::info!("connection closed");
+                            close.close();
+                        }
+                    }
+                    Ok(rsp)
+                }
+                Err(e) => Err(e.into()),
+            }
```
hawkw, line 82: nit, take it or leave it: i think we can simplify this a bit by using the question mark operator rather than `match` here:
```suggestion
            let rsp = response.await?;
            if let Some(message) =
                rsp.headers().get(linkerd_app_core::errors::L5D_PROXY_ERROR)
            {
                tracing::info!(?message, "response contained L5D_PROXY_ERROR header");
                // Gracefully teardown the accepted connection.
                if let Some(ClientHandle { close, .. }) = client {
                    tracing::info!("connection closed");
                    close.close();
                }
            }
            Ok(rsp)
```

hawkw (changes requested):
it looks like a number of integration tests are currently failing because the client handle extension is not present: https://github.com/linkerd/linkerd2-proxy/pull/1180/checks?check_run_id=3224409790#step:3:902. i noticed that [a majority of the failing tests involve retries](https://github.com/linkerd/linkerd2-proxy/pull/1180/checks?check_run_id=3224409790#step:3:913). i think the problem here is that when requests are cloned for retries, the `Extensions` of the original request are not cloned: https://github.com/linkerd/linkerd2-proxy/blob/1cff3aef82c203bf09ccce485506d7a29ca27308/linkerd/app/core/src/retry.rs#L125-L142

therefore, the retry request will have a new, empty set of `Extensions`, and be missing a `ClientHandle`. when those retry requests hit this middleware and it tries to clone the client handle, it doesn't find one, and hits the assertion that expects the `ClientHandle` to be present.

i believe we can fix this by updating the `clone_request` method to also clone the `ClientHandle` from the original request's `Extensions` and add it to the clone's extensions before returning it.

in addition, i also commented on a couple smaller style issues you may want to address.

linkerd/app/outbound/src/http/handle_proxy_error.rs
kleimkuhler: Oh good catch yea I'll make that change!

linkerd/app/outbound/src/http/handle_proxy_error.rs
```diff
@@ -0,0 +1,195 @@
...
+};
+use linkerd_proxy_http::ClientHandle;
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+#[derive(Clone, Debug)]
+pub struct NewHandleProxyError<N> {
```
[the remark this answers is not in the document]
kleimkuhler: Yep that's a good point I'll rename to PeerProxyErrors

linkerd/app/outbound/src/http/handle_proxy_error.rs
kleimkuhler: I made these separate because if for some reason the request does not have the client handle, the connection will not be closed. So the first line handles the fact that the `L5D-PROXY-ERROR` is set, and the other is when we actually close the connection.

The request _should_ have the client handle, but if for some reason it doesn't it could be misleading to say the connection is closed.

> not sure if we should have two log lines at the info level for this, since those will be disabled by default

Won't `info` level be enabled by default? I thought these may be important enough to be `info` instead of `debug`, but I'm also fine changing it!

linkerd/app/core/src/retry.rs
```diff
@@ -138,6 +139,12 @@ where
         *clone.headers_mut() = req.headers().clone();
         *clone.version_mut() = req.version();
 
+        // If the ClientHandle extension is present, clone it into the new
+        // request.
```
hawkw, line 143: i would prefer the comment to explain _why_ we are doing this, rather than _what_ we are doing ---  "if the client handle is present, clone it into the new request" is actually pretty obvious from the code, but something like this is less obvious:
```suggestion
        // The cloned request may also need the ability to tear down the
        // client connection if the remote proxy returns a `l5d-proxy-error`
        // header, so ensure it also has a `ClientHandle`.
        // If the ClientHandle extension is present, clone it into the new
        // request.
```

linkerd/app/outbound/src/http/endpoint.rs
```diff
@@ -39,6 +40,7 @@ impl<C> Outbound<C> {
                 .push_on_response(svc::MapErrLayer::new(Into::<Error>::into))
                 .check_service::<T>()
                 .into_new_service()
+                .push(peer_proxy_errors::PeerProxyErrors::layer())
```
hawkw, line 43: :+1: for the renaming

linkerd/app/outbound/src/http/peer_proxy_errors.rs
```diff
@@ -0,0 +1,186 @@
...
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+/// PeerProxyErrors will close the accepted connection if the response is from
+/// a peer proxy and the L5D_PROXY_ERROR header is set. This means the peer
+/// proxy encountered an inbound connection error with its application and
+/// therefore the accepted connection should be torn down.
```
hawkw, line 16: very minor style nit: including the name of the documented item in the doc comment is a Go idiom; in rust we generally omit it
```suggestion
/// Closes the accepted connection if the response is from a peer
/// proxy and the L5D_PROXY_ERROR header is set. This means the
/// peer proxy encountered an inbound connection error with
/// its application and therefore the accepted connection should
/// be torn down.
```

linkerd/app/outbound/src/http/peer_proxy_errors.rs
```diff
@@ -0,0 +1,186 @@
...
+        let client = req.extensions().get::<ClientHandle>().cloned();
+        debug_assert!(client.is_some(), "Missing client handle");
+        let response = self.inner.call(req);
+        Box::pin(async move {
+            let response = response.await?;
+            if let Some(message) = response.headers().get(L5D_PROXY_ERROR) {
+                tracing::info!(?message, "response contained `{}` header", L5D_PROXY_ERROR);
+                // Gracefully teardown the accepted connection.
+                if let Some(ClientHandle { close, .. }) = client {
+                    tracing::info!("connection closed");
```
hawkw, line 70: this should probably not be info
```suggestion
                    tracing::trace!("connection closed");
```

linkerd/app/outbound/src/http/handle_proxy_error.rs
hawkw: Sorry, I meant to say that info is enabled by default.

If we're so concerned about cases where the `ClientHandle` is missing that it's worth having two logs at the `INFO` level for it...why don't we just remove the `debug_assert!` that it's present, and `expect` it when extracting it from the request's `Extensions`? if it's ever not set, that's _definitely_ a bug, and shouldn't ever be possible, so it's fine to just panic in that case. Then, we don't have to worry about logging the presence of the header and the successful closure of the connection separately.

linkerd/app/core/src/retry.rs
kleimkuhler: Great point

linkerd/app/outbound/src/http/handle_proxy_error.rs
kleimkuhler: That makes sense. The `debug_asssert!` came from the other location where we expect a `ClientHandle`, but since it should always be set here adding an `expect` sounds good.

hawkw (approved):
seems good to me — let's see what @olix0r has to say?

linkerd/app/outbound/src/http/peer_proxy_errors.rs
```diff
@@ -0,0 +1,186 @@
...
+        let message = response
+            .headers()
+            .get(L5D_PROXY_ERROR)
+            .expect("response did not contain l5d-proxy-error header");
+        assert_eq!(message, "proxy received invalid response");
+
+        // The client's background task should be completed without dropping
+        // the client because the connection was closed after encountering a
+        // response that contains the l5d-proxy-error header.
+        let _ = bg.await;
```
olix0r, line 147: I don't understand what this is actually testing or how it's wired up with the client handle.

Specifically https://github.com/linkerd/linkerd2-proxy/pull/1183 changes the client handle test utility to return the close future; and the close future never fires? Why is that? How does the background task get notified about the client handle signaling a close?

olix0r: Or, presumably `NewServeHttp` adds a client handle? If so, why do we have to call `add_client_handle` in the test?

olix0r: It may be better to just narrow this test to only test the behavior in this module without actually initiating TCP clients and servers, and the whole stack. i.e. just test that close future fires.

olix0r: https://github.com/linkerd/linkerd2-proxy/pull/1183/commits/7e990193c81ecd0417f8e1c8bc0fb95722988ae5

linkerd/app/outbound/src/http/endpoint.rs
olix0r: Is there any particular reason this has to exist on the endpoint stack?

In https://github.com/linkerd/linkerd2-proxy/pull/1183/commits/c03b0493589bea172d26c373d52047c0037cbbfc, I've moved it into the server stack which gives us the advantage of not having to worry about retries and, more importantly, eliminates the need to add a clienthandle onto all of our endpoint-stack tests. I think this move may fit a bit nicer as then this logic can live right next to the other error-handling module

olix0r (approved):

linkerd/app/outbound/src/http/peer_proxy_errors.rs
```diff
@@ -0,0 +1,130 @@
...
+    type Error = S::Error;
+    type Future = Pin<Box<dyn Future<Output = Result<S::Response, S::Error>> + Send + 'static>>;
+
+    #[inline]
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: http::Request<A>) -> Self::Future {
+        let client = req
```
olix0r, line 57:
```suggestion
        let ClientHandle { close, .. } = req
```

linkerd/app/outbound/src/http/peer_proxy_errors.rs
```diff
@@ -0,0 +1,130 @@
...
+        Box::pin(self.inner.call(req).map_ok(move |rsp| {
+            if let Some(msg) = rsp.headers().get(L5D_PROXY_ERROR) {
+                tracing::debug!(
+                    ?msg,
+                    "Received an error response from a peer proxy; closing connection"
+                );
+
+                // Signal that the proxy's server-side connection should be terminated. This handles
+                // the remote error as if the local proxy encountered an error.
+                let ClientHandle { close, .. } = client;
```
olix0r, line 72:
```suggestion
```

olix0r (commented):
minor suggestion: just destructure the ClientHandle as you extract it

PR #1183, opened 2021-08-05
Suggestions for #1180

olix0r:
Suggestions for #1180 

* `outbound` should not pull in `proxy_http`. Instead, we use re-exports from `app_core` like `proxy::http`.
* We don't want to expose inner workings of how a `ClientHandle` is constructed. We have the `Close` & `Closed` types that hide the inner `Arc<Notify>`. This change adds `ClientHandle::new` to build these types without exposing the inner impl.
* I've renamed `add_client_handle` to `set_client_handle` and changed it to take a mutable reference and, rather than return the request, we return a future that fires when the client handle is closed.
* As discussed in https://github.com/linkerd/linkerd2-proxy/pull/1180/files#r683052774, the test seems a bit more complicated than it needs to be. I'd rather have a narrow unit test that exercises the behavior we care about without having to instrument a full proxy with TCP connections, etc. I've updated the test to only exercise the `PeerProxyErrors` logic without all of the extra proxy test scaffolding.

### DIFF
--- a/linkerd/app/outbound/src/http/peer_proxy_errors.rs
+++ b/linkerd/app/outbound/src/http/peer_proxy_errors.rs
@@ -86,40 +86,42 @@ mod test {
         Infallible,
     };
     use linkerd_tracing::test;
+    use tokio::time;
 
     #[tokio::test(flavor = "current_thread")]
     async fn connection_closes_after_response_header() {
         let _trace = test::trace_init();
 
-        let service = PeerProxyErrors::new(svc::mk(move |_: http::Request<hyper::Body>| {
-            let response = http::Response::builder()
-                .status(http::StatusCode::BAD_GATEWAY)
-                .header(L5D_PROXY_ERROR, "proxy received invalid response")
-                .body(hyper::Body::default())
-                .unwrap();
-            future::ok::<_, Infallible>(response)
-        }));
-
         // Build the client that should be closed after receiving a response
         // with the l5d-proxy-error header.
-        let mut request = http::Request::builder()
+        let mut req = http::Request::builder()
             .uri("http://foo.example.com")
             .body(hyper::Body::default())
             .unwrap();
-        let closed = test_util::set_client_handle(&mut request, ([192, 0, 2, 3], 50000).into());
-        let response = service
-            .oneshot(request)
-            .await
-            .expect("request must succeed");
-        assert_eq!(response.status(), http::StatusCode::BAD_GATEWAY);
-        let message = response
-            .headers()
-            .get(L5D_PROXY_ERROR)
-            .expect("response did not contain l5d-proxy-error header");
-        assert_eq!(message, "proxy received invalid response");
+        let closed = test_util::set_client_handle(&mut req, ([192, 0, 2, 3], 50000).into());
+
+        const ERROR_MSG: &str = "something bad happened";
+        let svc = PeerProxyErrors::new(svc::mk(move |_: http::Request<hyper::Body>| {
+            future::ok::<_, Infallible>(
+                http::Response::builder()
+                    .status(http::StatusCode::BAD_GATEWAY)
+                    .header(L5D_PROXY_ERROR, ERROR_MSG)
+                    .body(hyper::Body::default())
+                    .unwrap(),
+            )
+        }));
+
+        let rsp = svc.oneshot(req).await.expect("request must succeed");
+        assert_eq!(rsp.status(), http::StatusCode::BAD_GATEWAY);
+        assert_eq!(
+            rsp.headers()
+                .get(L5D_PROXY_ERROR)
+                .expect("response did not contain l5d-proxy-error header"),
+            ERROR_MSG
+        );
 
         // The client handle close future should fire.
-        tokio::time::timeout(tokio::time::Duration::from_secs(10), closed)
+        time::timeout(time::Duration::from_secs(10), closed)
             .await
             .expect("client handle must close");
     }
--- a/linkerd/app/outbound/src/http/peer_proxy_errors.rs
+++ b/linkerd/app/outbound/src/http/peer_proxy_errors.rs
@@ -42,7 +42,6 @@ where
     S: svc::Service<http::Request<A>, Response = http::Response<B>>,
     S::Response: Send,
     S::Future: Send + 'static,
-    A: Send + 'static,
     B: Send + 'static,
 {
     type Response = S::Response;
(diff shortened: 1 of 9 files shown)


linkerd/app/outbound/src/http/peer_proxy_errors.rs
```diff
@@ -1,9 +1,5 @@
-use http::{Request, Response};
```
olix0r, line 1: we tend not to import Request/Response -- referencing them via, e.g., `http::Request` isn't that much more verbose and is a little more readable -- there's no ambiguity about whether we're dealing with some other type of request/response.

linkerd/app/outbound/src/http/peer_proxy_errors.rs
```diff
@@ -24,14 +20,14 @@ impl<N> PeerProxyErrors<N> {
...
     }
 
-    pub fn layer() -> impl layer::Layer<N, Service = Self> + Clone + Copy {
-        layer::mk(Self::new)
+    pub fn layer() -> impl svc::layer::Layer<N, Service = Self> + Clone + Copy {
+        svc::layer::mk(Self::new)
     }
 }
 
-impl<T, N> NewService<T> for PeerProxyErrors<N>
```
olix0r, line 32: similarly, elsewhere in the code we tend to use `svc::NewService/Service`. The `svc` prefix isn't too verbose. I think we mostly do this because we used to have more structs named `Service` and `NewService`. This isn't so prevalent anymore but we still tend to do this elsewhere.

linkerd/app/outbound/src/http/peer_proxy_errors.rs
```diff
@@ -41,146 +37,89 @@ where
...
     S::Response: Send,
     S::Future: Send + 'static,
-    A: Send + 'static,
     B: Send + 'static,
 {
     type Response = S::Response;
     type Error = S::Error;
     type Future = Pin<Box<dyn Future<Output = Result<S::Response, S::Error>> + Send + 'static>>;
 
+    #[inline]
```
olix0r, line 51: We try to use `#[inline]` with trivial passthru implementations, especially for `poll_ready`.  Basically: there's no use seeing this scope in backtraces.

linkerd/app/outbound/src/http/peer_proxy_errors.rs
```diff
@@ -41,146 +37,89 @@ where
...
 }
 
-impl<S, A, B> Service<Request<A>> for PeerProxyErrors<S>
+impl<S, A, B> svc::Service<http::Request<A>> for PeerProxyErrors<S>
 where
-    S: Service<Request<A>, Response = Response<B>>,
+    S: svc::Service<http::Request<A>, Response = http::Response<B>>,
     S::Response: Send,
     S::Future: Send + 'static,
-    A: Send + 'static,
```
olix0r, line 49: This type constraint wasn't needed

linkerd/app/outbound/src/http/peer_proxy_errors.rs
```diff
@@ -41,146 +37,89 @@ where
...
         debug_assert!(client.is_some(), "Missing client handle");
-        let response = self.inner.call(req);
-        Box::pin(async move {
-            let response = response.await?;
-            if let Some(message) = response.headers().get(L5D_PROXY_ERROR) {
-                tracing::info!(?message, "response contained `{}` header", L5D_PROXY_ERROR);
+
+        Box::pin(self.inner.call(req).map_ok(move |rsp| {
+            if let Some(msg) = rsp.headers().get(L5D_PROXY_ERROR) {
+                tracing::debug!(header = %L5D_PROXY_ERROR, ?msg);
```
olix0r, line 62: Moved this to debug-- `info` logs are going to be user-facing (and may cost users money to store). It's not clear what a user should do when this situation is encountered, so it's more of a diagnostic, i think

linkerd/app/outbound/src/http/peer_proxy_errors.rs
```diff
@@ -41,146 +37,89 @@ where
...
                 // Gracefully teardown the accepted connection.
                 if let Some(ClientHandle { close, .. }) = client {
                     tracing::trace!("connection closed");
                     close.close();
                 }
             }
-            Ok(response)
-        })
+
+            rsp
```
olix0r, line 71: Generally, we use req/rsp for request and response. Especially when the variable is short-lived. Basically: variables with short scopes get shorter names. Important variables with large scopes get longer names.

linkerd/app/outbound/src/http/peer_proxy_errors.rs
```diff
@@ -41,146 +37,89 @@ where
...
-        // Build the outbound server that responds with the l5d-proxy-error
-        // header set.
-        let (rt, _shutdown) = test_util::runtime();
-        let addr = SocketAddr::new([127, 0, 0, 1].into(), 12345);
-        let connect = support::connect().endpoint_fn_boxed(addr, |_: Endpoint| serve());
-        let config = test_util::default_config();
-        let mut outbound = build_outbound(config, rt, connect);
-
-        // Build the otubound service.
-        let service = outbound.new_service(Endpoint {
```
olix0r, line 117: One important advantage to the test changes is that this test won't have to be updated as the outbound stack (or target types) change.

linkerd/app/outbound/src/http/peer_proxy_errors.rs
```diff
@@ -41,146 +37,89 @@ where
...
+        assert_eq!(rsp.status(), http::StatusCode::BAD_GATEWAY);
+        assert_eq!(
+            rsp.headers()
+                .get(L5D_PROXY_ERROR)
+                .expect("response did not contain l5d-proxy-error header"),
+            ERROR_MSG
+        );
+
+        // The client handle close future should fire.
+        time::timeout(time::Duration::from_secs(10), closed)
```
olix0r, line 121: It's important to use timeouts when the default (error) behavior would be to hang indefinitely

olix0r (commented):
some commentary

linkerd/app/outbound/src/http/peer_proxy_errors.rs
kleimkuhler: Using `closed` to test that the client handle is closed instead of going through the full setup of a client and having to satisfy the type constraints for `http_util::connect_and_accept` is much simpler; good call.

linkerd/app/outbound/src/http/server.rs
```diff
@@ -45,6 +47,8 @@ impl<N> Outbound<N> {
                         .push(svc::FailFast::layer("HTTP Server", dispatch_timeout))
                         .push_spawn_buffer(buffer_capacity)
                         .push(rt.metrics.http_errors.clone())
+                        // Tear down server connections when a peer proxy generates an error.
+                        .push(PeerProxyErrors::layer())
```
kleimkuhler, line 51: 👍

kleimkuhler (approved):

PR #1184, opened 2021-08-05
inbound: support configurable default port policies

hawkw:
The proxy supports a [default port policy][1]. We currently hardcode a
[default policy permitting all connections][2]. We should make this
behavior configurable. This will enable the proxy-injector to set
default behavior for each workload and, more imminently, we will be able
to configure the identity proxy's behavior to require TLS. See
linkerd/linkerd2#6606 for details.

This branch adds a `LINKERD2_PROXY_INBOUND_DEFAULT_POLICY` environment
configuration supporting the following values:

* `authenticated`
* `deny`
* `tis-unauthenticated`
* `unauthenticated`

This was implemented by:

* Adding a new `TlsUnauthenticated` variant to `AllowPolicy`. This
  variant allows all TLS connections, regardless of whether they have
  valid client identities.
* Adding `FromStr` implementations for `AllowPolicy` and
  `DefaultPolicy` to allow them to be parsed from the environment
  variable.
* Changing the env module to parse the default policy from the env var,
  rather than hardcoding it.

Closes linkerd/linkerd2#6606

[1]: https://github.com/linkerd/linkerd2-proxy/blob/2eb1671f0fde62dae5fb272608a276ad21a598ba/linkerd/app/inbound/src/port_policies.rs#L14-L26
[2]: https://github.com/linkerd/linkerd2-proxy/blob/2eb1671f0fde62dae5fb272608a276ad21a598ba/linkerd/app/src/env.rs#L526

### DIFF
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -75,6 +75,12 @@ pub enum ParseError {
     InvalidTokenSource,
     #[error("invalid trust anchors")]
     InvalidTrustAnchors,
+    #[error("not a valid port policy: {0}")]
+    NotAPortPolicy(
+        #[from]
+        #[source]
+        inbound::port_policies::ParsePolicyError,
+    ),
 }
 
 // Environment variables to look at when loading the configuration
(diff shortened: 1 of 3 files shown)


linkerd/app/src/env.rs
```diff
@@ -522,8 +536,14 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
                 }
             }
 
+            let default = parse(strings, ENV_INBOUND_DEFAULT_POLICY, |s| {
+                s.parse::<inbound::port_policies::DefaultPolicy>()
+                    .map_err(ParseError::from)
```
olix0r, line 541:
```suggestion
                s.parse().map_err(ParseError::from)
```

can be inferred

olix0r (approved):

PR #1185, opened 2021-08-05
inbound: test `AllowPolicy` variants that require TLS

hawkw:
There's currently [a TODO comment][1] in `inbound/detect.rs` about
adding tests for testing `AllowPolicy::Authenticated`, which is not
currently tested there because it requires mocking out TLS detection. PR
#1184 added an additional `AllowPolicy`, `TlsUnauthenticated`, which
also requires TLS detection to be simulated, so there are now two
untested cases.

This branch adds tests for the `AllowPolicy` variants that require TLS
detection. This is implemented by splitting out a (module-private)
`push_detect_http` method from `push_detect`, which pushes everything
below the TLS detection & TLS switch layer. This is then called by
`push_detect` after pushing the TLS detection layers. This way, we can
now write tests in the module where a `ConditionalServerTls` is passed
in, to test whether the `AllowPolicy` variants do the right thing with
different TLS detection results.

The new tests were initially written by writing out all the code
procedurally for each test, and then refactoring them into a table
driven test. Each test just configures the inbound stack and then calls
a helper method with an array of inputs (`ConditionalServerTls`
detection results) and expected outputs (should the connection be
allowed or denied?), and an `AllowPolicy`. The test function then runs
each case and asserts that the connection is allowed or denied depending
on the input array.

[1]: https://github.com/linkerd/linkerd2-proxy/blob/2706350ae060390c3ed20598aacb293a6996b80b/linkerd/app/inbound/src/detect.rs#L238

### DIFF
--- a/linkerd/app/inbound/Cargo.toml
+++ b/linkerd/app/inbound/Cargo.toml
@@ -14,7 +14,6 @@ bytes = "1"
 http = "0.2"
 futures = { version = "0.3", default-features = false }
 linkerd-app-core = { path = "../core" }
-linkerd-error = { path = "../../error" }
 thiserror = "1.0"
 tokio = { version = "1", features = ["sync"] }
 tower = { version = "0.4.8", features = ["util"] }
--- a/linkerd/app/inbound/Cargo.toml
+++ b/linkerd/app/inbound/Cargo.toml
@@ -14,6 +14,7 @@ bytes = "1"
 http = "0.2"
 futures = { version = "0.3", default-features = false }
 linkerd-app-core = { path = "../core" }
+linkerd-error = { path = "../../error" }
 thiserror = "1.0"
 tokio = { version = "1", features = ["sync"] }
 tower = { version = "0.4.8", features = ["util"] }
--- a/linkerd/app/inbound/src/detect.rs
+++ b/linkerd/app/inbound/src/detect.rs
@@ -276,6 +276,7 @@ mod tests {
 
     const HTTP: &[u8] = b"GET / HTTP/1.1\r\nhost: example.com\r\n\r\n";
     const NOT_HTTP: &[u8] = b"foo\r\nbar\r\nblah\r\n";
+
     #[tokio::test(flavor = "current_thread")]
     async fn skip_detect() {
         let (io, _) = io::duplex(1);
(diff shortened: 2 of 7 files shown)


linkerd/app/inbound/src/detect.rs
```diff
@@ -60,10 +60,56 @@ impl<N> Inbound<N> {
     {
         const TLS_PORT_SKIPPED: tls::ConditionalServerTls =
             tls::ConditionalServerTls::None(tls::NoServerTls::PortSkipped);
+        self.push_detect_http(forward.clone())
```
olix0r, line 63: Do we need to push detect_http here? would it be better to keep them separate and compose them only in the main stack?

linkerd/app/inbound/Cargo.toml
```diff
@@ -14,6 +14,7 @@ bytes = "1"
 http = "0.2"
 futures = { version = "0.3", default-features = false }
 linkerd-app-core = { path = "../core" }
+linkerd-error = { path = "../../error" }
```
olix0r, line 17:
```suggestion
```

We should pick this up from the reexports in core instead

linkerd/app/inbound/src/detect.rs
hawkw: We could do that. I erred on the side of making this change more self-contained, and just splitting it out into a private method so we can write new tests in this module. If you'd prefer to have separate detect TLS and detect HTTP public methods, we could do that instead!

linkerd/app/inbound/src/detect.rs
```diff
@@ -235,24 +261,28 @@ impl<T> svc::InsertParam<tls::ConditionalServerTls, T> for TlsParams {
...
+        is_error,
         svc::{NewService, ServiceExt},
-        Error,
+        trace, Error,
     };
+    use std::str::FromStr;
 
+    const HTTP: &[u8] = b"GET / HTTP/1.1\r\nhost: example.com\r\n\r\n";
+    const NOT_HTTP: &[u8] = b"foo\r\nbar\r\nblah\r\n";
     #[tokio::test(flavor = "current_thread")]
```
olix0r, line 279:
```suggestion

    #[tokio::test(flavor = "current_thread")]
```

olix0r (approved):

PR #1187, opened 2021-08-05
Remove unsafety from setting TcpStream keepalive

kleimkuhler:
As noted in the todo, once tokio's `TcpStream::into_std` merged we now have a way to convert `tokio` -> `std` `TcpStream`s and use `socket2`'s `Socket::From<TcpStream>` impl to safely set a socket's keepalive option.

### DIFF
--- a/linkerd/proxy/transport/src/lib.rs
+++ b/linkerd/proxy/transport/src/lib.rs
@@ -3,7 +3,7 @@
 //! Uses unsafe code to interact with socket options for keepalive and SO_ORIGINAL_DST.
 
 #![deny(warnings, rust_2018_idioms)]
-//#![forbid(unsafe_code)]
+#![forbid(unsafe_code)]
 
 pub mod addrs;
 mod connect;
--- a/linkerd/proxy/transport/src/lib.rs
+++ b/linkerd/proxy/transport/src/lib.rs
@@ -39,7 +39,10 @@ fn set_nodelay_or_warn(socket: &TcpStream) {
     }
 }
 
-fn set_keepalive(tcp: TcpStream, keepalive_duration: Option<Duration>) -> io::Result<TcpStream> {
+fn set_keepalive_or_warn(
+    tcp: TcpStream,
+    keepalive_duration: Option<Duration>,
+) -> io::Result<TcpStream> {
     let sock = {
         let stream = tokio::net::TcpStream::into_std(tcp)?;
         socket2::Socket::from(stream)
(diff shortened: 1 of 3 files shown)


linkerd/proxy/transport/src/lib.rs
```diff
@@ -38,47 +39,15 @@ fn set_nodelay_or_warn(socket: &TcpStream) {
...
     let ka = keepalive_duration
         .into_iter()
         .fold(TcpKeepalive::new(), |k, t| k.with_time(t));
-    if let Err(e) = sock.set_tcp_keepalive(&ka) {
-        tracing::warn!("failed to set keepalive: {}", e);
-    }
-
-    // Don't let the socket2 socket close the fd on drop!
-    std::mem::forget(sock);
+    sock.set_tcp_keepalive(&ka)?;
```
hawkw, line 50: I'm unsure about the error handling change here. I think the previous behavior (logging a warning rather than closing the connection) was intentional, and this is a change in behavior.

Is it possible to make the change that removes the use of unsafe APIs without also changing the error handling behavior?

hawkw (commented):
Overall, removing the use of unsafe APIs seems great!. I'm not sure if we want to change how we handle errors as part of that change.

Also, we should check if it's possible to add a `#![deny(unsafe_code)]` to this module now that we no longer use the unsafe socket conversions?

linkerd/proxy/transport/src/lib.rs
kleimkuhler: Yep I added back the warning. If this fails, we'll still return the TCP stream after converting it back.

kleimkuhler:
> Also, we should check if it's possible to add a #![deny(unsafe_code)] to this module now that we no longer use the unsafe socket conversions?

We still use `unsafe` in `orig_dst.rs` so adding back `#![forbid(unsafe_code)]` at the module level results in errors.

hawkw:
> > Also, we should check if it's possible to add a #![deny(unsafe_code)] to this module now that we no longer use the unsafe socket conversions?
> 
> We still use `unsafe` in `orig_dst.rs` so adding back `#![forbid(unsafe_code)]` at the module level results in errors.

ah, yup; forgot these were in the same crate. Nevermind!

linkerd/proxy/transport/src/lib.rs
```diff
@@ -3,7 +3,7 @@
 //! Uses unsafe code to interact with socket options for keepalive and SO_ORIGINAL_DST.
```
hawkw, line 3: should maybe update this comment
```suggestion
//! Uses unsafe code to interact with socket options for SO_ORIGINAL_DST.
```

hawkw (approved):
okay, this looks good to me now, thank you!

PR #1191, opened 2021-08-09
fuzzing: change `linkerd_dns::fuzz_target_1` to single-threaded

hawkw:
Something about the behavior of the fuzz logic for `linkerd_dns` is
hitting [a recently added assertion][1] in Tokio's multithreaded runtime
when shutting down. It's unclear what exactly triggers the assertion,
but it appears to be environment-dependent in some way: I can't
reproduce the failure on my machine using the reproducer generated by
`cargo-fuzz`, but it fails reliably in cluster-fuzz (possibly related to
the number of CPU cores on the cluster-fuzz servers?).

Since the purpose of this fuzz test is to fuzz the DNS name parsing
behavior, and this failure isn't actually related to the code being
fuzzed, just the environment the fuzz test runs in, this branch changes
the fuzz logic to use a single-threaded Tokio runtime. We'll continue
trying to get a minimal repro of the crash to report upstream, but this
should fix the fuzz test so that it can continue exercizing the DNS code
we actually care about here.

[1]: https://github.com/tokio-rs/tokio/blob/362df5a3172f6e1bdee2fd3808e5cfc730a111f6/tokio/src/runtime/thread_pool/worker.rs#L603

### DIFF
--- a/linkerd/app/inbound/fuzz/fuzz_targets/fuzz_target_1.rs
+++ b/linkerd/app/inbound/fuzz/fuzz_targets/fuzz_target_1.rs
@@ -13,7 +13,10 @@ fuzz_target!(|requests: Vec<http_fuzz::HttpRequestSpec>| {
         return;
     }
 
-    tokio::runtime::Runtime::new()
+    tokio::runtime::Builder::new_current_thread()
+        .enable_time()
+        .enable_io()
+        .build()
         .unwrap()
         .block_on(http_fuzz::fuzz_entry_raw(requests));
 });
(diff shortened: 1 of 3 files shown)


olix0r (commented):
Should we do this for all of our fuzzers?

hawkw:
https://github.com/tokio-rs/tokio/pull/4005/files/74acb2837826bbc2d2ea71457672af678e52e617#diff-89b1830bf4deaa3d4f631fa6f55970975adf4810e2d825ebdd79f877b9b0fcc9 should fix the upstream bug.

olix0r (approved):

PR #1192, opened 2021-08-10
inbound: Introduce client authorization policies

olix0r:
linkerd/linkerd2#6485 introduces a new discovery API that supports
richer authorization policies for inbound connections. This change
introduces a `server-policy` crate that models the types in this API (to
be built from gRPC types in a followup).  These types are used to
replace the simplistic `AllowPolicy` types, supporting full client
authorization policies in the inbound proxy.

The `inbound::detect` tests have been narrowed/simplified and
authorization checks have been moved into the `port_policies` module to
exercise only the relevant authorization code.


linkerd/app/inbound/src/port_policies.rs
```diff
@@ -36,36 +43,34 @@ pub enum AllowPolicy {
...
 
 #[derive(Clone, Debug, Error)]
 #[error("connection denied on unknown port {0}")]
-pub struct DeniedUnknownPort(u16);
+pub(crate) struct DeniedUnknownPort(u16);
 
-#[derive(Clone, Debug, Error, PartialEq, Eq)]
-#[error("expected one of `deny`, `authenticated`, `unauthenticated`, or `tls-unauthenticated`")]
-pub struct ParsePolicyError(());
+#[derive(Debug, thiserror::Error)]
```
hawkw, line 52: nit: `Error` is already imported
```suggestion
#[derive(Debug, Error)]
```

linkerd/app/inbound/src/port_policies.rs
```diff
@@ -76,52 +81,120 @@ impl From<DefaultPolicy> for PortPolicies {
...
-    fn from(default: AllowPolicy) -> Self {
+impl From<ServerPolicy> for PortPolicies {
+    fn from(default: ServerPolicy) -> Self {
         DefaultPolicy::from(default).into()
     }
 }
 
-// === impl DefaultPolicy ===
+impl PortPolicies {
+    pub(crate) fn check_allowed(
```
hawkw, line 91: nit/tioli: it might be nice if there was some kind of comment someplace explaining the relationship between `check_allowed` and `check_authorized` here?

linkerd/app/src/env.rs
```diff
@@ -537,18 +537,36 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
-                s.parse().map_err(ParseError::from)
+                parse_default_policy(s, detect_protocol_timeout)
             })?
-            .unwrap_or_default();
-
-            const ALLOW_OPAQUE: inbound::AllowPolicy =
-                inbound::AllowPolicy::Unauthenticated { skip_detect: true };
+            .unwrap_or_else(|| {
+                parse_default_policy("unauthenticated", detect_protocol_timeout)
+                    .expect("unauthenticated must be a valid default policy")
```
hawkw, line 544: not a big deal, but it seems a bit silly to try to parse this from a string? I'm guessing that the intent was to show _what_ policy was the default here, instead of just calling `Default::default()`...but maybe just having something like
```rust
.unwrap_or_else(|| DefaultPolicy::unauthenticated(detect_protocol_timeout))
```
would feel a little less circuitous?

not a blocker

linkerd/server-policy/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+authors = ["Linkerd Developers <cncf-linkerd-dev@lists.cncf.io>"]
+license = "Apache-2.0"
+edition = "2018"
+publish = false
+
+[dependencies]
+async-stream = "0.3"
+futures = { version = "0.3", default-features = false, features = ["alloc"]  }
+ipnet = "2"
+#linkerd2-proxy-api = { version = "0.2", features = ["inbound", "client"] }
```
hawkw, line 13: should the commented out line be removed?

linkerd/server-policy/src/network.rs
```diff
@@ -0,0 +1,69 @@
...
+        Self::from(ipnet::Ipv6Net::from(addr))
+    }
+}
+
+impl Network {
+    #[inline]
+    pub fn contains(&self, addr: &std::net::IpAddr) -> bool {
+        let contains_ip = self.net.contains(addr);
+        let exception = self.except.iter().any(|net| net.contains(addr));
+        contains_ip && !exception
```
hawkw, line 55: nit/tioli: if this was 
```suggestion
        contains_ip && self.except.iter().any(|net| net.contains(addr))
```
or something, instead of using a variable, we could avoid doing the iteration over `self.except` if the address is not contained in `self.net`. the way this is written now, we will *always* check if there's an exception, even if we don't need to.

not sure if this is a big enough overhead to actually matter, but felt like it was worth pointing out...

linkerd/app/src/env.rs
```diff
@@ -499,7 +499,7 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
             return Err(EnvError::InvalidEnvVar);
         }
 
-        // Ensure that connections thaat directly target the inbound port are
+        // Ensure that connections t directly target the inbound port are
```
hawkw, line 502: typo fix has a typo of its own:
```suggestion
        // Ensure that connections that directly target the inbound port are
```

hawkw (approved):
looks great! picked a few unimportant nits

Cargo.lock
```diff
@@ -198,6 +198,16 @@ dependencies = [
...
 
+[[package]]
+name = "env_logger"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a19187fea3ac7e84da7dacf48de0c45d63c6a76f9490dae389aead16c243fce3"
+dependencies = [
+ "log",
+ "regex",
+]
```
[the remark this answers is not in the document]
olix0r: https://github.com/linkerd/linkerd2-proxy/pull/1192/commits/18cf6a688325f105035971477884bbed18cdaf34

PR #1179, opened 2021-08-02
Split the inbound server into multiple modules

olix0r:
The `Inbound::push_server` stack builder is responsible for building all of the inbound stacks: the HTTP stack, the TCP forwarding stack, and the direct/gateway stack. Furthermore, most of these stacks rely on fixed target types (which makes them inflexible).

This change refactors the inbound stack to setup for upcoming policy changes:

1. A 'detect' stack is split out into a `push_detect` helper that only deals with protocol detection and dispatching connections to either the TCP forwarding stack or the HTTP handling stack.
2. An 'accept' stack is split out into a `push_accept` helper that is responsible for extracting relevant connection metadata (client and orig dst addrs) and determining the policy to use for the connection.
3. Each stack/module defines its own target types and other modules rely on param constraints instead of fixed target types.
4. A `port_policies` module is introduced to model the existing per-port policies: whether identity is required and whether protocol detection should be skipped. Various auxiliary modules have been eliminated in favor of inlined target filters. This helps to make these stacks easier to reason about, as relevant logic isn't spread over multiple files.

---

Depends on #1174

PR #1193, opened 2021-08-10
fuzz: fix inbound fuzz logic expecting wrong IP address

hawkw:
PR #1179 changed the inbound HTTP fuzz logic, and added a mocked
`Target` type similar to other stack tests. However, this introduced a
bug, because the `Target::addr()` function, which returns the socket
address that the mock TCP connector should expect the proxy to try to
connect to, returns the (mocked) original destination address. Since the
stack we build for the fuzz test maps the target IP address to localhost
before calling into the connect stack (as the real proxy currently
does), this means that we attempt to connect to an address the mock
connector doesn't expect. Then, it panics and the fuzz test fails.

This branch fixes this issue by changing `Target::addr` to return
127.0.0.1:80 instead of 192.0.2.2:80. This should match what the proxy
will actually attempt to connect to.

I also updated the fuzz logic to use exactly the same connect stack as
the HTTP stack tests (including a timeout). Although this shouldn't be
exercised, it makes the fuzz test more closely match the stack tests and
thus the real life proxy. I wonder if we can eventually share more code
between the tests and the fuzzers, but that seems like a follow-up.

### DIFF
--- a/linkerd/app/inbound/src/http/mod.rs
+++ b/linkerd/app/inbound/src/http/mod.rs
@@ -124,11 +124,7 @@ pub mod fuzz {
         I: io::AsyncRead + io::AsyncWrite + io::PeerAddr + Send + Unpin + 'static,
     {
         let connect = svc::stack(connect)
-            .map_stack(|cfg, _, s| {
-                s.push_map_target(|p| Remote(ServerAddr(([127, 0, 0, 1], p).into())))
-                    .push_map_target(|t| Param::<u16>::param(&t))
-                    .push_connect_timeout(cfg.proxy.connect.timeout)
-            })
+            .push_map_target(|t: Http| Remote(ServerAddr(([127, 0, 0, 1], t.param()).into())))
             .into_inner();
         Inbound::new(cfg, rt)
             .with_stack(connect)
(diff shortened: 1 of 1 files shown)


hawkw:
Also, another thought: Currently, we only try to _build_ the fuzz tests on CI, and never actually run them. This means that the CI build ensures they compile, but doesn't do anything to catch even simple logic errors in the fuzz logic itself. I wonder if we want to check in known-good minimized reproducers for each fuzz test, and run _those_ on CI. That way, the CI build doesn't spend a bunch of time actually running the fuzzer, but we would actually run the fuzz logic once as a sort of smoke test that we didn't just completely break it.

Just a thought.

linkerd/app/inbound/src/http/mod.rs
```diff
@@ -124,7 +124,11 @@ pub mod fuzz {
         I: io::AsyncRead + io::AsyncWrite + io::PeerAddr + Send + Unpin + 'static,
     {
         let connect = svc::stack(connect)
-            .push_map_target(|t: Http| Remote(ServerAddr(([127, 0, 0, 1], t.param()).into())))
+            .map_stack(|cfg, _, s| {
+                s.push_map_target(|p| Remote(ServerAddr(([127, 0, 0, 1], p).into())))
+                    .push_map_target(|t| Param::<u16>::param(&t))
+                    .push_connect_timeout(cfg.proxy.connect.timeout)
+            })
```
olix0r, line 131: is this change necessary? it seems to do the same thing afaict

olix0r (approved):

linkerd/app/inbound/src/http/mod.rs
hawkw: it's not necessary, the intent was just to ensure this code was in sync with the stack tests. i can back out this change.

PR #1199, opened 2021-08-11
Use FromIterator for NameAddr construction

olix0r:
If we provide a `FromIterator` implementation for `NameAddr`, we need
not have an extra import to construct `NameAddr`s when the type is
known.

No functional changes.


hawkw (approved):
awesome!

PR #1200, opened 2021-08-11
app: Defer stack construction until the proxy starts

olix0r:
We currently eagerly build the inbound and outbound stacks. In an
upcoming change, however, we'll want to ensure that we've obtained
inbound port policies before initializing the inbound stack.

This change modifies app initialization to move stack initialization
into the proxy's start task. In order to do this, we eagerly bind the
inbound/outbound listeners and provide the stream of incoming sockets to
the stacks at construction-time.

This doesn't actually change any behavior since stack construction was
moved into a task that was run lazily. The change is purely internal,
restructuring how/when this task is built.

### DIFF
--- a/linkerd/app/inbound/src/server.rs
+++ b/linkerd/app/inbound/src/server.rs
@@ -16,14 +16,13 @@ struct TcpEndpoint {
 // === impl Inbound ===
 
 impl Inbound<()> {
-    pub async fn serve<L, A, I, G, GSvc, P>(
+    pub async fn serve<A, I, G, GSvc, P>(
         self,
         addr: Local<ServerAddr>,
-        listen: L,
+        listen: impl Stream<Item = io::Result<(A, I)>> + Send + Sync + 'static,
         profiles: P,
         gateway: G,
     ) where
-        L: Stream<Item = io::Result<(A, I)>> + Send + Sync + 'static,
         A: svc::Param<Remote<ClientAddr>> + svc::Param<OrigDstAddr> + Clone + Send + Sync + 'static,
         I: io::AsyncRead + io::AsyncWrite + io::Peek + io::PeerAddr,
         I: Debug + Unpin + Send + Sync + 'static,
--- a/linkerd/app/inbound/src/server.rs
+++ b/linkerd/app/inbound/src/server.rs
@@ -1,11 +1,11 @@
 use crate::{direct, Inbound};
+use futures::Stream;
 use linkerd_app_core::{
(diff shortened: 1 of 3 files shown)


hawkw:
Release binary build seems to have timed out; I restarted it.

linkerd/app/inbound/src/server.rs
```diff
@@ -16,72 +16,64 @@ struct TcpEndpoint {
...
         profiles: P,
         gateway: G,
-    ) -> (Local<ServerAddr>, impl Future<Output = ()> + Send)
-    where
-        B: Bind<ServerConfig>,
-        B::Addrs: svc::Param<Remote<ClientAddr>>
-            + svc::Param<Local<ServerAddr>>
-            + svc::Param<OrigDstAddr>,
+    ) where
+        L: Stream<Item = io::Result<(A, I)>> + Send + Sync + 'static,
```
hawkw, line 26: nit/TIOLI: seems like `L` could just be an `impl Trait`, like
```rust
    pub async fn serve<A, I, G, GSvc, P>(
        self,
        addr: Local<ServerAddr>,
        listen: impl Stream<Item = io::Result<(A, I)>> + Send + Sync + 'static,
        profiles: P,
        gateway: G,
    ) where
```
to avoid adding another type param?

not a big deal either way

linkerd/app/outbound/src/lib.rs
```diff
@@ -119,15 +120,12 @@ impl<S> Outbound<S> {
...
-        self,
-        bind: B,
-        profiles: P,
-        resolve: R,
-    ) -> (Local<ServerAddr>, impl Future<Output = ()>)
+    pub async fn serve<L, A, I, P, R>(self, listen: L, profiles: P, resolve: R)
     where
-        B: Bind<ServerConfig>,
-        B::Addrs: Param<Remote<ClientAddr>> + Param<OrigDstAddr>,
+        L: Stream<Item = io::Result<(A, I)>> + Send + Sync + 'static,
```
hawkw, line 125: nit/tioli: similarly, since the `L` type param doesn't appear in any other `where` bounds, seems like it could be `impl Stream`:
```rust
    pub async fn serve<A, I, P, R>(
        self,
        listen: impl Stream<Item = io::Result<(A, I)>> + Send + Sync + 'static,
        profiles: P,
        resolve: R,
    ) where
```

hawkw (approved):
looks much nicer! i suggested one minor potential style improvement we could make, but it's not a blocker at all.

PR #1202, opened 2021-08-11
app: Block proxy initialization on identity readiness

olix0r:
Currently, the proxy may start serving connections even before identity
is provisioned. This won't be feasible once we introduce policy
discovery--the proxy will have to establish its inbound policy before
serving inbound connections.

This change modifies proxy initialization to block proxy initialization
on identity. When identity isn't established, warnings are logged every
15s.

While this is a behavior change from the proxy's point of view, this
shouldn't be a user-facing change, as container initialization is
blocked on the proxy's readiness, which is dependent on identity
initialization.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1168,6 +1168,7 @@ dependencies = [
  "linkerd-tls",
  "linkerd2-proxy-api",
  "pin-project",
+ "thiserror",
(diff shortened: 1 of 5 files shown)


hawkw (approved):
:thumbsup:

PR #1203, opened 2021-08-12
inbound: Use policy protocol configurations

olix0r:
Per-port server policies include protocol hinting that is not currently
honored.

This change modifies the inbound detect stacks to only perform HTTP
detection when the server policy indicates that detection should be
pefromed. It also honors the per-port detect timeout configuration. When
a port is documented as HTTP/1, HTTP/2, or gRPC, protocol detection is
skipped and the configuration is used instead.

When the port's protocol is documented to be application TLS, we check
that authorization policies apply but we do not explicitly enforce that
application TLS was present, since we may _additionally_ apply mesh TLS
in these situations.

### DIFF
--- a/linkerd/app/inbound/src/detect.rs
+++ b/linkerd/app/inbound/src/detect.rs
@@ -11,9 +11,10 @@ use linkerd_app_core::{
         addrs::{ClientAddr, OrigDstAddr, Remote},
         ServerAddr,
     },
-    Error,
+    Error, Infallible,
 };
-use std::fmt::Debug;
+use linkerd_server_policy::Protocol;
+use std::{fmt::Debug, time};
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Tls {
@@ -21,6 +22,15 @@ pub struct Tls {
     orig_dst_addr: OrigDstAddr,
     permit: Permitted,
 }
+
+#[derive(Clone, Debug)]
+struct Detect {
+    timeout: time::Duration,
+    tls: Tls,
+}
+
+#[derive(Copy, Clone, Debug)]
+struct ConfigureHttpDetect;
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Http {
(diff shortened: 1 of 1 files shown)


mateiidavid (commented):
Looks good to me, the `push_detect_http()` changes were actually pretty easy to reason with/about. I've left some questions that kind of gave me a hard time, they're mostly to do with `MapTargetLayer`, I'm not sure I understood fully what we are doing there.

linkerd/app/inbound/src/detect.rs
```diff
@@ -63,22 +73,34 @@ impl<N> Inbound<N> {
...
+                        // whether app TLS was employed, but we use this as a signal that we should
+                        // not perform additional protocol detection.
+                        if let Protocol::Tls = permit.protocol {
+                            return Ok(svc::Either::B(Tls::from_params(&t, permit)));
+                        }
+
+                        Ok(svc::Either::A(Tls::from_params(&t, permit)))
                     },
+                    svc::stack(forward.clone())
+                        .push_on_response(svc::MapTargetLayer::new(io::BoxedIo::new))
```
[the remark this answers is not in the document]
olix0r: we need to box the client socket because the `F`-typed forward-stack is constrained as:
```
        F: svc::NewService<Tls, Service = FSvc> + Clone + Send + Sync + Unpin + 'static,
        FSvc: svc::Service<io::BoxedIo, Response = ()> + Send + 'static,
```

that is, it expects a boxed I/o type. The service's request type has to be known, so we can either have some big Either-type that covers all of the ways the service can be called, or we can just put in a box so these various types look the same.

linkerd/app/inbound/src/detect.rs
```diff
@@ -63,22 +73,34 @@ impl<N> Inbound<N> {
...
                         let policy: AllowPolicy = t.param();
                         let permit = policy.check_authorized(tls)?;
-                        Ok(Tls::from_params(&t, permit))
+
+                        // If the port is configured to support application TLS, it may have also
+                        // been wrapped in mesh identity. In any case, we don't actually validate
+                        // whether app TLS was employed, but we use this as a signal that we should
+                        // not perform additional protocol detection.
+                        if let Protocol::Tls = permit.protocol {
+                            return Ok(svc::Either::B(Tls::from_params(&t, permit)));
```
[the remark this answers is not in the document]
mateiidavid: Thanks, that makes sense. Also, it just clicked that `Tls::from_params` produces a target type and whenever we use switches we actually use different target types, for some reason I was oblivious to that until now. Definitely helps a lot.

linkerd/app/inbound/src/detect.rs
```diff
@@ -63,22 +73,34 @@ impl<N> Inbound<N> {
...
+                    // detection.
+                    |(tls, t): (tls::ConditionalServerTls, T)| -> Result<_, Error> {
                         let policy: AllowPolicy = t.param();
                         let permit = policy.check_authorized(tls)?;
-                        Ok(Tls::from_params(&t, permit))
+
+                        // If the port is configured to support application TLS, it may have also
+                        // been wrapped in mesh identity. In any case, we don't actually validate
+                        // whether app TLS was employed, but we use this as a signal that we should
+                        // not perform additional protocol detection.
```
hawkw, line 89: do we want to eventually have this configuration validate that the app is using TLS?

linkerd/app/inbound/src/detect.rs
```diff
@@ -22,6 +23,15 @@ pub struct Tls {
...
 }
 
+#[derive(Clone, Debug)]
+struct Detect {
+    timeout: time::Duration,
+    tls: Tls,
+}
+
+#[derive(Copy, Clone, Debug)]
+struct ConfigureHttpDetect;
```
hawkw, line 33: guessing that subsequent changes will make this actually do something?

hawkw (approved):
looks good to me!

linkerd/app/inbound/src/detect.rs
olix0r: It's used now -- this is the ExtractParam that configures the http detect stack

PR #1205, opened 2021-08-18
inbound: Discover policies from the control plane

olix0r:
We've recently introduced support for authorization policies. These
policies are configured statically from the proxy's environment. This
change adds an optional new mode to support dynamic configuration from
the control plane.

The proxy now supports new environment variables,
`LINKERD2_PROXY_POLICY_SVC_{ADDR,NAME}` that may be set with the address
and identity of the policy controller. When this is set, other static
inbound port configurations (requiring identity, marking ports as
opaque, etc) are ignored in favor of the results returned by the API.
Instead, the proxy uses the `LINKER2_PROXY_INBOUND_PORTS` environment
variable to discover the list of ports configured in the pod's spec and
discovers the policies for each of these ports before admitting inbound
connections. When a connection is received for a port not in this list,
the `LINKERD2_PROXY_INBOUND_DEFAULT_POLICY` configuration is used to
determine whether the connection may be permitted or whether it should
be refused.

This change modifies the set of default policies to support only `deny`,
`all-authenticated` and `all-unauthenticated` (removing
`all-mtls-unauthenticated`). Instead a
`LINKERD2_PROXY_INBOUND_PORTS_REQUIRE_TLS` configuration is introduced
to enable tls requirements for individual pods. This static
configuration should only be necessary for the identity controller, in
which case we need to use an unauthenticated default policy to permit
kubernetes healtchecking probes; and we can use this new configuration
to enforce a default policy on the gRPC API port.


linkerd/app/src/env.rs
```diff
@@ -488,85 +486,164 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
-                    ),
-            )
+                            .filter_map(|p| match default {
+                                policy::DefaultPolicy::Allow(ref a) => {
+                                    let mut sp = (**a).clone();
+                                    sp.protocol = inbound::policy::Protocol::Opaque;
+                                    Some((p, sp))
+                                }
+                                policy::DefaultPolicy::Deny => {
+                                    tracing::warn!("inbound opaque ports configuration is ignored when the default policy is 'deny'");
```
hawkw, line 619: nit: i think `warn!` is imported in this module?
```suggestion
                                    warn!("inbound opaque ports configuration is ignored when the default policy is 'deny'");
```

linkerd/app/inbound/src/detect.rs
```diff
@@ -116,8 +153,6 @@ impl<N> Inbound<N> {
                         .push_on_response(svc::MapTargetLayer::new(io::BoxedIo::new))
                         .into_inner(),
                 )
-                .push_on_response(svc::BoxService::layer())
-                .push(svc::BoxNewService::layer())
```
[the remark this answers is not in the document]
olix0r: not that i could observe (it all still gets boxed up in `push_detect`) -- and it looks like the issues we hit are fixed on nightly/in the next stable release

olix0r:
I've tested this end-to-end by manually configuring the proxy for discovery. There's more we need to do testing-wise but for now I think we should try to get this merged and then focus on integration testing in the linkerd2 repo.

linkerd/app/core/src/serve.rs
```diff
@@ -75,7 +75,8 @@ pub async fn serve<M, S, I, A>(
                 }
             }
         }
-    };
+    }
+    .in_current_span();
```
olix0r, line 79: I'm surprised this is necessary but without it we don't see the inbound/outbound span on log messages from this future

linkerd/app/inbound/src/detect.rs
```diff
@@ -143,18 +183,26 @@ impl<N> Inbound<N> {
...
                             }
-                            Protocol::Http1 => http::Version::Http1,
+                            Protocol::Http1 => {
+                                // HTTP/1 services may actually be transported over HTTP/2
+                                // connections between proxies, so we have to do detection.
+                                return Ok(svc::Either::B(Detect {
+                                    timeout: detect_timeout,
+                                    tls,
+                                }));
+                            }
```
olix0r, line 205: This was an oversight before: we need to do detection in this case to support protocol upgrade (found during testing)

olix0r:
The release build is right at the 14-15 minute mark (even with boxing). Not much we can do, I think, but wait for the fixes for https://github.com/rust-lang/rust/issues/84873 to hit stable. Going to bump the timeout in the meantime

kleimkuhler (approved):
Code-wise looks good. Going to walk through the logic a little more with the changes actually in use in an installation, but nothing major standing out to me right now.

linkerd/app/inbound/src/direct.rs
```diff
@@ -98,31 +102,58 @@ impl<N> Inbound<N> {
...
+                                        allow.check_authorized(client.client_addr, tls)?;
+                                    // TODO(ver) Use the permit's labels in metrics...
+                                    Ok(svc::Either::A(port))
+                                }
+                                TransportHeader {
+                                    port,
+                                    name: Some(name),
+                                    protocol,
+                                } => {
+                                    let allow = policies.check_policy(client.local_addr)?;
```
hawkw, line 134: why does the case where the transport header includes a name and a port check the client's local addr, while the case where there's no name check the client's local addr IP and the port from the transport header?

is this an oversight, or is it on purpose? and if it's on purpose, can we have a comment here explaining why it's different?

hawkw (commented):
noticed one thing that seemed like a potential concern, plus a couple small nits (not blockers). once that question's addressed, lgtm!

linkerd/app/inbound/src/direct.rs
olix0r: when it includes a name/port, it's a gateway configuration targeting some other service; so the policy check determines whether the client can contact the _gateway_ itself. will add a comment

hawkw (approved):
thanks for clarifying the gateway behavior, lgtm!

PR #1210, opened 2021-08-20
policy: Add support for cluster-scoped default policies

olix0r:
The policy controller supports the default policies
`cluster-authenticated` and `cluster-unauthenticated` but the proxy
would error if configured with these policies. In order to use a single
policy annotation honored by both the policy controller and proxy, this
change adds support for a new configuration,
`LINKERD2_POLICY_CLUSTER_NETWORKS`, that allows the proxy to enforce
cluster-scoped defaults. When this configuration is not specified, the
cluster-scoped defaults are not supported.

This default will apply to all ports that are not documented in the
pod's spec (i.e. ports not in the `LINKERD2_PROXY_INBOUND_PORTS`
configuration).

This change also updates policy labeling to track server and authz labels
independently so that keys may overlap, as is required by https://github.com/linkerd/linkerd2/pull/6722

### DIFF
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -508,11 +508,7 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
                     "{} was not set; using `all-unauthenticated`",
                     ENV_INBOUND_DEFAULT_POLICY
                 );
-                policy::defaults::all_unauthenticated(
-                    "default:all-unauthenticated",
-                    detect_protocol_timeout,
-                )
-                .into()
+                policy::defaults::all_unauthenticated(detect_protocol_timeout).into()
             });
 
             match parse_control_addr(strings, ENV_POLICY_SVC_BASE, id_disabled)? {
@@ -586,7 +582,6 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
                                 .into_iter()
                                 .map(|p| {
                                     let allow = policy::defaults::all_authenticated(
-                                        format!("fixed:{}", p),
                                         detect_protocol_timeout,
                                     );
                                     (p, allow)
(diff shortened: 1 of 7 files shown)


linkerd/app/src/env.rs
```diff
@@ -1007,12 +1018,23 @@ fn parse_networks(list: &str) -> Result<HashSet<IpNet>, ParseError> {
...
+    cluster_nets: HashSet<IpNet>,
     detect_timeout: Duration,
 ) -> Result<policy::DefaultPolicy, ParseError> {
     match s {
         "deny" => Ok(policy::DefaultPolicy::Deny),
         "all-authenticated" => Ok(policy::defaults::all_authenticated(detect_timeout).into()),
         "all-unauthenticated" => Ok(policy::defaults::all_unauthenticated(detect_timeout).into()),
+
+        // If cluster networks are configured, support cluster-scoped default policies.
+        name if cluster_nets.is_empty() => Err(ParseError::InvalidPortPolicy(name.to_string())),
```
hawkw, line 1030: hmm, should there be a separate error case for this? this way, we will log an error that kind of implies "cluster-authenticated" isn't a valid default policy name, but it _is_ a valid policy, it's just not valid in the case that there's no cluster networks.

possibly improving the error message here is not that important, though, since users aren't setting these env vars manually?

linkerd/server-policy/src/lib.rs
```diff
@@ -2,17 +2,25 @@ mod network;
...
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct ServerPolicy {
     pub protocol: Protocol,
     pub authorizations: Vec<Authorization>,
-    pub labels: HashMap<String, String>,
+    pub labels: Labels,
 }
 
+/// Stores an ordered, cloneeable set of labels.
```
hawkw, line 20: nit: i don't _think_ there should be two 'e's here :)
```suggestion
/// Stores an ordered, cloneable set of labels.
```

hawkw (approved):
lgtm --- this change is pretty straightforward!

kleimkuhler (approved):
Aside from comment on some better error handling here looks good!

linkerd/app/src/env.rs
olix0r: > possibly improving the error message here is not that important, though, since users aren't setting these env vars manually?

Correct. Seeing this is basically a programing error: the injector isn't doing the right thing. There's nothing a user could change in their config other than to change/unset the default policy on the pod.

PR #1211, opened 2021-08-23
Use `ExtractParam` in transport metrics

olix0r:
Our transport metrics code is fairly old and doesn't take advantage of
newer primitives, namely `ExtractParam`.

This change:
* Moves transport metrics code into a dedicated `transport-metrics crate` (it's got nothing in common with the rest of the proxy-transport code). Each logical chunk of functionality is moved into its own module.
* Renames the `Connect` type to `Client; and `Accept` to `Server` -- this is where we've gone with TLS code, etc.
* Updates both the Connect and Server layers to be parameterized on an `ExtractParams` that is responsible for producing a metrics instance for each target. This eliminates the awkward mapping from target type to label type and in general decouples these layers from the details of how metrics are acquired.
* The `app_core::transport::metrics` module provides an implementation of `ExtractParams` that can be used to build transport labels.

This sets up upcoming changes to metrics infrastructure with the eventual goal being to decouple the metric names definition from the metrics tracking code.

No functional changes.

### DIFF
--- a/linkerd/app/core/src/transport/mod.rs
+++ b/linkerd/app/core/src/transport/mod.rs
@@ -1,6 +1,29 @@
 pub use linkerd_proxy_transport::*;
+use linkerd_stack::{ExtractParam, Param};
 pub use linkerd_transport_metrics as metrics;
+use std::sync::Arc;
 
 pub mod labels;
 
-pub type Metrics = metrics::Registry<labels::Key>;
+#[derive(Clone, Debug)]
+pub struct Metrics(metrics::Registry<labels::Key>);
+
+impl From<metrics::Registry<labels::Key>> for Metrics {
+    fn from(reg: metrics::Registry<labels::Key>) -> Self {
+        Self(reg)
+    }
+}
+
+impl<T: Param<labels::Key>> ExtractParam<Arc<metrics::Metrics>, T> for Metrics {
+    fn extract_param(&self, t: &T) -> Arc<metrics::Metrics> {
+        self.0.metrics(t.param())
+    }
+}
+
+impl std::ops::Deref for Metrics {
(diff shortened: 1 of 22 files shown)


linkerd/transport-metrics/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+    by_eos: Arc<Mutex<ByEos>>,
+}
+
+#[derive(Debug)]
+struct ByEos {
+    last_update: Instant,
+    metrics: HashMap<Eos, EosMetrics>,
+}
+
+/// Describes a classtransport end.
```
hawkw, line 63: typo (looks like it was copied and pasted from the old code though):
```suggestion
/// Describes a class of transport end.
```

linkerd/app/core/src/transport/mod.rs
```diff
@@ -1,5 +1,21 @@
 pub use linkerd_proxy_transport::*;
+use linkerd_stack::{ExtractParam, Param};
+pub use linkerd_transport_metrics as metrics;
+use std::sync::Arc;
 
 pub mod labels;
 
-pub type Metrics = metrics::Registry<labels::Key>;
+#[derive(Clone, Debug)]
+pub struct Metrics(metrics::Registry<labels::Key>);
```
hawkw, line 9: might be worth a comment explaining the relationship between this type and the `linkerd-transport-metrics` crate? it makes sense in the context of this PR and with the PR description, but i feel like it might be less obvious for someone looking at this code for the first time after this branch merges...

not a blocker though.

hawkw (approved):
looks good to me! it took me a minute to notice this, but this change is actually fairly simple; most of the diff is just moving files around with no code changes.

:+1:

PR #1215, opened 2021-08-23
Expose policy labels on inbound transport metrics

olix0r:
The inbound server only permits connections that are allowed by a
policy, but this decision is not observable at runtime.

This change modifies the server transport labels to include server and
authorization labels--prefixed by `srv_` and `saz_`, respectively
(matching the shortnames of the `Server` and `ServerAuthorization`
resources).

Furthermore, this change updates the `inbound_tcp_acept_errors_total`
metric to include an `unauthorized` error variant to expose a counter of
connections that were not permitted.

### DIFF
--- a/linkerd/app/core/src/transport/labels.rs
+++ b/linkerd/app/core/src/transport/labels.rs
@@ -1,6 +1,7 @@
 pub use crate::metrics::{Direction, OutboundEndpointLabels};
 use linkerd_conditional::Conditional;
 use linkerd_metrics::FmtLabels;
+use linkerd_server_policy as policy;
 use linkerd_tls as tls;
 use std::{fmt, net::SocketAddr};
 
@@ -11,13 +12,17 @@ use std::{fmt, net::SocketAddr};
 /// Implements `FmtLabels`.
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
 pub enum Key {
-    Accept {
-        direction: Direction,
-        tls: tls::ConditionalServerTls,
-        target_addr: SocketAddr,
-    },
-    OutboundConnect(OutboundEndpointLabels),
-    InboundConnect,
+    Server(ServerLabels),
+    OutboundClient(OutboundEndpointLabels),
+    InboundClient,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct ServerLabels {
+    direction: Direction,
+    tls: tls::ConditionalServerTls,
+    target_addr: SocketAddr,
+    policy: Option<PolicyLabels>,
 }
 
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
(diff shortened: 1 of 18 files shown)


linkerd/app/inbound/src/policy/mod.rs
```diff
@@ -86,20 +72,20 @@ impl AllowPolicy {
         &self,
         client_addr: Remote<ClientAddr>,
         tls: tls::ConditionalServerTls,
-    ) -> Result<Permitted, DeniedUnauthorized> {
+    ) -> Result<Permit, DeniedUnauthorized> {
```
hawkw, line 75: any particular rationale behind the name change for `Permit`? w/ the previous naming scheme, the two sides of the `Result` had consistent tense... not a big deal, just wondering.

linkerd/app/core/src/transport/labels.rs
```diff
@@ -74,6 +84,49 @@ impl FmtLabels for Key {
...
+        (TargetAddr(self.target_addr), TlsAccept(&self.tls)).fmt_labels(f)?;
+
+        if let Some(policy) = self.policy.as_ref() {
+            for (k, v) in policy.server.iter() {
+                write!(f, ",srv_{}=\"{}\"", k, v)?;
+            }
+            for (k, v) in policy.authz.iter() {
+                write!(f, ",saz_{}=\"{}\"", k, v)?;
+            }
+        }
```
hawkw, line 124: nit/TIOLI: should this just be a `FmtLabels` impl for `PolicyLabels`, or do we expect them to show up elsewhere with different prefixes??

linkerd/app/core/src/transport/labels.rs
```diff
@@ -11,13 +12,17 @@ use std::{fmt, net::SocketAddr};
...
-    Accept {
-        direction: Direction,
-        tls: tls::ConditionalServerTls,
-        target_addr: SocketAddr,
-    },
-    OutboundConnect(OutboundEndpointLabels),
-    InboundConnect,
+    Server(ServerLabels),
+    OutboundClient(OutboundEndpointLabels),
+    InboundClient,
```
hawkw, line 17: alternative naming suggestion, take it or leave it: maybe instead of `Server` and `Client` these should be `Src` and `Dest`, since that's what we'll actually emit for the `peer=` label? but, maybe consistent naming with the service that generates the labels is better... :woman_shrugging:

linkerd/app/inbound/src/direct.rs
```diff
@@ -28,31 +28,44 @@ pub struct RefusedNoIdentity(());
 #[error("a named target must be provided on gateway connections")]
 struct RefusedNoTarget;
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct Local {
+    pub port: u16,
+    pub permit: policy::Permit,
```
hawkw, line 34: nit: AFAICT these fields aren't actually accessed directly anywhere outside of this file (since they're exposed via `Param` impls), so can't this just be
```suggestion
    port: u16,
    permit: policy::Permit,
```

hawkw (approved):
a few minor bikeshedding comments, but this lgtm overall!

linkerd/app/core/src/transport/labels.rs
olix0r: the code for comma handling gets a bit more complicated than it's worth, imo

linkerd/app/core/src/transport/labels.rs
olix0r: yeah, we're going to move away from the peer labels sooner than later

linkerd/app/inbound/src/policy/mod.rs
olix0r: the target types hold "a permit"

PR #1216, opened 2021-08-23
inbound: Expose permit labels on HTTP metrics

olix0r:
This change follows linkerd/linkerd2-proxy#1215, which exposes policy
labels on inbound TCP server metrics, by adding these same labels to
HTTP metrics.

The HTTP router target types are updated to take a `Permit` which
includes server and authz labels so that the inbound HTTP metrics can
include these labels. This will enable `linkerd viz stat` to work on
these resources. This change shouldn't impact metrics cardinality, since
client IDs are already exposed in metrics labeling.

### DIFF
--- a/linkerd/app/core/src/metrics/mod.rs
+++ b/linkerd/app/core/src/metrics/mod.rs
@@ -14,6 +14,7 @@ use crate::{
 use linkerd_addr::Addr;
 use linkerd_metrics::FmtLabels;
 pub use linkerd_metrics::*;
+use linkerd_server_policy as policy;
 use std::{
     fmt::{self, Write},
     net::SocketAddr,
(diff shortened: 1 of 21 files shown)


linkerd/app/core/src/metrics/mod.rs
```diff
@@ -67,6 +68,13 @@ pub struct InboundEndpointLabels {
     pub tls: tls::ConditionalServerTls,
     pub authority: Option<http::uri::Authority>,
     pub target_addr: SocketAddr,
+    pub policy: PolicyLabels,
+}
+
+#[derive(Clone, Debug, Default, Eq, PartialEq, Hash)]
+pub struct PolicyLabels {
+    pub server: policy::Labels,
+    pub authz: policy::Labels,
```
hawkw, line 77: nit/tioli: kinda seems like this type and it's `FmtLabels` impl should just be provided by the `policy` crate, since there's an identical version of this in the `transport-metrics` crate? not a big deal though.

olix0r: There's just a single instance of this type -- transport-metrics doesn't define any labels, this is all in `app::core` (and it was moved from `core::transport::labels` to here). We could try to deduce in the for loops in a followup but my hunch is that it will be a net gain in LOC.

kleimkuhler (approved):

PR #1235, opened 2021-08-31
Rename push_on_response to push_on_service

olix0r:
The "on response" terminology is confusing: in all cases where we use
push_on_response, we're actually pushing a layer that operates on a
`NewService`'s returned `Service`. This rename helps to make this usage
clearer.


hawkw (approved):
definitely clearer, good call!

kleimkuhler (approved):
Yep +1 on this!

PR #1237, opened 2021-09-02
inbound: Improve policy metrics

olix0r:
We recently introduced metrics to help surface inbound policy decisions,
but in practice these haven't been as useful as we might hope.
Specifically, error metrics do not include the `target_addr` label so
these metrics can't be correlated with servers, etc. This change
improves error metrics and also introduces new metrics to describe
authorization decisions: authorization denials shouldn't be classified
as errors, really, anyway.

This change also improves TCP forwarding authorization so that policy
changes can be honored at runtime: previously authorized connections may
dropped if the policy is updated so that the connection is no longer
authorized.

The gateway is also updated to enforce HTTP policies at runtime as well
so that policy changes can be honored after the connection has been
established.

This change introduces new metrics:

* `inbound_http_authz_allow_total`
* `inbound_http_authz_deny_total`
* `inbound_tcp_authz_allow_total`
* `inbound_tcp_authz_deny_total`
* `inbound_tcp_authz_terminate_total`

_allow_ metrics include `target_addr`, `srv_name`, and `saz_name`
labels. _deny_ and _terminate_ metics include only `target_addr` and
`srv_name` labels.

Authorization denials are no longer reflected in inbound_tcp_error or
inbound_http_error metrics.

A number of internal changes have been made to support this:

* The `inbound::policy::authorize` module includes middlewares for TCP
  and HTTP authorization, replacing the prior method of enforcing policy
  in the stack/router. This module ensures that metrics are recorded for
  policy decisions.
* The `error-metrics` crate has been removed. In its place a `monitor`
  type has been added to the `stack` crate, supporting a general way to
  observe errors, decoupled from the metrics registry.
* Inbound and outbound error metrics are now tracked in the inbound and
  outbound crates, respectively. Inbound- and outbound-specific error
  types are also moved into their rspective crates.
* The `app_core::errors` module has been updated to only define the
  types it needs to instrument the error response layer. Error responses
  are now primarily instrumented via the `HttpError` type so that errors
  that should be handled can be configured where the error is thrown.
  The error type now holds an underlying source error so that the error
  metrics layer can see through this wrapper type to track the
  underlying error cause.
* Server & Authorization labels are no longer handled as a free-form
  maps. We currently read only the `name` label from each; and this
  label is required.


linkerd/app/core/src/errors/respond.rs
```diff
@@ -251,13 +141,80 @@ impl<RspB: Default + hyper::body::HttpBody> respond::Respond<http::Response<RspB
...
+
+    fn poll_data(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Result<Self::Data, Self::Error>>> {
+        match self.project() {
+            ResponseBodyProj::NonGrpc(inner) => inner.poll_data(cx),
+            ResponseBodyProj::Grpc { inner, trailers } => {
+                // should not be calling poll_data if we have set trailers derived from an error
+                assert!(trailers.is_none());
```
hawkw, line 161: nit/TIOLI: should we maybe just move the comment into the panic message for the assert?

linkerd/app/inbound/src/detect.rs
```diff
@@ -98,61 +96,74 @@ impl<N> Inbound<N> {
...
+                            }));
                         }
                         Ok(svc::Either::A(t))
                     },
                     svc::stack(forward)
                         .push_on_service(svc::MapTargetLayer::new(io::BoxedIo::new))
+                        .push_map_target(Forward::from)
+                        .push(policy::NewAuthorizeTcp::layer(rt.metrics.tcp_authz.clone()))
+                        .check_new_service::<Tls, I>()
                         .into_inner(),
```
hawkw, line 164: very minor nit, not a blocker: seems like there could be a `let` binding for this and we could clone it into each `push_switch`...or is there some type difference i'm missing?

linkerd/app/inbound/src/http/router.rs
```diff
@@ -355,6 +345,7 @@ impl tap::Inspect for Logical {
     }
 
     fn dst_labels<B>(&self, _: &http::Request<B>) -> Option<&tap::Labels> {
+        // TODO include policy labels here.
```
hawkw, line 348: does this need to be addressed as part of this PR, or is it a follow-up change?

linkerd/app/inbound/src/lib.rs
```diff
@@ -41,19 +46,61 @@ pub struct Config {
...
-        &self.runtime
+    pub fn identity(&self) -> Option<&LocalCrtKey> {
+        self.runtime.identity.as_ref()
+    }
+
+    pub fn proxy_metrics(&self) -> &metrics::Proxy {
+        &self.runtime.metrics.proxy
+    }
+
+    /// A helper fro gateways to instrument policy checks.
```
hawkw, line 92: typo:
```suggestion
    /// A helper for gateways to instrument policy checks.
```

linkerd/app/inbound/src/metrics/authz.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    inbound_tcp_authz_allow_total: Counter {
+        "The total number of inbound HTTP requests that were authorized"
+    },
+    inbound_tcp_authz_deny_total: Counter {
+        "The total number of inbound HTTP requests that were denied"
+    },
+    inbound_tcp_authz_terminate_total: Counter {
+        "The total number of inbound HTTP requests that were terminated due to an authorization change"
+    }
```
hawkw, line 25: looks like the help messages were copied and pasted incorrectly:
```suggestion
    inbound_tcp_authz_allow_total: Counter {
        "The total number of inbound TCP connections that were authorized"
    },
    inbound_tcp_authz_deny_total: Counter {
        "The total number of inbound TCP connections that were denied"
    },
    inbound_tcp_authz_terminate_total: Counter {
        "The total number of inbound TCP connections that were terminated due to an authorization change"
    }
```

linkerd/app/inbound/src/metrics/authz.rs
```diff
@@ -0,0 +1,153 @@
...
+struct HttpInner {
+    allow: RwLock<HashMap<(TargetAddr, AuthzLabels), Counter>>,
+    deny: RwLock<HashMap<(TargetAddr, ServerLabel), Counter>>,
+}
+
+#[derive(Debug, Default)]
+struct TcpInner {
+    allow: RwLock<HashMap<(TargetAddr, AuthzLabels), Counter>>,
+    deny: RwLock<HashMap<(TargetAddr, ServerLabel), Counter>>,
+    terminate: RwLock<HashMap<(TargetAddr, ServerLabel), Counter>>,
```
hawkw, line 44: AFAICT, these are never actually read concurrently --- it looks like every metrics update requires a write lock. Therefore, we're not really gaining anything from using a `RwLock` rather than a `Mutex` here...in fact, it may be a bit less efficient.

Unless we're planning to change the code that updates the metrics to try to increment with a read lock, and only acquire the write lock when we need to insert a new set of labels, we should *probably* just replace these with `Mutex`es).

linkerd/app/inbound/src/metrics/error/http.rs
```diff
@@ -0,0 +1,84 @@
...
+use std::{collections::HashMap, sync::Arc};
+
+metrics! {
+    inbound_http_errors_total: Counter {
+        "The total number of inbound HTTP requests that could not be processed due to a proxy error."
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct HttpErrorMetrics(Arc<RwLock<HashMap<(ErrorKind, (TargetAddr, ServerLabel)), Counter>>>);
```
hawkw, line 18: similarly, it looks like there are never concurrent reads of the `RwLock` so it should probably just be a `Mutex`?

linkerd/app/inbound/src/metrics/error/tcp.rs
```diff
@@ -0,0 +1,83 @@
...
+use std::{collections::HashMap, sync::Arc};
+
+metrics! {
+    inbound_tcp_errors_total: Counter {
+        "The total number of inbound TCP connections that could not be processed due to a proxy error."
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct TcpErrorMetrics(Arc<RwLock<HashMap<(ErrorKind, TargetAddr), Counter>>>);
```
hawkw, line 18: and again, probably should be a mutex?

linkerd/app/inbound/src/policy/authorize/tcp.rs
```diff
@@ -0,0 +1,158 @@
...
+                tls,
+                policy,
+                metrics,
+            }) => {
+                let client = *client;
+                let tls = tls.clone();
+                let mut policy = policy.clone();
+                let metrics = metrics.clone();
+
+                // FIXME increment counter.
```
hawkw, line 135: is this supposed to happen before merging this?

linkerd/app/inbound/src/policy/mod.rs
```diff
@@ -69,8 +84,19 @@ impl AllowPolicy {
...
+
+    #[inline]
+    pub fn server_label(&self) -> ServerLabel {
+        ServerLabel(self.server.borrow().name.clone())
+    }
+
+    async fn changed(&mut self) {
+        if self.server.changed().await.is_err() {
+            futures::future::pending::<()>().await;
+        }
```
hawkw, line 99: maybe worth a comment explaining why this should never complete if the watch was dropped? at a glance it seemed slightly scary, but I think I understand why we would want that --- we don't ever want to tear down those connections if the policy went away?

linkerd/app/outbound/src/metrics/error/http.rs
```diff
@@ -0,0 +1,54 @@
...
+use std::{collections::HashMap, sync::Arc};
+
+metrics! {
+    outbound_http_errors_total: Counter {
+        "The total number of inbound HTTP requests that could not be processed due to a proxy error."
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct Http(Arc<RwLock<HashMap<ErrorKind, Counter>>>);
```
hawkw, line 16: ...same rwlock comment...

linkerd/stack/src/monitor.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[inline]
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        match futures::ready!(this.inner.try_poll(cx)) {
+            Ok(rsp) => Poll::Ready(Ok(rsp)),
+            Err(err) => {
+                this.monitor.monitor_error(&err);
+                Poll::Ready(Err(err))
+            }
+        }
```
hawkw, line 135: nit/tioli: it seems like this (and the `poll_ready`) could _maybe_ be made slightly more concise with `map_err`? but, the match is fine too.

hawkw (approved):
overall, this looks good to me, i had mostly minor comments. i do think we may want to just replace the `RwLock`s with `Mutex`es, because it seems like we're not really getting any benefit from them --- we never have concurrent reads, and it might just be adding overhead.

PR #1238, opened 2021-09-02
metrics: Add `target_ip` and `target_port` labels

olix0r:
The `target_addr` label encodes both the IP address and port in a single
string. This makes certain grouping operations cumbersome.

In order to support more flexible queries, this change adds `target_ip`
and `target_port` labels in addition to the existing `target_addr`
label. This should have a negligible impact on prometheus, since it
doesn't actually increase timeseries cardinality.


hawkw (approved):
looks good to me!

PR #1245, opened 2021-09-06
orig-proto: Avoid emiting HTTP/2 errors for upgraded requests

olix0r:
Our error handlers need to account for the fact that HTTP/1 requests may
fail with HTTP/2 errors due to orig-proto upgrading.

To simplify error handling (in an upcoming change), the
`orig_proto::Upgrade` service introduces a new error type,
`orig_proto::DowngradedH2Error` that hides the original error source so
that these errors are not detected as being an `h2::Error`.

### DIFF
--- a/linkerd/proxy/http/src/orig_proto.rs
+++ b/linkerd/proxy/http/src/orig_proto.rs
@@ -1,13 +1,16 @@
-use super::{glue::UpgradeBody, h1, h2, upgrade};
+use super::{h1, h2, upgrade};
 use futures::{future, prelude::*};
 use http::header::{HeaderValue, TRANSFER_ENCODING};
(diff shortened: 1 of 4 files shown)


mateiidavid (approved):
Looks good!

linkerd/proxy/http/src/orig_proto.rs
```diff
@@ -89,28 +99,90 @@ where
...
+                return DowngradedH2Error(reason).into();
+            }
+        }
+
+        cause = error.source();
+    }
+
+    error.into()
+}
+// === impl UpgradeResponseBody ===
```
olix0r, line 147:
```suggestion

// === impl UpgradeResponseBody ===
```

PR #1246, opened 2021-09-07
errors: Support contextual error handling strategies

olix0r:
Currently `app_core::errors` handles synthesis of all error responses.
This means that this module needs to know about all possible error
types. To work around this, we wrap some errors with a special
`HttpError` type so this module can create a synthetic response.
Furthermore, this means that we can't handle errors differently in
inbound/outbound contexts.

This change modies the `error::respond` module with a new strategy
trait--`HttpRescue`--that is responsible for producing a
`SyntheticHttpResponse` if the error can be handled and returning an
error if it should not be handled gracefully. A `DefaultHttpRescue` type
encodes default behavior and inbound- and outbound-specific types extend
this functionality with appropriate error handling.

This change enables us to perform narrower error handling in more places
in the stack--for example, so that some error responses are included in
response metrics.

This change modifies HTTP/1 error responses to include a
`connection: close` header when the server-side connection is being torn
down.

### DIFF
--- a/linkerd/app/core/src/errors/respond.rs
+++ b/linkerd/app/core/src/errors/respond.rs
@@ -15,15 +15,6 @@ pub const L5D_PROXY_ERROR: &str = "l5d-proxy-error";
 pub trait HttpRescue<E> {
     /// Attempts to synthesize a response from the given error.
     fn rescue(&self, error: E) -> Result<SyntheticHttpResponse, E>;
-
-    /// A helper for inspecting potentially nested errors.
-    fn has_cause<C: std::error::Error + 'static>(
-        error: &(dyn std::error::Error + 'static),
-    ) -> bool {
-        // It's impractical for an error type to be so deeply nested that this recursion could
-        // overflow.
-        error.is::<C>() || error.source().map(Self::has_cause::<C>).unwrap_or(false)
-    }
 }
 
 #[derive(Clone, Debug)]
@@ -62,15 +53,80 @@ const GRPC_CONTENT_TYPE: &str = "application/grpc";
 const GRPC_STATUS: &str = "grpc-status";
 const GRPC_MESSAGE: &str = "grpc-message";
 
+// === impl HttpRescue ===
+
+impl<E, F> HttpRescue<E> for F
+where
+    F: Fn(E) -> Result<SyntheticHttpResponse, E>,
+{
+    fn rescue(&self, error: E) -> Result<SyntheticHttpResponse, E> {
+        (self)(error)
+    }
+}
+
 // === impl SyntheticHttpResponse ===
 
-impl Default for SyntheticHttpResponse {
-    fn default() -> Self {
+impl SyntheticHttpResponse {
+    pub fn unexpected_error() -> Self {
         Self {
             http_status: http::StatusCode::INTERNAL_SERVER_ERROR,
             grpc_status: tonic::Code::Internal,
             message: "unexpected error".to_string(),
             close_connection: true,
+        }
+    }
+
+    pub fn bad_gateway(msg: impl ToString) -> Self {
+        Self {
+            http_status: http::StatusCode::BAD_GATEWAY,
+            grpc_status: tonic::Code::Unavailable,
+            close_connection: true,
+            message: msg.to_string(),
+        }
+    }
+
+    pub fn gateway_timeout(msg: impl ToString) -> Self {
+        Self {
+            http_status: http::StatusCode::GATEWAY_TIMEOUT,
+            grpc_status: tonic::Code::Unavailable,
+            close_connection: true,
+            message: msg.to_string(),
+        }
+    }
+
+    pub fn unauthenticated(msg: impl ToString) -> Self {
+        Self {
+            http_status: http::StatusCode::FORBIDDEN,
+            grpc_status: tonic::Code::Unauthenticated,
+            close_connection: false,
+            message: msg.to_string(),
+        }
+    }
+
+    pub fn permission_denied(msg: impl ToString) -> Self {
+        Self {
+            http_status: http::StatusCode::FORBIDDEN,
+            grpc_status: tonic::Code::PermissionDenied,
+            close_connection: false,
+            message: msg.to_string(),
+        }
+    }
+
+    pub fn loop_detected(msg: impl ToString) -> Self {
+        Self {
+            http_status: http::StatusCode::LOOP_DETECTED,
+            grpc_status: tonic::Code::Aborted,
+            close_connection: true,
+            message: msg.to_string(),
+        }
+    }
+
+    pub fn not_found(msg: impl ToString) -> Self {
+        Self {
+            http_status: http::StatusCode::NOT_FOUND,
+            grpc_status: tonic::Code::NotFound,
+            close_connection: false,
+            message: msg.to_string(),
         }
     }
 }
--- a/linkerd/app/core/src/errors/respond.rs
+++ b/linkerd/app/core/src/errors/respond.rs
@@ -12,10 +12,6 @@ use tracing::{debug, info_span, warn};
 
 pub const L5D_PROXY_ERROR: &str = "l5d-proxy-error";
 
-pub fn layer() -> respond::RespondLayer<NewRespond<super::DefaultRescue>> {
-    respond::RespondLayer::new(NewRespond(super::DefaultRescue))
-}
-
 pub trait Rescue<E> {
     fn rescue(&self, error: E) -> Result<Rescued, E>;
 }
@@ -27,6 +23,8 @@ pub struct Rescued {
     pub close_connection: bool,
     pub message: String,
 }
+
+pub type Layer<R> = respond::RespondLayer<NewRespond<R>>;
 
 #[derive(Copy, Clone, Debug)]
 pub struct NewRespond<R>(R);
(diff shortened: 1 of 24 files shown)


linkerd/app/core/src/errors/respond.rs
```diff
@@ -76,76 +121,138 @@ impl<ReqB, RspB: Default + hyper::body::HttpBody>
...
+        info!(%code, %message, "Handling error on gRPC connection");
+
+        http::Response::builder()
+            .version(http::Version::HTTP_2)
+            .header(http::header::CONTENT_LENGTH, "0")
+            .header(http::header::CONTENT_TYPE, GRPC_CONTENT_TYPE)
+            .header(GRPC_STATUS, code_header(code))
+            .header(
+                GRPC_MESSAGE,
+                HeaderValue::from_str("request timed out").unwrap_or_else(|error| {
```
kleimkuhler, line 227: Should this always be "request timed out" regardless of what `message` is?

olix0r: no, good catch

linkerd/app/outbound/src/http/server.rs
```diff
@@ -64,3 +67,35 @@ impl<N> Outbound<N> {
...
+    }
+}
+
+impl errors::HttpRescue<Error> for ServerRescue {
+    fn rescue(&self, error: Error) -> Result<errors::SyntheticHttpResponse> {
+        if Self::has_cause::<errors::ResponseTimeout>(&*error) {
+            return Ok(errors::SyntheticHttpResponse {
+                http_status: http::StatusCode::GATEWAY_TIMEOUT,
+                grpc_status: errors::Grpc::DeadlineExceeded,
+                close_connection: false,
```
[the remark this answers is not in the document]
olix0r: the other errors can be indicative of the target going away. with response timeouts, we're dealing with some user-configured timeout. there doesn't seem to be any reason to tear down the connection in this case--the request has probably reached the application, it's just slow.

kleimkuhler (approved):
Looks good 👍

linkerd/app/core/src/errors/respond.rs
```diff
@@ -1,57 +1,100 @@
...
+        error.is::<C>() || error.source().map(Self::has_cause::<C>).unwrap_or(false)
+    }
 }
 
+#[derive(Clone, Debug)]
+pub struct SyntheticHttpResponse {
+    pub grpc_status: tonic::Code,
+    pub http_status: http::StatusCode,
+    pub close_connection: bool,
+    pub message: String,
```
hawkw, line 34: nit (possibly a premature optimization): I notice sometimes these strings are formatted dynamically, but sometimes they are fixed static strings that are converted to `String`s. I wonder if we want to use [`Cow<'static, str>`](https://doc.rust-lang.org/std/borrow/enum.Cow.html) here so that we don't have to allocate a bunch of copies of the same string when it isn't dynamic?

we would then match on the `Cow` and call `HeaderValue::from_static` when the `Cow` has a static string in it (as a side note, i floated the idea to Sean of adding a `TryFrom<Cow<'static, str>>` to `http::header::HeaderValue` that would do that for us).

linkerd/app/core/src/errors/respond.rs
```diff
@@ -76,76 +121,138 @@ impl<ReqB, RspB: Default + hyper::body::HttpBody>
...
-                }
+        let span = info_span!(
+            "rescue",
+            client.addr = %self.client
+                .as_ref()
+                .map(|ClientHandle { addr, .. }| *addr)
+                .unwrap_or_else(|| {
+                    tracing::debug!("Missing client address");
+                    ([0, 0, 0, 0], 0).into()
+                })
```
hawkw, line 158: nit/TIOLI: maybe this should go in a method, so that the span macro can just call `self.client_addr()` or something? not a blocker

linkerd/app/core/src/errors/respond.rs
```diff
@@ -76,76 +121,138 @@ impl<ReqB, RspB: Default + hyper::body::HttpBody>
...
             }
         }
+
+        let rsp = if self.is_grpc {
+            Self::grpc_response(grpc_status, &*message)
+        } else {
+            Self::http_response(self.version, http_status, &*message, close_connection)
+        };
+
+        Ok(rsp)
```
hawkw, line 185: nit/TIOLI: could golf this down to 
```suggestion
        if self.is_grpc {
            Ok(Self::grpc_response(grpc_status, &*message))
        } else {
            Ok(Self::http_response(self.version, http_status, &*message, close_connection))
        }
```

or even
```suggestion
        Ok(if self.is_grpc {
            Self::grpc_response(grpc_status, &*message)
        } else {
             Self::http_response(self.version, http_status, &*message, close_connection)
        })
```
although maybe that's less readable?

linkerd/app/core/src/errors/respond.rs
```diff
@@ -76,76 +121,138 @@ impl<ReqB, RspB: Default + hyper::body::HttpBody>
...
+        }
+
+        rsp.body(B::default())
+            .expect("error response must be valid")
+    }
+
+    fn grpc_response<B: Default>(code: tonic::Code, message: &str) -> http::Response<B> {
+        info!(%code, %message, "Handling error on gRPC connection");
+
+        http::Response::builder()
```
hawkw, line 220: nit/TIOLI (possibly a premature optimization): we always add exactly 5 headers to the headermap here, so I wonder if we want to consider pre-allocating enough capacity for them? that way, we know it won't allocate multiple times...
```suggestion
        let mut builder = http::Response::builder();
        // We always add 5 headers, so preallocate enough capacity for them.
        builder
            .headers_mut()
            .expect("builder will not have an error")
            .reserve(5);
        builder
```

linkerd/app/core/src/errors/respond.rs
```diff
@@ -174,150 +286,60 @@ where
...
-        headers.insert(GRPC_MESSAGE, HeaderValue::from_static("request timed out"));
-        code
-    } else if let Some(e) = error.downcast_ref::<FailFastError>() {
-        let code = Code::Unavailable;
-        headers.insert(GRPC_STATUS, code_header(code));
-        headers.insert(
+impl<R, B> ResponseBody<R, B> {
+    fn grpc_trailers(code: tonic::Code, message: &str) -> http::HeaderMap {
+        debug!(grpc.status = ?code, "Synthesizing gRPC trailers");
+        let mut t = http::HeaderMap::new();
```
hawkw, line 327: nit/tioli: could be
```suggestion
        let mut t = http::HeaderMap::new(2);
```
to be explicit, although i am pretty sure the first insert will reserve enough capacity for at least 2 headers anyway...

linkerd/app/inbound/src/http/router.rs
```diff
@@ -415,3 +423,35 @@ impl Param<transport::labels::Key> for Http {
...
+        }
+
+        if Self::has_cause::<errors::ConnectTimeout>(&*error) {
+            return Ok(errors::SyntheticHttpResponse {
+                http_status: http::StatusCode::GATEWAY_TIMEOUT,
+                grpc_status: errors::Grpc::DeadlineExceeded,
+                close_connection: true,
+                message: error.to_string(),
+            });
+        }
```
hawkw, line 453: it's a little bit of a bummer that the two `has_cause` calls will traverse the cause chain twice...ideally, we could do it once, and check for _both_ types at every cause. but, this isn't really possible with the `has_cause` function, since it would have to be generic over a variadic number of type parameters, which isn't possible. i guess the alternative would just be to not use that function.

this is probably fine, though, and it's clearer than rewriting the cause chain traversal here. i don't know if this is hot enough for it to matter?

linkerd/app/inbound/src/http/server.rs
```diff
@@ -83,3 +85,54 @@ impl<H> Inbound<H> {
...
+            });
+        }
+
+        if Self::has_cause::<crate::GatewayLoop>(&*error) {
+            return Ok(errors::SyntheticHttpResponse {
+                http_status: http::StatusCode::LOOP_DETECTED,
+                grpc_status: errors::Grpc::Aborted,
+                close_connection: true,
+                message: error.to_string(),
+            });
```
hawkw, line 133: ...okay, this will recurse over the whole cause chain as much as 4 times. i wonder if maybe what we should do instead is have something like

```rust
fn traverse_causes<T>(error: &(dyn Error + 'static), f: impl FnMut(&(dyn Error + 'static)) -> Option<T>) -> Option<T> {
     f(error).or_else(error.source().map(|s| traverse_causes(s, f)))
}
```
so that we can pass in an error that tries to downcast to all the error types on each cause? that way we only recurse once.

but, this may not be that big a deal...

hawkw (approved):
overall, this looks good to me! i picked a bunch of obnoxious minor nits, but you can feel free to ignore them!

linkerd/app/inbound/src/http/server.rs
olix0r: Yeah, I'm aware that this isn't ideal, but it's exceptional by definition: we only expect to hit this when the proxy hits a runtime error; and, conservatively, the max chain depth we're going to see is like 10 (probably much lower than that in practice). So O(40) vs O(10) doesn't seem like a huge concern. I'll look at optimizing it but it seems doubtful it will have any actual impact.

hawkw: it's definitely not actually important!

PR #1248, opened 2021-09-08
Use a dedicated error type for HTTP response timeouts

olix0r:
We have several layers that are backed by `linkerd_timeout::Timeout` so
that its error type--ResponseTimeout--is somewhat ambiguous

This change consolidates the modules from `linkerd-timoeut` into
`linkerd-stack` to serve as a general timeout mechanism.
`linkerd_proxy_http::timeout` provides a stack that is parameterized
with a use-case-specific target and emits a use-case-specific error
type.

Error handlers are updated to use the HTTP-specific error type.

---

This sets up for handling response errors within route metrics.


kleimkuhler (approved):
👍

linkerd/stack/src/map_err.rs
```diff
@@ -0,0 +1,57 @@
...
+    type Request = P::Request;
+    type Response = P::Response;
+    type Error = E;
+    type Future = futures::future::MapErr<P::Future, F>;
+
+    #[inline]
+    fn proxy(&self, inner: &mut S, req: Req) -> Self::Future {
+        self.inner.proxy(inner, req).map_err(self.f.clone())
+    }
+}
```
hawkw, line 57: I was going to ask why we have our own version of `MapErr` instead of using tower's, but I realized it's because we can't really implement `Proxy` for `tower::util::MapErr`. Maybe worth adding a comment in this file for future readers?

hawkw (approved):
lgtm!

PR #1249, opened 2021-09-09
inbound: Load policies lazily

olix0r:
Rather than block inbound communication on policy discovery, this change
modifies policy initialization to use the default policy until policy
has been synced from the control plane. This will allow us to use
policies with the admin server.


kleimkuhler (approved):
👍

PR #1250, opened 2021-09-09
admin: Enforce policy on admin requests

olix0r:
Previously, the admin server admitted all requests. Now the admin server
will apply policies as discovered from the control plane. The proxy's
default policy is used before policy has been synced from the
controller.


kleimkuhler (approved):
Looks good

PR #1252, opened 2021-09-10
Improve logs around authorization decisions

olix0r:


linkerd/app/inbound/src/policy/authorize/http.rs
```diff
@@ -92,11 +92,13 @@ where
     fn call(&mut self, req: Req) -> Self::Future {
         match self.policy.check_authorized(self.client_addr, &self.tls) {
             Ok(permit) => {
+                tracing::debug!(?permit, tls = ?self.tls, client = %self.client_addr, "Request authorized");
                 self.metrics.allow(&permit);
                 let svc = self.inner.new_service((permit, self.target.clone()));
                 future::Either::Left(svc.oneshot(req).err_into::<Error>())
             }
             Err(e) => {
+                tracing::info!(tls = ?self.tls, client = %self.client_addr, "Request denied");
```
hawkw, line 101: should this include the policy's server label like the TCP events do, or does that come from somewhere else?

linkerd/app/inbound/src/policy/authorize/tcp.rs
```diff
@@ -122,42 +122,39 @@ where
...
+
+        let call = inner.call(io);
+        future::Either::Left(Box::pin(async move {
+            tokio::pin!(call);
+            loop {
+                tokio::select! {
+                    res = &mut call => return res.map_err(Into::into),
+                    _ = policy.changed() => {
+                        if let Err(denied) = policy.check_authorized(client, &tls) {
+                            tracing::info!(server = %policy.server_label(), ?tls, %client, "Connection terminated");
```
hawkw, line 151: IMO "Connection terminated" seems a little vague, I'm not sure if it will be obvious to a reader that this connection was terminated because of policy --- it might just be interpreted as "the connection was closed".

Maybe rephrase it to something like
```suggestion
                            tracing::info!(
                                server = %policy.server_label(),
                                ?tls, 
                                %client,
                                "Connection terminated due to policy change",
                             );
```
or even
IMO "Connection terminated" seems a little vague, I'm not sure if it will be obvious to a reader that this connection was terminated because of policy --- it might just be interpreted as "the connection was closed".

Maybe rephrase it to something like
```suggestion
                            tracing::info!(
                                server = %policy.server_label(),
                                ?tls, 
                                %client,
                                "Connection was no longer authorized after policy change; terminating",
                             );
```

linkerd/app/inbound/src/policy/authorize/tcp.rs
```diff
@@ -122,42 +122,39 @@ where
...
-                tls,
-                policy,
-                metrics,
-            }) => {
-                let client = *client;
-                let tls = tls.clone();
-                let mut policy = policy.clone();
-                let metrics = metrics.clone();
-
-                // FIXME increment counter.
```
hawkw, line 140: we still need to increment the counter, right?

hawkw (approved):
Overall this seems good. There was one log line that I thought felt too vague, it might be nice to rephrase that?

linkerd/app/inbound/src/policy/authorize/tcp.rs
olix0r: no, it's incremented above.

hawkw: ah, i can't read! nvm.

PR #1253, opened 2021-09-12
duplex: Reduce copy buffer size from 64KB to 8KB

olix0r:
Each forward TCP connection currently allocates 128KB--64KB for each
direction, client-to-server and server-to-client. This is an excessive
amount to allocate, especially since this buffer can currently hold only
a single `read(2)`'s worth of data.

This change reduces the buffer size to 8KB so that each connection
allocates only 16KB. This will substantially reduce memory overhead when
a server must serve many concurrent non-HTTP connections.


hawkw (approved):
looks good to me!

PR #1254, opened 2021-09-12
 stack: Replace BoxNewService with ArcNewService

olix0r:
Our `BoxNewService` type implements `Clone`, but this means that each
time it is cloned, we incur heap allocations. This change modifies the
`NewService` trait to take `self` as an immutable reference instead of a
mutable reference so that `NewService` instances may be accessed in an
`Arc` and replaces `BoxNewService` with `ArcNewService` so that clones
do not require allocation.

It turns out that no implementations of NewService actually required
mutable access.


hawkw (approved):
very nice! no complaints here!

PR #1256, opened 2021-09-12
inbound: Reduce needless allocation in stack targets

olix0r:
Inbound stacks include labels for the inbound server and authorization
names. Every time a stack target is cloned, however, these strings are
re-allocated.

This change modifies these label types to hold an `Arc<str>` so that
cloning the stack target is cheap, reusing the initial allocation.

### DIFF
--- a/linkerd/app/core/src/metrics/mod.rs
+++ b/linkerd/app/core/src/metrics/mod.rs
@@ -13,6 +13,7 @@ pub use linkerd_metrics::*;
 use std::{
     fmt::{self, Write},
     net::SocketAddr,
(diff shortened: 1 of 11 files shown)


hawkw (approved):
lgtm! i wonder if there's any other stuff in target types that should be reference counted...

olix0r:
> lgtm! i wonder if there's any other stuff in target types that should be reference counted...

I think _most_ of them are already arc'd internally (i.e. dns::Name, identity::Name, etc). These showed up in heap track, but I don't really see others.

PR #1258, opened 2021-09-12
inbound: Improve policy-based protocol detection

olix0r:
When an inbound server is configured with `proxyProtocol: HTTP/1`, we
have to do detection since client proxies may connect to us over HTTP/2.
But if the client is unmeshed or if detection fails, we can assume the
connection wasn't H2 upgraded.

This change modifies the inbound detection stack to use this improved
default behavior.


linkerd/app/inbound/src/detect.rs
```diff
@@ -230,14 +248,20 @@ impl<N> Inbound<N> {
...
                                 return Ok(svc::Either::B(Detect { timeout, tls }));
                             }
-                            Protocol::Http1 => {
-                                // HTTP/1 services may actually be transported over HTTP/2
-                                // connections between proxies, so we have to do detection.
+                            // Meshed HTTP/1 services may actually be transported over HTTP/2 connections
+                            // between proxies, so we have to do detection.
+                            //
+                            // TODO(ver) outbound clients should hint this with ALPN so we don't
+                            // have to detect this situation.
```
hawkw, line 255: should there be a ticket for this?

hawkw (approved):
looks good to me, thanks for the detailed comments!

PR #1259, opened 2021-09-13
inbound: Always permit connections from localhost

olix0r:
While it's unexpected for the inbound proxy to receive connections from
localhost, other servers--especially the admin server--are likely to
receive connections from localhost. These connections should be
implicitly allowed.

This change updates policy discovery to always permit connections from
loopback addresses 127.0.0.1 and ::1. This authorization uses the name
`default:localhost`.


linkerd/app/inbound/src/policy/discover.rs
```diff
@@ -109,6 +109,21 @@ fn to_policy(proto: api::Server) -> Result<ServerPolicy> {
         _ => return Err("proxy protocol missing".into()),
     };
 
+    let loopback = Authorization {
+        name: "default:localhost".into(),
+        authentication: Authentication::Unauthenticated,
+        networks: vec![
+            Network {
+                net: "127.0.0.1".parse::<IpAddr>().unwrap().into(),
```
hawkw, line 117: nit/TIOLI: maybe a bit nicer to just write
```suggestion
                net: IpAddr::from([127, 0, 0, 1]).into(),
```
rather than trying to parse a hardcoded string? not that parsing the string will ever fail, but it's nicer to not parse things that don't need to be parsed IMO...

linkerd/app/inbound/src/policy/discover.rs
```diff
@@ -109,6 +109,21 @@ fn to_policy(proto: api::Server) -> Result<ServerPolicy> {
...
+                except: vec![],
+            },
+            Network {
+                net: "::1".parse::<IpAddr>().unwrap().into(),
```
hawkw, line 121: similarly this could be 
```suggestion
                net: IpAddr::from([0, 0, 0, 0, 0, 0, 0, 1]).into(),
```

hawkw (approved):
lgtm overall! i thought it might be nicer to construct the `IpAddr`s from array literals rather than parsing them from string literals, but it's not a blocker.

PR #1269, opened 2021-09-16
inbound: Include TLS labels in authz metrics

olix0r:
Our inbound authorization metrics include a target port but include no
client information. This change adds TLS lables to the authorization
metrics so it's possible to identify which types of clients are
allowed/denied to access a service.


mateiidavid (approved):
Looks good!

kleimkuhler (approved):

PR #1272, opened 2021-09-16
tracing: use `Span::or_current` when spawning tasks

hawkw:
`tracing` v0.1.27 added a new [`Span::or_current`] method to aid with
efficiently propagating spans to spawned tasks (or threads). The method
is intended to be used when a spawned task should be instrumented with a
provided span if it is enabled, or the current span if the provided span
is _not_ enabled. `span.or_current()` returns `span` if it is enabled,
or, if `span` is disabled, it clones the current span and returns it.

This offers better performance than writing code like this:
```rust
    .instrument(span)
    .in_current_span()
```

First, `or_current` will only look up the current span (requiring a
thread-local access) _if_ `span` is not enabled, while
`.instrument(span).in_current_span()` will do this unconditionally.

Second, and more importantly, using `.in_current_span()` will wrap the
spawned future in _two_ `Instrumented` futures. One will enter the
current span whenever the task is polled, and inside that, the second
`instrument` will enter the passed span. This means we are entering two
spans instead of one, in the case where both are enabled. That adds a
small amount of additional overhead in every poll.

Therefore, this branch changes code that instruments a future prior to
spawning it to use `Span::or_current`. A lot of this is test code, which
(it turns out) doesn't really handle span propagation at all. However,
this also touches the main accept loop, as well as where we spawn admin
background tasks.

[`Span::or_current`]: https://docs.rs/tracing/0.1.27/tracing/span/struct.Span.html#method.or_current

### DIFF
--- a/linkerd/app/core/src/serve.rs
+++ b/linkerd/app/core/src/serve.rs
@@ -39,39 +39,38 @@ pub async fn serve<M, S, I, A>(
                     };
 
                     // The local addr should be instrumented from the listener's context.
-                    debug_span!("accept", client.addr = %addrs.param()).in_scope(|| {
-                        let accept = new_accept.new_service(addrs);
+                    let span = debug_span!("accept", client.addr = %addrs.param()).entered();
+                    let accept = new_accept.new_service(addrs);
 
-                        // Dispatch all of the work for a given connection onto a
-                        // connection-specific task.
-                        tokio::spawn(
-                            async move {
-                                match accept.ready_oneshot().err_into::<Error>().await {
-                                    Ok(mut accept) => {
-                                        match accept
-                                            .call(io::ScopedIo::server(io))
-                                            .err_into::<Error>()
-                                            .await
-                                        {
-                                            Ok(()) => debug!("Connection closed"),
-                                            Err(reason) if is_io(&*reason) => {
-                                                debug!(%reason, "Connection closed")
-                                            }
-                                            Err(error) => info!(%error, "Connection closed"),
+                    // Dispatch all of the work for a given connection onto a
+                    // connection-specific task.
+                    tokio::spawn(
+                        async move {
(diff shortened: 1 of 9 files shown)


linkerd/app/core/src/serve.rs
```diff
@@ -39,39 +39,38 @@ pub async fn serve<M, S, I, A>(
                     };
 
                     // The local addr should be instrumented from the listener's context.
-                    debug_span!("accept", client.addr = %addrs.param()).in_scope(|| {
-                        let accept = new_accept.new_service(addrs);
+                    let span = debug_span!("accept", client.addr = %addrs.param()).entered();
+                    let accept = new_accept.new_service(addrs);
```
olix0r, line 43: I assume the `entered` here means this span applies for the `new_service` call?

linkerd/app/core/src/serve.rs
```diff
@@ -39,39 +39,38 @@ pub async fn serve<M, S, I, A>(
...
+                                }
+                                Err(error) => {
+                                    warn!(%error, "Server failed to become ready");
                                 }
                             }
-                            .in_current_span(),
-                        );
-                    });
+                        }
+                        .instrument(span.exit().or_current()),
```
olix0r, line 72: And this exits the above-entered span? and then we use that or the parent?

olix0r (approved):
LGTM once it passes CI (there's an error in the test controller)

linkerd/app/core/src/serve.rs
hawkw: yup, that's correct! i thought this was nicer than the previous version, especially since it means that the `spawn` call will be inside that span (which doesn't really matter currently, but might be nice for console purposes eventually)...

linkerd/app/core/src/serve.rs
hawkw: yup!

PR #1276, opened 2021-09-17
error-respond: Support stack target configuration

olix0r:
The error responder is currently a `Service` and not a `NewService`
meaning there's no means to configure the responder using the stack
target.

This change introduces a `NewRespondService` type that implements
`NewService` to instrument the error responder. This will be used in
follow-up changes to configure the error responder to only emit headers
in certain conditions.


hawkw (approved):
pretty straightforward change, lgtm!

PR #1281, opened 2021-09-21
Support a `l5d-proxy-connection: close` header

olix0r:
In 6cf1b2851, we updated the outbound proxy to automatically teardown
connections when the peer proxy sets the `l5d-proxy-error` response
header; but this header may be set in some situations when disconnecting
isn't appropriate and will only generate more load on the system.

This change introduces a `l5d-proxy-connection` header (like the
`connection` and `proxy-connection` headers) that the proxy can use to
signal that the proxied connection should be closed, independently of
the informational `l5d-proxy-error` message.

Some TODOs have been marked for followup improvements.

### DIFF
--- a/linkerd/app/core/src/errors/respond.rs
+++ b/linkerd/app/core/src/errors/respond.rs
@@ -192,6 +192,7 @@ impl SyntheticHttpResponse {
 
         if self.close_connection {
             if version == http::Version::HTTP_11 {
+                // Notify the (proxy or non-proxy) client that the connection will be closed.
                 rsp = rsp.header(http::header::CONNECTION, "close");
             }
 
--- a/linkerd/app/core/src/errors/respond.rs
+++ b/linkerd/app/core/src/errors/respond.rs
@@ -12,6 +12,7 @@ use std::{
 };
 use tracing::{debug, info_span, warn};
 
+pub const L5D_PROXY_CONNECTION: &str = "l5d-proxy-connection";
 pub const L5D_PROXY_ERROR: &str = "l5d-proxy-error";
 
 pub fn layer<R, P: Clone, N>(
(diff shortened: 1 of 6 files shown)


linkerd/app/core/src/errors/respond.rs
```diff
@@ -176,11 +185,20 @@ impl SyntheticHttpResponse {
...
-            .header(http::header::CONTENT_LENGTH, "0")
-            .header(L5D_PROXY_ERROR, self.message());
+            .header(http::header::CONTENT_LENGTH, "0");
+
+        // TODO only set when client is trusted.
+        rsp = rsp.header(L5D_PROXY_ERROR, self.message());
+
+        if self.close_connection {
+            if version == http::Version::HTTP_11 {
+                rsp = rsp.header(http::header::CONNECTION, "close");
```
hawkw, line 195: why do we need to add a `connection: close` header here as well? is that intended for situations where the client isn't a linkerd proxy, or is it in case the client proxy doesn't handle the `l5d-proxy-connection` header? a comment could be nice...

linkerd/app/outbound/src/http/proxy_connection_close.rs
```diff
@@ -46,34 +59,51 @@ where
...
-                );
+impl<B, F: TryFuture<Ok = http::Response<B>>> Future for ResponseFuture<F> {
+    type Output = Result<http::Response<B>, F::Error>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        let rsp = futures::ready!(this.inner.try_poll(cx))?;
+
+        if let Some(proxy_conn) = rsp.headers().get(L5D_PROXY_CONNECTION) {
+            if proxy_conn == "close" {
```
hawkw, line 89: nit: I _think_ this could just be
```suggestion
        if let Some("close") = rsp.headers().get(L5D_PROXY_CONNECTION) {
```

hawkw (approved):
looks good to me (assuming there'll be some follow-ups for the TODO comments)

linkerd/app/core/src/errors/respond.rs
olix0r: this advertises to _clients_ -- proxies and not -- that we are going to close the connection. the `l5d-proxy-connection` header advertises that the peer proxy should close the _proxied_ (server) connection.

linkerd/app/outbound/src/http/proxy_connection_close.rs
olix0r: Unfortunately not.
```
  --> linkerd/app/outbound/src/http/proxy_connection_close.rs:88:29
   |
88 |         if Some("close") == rsp.headers().get(L5D_PROXY_CONNECTION) {
   |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `str`, found struct `http::HeaderValue`
   |
   = note: expected enum `std::option::Option<&str>`
              found enum `std::option::Option<&http::HeaderValue>`
```

`HeaderValue` impls `PartialEq<str>`; but I assume that `Option<T>` doesn't impl `PartialEq<Option<U>> where T: PartialEq<U>`

olix0r: See https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=86bad11540423cd5ece18ed6fb8dbdda

PR #1282, opened 2021-09-22
Avoid emitting informational headers to untrusted clients

olix0r:
The inbound proxy sets the `l5d-proxy-error` and `l5d-proxy-connection`
headers when an error is encountered. This leaks information about the
proxy to potentially untrusted clients.

This change modifies the error handling behavior to only emit these
headers from the inbound proxy when the client uses mesh identity.

In doing so, we modify the `inbound::NewSetIdentiyHeader` layer to use
`ExtractParam<tls::ConditionalServerTls, _>` to parameterize the server
connection's identity to be consistent with the rescue parameters. This
precipitated modifying the default impls for `ExtractParam` so that we
can provide an impl for `()` that simply uses the target type. A new
`CloneParam` type is introduced to replicate the former behavior.

### DIFF
--- a/linkerd/app/inbound/src/http/router.rs
+++ b/linkerd/app/inbound/src/http/router.rs
@@ -1,9 +1,9 @@
 use crate::{policy, stack_labels, Inbound};
 use linkerd_app_core::{
-    classify, dst, http_tracing, io, metrics,
+    classify, dst, errors, http_tracing, io, metrics,
     profiles::{self, DiscoveryRejected},
     proxy::{http, tap},
-    svc::{self, Param},
+    svc::{self, ExtractParam, Param},
     tls,
     transport::{self, ClientAddr, Remote, ServerAddr},
     Error, Infallible, NameAddr, Result,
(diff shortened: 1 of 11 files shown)


linkerd/app/inbound/src/http/router.rs
```diff
@@ -98,11 +95,6 @@ impl<C> Inbound<C> {
                 .into_new_service()
                 .push_new_reconnect(config.proxy.connect.backoff)
                 .push_map_target(Http::from)
-                // Handle connection-level errors eagerly so that we can report 5XX failures in tap
-                // and metrics. HTTP error metrics are not incremented here so that errors are not
-                // double-counted--i.e., endpoint metrics track these responses and error metrics
-                // track proxy errors that occur higher in the stack.
```
hawkw, line 104: does moving everything to `ServerRescue` affect this?

hawkw (approved):
cool that we were able to back out the metrics change and it still works --- lgtm!

PR #1283, opened 2021-09-22
outbound: Only honor the `l5d-proxy-connection` header when meshed

olix0r:
The proxy uses the `l5d-proxy-connection` header to determine when
it should close proxied connections. This header must only be set by a
peer inbound proxy.

This change modifies the outbound proxy so that this connection is only
honored when the peer is actually meshed. If this header is set by an
unmeshed peer, the header is ignored.

This change also modifies the outbound proxy to always _clear_ this
header so that it can't be sent to the application.

### DIFF
--- a/linkerd/app/outbound/src/http/proxy_connection_close.rs
+++ b/linkerd/app/outbound/src/http/proxy_connection_close.rs
@@ -2,7 +2,7 @@ use futures::prelude::*;
 use linkerd_app_core::{
     errors::respond::{L5D_PROXY_CONNECTION, L5D_PROXY_ERROR},
     proxy::http::ClientHandle,
-    svc,
+    svc, tls,
 };
 use std::{
     future::Future,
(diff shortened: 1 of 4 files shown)


linkerd/proxy/http/src/insert.rs
```diff
@@ -26,46 +44,83 @@ pub struct FnLazy<F>(F);
 #[derive(Clone, Debug)]
 pub struct ValLazy<V>(V);
 
-/// Wraps an HTTP `Service` so that a `P`-typed `Param` is cloned into each
-/// request's extensions.
+#[pin_project::pin_project]
 #[derive(Debug)]
```
hawkw, line 48: nit: this `Debug` impl will require `V: Debug` and `B: Debug`, because they're parameters of the type...but they aren't needed to implement `Debug`. we _probably_ should replace this with a manual impl (if we care)

linkerd/proxy/http/src/insert.rs
```diff
@@ -1,20 +1,38 @@
...
+pub struct NewInsert<P, N> {
+    inner: N,
+    _marker: PhantomData<fn() -> P>,
+}
+
+/// Wraps an HTTP `Service` so that a `P`-typed `Param` is cloned into each
+/// response's extensions.
+#[derive(Debug)]
+pub struct NewResponseInsert<P, N> {
+    inner: N,
```
hawkw, line 25: nit: these should _probably_ have manual `Debug` implementations so that they don't require `P: Debug`...but, this probably doesn't actually matter

linkerd/app/outbound/src/http/proxy_connection_close.rs
```diff
@@ -83,23 +83,43 @@ impl<B, F: TryFuture<Ok = http::Response<B>>> Future for ResponseFuture<F> {
...
+                    Some(tls::ConditionalClientTls::Some(_)) => {
+                        if let Some(error) = rsp
+                            .headers()
+                            .get(L5D_PROXY_ERROR)
+                            .and_then(|v| v.to_str().ok())
+                        {
+                            tracing::info!(%error, "Closing application connection for remote proxy");
+                        } else {
+                            tracing::info!("Closing application connection for remote proxy");
+                        }
```
hawkw, line 101: ...whew okay i should really merge the `impl Value for Option<T: Value>` PR in `tracing`...

hawkw (approved):
lgtm, couple nits that don't actually matter

PR #1284, opened 2021-09-22
outbound: Disable informational headers by config

olix0r:
The proxy may set error messages in the outbound proxy's response
headers; but sometimes this isn't desirable, especially on ingresses
where the repsonses may be sent to external clients.

This change introduces a new configuration,
`LINKERD2_PROXY_OUTBOUND_DISABLE_INFORMATIONAL_HEADERS` that can be set
to `false` to disable these headers from being included in responses
from the outbound proxy. These headers are also disabled by default when
ingress mode is enabled.


hawkw (approved):
looks good to me!

PR #1285, opened 2021-09-22
outbound: Strip peer-sent `l5d-proxy-error` headers

olix0r:
If the outbound proxy is not configured to emit headers, then we
shouldn't surface `l5d-proxy-error` headers sent by peer proxies.

This change adds a module to the outbound HTTP endpoint stack to strip
these headers when appropriate.

### DIFF
--- a/linkerd/app/outbound/src/http/strip_proxy_error.rs
+++ b/linkerd/app/outbound/src/http/strip_proxy_error.rs
@@ -1,0 +1,37 @@
+use linkerd_app_core::{
+    errors::respond::L5D_PROXY_ERROR,
+    proxy::http,
+    svc::{self, NewService},
+};
+
+#[derive(Clone, Debug)]
+pub struct NewStripProxyError<N> {
+    strip: bool,
+    inner: N,
+}
+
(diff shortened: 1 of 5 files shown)


linkerd/app/outbound/src/http/strip_proxy_error.rs
```diff
@@ -0,0 +1,37 @@
...
+    type Service = http::strip_header::response::StripHeader<&'static str, N::Service>;
+
+    fn new_service(&self, target: T) -> Self::Service {
+        let header = if self.strip {
+            Some(L5D_PROXY_ERROR)
+        } else {
+            None
+        };
+        let inner = self.inner.new_service(target);
+        http::StripHeader::response(header, inner)
```
hawkw, line 35: another approach would be to return an `Either` service here that either includes the `StripHeader` or doesn't. but, that might make builds slower by increasing type complexity? it would, however, avoid having to make the changes in `StripHeader`.

hawkw (approved):
this seems fine to me! i'm unsure if modifying `strip-header` to take an `Option` is the _best_ approach, but it seems fine?

PR #1286, opened 2021-09-23
stack-tracing: Log when services are dropped

olix0r:
This change modifies `stack-tracing`'s `Instrument` service type to log
a message as instances are dropped, which can be useful for diagnosing
stack retention issues.

This change also modifies this type to stop using `pin-project` (since
it wasn't actually using it).

### DIFF
--- a/linkerd/stack/tracing/src/lib.rs
+++ b/linkerd/stack/tracing/src/lib.rs
@@ -192,12 +192,8 @@ where
 
 impl<T, G: GetSpan<T>, S> Drop for Instrument<T, G, S> {
     fn drop(&mut self) {
-        if let Some(span) = self.current_span.take() {
-            trace!(parent: &span, "drop");
-        } else {
-            let _span = self.get_span().entered();
-            trace!("drop");
-        }
+        let span = self.current_span.take().unwrap_or_else(|| self.get_span());
+        trace!(parent: &span, "drop");
     }
 }
 
--- a/linkerd/stack/tracing/src/lib.rs
+++ b/linkerd/stack/tracing/src/lib.rs
@@ -2,7 +2,6 @@
 #![forbid(unsafe_code)]
 
 use linkerd_stack::{layer, NewService, Proxy};
-use pin_project::pin_project;
 use std::task::{Context, Poll};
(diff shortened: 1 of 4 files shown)


linkerd/stack/tracing/src/lib.rs
```diff
@@ -25,9 +24,8 @@ pub struct NewInstrument<G, N> {
 }
 
 /// Instruments a service produced by `NewInstrument`.
-#[pin_project]
 #[derive(Debug)]
-pub struct Instrument<T, G, S> {
+pub struct Instrument<T, G: GetSpan<T>, S> {
```
hawkw, line 28: any particular reason to always require this bound?

hawkw (approved):
👍 looks good to me!

linkerd/stack/tracing/src/lib.rs
olix0r: We need it on the `Drop` and all bound on the `Drop` must also be on the struct definition.

linkerd/stack/tracing/src/lib.rs
```diff
@@ -193,6 +190,17 @@ where
...
 }
 
+impl<T, G: GetSpan<T>, S> Drop for Instrument<T, G, S> {
+    fn drop(&mut self) {
+        let _span = self
+            .current_span
+            .take()
+            .unwrap_or_else(|| self.get_span())
+            .entered();
+        trace!("drop");
```
hawkw, line 200: nit, take it or leave it: if this used
```suggestion
        trace!(parent:  parent: self.current_span.as_ref(), "drop");
```
we wouldn't enter the span when the trace event isn't enabled. this might be a minor perf improvement when the span is at a higher level (e.g. `DEBUG`), and the drop event isn't enabled.

we can't easily create the if we do this, though, because we would need to hold it for longer than passing it to `parent:` --- otherwise, if it had just been created, it will be immediately dropped.

anyway, not really a blocker; current approach is fine!

linkerd/stack/tracing/src/lib.rs
hawkw: yup, i just realized that!

PR #1287, opened 2021-09-23
Add trace-level logging around discovery updates

olix0r:
It's helpful to have more insight into what's going on during endpoint
resolution. This change adds trace-level logging messages to various
resolver modules.

### DIFF
--- a/linkerd/proxy/discover/src/from_resolve.rs
+++ b/linkerd/proxy/discover/src/from_resolve.rs
@@ -9,6 +9,7 @@ use std::{
     task::{Context, Poll},
 };
 use tower::discover::Change;
+use tracing::trace;
 
 #[derive(Clone, Debug)]
 pub struct FromResolve<R, E> {
@@ -97,7 +98,7 @@ impl<R: TryStream, E> Discover<R, E> {
 impl<R, E> Stream for Discover<R, E>
 where
     R: TryStream<Ok = Update<E>>,
-    E: Clone,
+    E: Clone + std::fmt::Debug,
 {
     type Item = Result<Change<SocketAddr, E>, R::Error>;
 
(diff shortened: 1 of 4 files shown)


linkerd/proxy/dns-resolve/src/lib.rs
```diff
@@ -84,7 +85,7 @@ async fn resolution(dns: dns::Resolver, na: NameAddr) -> Result<UpdateStream, Er
             loop {
                 match dns.resolve_addrs(na.name(), na.port()).await {
                     Ok((addrs, expiry)) => {
-                        debug!(?addrs);
+                        debug!(?addrs, name = %na);
```
hawkw, line 88: should we maybe just stick the spawned task in a debug span with the `name` field? (we could use `Span::or_current`)

that way, we would also get the nameaddr on the "Closed" and error events.

linkerd/proxy/discover/src/from_resolve.rs
```diff
@@ -97,29 +98,36 @@ impl<R: TryStream, E> Discover<R, E> {
...
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         loop {
             let this = self.as_mut().project();
             if let Some(change) = this.pending.pop_front() {
+                trace!(?change, "Changed");
                 return Poll::Ready(Some(Ok(change)));
             }
 
+            trace!("poll");
```
hawkw, line 113: i'm assuming we're inside a span that would give more context about what's being polled, here? so this won't just spam "poll" a bunch?

hawkw (approved):

linkerd/proxy/dns-resolve/src/lib.rs
olix0r: we should usually have it on the parent span already, maybe not really necessary here

PR #1288, opened 2021-09-23
Add tracing spans to controller clients

olix0r:
It can be difficult to identify when log messages are emitted from
controller clients. This change adds tracing spans to identify the
client and its endpoints.

### DIFF
--- a/linkerd/app/core/src/control.rs
+++ b/linkerd/app/core/src/control.rs
@@ -36,9 +36,7 @@ impl fmt::Display for ControlAddr {
 type BalanceBody =
     http::balance::PendingUntilFirstDataBody<tower::load::peak_ewma::Handle, hyper::Body>;
 
-type RspBody = linkerd_http_metrics::requests::ResponseBody<BalanceBody, classify::Eos>;
-
-pub type Client = svc::Buffer<http::Request<tonic::body::BoxBody>, http::Response<RspBody>, Error>;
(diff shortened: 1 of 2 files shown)


linkerd/app/core/src/control.rs
```diff
@@ -87,12 +93,14 @@ impl Config {
...
             // drive them all directly.
             .push_on_service(svc::layer::mk(svc::SpawnReady::new))
+            .instrument(|t: &self::client::Target| tracing::info_span!("endpoint", addr = %t.addr))
             .push(self::resolve::layer(dns, resolve_backoff))
             .push_on_service(self::control::balance::layer())
             .into_new_service()
             .push(metrics.to_layer::<classify::Response, _, _>())
             .push(self::add_origin::layer())
             .push_on_service(svc::layers().push_spawn_buffer(self.buffer_capacity))
+            .instrument(|c: &ControlAddr| tracing::info_span!("controller", addr = %c.addr))
```
hawkw, line 103: just double-checking that there are, in fact, `INFO`-level events (or `WARN`/`ERROR`) that will occur inside these, right?

hawkw (approved):

linkerd/app/core/src/control.rs
olix0r: yeah we mostly want connection errors, etc, to be logged with context.

PR #1289, opened 2021-09-23
Avoid attempting to reconnect to defunct endpoints

olix0r:
Load balanced clients can get stuck continually trying to reconnect to
defunct endpoints in situations like the following:

1. A balancer has an existing endpoint but it is pending;
2. A new ready endpoint is added to the balancer;
3. Requests are sent to the new endpoint and the balancer ends up with
   the new endpoint being ready.
4. Service discovery issues an update indicating that we should no
   longer use the old endpoint, which has now shutdown.

This discovery update won't be processed by the balancer until we
attempt to issue a request on the balancer. But, because each endpoint
uses `SpawnReady` to attempt reconnection on a background task, we
continually attempt to reconnect to the defunct endpoint even though
we'll never actually issue requests to it (because it will be removed as
soon as the balancer is polled again).

There's a simple fix to this: we shouldn't put reconnection inside of
`SpawnReady`. Instead, we use a `SpawnReady` to drive each individual
connection attempt to readiness, but we don't drive reconnect to drive a
failed connection to be retried until the balancer has a chance to be
updated.

This may address linkerd/linkerd2#6842 and should fix another issue
reported in Slack where controller pods would continually log connection
failures after the destination controller was rescheduled.

### DIFF
--- a/linkerd/app/outbound/src/http/logical.rs
+++ b/linkerd/app/outbound/src/http/logical.rs
@@ -60,17 +60,12 @@ impl<E> Outbound<E> {
                 .clone()
                 .check_new_service::<Endpoint, http::Request<http::BoxBody>>()
                 .push_on_service(
-                    svc::layers()
-                        .push(http::BoxRequest::layer())
-                        .push(
-                            rt.metrics
-                                .proxy
-                                .stack
-                                .layer(stack_labels("http", "balance.endpoint")),
-                        )
-                        // Ensure individual endpoints are driven to readiness so that
-                        // the balancer need not drive them all directly.
-                        .push(svc::layer::mk(svc::SpawnReady::new)),
+                    svc::layers().push(http::BoxRequest::layer()).push(
+                        rt.metrics
+                            .proxy
+                            .stack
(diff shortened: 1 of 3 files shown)


linkerd/app/outbound/src/http/logical.rs
```diff
@@ -60,24 +60,25 @@ impl<E> Outbound<E> {
...
                 //
                 // If the balancer has been empty/unavailable, eagerly fail requests.
                 // When the balancer is in failfast, spawn the service in a background
                 // task so it becomes ready without new requests.
+                //
+                // We *don't* ensure that the endpoint is driven to readiness here, because this
+                // might cause us to continually attempt to reestablish connections without
+                // consulting discovery to see whether the endpoint has been removed. Instead, the
+                // endpoint layer spawns each _connection_ attempt on a background task, but the
+                // decision to attempt the connection must be driven by the balancer.
```
hawkw, line 81: <3

hawkw (approved):
okay, i see what's going on here, this is hairy but the change makes sense! thanks for the comments!

PR #1295, opened 2021-10-01
Adopt Rust edition 2021

olix0r:
Depends on Rust 1.56.0 which is due out on 10/09.

### DIFF
--- a/hyper-balance/Cargo.toml
+++ b/hyper-balance/Cargo.toml
@@ -3,7 +3,7 @@ name = "hyper-balance"
 version = "0.1.0"
 authors = ["Linkerd Developers <cncf-linkerd-dev@lists.cncf.io>"]
 license = "Apache-2.0"
-edition = "2018"
+edition = "2021"
 publish = false
 
 [dependencies]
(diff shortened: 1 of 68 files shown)


hawkw:
It occurs to me that we may also want to consider adopting [Rust 2018-style modules](https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-modrs)? 

In the new style modules, the module goes in a file named `<module_name>.rs` even if it has submodules in separate files, and they go in `<module_name>/<submodule>.rs`, rather than having to move the module to `<module_name>/mod.rs` when adding submodules. This would mean changing a _bunch_ of existing code away from using `mod.rs`, but it does have the advantage that submodules can be added to a module in separate files _without_ having to rename and move the module file. This might make commits that add submodules less painful to review.

Obviously, this isn't directly linked to the Rust 2018 -> 2021 update, and should be addressed separately, but it's just a thought that occurred to me...

olix0r:
@hawkw interesting! that's good to know. If it's only moving and doesn't require code changes, seems easy to do (and automate ;). I'll take look at that.

PR #1297, opened 2021-10-01
Require ALPN on gateway connections

olix0r:
In Linkerd stable-2.10.0, the proxy started using a specialized protocol
header for multicluster gateway connections (`transport.l5d.io`). Now
that stable-2.11.0 has been released, we can drop support for gateway
connections that do not use this protocol.


hawkw (approved):
yay for deleting things!

mateiidavid (approved):

PR #1299, opened 2021-10-06
tracing: use `Option`s as `tracing` values

hawkw:
Tracing 0.1.29 added a `tracing::field::Value` implementation for
`Option<T> where T: Value`. This means that `Option`s can now be used as
structured field values, rather than recording them with `fmt::Debug`.
When the `Option` is `None`, the field will simply be skipped.

This allows us to make a bunch of existing code simpler. There are a
number of places where we create one span or event if some value is
`Some`, and a different span/event without that field if the value is
`None`. This can be changed to just use the `Option` as a field value
directly.

I'm almost certain I missed some of those places, but I changed the ones
I could find.

This also updates `tracing` to 0.1.29, of course.

### DIFF
--- a/linkerd/app/outbound/src/tcp/logical.rs
+++ b/linkerd/app/outbound/src/tcp/logical.rs
@@ -70,7 +70,7 @@ where
                     debug_span!(
                         "endpoint",
                         server.addr = %t.addr,
-                        server.id = t.tls.value().map(|tls| tracing::field::debug(&tls.server_id)),
+                        server.id = t.tls.value().map(|tls| tracing::field::display(&tls.server_id)),
                     )
                 })
                 .push(resolve::layer(resolve, config.proxy.cache_max_idle_age * 2))


linkerd/app/inbound/src/http/router.rs
```diff
@@ -219,11 +219,12 @@ impl<C> Inbound<C> {
                         .push(http::BoxResponse::layer()),
                 )
                 .check_new_service::<Logical, http::Request<http::BoxBody>>()
-                .instrument(|t: &Logical| match (t.http, t.logical.as_ref()) {
-                    (http::Version::H2, None) => debug_span!("http2"),
-                    (http::Version::H2, Some(name)) => debug_span!("http2", %name),
-                    (http::Version::Http1, None) => debug_span!("http1"),
-                    (http::Version::Http1, Some(name)) => debug_span!("http1", %name),
+                .instrument(|t: &Logical| {
+                    let name = t.logical.as_ref().map(tracing::field::display);
```
olix0r, line 223: I assume this is basically ~free? since this is being used in a debug_span, we wouldn't want to do any work in most cases...

linkerd/app/outbound/src/tcp/logical.rs
```diff
@@ -66,13 +66,12 @@ where
...
-                        debug_span!("endpoint", server.addr = %t.addr, server.id = ?tls.server_id)
-                    }
-                    Conditional::None(_) => {
-                        debug_span!("endpoint", server.addr = %t.addr)
-                    }
+                .instrument(|t: &Endpoint| {
+                    debug_span!(
+                        "endpoint",
+                        server.addr = %t.addr,
+                        server.id = t.tls.value().map(|tls| tracing::field::debug(&tls.server_id)),
```
olix0r, line 73:
```suggestion
                        server.id = t.tls.value().map(|tls| tracing::field::display(&tls.server_id)),
```

I think this was debug because it was an option

olix0r (approved):

PR #1318, opened 2021-10-19
Fix an infinite loop when downgrading HTTP/2 errors

olix0r:
578d979 introduced a bug: when the proxy handles errors on
HTTP/2-upgraded connections, it can get stuck in an infinite loop when
searching the error chain for an HTTP/2 error.

This change fixes this inifite loop and adds a test that exercises this
behavior to ensure that `downgrade_h2_error` behaves as expected.

Fixes linkerd/linkerd2#7103


olix0r:
@hawkw quick spot check shows that the other error handlers look okay. most of them are recursive, which is probably better in this case ;)

PR #1319, opened 2021-10-20
trace: Apache Common Log Format access logging

hawkw:
This branch builds on @tustvold's work in #601. The original PR
description from that branch:

> Access logging is very important functionality for my team as we wish
> to maintain feature-parity with our existing AWS ALB-based approach.
> This functionality was requested
> [here](https://github.com/linkerd/linkerd2/issues/1913) and was marked
> as help wanted, so thought I'd take a stab at implementing it.
>
> Creating as a draft as still needs more testing and benchmarking, and
> I'm new to tower and so might have made some rookie errors. However, I
> wanted to create a draft as an opportunity to get some early feedback.
>
> The basic design consists of an AccessLogLayer that instruments both
> requests and responses that flow through it, in a similar manner to
> how handle_time is already computed. I'm not a massive fan of this,
> but it was the only way I could easily see to get accurate processing
> time metrics. I've tried to avoid any memory allocation on the hot
> path, although there are possibly more atomic increments than I would
> like. The performance impact with the feature disabled, i.e.
> LINKERD2_PROXY_ACCESS_LOG_FILE, not set should be minimal.
>
> The results of this instrumentation are then sent over a mpsc channel
> to an AccessLogCollector that writes them in a space-delimited format
> to a file specified as an environment variable. It buffers in memory
> and flushes on termination and on write if more than
> FLUSH_TIMEOUT_SECS since the last flush. This makes the access logging
> best effort much like AWS ALBs.
>
> An example deployment scenario using this functionality might deploy a
> fluent-bit sidecar to ship the logs, or write to /dev/stderr and use a
> log shipper deployed as a DaemonSet.

The additional changes in this branch are:

 - Updated against the latest state of the `main` branch.
 - Changed the `tracing` configuration to use per-layer filtering, so that
   the access log layer _only_ sees access log spans, while the stdout
   logging layer does not see the access log spans (although, it _could_
   if we wanted it to...)
 - Changed the format for outputting the access log to the Apache Common
   Log Format. Note that this format does *not* include all the data that
   the access log spans currently collect; I excluded that data so that
   the output is compatible with tools that ingest the Apache log format.
   In a follow-up PR, we can add the ability to control what format the
   access log is written in, and add an alternative format that includes
   all the access log data that Linkerd's spans can collect (I suggest
   newline-delimited JSON for this).

Of course, a huge thank you to @tustvold for all their work on this; I
only updated the branch with the latest changes and made some minor
improvements. :)

Co-authored-by: Raphael Taylor-Davies <r.taylordavies@googlemail.com>

Closes #601


linkerd/tracing/src/access_log.rs
```diff
@@ -0,0 +1,187 @@
...
+use tracing_appender::non_blocking::{NonBlocking, WorkerGuard};
+use tracing_subscriber::{
+    field::RecordFields,
+    filter::{Directive, FilterFn, Filtered},
+    fmt::{format, FormatFields, FormattedFields, MakeWriter},
+    layer::{Context, Layer},
+    registry::LookupSpan,
+};
+
+const ENV_ACCESS_LOG: &str = "LINKERD2_PROXY_ACCESS_LOG";
```
olix0r, line 12: Stepping back from the code for a second: how is this actually supposed to be used in practice? How would a Linkerd user configure the proxy so that it had a volume to which access log files could be written?

I can understand emitting logs to stdout or stderr, but it's not clear to me how output to a file could actually be used.

hawkw: Honestly, I have no idea; this is just what @tustvold's PR #601 did, and since that was opened by a potential user of this feature, I had kind of assumed they already knew how the logs were going to be collected, and that emitting the logs to a user configured file was the best way to integrate with whatever existing systems users are already using...

But, thinking through it...this would probably require adding some kind of volume mount to the proxy container, and because the proxy containers are added to pods by the proxy-injector _after_ the user applies the configuration for a pod, this might be kind of hard to do? Perhaps it's worth getting additional user input about how people want to get access logs out of Linkerd...

tustvold: I'm afraid my memory is a little bit hazy as to why I implemented it this way, but I think it was simply a case of replicating what Envoy does, combined with it being a simple way for the user to choose to be able to choose which of stdout or stderr to route logs to... There might have also been something about it being easier to write tests for, but I don't really recall I'm afraid.

For the reasons you articulate logging to a file in an environment making using of the admission webhook to inject the sidecars would be non-trivial, however, it could theoretically be used when doing injection manually. That being said I'm not aware of any environments doing this.

TLDR I don't think file output is needed per se, but I'm not sure it particularly hurts either?

hawkw: > combined with it being a simple way for the user to choose to be able to choose which of stdout or stderr to route logs to.

Ah, if the intention is that the user would just set the environment variable to `/dev/stderr` or `/dev/stdout` as appropriate, that _should_ work. I guess I'd have to revise the follow-up PR #1350 to add a separate env variable for controlling the access log format, though --- if the user is going to set the env var to `/dev/stderr`, we obviously can't enable JSON access logging with a filename extension in that case! :)

jon-depop:
New year bump :balloon:

olix0r:
@jon-depop While the code in this PR is in good shape, it is currently blocked on having a story on how users would interact with this feature. Specifically: where should these logs be written? It's not feasible for Linkerd to emit these logs to a file, as this would require volume mounts in all pods that emit access logs. We could alternatively intermingle these logs with the other proxy diagnostic logs, but I'm not sure how valuable that would actually be. If you are interested in helping to flesh this out, we should probably move that discussion into https://github.com/linkerd/linkerd2/issues/1913

0x91:
> @jon-depop While the code in this PR is in good shape, it is currently blocked on having a story on how users would interact with this feature. Specifically: where should these logs be written? It's not feasible for Linkerd to emit these logs to a file, as this would require volume mounts in all pods that emit access logs. We could alternatively intermingle these logs with the other proxy diagnostic logs, but I'm not sure how valuable that would actually be. If you are interested in helping to flesh this out, we should probably move that discussion into [linkerd/linkerd2#1913](https://github.com/linkerd/linkerd2/issues/1913)

Intermingling them isn't perfect, but it would still yield useful information. I believe this is what the default nginx configuration in a container does too.

jon-depop:
As @0x91 said, we are currently intermingling the output as part of our POC (using our log processor to separate them into different streams). We would like to eventually have access logs to stderr and linkerd proxy diagnostics to stdout for more streamline log forwarding. We do not have any requirement for the logs to be written to a specific filepath.

Apologies for the delay in response @olix0r

jon-depop:
thoughts @olix0r?

olix0r:
@jon-depop I think that's a reasonable approach.

I'm not an expert on kubernetes' logging infrastructure, but as far as I know, it's not possible to decouple stdout and stderr (based on https://kubernetes.io/docs/concepts/cluster-administration/logging/ and https://github.com/kubernetes/kubernetes/issues/28167); so I'm not sure that this distinction actually matters, practically. Am I wrong about this?

I only bring this up because unifying the output streams might simplify implementation (I'm not sure). If Kubernetes doesn't support demultiplexing the log streams, I'd opt for whatever is simplest code-wise.

I'll also note for the sake of completeness that this feature will need to be opt-in (via annotation). Per-request logging will definitely impact performance and logging costs, so we'll want to avoid this overhead by default.

jon-depop:
From our current investigations, it seems we should be able to achieve this as the Kubelet supports [splitting stdout and stderr log streams](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#JSONOptions)

We will be confirming this with a spike this week, will report as soon as we find out to allow you to make your preference in implementation.

I also 100% agree that this should be opt in, thanks for the heads up @olix0r.

linkerd/tracing/src/access_log.rs
```diff
@@ -0,0 +1,166 @@
...
+    res: fmt::Result,
+    writer: format::Writer<'writer>,
+}
+
+pub(super) fn build<S>() -> Option<(AccessLogLayer<S>, Directive)>
+where
+    S: Subscriber + for<'span> LookupSpan<'span>,
+{
+    // Is access logging enabled?
+    let _ = std::env::var(ENV_ACCESS_LOG).ok()?;
```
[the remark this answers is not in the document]
olix0r: Or if we can't move this into `linkerd_app`, could we at least make this part of `linkerd_tracing::Settings::from_env`?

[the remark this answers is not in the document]
hawkw: i don't remember off the top of my head, but i think there might have been a reason we defined the logging env vars here instead, maybe to avoid a cyclic dependency? but i'll take a look and see if that's possible!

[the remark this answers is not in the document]
hawkw: can definitely at least move it into `from_env` though.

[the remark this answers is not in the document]
hawkw: forgot to reply to this earlier, but, yeah, the reason the logging env vars aren't in `linkerd-app` is because `linkerd-app` depends on `linkerd-tracing`, so it would create a cyclical dependency (i double checked). i did move it into `Settings::from_env` though.

linkerd/http-access-log/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+            request_bytes = get_header(http::header::CONTENT_LENGTH),
+            status = field::Empty,
+            response_bytes = field::Empty,
+            total_ns = field::Empty,
+            processing_ns = field::Empty,
+            user_agent = get_header(http::header::USER_AGENT),
+            host = get_header(http::header::HOST),
+        );
+
+        if span.is_disabled() {
```
olix0r, line 125: I think it's worth leaving a comment here making it clear that the `span!` body is only materialized when access logging is enabled by `linkerd_tracing::access_log`'s initializer. I.e. this is expected to be free when tracing is not enabled.

Maybe also worth leaving a comment on the `fn layer` summarizing this, too.

olix0r (commented):
I'm going to test this a little and then I think this is basically good to merge!

My only feedback asks to make the contract/overhead a bit clearer in comments.

hawkw:
Great, I'll add some comments.

FWIW, you can't currently test the access logging functionality using the integration tests, as the `linkerd-tracing::Settings::for_test` constructor doesn't enable access logging. I can change that if we care!

olix0r:
ran a load test (~20K RPS) with logging disabled and then enabled

image

about 5% more CPU when logging, and higher tail latencies.

olix0r:
```text
:; k logs -n ort deploy/server-000 -c linkerd-proxy  |head
10.42.0.31:46616 load-http.ort.serviceaccount.identity.linkerd.cluster.local - [2022-01-19T23:15:55.389555918Z] "GET http://logical:8080/?latency_ms=0&size=0 HTTP/2.0" 200
10.42.0.31:46616 load-http.ort.serviceaccount.identity.linkerd.cluster.local - [2022-01-19T23:15:55.388657397Z] "GET http://logical:8080/?latency_ms=0&size=0 HTTP/2.0" 200
10.42.0.31:46616 load-http.ort.serviceaccount.identity.linkerd.cluster.local - [2022-01-19T23:15:55.388702235Z] "GET http://logical:8080/?latency_ms=0&size=0 HTTP/2.0" 200
10.42.0.31:46616 load-http.ort.serviceaccount.identity.linkerd.cluster.local - [2022-01-19T23:15:55.388907033Z] "GET http://logical:8080/?latency_ms=0&size=0 HTTP/2.0" 200
10.42.0.31:46616 load-http.ort.serviceaccount.identity.linkerd.cluster.local - [2022-01-19T23:15:55.388918324Z] "GET http://logical:8080/?latency_ms=0&size=0 HTTP/2.0" 200
10.42.0.31:46616 load-http.ort.serviceaccount.identity.linkerd.cluster.local - [2022-01-19T23:15:55.388914521Z] "GET http://logical:8080/?latency_ms=0&size=0 HTTP/2.0" 200
10.42.0.31:46616 load-http.ort.serviceaccount.identity.linkerd.cluster.local - [2022-01-19T23:15:55.389215678Z] "GET http://logical:8080/?latency_ms=0&size=0 HTTP/2.0" 200
10.42.0.31:46616 load-http.ort.serviceaccount.identity.linkerd.cluster.local - [2022-01-19T23:15:55.389223877Z] "GET http://logical:8080/?latency_ms=0&size=0 HTTP/2.0" 200
10.42.0.31:46616 load-http.ort.serviceaccount.identity.linkerd.cluster.local - [2022-01-19T23:15:55.389228151Z] "GET http://logical:8080/?latency_ms=0&size=0 HTTP/2.0" 200
10.42.0.31:46616 load-http.ort.serviceaccount.identity.linkerd.cluster.local - [2022-01-19T23:15:55.389231928Z] "GET http://logical:8080/?latency_ms=0&size=0 HTTP/2.0" 200
```

hawkw:
hmm, tail latency bump is kinda bad...there's probably some things we can do about that.

olix0r (approved):

0x91:
thank you @olix0r and @hawkw for the work on this!

jon-depop:
Many thanks @olix0r and @hawkw, also a really special thanks to @tustvold for starting this all in the first place! :tada:

PR #1341, opened 2021-10-29
retry: allow retrying requests without Content-Length headers 

hawkw:
Currently, the proxy will retry requests with bodies if and only if they
have a `Content-Length` header with a value <= 64 KB. If the request has
a body but no `Content-Length` header, we currently assume that its body
will exceed the maximum buffered size, and skip trying to retry it.
However, this means gRPC requests will never be retried, because it
turns out gRPC requests don't include a `Content-Length` header (see
linkerd/linkerd2#7141). Whoops!

This PR fixes this by changing the retry logic so that `Content-Length`
is treated as a _hint_, rather than a requirement. Now, we will
preemptively skip retrying requests with bodies when the there is a
`Content-Length` header with a value that exceeds the limit, but if no
`Content-Length` header is present, we will still attempt to buffer the
body for a potential retry.

We are still protected against unbounded buffering because the buffering
body will stop buffering and discard any previously buffered data if the
buffered length ever exceeds the maximum. Requiring a `Content-Length`
simply allowed us to skip _trying_.

Furthermore, now that we actually expect the cap on buffer length to
actually be applied as part of normal operation (rather than as a
safeguard against requests where the `Content-Length` header is a lie),
I had to make some additional changes to the retry policy to ensure that
the original error is propagated from the server when hitting the cap.

Currently, when we have capped a body that exceeded the buffered bytes
limit, the `ReplayBody` implementation of `poll_data` returns an error
indicating that the body was capped. When this occurs, `hyper` fails the
request, and we synthesize a 500 error. This was fine when the capping
was only triggered as a safeguard against an evil request with a lying
`Content-Length`. Now, however, we will _often_ hit the cap, for
requests that omit a `Content-Length` header. In this case, it would be
preferable to return the original error from the server. Therefore, I've
modified the retry policy to check if the body was capped on a previous
iteration of the request, and skip retrying if it was.

This should probably fix linkerd/linkerd2#7141, although it might be
worth testing with a non-Rust gRPC implementation to make sure it works?

I did, however, change the existing test that expected requests without
`Content-Length` to never be retried, and added a  new one to exercise
the capping behavior when the body does exceed the max length. It turns
out that some of the retry tests were actually subtly broken, in that chunked
bodies were never actually being sent (we forgot to `await` the `send_data`
futures...).


olix0r:
Not necessarily a blocker but as I read through this I have a question. What happens with retries in the following scenario:

* A POST request is issued
  * That is retryable;
  * With a body <64K.
* A response is received
  * With a 2xx response;
  * With a body of an arbitrary/streamed length;
  * That does not complete.

When is the request body data dropped?

olix0r:
> When is the request body data dropped?

Refreshed myself on the retry impl. If I understand correctly the cloned request (and its body) are dropped as soon as the call completes; and the original request (and its body) should have been consumed by the client.

olix0r:
After looking at this a bit, it turns out we can completely avoid using `content-length` and instead use `Body::size_hint`, which will be set by the server when `content-length` was set and may also be set when it is not. PR on this branch here: https://github.com/linkerd/linkerd2-proxy/pull/1345

PR #1345, opened 2021-10-31
Simplify retry policy to use `Body::size_hint`

olix0r:
This change rewrites `RetryPolicy::can_retry` as
`RetryPolicy::body_too_big` and updates it to avoid reading
`content-length` header, instead relying on the `Body::size_hint` value
(which is set by the HTTP server if a `content-length` header is
present.

Renaming the function helps us express what the function actually checks
to help disambiguate it from all of the other 'retry' related
terminology in this file and is more explicit about what it does: we
don't precicesly know if the request is retryable, rather we know
whether the request is obviously too large to be retried. This also lets
us simplify comments and logging as the flow is more self-explanatory.

This change also adds unit tests for `RetryPolicy::body_too_big` to
exercise the variety of scenarios this function handles..

Furthermore, this change ensures that all `Body` implementations
correctly proxy `Body::size_hint`. `ReplayBody` is updated to cache the
the inner body's size hint value in the shared state so it can be reused
across all clones of the `ReplayBody` regardless of whether the current
clone holds the body state or not.

### DIFF
--- a/linkerd/app/core/src/retry.rs
+++ b/linkerd/app/core/src/retry.rs
@@ -4,6 +4,7 @@ use super::http_metrics::retries::Handle;
 use super::metrics::HttpRouteRetry;
 use crate::profiles;
 use futures::future;
+use http_body::SizeHint;
 use linkerd_error::Error;
 use linkerd_http_classify::{Classify, ClassifyEos, ClassifyResponse};
 use linkerd_http_retry::ReplayBody;
@@ -29,9 +30,6 @@ pub struct RetryPolicy {
     budget: Arc<retry::Budget>,
     response_classes: profiles::http::ResponseClasses,
 }
-
-/// Allow buffering requests up to 64 kb
-const MAX_BUFFERED_BYTES: usize = 64 * 1024;
 
 // === impl NewRetryPolicy ===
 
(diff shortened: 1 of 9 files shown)


linkerd/app/integration/src/tests/transparency.rs
```diff
@@ -843,9 +843,7 @@ macro_rules! http1_tests {
 
             let srv = server::http1()
                 .route_fn("/", |req| {
-                    let has_body_header = req.headers().contains_key("transfer-encoding")
-                        || req.headers().contains_key("content-length");
```
olix0r, line 847: I have no idea why this test--which only mentions `transfer-encoding` in its description also checks for `content-length`...  Now that we're properly preserving the size hint, we can set the `content-length` header to `0` if we know it's empty. This seems like a strictly good thing.

linkerd/http-retry/src/lib.rs
```diff
@@ -301,31 +311,29 @@ where
...
-            }
-
-            // Otherwise, the size is the size of the current buffer plus the
-            // size hint returned by the inner body.
-            let (rest_lower, rest_upper) = state
-                .rest
-                .as_ref()
-                .map(|rest| {
-                    let hint = rest.size_hint();
-                    (hint.lower(), hint.upper().unwrap_or(0))
```
olix0r, line 322: I think that `hint.upper().unwrap_or(0)` is a subtle bug -- if the lower bound is non-zero, we'll end up having an upper limit that is less than the lower limit, which is bad.

linkerd/app/core/src/retry.rs
```diff
@@ -183,9 +134,20 @@ where
         &self,
         req: http::Request<A>,
     ) -> Either<Self::RetryRequest, http::Request<A>> {
-        if self.can_retry(&req) {
-            return Either::A(req.map(|body| ReplayBody::new(body, MAX_BUFFERED_BYTES)));
-        }
-        Either::B(req)
+        let (head, body) = req.into_parts();
+        let replay_body = match ReplayBody::try_new(body, MAX_BUFFERED_BYTES) {
```
hawkw, line 138: this is really not important at all, but i _kind of_ feel like the `prepare_request` method should be responsible for checking the `size_hint` and `ReplayBody` should just have an unconditional `new` constructor...that kind of feels like a better separation of concerns to me, with the retry policy code being responsible for determining if the body is too large to retry, and `ReplayBody` just providing the buffering mechanism.

but, since `ReplayBody` _does_ still need to be aware of the body size limit so that it can stop buffering if necessary, maybe this is fine...either way, it's not a big deal...

linkerd/app/core/src/retry.rs
```diff
@@ -183,9 +134,20 @@ where
...
+                    size = body.size_hint().lower(),
+                    "Body is too large to buffer"
+                );
+                return Either::B(http::Request::from_parts(head, body));
+            }
+        };
+
+        // The body may still be too large to be buffered if the body's length was not known.
+        // `ReplayBody` handles this gracefully.
+        Either::A(http::Request::from_parts(head, replay_body))
```
hawkw, line 151: nit, take it or leave it: the early return here feels slightly convoluted, can't this just be
```suggestion
        match ReplayBody::try_new(body, MAX_BUFFERED_BYTES) {
            Ok(replay_body) => {
                // The body may still be too large to be buffered if the body's length was not known.
                // `ReplayBody` handles this gracefully.
                Either::A(http::Request::from_parts(head, replay_body))
            >,
            Err(body) => {
                tracing::debug!(
                    size = body.size_hint().lower(),
                    "Body is too large to buffer"
                );
                Either::B(http::Request::from_parts(head, body))
            }
        }
```

linkerd/http-retry/src/lib.rs
```diff
@@ -301,31 +311,29 @@ where
...
+            None => return SizeHint::with_exact(buffered),
+        };
+
+        // Otherwise, add the inner body's size hint to the amount of buffered data. An upper limit
+        // is only set if the inner body has an upper limit.
+        let mut hint = SizeHint::default();
+        hint.set_lower(buffered + rest_hint.lower());
+        if let Some(rest_upper) = rest_hint.upper() {
+            hint.set_upper(buffered + rest_upper);
         }
```
hawkw, line 336: hmm, if `rest` _doesn't_ have an upper bound, can't we just use the max buffered length as the upper bound, instead of not returning one?

hawkw (approved):
yup, this seems good to me --- i hadn't realized that `size_hint` would be set from the request's `content-length`; i had assumed it was *just* the size of any currently-received body chunks. this definitely seems like the right approach since it will also let us make better decisions in the case where the whole body has been received but it didn't have a `content-length`.

i had some minor nits and suggestions, but no hard blockers --- let me know what you think?

linkerd/app/core/src/retry.rs
olix0r: I initially had it the way you suggest, but I think I liked to make it impossible to create a `ReplayBody` that is 'illegal' -- this means that the redundant check in `clone_request` becomes totally unnecessary: by virtue of having constructed a `ReplayBody`, we know that it doesn't violate our expectations on body size.

linkerd/app/core/src/retry.rs
olix0r: yeah, I had it that way initially. but it feels like the secondary case is exceptional, so it feels appropriate to early return. don't feel strongly either way, though.

linkerd/http-retry/src/lib.rs
olix0r: But we don't know that it's actually the upper limit of the body, do we? My understanding is that `ReplayBody` should gracefully handle the case when the inner body exceeds the buffer capacity?

hawkw: ah, yeah, it's only the upper limit _if_ we're a subsequent clone of the body --- if we're the initial body, we will continue reading body chunks if we go past the limit; we'll just stop _buffering_ them. we probably _could_ add a check here and only return the max length if we didn't get an upper bound from `rest_hint` *and* we know we're not the original body...but, this is probably fine as-is and it may not be worth bothering with that...

PR #1346, opened 2021-11-01
retry: Simplify ReplayBody::poll_data for readability

olix0r:
`ReplayBody::poll_data` includes some deeply nested logic that can be
avoided with early returns. This restructures `ReplayBody::poll_data` so
that the control flow is easier to follow.

### DIFF
--- a/linkerd/http-retry/src/lib.rs
+++ b/linkerd/http-retry/src/lib.rs
@@ -199,13 +199,6 @@ where
             }
         }
 
-        // Get access to the initial body. If we don't have access to the inner body, there's no
-        // more work to do.
-        let rest = match state.rest.as_mut() {
-            Some(rest) => rest,
-            None => return Poll::Ready(None),
-        };
-
         // If the inner body has previously ended, don't poll it again.
         //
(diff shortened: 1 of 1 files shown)


linkerd/http-retry/src/lib.rs
```diff
@@ -180,79 +180,73 @@ where
...
-                                "buffered maximum capacity, discarding buffer"
-                            );
-                            state.buf = Default::default();
-                        }
-                        return Data::Initial(data.copy_to_bytes(length));
-                    }
-
-                    if state.is_capped() {
-                        return Data::Initial(data.copy_to_bytes(length));
-                    }
```
olix0r, line 245: AFAICT this is a redundant check? This is the same as the above conditional block?

olix0r, line 245: AFAICT the second block is redundant?

hawkw: oh, huh, yeah, that looks like a copy-paste error or something...

hawkw (approved):
:+1: this is definitely clearer than the `map` version, and doesn't seem to change any behavior as far as i can tell!

PR #1349, opened 2021-11-01
reorg: Decouple TLS implementation from proxy client

olix0r:
Currently the TLS implementation in `proxy-identity-default` depends on
`proxy-identity`, which depends on the proxy API to implement its
refreshing identity client.

This change updates these crates so that the TLS implementation only
depends on the core identity crate as follows:

* Move `proxy_identity::Credentials` to `identity::Credentials`;
* Move `proxy/identity/default` to `identity/default`;
* Rename `proxy/identity` to `proxy/identity-client` to help
  disambiguate it from the core identity crate.


hawkw (approved):
yup, this makes sense to me, ship it!

PR #1351, opened 2021-11-02
meshtls: Add a `boring` backend

olix0r:
This change adds a `meshtls-boring` proxy feature that can be used to
compile the proxy with an alternate TLS implementation. The
`meshtls-rustls` feature should be disabled to take advantage of this
alternate backend.
    
In its current mode, the boring backend is compatible with the existing
identity credentials and algorithms (specifically TLSv1.3 and
ECDSA-P256-SHA256 with CHACHA20-POLY1305-SHA256).
    
In future changes--once `boring` has been updated--we can:
- Improve error handling, especially for SSL errors
- Relax deny.toml changes needed by bindgen features
- Add a FIPS mode
    
Co-authored-by: @arnarpall


olix0r:
Related https://github.com/cloudflare/boring/pull/55

linkerd/meshtls/boring/src/creds.rs
```diff
@@ -0,0 +1,42 @@
...
+        // FIXME(ver) This should handle a list of PEM-encoded certificates.
+        let cert = X509::from_pem(roots_pem.as_bytes())?;
+        store.add_cert(cert)?;
+        store.build()
+    };
+
+    let key = PKey::private_key_from_pkcs8(key_pkcs8)?;
+
+    let (client_tx, client_rx) = {
+        let conn = ssl::SslConnector::builder(ssl::SslMethod::tls_client())?;
```
olix0r, line 31: This isn't appropriate -- this does not configure the client to use our roots/key.

linkerd/meshtls/boring/src/creds.rs
```diff
@@ -0,0 +1,42 @@
...
+        watch::channel(conn.build())
+    };
+    let (server_tx, server_rx) = {
+        let acc = ssl::SslAcceptor::mozilla_intermediate_v5(ssl::SslMethod::tls_server())?;
```
olix0r, line 35: Similarly

linkerd/io/src/boxed.rs
```diff
@@ -10,14 +10,14 @@ pub struct BoxedIo(Pin<Box<dyn Io + Unpin>>);
 /// This is necessary for `BoxedIo`, as `dyn AsyncRead + AsyncWrite + PeerAddr`
 /// is not a valid trait object. However, it needn't be public --- it's just
 /// used internally.
-trait Io: AsyncRead + AsyncWrite + PeerAddr + Send {}
+trait Io: AsyncRead + AsyncWrite + PeerAddr + Send + Sync {}
```
hawkw, line 13: what necessitated the `Sync` bound?

linkerd/meshtls/boring/src/client.rs
```diff
@@ -0,0 +1,163 @@
...
+        std::task::Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, io: I) -> Self::Future {
+        let conn = self.connector.clone();
+        let id = self.server_id.clone();
+        Box::pin(async move {
+            let config = conn
+                .configure()
+                .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
```
hawkw, line 88: is there a reason this has to be inside the future, rather than in `call` but outside of the future? if this wasn't inside the async block, we could presumably replace the `Box::pin(async move { ... })` with just returning a `futures::MapErr` and a `futures::Map` over the future returned by `tokio_boring::connect`.

not a blocker, just wondering.

linkerd/meshtls/boring/src/client.rs
```diff
@@ -0,0 +1,163 @@
...
+{
+    type Response = ClientIo<I>;
+    type Error = io::Error;
+    type Future = ConnectFuture<I>;
+
+    fn poll_ready(
+        &mut self,
+        _cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        std::task::Poll::Ready(Ok(()))
```
hawkw, line 79: nit, take it or leave it: maybe worth importing `Poll` and `Context` in this module?

linkerd/meshtls/boring/src/creds/store.rs
```diff
@@ -0,0 +1,135 @@
...
+            name,
+            client_tx,
+            server_tx,
+        }
+    }
+
+    fn cert_matches_name(&self, cert: &X509) -> bool {
+        for san in cert.subject_alt_names().into_iter().flatten() {
+            if let Some(n) = san.dnsname() {
+                if let Ok(name) = n.parse::<linkerd_dns_name::Name>() {
```
hawkw, line 43: presumably `boring` also guarantees that the values returned by `dnsname()` are DNS names...does this differ significantly from our own validation in `linkerd_dns_name`?

linkerd/meshtls/boring/src/creds/store.rs
```diff
@@ -0,0 +1,135 @@
...
+        _expiry: std::time::SystemTime,
+    ) -> Result<()> {
+        let cert = X509::from_der(&leaf)?;
+        if !self.cert_matches_name(&cert) {
+            return Err("certificate does not have a DNS name SAN for the local identity".into());
+        }
+
+        let mut chain = boring::stack::Stack::new()?;
+        chain.push(cert.clone())?;
+        for id::DerX509(der) in intermediates.iter() {
```
hawkw, line 91: nit, TIOLI: i think the `.iter()` shouldn't be necessary here?
```suggestion
        for id::DerX509(der) in &intermediates {
```

linkerd/meshtls/boring/src/creds/store.rs
```diff
@@ -0,0 +1,135 @@
...
+            return Err("certificate could not be validated against the trust chain".into());
+        }
+
+        let conn = {
+            // FIXME(ver) Restrict TLS version, algorithms, etc.
+            let mut b = ssl::SslConnector::builder(ssl::SslMethod::tls_client())?;
+            b.set_private_key(self.key.as_ref())?;
+            b.set_cert_store(self.clone_roots()?);
+            b.set_certificate(cert.as_ref())?;
+            for id::DerX509(der) in intermediates.iter() {
```
hawkw, line 107: nit/tioli: similarly, this could be
```suggestion
            for id::DerX509(der) in &intermediates {
```

linkerd/meshtls/boring/src/creds/store.rs
```diff
@@ -0,0 +1,135 @@
...
+                let cert = X509::from_der(der)?;
+                b.add_extra_chain_cert(cert)?;
+            }
+            b.set_verify(ssl::SslVerifyMode::PEER);
+            b.check_private_key()?;
+            b.build()
+        };
+
+        let _ = self.server_tx.send(acc);
+        let _ = self.client_tx.send(conn);
```
hawkw, line 131: should we error if there are no receivers?

linkerd/meshtls/boring/src/client.rs
```diff
@@ -0,0 +1,163 @@
...
+        let conn = self.connector.clone();
+        let id = self.server_id.clone();
+        Box::pin(async move {
+            let config = conn
+                .configure()
+                .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
+            let io = tokio_boring::connect(config, id.as_str(), io)
+                .await
+                .map_err(|e| match e.as_io_error() {
+                    Some(ioe) => io::Error::new(ioe.kind(), ioe.to_string()),
```
hawkw, line 92: wait...what's going on here? presumably, `as_io_error()` returns an `Option<io::Error>`; why do we create a new `io::Error` rather than just returning that error?

hawkw: ...oh, I see, `as_io_error()` returns a reference, and there's no way to get the inner `io::Error` out of the `HandshakeError`. Gross. That could be worth an upstream PR?

linkerd/meshtls/boring/src/client.rs
```diff
@@ -0,0 +1,163 @@
...
+                    None => io::Error::new(io::ErrorKind::Other, "unexpected TLS error"),
```
hawkw, line 93: this looks like it completely discards any error information if the error _isn't_ an IO error, which could be useful to have --- maybe this should wrap the `HandshakeError` as the inner error's source, so that we still print its message?
```suggestion
                    None => io::Error::new(io::ErrorKind::Other, e)
```

linkerd/meshtls/boring/src/server.rs
```diff
@@ -0,0 +1,154 @@
...
+    fn call(&mut self, io: I) -> Self::Future {
+        let acc = (*self.rx.borrow()).clone();
+        Box::pin(async move {
+            let io = tokio_boring::accept(&acc, io)
+                .await
+                .map(ServerIo)
+                .map_err(|e| match e.as_io_error() {
+                    Some(ioe) => io::Error::new(ioe.kind(), ioe.to_string()),
+                    None => io::Error::new(io::ErrorKind::Other, e),
+                })?;
```
hawkw, line 71: the same comments as on the client-side error handling apply here.

linkerd/meshtls/boring/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+name = "linkerd-meshtls-boring"
+version = "0.1.0"
+authors = ["Linkerd Developers <cncf-linkerd-dev@lists.cncf.io>"]
+license = "Apache-2.0"
+edition = "2018"
+publish = false
+
+[dependencies]
+boring = { version = "1" }
+boring-sys = { version = "1"}
```
hawkw, line 11: do we actually need to depend on the `sys` crate here? i assumed that would be a transitive dependency of `boring`, and i don't see anywhere where we actually use anything from the `sys` crate?

linkerd/meshtls/boring/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+futures = { version = "0.3", default-features = false }
+linkerd-error = { path = "../../error" }
+linkerd-dns-name = { path = "../../dns/name" }
+linkerd-identity = { path = "../../identity" }
+linkerd-io = { path = "../../io" }
+linkerd-stack = { path = "../../stack" }
+linkerd-tls = { path = "../../tls" }
+thiserror = "1"
```
hawkw, line 19: looks like we never actually use the `thiserror` dependency in this crate --- are we planning to?

hawkw (commented):
handful of nitpicks and stuff, but this looks good so far!

linkerd/meshtls/boring/src/client.rs
hawkw: ...oh, i see; `connect` is an `async fn`. never mind, then!

linkerd/meshtls/boring/src/creds/store.rs
olix0r: they should be the same, but is it really better to panic if there's a mismatch? it seems sufficient to fail to validate that the cert matches the name.

linkerd/meshtls/boring/src/client.rs
olix0r: boring's error implementation requires the `I: std::fmt::Debug`, which is a bit heavyweight.. I'll add a note about that.

linkerd/meshtls/boring/src/creds/store.rs
hawkw: yeah, i think this is fine, i was mostly curious about whether there was a possible difference.

linkerd/meshtls/boring/src/client.rs
hawkw: ohh, because it contains the IO stream in the error, presumably so that you can get the stream out if you want to use it without TLS. hmm... it would be _really_ nice if they had some way to separate the actual error from the returned IO stream.  but, this is fine, then, given the API constraints!

linkerd/meshtls/boring/src/client.rs
arnarpall: Yeah this is definitely not ideal :(
Most definitely would warrant and upstream PR with some changes to the error handling for handshake errors.

linkerd/io/src/boxed.rs
olix0r: We need this so that boring's error implements `Error`
```
error[E0277]: `I` cannot be shared between threads safely
   --> linkerd/meshtls/boring/src/server.rs:68:66
    |
68  |                     None => io::Error::new(io::ErrorKind::Other, e),
    |                                                                  ^ `I` cannot be shared between threads safely
    |
    = note: required because it appears within the type `tokio_boring::StreamWrapper<I>`
    = note: required because it appears within the type `PhantomData<tokio_boring::StreamWrapper<I>>`
    = note: required because it appears within the type `boring::ssl::SslStream<tokio_boring::StreamWrapper<I>>`
    = note: required because it appears within the type `MidHandshakeSslStream<tokio_boring::StreamWrapper<I>>`
    = note: required because it appears within the type `boring::ssl::HandshakeError<tokio_boring::StreamWrapper<I>>`
    = note: required because it appears within the type `tokio_boring::HandshakeError<I>`
    = note: required because of the requirements on the impl of `From<tokio_boring::HandshakeError<I>>` for `Box<(dyn std::error::Error + Send + Sync + 'static)>`
    = note: required because of the requirements on the impl of `Into<Box<(dyn std::error::Error + Send + Sync + 'static)>>` for `tokio_boring::HandshakeError<I>`
note: required by a bound in `linkerd_io::Error::new`
help: consider further restricting type parameter `I`
    |
49  |     I: std::fmt::Debug, I: std::marker::Sync // Needed so that boring's error type implements Error.
    |                       ~~~~~~~~~~~~~~~~~~~~~~
```

linkerd/meshtls/boring/src/client.rs
hawkw: i went ahead and opened https://github.com/cloudflare/boring/pull/57 upstream, which should hopefully help with this!

linkerd/meshtls/boring/src/client.rs
hawkw: i went ahead and opened https://github.com/cloudflare/boring/pull/57 upstream, which should hopefully help with this as well.

olix0r: We'll at least be able to use `to_string()` here https://github.com/cloudflare/boring/commit/c4e1f474973744b32579797c1ff51335002a4491

linkerd/meshtls/boring/src/client.rs
olix0r: I think stringifying the underlying error here is fine for now. I looked at submitting a PR and it's a little more complicated than I expected. If someone else wants to submit an upstream change, we'll use it; but I'm not sure it's worth the effort.

olix0r:
@hawkw @arnarpall This change should be mergeable now (once CI clears). We run our normal end-to-end tests against the boring module and I've tested this branch manually as well.

This change does **not** include the FIPS functionality being discussed in https://github.com/cloudflare/boring/pull/52/files; and it sounds like even once that change merges (if it lands in its current state), it will be difficult to build in that mode (as a FIPS-compliant boring library will have to be provided externally).

I haven't taken advantage of any of the recently changed error functionality, either. We'll pick that up once boring does a proper crate release.

arnarpall:
> @hawkw @arnarpall This change should be mergeable now (once CI clears). We run our normal end-to-end tests against the boring module and I've tested this branch manually as well.
> 
> This change does **not** include the FIPS functionality being discussed in https://github.com/cloudflare/boring/pull/52; and it sounds like even once that change merges (if it lands in its current state), it will be difficult to build in that mode (as a FIPS-compliant boring library will have to be provided externally).
> 
> I haven't taken advantage of any of the recently changed error functionality either. We'll pick that up once boring does a proper crate release.

Truly awesome work.

I believe I could rework my original boring pull request once this one is merged where we hook into the boring build process to enable fips mode without having to to provide it externally. I'd like to get to a point where this is as transparent as possible

linkerd/meshtls/boring/src/creds.rs
```diff
@@ -0,0 +1,210 @@
...
+
+        Ok(conn.build())
+    }
+
+    // TODO(ver) Specify certificate types, signing algorithms, cipher suites..
+    pub(crate) fn connector(&self, alpn_protocols: &[Vec<u8>]) -> Result<ssl::SslConnector> {
+        // XXX(ver) This function reads from the environment and/or the filesystem. This likely is
+        // at best wasteful and at worst unsafe (if another thread were to mutate these environment
+        // variables simultaneously, for instance). Unfortunately, the boring APIs don't really give
+        // us an alternative AFAICT.
```
hawkw, line 112: do you know if this is something that can be fixed upstream in the `boring` crate, or is it a boringssl behavior that we can't avoid?

linkerd/meshtls/boring/src/creds.rs
```diff
@@ -0,0 +1,210 @@
...
+        let mut conn = ssl::SslConnector::builder(ssl::SslMethod::tls_client())?;
+
+        // Explicitly enable use of TLSv1.3
+        conn.set_options(ssl::SslOptions::NO_TLSV1 | ssl::SslOptions::NO_TLSV1_1);
+        // XXX(ver) if we disable use of TLSv1.2, connections just hang.
+        //conn.set_options(ssl::SslOptions::NO_TLSV1_2);
```
hawkw, line 118: should we report a bug for this upstream?

linkerd/meshtls/boring/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+//! In its current form, this crate is compatible with the `meshtls-rustls`
+//! implementation, which requires of ECDSA-P256-SHA256 keys & signature
+//! algorithms. This crate doesn't actually constrain the algorithms beyond the
+//! Mozilla's 'intermediate' (v5) [defaults][defaults]. But, the goal for
+//! supporting `boring` is to provide a FIPS 140-2 compliant mode. There's a
+//! [PR][fips-pr] that implements this, but code changes will likely be required
+//! to enable this once it's merged/released.
+//!
+//! A new SSL context is created for each connection. This is probably
+//! unnecessary, but it's simpler for now. We can revisit this if needed.
```
hawkw, line 18: not a blocker at all, just curious: how much do we know about what this actually *means*? e.g. do we know what a "SSL context" in `boring` consists of, and what are the implications of creating them?

linkerd/meshtls/boring/src/server.rs
```diff
@@ -0,0 +1,177 @@
...
+            rx,
+            alpn: None,
+        }
+    }
+
+    pub fn with_alpn(mut self, alpn_protocols: Vec<Vec<u8>>) -> Self {
+        if !alpn_protocols.is_empty() {
+            self.alpn = Some(alpn_protocols.into());
+        }
+
```
hawkw, line 37: nit, take it or leave it: should this maybe do...something...if `alpn_protocols` is _not_ empty? maybe a warning or debug assertion? it seems like we expect this shouldn't happen, but it would be nice to actually know if it ever happens...

linkerd/meshtls/boring/src/server.rs
```diff
@@ -0,0 +1,177 @@
...
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> io::Poll<()> {
+        io::Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, io: I) -> Self::Future {
+        // TODO(ver) we should avoid creating a new context for each connection.
+        let acceptor = match &self.alpn {
+            Some(alpn) => self.rx.borrow().acceptor(alpn),
+            None => self.rx.borrow().acceptor(&[]),
+        };
```
hawkw, line 66: nit/TIOLI: i think this could be
```suggestion
        let acceptor = self.rx.borrow().acceptor(self.alpn.as_deref().unwrap_or(&[]);
```

linkerd/meshtls/src/server.rs
```diff
@@ -74,14 +93,18 @@ impl Server {
...
 {
     type Response = (ServerTls, ServerIo<I>);
     type Error = io::Error;
     type Future = TerminateFuture<I>;
 
+    #[inline]
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
         match self {
+            #[cfg(feature = "boring")]
+            Self::Boring(svc) => <boring::Server as Service<I>>::poll_ready(svc, cx),
```
hawkw, line 106: looking at this code it occurs to me that it's _probably_ possible to have a macro that, given a list of feature names and type names, generates all of these match arms automagically...but this is probably not worth bothering with unless we end up supporting a larger number of TLS implementations...

hawkw (approved):
overall, this looks good to me. i had a couple questions and minor nits but nothing stood out as a blocker!

Dockerfile
```diff
@@ -61,8 +67,9 @@ ARG SKIP_IDENTITY_WRAPPER
 WORKDIR /linkerd
 COPY --from=build /out/linkerd2-proxy /usr/lib/linkerd/linkerd2-proxy
 ENV LINKERD2_PROXY_LOG=warn,linkerd=info
-RUN if [ -n "$SKIP_IDENTITY_WRAPPER" ] ; then \
-  rm -f /usr/bin/linkerd2-proxy-run && \
-  ln /usr/lib/linkerd/linkerd2-proxy /usr/bin/linkerd2-proxy-run ; \
+RUN \
+  if [ -n "$SKIP_IDENTITY_WRAPPER" ] ; then \
+    rm -f /usr/bin/linkerd2-proxy-run && \
+    ln /usr/lib/linkerd/linkerd2-proxy /usr/bin/linkerd2-proxy-run ; \
```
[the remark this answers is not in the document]
olix0r: yep

linkerd/meshtls/boring/src/creds.rs
olix0r: maybe? this is probably a bug in boringssl and not the rust bindings, if it's a bug.

deny.toml
```diff
@@ -47,8 +47,15 @@ highlight = "all"
 deny = [
     { name = "rustls", wrappers = ["tokio-rustls"] }
 ]
-skip = []
-skip-tree = []
+skip = [
+    # boring-sys pulls in an old version via bindgen. See
+    # https://github.com/cloudflare/boring/pull/55.
+    { name = "ansi_term" },
```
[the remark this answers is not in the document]
olix0r: this would also ignore all _other_ bindgen deps, though. I think the `skip` is narrower.

linkerd/meshtls/boring/src/creds.rs
olix0r: boring could probably have other constructors that don't load default trust anchors. I don't think it actually matters for us, in practice, though.

linkerd/meshtls/boring/src/lib.rs
olix0r: Some allocations + everything in the connector/acceptor constructors (i.e. trust anchors, private key, certs).

linkerd/meshtls/boring/src/creds.rs
hawkw: that's fair; it seems like something that could be worth opening an issue for, but maybe not worth putting effort into implementing?

PR #1353, opened 2021-11-03
Introduce `meshtls` facade to hide rustls crate

olix0r:
In #1351, we add an alternate identity/mtls implementation that uses
`boring`. To setup for that, this change introduces a new `meshtls`
crate that serves as a facade for application crates to depend on,
independently of the actual crypto implementation.

This change does not change any runtime logic and sets up for #1351 to
enable an alternate TLS implementation as a build-time configuration.


linkerd/meshtls/src/client.rs
```diff
@@ -0,0 +1,200 @@
...
+impl NewService<ClientTls> for NewClient {
+    type Service = Connect;
+
+    fn new_service(&self, target: ClientTls) -> Self::Service {
+        #[cfg(feature = "rustls")]
+        if let Self::Rustls(new_client) = self {
+            return Connect::Rustls(new_client.new_service(target));
+        }
+
+        unreachable!()
```
hawkw, line 49: nit/TIOLI: not a big deal, but maybe this should say something like
```suggestion
        unreachable!("attempted o initiate a mTLS connection, but no TLS backends were enabled at compile time!")
```
or something, _just in case_ someone makes a build of the proxy where no backends were enabled?

hawkw: alternatively, we could probably make this unreachable at compile-time by putting `#[cfg(any(feature = "rustls", feature = "boring", ...etc))]` on the `Service` impl (and other stuff in this crate), but i'm not sure if that would make other stuff more complicated...

linkerd/meshtls/src/client.rs
```diff
@@ -0,0 +1,200 @@
...
+        if let ClientIoProj::Rustls(io) = this {
+            return io.poll_write_vectored(cx, bufs);
+        }
+
+        unreachable!()
+    }
+
+    #[inline]
+    fn is_write_vectored(&self) -> bool {
+        unimplemented!()
```
hawkw, line 179: the rustls impl should probably implement this?
```suggestion
        #[cfg(feature = "rustls")]
        if let ClientIo::Rustls(ref io) = self {
            return io.is_write_vectored();
        }
        
        unreachable!()
```

linkerd/meshtls/src/client.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+}
+
+impl<I> HasNegotiatedProtocol for ClientIo<I> {
+    #[inline]
+    fn negotiated_protocol(&self) -> Option<NegotiatedProtocolRef<'_>> {
+        unimplemented!()
```
hawkw, line 186: similarly, this needs to be filled in before we merge, right?

linkerd/meshtls/src/lib.rs
```diff
@@ -0,0 +1,51 @@
+#![allow(irrefutable_let_patterns)]
```
hawkw, line 1: nit: we should make sure to remove this when we do add additional variants...

hawkw (approved):
overall, this seems like the best possible way i could think of to structure the facade crate; it feels a _big_ ugly, but i feel like abstracting over the different implementations with feature flags is _always_ going to be ugly, so this looks like the neatest way of doing it.

i think it *could* be worth having the various `unreachable!()` panics include messages, just in case someone compiles a build where all the TLS implementations are disabled? but, that's not really a blocker.

also, i noticed a couple places where it looks like we forgot to forward an implementation through a wrapper; we should probably fix that before merging this?

other than that, lgtm!

linkerd/meshtls/src/lib.rs
olix0r: No, we'll continue to want this so that things compile when there's only one variant enabled.

hawkw: ah, right, i didn't think that through...

linkerd/meshtls/src/client.rs
olix0r: I changed this to use a match for all of these so there's no longer an unreachable case. I've added a build.rs check that ensures at least one feature is enabled.

linkerd/meshtls/build.rs
```diff
@@ -0,0 +1,16 @@
...
+        .any(|f| std::env::var_os(&*format!("CARGO_FEATURE_{}", f.to_ascii_uppercase())).is_some())
+    {
+        return Err(format!(
+            "at least one of the following TLS implementations must be enabled: '{}'",
+            TLS_FEATURES.join("', '"),
+        )
+        .into());
+    }
+
+    Ok(())
```
hawkw, line 15: FWIW, I think this could be achieved without a build script by just sticking
```rust
#![cfg(not(any(feature = "rustls"))]
compile_error!("at least one of the following TLS implementations must be enabled: 'rustls}''")
```
in `lib.rs` (and adding the other implementation feature flags as needed)

PR #1354, opened 2021-11-03
build: add nix-env for building the proxy

hawkw:
This branch adds a simple nix-env configuration for building the proxy
locally on NixOS. This does *not* yet introduce a derivation for
*packaging* the proxy as a Nix package, just a nix-env for local
development for Nix users.

Actually packaging the proxy for Nix could be
fun, but it's not really necessary, since the proxy is not currently
distributed as a package for other package managers --- it's distributed
as a docker image. If we were going to actually distribute something as a
Nix package, it would be the Linkerd CLI (which could be worth doing!).

This branch *does* include the necessary configuration to build
`cloudflare/boring`, which is added as a dependency by @olix0r's
PR #1351. I've confirmed that it is possible to build that branch with
these configs, but I opened this as a separate PR against `main` so that
we can merge it separately. Currently, we do have to build `boringssl`
from source, rather than depending on it from nixpkgs, which is kind of
a bummer, but this can be fixed later.


olix0r:
Closing this in favor of #1432

hawkw:
fair enough!

PR #1356, opened 2021-11-05
ci: Split actions into several workflows

olix0r:
This change splits our CI workflows into smaller groups so that when
we need to retry a workflow (i.e. due to CI flakiness), it has a smaller
blast radius. It also splits out integration tests from other (unit) tests
which should hopefully get us quicker feedback on crate-level tests.


olix0r:
I'll update the branch requirements after this merges.

.github/workflows/check.yml
```diff
@@ -44,41 +31,43 @@ jobs:
...
+    timeout-minutes: 20
     runs-on: ubuntu-latest
-    permissions:
-      contents: read
     steps:
       - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
-      - run: cargo doc
+      - env:
+          CARGO_RELEASE: true
+        run: make build
```
hawkw, line 73: since the release build could potentially take way longer than the other steps, should this go in a separate workflow?

.github/workflows/test.yml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+
+  # Run all non-integration tests.
+  unit:
+    timeout-minutes: 15
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
+      - run: cargo test --all --exclude=linkerd-app-integration --no-run
+      - run: cargo test --all --exclude=linkerd-app-integration
```
hawkw, line 24: is the idea behind using `--no-run` first that there's a separate step that builds the tests, so we can differentiate between build failures and test failures?

.github/workflows/coverage.yml
```diff
@@ -5,6 +5,9 @@ on:
   schedule:
     - cron: '0 0 * * 0'
```
hawkw, line 7: should this set the same cargo env vars we set for other Rust build jobs?

hawkw (approved):
looks good to me, this definitely makes sense (and being able to rerun fewer jobs seems like a good idea!)

i commented on a couple small notes/questions, but no blockers!

.github/workflows/test.yml
olix0r: yep, recommended in matklad's post. it also helps us be clear about where we're spending time (in ci logs)

PR #1359, opened 2021-11-05
meshtls: allow building without any TLS impls enabled

hawkw:
Currently, the `linkerd-meshtls` crate cannot be compiled when no TLS
implementation feature flags are enabled. This is an issue, since it
means the `rustls` implementation must be enabled by default, requiring
all crates that depend on `meshtls`, or on a crate that depends on it,
to expose feature flags for controlling what TLS implementation(s) are
enabled. This is not ideal.

This branch changes the `meshtls` so that it can compile even when no
TLS implementations are enabled. The crates which actually need TLS
implementations to be enabled (the `linkerd2-proxy` application crate,
and `linkerd-app-integration`) now depend on the `linkerd-meshtls` crate
for the purposes of enabling feature flags. Other crates that depend on
`meshtls` can do so without having to propagate its feature flags.

The compile error when no TLS implementations are enabled is moved to
the `linkerd2-proxy` crate. `linkerd-meshtls` will now panic at runtime
(rather than failing to compile) if built with all TLS features disabled
(so that dependent crates don't have to enable its features), but the
proxy crate statically ensures that this cannot happen when building a
proxy.

### DIFF
--- a/linkerd/meshtls/Cargo.toml
+++ b/linkerd/meshtls/Cargo.toml
@@ -7,7 +7,9 @@ edition = "2018"
 publish = false
 
 [features]
-rustls = ["linkerd-meshtls-rustls"]
+rustls = ["linkerd-meshtls-rustls", "has_any_tls_impls"]
+# Enabled if *any* TLS impl is enabled.
+has_any_tls_impls = []
 
 [dependencies]
 futures = { version = "0.3", default-features = false }
--- a/linkerd/meshtls/Cargo.toml
+++ b/linkerd/meshtls/Cargo.toml
@@ -7,9 +7,7 @@ edition = "2018"
 publish = false
 
 [features]
-default = ["rustls"]
 rustls = ["linkerd-meshtls-rustls"]
-
 
 [dependencies]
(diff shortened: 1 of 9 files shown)


linkerd/meshtls/Cargo.toml
```diff
@@ -7,9 +7,9 @@ edition = "2018"
 publish = false
 
 [features]
-default = ["rustls"]
-rustls = ["linkerd-meshtls-rustls"]
-
+rustls = ["linkerd-meshtls-rustls", "has_any_tls_impls"]
+# Enabled if *any* TLS impl is enabled.
+has_any_tls_impls = []
```
olix0r, line 12: tioli: wdyt about prefixing this like `__has_any_tls_impls` to show it's internal?

olix0r (approved):

linkerd/meshtls/Cargo.toml
hawkw: i actually considered that but wasn't sure if it was necessary. i'll change it!

PR #1362, opened 2021-11-06
meshtls-rustls: update to `rustls` 0.20 and `tokio-rustls` 0.23

hawkw:
This branch updates `linkerd-meshtls-rustls` to use `rustls` version
0.20, `tokio-rustls` 0.23, and `webpki` 0.22.

Hopefully, this change should be relatively straightforward --- I've
just updated the proxy's usage of these crates to use their new APIs.
Perhaps the most significant change in `rustls` is the new builder APIs
for `ClientConfig` and `ServerConfig`. Besides that, there were a couple
places where we had to restructure our code a bit to use the new API ---
we can no longer access the server cert verifier from the
`ClientConfig`, which we previously used in `Store::validate` for
validating the proxy's cert. Now, instead of accessing it _from_ the
`ClientConfig`, we pass in an `Arc`ed server cert verifier into the
client config, and hold onto a clone of it for use in `Store::validate`.
This is effectively the same as what we were doing before, but required
some restructuring.

Also, since `rustls` no longer exposes `webpki` types in its public API,
this change reduces our `webpki` usage significantly.


linkerd/dns/name/src/name.rs
```diff
@@ -165,7 +165,7 @@ fn is_valid_reference_dns_id(hostname: untrusted::Input<'_>) -> bool {
 // https://tools.ietf.org/html/rfc5280#section-4.2.1.6:
 //
 //   When the subjectAltName extension contains a domain name system
-//   label, the domain name MUST be stored in the dNSName (an IA5String).
+//   label, the domain name MUST be stored in the DnsName (an IA5String).
```
hawkw, line 168:
```suggestion
//   label, the domain name MUST be stored in the DNSName (an IA5String).
```
overeager find and replace for the `webpki` type, lol...

linkerd/meshtls/rustls/src/client.rs
```diff
@@ -93,7 +92,8 @@ where
 
     fn call(&mut self, io: I) -> Self::Future {
         tokio_rustls::TlsConnector::from(self.config.clone())
-            .connect(self.server_id.as_ref(), io)
+            // XXX(eliza): it's a bummer that the server name has to be cloned here...
+            .connect(self.server_id.clone(), io)
```
hawkw, line 96: this is not my favorite thing about the new `rustls` API, but i don't think we can fix it in the proxy...

hawkw:
`cargo deny` is failing because i forked `webpki` under my personal github account rather than under the linkerd org way back when i was initially experimenting with the webpki update...i think the reason for that was maybe because i wanted to patch the dep on the `linkerd/webpki` repo, or something?

hawkw:
> `cargo deny` is failing because i forked `webpki` under my personal github account rather than under the linkerd org way back when i was initially experimenting with the webpki update...i think the reason for that was maybe because i wanted to patch the dep on the `linkerd/webpki` repo, or something?

...oh, I figured out why this is in my fork rather than linkerd's fork --- it's obvious in hindsight. turns out i don't have push permission for `linkerd/webpki`. @olix0r, mind adding me to the contributors for that repo so I can push the 0.22 branch?

olix0r:
@hawkw you should have access to linkerd/webpki now.

olix0r (approved):

linkerd/meshtls/rustls/Cargo.toml
```diff
@@ -18,11 +18,10 @@ linkerd-stack = { path = "../../stack" }
 linkerd-tls = { path = "../../tls" }
 linkerd-tls-test-util = { path = "../../tls/test-util", optional = true }
 ring = { version = "0.16.19", features = ["std"] }
+rustls-pemfile = "0.2"
 thiserror = "1"
 tokio = { version = "1", features = ["macros", "rt", "sync"] }
-tokio-rustls = "0.23.1"
-rustls = { version = "0.20", features = ["dangerous_configuration"] }
-rustls-pemfile = "0.2"
+tokio-rustls = { version = "0.23.1", features = ["dangerous_configuration"] }
```
hawkw, line 24: ah, i hadn't realized tokio-rustls also exposed that!

PR #1365, opened 2021-11-08
ci: Split jobs into 'fast' and 'slow' workflows

olix0r:
This change sets up CI to add support for the boring meshtls backend. We
put most of the fast jobs into a single workflow and move slower jobs
into another workflow.

Unit tests are updated to exclude all app-level crates, which take a
while to compile; and the integration test job now runs all of these app
tests.


.github/workflows/fast.yml
```diff
@@ -0,0 +1,136 @@
...
+  RUSTUP_MAX_RETRIES: 10
+
+permissions:
+  contents: read
+
+jobs:
+
+  # Linting
+  check-clippy:
+    timeout-minutes: 10
```
hawkw, line 20: do you know if these configs can be set for all jobs in this file? could be nice to do now that we split them based on fast/slow...?

hawkw (approved):
lgtm!

.github/workflows/fast.yml
olix0r: afaict, no https://json.schemastore.org/github-workflow.json

PR #1371, opened 2021-11-11
Eliminate needless compiler directives

olix0r:
We've accrued various compiler directives and clippy exceptions that are
no longer necesary. This change removes these unnecessary settings.

This change also addresses a lint in our integration tests.


hawkw (approved):
lgtm!

PR #1372, opened 2021-11-12
transport: Improve TCP server error messages

olix0r:
The TCP server can fail connections in a few situations:

* `accept(2)` can fail
* setting thte TCP keepalive can fail
* getting the client peer address can fail

It's not easy to differentiate these errors when inspecting proxy logs.

This change adds wrapper error types with custom error messages that
disambiguate the error cause.

Additionally, this change includes the client address when logging
information/warnings about connection errors in general. This is
redundant when logging is set at debug, but will be helpful when logging
is set at the normal level.

### DIFF
--- a/linkerd/proxy/transport/src/listen.rs
+++ b/linkerd/proxy/transport/src/listen.rs
@@ -3,6 +3,7 @@ use futures::prelude::*;
 use linkerd_io as io;
 use linkerd_stack::Param;
 use std::{fmt, pin::Pin};
+use thiserror::Error;
 use tokio::net::TcpStream;
 use tokio_stream::wrappers::TcpListenerStream;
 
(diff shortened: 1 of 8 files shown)


linkerd/proxy/transport/src/listen.rs
```diff
@@ -64,10 +77,14 @@ where
...
-            let tcp = res?;
+            let tcp = res.map_err(|e| io::Error::new(e.kind(), AcceptError(e)))?;
             super::set_nodelay_or_warn(&tcp);
-            let tcp = super::set_keepalive_or_warn(tcp, keepalive)?;
-            let client = Remote(ClientAddr(tcp.peer_addr()?));
+            let tcp = super::set_keepalive_or_warn(tcp, keepalive)
+                .map_err(|e| io::Error::new(e.kind(), KeepaliveError(e)))?;
+            let client = Remote(ClientAddr(
+                tcp.peer_addr()
+                    .map_err(|e| io::Error::new(e.kind(), PeerAddrError(e)))?,
```
hawkw, line 86: it seems a little weird that we wrap an `io::Error` in our various error types, and then wrap that error type in another `io::Error`? is there any reason that the `Bind` trait _needs_ to return an `io::Error` rather than a `linkerd_error::Error`? if we just changed the trait to return a `linkerd_error::Error`, we wouldn't need to wrap our errors in another layer of `io::Error`...

hawkw (approved):
looks good to me overall, one thought that we could maybe clean up a bit, but it would be a bigger change.

linkerd/proxy/transport/src/listen.rs
olix0r: There are assumptions in various places that the listeners/connectors return `io::Error` and I don't want to have to tackle all of that right now. I guess it would probably be fine to omit the `#[source]` annotation on these error types though.

olix0r: nevermind. fixed in https://github.com/linkerd/linkerd2-proxy/pull/1372/commits/f4b9d6f336e2990d778793ca5781cc027e99d663

PR #1378, opened 2021-11-16
app: Move retry types into the app-outbound crate

olix0r:
The retry crate is hardcoded to use the `linkerd_app_core::dst::Route`
type as its target. This is inflexible, especially if we want to
reorganize the outbound stack to support things like header-based
overrides.

This change movers the retry layer into the outbound crate so that it
can be changed to use outbound-specific types.

No functional changes.


hawkw (approved):
lgtm!

PR #1380, opened 2021-11-17
outbound: Use per-route services in routing stack

olix0r:
The outbound proxy currently maintains a `Proxy` instance for each
profile-defined `Route`. This allows the router to use a single
underlying logical service, but this prevents per-route logic from
influencing the logical target to which the request is dispatched.

This change modifies the outbound stack to use an alternate profile
router that dispatches over per-route `Service`s instead of per-route
`Proxy`s. This is possible because the outbound logical stack is
buffered (and implements `Clone`), while the inbound stack is not.

To support this, the following changes have been made:

* The `linkerd_app_core::dst` module has been eliminated in favor of
  inbound- and outbound-specific Route target types;
* `linkerd_http_classify::Classify` now implements `Service` in addition
  to `Proxy`;
* `linkerd_retry::Retry` now implements `Service` instead of `Proxy`;
* `linkerd_service_profiles`now implements two caching routers:
  `NewServiceRouter` for `Service`s and `NewProxyRouter` for `Proxy`s;
* The `linkerd_stack::ProxyService` helper has been removed, as it's not
  used; and
* Various unneeded `http::BoxResponse` layers have been removed from the
  outound stack.

This change does not introduce any new buffers and should not change
runtime behavior. This change sets up future changes to the outbound
stack.


linkerd/app/outbound/src/http/logical.rs
```diff
@@ -157,17 +165,16 @@ impl<E> Outbound<E> {
                         // Sets the per-route response classifier as a request
                         // extension.
                         .push(classify::NewClassify::layer())
-                        .push_map_target(Logical::mk_route)
                         .push_on_service(http::BoxResponse::layer())
+                        // TODO(ver): CloneBoxService to flatten the stack type?
```
hawkw, line 169: this is released now fwiw: https://docs.rs/tower/0.4.11/tower/util/struct.BoxCloneService.html

linkerd/http-classify/src/service.rs
```diff
@@ -48,3 +49,24 @@ where
...
+    type Future = S::Future;
+
+    #[inline]
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, mut req: http::Request<B>) -> Self::Future {
+        let classify_rsp = self.classify.classify(&req);
+        let _ = req.extensions_mut().insert(classify_rsp);
```
hawkw, line 69: nit: should this `debug_assert!` that there wasn't already a response classifier inserted? not a big deal.

linkerd/retry/src/lib.rs
```diff
@@ -94,52 +78,38 @@ where
...
-            return match policy.prepare_request(req) {
-                Either::A(retry_req) => {
-                    let inner =
-                        Proxy::<R::RetryRequest, S>::wrap_service(self.inner.clone(), svc.clone());
-                    let retry = tower::retry::Retry::new(policy.clone(), inner);
-                    ResponseFuture::Retry(retry.oneshot(retry_req))
-                }
-                Either::B(req) => ResponseFuture::Disabled(self.inner.proxy(svc, req)),
-            };
-        }
```
hawkw, line 124: new version of this feels much easier to read! <3

linkerd/service-profiles/src/http/proxy.rs
```diff
@@ -0,0 +1,136 @@
...
+            }
+        }
+
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, req: http::Request<B>) -> Self::Future {
+        match super::route_for_request(&self.http_routes, &req) {
+            None => future::Either::Left({
+                // Otherwise, use the inner service directly.
```
hawkw, line 120: _very_ silly nitpick: it's mildly weird to me that the comment starting with "otherwise" comes _before_ the other comment? not a blocker of course.

linkerd/service-profiles/src/http/proxy.rs
```diff
@@ -0,0 +1,136 @@
...
+        futures::ready!(self.inner.poll_ready(cx).map_err(Into::into))?;
+
+        // If the routes have been updated, update the cache.
+        if let Poll::Ready(Some(Profile { http_routes, .. })) = self.rx.poll_next_unpin(cx) {
+            debug!(routes = %http_routes.len(), "Updating HTTP routes");
+            let routes = http_routes
+                .iter()
+                .map(|(_, r)| r.clone())
+                .collect::<HashSet<_>>();
+            self.http_routes = http_routes;
```
hawkw, line 100: nit, take it or leave it: this seems like it could be
```suggestion
            self.http_routes.clear();
            self.http_routes.extend(routes.iter().map(|(_, r)| r.clone()));
```
which has the nice side benefit of letting us reuse the prior allocation for the hashmap if the new `routes` is the same size or smaller...not a blocker though.

linkerd/service-profiles/src/http/service.rs
```diff
@@ -0,0 +1,116 @@
...
+            self.services.retain(|r, _| routes.contains(r));
+            for route in routes.into_iter() {
+                if let hash_map::Entry::Vacant(ent) = self.services.entry(route) {
+                    let route = ent.key().clone();
+                    let svc = self
+                        .new_route
+                        .new_service((Some(route), self.target.clone()));
+                    ent.insert(svc);
+                }
+            }
```
hawkw, line 94: nit/TIOLI: it seems like we _could_ probably stick  this part in some common `Routes` type or something so that it's shared by `ProxyRouter` and `ServiceRouter`? it's only repeated in two places, but there's enough going on here that _maybe_ we want to define it in one place?

hawkw (approved):
sorry it took a bit to read over this, i wanted to make sure we didn't accidentally change something subtle...looks good to me!

i had a few nitpicks but no blocking concerns.

linkerd/service-profiles/src/http/proxy.rs
olix0r: I don't follow?

`self.http_routes = http_routes;` shouldn't allocate.

The issue is that we temporarily allocate a `HashSet` so that we can do quick lookups in the following `self.proxies.retain(...)`. I think we basically need to have some sort of temporary allocation, since we can't iterate over `self.proxies` while modifying it.

linkerd/service-profiles/src/http/service.rs
olix0r: I couldn't work out a way to make this better (in the few minutes I tried). If it's doable, easy to fix later.

linkerd/service-profiles/src/http/proxy.rs
hawkw: hahahaha, i can't read --- i mixed up `http_routes` and `routes`. my bad.

PR #1382, opened 2021-11-18
ci: Fix coverage workflow

olix0r:
* Avoid building the binary target, since it slows down the build
* Use --skip-clean to use the cached build
* Fetch a pre-build tarpaulin binary (saves ~5m)
* Mark flakey tests as flakey

### DIFF
--- a/linkerd/app/integration/src/tests/telemetry.rs
+++ b/linkerd/app/integration/src/tests/telemetry.rs
@@ -493,7 +493,7 @@ where
 // Eventually, we can add some kind of mock timer system for simulating latency
 // more reliably, and re-enable this test.
 #[tokio::test]
-#[cfg_attr(not(feature = "flaky_tests"), ignore)]
+#[cfg_attr(not(feature = "flakey-in-ci"), ignore)]
 async fn inbound_response_latency() {
     test_response_latency(Fixture::inbound_with_server).await
 }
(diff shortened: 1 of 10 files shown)


kleimkuhler (approved):

linkerd/app/integration/src/tests/telemetry.rs
```diff
@@ -1191,6 +1191,8 @@ mod transport {
         test_tcp_connect(TcpFixture::outbound()).await;
     }
 
+    // XXX This test is flakey when running coverage tests.
+    #[cfg_attr(not(feature = "flaky_tests"), ignore)]
```
hawkw, line 1195: i believe the `flakey_tests` feature is disabled when running the normal CI test job, as well, so these tests will now never run on CI. should we have a separate cfg to disable them only under coverage?

olix0r: I'm pretty sure these tests can actually be flakey in ci too :/ I don't have a good catalog of flakey test failures, but they show up in normal CI runs from time to time.  So, I'd really _love_ to figure out how to test this stuff better in general...

I guess we could add a `very_flaky_test` feature for now, though.

hawkw (approved):

olix0r:
@hawkw please take another look -- we add a `flakey-in-coverage` feature which includes these tests. I've updated integration tests to run all tests by default; and in doing so I've found a few tests that never pass for me (now marked as `#[ignore]`)

hawkw (approved):
okay, lgtm!

 the fact that so many of the "flaky" tests are just consistently broken is pretty sad, we should definitely fix those (after determining if they're even still relevant?) but the coverage change is :+1:

PR #1390, opened 2021-11-29
ci: Skip unit tests in coverage report

olix0r:
The coverage workflow is misconfigured so that it fails to run:
`cargo-tarpaulin` has command line arguments that don't match other
cargo commands (`--packages` instead of `--package`, which cannot be
mixed with `--workspace`). Furthermore, it doesn't seem that we actually
benefit from running tarpaulin over our non-integration tests, since the
coverage reports are not merged.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -31,8 +31,12 @@ jobs:
         run: |
           cd "${CARGO_HOME}/bin"
           curl -sL https://github.com/xd009642/tarpaulin/releases/download/${CARGO_TARPAULIN_VERSION}/cargo-tarpaulin-${CARGO_TARPAULIN_VERSION}-travis.tar.gz | tar xzvf -
-      - run: cargo tarpaulin --locked --workspace --exclude=linkerd2-proxy --exclude=linkerd-app-integration --no-run
-      - run: cargo tarpaulin --locked --workspace --exclude=linkerd2-proxy --exclude=linkerd-app-integration --skip-clean --ignore-tests --no-fail-fast --out=Xml
-      - run: cargo tarpaulin --locked --workspace --package=linkerd-app-integration --no-default-features --skip-clean --no-run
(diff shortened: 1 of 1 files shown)


.github/workflows/coverage.yml
```diff
@@ -31,8 +31,12 @@ jobs:
...
-      - run: cargo tarpaulin --locked --workspace --exclude=linkerd2-proxy --exclude=linkerd-app-integration --skip-clean --ignore-tests --no-fail-fast --out=Xml
-      - run: cargo tarpaulin --locked --workspace --package=linkerd-app-integration --no-default-features --skip-clean --no-run
-      - run: cargo tarpaulin --locked --workspace --package=linkerd-app-integration --no-default-features --skip-clean --ignore-tests --no-fail-fast --out=Xml
+      # XXX(ver) AFAICT, Tarpaulin doesn't allow us to compose a report over multiple invocations,
+      # so we have to choose between getting coverage from unit tests and integration tests (since
+      # integration tests require --no-default-features to avoid flakiness). Currently the
+      # integration tests seem to cover more code, so we skip the unit tests for now :(.
+      #- run: cargo tarpaulin --locked --workspace --exclude=linkerd2-proxy --exclude=linkerd-app-integration --no-run
+      #- run: cargo tarpaulin --locked --workspace --exclude=linkerd2-proxy --exclude=linkerd-app-integration --skip-clean --ignore-tests --no-fail-fast --out=Xml
```
hawkw, line 39: hmm, that's kind of a bummer. Looking at the docs it looks like we might be able to define different configurations (with different featuresets) in [Tarpaulin's config file](https://github.com/xd009642/tarpaulin#config-file), I wonder if we could use that instead of separate runs?

hawkw (approved):
looks good to me! i did notice that there is a mechanism for running Tarpaulin with multiple feature sets, so it might be worth investigating that in a follow-up?

.github/workflows/coverage.yml
olix0r: Cool! If you can help figure that out, would be happy to use it :)

PR #1392, opened 2021-12-03
Fixes for compiling on macOS

tychedelia:
A few fixes to compile the integration tests on macOS. Unfortunately, [4 of the tests are failing](https://gist.github.com/tychedelia/301e82856d85c0f59d649e46c5d0c1ca) (2 intermittently so). I tried to dig in a bit, but wasn't able to figure out what was up. Happy to continue investigating with further direction.

This also includes an ugly Makefile hack to exclude the `linekerd-system` on macOS during tests. The conditional dependency inside `linkerd-app-core` works when running `cargo build`, but I guess `cargo test` wants to consume all workspace members (in order to run tests inside the crate, presumably).

(Also, for any future readers I had to increase the file descriptor limit in order to get these to run -- the default on macOS is randomly super low).

Signed-off-by: Jem McElwain <jmcelwain@gmail.com>

### DIFF
--- a/linkerd/app/integration/src/tests/telemetry.rs
+++ b/linkerd/app/integration/src/tests/telemetry.rs
@@ -1199,10 +1199,7 @@ mod transport {
     async fn inbound_tcp_connect_err() {
         let _trace = trace_init();
         let srv = tcp::server()
-            .accept_fut(move |sock| {
-                drop(sock);
-                future::ok(())
-            })
+            .accept_fut(move |sock| async { drop(sock) })
             .run()
             .await;
         let proxy = proxy::new().inbound(srv).run().await;
@@ -1213,7 +1210,7 @@ mod transport {
         let tcp_client = client.connect().await;
 
         tcp_client.write(TcpFixture::HELLO_MSG).await;
-        assert_eq!(tcp_client.read().await, &[]);
+        assert_eq!(tcp_client.read().await, &[] as &[u8]);
         // Connection to the server should be a failure with the EXFULL error
         metrics::metric("tcp_close_total")
             .label("peer", "dst")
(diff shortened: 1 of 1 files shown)


tychedelia:
I also just realize by trying to run `make lint` that a lot of these common commands will fail because they try to include the `linkerd-system` crate from workspace. I realize y'all aren't working on macOS, but wondering about possibly just removing the forced compilation failure instead?

Makefile
```diff
@@ -101,7 +101,11 @@ shellcheck:
 
 .PHONY: test
 test: fetch
+ifeq ($(findstring darwin,$(CARGO_TARGET)),darwin)
+	$(CARGO_TEST) --workspace --exclude linkerd-system
+else
 	$(CARGO_TEST)
+endif
```
olix0r, line 108: Presumably a bunch of the other targets suffer from the same problem? I think it's probably better to omit this for now in favor of running cargo directly. If we want to support using the makefile on macOS, we probably need to pursue a more holistic approach.

linkerd/app/integration/src/tests/telemetry.rs
```diff
@@ -1233,15 +1230,12 @@ mod transport {
             .await;
     }
 
-    #[test]
+    #[tokio::test]
     #[cfg(target_os = "macos")]
```
olix0r, line 1234: @hawkw do you recall why these tests are macOS-specific? should they instead be marked flakey so we can at least catch build failures? Do we really depend on some macOS specific behavior here, or are these tests just broken?

Makefile
tychedelia: Yeah, it's an issue for many of the targets so this fix wont work. Running cargo manually is totally fine. I was also wondering about the purpose of forcing a compilation error in the system crate - would it be okay to just let it build empty?

linkerd/app/integration/src/tests/telemetry.rs
hawkw: they are macos-specific due to errno labels; some syscalls return different error codes on macos than on Linux, so the errno labels are different on macOS. the tests assert that a label is added for "whatever error code the OS returns": https://github.com/linkerd/linkerd2-proxy/pull/1392/files/92538c2c2764bd22802ad3db9eac676828f070d7#diff-22fea522d6f7f7482dc4554c9d57ffb7bb433c7dc1729ef93ed13864af712cb2R1254

hawkw: as a potential alternative, we could just only test errno labels on Linux!

Makefile
tychedelia: Backed this out.

olix0r: Yeah, removing the compile error is also probably fine.

olix0r (approved):

PR #1396, opened 2021-12-06
Remove macOS-specific transport label tests

olix0r:
The proxy doesn't actually run on macOS, so there's not a lot of value
in having macOS-specific tests, especially as these tests aren't
executed as a part of CI.

This change removes two macOS-specific transport label tests.


hawkw (approved):
FWIW, the intention behind these tests was to allow testing transport error labels while _developing_ on macOS. But, I'm not sure how much value there is in keeping them around, so I'm fine with removing these, especially given that they're not tested on CI so it's easy for developers _not_ on macOS to forget to update them...

PR #1414, opened 2021-12-17
inbound: Simplify GatewayIo type

olix0r:
Since the legacy gateway mode has been removed from the proxy, we no
longer need to support multiple I/O types for the gateway. This change
drops the unused I/O variant from the GatewayIo type.


mateiidavid (approved):

hawkw (approved):
yay!

PR #1417, opened 2021-12-20
Introduce a dedicated direct::LocalHttp target type

olix0r:


linkerd/app/inbound/src/direct.rs
```diff
@@ -137,20 +135,46 @@ impl<N> Inbound<N> {
...
+                                    let policy = policies.check_policy(OrigDstAddr(addr))?;
+                                    let local = match protocol {
+                                        None => {
+                                            let tls = tls::ConditionalServerTls::Some(
+                                                tls::ServerTls::Established {
+                                                    client_id: Some(client.client_id.clone()),
+                                                    negotiated_protocol: client.alpn,
+                                                },
+                                            );
+                                            let permit = policy
```
mateiidavid, line 154: Ah nice, so when we go direct w/o a protocol we can just check the policy here. This is something I was wondering about in my approach; if we check the policy in the direct stack, it'll be checked once again in the http stack?

linkerd/app/inbound/src/direct.rs
```diff
@@ -137,20 +135,46 @@ impl<N> Inbound<N> {
...
+                                                client_id: client.client_id,
+                                            })
+                                        }
+                                        Some(protocol) => {
+                                            // When TransportHeader includes the protocol, but does not
+                                            // include an alternate name we go through the Inbound HTTP
+                                            // stack.
+                                            let addr = (client.local_addr.ip(), port).into();
+                                            let policy =
+                                                policies.check_policy(OrigDstAddr(addr))?;
```
mateiidavid, line 168: Do we still need to check the policy here since we've done that already on L145? Can we re-use it here?

linkerd/app/inbound/src/direct.rs
```diff
@@ -25,14 +25,23 @@ pub struct RefusedNoIdentity(());
...
-pub(crate) struct Local {
+pub(crate) struct LocalHttp {
     addr: Remote<ServerAddr>,
     policy: policy::AllowPolicy,
     client: ClientInfo,
-    protocol: Option<SessionProtocol>,
+    protocol: SessionProtocol,
 }
 
+type Local = svc::Either<LocalTcp, LocalHttp>;
```
mateiidavid, line 43: This is neat! So this basically allows us to just return the inner predicate on L129 and have the switch layer in one line? That's really cool.

mateiidavid (approved):
Makes for a much nicer read now :) It also seems to put me at ease with the policy questions I had. The only thing that I'm really unsure of (and that I put in a comment) is whether through the old approach we'd end up checking the policy twice: once here, and once through the http stack.

linkerd/app/inbound/src/direct.rs
olix0r: correct, thanks. copy-paste error :/

linkerd/app/inbound/src/direct.rs
olix0r: yeah, in this case we only check the authorization for the tcp stack -- this is why the LocalTcp type takes a `permit`: to prove that it has been authorized.

the http stack will do its own authorization checks on requests.

PR #1420, opened 2021-12-21
Update Rust to v1.57.0

olix0r:
Addresses new lints.


olix0r:
Well, it seems like we can't update to v1.57.0 due to the compile-time regressions (as described in https://github.com/rust-lang/rust/issues/89195). Hopefully the fixes in flight will address this.

olix0r:
Per https://github.com/rust-lang/rust/issues/91830#issuecomment-992513380, we'll likely need to wait for 1.59.0

PR #1426, opened 2022-01-02
Decouple client connection metadata from the I/O type

olix0r:
Some information about client connections may not be known a priori. For
example, the local socket address is only known after the connection is
established and the TLS handshake may negotiate additional information
(like the ALPN protocol).

We can use traits on the I/O type, but this means that all wrappers
types need implement all of these traits. This is cumbersome.

This change replaces the `tower::make::MakeConnection` helper trait with
a new `linkerd_stack::MakeConnection` trait that explicity returns a
`Metadata` instance so that connectors can return arbitrary information.

Now, the TCP connector returns a `Local<ClientAddr>` and the TLS
connector conditionally includes the negotiated protocol.

This change sets up the ability for the HTTP client to negotiate
protocol upgrading via ALPN instead of using per-request headers.

---

Changes are broken into distinct commits. No functional changes.


linkerd/stack/src/connect.rs
```diff
@@ -0,0 +1,103 @@
...
+use std::task::{Context, Poll};
+use tokio::io::{AsyncRead, AsyncWrite};
+
+#[derive(Clone, Debug)]
+pub struct WithoutConnectionMetadata<S>(S);
+
+#[derive(Clone, Debug)]
+pub struct MakeConnectionService<S>(S);
+
+pub trait MakeConnection<T> {
```
olix0r, line 13: needs a comment

linkerd/app/outbound/src/endpoint.rs
```diff
@@ -209,10 +209,14 @@ impl<S> Outbound<S> {
...
-        S: svc::Service<tcp::Connect, Error = io::Error> + Clone + Send + Sync + Unpin + 'static,
-        S::Response:
-            tls::HasNegotiatedProtocol + io::AsyncRead + io::AsyncWrite + Send + Unpin + 'static,
-        S::Future: Send + Unpin,
+        S: svc::MakeConnection<tcp::Connect, Metadata = Local<ClientAddr>, Error = io::Error>
+            + Clone
+            + Send
+            + Sync
+            + Unpin
+            + 'static,
```
hawkw, line 217: style nit, doesn't actually matter: consider breaking this into two lines?
```suggestion
        S: svc::MakeConnection<tcp::Connect, Metadata = Local<ClientAddr>, Error = io::Error>
        S: Clone + Send + Sync + Unpin + 'static,
```

linkerd/app/outbound/src/tcp/connect.rs
```diff
@@ -50,9 +51,12 @@ impl<C> Outbound<C> {
             + svc::Param<Option<http::AuthorityOverride>>
             + svc::Param<Option<SessionProtocol>>
             + svc::Param<transport::labels::Key>,
-        C: svc::Service<Connect, Error = io::Error> + Clone + Send + 'static,
-        C::Response: tls::HasNegotiatedProtocol,
-        C::Response: io::AsyncRead + io::AsyncWrite + Send + Unpin + 'static,
+        C: svc::MakeConnection<Connect, Metadata = Local<ClientAddr>, Error = io::Error>
+            + Clone
+            + Send
+            + 'static,
```
hawkw, line 57: style nit, TIOLI: consider
```suggestion
        C: svc::MakeConnection<Connect, Metadata = Local<ClientAddr>, Error = io::Error>
        C: Clone + Send + 'static,
```

linkerd/meshtls/src/client.rs
```diff
@@ -128,22 +130,29 @@ impl<I> Future for ConnectFuture<I>
...
 
     #[inline]
     fn poll(self: std::pin::Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         match self.project() {
             #[cfg(feature = "boring")]
             ConnectFutureProj::Boring(f) => {
                 let res = futures::ready!(f.poll(cx));
-                Poll::Ready(res.map(ClientIo::Boring))
+                Poll::Ready(res.map(|io| {
+                    let np = io.negotiated_protocol().map(|np| np.to_owned());
```
hawkw, line 142: huh, i was surprised that `std` doesn't have an `impl<T: ToOwned> ToOwned for Option<T>` but i guess it would conflict with [the blanket impl for `T: Clone`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#impl-ToOwned)...

linkerd/proxy/http/src/glue.rs
```diff
@@ -191,16 +193,16 @@ where
...
+    F: TryFuture<Ok = (I, M)> + 'static,
     F::Error: Into<Error>,
 {
-    type Output = Result<Connection<F::Ok>, Error>;
+    type Output = Result<Connection<I>>;
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         let this = self.project();
-        let transport = futures::ready!(this.inner.try_poll(cx)).map_err(Into::into)?;
+        let (transport, _meta) = futures::ready!(this.inner.try_poll(cx)).map_err(Into::into)?;
```
hawkw, line 205: nit, TIOLI: using a named ignored argument for the metadata (`_meta`) kind of implies to me that we want to keep it around until the end of the function even though we're not using it, but we _could_ drop it immediately by just writing
```suggestion
        let (transport, _) = futures::ready!(this.inner.try_poll(cx)).map_err(Into::into)?;
```

on the other hand, maybe it's nice to use the name to make it explicit _what_ we're ignoring here? idk, maybe disregard this one. :)

linkerd/proxy/transport/src/connect.rs
```diff
@@ -38,12 +37,13 @@ impl<T: Param<Remote<ServerAddr>>> Service<T> for ConnectTcp {
             let io = TcpStream::connect(&addr).await?;
             super::set_nodelay_or_warn(&io);
             let io = super::set_keepalive_or_warn(io, keepalive)?;
+            let local_addr = io.local_addr()?;
             debug!(
-                local.addr = %io.local_addr().expect("cannot load local addr"),
+                local.addr = %local_addr,
```
hawkw, line 42: this will now bail rather than panicking if the local address is missing...is that what we want? should we be logging an error or something before bailing?

linkerd/stack/src/connect.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    type Future: std::future::Future<
+        Output = Result<(Self::Connection, Self::Metadata), Self::Error>,
+    >;
+
+    /// Determines whether the connector is ready to establish a connection.
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>>;
+
+    /// Establishes a connection.
+    fn make_connection(&mut self, t: T) -> Self::Future;
```
hawkw, line 36: nit/TIOLI: i might have just named this `connect`? but maybe it's better for it to be the trait name. not a blocker either way...

linkerd/stack/src/connect.rs
```diff
@@ -0,0 +1,124 @@
...
+pub trait MakeConnection<T> {
+    /// An I/O type that represents a connection to the remote endpoint.
+    type Connection: AsyncRead + AsyncWrite;
+
+    /// Metadata associated with the established connection.
+    type Metadata;
+
+    type Error: Into<Error>;
+
+    type Future: std::future::Future<
```
hawkw, line 28: nit: i think `Future` is imported via `futures::prelude::*`?
```suggestion
    type Future: Future<
```

linkerd/stack/src/connect.rs
```diff
@@ -0,0 +1,124 @@
...
+    {
+        MakeConnectionService(self)
+    }
+}
+
+impl<T, S, I, M> MakeConnection<T> for S
+where
+    S: Service<T, Response = (I, M)>,
+    S::Error: Into<linkerd_error::Error>,
+    I: tokio::io::AsyncRead + tokio::io::AsyncWrite,
```
hawkw, line 59: nit: `AsyncRead` and `AsyncWrite` are imported in this file:
```suggestion
    I: AsyncRead + AsyncWrite,
```

linkerd/tls/src/client.rs
```diff
@@ -58,9 +58,18 @@ pub struct Client<L, C> {
...
 #[derive(Debug)]
-pub enum Connect<F, I, H: Service<I>> {
-    Connect(#[pin] F, Option<H>),
-    Handshake(#[pin] Oneshot<H, I>),
+pub enum Connect<F, I, H: Service<I>, M> {
+    Connect(#[pin] F, Option<Conditional<H, NoClientTls>>),
+    Handshake(
+        #[pin] Oneshot<H, I>,
+        Option<(Conditional<(), NoClientTls>, M)>,
+    ),
```
hawkw, line 66: nit, take it or leave it: i might give these named fields? not a blocker though.

hawkw (approved):
this looks great --- not having to implement a bunch of additional traits on every I/O wrapper type seems much nicer IMO. i commented on several minor style nits, but I didn't see any blockers! :)

linkerd/proxy/transport/src/connect.rs
olix0r: This will manifest as a connection error and be logged where those errors are handled.

PR #1429, opened 2022-01-03
refactor: rename 'metrics' addr to 'admin' in tests

AaronFriel:
While working on #1428, I was a bit confused reading through tests and noticed that it looked like at some point the admin server was just a metrics server, or referred to as such. This refactor just renames the address in integration tests to be consistent with the name used in the proxy.

Signed-off-by: Aaron Friel <mayreply@aaronfriel.com>

### DIFF
--- a/linkerd/app/integration/src/proxy.rs
+++ b/linkerd/app/integration/src/proxy.rs
@@ -37,7 +37,7 @@ pub struct Listening {
     pub tap: Option<SocketAddr>,
     pub inbound: SocketAddr,
     pub outbound: SocketAddr,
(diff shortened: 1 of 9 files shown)


hawkw (approved):
This seems like a good idea to me! :+1:

olix0r (approved):

hawkw:
Hmm, looks like one of the integration tests still references `metrics`: https://github.com/linkerd/linkerd2-proxy/runs/4705715328?check_suite_focus=true#step:4:937

AaronFriel:
@hawkw @olix0r thanks for the fix!

PR #1434, opened 2022-01-16
Use `AHash` for caches

olix0r:
[`AHash`][ahash] pruports to the be "the fastest DoS-resistant hash currently
available in Rust." It boasts substantial (~10x) latency improvements
over the default hash implementation (SipHash-1-3). It appears to be
an obvious choice for use in our maps, especially where hashing is
frequent and/or performed while a lock is held.

According to <https://lib.rs/crates/ahash>, it has substantial usage.

This change replaces the hash implementation for our metrics stores
and service cache (where locks are employed). It also updates the
service-profile router and traffic splitter, where potentially large
targets may be hashed for each request.

Other uses of the standard `HashMap`--mostly in metadata labels, where
hashing is not performed frequently--are currently left unchanged.

[ahash]: https://github.com/tkaitchuck/aHash/blob/e77cab8c1e15bfc9f54dfd28bd8820c2a7bb27c4/compare/readme.md

### DIFF
--- a/linkerd/app/inbound/src/policy/config.rs
+++ b/linkerd/app/inbound/src/policy/config.rs
@@ -1,6 +1,7 @@
 use super::{discover::Discover, DefaultPolicy, ServerPolicy, Store};
+use ahash::AHashMap as HashMap;
 use linkerd_app_core::{control, dns, identity, metrics, svc::NewService};
-use std::collections::{HashMap, HashSet};
+use std::collections::HashSet;
 
 /// Configures inbound policies.
 ///
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,6 +15,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
+name = "ahash"
+version = "0.7.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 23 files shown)


linkerd/app/inbound/src/policy/config.rs
```diff
@@ -1,6 +1,7 @@
 use super::{discover::Discover, DefaultPolicy, ServerPolicy, Store};
+use ahash::AHashMap as HashMap;
 use linkerd_app_core::{control, dns, identity, metrics, svc::NewService};
-use std::collections::{HashMap, HashSet};
+use std::collections::HashSet;
```
mateiidavid, line 4: Would it make sense to swap out `collections::HashSet` for ahash's HashSet wrapper? 

I'm not sure how frequently hashing would be done here since it's unlikely for ports to change that much in this context.

mateiidavid (approved):

linkerd/app/inbound/src/policy/config.rs
olix0r: this is really just used for configuration and shouldn't factor into application performance. i had tried to change but it forces some annoying changes into `linkerd_app::config`, so i left that out for now.

hawkw:
I discussed some of the criticism of AHash's DoS-resistance claims here: https://github.com/linkerd/linkerd2/pull/7616#issuecomment-1015720869. I don't have a deep enough understanding of the cryptanalysis of hashing algorithms to assess this myself, but the fact that cryptographers have expressed concerns about AHash makes me think we may want to be a bit cautious about using it in the proxy. Some of the data we're storing in these hashmaps can come from completely untrusted sources (e.g. request paths in metrics).

PR #1437, opened 2022-01-20
test: remove `env::set_var` call in test logging initialization 

hawkw:
Per rust-lang/rust#90308, this is potentially a data race. In practice,
I don't think it was actually problematic here, but it also wasn't doing
anything of value, so we should remove it.

This is currently the only `env::set_var` or `env::remove_var` call in
the proxy.

To prevent uses of these functions in the future, I also added a 
`disallowed-methods` configuration in `.clippy.toml` to emit a
warning for any uses of `std::env::set_var` and
`std::env::remove_var`. Unfortunately, this required adding a
`deny` attribute basically everywhere, which is why this diff touches
so many files.

Closes linkerd/linkerd2#7651

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,1 +1,8 @@
 type-complexity-threshold = 500
+disallowed-methods = [
+    # mutating environment variables in a multi-threaded context can
+    # cause data races.
+    # see https://github.com/rust-lang/rust/issues/90308 for details.
+    "std::env::set_var",
+    "std::env::remove_var",
+]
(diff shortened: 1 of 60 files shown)


.clippy.toml
```diff
@@ -1 +1,8 @@
 type-complexity-threshold = 500
+disallowed-methods = [
+    # mutating environment variables in a multi-threaded context can
+    # cause data races.
+    # see https://github.com/rust-lang/rust/issues/90308 for details.
+    "std::env::set_var",
+    "std::env::remove_var",
+]
```
olix0r, line 8:
```suggestion
]
```

olix0r (approved):

PR #1442, opened 2022-01-25
Make OUTBOUND_LISTEN listen all addresses

kebe7jun:
I am writing a project to enhance network performance for Linkerd using eBPF technology, and I have encountered some problems and would like to make some minor modifications to Linkerd. Project address: https://github.com/merbridge/merbridge

In my cases, I want to change the original ip address to 127.x.y.z, but this line abort me do that.

I think that since INBOUND can listen to all addresses, there should be no problem for OUTBOUND to do the same.


olix0r:
I'm not sure that this is appropriate: unlike the inbound proxy, the outbound proxy is only intended to receive connections that originate from the local pod. It would not be suitable for the outbound proxy to handle connections that originate from other pods. I'm concerned that this change could permit unauthorized clients to bypass authorization and to assume another pod's identity.

For example, let's say *pod-alice* is authorized to access *pod-vault*. If any arbitrary client can connect to *pod-alice:4140* and cause it to initiate connections to *pod-vault*, this could pose a substantial security risk.

In the meantime, you can configure the proxy via the environment:

https://github.com/linkerd/linkerd2-proxy/blob/e1e470e8a1cd46345a587542e5188621b2907cad/linkerd/app/src/env.rs#L90

(This would generally entail using `linkerd inject --manual` or modify the proxy injector to support a configuration to change this value).

We'd love to hear more about how merbridge can be used with Linkerd, and I'd be happy to make changes so that we can leverage ebpf in Linkerd; but I don't think that this default change is sound in the current context of Linkerd.

kebe7jun:
Thank you very much for your patient reply.

As for your comment that opening all addresses may make the proxy insecure, I think it may not have much effect because the proxy needs to read ORIGINAL_DST through the kernel when it does the processing of the packet, and if the application sends the packet directly to this port, it will not be allowed. In istio it will be routed to blackhole, so I don't think non-local traffic will have this problem under normal circumstances.


> I'm not sure that this is appropriate: unlike the inbound proxy, the outbound proxy is only intended to receive connections that originate from the local pod. It would not be suitable for the outbound proxy to handle connections that originate from other pods. I'm concerned that this change could permit unauthorized clients to bypass authorization and to assume another pod's identity.
> 
> For example, let's say _pod-alice_ is authorized to access _pod-vault_. If any arbitrary client can connect to _pod-alice:4140_ and cause it to initiate connections to _pod-vault_, this could pose a substantial security risk.
> 
> In the meantime, you can configure the proxy via the environment:
> 
> https://github.com/linkerd/linkerd2-proxy/blob/e1e470e8a1cd46345a587542e5188621b2907cad/linkerd/app/src/env.rs#L90
> 
> (This would generally entail using `linkerd inject --manual` or modify the proxy injector to support a configuration to change this value).
> 
> We'd love to hear more about how merbridge can be used with Linkerd, and I'd be happy to make changes so that we can leverage ebpf in Linkerd; but I don't think that this default change is sound in the current context of Linkerd.

kebe7jun:
If the proxy side can't change this default value, I've adapted it in Merbridge to change the original binding IP through the hook bind4 function to achieve the same effect.

But I don't think this change should have much impact, and I hope the community will accept this PR.

Details: https://github.com/merbridge/merbridge/pull/26

olix0r:
We've confirmed that this change would make Linkerd susceptible to be an open relay in Linkerd's default configuration. Additional changes to Linkerd's proxy-init/CNI would be needed to support this configuration. I'm going to close this PR for now. Feel free to open an issue against the linkerd2 repository if you feel that supporting this change is important.

kebe7jun:
Ok, thanks for you response.

PR #1456, opened 2022-01-31
Ban uses of `Instant` operations that can panic

olix0r:
When comparing instances, we should use saturating varieties to help
ensure that we can't hit panics.

This change bans uses of `std::time::Instant::{duration_since, elapsed,
sub}` via clippy. Uses are ported to using
`Instant::saturating_duration_since`.

Related to linkerd/linkerd2#7748

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -5,4 +5,9 @@ disallowed-methods = [
     # see https://github.com/rust-lang/rust/issues/90308 for details.
     "std::env::set_var",
     "std::env::remove_var",
+
+    # Avoid instances of https://github.com/rust-lang/rust/issues/86470
+    "std::time::Instant::duration_since",
+    "std::time::Instant::elapsed",
+    "std::time::Instant::sub",
 ]
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -9,5 +9,5 @@ disallowed-methods = [
     # Avoid instances of https://github.com/rust-lang/rust/issues/86470
     "std::time::Instant::duration_since",
     "std::time::Instant::elapsed",
-    "std::time::Instant::sub",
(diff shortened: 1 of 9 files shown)


.clippy.toml
```diff
@@ -5,4 +5,9 @@ disallowed-methods = [
     # see https://github.com/rust-lang/rust/issues/90308 for details.
     "std::env::set_var",
     "std::env::remove_var",
-]
\ No newline at end of file
+
+    # Avoid instances of https://github.com/rust-lang/rust/issues/86470
+    "std::time::Instant::duration_since",
+    "std::time::Instant::elapsed",
+    "std::time::Instant::sub",
```
olix0r, line 12: we can't actually detect uses of `sub` via clippy, afaict

olix0r: we should use the `{ path = "...", reason = "..."}` form so that errors include the commented rationale.

hawkw: I believe that's why we've missed the `Uptime` timestamp formatter for the logs, which uses `-` (and clippy maybe can't detect banned methods that are called via operators?): https://github.com/linkerd/linkerd2-proxy/blob/d9dbdece1edf765b6c8d9ccdd16ca611c70c8669/linkerd/tracing/src/uptime.rs#L24

hawkw (commented):
this looks like the correct solution to the problem. however, since clippy's disallowed methods lint doesn't seem to work with operators, it looks like this missed all the uses of `-` with instants, which are equivalent to `elapsed`. I opened #1457 to fix them, as well.

.clippy.toml
olix0r: not supported in our version of rust

PR #1457, opened 2022-02-01
fix uses of `Instant::sub`

hawkw:
It looks like clippy's disallowed methods can't detect the use of
disallowed *operators*, so #1456 misses a bunch of uses of the `-`
operator on `Instant`s, mostly in metrics, tap, and the log timestamp
formatter.

This fixes all of them that I could find.

### DIFF
--- a/linkerd/http-metrics/src/requests/report.rs
+++ b/linkerd/http-metrics/src/requests/report.rs
@@ -137,7 +137,7 @@ where
         metric.fmt_help(f)?;
         Self::fmt_by_class(&registry, f, metric, |s| &s.total)?;
 
-        registry.retain_since(Instant::now().saturating_duration_since(self.retain_idle));
+        registry.retain_since(Instant::now() - self.retain_idle);
 
         Ok(())
     }
--- a/linkerd/http-metrics/src/requests/report.rs
+++ b/linkerd/http-metrics/src/requests/report.rs
@@ -137,7 +137,7 @@ where
         metric.fmt_help(f)?;
         Self::fmt_by_class(&registry, f, metric, |s| &s.total)?;
 
-        registry.retain_since(Instant::now() - self.retain_idle);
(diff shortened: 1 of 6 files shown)


linkerd/transport-metrics/src/report.rs
```diff
@@ -74,7 +74,7 @@ impl<K: Eq + Hash + FmtLabels + 'static> FmtMetrics for Report<K> {
         tcp_close_total.fmt_help(f)?;
         Self::fmt_eos_by(&*metrics, f, tcp_close_total, |e| &e.close_total)?;
 
-        metrics.retain_since(Instant::now() - self.retain_idle);
+        metrics.retain_since(Instant::now().saturating_duration_since(self.retain_idle));
```
hawkw, line 77: oh, this was a `Duration` sub, rather than an `Instant` sub. derp.
```suggestion
        metrics.retain_since(Instant::now() - self.retain_idle);
```

linkerd/http-metrics/src/requests/report.rs
```diff
@@ -137,7 +137,7 @@ where
         metric.fmt_help(f)?;
         Self::fmt_by_class(&registry, f, metric, |s| &s.total)?;
 
-        registry.retain_since(Instant::now() - self.retain_idle);
+        registry.retain_since(Instant::now().saturating_duration_since(self.retain_idle));
```
hawkw, line 140:
```suggestion
        registry.retain_since(Instant::now() - self.retain_idle);
```

hawkw (commented):
i missed that a couple of those were actually `<Instant> - <Duration>` instead, which broke when changed to `saturating_duration_since` (and, the non-saturating subs were legit in that case since the `Duration` is not going to time travel)

olix0r (approved):
Excellent, thanks!

PR #1458, opened 2022-02-01
clippy: Disallow lock and instant types from `std`

olix0r:
We use `parking_lot` locks throughout our code. This change disallows the
introduction of std::sync's locks.

This change also enforces the use of `tokio::time::Instant`, which
allows for a mockable time source in tests.


olix0r:
Added benefit is that we can pickup panic avoidance if/when https://github.com/tokio-rs/tokio/pull/4461 is released

.clippy.toml
```diff
@@ -1,13 +1,21 @@
...
 
-    # Avoid instances of https://github.com/rust-lang/rust/issues/86470
-    "std::time::Instant::duration_since",
-    "std::time::Instant::elapsed",
-    # Clippy doesn't let us ban std::time::Instant::sub, but it knows what it did.
+    # Use parking_lot instead.
+    "std::sync::Mutex",
+    "std::sync::RwLock",
+
+    # Use tokio::time::Instant instead..
```
olix0r, line 13:
```suggestion
    # Use tokio::time::Instant instead.
```

.clippy.toml
```diff
@@ -1,13 +1,21 @@
...
     "std::env::set_var",
     "std::env::remove_var",
 
-    # Avoid instances of https://github.com/rust-lang/rust/issues/86470
-    "std::time::Instant::duration_since",
-    "std::time::Instant::elapsed",
-    # Clippy doesn't let us ban std::time::Instant::sub, but it knows what it did.
+    # Use parking_lot instead.
+    "std::sync::Mutex",
+    "std::sync::RwLock",
```
hawkw, line 11: i don't think this works, this is the `disallowed-methods` lint, but these are type names, rather than method/function names. i believe there's a separate [`disallowed-types` lint](https://rust-lang.github.io/rust-clippy/master/#disallowed_types) for banning types.

i think this will need to go in a separate list, like:
```toml
disallowed-types = [
    # Use parking_lot instead.
    "std::sync::Mutex",
    "std::sync::RwLock",
]
disallowed-methods = [
    # ...
]
```

linkerd/app/admin/src/server.rs
```diff
@@ -201,7 +201,7 @@ where
 mod tests {
     use super::*;
     use http::method::Method;
-    use std::time::Duration;
+    use tokio::time::Duration;
```
hawkw, line 204: i believe `tokio`'s `Duration` is [just a re-export of `std::time::Duration`](https://docs.rs/tokio/latest/tokio/time/index.html#reexports), so we really don't *need* to change this...it's fine that we changed it, but if we wanted to keep the diff small, we could remove it...

i believe the re-export exists mainly so that `Instant` and `Duration` can be imported from the same place?

linkerd/app/core/src/config.rs
```diff
@@ -4,7 +4,7 @@ use crate::{
     svc::{stack::CloneParam, Param},
     transport::{Keepalive, ListenAddr},
 };
-use std::time::Duration;
+use tokio::time::Duration;
```
hawkw, line 7: similarly, this is fine but not necessary.

linkerd/app/core/src/control.rs
```diff
@@ -221,7 +221,7 @@ mod resolve {
 
 mod balance {
     use crate::proxy::http;
-    use std::time::Duration;
+    use tokio::time::Duration;
```
hawkw, line 224: ...and this...

linkerd/app/core/src/dns.rs
```diff
@@ -1,6 +1,6 @@
 pub use linkerd_dns::*;
 use std::path::PathBuf;
-use std::time::Duration;
+use tokio::time::Duration;
```
hawkw, line 3: ...and this...

linkerd/app/core/src/errors.rs
```diff
@@ -8,7 +8,7 @@ pub use tonic::Code as Grpc;
 
 #[derive(Debug, Error)]
 #[error("connect timed out after {0:?}")]
-pub struct ConnectTimeout(pub(crate) std::time::Duration);
+pub struct ConnectTimeout(pub(crate) tokio::time::Duration);
```
hawkw, line 11: ...and this...

linkerd/app/core/src/transport.rs
```diff
@@ -11,7 +11,7 @@ pub use self::allow_ips::AllowIps;
 pub struct Metrics(metrics::Registry<labels::Key>);
 
 impl Metrics {
-    pub fn new(retain_idle: std::time::Duration) -> (Self, metrics::Report<labels::Key>) {
+    pub fn new(retain_idle: tokio::time::Duration) -> (Self, metrics::Report<labels::Key>) {
```
hawkw, line 14: ...and this...

linkerd/app/inbound/src/detect.rs
```diff
@@ -504,7 +504,7 @@ mod tests {
                 negotiated_protocol: None,
             }),
             policy: allow(Protocol::Detect {
-                timeout: std::time::Duration::from_secs(10),
+                timeout: tokio::time::Duration::from_secs(10),
```
hawkw, line 507: ...and this...

linkerd/app/inbound/src/detect.rs
```diff
@@ -533,7 +533,7 @@ mod tests {
...
```
hawkw, line 536: ...and this...

linkerd/app/inbound/src/http/tests.rs
```diff
@@ -513,7 +513,7 @@ fn connect_timeout(
         Box::pin(
             async move {
                 tracing::info!("sleeping so that the proxy hits a connect timeout");
-                tokio::time::sleep(std::time::Duration::from_secs(3)).await;
+                tokio::time::sleep(tokio::time::Duration::from_secs(3)).await;
```
hawkw, line 516: ...and this...

linkerd/app/inbound/src/policy/defaults.rs
```diff
@@ -1,6 +1,6 @@
 use linkerd_app_core::{IpNet, Ipv4Net, Ipv6Net};
 use linkerd_server_policy::{Authentication, Authorization, Protocol, ServerPolicy, Suffix};
-use std::time::Duration;
+use tokio::time::Duration;
```
hawkw, line 3: ...and this...

linkerd/app/inbound/src/test_util.rs
```diff
@@ -15,7 +15,7 @@ use linkerd_app_core::{
 };
 pub use linkerd_app_test as support;
 use linkerd_server_policy::{Authentication, Authorization, Protocol, ServerPolicy};
-use std::time::Duration;
+use tokio::time::Duration;
```
hawkw, line 18: ...and this...

linkerd/app/inbound/src/test_util.rs
```diff
@@ -53,7 +53,7 @@ pub fn default_config() -> Config {
         policy: policy::Config::Fixed {
             default: ServerPolicy {
                 protocol: Protocol::Detect {
-                    timeout: std::time::Duration::from_secs(10),
+                    timeout: tokio::time::Duration::from_secs(10),
```
hawkw, line 56: ...and this...

linkerd/app/inbound/src/test_util.rs
```diff
@@ -73,7 +73,7 @@ pub fn default_config() -> Config {
 pub fn runtime() -> (ProxyRuntime, drain::Signal) {
     let (drain_tx, drain) = drain::channel();
     let (tap, _) = tap::new();
-    let (metrics, _) = metrics::Metrics::new(std::time::Duration::from_secs(10));
+    let (metrics, _) = metrics::Metrics::new(tokio::time::Duration::from_secs(10));
```
hawkw, line 76: ...and this...

linkerd/app/integration/src/tests/profile_dst_overrides.rs
```diff
@@ -61,7 +61,7 @@ async fn wait_for_profile_stage(client: &client::Client, metrics: &client::Clien
             break;
         }
 
-        tokio::time::sleep(std::time::Duration::from_millis(200)).await;
+        tokio::time::sleep(tokio::time::Duration::from_millis(200)).await;
```
hawkw, line 64: ...and this...

linkerd/app/integration/src/tests/profiles.rs
```diff
@@ -129,7 +129,7 @@ macro_rules! profile_test {
                 break;
             }
 
-            tokio::time::sleep(std::time::Duration::from_millis(200)).await;
+            tokio::time::sleep(tokio::time::Duration::from_millis(200)).await;
```
hawkw, line 132: ...and this...

linkerd/app/integration/src/tests/transparency.rs
```diff
@@ -1367,7 +1367,7 @@ async fn retry_reconnect_errors() {
 
     drop(tx); // start `listen` now
               // This may be flaky on CI.
-    tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+    tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
```
hawkw, line 1370: ...and this...

linkerd/app/outbound/src/resolve.rs
```diff
@@ -5,7 +5,7 @@ use linkerd_app_core::{
     },
     svc::{layer, NewService},
 };
-use std::time::Duration;
+use tokio::time::Duration;
```
hawkw, line 8: ...and this...

linkerd/app/outbound/src/test_util.rs
```diff
@@ -51,7 +51,7 @@ pub(crate) fn default_config() -> Config {
 pub(crate) fn runtime() -> (ProxyRuntime, drain::Signal) {
     let (drain_tx, drain) = drain::channel();
     let (tap, _) = tap::new();
-    let (metrics, _) = metrics::Metrics::new(std::time::Duration::from_secs(10));
+    let (metrics, _) = metrics::Metrics::new(tokio::time::Duration::from_secs(10));
```
hawkw, line 54: ...and this...

linkerd/app/src/tap.rs
```diff
@@ -110,7 +110,7 @@ impl Tap {
 impl<T> ExtractParam<tls::server::Timeout, T> for TlsParams {
     #[inline]
     fn extract_param(&self, _: &T) -> tls::server::Timeout {
-        tls::server::Timeout(std::time::Duration::from_secs(1))
+        tls::server::Timeout(tokio::time::Duration::from_secs(1))
```
hawkw, line 113: ...and this...

linkerd/exp-backoff/src/lib.rs
```diff
@@ -7,9 +7,9 @@ use rand::{rngs::SmallRng, thread_rng, SeedableRng};
 use std::future::Future;
 use std::pin::Pin;
 use std::task::{Context, Poll};
-use std::time::Duration;
 use thiserror::Error;
 use tokio::time;
+use tokio::time::Duration;
```
hawkw, line 12: ...and this...

linkerd/meshtls/boring/src/tests.rs
```diff
@@ -1,6 +1,6 @@
 use linkerd_identity::{Credentials, DerX509};
 use linkerd_tls_test_util::*;
-use std::time::Duration;
+use tokio::time::Duration;
```
hawkw, line 3: ...and this...

linkerd/meshtls/rustls/src/tests.rs
```diff
@@ -1,6 +1,6 @@
 use linkerd_identity::{Credentials, DerX509};
 use linkerd_tls_test_util::*;
-use std::time::Duration;
+use tokio::time::Duration;
```
hawkw, line 3: ...and this...

linkerd/proxy/discover/src/buffer.rs
```diff
@@ -4,8 +4,8 @@ use pin_project::pin_project;
 use std::future::Future;
 use std::pin::Pin;
 use std::task::{Context, Poll};
-use std::time::Duration;
 use tokio::sync::{mpsc, oneshot};
+use tokio::time::Duration;
```
hawkw, line 8: ...and this...

linkerd/proxy/http/src/timeout.rs
```diff
@@ -1,7 +1,7 @@
 use linkerd_error::Error;
 use linkerd_stack::{layer, MapErr, NewService, Param, Timeout, TimeoutError};
-use std::time::Duration;
 use thiserror::Error;
+use tokio::time::Duration;
```
hawkw, line 4: ...and this...

linkerd/proxy/transport/src/lib.rs
```diff
@@ -17,8 +17,8 @@ pub use self::{
 };
 use linkerd_io as io;
 use socket2::TcpKeepalive;
-use std::time::Duration;
 use tokio::net::TcpStream;
+use tokio::time::Duration;
```
hawkw, line 21: ...and this...

linkerd/stack/src/failfast.rs
```diff
@@ -171,7 +171,7 @@ where
 #[cfg(test)]
 mod test {
     use super::FailFast;
-    use std::time::Duration;
+    use tokio::time::Duration;
```
hawkw, line 174: ...and this...

hawkw (commented):
overall, i think this is a good idea. my one issue with the current change is that i don't think the Clippy bans for `Mutex` and `RwLock` actually work --- those need to go in the "disallowed-types" lint instead. I suspect if we add them to that, we'll find a bunch of stdlib locks that need to be replaced.

i also noticed that the majority of this diff is changing `std::time::Duration` to `tokio::time::Duration`. this is _fine_, but it isn't actually necessary, as that's just a re-export of `std::time::Duration`. I think it's main purpose is to allow `Instant` and `Duration` to be imported from the same location. in places where we only use `Duration`, changing the import just makes the diff bigger, so if we wanted this to be a smaller diff, we could avoid changing those files. but, i'm fine with making the change for consistency's sake, i just wanted to point that out.

.clippy.toml
```diff
@@ -1,13 +1,22 @@
...
+    "tokio::time::Instant::elapsed",
+    # Clippy doesn't let us ban tokio::time::Instant::sub, but it knows what it did.
+]
+disallowed-types = [
+    # Use parking_lot instead.
+    "std::sync::Mutex",
+    "std::sync::RwLock",
+
+    # Use tokio::time::Instant instead.
+    "std::time::Instant",
```
hawkw, line 21: looking at the clippy docs, it seems like we can also [add reasons](https://github.com/rust-lang/rust-clippy/issues/7609) to these, so that they're displayed with the warning when a disallowed type is used. we may want to add that, but not a blocker.

.clippy.toml
```diff
@@ -6,16 +6,17 @@ disallowed-methods = [
...
-    # Use tokio::time::Instant instead.
-    "std::time::Instant::now",
-
     # Avoid instances of https://github.com/rust-lang/rust/issues/86470 until tokio/tokio#4461 is
     # available.
     "tokio::time::Instant::duration_since",
     "tokio::time::Instant::elapsed",
     # Clippy doesn't let us ban tokio::time::Instant::sub, but it knows what it did.
 ]
+disallowed-types = [
```
hawkw, line 15: i think we might also need to add `#![deny(clippy::disallowed_types)]` to every crate in the project. based on a quick grep, there are definitely a few uses of `std::sync` locks in the codebase, so i'm surprised this isn't catching them...

most uses are in tests, but there's at least one `std::sync::RwLock` in `linkerd/stack/src/fail_on_error.rs` according to grep.

hawkw (commented):
not totally sure if i trust that the disallowed types lint is working, because i found some instances of `std::sync` locks via grepping. we may need to add `#![deny(...)]` attributes for it to work?

.clippy.toml
olix0r: afaict this isn't supported in our version of clippy

.clippy.toml
olix0r: thanks. fixed

.clippy.toml
hawkw: ah, got it. could be nice to add later when it's available.

linkerd/app/integration/src/controller.rs
```diff
@@ -232,7 +228,7 @@ impl pb::destination_server::Destination for Controller {
         let _e = span.enter();
         tracing::debug!(request = ?req.get_ref(), "received");
 
-        if let Ok(mut calls) = self.expect_dst_calls.lock() {
+        let mut calls = self.expect_dst_calls.lock();
             if self.unordered {
```
hawkw, line 232: i believe this formatting change is why rustfmt is failing :) since the lock is no longer an `if let`, the rest of the code shouldn't be indented any longer.

hawkw (approved):
looks good now modulo that rustfmt is sad!

PR #1478, opened 2022-02-09
Do not error when `LINKERD2_PROXY_INBOUND_PORTS` is empty

kleimkuhler:
As part of #7816, we want to be able to set the `LINKERD2_PROXY_INBOUND_PORTS` env var to `""` when an app does not specify its ports. This means that in the proxy environment, `LINKERD2_PROXY_INBOUND_PORTS = ""`.

In order to handle this case, we now need to handle `LINKERD2_PROXY_INBOUND_PORTS` being present or not; if it is present then we need to handle it being `""` or some other value.

With this change, we now parse `LINKERD2_PROXY_INBOUND_PORTS` and if it is empty, handle it the same way that we handle it being absent.

I've tested in this change in an install and the logs now report:

```
time="2022-02-09T22:59:34Z" level=info msg="running version edge-21.4.5"
[     0.000268s]  WARN ThreadId(01) linkerd_app::env: No inbound ports specified via LINKERD2_PROXY_INBOUND_PORTS
[     0.000412s]  INFO ThreadId(01) linkerd2_proxy::rt: Using single-threaded proxy runtime
[     0.000739s]  INFO ThreadId(01) linkerd2_proxy: Admin interface on 0.0.0.0:4191
[     0.000745s]  INFO ThreadId(01) linkerd2_proxy: Inbound interface on 0.0.0.0:4143
[     0.000746s]  INFO ThreadId(01) linkerd2_proxy: Outbound interface on 127.0.0.1:4140
[     0.000747s]  INFO ThreadId(01) linkerd2_proxy: Tap DISABLED
[     0.000748s]  INFO ThreadId(01) linkerd2_proxy: Local identity is default.default.serviceaccount.identity.linkerd.cluster.local
[     0.000750s]  INFO ThreadId(01) linkerd2_proxy: Identity verified via linkerd-identity-headless.linkerd.svc.cluster.local:8080 (linkerd-identity.linkerd.serviceaccount.identity.linkerd.cluster.local)
[     0.000751s]  INFO ThreadId(01) linkerd2_proxy: Destinations resolved via linkerd-dst-headless.linkerd.svc.cluster.local:8086 (linkerd-destination.linkerd.serviceaccount.identity.linkerd.cluster.local)
[     0.008245s]  INFO ThreadId(02) daemon:identity: linkerd_app: Certified identity id=default.default.serviceaccount.identity.linkerd.cluster.local
```

### DIFF
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -547,25 +547,18 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
                 Some(addr) => {
                     // If the inbound is proxy is configured to discover policies, then load the set
                     // of all known inbound ports to be discovered during initialization.
-                    let mut ports = match strings.get(ENV_INBOUND_PORTS)? {
-                        Some(s) => {
-                            if s.is_empty() {
-                                warn!("No inbound ports specified via {}", ENV_INBOUND_PORTS,);
-                                Default::default()
-                            } else {
-                                parse_port_set(&s).map_err(|parse_error| {
-                                    error!(
-                                        "{}={:?} is not valid: {:?}",
-                                        ENV_INBOUND_PORTS, s, parse_error
-                                    );
-                                    EnvError::InvalidEnvVar
-                                })?
-                            }
-                        }
-                        None => {
+                    let mut ports = match strings.get(ENV_INBOUND_PORTS)?.as_deref() {
+                        Some("") | None => {
                             warn!("No inbound ports specified via {}", ENV_INBOUND_PORTS,);
                             Default::default()
                         }
+                        Some(s) => parse_port_set(&s).map_err(|parse_error| {
+                            error!(
+                                "{}={:?} is not valid: {:?}",
+                                ENV_INBOUND_PORTS, s, parse_error
+                            );
+                            EnvError::InvalidEnvVar
+                        })?,
                     };
                     if !gateway.allow_discovery.is_empty() {
                         // Add the inbound port to the set of ports to be discovered if the proxy is
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -547,23 +547,10 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
                 Some(addr) => {
                     // If the inbound is proxy is configured to discover policies, then load the set
                     // of all known inbound ports to be discovered during initialization.
-                    let mut ports = match strings.get(ENV_INBOUND_PORTS)? {
-                        Some(s) => {
-                            if s.is_empty() {
-                                warn!("No inbound ports specified via {}", ENV_INBOUND_PORTS,);
-                                Default::default()
-                            } else {
-                                parse_port_set(&s).map_err(|parse_error| {
-                                    error!(
-                                        "{}={:?} is not valid: {:?}",
-                                        ENV_INBOUND_PORTS, s, parse_error
-                                    );
-                                    EnvError::InvalidEnvVar
-                                })?
-                            }
-                        }
+                    let mut ports = match parse(strings, ENV_INBOUND_PORTS, parse_port_set)? {
+                        Some(ports) => ports,
                         None => {
-                            warn!("No inbound ports specified via {}", ENV_INBOUND_PORTS,);
+                            error!("No inbound ports specified via {}", ENV_INBOUND_PORTS,);
                             Default::default()
                         }
                     };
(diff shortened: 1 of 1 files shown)


linkerd/app/src/env.rs
```diff
@@ -547,10 +547,23 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
+                                        "{}={:?} is not valid: {:?}",
+                                        ENV_INBOUND_PORTS, s, parse_error
+                                    );
+                                    EnvError::InvalidEnvVar
+                                })?
+                            }
+                        }
                         None => {
-                            error!("No inbound ports specified via {}", ENV_INBOUND_PORTS,);
+                            warn!("No inbound ports specified via {}", ENV_INBOUND_PORTS,);
```
kleimkuhler, line 566: I'm open to suggestions on simplifying this. I originally tried something similar to

```rust
Some("") | None => {
    warn!("No inbound ports specified via {}", ENV_INBOUND_PORTS,);
    Default::default()
}
Some(s) => {
    ...
}
```

But we cannot match on `Some("")` unfortunately. Therefore, we have the duplicate handling of issuing the `warn!` and setting `ports = Default::default()`.

hawkw: I believe the reason we cannot match `Some` here is because the value is an `Option<&String>` rather than an `Option<&str>`, if it was an `&str`, `Some(''")` should work. I think if we call `as_deref()` on the option before matching it, the empty string pattern should work. See this playground as an example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6fd8540f90081c6360ac5c10de336ae8

kleimkuhler: Ah that's great!

linkerd/app/src/env.rs
```diff
@@ -547,12 +547,18 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
-                    let mut ports = match parse(strings, ENV_INBOUND_PORTS, parse_port_set)? {
-                        Some(ports) => ports,
-                        None => {
-                            error!("No inbound ports specified via {}", ENV_INBOUND_PORTS,);
+                    let mut ports = match strings.get(ENV_INBOUND_PORTS)?.as_deref() {
+                        Some("") | None => {
+                            warn!("No inbound ports specified via {}", ENV_INBOUND_PORTS,);
                             Default::default()
                         }
+                        Some(s) => parse_port_set(&s).map_err(|parse_error| {
```
olix0r, line 555: `s` is already a reference

```suggestion
                        Some(s) => parse_port_set(s).map_err(|parse_error| {
```

olix0r:
> With this change, we now parse LINKERD2_PROXY_INBOUND_PORTS and if it is empty, handle it the same way that we handle it being absent.

I'm a little confused by this vs the title of the PR. I'd think we want to emit an error when it is absent, and if it is set and empty, then we should continue on happily without logging. I'd expect the current behavior to basically work this way.

What about the current behavior is problematic? Why do we want to handle missing and empty as if they're the same?

kleimkuhler:
> What about the current behavior is problematic? Why do we want to handle missing and empty as if they're the same?

If the env var is `""`, the proxy tries to parse this as an integer

```bash
[     0.000258s] ERROR ThreadId(01) linkerd_app::env: LINKERD2_PROXY_INBOUND_PORTS="" is not valid: NotAnInteger(ParseIntError { kind: Empty })
Invalid configuration: invalid environment variable
```

This change handles missing and empty as if they are same because when it's empty, we don't want to try parsing it as a port set.

> I'm a little confused by this vs the title of the PR.

So the title is in regards to the fact that the proxy fails to start when `LINKERD2_PROXY_INBOUND_PORTS=""`. Yes we do still emit warnings, but those could be removed.

hawkw:
> > If the env var is "", the proxy tries to parse this as an integer
> 
> Ahh, ok.
> 
> Can we fix that more simply with:
> 
> ```diff
> diff --git a/linkerd/app/src/env.rs b/linkerd/app/src/env.rs
> index 00069a60..bcb0eeb6 100644
> --- a/linkerd/app/src/env.rs
> +++ b/linkerd/app/src/env.rs
> @@ -935,6 +935,9 @@ fn parse_addr(s: &str) -> Result<Addr, ParseError> {
>  
>  fn parse_port_set(s: &str) -> Result<HashSet<u16>, ParseError> {
>      let mut set = HashSet::new();
> +    if s.is_empty() {
> +        return Ok(set);
> +    }
>      for num in s.split(',') {
>          set.insert(parse_number::<u16>(num)?);
>      }
> ```

hmm, we may want to double-check other env var parsing functions for how _they_ handle empty strings, while we're looking at this code...

linkerd/app/src/env.rs
olix0r: @kleimkuhler clippy is still complaining about this ^

olix0r: also I thought we were going to back out this change in favor of a change to parse_port_set

olix0r (approved):

PR #1483, opened 2022-02-11
ci: Add workflows to run beta, nightly builds

olix0r:
We currently only really catch Rust regressions as they hit stable. This
change adds workflows to build the proxy daily against `nightly` and
weekly against `beta`.

These workflows are separated so that they can run on different
schedules and so that a failure on one toolchain does not interfere with
the other.

### DIFF
--- a/.github/workflows/beta.yml
+++ b/.github/workflows/beta.yml
@@ -1,0 +1,31 @@
+# Builds the proxy on the beta toolchain to help catch Rust regressions before they bite us hit stable.
+name: beta
+
+on:
+  pull_request:
+    paths:
+      - .github/workflows/beta.yml
+  schedule:
+    # Run weekly on wednesday @ midnightish Pacific-time.
+    - cron: "30 7 * * 3"
+
+env:
+  CARGO_INCREMENTAL: 0
+  CARGO_NET_RETRY: 10
+  RUST_BACKTRACE: short
+  RUSTUP_MAX_RETRIES: 10
+
+permissions:
+  contents: read
+
+jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/beta.yml
```diff
@@ -0,0 +1,31 @@
+# Builds the proxy on the beta toolchain to help catch Rust regressions before they bite us hit stable.
```
hawkw, line 1: "before they bite us hit stable" seems grammatically weird...was one of those supposed to be deleted?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,32 @@
+# Builds the proxy on the nightly toolchain to help catch Rust regressions before they bite us hit
```
hawkw, line 1: same grammar bug here

hawkw (approved):
looks good to me modulo a small grammar bug in the comment.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,32 @@
...
+# stable.
+name: nightly
```
hawkw, line 3: perhaps this should be named "nightly rust" or something, so it doesn't imply that it's a nightly build of the _proxy_? not a hard blocker

hawkw:
failure on nightly appears to be due to a compiler regression (https://github.com/rust-lang/rust/issues/93821), i believe there's a PR fixing it (https://github.com/rust-lang/rust/pull/93893), but it's not in yesterday's nightly. we can ignore that for now, as it will be fixed shortly.

PR #1484, opened 2022-02-11
ci: Add a toolchain workflow

olix0r:
The toolchain workflow validates that the same rust toolchain version is
used in all docker images, matching the value in the `rust-toolchain`
file.

This change also updates the devcontainer to include the `yq` utility.

### DIFF
--- a/.github/workflows/toolchain.yml
+++ b/.github/workflows/toolchain.yml
@@ -1,0 +1,60 @@
+name: rust-toolchain
+
+on:
+  pull_request:
+    paths:
+      - rust-toolchain
+      - "**Dockerfile"
+      - ".github/workflows/*"
+      - ".github/**/Dockerfile"
+
+permissions:
+  contents: read
+
+env:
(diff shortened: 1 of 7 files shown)


.github/workflows/toolchain.yml
```diff
@@ -0,0 +1,60 @@
...
+            for mismatch in $(echo "$version" | grep -vF "$VERSION" || true) ; do
+              echo "::error file=$f::$f uses incorrect rust version(s): $mismatch"
+              ex=$((ex + 1))
+            done
+          done
+          exit $ex
+
+  workflows:
+    runs-on: ubuntu-latest
+    steps:
```
mateiidavid, line 40: Might be a silly question but would the yq action simplify anything here?

https://github.com/mikefarah/yq/issues/844#issuecomment-856700574

I guess it might not be as easy to check the result if we used the action as opposed to doing it in bash during code checkout?

mateiidavid (approved):
Looks good

PR #1491, opened 2022-02-11
ci: Only run checks on Cargo.toml changes

olix0r:
The point of this job is to check that each crate can be built
independently (i.e., that it doesn't depend on features pulled in by
another crate).

This change also updates the workflow to check linkerd-meshtls-boring by
installing its build deps.


hawkw (approved):
lgtm!

PR #1500, opened 2022-02-15
Fix dockerfile

olix0r:
PR #1479 changed the docker base image to not include a shell, which
broke the dockerfile's logic for skipping the identity wrapper. It's no
longer possible to run the proxy without identity, so this functionality
isn't needed in any case.

This change fixes the Dockerfile and adds a CI workflow to test docker
builds on Dockerfile changes.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -11,6 +11,9 @@ permissions:
 jobs:
   build:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
+    env:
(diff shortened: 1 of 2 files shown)


hawkw:
Looks like the CI job doesn't have BuildKit enabled, which the dockerfile requires: https://github.com/linkerd/linkerd2-proxy/runs/5204736211?check_suite_focus=true#step:3:50

It brings me no pleasure to report this.

hawkw (approved):
looks good to me provided CI passes!

PR #1501, opened 2022-02-16
ci: Run fuzzing only on related changes

olix0r:
We previously always built fuzzers. This change updates the fuzzers
workflow as follows:

* the workflow is only invoked when a fuzz crate changes (or when the
  workflow itself changes).
* fuzzer builds are only executed for changed crate directories


hawkw (approved):
seems good to me!

PR #1503, opened 2022-02-16
deps: update to `tokio` 1.17.0 and `tower` 0.4.12, and unpatch

hawkw:
This branch updates the proxy's dependencies on Tokio and Tower to
1.17.0 and 0.4.12, respectively. These releases contain fixes for panics
when `std::time::Instant` fails to be properly monotonic.

Previously, we were using Git dependencies to ensure we got these fixes.
Now, we can remove those patches, as the released versions of these
crates have the necessary changes.


deny.toml
```diff
@@ -48,7 +48,7 @@ deny = [
     { name = "rustls", wrappers = ["tokio-rustls"] },
 ]
 skip = [
-    # Waiting on a tokio release that updates parking_lot to v0.12.
+    # waiting on tracing-subscriber
```
olix0r, line 51: note that tracing-subscriber is still pulling in parking_lot 0.11

olix0r (approved):

PR #1504, opened 2022-02-17
cargo: Omit patch versions from Cargo.toml files

olix0r:
There's no need for us to specify patch versions in Cargo.toml files,
since we commit our Cargo.lock file. These versions mean that dependency
changes touch unnecessary files, requiring "shotgun surgery" for routine
dependency bumps (and also triggering CI workflows for these subcrates).

Let's omit patch versions from Cargo.toml


.github/workflows/check-all.yml
```diff
@@ -0,0 +1,42 @@
...
+      - run: |
+          curl --proto =https --tlsv1.3 -vsSfLo /usr/local/bin/cargo-action-fmt "https://github.com/olix0r/cargo-action-fmt/releases/download/release%2F${CARGO_ACTION_FMT_VERSION}/cargo-action-fmt-x86_64-unknown-linux-gnu"
+          chmod 755 /usr/local/bin/cargo-action-fmt
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
+      - run: cargo fetch
+      - run: |
+          cargo check --frozen \
+            --workspace \
+            --all-targets \
+            --exclude=linkerd-meshtls-boring \
```
hawkw, line 40: why is `meshtls-boring` excluded from this?

hawkw (approved):
per our discussion on discord, i'm okay with this! :+1:

.github/workflows/check-all.yml
olix0r: it needs external deps (clang & cmake) to build; and it's not actually enabled in builds right now.

hawkw: ah, got it! carry on then!

PR #1517, opened 2022-02-23
Add ExponentialBackoff::new_unchecked

olix0r:
Now that we can use const functions, we can update `ExponentialBackoff`
to expose a constructor instead of having public members.


hawkw (approved):
lovely!

PR #1532, opened 2022-03-08
Use `mod.rs` files for modules

olix0r:
In 23be602, we updated all modules so that the `mod.rs` files were moved
out of the parent directory. For example, `server/mod.rs` was renamed to
`server.rs`. In practice, it's generally more confusing and harder to
navigate when module definitions are decoupled from the rest of the
module.

This change reverts to using `mod.rs` files for modules.


olix0r:
eh, I'll be less grumpy and get used to it I guess

PR #1533, opened 2022-03-09
Use the connection's HTTP version in transport header

olix0r:
Currently, when the outbound proxy communicates with a meshed endpoint
that uses opaque transport (i.e. multi-cluster gateways), it *always*
sets the gateway header's session protocol to HTTP/2, since the meshed
endpoint supports HTTP/2 protocol upgrading. But the HTTP client may
choose not to use HTTP/2 if the request includes the `Upgrade` header,
as it does for WebSocket connections. In these cases, the transport
header should indicate that the connection is HTTP/1.

This change modifies the HTTP client to pass the used protocol version
when building a connection. This value is then used to set the session
protocol header when it is required..

### DIFF
--- a/linkerd/app/inbound/src/http/router.rs
+++ b/linkerd/app/inbound/src/http/router.rs
@@ -100,6 +100,7 @@ impl<C> Inbound<C> {
                 .check_service::<Http>()
                 .push(transport::metrics::Client::layer(rt.metrics.proxy.transport.clone()))
                 .check_service::<Http>()
+                .push_map_target(|(_version, target)| target)
                 .push(http::client::layer(
                     config.proxy.connect.h1_settings,
                     config.proxy.connect.h2_settings,
--- a/linkerd/app/outbound/src/http/endpoint.rs
+++ b/linkerd/app/outbound/src/http/endpoint.rs
@@ -193,7 +193,7 @@ impl<T: svc::Param<transport::labels::Key>> svc::Param<transport::labels::Key> f
 #[cfg(test)]
 mod test {
     use super::*;
-    use crate::{http, test_util::*, transport::addrs::*};
+    use crate::{http, test_util::*};
     use linkerd_app_core::{
         io,
         proxy::api_resolve::Metadata,
@@ -212,7 +212,7 @@ mod test {
         let addr = SocketAddr::new([192, 0, 2, 41].into(), 2041);
 
         let connect = support::connect()
-            .endpoint_fn_boxed(addr, |_: http::Endpoint| serve(::http::Version::HTTP_11));
+            .endpoint_fn_boxed(addr, |_: http::Connect| serve(::http::Version::HTTP_11));
 
         // Build the outbound server
         let (rt, _shutdown) = runtime();
@@ -249,7 +249,7 @@ mod test {
         let addr = SocketAddr::new([192, 0, 2, 41].into(), 2042);
 
         let connect = support::connect()
-            .endpoint_fn_boxed(addr, |_: http::Endpoint| serve(::http::Version::HTTP_2));
+            .endpoint_fn_boxed(addr, |_: http::Connect| serve(::http::Version::HTTP_2));
 
         // Build the outbound server
         let (rt, _shutdown) = runtime();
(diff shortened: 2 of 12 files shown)


mateiidavid (approved):
It looks good _to me_

linkerd/app/inbound/src/http/router.rs
```diff
@@ -100,6 +100,7 @@ impl<C> Inbound<C> {
                 .check_service::<Http>()
                 .push(transport::metrics::Client::layer(rt.metrics.proxy.transport.clone()))
                 .check_service::<Http>()
+                .push_map_target(|(_version, target)| target)
```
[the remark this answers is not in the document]
olix0r: Only in the outbound case can the protocol be upgraded from HTTP/1 to HTTP/2; so in both of these cases:
1. The target already has the correct HTTP version, and the client is bound to a single HTTP version; and
2. The inner stack doesn't actually care about the HTTP version (at the moment, at least).

linkerd/app/outbound/src/http/endpoint.rs
```diff
@@ -132,17 +143,66 @@ impl errors::HttpRescue<Error> for ClientRescue {
...
+{
+    fn param(&self) -> Option<http::AuthorityOverride> {
+        self.inner.param()
+    }
+}
+
+impl<T: svc::Param<transport::labels::Key>> svc::Param<transport::labels::Key> for Connect<T> {
+    fn param(&self) -> transport::labels::Key {
+        self.inner.param()
+    }
```
hawkw, line 190: nit: should these `param` methods all be inlined?
```suggestion
    #[inline]
    fn param(&self) -> Remote<ServerAddr> {
        self.inner.param()
    }
}

impl<T: svc::Param<tls::ConditionalClientTls>> svc::Param<tls::ConditionalClientTls>
    for Connect<T>
{
    #[inline]
    fn param(&self) -> tls::ConditionalClientTls {
        self.inner.param()
    }
}

impl<T: svc::Param<Option<opaque_transport::PortOverride>>>
    svc::Param<Option<opaque_transport::PortOverride>> for Connect<T>
{
    #[inline]
    fn param(&self) -> Option<opaque_transport::PortOverride> {
        self.inner.param()
    }
}

impl<T: svc::Param<Option<http::AuthorityOverride>>> svc::Param<Option<http::AuthorityOverride>>
    for Connect<T>
{
    #[inline]
    fn param(&self) -> Option<http::AuthorityOverride> {
        self.inner.param()
    }
}

impl<T: svc::Param<transport::labels::Key>> svc::Param<transport::labels::Key> for Connect<T> {
    #[inline]
    fn param(&self) -> transport::labels::Key {
        self.inner.param()
    }
```

linkerd/app/outbound/src/http/endpoint.rs
```diff
@@ -270,6 +324,63 @@ mod test {
...
 
+    #[tokio::test(flavor = "current_thread")]
+    async fn orig_proto_skipped_on_http_upgrade() {
+        let _trace = linkerd_tracing::test::trace_init();
+
+        let addr = SocketAddr::new([192, 0, 2, 41].into(), 2041);
+
+        // Pretend the upstream is a proxy that supports proto upgrades. The service needs to
+        // support both HTTP/1 and HTTP/2 because an HTTP/2 connection is maintained by default and
+        // HTTP/1 connections are created as-needed.
```
hawkw, line 335: <3 for the comment

linkerd/proxy/http/src/glue.rs
```diff
@@ -163,15 +163,15 @@ impl<C, T> HyperConnect<C, T> {
...
             target,
+            absolute_form,
         }
     }
 }
 
 impl<C, T> Service<hyper::Uri> for HyperConnect<C, T>
 where
-    C: MakeConnection<T> + Clone + Send + Sync,
+    C: MakeConnection<(crate::Version, T)> + Clone + Send + Sync,
```
hawkw, line 174: this kind of makes me wonder if the `Connect<T>` type should just live in `linkerd-proxy-http`, but i guess then we wouldn't be able to write all the necessary `Param` impls for `Connect<T>`...it's a shame we can't just have a 

```rust
impl<T, P> Param<P> for Connect<T> where T: Param<P> { ... }
```
without conflicting with the existing `Param` blanket impls :(

hawkw (approved):
looks good to me!

linkerd/proxy/http/src/glue.rs
olix0r: yeah, agreed that it would be a lot nicer with blanket impulse. in general, though, I prefer to keep the target types in the modules from which they are emitted.

PR #1560, opened 2022-03-24
inbound: Support multiple authorization types

olix0r:
The inbound policy module uses the label `saz_name` to indicate the
authorization resource being employed to allow/deny traffic. This
corresponds to the `ServerAuthorization` kubernetes resource (with the
`saz` shortname). This resource type is going to be deprecated in favor
of a new, more general, `AuthorizationPolicy` resource.

When this change is made in the control plane, the policy controller
will include a `kind` label on gRPC messages indicating whether the
resource type, or `default` if a default policy is in effect.

This change honors this new `kind` field and adds a dedicated label to
indicate the kind.

Server labels are changed from:

    srv_name="default:foo"
    srv_name="fah"

to:

    srv_kind="default",srv_name="foo"
    srv_kind="server",srv_name="fah"

Authorization labels are changed from:

    saz_name="default:bar"
    saz_name="bah"

to:

    authz_kind="default",authz_name="bar"
    authz_kind="serverauthorization",authz_name="bah"

### DIFF
--- a/linkerd/app/inbound/src/policy/discover.rs
+++ b/linkerd/app/inbound/src/policy/discover.rs
@@ -110,7 +110,8 @@ fn to_policy(proto: api::Server) -> Result<ServerPolicy> {
     };
 
     let loopback = Authorization {
-        name: "default:localhost".into(),
+        kind: "default".into(),
+        name: "localhost".into(),
         authentication: Authentication::Unauthenticated,
         networks: vec![
             Network {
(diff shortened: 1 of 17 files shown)


linkerd/app/inbound/src/policy/discover.rs
```diff
@@ -176,32 +177,60 @@ fn to_policy(proto: api::Server) -> Result<ServerPolicy> {
...
+                    match (parts.next().unwrap(), parts.next()) {
+                        (kind, Some(name)) => (kind.into(), name.into()),
+                        (name, None) => {
+                            let kind = labels
+                                .get("kind")
+                                .cloned()
+                                .unwrap_or_else(|| "serverauthorization".to_string());
+                            (kind.into(), name.into())
+                        }
+                    }
```
hawkw, line 196: nit/TIOLI: this looks like a parsing helper that can be factored out? not a big deal though, as it's only repeated twice

hawkw (approved):
this looks good to me!

PR #1570, opened 2022-03-30
2.11.2 cherrypick

olix0r:


hawkw (approved):
It looks like there's some fixup commits in here, do we intend to merge them as-is, or squash them prior to merging?

Fine with me either way, just curious!

PR #1617, opened 2022-04-19
outbound: Preserve opaqueness on unknown endpoints

olix0r:
The outbound stack only honors opaqueness when the profile response
clearly indicates that the target is a known endpoint or logical
service. This ignores the case when the target is unknown but the target
port is in the default opaque ports list, in which case the profile
response has no metadata except for the opaqueness setting.

This change handles this case explicitly and adds a test for the
`switch_logical` stack to ensure that these profile responses are
honored.

Fixes linkerd/linkerd2#8273

### DIFF
--- a/linkerd/app/outbound/src/switch_logical.rs
+++ b/linkerd/app/outbound/src/switch_logical.rs
@@ -32,10 +32,11 @@ impl<S> Outbound<S> {
                 .push_switch(
                     move |(profile, target): (Option<profiles::Receiver>, T)| -> Result<_, Infallible> {
                         if let Some(rx) = profile {
+                            let is_opaque = rx.is_opaque_protocol();
+
                             // If the profile provides an endpoint, then the target is single endpoint and
                             // not a logical/load-balanced service.
                             if let Some((addr, metadata)) = rx.endpoint() {
-                                let is_opaque = rx.is_opaque_protocol();
                                 tracing::debug!(%is_opaque, "Profile describes an endpoint");
                                 return Ok(svc::Either::A(Endpoint::from_metadata(
                                     addr,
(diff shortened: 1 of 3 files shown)


linkerd/app/outbound/src/switch_logical.rs
```diff
@@ -52,14 +53,28 @@ impl<S> Outbound<S> {
                                 tracing::debug!("Profile describes a logical service");
                                 return Ok(svc::Either::B(Logical::new(logical_addr, rx)));
                             }
+
+                            // If there was a profile but it didn't include an
+                            // endpoint or logical address, create a bare
+                            // endpoint from the original destination address,
+                            // using the profile-provided opaqeuness. This
+                            // mostly applies to external targets that may be
+                            // included in the default opaque list.
```
hawkw, line 62: super nitpicky: this comment appears to be wrapped at a different column than the previous comments in this file? obviously, this isn't actually important but it looks slightly weird.

linkerd/app/outbound/src/switch_logical.rs
```diff
@@ -52,14 +53,28 @@ impl<S> Outbound<S> {
                                 tracing::debug!("Profile describes a logical service");
                                 return Ok(svc::Either::B(Logical::new(logical_addr, rx)));
                             }
+
+                            // If there was a profile but it didn't include an
+                            // endpoint or logical address, create a bare
+                            // endpoint from the original destination address,
+                            // using the profile-provided opaqeuness. This
```
hawkw, line 60: little typo here:
```suggestion
                            // using the profile-provided opaqueness. This
```

hawkw, line 60: super unimportant nit, take it or leave it: this sentence is kind of long, i might break it up
```suggestion
                            // There was a profile but it didn't include an
                            // endpoint or logical address. Create a bare
                            // endpoint from the original destination address,
                            // using the profile-provided opaqueness. This
```

linkerd/app/outbound/src/switch_logical.rs
```diff
@@ -52,14 +53,28 @@ impl<S> Outbound<S> {
...
+                            // mostly applies to external targets that may be
```
hawkw, line 61: the "mostly" here is a little bit unclear to me --- reading this, i would want to know whether there are any other cases.

hawkw (approved):
this change looks good to me, makes sense. i left a handful of minor comments on the comments you added, but none of them are blockers.

kleimkuhler (approved):
👍

PR #1619, opened 2022-04-20
Update deprecated codes

jelipo:
Signed-off-by: Haining <haining.cao@daocloud.io>

`std::usize::MAX` is a deprecated way.

https://doc.rust-lang.org/core/usize/constant.MAX.html


olix0r:
@jelipo Thanks. If I'm reading that correctly, it sounds like it's not actually deprecated yet. When was this change introduced in Rust?

jelipo:
@olix0r 
Look like this commit.
https://github.com/rust-lang/rust/commit/6f3df006105ee637f91f0e2af8ecd00c0fda03e6

olix0r:
OK, it looks like this was added in [1.43.0](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1430-2020-04-23), which is long enough ago that this seems okay. We generally want the ability to degrade rust versions if we hit a regression, but 1.43 is plenty old.

PR #1642, opened 2022-05-02
trace: add `/logs.json` endpoint to admin server

hawkw:
This branch adds a new `/logs.json` endpoint to the proxy's admin
server. This endpoint accepts HTTP `GET` requests with a log filter as a
query string, or `QUERY` requests with a filter string as a body, and
responds with a stream of JSON objects representing each `tracing` event
that matches that filter, as described in linkerd/linkerd2#7250.
Multiple concurrent `/logs.json` requests with different filters are
supported, and each request will only recieve the logged events that
match the filter it requested.

The `/logs.json` endpoint is feature flagged, and is only accessible
from `localhost`, similarly to the `/proxy-log-level` endpoint. If the
feature flag is not enabled, `/logs.json` will respond with a 404 error.
Currently, the `log-streaming` feature is not enabled by default, so the
`/logs.json` endpoint won't do anything unless the proxy is compiled
with the feature flag enabled.

In the future, we'll want to figure out a better auth story for the logs
endpoint (and the proxy's admin server more generally); when we do, we
can enable the feature flag in release builds. There's also room for
some additional work to make the implementation cleaner and improve
performance, but some of this will require upstream library changes as
well. I've tried to comment on most of the future work where
appropriate.

Closes linkerd/linkerd2#7250


olix0r:
Unfortunately there seems to be some conflict with the recently merged uptime PR. Looks like there are some merge conflicts.

olix0r:
Unfortunately there seems to be some conflict with the recently merged uptime PR. Looks like there are some merge conflicts.

hawkw:
> Unfortunately there seems to be some conflict with the recently merged uptime PR. Looks like there are some merge conflicts.

fixed it, thanks! i also added some additional comments; hope that's helpful!

linkerd/tracing/src/stream.rs
```diff
@@ -0,0 +1,365 @@
...
+
+pub(crate) type StreamLayer<S> = Filtered<WriterLayer<S>, StreamFilter, S>;
+
+/// A handle for starting new log streams, and removing old ones.
+#[derive(Debug)]
+
+pub struct StreamHandle<S> {
+    handle: reload::Handle<StreamLayer<S>, S>,
+    channel_capacity: usize,
+    channel_settings: WithCapacity,
```
olix0r, line 54: it's not really clear to me how these two capacities relate to each other. Is one of these "number of lines" to buffer and the other is "max size of a line?"

olix0r (approved):
lgtm! one last question

linkerd/tracing/src/stream.rs
hawkw: i'll add a comment

hawkw: 8fb8bc8c095022686c8dd7195393535fecb749d4 adds docs for this.

PR #1649, opened 2022-05-05
ingress: Fall back to normal TCP forwarding

olix0r:
Currently, ingress-mode proxies ONLY support HTTP traffic. This is an
unfortunate tradeoff, as ingress may make arbitrary outbound connections
(e.g., TLS calls to external services). To avoid requiring that these
connections be completely skipped from the proxy, we can fallback to
transporting these connections after HTTP detection fails.

Ingress-mode proxies DO NOT fully honor port opacity configurations:
HTTP detection is always performed before discovery is attempted.

Fixes linkerd/linkerd2#7238

### DIFF
--- a/linkerd/app/outbound/src/ingress.rs
+++ b/linkerd/app/outbound/src/ingress.rs
@@ -8,7 +8,7 @@ use linkerd_app_core::{
     },
     svc::{self, stack::Param},
     tls,
-    transport::{self, OrigDstAddr, Remote, ServerAddr},
+    transport::{OrigDstAddr, Remote, ServerAddr},
     AddrMatch, Error, Infallible, NameAddr,
 };
 use thiserror::Error;
@@ -34,14 +34,12 @@ enum Target {
 struct ProfileRequired;
 
 #[derive(Debug, Default, Error)]
-#[error("ingress-mode routing is HTTP-only")]
-struct IngressHttpOnly;
-
-#[derive(Debug, Default, Error)]
 #[error("l5d-dst-override is not a valid host:port")]
 struct InvalidOverrideHeader;
 
 const DST_OVERRIDE_HEADER: &str = "l5d-dst-override";
+
+type DetectIo<I> = io::PrefixedIo<I>;
 
 // === impl Outbound ===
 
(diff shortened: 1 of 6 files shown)


linkerd/app/outbound/src/ingress.rs
```diff
@@ -61,185 +71,202 @@ impl Outbound<svc::ArcNewHttp<http::Endpoint>> {
...
+                            )
+                            .instrument(|_: &_| info_span!("forward"))
+                            .into_inner(),
+                    )
+                    .push(svc::ArcNewService::layer())
+                    // Obtain a new inner service for each request. Override stacks are
+                    // cached, as they depend on discovery that should not be performed
+                    // many times. Forwarding stacks are not cached explicitly, as there
+                    // are no real resources we need to share across connections. This
+                    // allows us to avoid buffering requests to these endpoints.
```
hawkw, line 206: thanks for the comment here!

linkerd/app/outbound/src/ingress.rs
```diff
@@ -61,185 +71,202 @@ impl Outbound<svc::ArcNewHttp<http::Endpoint>> {
...
+                new_http
+                    .push(http::NewServeHttp::layer(*h2_settings, rt.drain.clone()))
+                    .push_switch(
+                        |(http, t): (Option<http::Version>, T)| -> Result<_, Infallible> {
+                            let accept = tcp::Accept::from(t.param());
+                            if let Some(version) = http {
+                                Ok(svc::Either::A(http::Accept::from((version, accept))))
+                            } else {
+                                Ok(svc::Either::B(accept))
+                            }
```
hawkw, line 261: very unimportant code golf: a `match` would actually be one line shorter here:
```suggestion
                            match http {
                                Some(version) => Ok(svc::Either::A(http::Accept::from((version, accept)))),
                                None => Ok(svc::Either::B(accept)),
                            }
```

hawkw (approved):
looks good to me! commented on a few very minor nits.

linkerd/app/outbound/src/ingress.rs
olix0r: rustfmt makes the match statement longer, though.

I think the shortest form is actually
```rust
                            if let Some(version) = http {
                                return Ok(svc::Either::A(http::Accept::from((version, accept))));
                            }
                            Ok(svc::Either::B(accept))
```

PR #1659, opened 2022-05-09
metrics: add `process_uptime_seconds_total` metric

hawkw:
The proxy exposes a `process_cpu_seconds_total` metric. It also exposes
a `process_start_time_seconds` metric. But given a random metrics
snapshot, we have no way of knowing the proxy's uptime unless we know
the current time at the time of the snapshot. linkerd/linkerd2#8443
proposed adding a `process_uptime_seconds_total` metric tracking the
(fractional) number of seconds since the process started, to make these
other process metrics easier to interpret.

This branch implements the `process_uptime_seconds_total` metric.
Actually adding the metric is rather simple.

However, it occurred to me that uptimes are also used as timestamps in
the proxy's `tracing` logs. Currently, the start time used to calculate
log timestamps is just recorded when constructing the logger. If
separate `Instant`s are used as the start time for log uptime timestamps
and for the `process_uptime_seconds_total` metric, this means the
process' uptime as reported by the metrics endpoint and the process'
uptime according to log messages may not quite agree. Therefore, I
thought it would be good if a single `Instant` was used for both logging
and metrics, and I made some additional changes so that we can record a
single timestamp when the proxy starts, and pass it into both the
metrics endpoint and the logger. This does make this branch a slightly
larger change, though.

Closes linkerd/linkerd2#8443

### DIFF
--- a/linkerd/tracing/src/lib.rs
+++ b/linkerd/tracing/src/lib.rs
@@ -14,6 +14,7 @@ mod uptime;
 use self::uptime::Uptime;
 use linkerd_error::Error;
 use std::str;
+use tokio::time::Instant;
 use tracing::{Dispatch, Subscriber};
 use tracing_subscriber::{
     filter::{FilterFn, LevelFilter},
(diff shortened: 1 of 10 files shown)


linkerd/tracing/src/lib.rs
```diff
@@ -74,26 +60,28 @@ pub fn init_log_compat() -> Result<(), Error> {
...
-            if level.to_uppercase().trim() == "OFF" {
-                return None;
-            }
-        }
-
-        Some(Self {
-            filter,
+    pub fn from_env() -> Self {
+        Self {
+            filter: std::env::var(ENV_LOG_LEVEL).ok(),
```
olix0r, line 65: IIUC this is only called in one place. Should it just take an `Instant`?

olix0r (approved):

linkerd/tracing/src/lib.rs
hawkw: good call, changed it to just pass in the `Instant` there!

PR #1663, opened 2022-05-12
Fall back to A record when SRV resolution fails

adleong:
Fixes #8296

Adapted from: https://github.com/linkerd/linkerd2/issues/8296#issuecomment-1104093285 by @magec

When we do a DNS SRV lookup, if that lookup succeeds but the response has no labels, we fall back and issue a DNS A lookup instead.  However, if the initial SRV lookup fails to resolve at all, we do not fallback to an A lookup.

We change this logic to fallback to A lookups in either case.  It's not clear under what conditions a SRV lookup will fail vs returning an entry with no labels, but falling back to an A lookup seems like reasonable behavior and there is anecdotal evidence from #8296 that this resolves certain issues.

Signed-off-by:_ Alex Leong <alex@buoyant.io>

### DIFF
--- a/linkerd/dns/src/lib.rs
+++ b/linkerd/dns/src/lib.rs
@@ -70,7 +70,7 @@ impl Resolver {
     ) -> Result<(Vec<net::SocketAddr>, time::Sleep), Error> {
         match self.resolve_srv(name).await {
             Ok(res) => Ok(res),
-            Err(e) if e.is::<InvalidSrv>() => {
+            Err(e) if e.is::<ResolveError>() || e.is::<InvalidSrv>() => {
                 let (ips, delay) = self.resolve_a(name).await?;
                 let addrs = ips
                     .into_iter()


linkerd/dns/src/lib.rs
```diff
@@ -70,7 +70,7 @@ impl Resolver {
     ) -> Result<(Vec<net::SocketAddr>, time::Sleep), Error> {
         match self.resolve_srv(name).await {
             Ok(res) => Ok(res),
-            Err(e) if e.is::<InvalidSrv>() => {
+            Err(e) if e.is::<ResolveError>() || e.is::<InvalidSrv>() => {
```
hawkw, line 73: I believe the error returned by `resolve_srv` is currently going to _always_ be either a `ResolveError` (if the error was returned by `trust-dns-resolver`, or `InvalidSrv` (if the error was returned by `srv_to_socket_addr`).

The [`AsyncResolver::srv_lookup` function](https://docs.rs/trust-dns-resolver/latest/trust_dns_resolver/struct.AsyncResolver.html#method.srv_lookup) in `trust-dns-resolver` always returns `ResolveError`, and `srv_to_socket_addr` always returns `InvalidSrv`, and `resolve_srv` doesn't call any other fallible functions, so the error type is always one of these two errors.

That means we should be able to simplify this to just
```suggestion
            Err(e) => {
```
and remove the match arm on line 81 one that returns any other error from `resolve_srv`.

hawkw (changes requested):
it looks like we are now always falling back to an A record lookup no matter what error type is returned by `resolve_srv`, so i think we could simplify this code a bit more.

linkerd/dns/src/lib.rs
olix0r: Mmm, I think the explicit check is better. Otherwise, this is fragile if any of the other code changes. Otherwise, we should use an explicit error enum so that we can verify this statically.

olix0r (approved):

linkerd/dns/src/lib.rs
hawkw: :woman_shrugging: hmm...I see your point. on the other hand, I'm not a huge fan of having the catchall match arm `Err(e) => Err(e)`, because it seems like it suggests that it's possible that we won't retry some error cases, which is misleading. 

maybe we should make that arm be `unreachable!()` or something instead, or remove the guards from the match arm and add
```rust
assert(e.is::<ResolveError>() || e.is::<InvalidSrv>())
```
in the body to make this explicit?

hawkw: honestly we could also just change the error types returned from `resolve_srv` and `srv_to_socket_addr` from a boxed error to `()` or something here, since we never actually care about any information contained in the error type...

adleong: an error enum seems like the most complete solution here but I tried going down that path and it added a lot of boilerplate.  I'm hesitant to add `unreachable!()` or an assert because even though there aren't currently any other error types returned, if there were any, we would want them to return that error rather than panic.  In other words, that code path is currently unreachable, but it's not invalid.  With all this in mind, I'm inclined to leave it as-is.

hawkw: i'm not pulling for adding an error enum that we wouldn't use: we could just discard the errors entirely, which is _essentially_ what we're doing now, but with more steps...but i'm not going to block on this if no one else agrees with me.

olix0r:
@adleong @hawkw wdyt about https://github.com/linkerd/linkerd2-proxy/pull/1670

PR #1670, opened 2022-05-16
dns: Use typed errors

olix0r:
The `dns` module uses boxed errors, which makes it difficult to reason
about the types of errors we may encounter.

This change adds error newtypes so that we can at least print more
detailed information about the type of resolution that failed.

### DIFF
--- a/linkerd/app/core/src/errors.rs
+++ b/linkerd/app/core/src/errors.rs
@@ -19,3 +19,16 @@ pub fn root_cause<'e>(
     }
     error
 }
+
+/// Determines if the provided error was caused by an `E` typed error.
+pub fn caused_by<'e, E: std::error::Error + 'static>(
+    mut error: &'e (dyn std::error::Error + 'static),
+) -> Option<&'e E> {
+    while let Some(src) = error.source() {
+        if let Some(e) = src.downcast_ref::<E>() {
+            return Some(e);
+        }
+        error = src;
+    }
+    None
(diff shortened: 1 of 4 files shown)


adleong (approved):
Neat!

linkerd/app/core/src/errors.rs
```diff
@@ -19,3 +19,16 @@ pub fn root_cause<'e>(
...
+    mut error: &'e (dyn std::error::Error + 'static),
+) -> Option<&'e E> {
+    while let Some(src) = error.source() {
+        if let Some(e) = src.downcast_ref::<E>() {
+            return Some(e);
+        }
+        error = src;
+    }
+    None
+}
```
hawkw, line 34: very minor nit: i might consider putting this in `linkerd-error`, so it could later be used in other crates. also, we already have https://github.com/linkerd/linkerd2-proxy/blob/a8e4a2e3f713d7417b51cd8ffd408d481823e49c/linkerd/error/src/lib.rs#L18-L20 which i think could be replaced with this function in a few places.

(not a big deal, i just feel like we've written a bunch of variants of this function in different places...)

hawkw (approved):
this looks great, one tiny nit

linkerd/app/core/src/errors.rs
olix0r: I'll followup with a cleanup

PR #1671, opened 2022-05-16
Consolidate error chain inspection with `cause_ref`

olix0r:
We have similar/redundant code that handles inspecting (boxed) errors to
determine whether they are caused by a given error type. This change
replaces the other methods for expecting with utility functions
`is_caused_by` and `cause_ref`.

These helpers are now tested.

### DIFF
--- a/linkerd/app/core/src/errors.rs
+++ b/linkerd/app/core/src/errors.rs
@@ -1,34 +1,11 @@
 pub mod respond;
 
 pub use self::respond::{HttpRescue, NewRespond, NewRespondService, SyntheticHttpResponse};
+pub use linkerd_error::{cause_ref, is_caused_by};
 pub use linkerd_proxy_http::h2::H2Error;
 pub use linkerd_stack::FailFastError;
-use thiserror::Error;
 pub use tonic::Code as Grpc;
 
-#[derive(Debug, Error)]
+#[derive(Debug, thiserror::Error)]
 #[error("connect timed out after {0:?}")]
 pub struct ConnectTimeout(pub(crate) std::time::Duration);
-
-/// Obtain the source error at the end of a chain of `Error`s.
-pub fn root_cause<'e>(
-    mut error: &'e (dyn std::error::Error + 'static),
-) -> &'e (dyn std::error::Error + 'static) {
-    while let Some(e) = error.source() {
-        error = e;
-    }
-    error
-}
-
-/// Determines if the provided error was caused by an `E` typed error.
-pub fn caused_by<'e, E: std::error::Error + 'static>(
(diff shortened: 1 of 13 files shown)


linkerd/app/core/src/errors.rs
```diff
@@ -1,34 +1,11 @@
...
-        error = e;
-    }
-    error
-}
-
-/// Determines if the provided error was caused by an `E` typed error.
-pub fn caused_by<'e, E: std::error::Error + 'static>(
-    mut error: &'e (dyn std::error::Error + 'static),
-) -> Option<&'e E> {
-    while let Some(src) = error.source() {
```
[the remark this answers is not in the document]
hawkw: whoops!

linkerd/app/inbound/src/http/server.rs
```diff
@@ -127,23 +127,23 @@ impl<T: Param<tls::ConditionalServerTls>> ExtractParam<errors::respond::EmitHead
...
             return Ok(errors::SyntheticHttpResponse::unauthenticated(cause));
         }
-        if cause.is::<crate::GatewayLoop>() {
+        if let Some(cause) = errors::cause_ref::<crate::GatewayLoop>(&*error) {
             return Ok(errors::SyntheticHttpResponse::loop_detected(cause));
         }
-        if cause.is::<errors::FailFastError>() {
+        if let Some(cause) = errors::cause_ref::<errors::FailFastError>(&*error) {
             return Ok(errors::SyntheticHttpResponse::gateway_timeout(cause));
         }
```
hawkw, line 144: this does mean that we will now traverse the entire cause chain for _every_ separate `if` condition, rather than doing it once to find the root cause and then testing that error's type. OTTOH, this code will now handle cases where the error of a desired type is not the ultimate root cause...although i believe all of these error types are always roots.

i don't know if the performance implications of iterating over the cause chain multiple times is really worth worrying about, but i wanted to point it out — i think having a single correct implementation of this is worth more than not traversing a couple nested causes a few more times, in any case.

linkerd/error/src/lib.rs
```diff
@@ -15,6 +15,64 @@ pub type Error = Box<dyn std::error::Error + Send + Sync>;
...
+    loop {
+        if let Some(e) = error.downcast_ref::<E>() {
+            return Some(e);
+        }
+        error = error.source()?;
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
hawkw, line 46: <3

hawkw (approved):

linkerd/app/inbound/src/http/server.rs
olix0r: yeah that was my thought process as well.

PR #1676, opened 2022-05-17
opencensus: Include empty generated protobuf

olix0r:
prost generates empty files for google protobuf types. This change adds
these generated files so that they are not regenerated on `make test`.


hawkw (approved):
i was wondering where that came from, lgtm!

PR #1677, opened 2022-05-18
Dynamically discover policies for undocumented ports

olix0r:
During process startup, proxies discover policies for all documented
inbound ports. Undocumented ports use the default policy and cannot be
configured by the policy controller.

This is an annoying limitation, as Kubernetes does not otherwise require
a pod's ports to be explicitly documented.

This change lifts this limitation. We continue to eagerly
discover policies for documented ports during startup, but we now fall
back to dynamically discovering policies for undocumented ports. This
discovery does not block processing a connection. The default policy is
honored until a policy is provided by the control plane.

Fixes https://github.com/linkerd/linkerd2/issues/7640

### DIFF
--- a/linkerd/app/inbound/src/policy.rs
+++ b/linkerd/app/inbound/src/policy.rs
@@ -2,7 +2,7 @@ mod api;
 mod authorize;
 mod config;
 pub mod defaults;
-mod store;
+pub(crate) mod store;
 #[cfg(test)]
 mod tests;
 
--- a/linkerd/app/inbound/src/policy.rs
+++ b/linkerd/app/inbound/src/policy.rs
@@ -29,7 +29,7 @@ pub struct DeniedUnauthorized {
 }
 
 pub trait GetPolicy {
-    /// Checks that the destination address is configured to allow traffic.
+    // Returns the traffic policy configured for the destination address.
     fn get_policy(&self, dst: OrigDstAddr) -> AllowPolicy;
 }
 
--- a/linkerd/app/inbound/src/policy.rs
+++ b/linkerd/app/inbound/src/policy.rs
@@ -115,18 +115,23 @@ impl AllowPolicy {
         }
     }
 
+    /// Checks whether the server has any authorizations at all. If it does not,
+    /// a denial error is returned.
     pub(crate) fn check_port_allowed(&self) -> Result<(), DeniedUnauthorized> {
         let server = self.server.borrow();
+
         if server.authorizations.is_empty() {
             return Err(DeniedUnauthorized {
                 kind: server.kind.clone(),
                 name: server.name.clone(),
             });
         }
+
         Ok(())
     }
-    /// Checks whether the destination port's `AllowPolicy` is authorized to accept connections
-    /// given the provided TLS state.
+
+    /// Checks whether the destination port's `AllowPolicy` is authorized to
+    /// accept connections given the provided TLS state.
     pub(crate) fn check_authorized(
         &self,
         client_addr: Remote<ClientAddr>,
--- a/linkerd/app/inbound/src/policy.rs
+++ b/linkerd/app/inbound/src/policy.rs
@@ -1,7 +1,7 @@
+mod api;
 mod authorize;
 mod config;
 pub mod defaults;
-mod discover;
 mod store;
 #[cfg(test)]
 mod tests;
--- a/linkerd/app/inbound/src/policy.rs
+++ b/linkerd/app/inbound/src/policy.rs
@@ -2,7 +2,7 @@ mod api;
 mod authorize;
 mod config;
 pub mod defaults;
-pub(crate) mod store;
+mod store;
 #[cfg(test)]
 mod tests;
 
@@ -22,18 +22,15 @@ use thiserror::Error;
 use tokio::sync::watch;
 
 #[derive(Clone, Debug, Error)]
-#[error("unauthorized connection on unknown port {0}")]
-pub struct DeniedUnknownPort(pub u16);
-
-#[derive(Clone, Debug, Error)]
-#[error("unauthorized connection on server {server}")]
+#[error("unauthorized connection on {kind}/{name}")]
 pub struct DeniedUnauthorized {
-    server: std::sync::Arc<str>,
+    kind: std::sync::Arc<str>,
+    name: std::sync::Arc<str>,
 }
 
-pub trait CheckPolicy {
+pub trait GetPolicy {
     /// Checks that the destination address is configured to allow traffic.
-    fn check_policy(&self, dst: OrigDstAddr) -> Result<AllowPolicy, DeniedUnknownPort>;
+    fn get_policy(&self, dst: OrigDstAddr) -> AllowPolicy;
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
@@ -114,11 +111,20 @@ impl AllowPolicy {
     async fn changed(&mut self) {
         if self.server.changed().await.is_err() {
             // If the sender was dropped, then there can be no further changes.
-
             futures::future::pending::<()>().await;
         }
     }
 
+    pub(crate) fn check_port_allowed(&self) -> Result<(), DeniedUnauthorized> {
+        let server = self.server.borrow();
+        if server.authorizations.is_empty() {
+            return Err(DeniedUnauthorized {
+                kind: server.kind.clone(),
+                name: server.name.clone(),
+            });
+        }
+        Ok(())
+    }
     /// Checks whether the destination port's `AllowPolicy` is authorized to accept connections
     /// given the provided TLS state.
     pub(crate) fn check_authorized(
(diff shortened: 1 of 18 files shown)


linkerd/app/inbound/src/policy.rs
```diff
@@ -16,23 +16,21 @@ use linkerd_app_core::{
...
-#[error("unauthorized connection on unknown port {0}")]
-pub struct DeniedUnknownPort(pub u16);
-
-#[derive(Clone, Debug, Error)]
-#[error("unauthorized connection on server {server}")]
+#[error("unauthorized connection on {kind}/{name}")]
 pub struct DeniedUnauthorized {
-    server: std::sync::Arc<str>,
+    kind: std::sync::Arc<str>,
+    name: std::sync::Arc<str>,
```
hawkw, line 28: very tiny nitpick: should we maybe import `Arc`?

linkerd/app/inbound/src/policy.rs
```diff
@@ -16,23 +16,21 @@ use linkerd_app_core::{
...
 }
 
-pub trait CheckPolicy {
+pub trait GetPolicy {
     /// Checks that the destination address is configured to allow traffic.
```
hawkw, line 32: doc comment should maybe be changed to 
```suggestion
    /// Returns the allowed traffic policy configured for the destination address.
```
or similar?

linkerd/app/inbound/src/policy/config.rs
```diff
@@ -29,29 +31,23 @@ impl Config {
         dns: dns::Resolver,
         metrics: metrics::ControlHttp,
         identity: identity::NewClient,
-    ) -> Store {
+    ) -> impl GetPolicy + Clone + Send + Sync + 'static {
```
hawkw, line 34: cute way to erase the type parameter of `Store`...

linkerd/app/inbound/src/policy/store.rs
```diff
@@ -31,51 +27,46 @@ struct PortHasher(u16);
...
 
-        let store = Self {
-            default,
-            ports: Arc::new(rxs),
-        };
-        (store, default_tx)
+        let (default_tx, default_rx) = watch::channel(ServerPolicy::from(default));
+        tokio::spawn(async move {
+            default_tx.closed().await;
+        });
```
hawkw, line 49: maybe worth a comment here saying that this task's job is to keep the watch from closing if the tx is dropped?

linkerd/app/inbound/src/policy/store.rs
```diff
@@ -31,51 +27,46 @@ struct PortHasher(u16);
...
-    /// The returned future completes when a watch has been successfully created for all of the
-    /// provided ports. The store maintains these watches so that each time a policy is checked, it
-    /// may obtain the latest policy provided by the watch. An error is returned if any of the
-    /// watches cannot be established.
-    pub(super) fn spawn_discover<S>(
+    /// The returned future completes when a watch has been successfully created
+    /// for all of the provided ports. The store maintains these watches so that
+    /// each time a policy is checked, it may obtain the latest policy provided
+    /// by the watch. An error is returned if any of the watches cannot be
+    /// established.
```
hawkw, line 64: i don't think this comment is correct --- the function returns a `Store`, which is not a future, and a separate watch task is spawned for each provided port. should this be updated?

linkerd/app/inbound/src/policy/store.rs
```diff
@@ -31,51 +27,46 @@ struct PortHasher(u16);
...
                 let (_, rx) = watch::channel(s);
                 (p, rx)
-            })
-            .collect();
-
-        let (default_tx, default) = match Self::mk_default(default.into()) {
-            Some((tx, rx)) => (Some(tx), Some(rx)),
-            None => (None, None),
+            });
+            Cache::with_permanent_from_iter(Duration::MAX, rxs)
```
hawkw, line 43: hmm, i'm not sure about this...i feel like we should still pass in the idle timeout in the fixed policy case. we won't spawn a bunch of background tasks in the case where `discover` is `None`, but we'll still insert a bunch of clones of the default policy into the cache which will never go away. this means that something like a port scanner enumerating all ports will blow up the cache to 65535 entries, and it'll stay there forever...

linkerd/app/inbound/src/policy/store.rs
```diff
@@ -84,53 +75,57 @@ impl Store {
...
+        // already exist, we spawn a watch on the API (if it is configured). If
+        // no discovery API is configured we use the default policy.
+        let server =
+            self.cache
+                .get_or_insert_with(dst.port(), |port| match self.discover.clone() {
+                    Some(disco) => info_span!("watch", %port).in_scope(|| {
+                        disco.spawn_with_init(*port, self.default_rx.borrow().clone())
+                    }),
+                    None => self.default_rx.clone(),
+                });
```
hawkw, line 126: i wonder if it was actually better not to hit the cache and just return `Cached::uncached` in the case where the policy is fixed and lookups are disabled. In the fixed policy case we're _just_ cloning the default policy, but the current code here will still insert a clone of the default policy into the map for each port. even though we won't spawn a background task in the fixed case, we still put stuff into the map, so worst case, we have a map with `u16::MAX` u16s and `watch::Receiver` clones.

i [worked it out](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=405c0630fcd70f75a8d8c95617626805) and it looks like the actual data stored in such a map would max out at about 1.7MB, plus however much additional overhead a hashmap takes up.

we could avoid this if we just wrote something like this
```suggestion
        // Lookup the polcify for the target port in the cache. If it doesn't
        // already exist, we spawn a watch on the API (if it is configured). If
        // no discovery API is configured we use the default policy.
        let server = match self.discover {
            Some(ref disco) => self.cache.get_or_insert_with(dst.port(), |port| {
                info_span!("watch", %port).in_scope(|| {
                    disco.spawn_with_init(*port, self.default_rx.borrow().clone())
                })
            },
            None => Cached::uncached(self.default_rx.clone()),
        };
```

but, maybe this is a premature optimization that isn't worth using the `uncached` constructor outside of tests? the maximum memory use of the cache _is_ bounded, but it seems unfortunate to have a giant map where every value is the same thing...

linkerd/app/inbound/src/policy/store.rs
```diff
@@ -84,53 +75,57 @@ impl Store {
...
+        //
+        // Policies that are dynamically discovered at runtime will expire after
+        // `idle_timeout` to prevent holding policy watches indefinitely for
+        // ports that are generally unused.
+        let cache = {
+            let rxs = ports.into_iter().map(|port| {
+                let watch = watch.clone();
+                let default = default_rx.borrow().clone();
+                let rx =
+                    info_span!("watch", %port).in_scope(|| watch.spawn_with_init(port, default));
```
hawkw, line 93: tiny nit: `port` is a[ `u16`, which implements `tracing::Value`](https://docs.rs/tracing/latest/tracing/trait.Value.html#impl-Value-for-u16), so we don't _actually_ need to record it using `fmt::Display`:
```suggestion
                    info_span!("watch", port).in_scope(|| watch.spawn_with_init(port, default));
```

linkerd/app/inbound/src/policy/store.rs
```diff
@@ -84,53 +75,57 @@ impl Store {
...
+        http::HttpBody<Data = tonic::codegen::Bytes, Error = Error> + Default + Send + 'static,
+{
+    fn get_policy(&self, dst: OrigDstAddr) -> AllowPolicy {
+        // Lookup the polcify for the target port in the cache. If it doesn't
+        // already exist, we spawn a watch on the API (if it is configured). If
+        // no discovery API is configured we use the default policy.
+        let server =
+            self.cache
+                .get_or_insert_with(dst.port(), |port| match self.discover.clone() {
+                    Some(disco) => info_span!("watch", %port).in_scope(|| {
```
hawkw, line 122: since [`u16` implements `tracing::Value`](https://docs.rs/tracing/latest/tracing/trait.Value.html#impl-Value-for-u16), we don't need `Display` here
```suggestion
                    Some(disco) => info_span!("watch", port).in_scope(|| {
```

hawkw (commented):
overall, this looks good to me. 

i had some concerns on the most recent changes to avoid the use of `Cached::uncached` in non-test code, but i'm open to being convinced that i'm wrong about it. 

i also commented on a few minor nits, and it _would_ be nice to have a test that watches are actually torn down when evicted from the cache, but it's not a hard blocker for me.

linkerd/app/inbound/src/policy/store.rs
olix0r: We don't really care about the fixed policy case being efficient. I'll document that. We expect proxies to always run with policy discovery, basically. I'm really trying to avoid making `Cached::uncached` a thing if at all possible :)

hawkw (approved):
looks good to me

linkerd/app/inbound/src/policy/store.rs
hawkw: yeah...i guess as long as it's not a configuration that will use an *unbounded* amount of memory, it doesn't really matter...

PR #1682, opened 2022-05-18
Simplify admin server feature-flagging in #1642

olix0r:
* Rather than push feature flagging down into the `serve` logic, we can
  avoid even matching the logs.json endpoint when the feature isn't
  disabled. This avoids the need for conditional feature gatingc in the
  handler.
* The log module is split into two submodules: level and stream. The
  stream module is feature-gated. This allows to extract function-local
  defintions into the module, which simplifies scoping and makes things
  easier to read (imo).
* merge main


linkerd/app/admin/src/server/log/stream.rs
```diff
@@ -0,0 +1,160 @@
...
+                }))?;
+                return Ok(mk_rsp($status, json));
+            }
+        }
+    }
+}
+
+// If log streaming support is enabled, start a new log stream
+pub async fn serve<B>(
+    handle: trace::Handle,
```
olix0r, line 34: Changed the type here to avoid taking a generic parameter. Cloning the handle should be cheap (and this isn't really allocation-sensitive)

hawkw (approved):
yeah, this feels a bit neater --- thanks!

PR #1683, opened 2022-05-18
cache: generalize `Cache` into a key-value cache

hawkw:
Currently, the `linkerd-cache` crate contains an implementation of a
cache where entries are evicted if they have not been accessed for a
timeout duration. This type currently implements the `NewService` trait
and requires that cached values are `Service`s. It does not provide a
more general purpose key-value cache interface.

PR #1677 introduces dynamic discovery for inbound port policies. We
would like to implement a similar eviction strategy for these policy
watches. However, the policy watches are *not* used in a `NewService`
implementation, but in a `CheckPolicy` implementation.

Therefore, this branch rewrites the `Cache` type to serve as a
general-purpose key-value cache, and adds a new `NewCachedService` type
that implements `NewService` on top of `Cache`.

### DIFF
--- a/linkerd/cache/src/lib.rs
+++ b/linkerd/cache/src/lib.rs
@@ -5,7 +5,7 @@ use parking_lot::RwLock;
 use std::{
     borrow::Borrow,
     collections::{hash_map::{Entry, RandomState}, HashMap},
-    hash::{BuildHasher, Hash},
+    hash::{BuildHasher, Hash, BuildHasherDefault},
     ops::Deref,
     sync::{Arc, Weak},
     task::{Context, Poll},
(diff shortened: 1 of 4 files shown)


linkerd/cache/src/lib.rs
```diff
@@ -1,86 +1,180 @@
...
+    }
+
+    pub fn get<'a, Q: ?Sized>(&self, key: &Q) -> Option<Cached<V>>
+    where
+        K: Borrow<Q>,
+        Q: Hash + Eq,
+        V: Clone,
+    {
+        let lock = self.inner.read();
+        let (value, weak) = lock.get(&key)?;
```
olix0r, line 98:
```suggestion
        let (value, weak) = lock.get(key)?;
```

linkerd/cache/src/lib.rs
```diff
@@ -1,86 +1,180 @@
...
+    K: Clone + std::fmt::Debug + Eq + Hash + Send + Sync + 'static,
+    V: Send + Sync + 'static,
+    S: BuildHasher + Send + Sync + 'static,
+{
+    pub fn with_hasher(idle: time::Duration, hasher: S) -> Self {
+        let inner = Arc::new(RwLock::new(HashMap::with_hasher(hasher)));
+        Self { inner, idle }
+    }
+
+    pub fn get<'a, Q: ?Sized>(&self, key: &Q) -> Option<Cached<V>>
```
olix0r, line 91:
```suggestion
    pub fn get<Q: ?Sized>(&self, key: &Q) -> Option<Cached<V>>
```

olix0r (commented):
note that there are a few warnings:

hawkw:
@olix0r I merged #1685 into this PR (after some manual cleanup). WDYT?

olix0r (approved):

PR #1684, opened 2022-05-18
evict inbound dynamic policy discovery watches

hawkw:
This branch adds idle eviction for the dynamic policy lookups for
inbound ports added in #1677. Idle lookups are evicted by replacing the
`HashMap` of ports to watches with the the `Cache` type from
`linkerd-cache`, which tracks when a policy watch is in use, and removes
them from the cache when they have been idle for too long. PR #1683
modifies `Cache` to allow it to be used as a general-purpose key-value
cache in addition to a `NewService` implementation, so this branch
depends on #1683.


hawkw:
I'd like to also add some tests asserting that the policy watches are actually completely torn down on eviction, but AFAICT they _should_ be...

linkerd/app/inbound/src/policy.rs
```diff
@@ -44,7 +45,7 @@ pub enum DefaultPolicy {
 #[derive(Clone, Debug)]
 pub struct AllowPolicy {
     dst: OrigDstAddr,
-    server: watch::Receiver<ServerPolicy>,
+    server: Cached<watch::Receiver<ServerPolicy>>,
```
olix0r, line 48: This is a little unfortunate. Could we get away with doing something like `T: ToOwned<Owned = watch::Receiver<ServerPolicy>>`?

linkerd/app/inbound/src/policy/store.rs
```diff
@@ -76,21 +77,19 @@ impl<S> Store<S> {
...
-            })
-            .collect::<PortMap<_>>();
+        let rxs = ports.into_iter().map(|port| {
+            let watch = watch.clone();
+            let default = default.clone();
+            let rx =
+                info_span!("watch", %port).in_scope(|| watch.spawn_with_init(port, default.into()));
+            (port, rx)
+        });
+        let ports = Cache::from_iter(idle_timeout, rxs);
```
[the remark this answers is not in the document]
hawkw: okay, we can make these initial values non-expiring.

linkerd/app/inbound/src/policy.rs
hawkw: Yeah...I thought about something like that. I think it would be `Borrow<watch::Receiver<ServerPolicy>>` or `AsRef<...>` instead of `ToOwned`, though — a `ToOwned` impl would have to clone the receiver out of the `Cached` handle, so the receiver could be held even while the cache access is considered over...

hawkw:
@olix0r updated based on your suggestions, plmkwyt!

hawkw:
@olix0r now that #1683 has merged, I've rebased your PR #1677 onto `main` and this branch onto #1677. This should now only contain changes to your branch, and not all of the cache changes.

linkerd/app/inbound/src/policy/config.rs
```diff
@@ -52,7 +53,10 @@ impl Config {
                     let c = control.build(dns, metrics, identity).new_service(());
                     Api::new(workload, c).into_watch(backoff)
                 };
-                Store::spawn_discover(default, ports, watch)
+                // TODO(eliza): should the idle timeout for policy discovery be
+                // configurable? One minute is totally arbitrary...
+                let idle_timeout = Duration::from_secs(60);
```
hawkw, line 58: @olix0r do you think this seems reasonable or should we change it?

olix0r: I think we should use the `cache_max_idle_age` from the top-level config

hawkw: sounds good, will do!

PR #1685, opened 2022-05-19
Store permanent values in the cache (for #1683)

olix0r:
In some cases--like inbound policy discovery--we have a set of permanent
values that should never be removed from the cache. This change updates
the cache to only hold a handle when eviction is scheduled. No handle is
held when the cache entry is permanent.

This change also fixes a possible race condition. Previously, the
eviction task could consume an entry's handle without locking the cache.
This made it possible for a `get_or_insert_with` call to race against
the eviction task, adding an entry that would be immediately evicted by
the eviction task.

### DIFF
--- a/linkerd/cache/src/lib.rs
+++ b/linkerd/cache/src/lib.rs
@@ -360,7 +360,7 @@ async fn test_idle_retain() {
 
     let handle = cache.spawn_idle(());
     let weak = Arc::downgrade(&handle);
-    cache.inner.write().insert((), ((), Some(weak.clone())));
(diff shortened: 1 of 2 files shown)


hawkw:
Oh, whoops, I missed this and made my own almost-identical version of the permanent entry part of the change: https://github.com/linkerd/linkerd2-proxy/pull/1683/commits/3525055bf5306c12869696fcde9fb2f0391ffbcd

I'm going to merge this anyway to pick up the race fix --- good catch.

PR #1689, opened 2022-05-20
Don't allow a policy to be used if `check_port_allowed` fails

olix0r:


hawkw (approved):
looks good to me!

just to confirm, this doesn't actually fix a bug, right? AFAICT, the current code *is* correct, because an error is always returned anyway, and this change just makes it harder to misuse by *ensuring* this is the case?

olix0r:
@hawkw yep!

PR #1691, opened 2022-05-22
dev: Replace `Makefile` with `justfile`

olix0r:
Other projects in the ecosystem (like kube-rs) are ditching Make's
obtuse syntax for a newer tool called [`just`][just].

This change removes our `Makefile` with a simpler `justfile`. The
release workflow is updated to use `just`.

While doing this, a new version of the devcontainer has been published
with `just`, `cargo-nextest`, and various dependency updates.

[just]: https://just.systems/man/


hawkw (approved):
looks good to me --- i've never used `just` before but the syntax seems _pretty_ obvious.

PR #1698, opened 2022-05-23
http: Strip illegal headers from CONNECT responses

olix0r:

Fixes linkerd/linkerd2#8539

### DIFF
--- a/linkerd/app/integration/src/tests/transparency.rs
+++ b/linkerd/app/integration/src/tests/transparency.rs
@@ -709,6 +709,96 @@ macro_rules! http1_tests {
             proxy.join_servers().await;
         }
 
+        /// Tests that the proxy drops headers that can't be used in CONNECT responses.
+        ///
+        /// Exercises https://tools.ietf.org/html/rfc7231#section-4.3.6:
+        ///
+        /// A client MUST ignore any Content-Length or Transfer-Encoding header
+        /// fields received in a successful response to CONNECT.
+        #[tokio::test]
+        async fn http11_connect_headers_stripped() {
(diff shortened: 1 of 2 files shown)


linkerd/app/integration/src/tests/transparency.rs
```diff
@@ -709,6 +709,96 @@ macro_rules! http1_tests {
...
+                        let n = sock.read(&mut vec).await?;
+                        assert_eq!(s(&vec[..n]), s(&tunneled_req[..]));
+
+                        // Some processing... and then write back tunneled res...
+                        sock.write_all(&tunneled_res[..]).await
+                    }
+                    .map(|res: std::io::Result<()>| match res {
+                        Ok(()) => {}
+                        Err(e) => panic!("tcp server error: {}", e),
+                    })
```
hawkw, line 768: tiny nit, TIOLI: i believe this could avoid matching the `Ok(())` case if it used the [`unwrap_or_else`combinator](https://docs.rs/futures/latest/futures/future/trait.TryFutureExt.html#method.unwrap_or_else) rather than `map`:
```suggestion
                    .unwrap_or_else(|e| panic!("tcp server error: {}", e))
```

hawkw (approved):
looks good to me! tiny nit on the test

linkerd/app/integration/src/tests/transparency.rs
olix0r: this is copy/paste from above. would love help cleaning up the tests in a separate PR

olix0r:
ugh this was based on the wrong branch :/. will resubmit

PR #1699, opened 2022-05-23
http: Strip illegal headers from CONNECT responses

olix0r:

Fixes linkerd/linkerd2#8539


olix0r:
Reviewed/approved in https://github.com/linkerd/linkerd2-proxy/pull/1698

PR #1704, opened 2022-05-24
ci: Run all CI workflows on all changes

olix0r:
We currently can't require that CI completes successfully because
actions may not necessarily run on a given change.

This change uses `tj-actions/changed-files` to detect whether files
change and then conditionally runs CI jobs depending on whether files
changed.


olix0r:
Unfortunately this is probably a bit more cumbersome than we'd like: we'd have to put conditions on every _step_ for the jobs to appear successful.

.github/workflows/fuzzers.yml
```diff
@@ -72,18 +71,20 @@ jobs:
...
+    needs: [changes]
+    if: needs.changes.outputs.changed
     timeout-minutes: 40
     runs-on: ubuntu-latest
     container:
       image: docker://rust:1.60.0-bullseye
     strategy:
       matrix:
-        dir: ${{ fromJson(needs.list-changed.outputs.dirs) }}
+        dir: ${{ fromJson(needs.changes.outputs.dirs) }}
```
hawkw, line 87: oh this is clever!

hawkw (approved):
looks good to me!

PR #1705, opened 2022-05-24
test: replace `profile_test!` macro with builder

hawkw:
Currently, the service profiles integration test uses a very complex
declarative macro to generate test bodies. This approach has some
downsides:

1. All of the test code is contained inside a macro. This means that
   `rustfmt` will not reformat it, and (in some cases) compilation
   errors can be unclear.
2. The macro-generated code is relatively inflexible. This makes it
   difficult to add new tests that need to do slightly different things.
3. Adding new configurations to the macro is complex. If we want to add
   some new thing that can also be passed in to change the generated
   code, we have to modify a bunch of macro match arms, which is
   annoying and time-consuming.

This branch rewrites the profiles integration tests to use a
builder-style design instead of a macro. This way, the repetitive
boilerplate necessary to set up the tests is still factored out, and can
still be overridden by passing in different parameters to the builder,
but instead of having a giant macro generate the test, the test itself
is just normal Rust code.

The builder spins up the test proxy and servers, and returns the test
client and metrics client to the test itself, and the test can now do
whatever it wants with them, rather than having to pass bits of test
code into a macro. This is much nicer, because the ests are just normal
tests now. Additionally, the builder can accept an arbitrary server
configuration, which will be necessary for some new tests I'm working on
in a separate branch.

I've also changed the profiles integration test so that most of the
tests are run with both HTTP/1 *and* HTTP/2 clients and servers. This is
still accomplished using macros, but now, the macro is much simpler. The
tests themselves are factored out into functions that take either an
HTTP/1 or HTTP/2 server, and all the macro does is generate wrapper
functions that call the actual test function with the appropriate server
version. This way, the actual test code is never inside of a macro
expansion.

I'd like to change more of the integration tests to work similarly to
these, removing other cases where a macro is used to generate test
boilerplate, and replacing it with a builder.  Also, and more importantly,
this change is necessary for some additional tests I'll be adding in a 
separate PR.

### DIFF
--- a/linkerd/app/integration/src/tests/profiles.rs
+++ b/linkerd/app/integration/src/tests/profiles.rs
@@ -168,9 +168,8 @@ impl TestBuilder {
     }
 }
 
-#[tokio::test]
-async fn retry_if_profile_allows() {
-    let test = TestBuilder::new(server::http1())
+async fn test_retry_if_profile_allows(version: server::Server) {
+    let test = TestBuilder::new(version)
         .with_profile_route(
             controller::route()
                 .request_any()
@@ -183,9 +182,8 @@ async fn retry_if_profile_allows() {
     assert_eq!(test.client.get("/0.5").await, "retried");
 }
 
-#[tokio::test]
-async fn retry_uses_budget() {
-    let test = TestBuilder::new(server::http1())
+async fn test_retry_uses_budget(version: server::Server) {
+    let test = TestBuilder::new(version)
         .with_profile_route(
             controller::route()
                 .request_any()
@@ -216,9 +214,8 @@ async fn retry_uses_budget() {
         .await;
 }
 
-#[tokio::test]
-async fn retry_with_small_post_body() {
-    let test = TestBuilder::new(server::http1())
+async fn test_retry_with_small_post_body(version: server::Server) {
+    let test = TestBuilder::new(version)
         .with_profile_route(
             controller::route()
                 .request_any()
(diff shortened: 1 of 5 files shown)


linkerd/app/integration/src/tests/profiles.rs
```diff
@@ -1,123 +1,142 @@
...
-                    // client doesn't fail when writing it out.
-                    let _body = hyper::body::to_bytes(req.into_body()).await;
-                    tracing::debug!(body = ?_body.as_ref().map(|body| body.len()), "recieved body");
-                    Ok::<_, Error>(if fail {
+                Response::builder().status(201).body("".into()).unwrap()
+            });
+        if self.default_routes {
+            let counter = AtomicUsize::new(0);
+            let counter2 = AtomicUsize::new(0);
+            let counter3 = AtomicUsize::new(0);
```
olix0r, line 68: I know this isn't new code but it's a bit weird that these endpoints alternate between failure and success without any explicit control

linkerd/app/integration/src/tests/profiles.rs
```diff
@@ -132,334 +151,427 @@ macro_rules! profile_test {
...
+            )
+            .label("classification", "failure")
+            .label("error", "timeout")
+            .value(1u64)
+            .assert_in(&test.metrics)
+            .await;
+    }
+}
+
+macro_rules! cross_version {
```
olix0r, line 507: perhaps better named `version_tests`?

linkerd/app/integration/src/tests/profiles.rs
```diff
@@ -132,334 +151,427 @@ macro_rules! profile_test {
...
-                .label("error", "timeout")
-                .value(1u64)
-                .assert_in(&metrics)
-                .await;
-        }
+mod http2 {
+    use super::*;
+
+    cross_version! {
+        server::http1() =>
```
olix0r, line 543: should this be http2?

olix0r (approved):

PR #1706, opened 2022-05-24
retry gRPC requests are immediately terminated by trailers

hawkw:
Currently, gRPC responses are classified for metrics based on the value
of the `grpc-status` header in either the initial `HEADERS` frame *or*
in the `TRAILERS` frame that terminates the response body. However, this
is *not* the case for retries. Because we don't want to buffer the
entire response body before forwarding it to the client (introducing
latency, and potentially breaking client behavior in the case of
long-running streams), we make retry decisions for gRPC based only on
`grpc-status` in the initial `HEADERS` frame.

Unfortunately, some gRPC implementations (such as Akka) will send the
`grpc-status` header in a `TRAILERS` frame even when the body is empty
(see linkerd/linkerd2#7701). In that case, Linkerd won't retry the
request, even though it wouldn't actually have to wait to buffer a long
body before making a retry decision.

This branch resolves this issue by changing the retry implementation so
that the proxy will wait for a single additional frame after the first
`HEADERS` frame before making a retry decision. If the next frame is a
`TRAILERS` frame that terminates the response stream, we can include the
value of the `grpc-status` in that `TRAILERS` frame in the retry
decision. Otherwise, if the next frame is a `DATA` frame, we buffer only
a single frame, give up on trying to retry, and forward that frame to
the client, followed by the rest of the body.

Unfortunately, this change is a bit complicated, as it was necessary to
introduce the ability for retries to modify the response body as well as
the request body. This then implies that we need to be able to erase
response body types as well, when they may differ based on whether or
not retries are enabled for a route. I tried a number of different ways
of structuring this change, such as passing an additional `Layer` into
the `Retry` service that wraps the service constructed for retries, but
this was kind of a mess. Ultimately, I felt like the cleanest way of
doing this was changing the `linkerd_retry::PrepareRequest` trait to a
`PrepareRetry` trait that can also modify response bodies.

I'm open to suggestions for a nicer way of structuing this change. I
also considered the alternative of giving up on `tower::retry` entirely
and writing our own retry service from scratch, but I wanted to try to
avoid that if it was possible.

Depends on #1705 and #1726.

Fixes linkerd/linkerd2#7701.


linkerd/stack/src/proxy.rs
```diff
@@ -39,3 +51,45 @@ where
...
+                StateProj::PollReady { proxy, svc } => {
+                    futures::ready!(svc.poll_ready(cx))?;
+                    let f = proxy.proxy(svc, this.req.take().expect("already called"));
+                    this.state.as_mut().set(State::Future(f));
+                }
+                StateProj::Future(f) => return f.poll(cx),
+            }
+        }
+    }
+}
```
hawkw, line 95: happy to factor this change out into a smaller branch that can merge independently

linkerd/http-retry/src/with_trailers.rs
```diff
@@ -0,0 +1,154 @@
...
+            inner: body,
+            first_data: None,
+            trailers: None,
+        };
+
+        if let Some(data) = body.inner.data().await {
+            // body has data; stop waiting for trailers
+            body.first_data = Some(data?);
+
+            // peek to see if there's immediately a trailers frame, and grab
```
olix0r, line 76: nit: sentence capitalization

```suggestion
            // Peek to see if there's immediately a trailers frame, and grab
```

linkerd/http-retry/src/with_trailers.rs
```diff
@@ -0,0 +1,154 @@
...
+            // it if so. otherwise, bail.
+            if let Some(trailers) = body.inner.trailers().now_or_never() {
+                body.trailers = trailers?;
+            }
+
+            return Ok(http::Response::from_parts(parts, body));
+        }
+
+        // okay, let's see if there's trailers...
+        body.trailers = body.inner.trailers().await?;
```
olix0r, line 86: here, too.

linkerd/http-retry/src/with_trailers.rs
```diff
@@ -0,0 +1,154 @@
...
+        let (parts, body) = rsp.into_parts();
+        let mut body = Self {
+            inner: body,
+            first_data: None,
+            trailers: None,
+        };
+
+        if let Some(data) = body.inner.data().await {
+            // body has data; stop waiting for trailers
+            body.first_data = Some(data?);
```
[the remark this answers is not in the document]
hawkw: oh, that's a good point. i'll change that.

hawkw:
@olix0r I've made the error handling changes you suggested --- thanks for that, good catch! Also, I did a bit of cleaning up the `WithTrailers` implementation to make it clearer that it can't eat additional data frames.

linkerd/http-retry/src/with_trailers.rs
```diff
@@ -70,22 +89,25 @@ impl<B: Body> WithTrailers<B> {
...
+            body.first_data = Some(data);
+
+            // Peek to see if there's immediately a trailers frame, and grab
+            // it if so. Otherwise, bail.
+            // XXX(eliza): the documentation for the `http::Body` trait says
+            // that `poll_trailers` should only be called after `poll_data`
+            // returns `None`...but, in practice, I'm fairly sure that this just
+            // means that it *will not return `Ready`* until there are no data
+            // frames left, which is fine for us here, because we `now_or_never`
+            // it.
```
olix0r, line 102: maybe worth a clarification on `http_body`...

olix0r (approved):
LGTM!

PR #1716, opened 2022-05-27
justfile: generate a default docker tag name

hawkw:
Currently, when running `just docker`, the image is not given a
meaningful tag name by default. This commit changes the `justfile` so
that the default tag is something useful, rather than `''`. If no tag
name is provided from the command line, the Docker image will be given
the tag `gchr.io/linkerd-io/proxy:$USER-$GIT_SHA`. The image name of the
actual Linkerd repository is used so that dev images can be selected
using only the `--proxy-version` flag of `linkerd inject`, and
`--proxy-image` is not needed.

If we'd prefer not to output the same image name as release images by
default, I could that part to something different.

### DIFF
--- a/justfile
+++ b/justfile
@@ -22,7 +22,7 @@ features := ""
 package_version := `git rev-parse --short HEAD`
 
 # Docker image name & tag.
-docker_repo := "localhost/linked/proxy"
+docker_repo := "localhost/linkerd/proxy"
 docker_tag := `git rev-parse --abbrev-ref HEAD | sed 's|/|.|'` + "." + `git rev-parse --short HEAD`
 docker_image := docker_repo + ":" + docker_tag
 
--- a/justfile
+++ b/justfile
@@ -20,7 +20,7 @@ cargo := "cargo" + if toolchain != "" { " +" + toolchain } else { "" }
 package_version := `git rev-parse --short HEAD`
 
 # Default docker tag name
-default_docker_tag := "gchr.io/linkerd-io/proxy:" + env_var_or_default("USER", "dev") + "-" + package_version
+default_docker_tag := env_var_or_default("USER", "dev") + "/l2-proxy:" + package_version
 
 # The architecture name to use for packages. Either 'amd64', 'arm64', or 'arm'.
 package_arch := "amd64"
--- a/justfile
+++ b/justfile
@@ -19,6 +19,9 @@ cargo := "cargo" + if toolchain != "" { " +" + toolchain } else { "" }
 # The version name to use for packages.
 package_version := `git rev-parse --short HEAD`
 
+# Default docker tag name
+default_docker_tag := "gchr.io/linkerd-io/proxy:" + env_var_or_default("USER", "dev") + "-" + package_version
+
 # The architecture name to use for packages. Either 'amd64', 'arm64', or 'arm'.
 package_arch := "amd64"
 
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -19,6 +19,9 @@ cargo := "cargo" + if toolchain != "" { " +" + toolchain } else { "" }
 # The version name to use for packages.
 package_version := `git rev-parse --short HEAD`
 
+# Default docker tag name
+default_docker_tag := "gchr.io/linkerd-io/proxy:" + env_var_or_default("USER", "dev") + "-" + package_version
```
olix0r, line 23: What is `linkerd-io`?

I chose not to do something like this originally because we definitely do not want to push these images to the container registry.

Perhaps better to do something like
```suggestion
default_docker_tag := env_var_or_default("USER", "dev") + "/l2-proxy:" + package_version
```

Also, AFAICT this variable is never used. Maybe we should stop taking the tag as a recipe argument and instead always use the variable?

hawkw: Sure, I'm fine with using a local image tag that differs from the published one, I wasn't sure what the preference would be.

hawkw: > Also, AFAICT this variable is never used. Maybe we should stop taking the tag as a recipe argument and instead always use the variable?

agh, that's my fault, i must have messed up as a rebase. it was passed as the default value for the `tag` argument in the `docker` recipe. i would also be fine with just always using the variable, since currently, if you want to pass other arguments to `just docker` (such as a mode), you would have to override the tag name as well...

justfile
```diff
@@ -19,6 +19,9 @@ cargo := "cargo" + if toolchain != "" { " +" + toolchain } else { "" }
 # The version name to use for packages.
 package_version := `git rev-parse --short HEAD`
 
+# Default docker tag name
+default_docker_tag := env_var_or_default("USER", "dev") + "/l2-proxy:" + package_version
+
```
olix0r, line 24:
```suggestion
# Docker tag
docker_tag := env_var_or_default("USER", "dev") + "/l2-proxy:" + package_version

```

justfile
```diff
@@ -151,7 +154,7 @@ fuzzers:
     done
 
 # Build a docker image (FOR TESTING ONLY)
-docker tag='' mode='load':
+docker tag=default_docker_tag mode='load':
     docker buildx build . \
         {{ if build_type != 'release' { "--build-arg PROXY_UNOPTIMIZED=1" } else { "" } }} \
         {{ if tag != "" { "--tag=" + tag + " " + (if mode == "push" { "--push" } else { "--load" }) } else { "" } }}
```
olix0r, line 160:
```suggestion
docker mode='load':
    docker buildx build . \
        --tag={{ tag }} \
        {{ if build_type != 'release' { "--build-arg PROXY_UNOPTIMIZED=1" } else { "" } }} \
        {{ if mode == "push" { "--push" } else { "--load" } }}
```

olix0r (commented):
Now that the docker tag is always set, there's no use to take the tag as an argument or to have conditional logic

justfile
```diff
@@ -19,6 +19,9 @@ cargo := "cargo" + if toolchain != "" { " +" + toolchain } else { "" }
 # The version name to use for packages.
 package_version := `git rev-parse --short HEAD`
 
+# Docker tag
+docker_tag := env_var_or_default("USER", "dev") + "/l2-proxy:" + package_version
```
hawkw, line 23: another thought...I wonder if we want to include the branch name for this as well, since it's not intended to be used for actual releases?
```suggestion
docker_tag := env_var_or_default("USER", "dev") + "/l2-proxy:" + `git rev-parse --abbrev-ref HEAD` + "." package_version
```
or similar?

olix0r: i'm pretty sure tags cannot have `/` in them, so we'd need to sed out illegal chars

olix0r (approved):

PR #1725, opened 2022-06-01
stack: add `Proxy::proxy_oneshot`

hawkw:
Currently, `ServiceExt::oneshot` is used to take a `Service` by value,
poll it until it's ready, and then call that service. This composes well
with `Service` middleware, but it does not compose with our `Proxy`
trait. If a `Service` must be wrapped in a `Proxy` call, it is
impossible to oneshot it.

This branch adds a new `Proxy::proxy_oneshot` method to `Proxy` that
takes a `Proxy` and a `Service` by value, and returns a `Future` that
drives the `Service` to readiness and then calls it through that
`Proxy`.

Currently, this is unused, but it will be used in PR #1706 (from which
this change was factored out).

### DIFF
--- a/linkerd/stack/src/proxy.rs
+++ b/linkerd/stack/src/proxy.rs
@@ -19,6 +19,18 @@ pub trait Proxy<Req, S: tower::Service<Self::Request>> {
 
     /// Usually invokes `S::call`, potentially modifying requests or responses.
     fn proxy(&self, inner: &mut S, req: Req) -> Self::Future;
+
+    /// Like [`ServiceExt::oneshot`](tower::util::ServiceExt::oneshot), but with
+    /// a `Proxy` too.
+    fn proxy_oneshot(self, svc: S, req: Req) -> Oneshot<Self, S, Req>
+    where
+        Self: Sized,
+    {
+        Oneshot {
+            req: Some(req),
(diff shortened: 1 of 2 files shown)


linkerd/stack/src/proxy.rs
```diff
@@ -19,6 +19,18 @@ pub trait Proxy<Req, S: tower::Service<Self::Request>> {
 
     /// Usually invokes `S::call`, potentially modifying requests or responses.
     fn proxy(&self, inner: &mut S, req: Req) -> Self::Future;
+
+    /// Like [`ServiceExt::oneshot`](tower::util::ServiceExt::oneshot), but with
+    /// a `Proxy` too.
+    fn proxy_oneshot(self, svc: S, req: Req) -> Oneshot<Self, S, Req>
```
olix0r, line 25: I'm not necessarily opposed to this but I'm having a little trouble understanding the use case. The whole point of a proxy is to take a service by-reference. It sounds like  what we really want is to fuse the proxy to a service in this case?

olix0r:
Can this be closed in favor of #1726?

hawkw:
> Can this be closed in favor of #1726?

Yes, should have just closed it.

PR #1726, opened 2022-06-02
stack: add `ProxyService` for composing a `Proxy` with a `Service`

hawkw:
This branch adds a `Proxy::into_service` method that composes a `Proxy`
with a `Service`, returning a new `Service` that calls the inner
`Service` through that `Proxy`.

This negates the need for `Proxy::proxy_oneshot`, so this also
closes #1725.

Currently, this is unused, but it will be used in PR #1706 (from which
this change was factored out).


olix0r (commented):
I thought we already had this but maybe it was removed at some point. Note that there are several docs warnings

olix0r (approved):

PR #1759, opened 2022-06-18
Dedupe discovery updates

olix0r:
The proxy can receive redundant discovery updates. When this occurs, it
causes the balancer to churn, replacing an endpoint stack (and
therefore needlessly dropping a connection).

This change updates the discovery module to keep clones of the
discovered endpoint metadata, so that updated values can be compared to
eliminate duplicate updates.

Relates to linkerd/linkerd2#8677

### DIFF
--- a/linkerd/proxy/discover/src/from_resolve.rs
+++ b/linkerd/proxy/discover/src/from_resolve.rs
@@ -184,11 +184,11 @@ where
 #[cfg(test)]
 mod tests {
     use super::Discover;
-    use async_stream::stream;
     use futures::prelude::*;
     use linkerd_error::Infallible;
     use linkerd_proxy_core::resolve::Update;
     use std::net::SocketAddr;
+    use tokio_stream::wrappers::ReceiverStream;
     use tower::discover::Change;
 
     const PORT: u16 = 8080;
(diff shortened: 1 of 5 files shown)


linkerd/proxy/discover/src/from_resolve.rs
```diff
@@ -111,50 +113,70 @@ where
...
                     }
-                },
-                None => return Poll::Ready(None),
+                }
+
+                Update::DoesNotExist => {
+                    trace!("Scheduling removals");
+                    this.pending
+                        .extend(this.active.drain(..).map(|(sa, _)| Change::Remove(sa)));
+                    this.active.clear();
```
hawkw, line 178: i don't *think* the `clear` call should be necessary here, since the `drain(..)` iterator should remove everything from `this.pending`? if this actually is load-bearing, i would kind of like a comment explaining why, otherwise, we could just remove it.

linkerd/proxy/discover/src/from_resolve.rs
```diff
@@ -111,50 +113,70 @@ where
...
+                                .push_back(Change::Insert(*addr, endpoint.clone()));
                         }
                     }
-                    Update::Remove(addrs) => {
-                        for addr in addrs.into_iter() {
-                            if this.active.remove(&addr) {
-                                trace!(%addr, "Scheduling removal");
-                                this.pending.push_back(Change::Remove(addr));
+
+                    *this.active = new_active;
```
hawkw, line 144: spent a minute trying to figure out if there was a nice way to do this without creating a new map and dropping the old one...but i think this approach is probably the best, so :+1:

linkerd/proxy/discover/src/from_resolve.rs
```diff
@@ -163,50 +185,112 @@ where
...
+
+        // Reset to a new state with 3 addresses, one of which is unchanged, one of which is
+        // changed, and one of which is added.
+        tx.try_send(Ok(Update::Reset(
+            // Restore the original `2` value. We shouldn't see an update for it.
+            Some((addr(2), 2))
+                .into_iter()
+                // Set a new value for `3`. and add a 4th as well.
+                .chain((3..=4).map(|n| (addr(n), n + 1)))
+                .collect(),
```
hawkw, line 225: tbh, i find this a little weird, it might be clearer to just write
```suggestion
            vec![
                // Restore the original `2` value. We shouldn't see an update for it.
                (addr(2), 2),
                // Set a new value for `3`.
                (addr(3), 4),
                // Add a fourth endpoint as well.
                (addr(4), 5),
            ]
```

linkerd/proxy/discover/src/from_resolve.rs
```diff
@@ -163,50 +185,112 @@ where
...
+            ));
         }
+
+        // No more updates.
+        drop(tx);
         assert!(disco.next().await.is_none());
     }
+
+    #[tokio::test(flavor = "current_thread")]
+    async fn dedupe() {
```
hawkw, line 247: tiny, silly nit: generally i think i've seen this spelt "dedup" (it's not spelled "dupelicate"...)

linkerd/proxy/discover/src/from_resolve.rs
```diff
@@ -31,7 +32,7 @@ pub struct DiscoverFuture<F, E> {
 pub struct Discover<R: TryStream, E> {
     #[pin]
     resolution: R,
-    active: HashSet<SocketAddr>,
+    active: IndexMap<SocketAddr, E>,
```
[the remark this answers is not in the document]
olix0r: it's mostly so that the tests are predictable. noted in comment

PR #1765, opened 2022-06-20
dev: Update build settings

olix0r:
* Update the devcontainer so that rust-analyzer uses the same
  `RUSTFLAGS` as our `justfile`. This helps to eliminate needless
  recompilation.
* Support `features=all` to enable all features.
* Update test targets to use nextest if it is available.
* Update all build targets to run `fmt` first.
* Add `clippy-dir` and `test-dir` directives that accept a path instead
  of a crate name.


justfile
```diff
@@ -60,7 +60,21 @@ _fmt := if env_var_or_default("GITHUB_ACTIONS", "") != "true" { "" } else {
...
+_features := if features == "all" {
+        "--all-features"
+    } else if features != "" {
+        "--no-default-features --features=" + features
+    } else { "" }
+
+
+# Use nextest if it's available.
+_test := ```
+        if command -v cargo-nextest >/dev/null 2>&1; then
```
hawkw, line 72: nit, take it or leave it: you can also detect if a cargo subcommand is available using `cargo --list`; this _might_ be more robust than checking if a `cargo-nextest` executable is on the `PATH`?

.devcontainer/devcontainer.json
```diff
@@ -11,6 +11,11 @@
 		"tamasfe.even-better-toml",
 		"zxh404.vscode-proto3"
 	],
+	"settings": {
+		"rust-analyzer.server.extraEnv": {
+			"RUSTFLAGS": "-D warnings -A deprecated"
```
hawkw, line 16: i'm...not totally sure how i feel about setting `-D warnings` in dev builds? i know the `justfile` already does it (https://github.com/linkerd/linkerd2-proxy/blob/f3cf2e8b8556502bfbf7138d9b26517ff0fd6afe/justfile#L8) but i'm not sure if we should be doing that; i would like to be able to run tests from the justfile even if the proxy builds with warnings, while working on local dev branches. maybe we should change this to only add `-D warnings` if on CI? idk.

in any case, that should probably be addressed separately; if we *are* going to set `-D warnings` in `just test`, we should do it in rust-analyzer as well.

hawkw (approved):

justfile
```diff
@@ -4,8 +4,7 @@
 # Configuration
 #
 
-export RUST_BACKTRACE := "short"
-export RUSTFLAGS := env_var_or_default("RUSTFLAGS", "-D warnings -A deprecated")
```
hawkw, line 8: TIOLI: the `justfile` already [detects whether we're on github actions](https://github.com/linkerd/linkerd2-proxy/blob/45d6cf386f71cf284d238e35f5ff149d535cbdf3/justfile#L54), so we could alternatively just set this based on whether the `GITHUB_ACTIONS` env var is set or not? that way, we wouldn't have to stick the env var in every CI config. but, on the other hand, it's maybe more explicit to have all the CI workflows explicitly state that they disable warnings? i dunno...

hawkw (approved):

PR #1774, opened 2022-06-22
inbound: Add route authorization labels

olix0r:
This change adds `route_{kind,group,name}` labels for inbound HTTP
authorization metrics. In this first implementation, a default route is
used for all metrics. In followup changes these labels will vary based
on control plane responses; and they will be wired into additional
inbound HTTP metrics.

Distinct error and permit types are introduced so that TCP-only
types are differentiated.

### DIFF
--- a/linkerd/app/admin/src/stack.rs
+++ b/linkerd/app/admin/src/stack.rs
@@ -52,7 +52,7 @@ struct Http {
 
 #[derive(Clone, Debug)]
 struct Permitted {
-    permit: inbound::policy::ServerPermit,
+    permit: inbound::policy::RoutePermit,
     http: Http,
 }
 
@@ -115,7 +115,7 @@ impl Config {
                         // - If we received some unexpected SNI, the client is mostly likely
                         //   confused/stale.
                         Err(_timeout) => {
-                            let version = match tcp.tls.clone() {
+                            let version = match tcp.tls {
                                 tls::ConditionalServerTls::None(_) => http::Version::Http1,
                                 tls::ConditionalServerTls::Some(tls::ServerTls::Established {
                                     ..
@@ -291,7 +291,7 @@ impl<T: Param<tls::ConditionalServerTls>> ExtractParam<errors::respond::EmitHead
 
 impl errors::HttpRescue<Error> for Rescue {
     fn rescue(&self, error: Error) -> Result<errors::SyntheticHttpResponse> {
-        if let Some(cause) = errors::cause_ref::<inbound::policy::DeniedUnauthorized>(&*error) {
+        if let Some(cause) = errors::cause_ref::<inbound::policy::HttpRouteUnauthorized>(&*error) {
             return Ok(errors::SyntheticHttpResponse::permission_denied(cause));
         }
 
--- a/linkerd/app/admin/src/stack.rs
+++ b/linkerd/app/admin/src/stack.rs
@@ -52,7 +52,7 @@ struct Http {
 
 #[derive(Clone, Debug)]
 struct Permitted {
-    permit: inbound::policy::RoutePermit,
+    permit: inbound::policy::HTTPRoutePermit,
     http: Http,
 }
 
--- a/linkerd/app/inbound/src/http/router.rs
+++ b/linkerd/app/inbound/src/http/router.rs
@@ -16,7 +16,7 @@ use tracing::{debug, debug_span};
 pub struct Http {
     addr: Remote<ServerAddr>,
     settings: http::client::Settings,
-    permit: policy::RoutePermit,
+    permit: policy::HTTPRoutePermit,
 }
 
 /// Builds `Logical` targets for each HTTP request.
@@ -24,7 +24,7 @@ pub struct Http {
 struct LogicalPerRequest {
     server: Remote<ServerAddr>,
     tls: tls::ConditionalServerTls,
-    permit: policy::RoutePermit,
+    permit: policy::HTTPRoutePermit,
     labels: tap::Labels,
 }
 
@@ -36,7 +36,7 @@ struct Logical {
     addr: Remote<ServerAddr>,
     http: http::Version,
     tls: tls::ConditionalServerTls,
-    permit: policy::RoutePermit,
+    permit: policy::HTTPRoutePermit,
     labels: tap::Labels,
 }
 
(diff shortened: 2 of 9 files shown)


linkerd/app/inbound/src/policy.rs
```diff
@@ -52,10 +56,15 @@ pub struct AllowPolicy {
...
     pub protocol: Protocol,
-
     pub labels: ServerAuthzLabels,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct RoutePermit {
+    pub dst: OrigDstAddr,
+    pub labels: RouteAuthzLabels,
+}
```
hawkw, line 66: might be nice to have comments on these explaining the relationship between `ServerPermit` and `RoutePermit`? nbd though.

linkerd/app/inbound/src/http/router.rs
```diff
@@ -246,47 +246,32 @@ impl<C> Inbound<C> {
...
+            ("srv", &permit.labels.route.server.0),
+            ("route", &permit.labels.route.route),
+            ("authz", &permit.labels.authz),
+        ]
+        .into_iter()
+        .flat_map(|(k, v)| {
+            [
+                (format!("{k}_group"), v.group().to_string()),
+                (format!("{k}_kind"), v.kind().to_string()),
+                (format!("{k}_name"), v.name().to_string()),
```
hawkw, line 265: i was kind of bummed out that we had to `to_string`-ify all the meta bits here where they could be `Arc`ed and just implement `FormatLabels` everywhere else...but i realized this is because of the current way `tap::Labels` works. might be worth messing with that separately i guess...

linkerd/app/inbound/src/metrics/authz.rs
```diff
@@ -35,48 +37,62 @@ pub(crate) struct TcpAuthzMetrics(Arc<TcpInner>);
...
+    tls: tls::ConditionalServerTls,
+    labels: RouteLabels,
+}
+
+#[derive(Debug, Hash, PartialEq, Eq)]
+struct RouteAuthzKey {
+    target: TargetAddr,
+    tls: tls::ConditionalServerTls,
+    labels: RouteAuthzLabels,
+}
```
hawkw, line 77: kinda wondered if it would make sense to do something like
```rust
#[derive(Debug, Hash, PartialEq, Eq)]
struct Key<T> {
    target: TargetAddr,
    tls: tls::ConditionalServerTls,
    labels: T,
}

type ServerAuthzKey = Key<ServerAuthzLabels>;
type RouteKey = Key<RouteLabels>;
type RouteAuthzKey = Key<RouteAuthzLabels>;
```
but, not sure if it's worth it.

linkerd/app/inbound/src/policy/http.rs
```diff
@@ -22,24 +24,41 @@ use std::task;
...
+
+#[derive(Clone, Debug)]
+struct ConnectionMeta {
+    dst: OrigDstAddr,
+    client: Remote<ClientAddr>,
+    tls: tls::ConditionalServerTls,
 }
 
+#[derive(Debug, thiserror::Error)]
+#[error("unauthorized request on route kind={:?} name={:?}", .0.kind(), .0.name())]
```
hawkw, line 48: tiny nits: this formats the `kind` and `name` strings using `fmt::Debug` rather than `fmt::Display`, so they'll be quoted and have special characters escaped...do we actually want them to be quoted strings? 

i notice that the `ServerUnauthorized` error formats the server meta differently:
https://github.com/linkerd/linkerd2-proxy/blob/15951c535c6d2cc734cc189adff51964749a69cb/linkerd/app/inbound/src/policy.rs#L33

should we match that here?
```suggestion
#[error("unauthorized request on route {}/{}", .0.kind(), .0.name())]
```

linkerd/app/inbound/src/policy.rs
```diff
@@ -149,51 +148,55 @@ impl AllowPolicy {
...
+            tls::ConditionalServerTls::Some(tls::ServerTls::Established {
+                client_id: Some(tls::server::ClientId(ref id)),
+                ..
+            }) => {
+                identities.contains(id.as_str()) || suffixes.iter().any(|s| s.contains(id.as_str()))
+            }
+            _ => false,
+        },
+    }
+}
```
hawkw, line 197: factoring this out seems good :+1:

linkerd/app/inbound/src/http/server.rs
```diff
@@ -127,9 +127,10 @@ impl<T: Param<tls::ConditionalServerTls>> ExtractParam<errors::respond::EmitHead
 
 impl errors::HttpRescue<Error> for ServerRescue {
     fn rescue(&self, error: Error) -> Result<errors::SyntheticHttpResponse> {
-        if let Some(cause) = errors::cause_ref::<crate::policy::DeniedUnauthorized>(&*error) {
+        if let Some(cause) = errors::cause_ref::<policy::HttpRouteUnauthorized>(&*error) {
```
hawkw, line 130: does this also want to handle the `ServerUnauthorized` error, or will that never make it here?

linkerd/app/inbound/src/policy/http.rs
```diff
@@ -90,32 +110,90 @@ where
...
+            route: self.default_route_meta.clone(),
+            server: labels,
+        };
+
+        let permit = match Self::check_authorized(
+            &*server.authorizations,
+            &self.meta,
+            labels,
+            &self.metrics,
+        ) {
```
hawkw, line 127: unimportant style nit: is there a reason `check_authorized` can't just take `&self`, rather than having to call it with `&self.meta` and `&self.metrics`?  or is `self` mutably borrowed somewhere i'm missing?

linkerd/app/admin/src/stack.rs
```diff
@@ -291,7 +291,7 @@ impl<T: Param<tls::ConditionalServerTls>> ExtractParam<errors::respond::EmitHead
 
 impl errors::HttpRescue<Error> for Rescue {
     fn rescue(&self, error: Error) -> Result<errors::SyntheticHttpResponse> {
-        if let Some(cause) = errors::cause_ref::<inbound::policy::DeniedUnauthorized>(&*error) {
+        if let Some(cause) = errors::cause_ref::<inbound::policy::HttpRouteUnauthorized>(&*error) {
```
hawkw, line 294: does this also need to handle the `ServerUnauthorized` error or will we never construct an `HttpRescue` in that case?

hawkw (approved):
overall, this looks good to me! i had a few questions & tiny style suggestions, but besides the question about whether the `ServerUnauthorized` error needs to be handled in `HttpRescue` impls, i didn't notice anything major.

linkerd/app/inbound/src/policy/http.rs
olix0r: i left them quoted in case they're empty. In practice, I don't think this error type will actually get rendered, but I think we can expect that both kind and name won't actually be empty so this change is probably fine.

linkerd/app/inbound/src/http/server.rs
olix0r: we shouldn't ever hit that now, since HTTP authorization decisions are made per-request and not at the server level.

linkerd/app/inbound/src/metrics/authz.rs
olix0r: Yeah this cleans up a lot actually

PR #1780, opened 2022-06-23
inbound: Record policy metrics for opaque-transport connections

olix0r:
When the inbound receives a "direct" inbound connection--a connection
that targets the proxy's inbound port--it uses the "opaque transport"
wrapper to discover the actual destination port for the connection. In
this case, the direct stack does a one-off authorization check instead
of using the proper TCP policy layer, which means that these connections
don't get the proper authorization metrics.

This change updates the direct TCP stack to use the TcpPolicy middleware
to enforce authorization. It also removes the authorization checking
utilities from the public `AllowPolicy` type so that all authorizations
must go through the metrics-tracking middlewares. The `policy` tests are
moved under the `policy::tcp` module so they have access to these
functions.

The TCP accept stack no longer performs eager authorization denials when
a default-deny policy is used, as these failures bypassed the metrics
code also.

### DIFF
--- a/linkerd/app/inbound/src/direct.rs
+++ b/linkerd/app/inbound/src/direct.rs
@@ -157,8 +157,7 @@ impl<N> Inbound<N> {
                                 } => {
                                     // When the transport header targets an alternate port (but does
                                     // not identify an alternate target name), we check the new
-                                    // target's policy to determine whether the client can access
-                                    // it.
+                                    // target's policy (rather than the inbound proxy's address).
                                     let addr = (client.local_addr.ip(), port).into();
                                     let policy = policies.get_policy(OrigDstAddr(addr));
                                     let local = match protocol {
@@ -189,8 +188,7 @@ impl<N> Inbound<N> {
                                 } => {
                                     // When the transport header provides an alternate target, the
                                     // connection is a gateway connection. We check the _gateway
-                                    // address's_ policy to determine whether the client is
-                                    // authorized to use this gateway.
+                                    // address's_ policy (rather than the target address).
                                     let policy = policies.get_policy(client.local_addr);
                                     Ok(svc::Either::B(GatewayTransportHeader {
                                         target: NameAddr::from((name, port)),
--- a/linkerd/app/inbound/src/direct.rs
+++ b/linkerd/app/inbound/src/direct.rs
@@ -1,4 +1,7 @@
-use crate::{policy, Inbound};
+use crate::{
+    policy::{self, AllowPolicy},
+    Inbound,
+};
 use linkerd_app_core::{
     identity, io,
     proxy::http,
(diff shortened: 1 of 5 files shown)


linkerd/app/inbound/src/direct.rs
```diff
@@ -120,6 +128,16 @@ impl<N> Inbound<N> {
...
+                .check_new_service::<AuthorizedLocalTcp, _>()
+                .push_map_target(|(permit, tcp): (policy::ServerPermit, LocalTcp)| {
+                    AuthorizedLocalTcp {
+                        addr: tcp.server_addr,
+                        client_id: tcp.client_id,
+                        permit,
+                    }
+                })
+                .check_new_service::<(policy::ServerPermit, LocalTcp), _>()
+                .push(policy::NewTcpPolicy::layer(rt.metrics.tcp_authz.clone()))
```
hawkw, line 140: IMO it feels kind of like it would be nicer to me if the state transition from `LocalTcp` -> `AuthorizedLocalTcp` happened _inside_ of the `NewTcpPolicy` layer...but, i guess trying to abstract over types that can be turned into an authorized target type by adding a `ServerPermit` with some kind of trait is probably not worth the complexity, so, this is fine IMO...

linkerd/app/inbound/src/direct.rs
```diff
@@ -182,9 +191,7 @@ impl<N> Inbound<N> {
                                     // connection is a gateway connection. We check the _gateway
                                     // address's_ policy to determine whether the client is
```
hawkw, line 192: does this comment need to be updated to reflect that we don't actually check it here? i guess the wording is ambiguous enough that maybe this is fine, but it might be a little clearer.

maybe just changing it to "we *will* check..." is sufficient?

hawkw: there's another comment here that maybe should also be changed: https://github.com/linkerd/linkerd2-proxy/blob/debb9cb802e43d6ea946a97428f61cea6a73d505/linkerd/app/inbound/src/direct.rs#L158-L161
but i can't leave an inline comment on that one in the diff.

linkerd/app/inbound/src/policy/tcp.rs
```diff
@@ -169,3 +176,22 @@ where
         }))
     }
 }
+
+/// Checks whether the destination port's `AllowPolicy` is authorized to
+/// accept connections given the provided TLS state.
+fn check_authorized(
+    server: &ServerPolicy,
+    dst: OrigDstAddr,
```
hawkw, line 184: nit, not a big deal: it kinda feels like this should just take an `&AllowPolicy` instead of two pieces of it (the `ServerPolicy` and orig dst)... it seems like the only reason it doesn't is because we want to borrow the `ServerPolicy` for the tracing event, but...we could just move that into this function?

hawkw: or maybe it's like that for the tests? in which case, no worries. not a big deal regardless.

hawkw (approved):
looks good to me! i commented on a couple very minor nits, but no blockers!

linkerd/app/inbound/src/policy/tcp.rs
olix0r: yeah, i want to test this function without all of the extra watch stuff.

PR #1781, opened 2022-06-24
Introduce per-route authorization policies

olix0r:
This change adds per-route authorization policies to the inbound proxy.
This change does NOT introduce an API client to configure these
policies. That will be added in follow-up changes, as will richer
per-route configurations.

This change:

* Modifies the inbound policy `Protocol` type so that each protocol
  variant includes policy configuration. TCP protocols now include
  authorization policies (moved from the `Server` type).
* Implements an HTTP policy router that determines the route for each
  request and applies its authorization policy.
* Fails requests for which no route matches with a `404 Not Found`
  response. A new metric, `inbound_http_route_not_found_total`, is
  introduced to count occurrences of this error.
* Includes a default route for all inbound HTTP communication, using the
  server's authorization policies.


linkerd/app/inbound/src/policy/http.rs
```diff
@@ -81,16 +81,30 @@ where
...
+}
+
+macro_rules! try_fut {
+    ($e:expr) => {
+        match $e {
+            Ok(v) => v,
+            Err(e) => err!(e),
+        }
+    };
+}
```
hawkw, line 106: occurs to me that there's probably some code elsewhere in the proxy that could be cleaned up with these...

linkerd/app/inbound/src/policy/http/tests.rs
```diff
@@ -0,0 +1,230 @@
...
+            dst: OrigDstAddr(([192, 168, 3, 4], 8080).into()),
+            client: Remote(ClientAddr(([192, 168, 3, 3], 30120).into())),
+            tls: tls::ConditionalServerTls::Some(tls::ServerTls::Established {
+                client_id: Some("foo.bar.bah".parse().unwrap()),
+                negotiated_protocol: None,
+            }),
+        };
+        new_svc!($proto, conn)
+    }};
+}
```
hawkw, line 51: nit, TIOLI: this kind of feels like it _could_ just be a pair of functions rather than a macro, but i'm not sure if it's really a big deal either way...

linkerd/grpc-route/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
```
hawkw, line 1: am i losing my mind or does this crate not have a `lib.rs`? i can't seem to find it in the diff...

hawkw (commented):
i read through this a couple times and left some comments, mostly on minor style nits. this seems good overall!

i am a bit confused about the `linkerd-grpc-route` crate added in this PR, which appears to be missing a `lib.rs` file? i'm surprised that compiles...it looks like it wasn't added to the list of crates in the workspace's `Cargo.toml`, which i _think_ is the only reason we're not getting a build failure. was that crate intended to be added in this PR, or did it accidentally slip in from a different change or something?

linkerd/app/inbound/src/metrics/authz.rs
```diff
@@ -109,6 +127,19 @@ impl FmtMetrics for HttpAuthzMetrics {
...
+        let route_not_found = self.0.route_not_found.lock();
+        if !route_not_found.is_empty() {
+            inbound_http_route_not_found_total.fmt_help(f)?;
+            inbound_http_route_not_found_total.fmt_scopes(
+                f,
+                route_not_found
+                    .iter()
+                    .map(|(k, c)| ((k.target, (&k.labels, TlsAccept(&k.tls))), c)),
+                |c| c,
+            )?;
```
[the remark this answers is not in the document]
olix0r: that would be inconsistent with the rest of the function. I'm not opposed to cleaning that up, but let's make that separate/consistent.

linkerd/grpc-route/Cargo.toml
olix0r: crate should have been removed

linkerd/app/inbound/src/policy/http/tests.rs
olix0r: Yeah, but I don't want to have to name the response type.

hawkw: oh, missed that. yeah, let's keep the macro then!

hawkw (approved):
okay, this looks good to me whenever you feel like it's ready!

PR #1786, opened 2022-06-24
inbound: Cleanup in preparation for route policies #1781

olix0r:
Before making changes that support inbound HTTP route authorization
policies, this change makes some cosmetic changes that will help
minimize the changes needed for new funcationality.

No funtional changes.

### DIFF
--- a/linkerd/app/inbound/src/policy/store.rs
+++ b/linkerd/app/inbound/src/policy/store.rs
@@ -120,6 +120,7 @@ where
             self.cache
                 .get_or_insert_with(dst.port(), |port| match self.discover.clone() {
                     Some(disco) => info_span!("watch", port).in_scope(|| {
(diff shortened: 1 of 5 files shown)


linkerd/app/inbound/src/policy/store.rs
```diff
@@ -120,14 +120,18 @@ where
             self.cache
                 .get_or_insert_with(dst.port(), |port| match self.discover.clone() {
                     Some(disco) => info_span!("watch", port).in_scope(|| {
+                        tracing::trace!(%port, "spawning policy discovery");
```
hawkw, line 123: if `trace` is enabled, the port _should_ already be present on the span, which is `info`, right?

hawkw (approved):
looks good to me!

PR #1787, opened 2022-06-24
Add HTTP route matchers to support the Gateway API

olix0r:
This change adds a standalone utility crate that will support matching
routes for the Gateway API's HTTP route types. This crate is not yet
used. In followup changes:

* filter types will be added to support header-rewriting, HTTP
  redirection, and error injection filters;
* the inbound proxy will be updated to supcort route-oriented
  authorization policies; and
* the inbound policy API client will be updated to configure inbound
  server routes.

### DIFF
--- a/linkerd/http-route/src/grpc/match.rs
+++ b/linkerd/http-route/src/grpc/match.rs
@@ -27,7 +27,9 @@ pub(crate) struct MatchRpc {
 /// Summarizes a matched gRPC endpoints.
 #[derive(Clone, Debug, Default, Hash, PartialEq, Eq, PartialOrd, Ord)]
 pub(crate) struct RpcMatch {
+    /// The number of characters matched in the service name.
     service: usize,
+    /// The number of characters matched in the method name.
     method: usize,
 }
 
--- a/linkerd/http-route/src/grpc/match.rs
+++ b/linkerd/http-route/src/grpc/match.rs
@@ -36,7 +36,7 @@ pub(crate) struct RpcMatch {
 impl crate::Match for MatchRoute {
     type Summary = RouteMatch;
 
-    fn r#match<B>(&self, req: &http::Request<B>) -> Option<RouteMatch> {
+    fn match_request<B>(&self, req: &http::Request<B>) -> Option<RouteMatch> {
         if req.method() != http::Method::POST {
             return None;
         }
@@ -64,11 +64,9 @@ impl std::cmp::PartialOrd for RouteMatch {
 
 impl std::cmp::Ord for RouteMatch {
     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
-        use std::cmp::Ordering;
-        match self.rpc.cmp(&other.rpc) {
-            Ordering::Equal => self.headers.cmp(&other.headers),
-            ord => ord,
-        }
+        self.rpc
+            .cmp(&other.rpc)
+            .then_with(|| self.headers.cmp(&other.headers))
     }
 }
 
--- a/linkerd/http-route/src/grpc/match.rs
+++ b/linkerd/http-route/src/grpc/match.rs
@@ -1,0 +1,104 @@
+#[cfg(test)]
+mod tests;
+
+use crate::http::MatchHeader;
+
+/// Matches gRPC routes.
+#[derive(Clone, Debug, Default, Hash, PartialEq, Eq)]
+pub struct MatchRoute {
+    pub(crate) rpc: MatchRpc,
+    pub(crate) headers: Vec<MatchHeader>,
+}
+
+/// Summarizes a matched gRPC route.
+#[derive(Clone, Debug, Default, Hash, PartialEq, Eq)]
+pub struct RouteMatch {
+    rpc: RpcMatch,
+    headers: usize,
+}
+
+/// Matches gRPC endpoints.
+#[derive(Clone, Debug, Default, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) struct MatchRpc {
+    pub(crate) service: Option<String>,
+    pub(crate) method: Option<String>,
+}
+
+/// Summarizes a matched gRPC endpoints.
+#[derive(Clone, Debug, Default, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) struct RpcMatch {
+    service: usize,
+    method: usize,
+}
+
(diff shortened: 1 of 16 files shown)


linkerd/http-route/src/grpc/match.rs
```diff
@@ -0,0 +1,104 @@
...
+pub(crate) struct MatchRpc {
+    pub(crate) service: Option<String>,
+    pub(crate) method: Option<String>,
+}
+
+/// Summarizes a matched gRPC endpoints.
+#[derive(Clone, Debug, Default, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) struct RpcMatch {
+    service: usize,
+    method: usize,
```
hawkw, line 31: IMO it would be nice if it was documented somewhere that the match results are represented as the length of the prefix that was matched?

linkerd/http-route/src/grpc/tests.rs
```diff
@@ -0,0 +1,171 @@
...
+        .uri("http://foo.example.com/foo/bar")
+        .body(())
+        .unwrap();
+    let (_, policy) = find(&rts, &req).expect("must match");
+    assert_eq!(*policy, Policy::Expected, "incorrect rule matched");
+}
+
+#[test]
+fn method_precedence() {
+    // Given two equivalent routes, choose the longer path match.
```
hawkw, line 60: very silly nit: most of the tests in this file have doc comments outside of the test, but this one has a non-doc-comment inside of the test. let's change this to
```suggestion
/// Given two equivalent routes, choose the longer path match.
#[test]
fn method_precedence() {
```
for consistency?

linkerd/http-route/src/http/match/host.rs
```diff
@@ -0,0 +1,205 @@
...
+            Self::Suffix(suffix) => {
+                if suffix.first().map(|s| &**s) != Some("") {
+                    host = host.strip_suffix('.').unwrap_or(host);
+                }
+                let mut length = 0;
+                for sfx in suffix.iter() {
+                    if !host.ends_with(sfx) {
+                        return None;
+                    }
+                    host = &host[..host.len() - sfx.len()];
```
hawkw, line 75: nit, take it or leave it: i think this is equivalent to
```suggestion
                    host = host.strip_suffix(sfx)?;
```

linkerd/http-route/src/http/match/host.rs
```diff
@@ -0,0 +1,205 @@
...
+                    if !host.ends_with('.') {
+                        return None;
+                    }
+                    host = &host[..host.len() - 1];
```
hawkw, line 79: and this is
```suggestion
                    host = host.strip_suffix('.')?;
```

linkerd/http-route/src/http/match/path.rs
```diff
@@ -0,0 +1,153 @@
...
+            }
+
+            Self::Regex(re) => {
+                if let Some(m) = re.find(uri.path()) {
+                    let len = uri.path().len();
+                    // Check that the regex is anchored at the start and end of
+                    // the value.
+                    if m.start() == 0 && m.end() == len {
+                        return Some(PathMatch::Regex(len));
+                    }
```
hawkw, line 36: nit: could we change this so that we always construct anchored regexen, and then just use `Regex::is_match` instead? the `regex` crate's docs suggest that `is_match` may be more efficient than `find`.

linkerd/http-route/src/http/match/query_param.rs
```diff
@@ -0,0 +1,96 @@
...
+        uri.query().map_or(false, |qs| {
+            url::form_urlencoded::parse(qs.as_bytes()).any(|(q, p)| match self {
+                Self::Exact(n, v) => *n == *q && *v == *p,
+                Self::Regex(n, r) => {
+                    if *n == *q {
+                        if let Some(m) = r.find(&*p) {
+                            // Check that the regex is anchored at the start and
+                            // end of the value.
+                            return m.start() == 0 && m.end() == p.len();
+                        }
```
hawkw, line 23: similarly, it would be nice IMO if we ensured all regexes are anchored and then used `Regex::is_match`...not a blocker, though.

linkerd/http-route/src/http/match.rs
```diff
@@ -0,0 +1,104 @@
...
+        match self.path_match.cmp(&other.path_match) {
+            Ordering::Equal => match self.headers.cmp(&other.headers) {
+                Ordering::Equal => match self.query_params.cmp(&other.query_params) {
+                    Ordering::Equal => self.method.cmp(&other.method),
+                    ord => ord,
+                },
+                ord => ord,
+            },
+            ord => ord,
+        }
```
hawkw, line 102: nit, take it or leave it: this feels like a use-case for `Ordering::then_with`:
```suggestion
        self.path_match
            .cmp(&other.path_match)
            .then_with(|| self.headers.cmp(&other.headers))
            .then_with(|| self.query_params.cmp(&other.query_params))
            .then_with(|| self.method.cmp(&other.method))
```

linkerd/http-route/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+pub struct RouteMatch<T> {
+    host: Option<http::HostMatch>,
+    route: T,
+}
+
+/// A strategy for matching a request to a route.
+pub trait Match {
+    type Summary: Default + Ord;
+
+    fn r#match<B>(&self, req: &::http::Request<B>) -> Option<Self::Summary>;
```
hawkw, line 53: this is maybe silly, but there is a part of me that feels like maybe we should just rename this to `find_match` or something like that rather than having to type `r#match`? :woman_shrugging:

hawkw (approved):
overall, this looks good to me! i was a bit confused about the meaning of the match summaries at first, but then i realized i should have started with `lib.rs`, which GitHub had to put at the end of the review --- this makes sense given the docs there, so i don't think any additional docs are needed.

i left a few minor suggestions, but :+1:

linkerd/http-route/src/http/match/path.rs
olix0r: > could we change this so that we always construct anchored regexen

While I think I could get something that would anchor some regexes, I'm not confident that my solution wouldn't have some unexpected corner cases. This solution is at least simple and relatively obvious--it seems unlikely there could be false positives,at least. We can always improve this later if we find it's worth the effort.

PR #1788, opened 2022-06-24
inbound: Add HTTP route filters

olix0r:
This change adds several route filters:

* An HTTP header rewriting;
* An HTTP redirection filter;
* An HTTP error filter; and
* A gRPC error filter.

The header-rewriting and HTTP redirection filters are implemented as
specified by the Kubernetes Gateway API HTTPRoute spec.

These filters apply only on inbound policies are not yet configurable
via the control plane. A followup change will be made to support
configuring these filters. Depends on #1781.

### DIFF
--- a/linkerd/app/inbound/src/policy/http.rs
+++ b/linkerd/app/inbound/src/policy/http.rs
@@ -1,4 +1,4 @@
-use super::Routes;
+use super::{RoutePolicy, Routes};
 use crate::{
     metrics::authz::HttpAuthzMetrics,
     policy::{AllowPolicy, HttpRoutePermit},
@@ -11,7 +11,8 @@ use linkerd_app_core::{
     transport::{ClientAddr, OrigDstAddr, Remote},
     Error, Result,
 };
-use std::task;
+use linkerd_server_policy::{grpc, http, route::RouteMatch, Meta as RouteMeta};
+use std::{sync::Arc, task};
 
 /// A middleware that enforces policy on each HTTP request.
 ///
@@ -45,6 +46,26 @@ struct ConnectionMeta {
 #[derive(Debug, thiserror::Error)]
 #[error("no route found for request")]
 pub struct HttpRouteNotFound(());
+
+#[derive(Debug, thiserror::Error)]
+#[error("invalid redirect: {0}")]
+pub struct HttpRouteInvalidRedirect(#[from] pub http::filter::InvalidRedirect);
+
+#[derive(Debug, thiserror::Error)]
+#[error("request redirected to {}", .0.location)]
+pub struct HttpRouteRedirect(pub http::filter::Redirection);
+
+#[derive(Debug, thiserror::Error)]
+#[error("API indicated an HTTP error response: {}: {}", .0.status, .0.message)]
+pub struct HttpRouteErrorResponse(pub http::filter::RespondWithError);
+
+#[derive(Debug, thiserror::Error)]
+#[error("API indicated an gRPC error response: {}: {}", .0.code, .0.message)]
+pub struct GrpcRouteErrorResponse(pub grpc::filter::RespondWithError);
+
+#[derive(Debug, thiserror::Error)]
+#[error("unknown filter type in route: {} {} {}", .0.group(), .0.kind(), .0.name())]
+pub struct HttpRouteUnknownFilter(Arc<RouteMeta>);
 
 #[derive(Debug, thiserror::Error)]
 #[error("unauthorized request on route")]
@@ -121,15 +142,23 @@ where
         task::Poll::Ready(Ok(()))
     }
 
-    fn call(&mut self, req: ::http::Request<B>) -> Self::Future {
+    fn call(&mut self, mut req: ::http::Request<B>) -> Self::Future {
         // Find an appropriate route for the request and ensure that it's
         // authorized.
         //
         // TODO Apply filters...
         let permit = match self.policy.routes() {
             None => err!(self.mk_route_not_found()),
-            Some(Routes::Http(routes)) => try_fut!(self.authorize(&routes, &req)),
-            Some(Routes::Grpc(routes)) => try_fut!(self.authorize(&routes, &req)),
+            Some(Routes::Http(routes)) => {
+                let (permit, r#match, route) = try_fut!(self.authorize(&routes, &req));
+                try_fut!(Self::apply_http_filters(r#match, route, &mut req));
+                permit
+            }
+            Some(Routes::Grpc(routes)) => {
+                let (permit, _, route) = try_fut!(self.authorize(&routes, &req));
+                try_fut!(Self::apply_grpc_filters(route, &mut req));
+                permit
+            }
         };
 
         future::Either::Left(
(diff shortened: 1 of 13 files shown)


linkerd/app/inbound/src/policy/http.rs
```diff
@@ -221,3 +248,64 @@ impl<T, N> HttpPolicyService<T, N> {
...
+                Ok(Some(redirection)) => {
+                    return Err(HttpRouteRedirect(redirection).into());
+                }
+
+                Err(invalid) => {
+                    return Err(HttpRouteInvalidRedirect(invalid).into());
+                }
+
+                Ok(None) => {
+                    tracing::debug!("Ignoring irrelvant redirect");
```
hawkw, line 274: typo: 
```suggestion
                    tracing::debug!("Ignoring irrelevant redirect");
```

linkerd/http-route/src/http/filter/redirect.rs
```diff
@@ -0,0 +1,100 @@
...
+                .scheme
+                .clone()
+                .or_else(|| orig_uri.scheme().cloned())
+                .unwrap_or(Scheme::HTTP);
+
+            let authority: Authority = match (self.host.as_deref(), self.port) {
+                // If a host is configured, use it and whatever port is configured.
+                (Some(h), p) => p
+                    .map(|p| format!("{}:{}", h, p).parse())
+                    .unwrap_or_else(|| h.parse())?,
```
hawkw, line 57: seems like we could potentially avoid doing this for every request, in the case that there's a port configured --- this authority doesn't depend on the request at all, unlike in the case where the redirect is configured without a host...

linkerd/http-route/src/http/filter/redirect.rs
```diff
@@ -0,0 +1,100 @@
...
+                // If a host is NOT configured, use the request's original host and either an
+                // overridden port or the original port.
+                (None, p) => {
+                    let h = orig_uri.host().ok_or(InvalidRedirect::MissingAuthority)?;
+                    p.or_else(|| orig_uri.port_u16())
+                        .map(|p| format!("{}:{}", h, p).parse())
+                        .unwrap_or_else(|| h.parse())?
```
hawkw, line 64: also, we may want to use the `TryFrom<String>` impl for `Authority` rather than `FromStr`? the `TryFrom<String>` impl will [try to construct an `Authority` from the `String` converted into a `Bytes`](https://docs.rs/http/latest/src/http/uri/authority.rs.html#454-461) while [the `FromStr` impl calls the `TryFrom<&'a str>` impl](https://docs.rs/http/latest/src/http/uri/authority.rs.html#463-469), which [will copy the `&str`'s bytes into a new `Bytes`](https://docs.rs/http/latest/src/http/uri/authority.rs.html#425-443). since we have ownership of the strings returned by `format!` here, we may want to use the conversion that takes the `String` by value, as it won't allocate a new `Bytes` and copy the string into it, but instead [uses the same allocation](https://docs.rs/bytes/1.1.0/src/bytes/bytes.rs.html#836-840).

linkerd/http-route/src/http/filter/redirect.rs
```diff
@@ -0,0 +1,100 @@
...
+                use crate::http::r#match::PathMatch;
+
+                let orig_path = orig_uri.path();
+                match &self.path {
+                    None => orig_path.to_string(),
+                    Some(ModifyPath::ReplaceFullPath(p)) => p.clone(),
+                    Some(ModifyPath::ReplacePrefixMatch(new_pfx)) => match rm.route.path() {
+                        PathMatch::Prefix(pfx_len) if *pfx_len <= orig_path.len() => {
+                            let (_, rest) = orig_path.split_at(*pfx_len);
+                            format!("{}{}", new_pfx, rest)
```
hawkw, line 78: it might be a little bit more efficient to do something like
```suggestion
                            let mut path = String::with_capacity(new_pfx.len() + rest.len());
                            path.push_str(new_pfx);
                            path.push_str(rest);
                            path
```

but, that's also several more lines of code...

linkerd/http-route/src/http/filter/redirect.rs
```diff
@@ -0,0 +1,100 @@
...
+                }
+            };
+
+            let path = {
+                use crate::http::r#match::PathMatch;
+
+                let orig_path = orig_uri.path();
+                match &self.path {
+                    None => orig_path.to_string(),
+                    Some(ModifyPath::ReplaceFullPath(p)) => p.clone(),
```
hawkw, line 74: if we stored the full path as a `Bytes` rather than a `String`, the clone here would just be a reference count bump, rather than an alloc + memcpy (and `http::uri::Builder::path_and_query` will accept a `Bytes`). we would also have to convert the modified path in that case, though.

linkerd/http-route/src/http/filter/redirect.rs
```diff
@@ -0,0 +1,100 @@
...
+                    Some(ModifyPath::ReplacePrefixMatch(new_pfx)) => match rm.route.path() {
+                        PathMatch::Prefix(pfx_len) if *pfx_len <= orig_path.len() => {
+                            let (_, rest) = orig_path.split_at(*pfx_len);
+                            format!("{}{}", new_pfx, rest)
+                        }
```
hawkw, line 79: if we converted the paths in the other match arms to `http::uri::PathAndQuery` here, we could re-use the original request path without formatting it to a string and then back again:
```suggestion
            let path: http::uri::PathAndQuery = {
                use crate::http::r#match::PathMatch;

                let orig_path = orig_uri.path();
                match &self.path {
                    None => orig_path,
                    Some(ModifyPath::ReplaceFullPath(p)) => p.clone().try_into()?,
                    Some(ModifyPath::ReplacePrefixMatch(new_pfx)) => match rm.route.path() {
                        PathMatch::Prefix(pfx_len) if *pfx_len <= orig_path.len() => {
                            let (_, rest) = orig_path.split_at(*pfx_len);
                            format!("{}{}", new_pfx, rest).try_into()?
                        }
```

we would need to change the error a bit in that case, though, because the `String` -> `PathAndQuery` conversion also returns `InvalidUrl`

hawkw (commented):
i got kind of nerd-sniped by trying to reduce allocation & copying in the `RedirectRequest`'s URI-rewriting code, so i left a bunch of comments there. if you like, i'd be happy to open a PR against this with what i believe would be a more efficient implementation, but it's also something we can come back to later.

linkerd/http-route/src/http/filter/redirect.rs
```diff
@@ -0,0 +1,100 @@
...
+                        _ => return Err(InvalidRedirect::InvalidReplacePrefix),
+                    },
+                }
+            };
```
hawkw, line 83: Noticed what I believe is a bug here: in all cases (even when just using the original path), this filter will clobber any query params. This is because we get `orig_path` by calling `orig_uri.path()`, [which returns _just_ the path component off the URI](https://docs.rs/http/latest/http/uri/struct.Uri.html#method.path), and not the query component, but then we pass a value constructed using `orig_path` to `uri::Builder::path_and_query` when constructing the new URI. If the original URL had a query component, it's discarded here.

I think we should probably call [`orig_uri.query()`](https://docs.rs/http/latest/http/uri/struct.Uri.html#method.query) at some point and ensure that the query parameters are added to the new path-and-query part passed to `uri::Builder::path_and_query`.

hawkw: (as a side note, it also appears that the redirect filter will currently always throw out any [fragment identifier](https://www.rfc-editor.org/rfc/rfc3986#section-3.5) component of the URI...but unfortunately, i don't think we can actually fix that one in Linkerd, because the `http` crate's `uri` module does not currently appear to expose any APIs for getting or setting a fragment?)

PR #1793, opened 2022-06-28
inbound: Add a header-modification route filter

olix0r:
The Gateway API defines a [header-rewriting filter](gwapi) that may be
attached to HTTP routes. This change updates the HTTP route types to
support a list of filters. The inbound proxy supports only the
header modifier filter. Additional filters will be added in follow-up
changes (see #1788 for an example of these additional filters).

[gwapi]: https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1alpha2.HTTPRequestHeaderFilter

### DIFF
--- a/linkerd/app/inbound/src/policy/http/tests.rs
+++ b/linkerd/app/inbound/src/policy/http/tests.rs
@@ -3,8 +3,24 @@ use crate::policy::{Authentication, Authorization, Meta, Protocol, ServerPolicy}
 use linkerd_app_core::{svc::Service, Infallible};
 use std::sync::Arc;
 
+macro_rules! conn {
+    ($client:expr, $dst:expr) => {{
+        ConnectionMeta {
+            dst: OrigDstAddr(($dst, 8080).into()),
+            client: Remote(ClientAddr(($client, 30120).into())),
+            tls: tls::ConditionalServerTls::Some(tls::ServerTls::Established {
+                client_id: Some("foo.bar.bah".parse().unwrap()),
+                negotiated_protocol: None,
+            }),
+        }
+    }};
+    () => {{
+        conn!([192, 168, 3, 3], [192, 168, 3, 4])
+    }};
+}
+
 macro_rules! new_svc {
-    ($proto:expr, $conn:expr) => {{
+    ($proto:expr, $conn:expr, $rsp:expr) => {{
         let (policy, tx) = AllowPolicy::for_test(
             $conn.dst,
             ServerPolicy {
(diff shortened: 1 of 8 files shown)


linkerd/app/inbound/src/policy/http/tests.rs
```diff
@@ -130,6 +143,69 @@ async fn http_route() {
...
+                    networks: vec![std::net::IpAddr::from([192, 168, 3, 3]).into()],
+                    meta: Arc::new(Meta::Resource {
+                        group: "policy.linkerd.io".into(),
+                        kind: "server".into(),
+                        name: "testsaz".into(),
+                    }),
+                }]),
+                filters: vec![Filter::RequestHeaders(filter::ModifyRequestHeader {
+                    add: vec![("testkey".parse().unwrap(), "testval".parse().unwrap())],
+                    ..filter::ModifyRequestHeader::default()
```
hawkw, line 174: seems like it might be worth also having tests for `set` and `remove`, but since the actual filter mechanics are fairly simple and just call a `HeaderMap` method, that may not be worth the boilerplate?

linkerd/http-route/src/http/filter/modify_request_header.rs
```diff
@@ -0,0 +1,24 @@
+use http::header::{HeaderMap, HeaderName, HeaderValue};
+
+#[derive(Clone, Debug, Default, Hash, PartialEq, Eq)]
+pub struct ModifyRequestHeader {
+    pub add: Vec<(HeaderName, HeaderValue)>,
+    pub set: Vec<(HeaderName, HeaderValue)>,
+    pub remove: Vec<HeaderName>,
```
hawkw, line 7: it makes sense for `add` to be a `Vec`, since we might want to add multiple values for the same header name, but i wonder if `set` (and `remove`?) ought to be a `HashMap` or something? allowing multiple entries in `set` for the same header name feels kinda sketchy, since then we would end up with a final header value that depends on the order of the entries in the `Vec`...and for removing headers, removing the same header name multiple times is just an unnecessary hashmap lookup...

linkerd/http-route/src/http/filter/modify_request_header.rs
```diff
@@ -0,0 +1,24 @@
+use http::header::{HeaderMap, HeaderName, HeaderValue};
+
+#[derive(Clone, Debug, Default, Hash, PartialEq, Eq)]
+pub struct ModifyRequestHeader {
```
hawkw, line 4: nit/TIOLI: it seems like this type doesn't _really_ need to be named "modify *request* header", since there isn't actually anything request-specific about it? the `apply` method just mutates a `HeaderMap`, so it seems like the same struct could eventually be used to modify response headers too...it's the `Filter` enum that's going to be responsible for determining which `HeaderMap` is modified...

hawkw (approved):
looks good to me overall! couple of small suggestions, but they're not blockers.

hawkw:
oh, another thought: at _some_ level, it seems like we should probably be validating that a header modification route filter isn't setting any connection-level headers, or headers that are set by the underlying HTTP implementation. for example, if a route has a filter that adds a `Content-Length` header with an arbitrary value...that's probably always gonna be bad news, and we should maybe validate such filters at some point? but it could also be done when the route is _created_ rather than in the proxy, i guess...

linkerd/app/inbound/src/policy/http/tests.rs
olix0r: i don't think that needs to be tested from here. We could have tests for the `filter::ModifyHeader` type that exercised that, but this is just ensuring that the filter gets applied by the service. There's nothing that could change in this service to keep this test passing while breaking set/remove.

olix0r:
> we should probably be validating that a header modification route filter isn't setting any connection-level headers

agreed. we'll address that when we deal with creating the filters.

PR #1799, opened 2022-06-30
Add an HTTP route redirect filter

olix0r:
The Gateway API requires that `HTTPRoute` implementations support a
[redirection filter][gwapi].

This change adds a filter implementation to the `http-route` crate.
Inbound server policies are updated to apply this filter on HTTP routes.
When a request is redirected, the policy layer returns an error and the
error crate is responsible for turning that error into an HTTP response.

This change fixes path matching so that trailing slashes do not impact
prefix matching.

[gwapi]: https://github.com/kubernetes-sigs/gateway-api/blob/541e9fc2b3c2f62915cb58dc0ee5e43e4096b3e2/apis/v1beta1/httproute_types.go#L772-L821

### DIFF
--- a/linkerd/app/core/src/errors/respond.rs
+++ b/linkerd/app/core/src/errors/respond.rs
@@ -36,7 +36,7 @@ pub struct SyntheticHttpResponse {
     http_status: http::StatusCode,
     close_connection: bool,
     message: Cow<'static, str>,
-    location: Option<http::header::HeaderValue>,
+    location: Option<HeaderValue>,
 }
 
 #[derive(Copy, Clone, Debug)]
@@ -167,7 +167,7 @@ impl SyntheticHttpResponse {
             close_connection: false,
             message: Cow::Borrowed("redirected"),
             location: Some(
-                http::header::HeaderValue::from_str(&*location.to_string())
+                HeaderValue::from_str(&*location.to_string())
                     .expect("location must be a valid header value"),
             ),
         }
--- a/linkerd/app/core/src/errors/respond.rs
+++ b/linkerd/app/core/src/errors/respond.rs
@@ -1,5 +1,5 @@
 use crate::svc;
-use http::header::HeaderValue;
+use http::header::{HeaderValue, LOCATION};
 use linkerd_error::{Error, Result};
 use linkerd_error_respond as respond;
 pub use linkerd_proxy_http::{ClientHandle, HasH2Reason};
@@ -32,10 +32,11 @@ pub trait HttpRescue<E> {
 
 #[derive(Clone, Debug)]
 pub struct SyntheticHttpResponse {
-    pub grpc_status: tonic::Code,
-    pub http_status: http::StatusCode,
-    pub close_connection: bool,
-    pub message: Cow<'static, str>,
+    grpc_status: tonic::Code,
+    http_status: http::StatusCode,
+    close_connection: bool,
+    message: Cow<'static, str>,
+    location: Option<http::header::HeaderValue>,
 }
 
 #[derive(Copy, Clone, Debug)]
@@ -95,6 +96,7 @@ impl SyntheticHttpResponse {
             http_status: http::StatusCode::INTERNAL_SERVER_ERROR,
             grpc_status: tonic::Code::Internal,
             message: Cow::Borrowed("unexpected error"),
+            location: None,
         }
     }
 
@@ -104,6 +106,7 @@ impl SyntheticHttpResponse {
             http_status: http::StatusCode::BAD_GATEWAY,
             grpc_status: tonic::Code::Unavailable,
             message: Cow::Owned(msg.to_string()),
+            location: None,
         }
     }
 
@@ -113,6 +116,7 @@ impl SyntheticHttpResponse {
             http_status: http::StatusCode::GATEWAY_TIMEOUT,
             grpc_status: tonic::Code::Unavailable,
             message: Cow::Owned(msg.to_string()),
+            location: None,
         }
     }
 
@@ -122,6 +126,7 @@ impl SyntheticHttpResponse {
             grpc_status: tonic::Code::Unauthenticated,
             close_connection: false,
             message: Cow::Owned(msg.to_string()),
+            location: None,
         }
     }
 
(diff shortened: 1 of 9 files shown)


linkerd/http-route/src/http/filter/redirect.rs
```diff
@@ -0,0 +1,416 @@
...
+        let status = self.status.unwrap_or(http::StatusCode::MOVED_PERMANENTLY);
+
+        Ok(Some(Redirection { status, location }))
+    }
+
+    fn authority(&self, orig_uri: &http::Uri) -> Result<Authority, InvalidRedirect> {
+        match (self.host.as_deref(), self.port) {
+            // If a host is configured, use it and whatever port is configured.
+            (Some(h), Some(p)) => format!("{}:{}", h, p).parse().map_err(Into::into),
+            (Some(h), None) => h.parse().map_err(Into::into),
```
hawkw, line 76: i still think it would be preferable to perform this parsing when the `RedirectRequest` is _constructed_, rather than every time a redirect is processed --- these authorities are always going to be the same when a host is configured...

linkerd/http-route/src/http/filter/redirect.rs
```diff
@@ -0,0 +1,416 @@
...
+
+            // If a host is NOT configured, use the request's original host
+            // and either an overridden port or the original port.
+            (None, p) => {
+                let h = orig_uri.host().ok_or(InvalidRedirect::MissingAuthority)?;
+                match p.or_else(|| orig_uri.port_u16().and_then(|p| p.try_into().ok())) {
+                    Some(p) => format!("{}:{}", h, p).parse().map_err(Into::into),
```
hawkw, line 83: using the `TryFrom<String>` impl rather than `FromStr` would avoid an allocation/deallocation and is a pretty trivial change here
```suggestion
            (Some(h), Some(p)) => format!("{}:{}", h, p).try_into().map_err(Into::into),
            (Some(h), None) => h.parse().map_err(Into::into),

            // If a host is NOT configured, use the request's original host
            // and either an overridden port or the original port.
            (None, p) => {
                let h = orig_uri.host().ok_or(InvalidRedirect::MissingAuthority)?;
                match p.or_else(|| orig_uri.port_u16().and_then(|p| p.try_into().ok())) {
                    Some(p) => format!("{}:{}", h, p).try_into().map_err(Into::into),
```

linkerd/http-route/src/http/filter/redirect.rs
```diff
@@ -0,0 +1,416 @@
...
+                    let mut new_path = new_pfx.to_string();
+                    let (_, rest) = orig_uri.path().split_at(*pfx_len);
+                    if !rest.is_empty() && !rest.starts_with('/') {
+                        new_path.push('/');
+                    }
+                    new_path.push_str(rest);
+                    if let Some(q) = orig_uri.query() {
+                        new_path.push('?');
+                        new_path.push_str(q);
+                    }
```
hawkw, line 126: :+1:

linkerd/http-route/src/http/filter/redirect.rs
```diff
@@ -0,0 +1,416 @@
...
+                // filter is invalid. This should cause us to fail requests with
+                // a 5XX.
+                _ => Err(InvalidRedirect::InvalidReplacePrefix),
+            },
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
hawkw, line 140: thanks for adding tests for this!

as a side note, this feels like the kind of code that _might_ be well-suited by property testing...but i'm not sure if it's worth the effort to get `quickcheck`/`proptest` to generate arbitrary valid URIs or not...

linkerd/app/core/src/errors/respond.rs
```diff
@@ -149,6 +156,30 @@ impl SyntheticHttpResponse {
...
+
+    pub fn redirect(http_status: http::StatusCode, location: &http::Uri) -> Self {
+        Self {
+            http_status,
+            grpc_status: tonic::Code::NotFound,
+            close_connection: false,
+            message: Cow::Borrowed("redirected"),
+            location: Some(
+                http::header::HeaderValue::from_str(&*location.to_string())
+                    .expect("location must be a valid header value"),
```
hawkw, line 171: tiny nit: `HeaderValue` has [a `TryFrom<String>` impl](https://docs.rs/http/latest/http/header/struct.HeaderValue.html#impl-TryFrom%3CString%3E), which would take the string by-value and turn it into a `Bytes`, rather than copying from a borrowed string and then dropping the original string:
```suggestion
                http::header::HeaderValue::try_from(location.to_string())
                    .expect("location must be a valid header value"),
```

hawkw: (we would also need to import `TryFrom` if we apply this change)

linkerd/app/core/src/errors/respond.rs
```diff
@@ -32,10 +32,11 @@ pub trait HttpRescue<E> {
...
 pub struct SyntheticHttpResponse {
-    pub grpc_status: tonic::Code,
-    pub http_status: http::StatusCode,
-    pub close_connection: bool,
-    pub message: Cow<'static, str>,
+    grpc_status: tonic::Code,
+    http_status: http::StatusCode,
+    close_connection: bool,
+    message: Cow<'static, str>,
+    location: Option<http::header::HeaderValue>,
```
hawkw, line 39: tiny style nit: `HeaderValue` is imported:
```suggestion
    location: Option<HeaderValue>,
```

linkerd/app/core/src/errors/respond.rs
hawkw: also, a very tiny nit: `HeaderValue` is imported already:
```suggestion
                HeaderValue::from_str(&*location.to_string())
                    .expect("location must be a valid header value"),
```

hawkw (approved):
overall, this looks good to me! i commented on a few relatively minor nits.

i do still think it would be preferable to parse the authority once in the overridden host and port case, when constructing the `RedirectRequest` instance, rather than parsing it every time we redirect a request. but, this can be changed in a follow-up.

hawkw (approved):
looks good to me!

PR #1802, opened 2022-06-30
2.161.3: cherry-pick #1670

hawkw:
When resolving control plane components, we first resolve a SRV
record. If we're unable to parse IPv4 addresses from the record,
then we fall back to an A record lookup. But if the initial resolution
fails completely (i.e. doesn't produce any record data), then we don't
fall back to an A record.

This change fixes this fallback behavior so that we attempt A record
lookups whenever a SRV lookup fails. In doing so, we change the dns
module to use typed errors (avoiding boxing) so that we can print
better error messages and have static guarantees about the error
types being handled.

Fixes linkerd/linkerd2#8296

Co-authored-by: Alex Leong <alex@buoyant.io>

### DIFF
--- a/linkerd/app/core/src/control.rs
+++ b/linkerd/app/core/src/control.rs
@@ -64,13 +64,8 @@ impl Config {
             let backoff = self.connect.backoff;
             move |error: Error| {
                 warn!(%error, "Failed to resolve control-plane component");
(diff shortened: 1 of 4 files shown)


hawkw:
Part of linkerd/linkerd2#8779

kleimkuhler (approved):

PR #1803, opened 2022-06-30
2.161.3: cherry-pick #1699

hawkw:

Fixes linkerd/linkerd2#8539


hawkw:
Part of linkerd/linkerd2#8779

kleimkuhler (approved):

PR #1804, opened 2022-06-30
2.161.3: cherry-pick #1759

hawkw:
The proxy can receive redundant discovery updates. When this occurs, it
causes the balancer to churn, replacing an endpoint stack (and
therefore needlessly dropping a connection).

This change updates the discovery module to keep clones of the
discovered endpoint metadata, so that updated values can be compared to
eliminate duplicate updates.

Relates to linkerd/linkerd2#8677

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1087,13 +1087,14 @@ dependencies = [
 name = "linkerd-proxy-discover"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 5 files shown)


hawkw:
Part of linkerd/linkerd2#8779

kleimkuhler (approved):

PR #1805, opened 2022-07-01
Add HTTP and gRPC route failure-injection filters

olix0r:
The control plane may choose to inject failures for a route. It may
cause all requests to fail, or only a random subset of requests to fail.

This change adds HTTP and gRPC route filters that can cause error
responses for inbound requests.

### DIFF
--- a/linkerd/app/inbound/src/policy/http.rs
+++ b/linkerd/app/inbound/src/policy/http.rs
@@ -12,7 +12,7 @@ use linkerd_app_core::{
     Error, Result,
 };
 use linkerd_server_policy::{grpc, http, route::RouteMatch};
-use std::task;
+use std::{sync::Arc, task};
 
 #[cfg(test)]
 mod tests;
@@ -64,6 +64,20 @@ pub struct HttpRouteRedirect {
 #[derive(Debug, thiserror::Error)]
 #[error("unauthorized request on route")]
 pub struct HttpRouteUnauthorized(());
+
+#[derive(Debug, thiserror::Error, Clone, PartialEq)]
+#[error("HTTP request configured to fail with {status}: {message}")]
+pub struct HttpRouteInjectedFailure {
+    pub status: ::http::StatusCode,
+    pub message: Arc<str>,
+}
+
+#[derive(Debug, thiserror::Error, Clone, PartialEq)]
+#[error("gRPC request configured to fail with {code}: {message}")]
+pub struct GrpcRouteInjectedFailure {
+    pub code: u16,
+    pub message: Arc<str>,
+}
 
 // === impl NewHttpPolicy ===
 
@@ -248,6 +262,12 @@ fn apply_http_filters<B>(
     // TODO Do any metrics apply here?
     for filter in &route.filters {
         match filter {
+            http::Filter::InjectFailure(fail) => {
+                if let Some(http::filter::FailureResponse { status, message }) = fail.apply() {
+                    return Err(HttpRouteInjectedFailure { status, message }.into());
+                }
+            }
+
             http::Filter::Redirect(redir) => match redir.apply(req.uri(), &r#match) {
                 Ok(Some(http::filter::Redirection { status, location })) => {
                     return Err(HttpRouteRedirect { status, location }.into());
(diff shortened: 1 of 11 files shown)


linkerd/app/inbound/src/policy/http.rs
```diff
@@ -65,6 +65,20 @@ pub struct HttpRouteRedirect {
 #[error("unauthorized request on route")]
 pub struct HttpRouteUnauthorized(());
 
+#[derive(Debug, thiserror::Error, Clone, PartialEq)]
```
hawkw, line 68: nit, not important: i _think_ we can still import `thiserror::Error` even though `linkerd_app_core::Error` is also in scope; IIRC derive macros occupy a different namespace from traits.

linkerd/http-route/src/http/filter/inject_failure.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+impl Eq for Distribution {}
+
+impl Hash for Distribution {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.numerator.hash(state);
+        self.denominator.hash(state);
+    }
+}
```
hawkw, line 74: not *totally* sure why we need our own `Distribution` type when it appears to just be a wrapper over `rand`'s `distributions::Bernoulli` type --- is this just so we can impl `Hash` on it, or something? maybe worth a comment indicating that...

linkerd/http-route/src/http/filter/inject_failure.rs
```diff
@@ -0,0 +1,74 @@
...
+
+// === impl InjectFailure ===
+
+impl Distribution {
+    pub fn from_ratio(
+        numerator: u32,
+        denominator: u32,
+    ) -> Result<Self, rand::distributions::BernoulliError> {
+        let inner = rand::distributions::Bernoulli::from_ratio(numerator, denominator)?;
```
hawkw, line 45: this seems like it could eventually be a cool use case for the [`fast_bernoulli` crate](https://crates.io/crates/fast-bernoulli), rather than `rand` --- that crate avoids generating random numbers when an event is *not* sampled, which could be a kind of nice optimization. but, we can always play with that later; it's not a blocker for now!

linkerd/http-route/src/grpc/filter/inject_failure.rs
```diff
@@ -0,0 +1,30 @@
+use std::hash::Hash;
+
+pub use crate::http::filter::Distribution;
+
+/// A filter that responds with an HTTP error.
```
hawkw, line 5: nit:
```suggestion
/// A filter that responds with a gRPC error.
```

linkerd/http-route/src/grpc/filter/inject_failure.rs
```diff
@@ -0,0 +1,30 @@
...
+    pub fn apply(&self) -> Option<FailureResponse> {
+        use rand::distributions::Distribution;
+
+        if self.distribution.sample(&mut rand::thread_rng()) {
+            return Some(self.response.clone());
+        }
+
+        None
+    }
+}
```
hawkw, line 30: it occurs to me that we _could_ get away with just having a single 
```rust
pub struct InjectFailure<R: Clone> {
    pub response: R,
    pub distribution: Distribution,
}

impl<R: Clone> InjectFailure<R> {
    pub fn apply(&self) -> Option<R> {
        use rand::distributions::Distribution;

        if self.distribution.sample(&mut rand::thread_rng()) {
            return Some(self.response.clone());
        }

        None
    }
}
```
and use that for both HTTP and gRPC. But, this may be kind of a premature genericization, so :woman_shrugging:

linkerd/http-route/src/http/filter/inject_failure.rs
olix0r: https://github.com/linkerd/linkerd2-proxy/blob/e8220da644fb4fc5e01b386c256cc7905a0eacd5/linkerd/http-route/src/http/filter/inject_failure.rs#L16

linkerd/http-route/src/http/filter/inject_failure.rs
olix0r: yeah, if we find other uses for that crate, i'd be more inclined to use it here. it doesn't seem worth a dependency just for this use case though.

linkerd/app/inbound/src/policy/http.rs
olix0r: it's done this way elsewhere in the file already. would rather keep the change small/consistent for now.

hawkw (approved):
looks good to me!

PR #1814, opened 2022-07-06
Configure inbound HTTP routes via gRPC

olix0r:
The policy controller serves inbound server configuration. Recent
changes have updated the inbound proxy to support applying HTTP
route-specific policies, but only a single default route configuration
was used.

This change updates the proxy to use a (not yet released) new proxy API
version that includes HTTP routes in server responses. This change adds
protobuf conversions from these types to the proxy's HTTP route types.

If the proxy receives a router filter of an unknown type (i.e., because
the controller is running a later version of the API that includes new
types), then the proxy will FAIL all requests on that route with an
internal server error. It's considered safer to fail hard in this case,
rather than to silently ignore a configured policy that could
potentially be security-sensitive.

---

## TODO

- [x] Stack test that ensures policies are enforced
- [x] What happens when an HTTP route includes no authorizations? Do we have to support this?
- [x] Do we need to apply further validations (e.g. path, hostname)? Bad data can't cause false-positive matches, really; so maybe not a concern.


hawkw (commented):
looks good to me overall! i commented on some fairly minor nits.

olix0r:
test image up at `ghcr.io/olix0r/l2-proxy:ver.inbound-route-proto.7b9ea575`

linkerd/http-route/src/grpc/filter/inject_failure.rs
```diff
@@ -10,3 +10,47 @@ pub struct FailureResponse {
...
+        type Error = InvalidFailureResponse;
+
+        fn try_from(proto: api::GrpcFailureInjector) -> Result<Self, Self::Error> {
+            let response = {
+                if proto.code > u16::MAX as u32 {
+                    return Err(InvalidFailureResponse::StatusNonU16(proto.code));
+                }
+
+                FailureResponse {
+                    code: proto.code as u16,
```
hawkw, line 40: it occurs to me that we _could_ use a checked cast here with `try_into`, rather than checking and then casting...but i'm not sure if that would actually make the code that much clearer, so this is fine.

linkerd/http-route/src/http/filter/inject_failure.rs
```diff
@@ -73,3 +80,59 @@ impl Hash for Distribution {
...
+pub mod proto {
+    use super::*;
+    use linkerd2_proxy_api::http_route as api;
+
+    #[derive(Debug, thiserror::Error)]
+    pub enum InvalidFailureResponse {
+        #[error("invalid HTTP status code: {0}")]
+        Status(#[from] http::status::InvalidStatusCode),
+
+        #[error("invalid HTTP status code: {0}")]
```
hawkw, line 94: should the message here maybe indicate that the number was too large, so that it can be distinguished from the other error? not a big deal, but just a thought.

linkerd/http-route/src/http/filter/redirect.rs
```diff
@@ -193,6 +193,84 @@ impl RedirectRequest {
...
+                    (h, None) => {
+                        let a = h.try_into()?;
+                        Some(AuthorityOverride::Host(a))
+                    }
+                }
+            };
+
+            let path = rr.path.and_then(|p| p.replace).map(|p| match p {
+                api::path_modifier::Replace::Full(path) => {
+                    // TODO ensure path is valid.
```
hawkw, line 249: assuming these TODOs are going to be addressed in a subsequent PR?

linkerd/http-route/src/http/match/path.rs
```diff
@@ -102,6 +102,36 @@ impl std::cmp::Ord for PathMatch {
...
+        Regex(#[from] regex::Error),
+    }
+
+    // === impl MatchPath ===
+
+    impl TryFrom<api::PathMatch> for MatchPath {
+        type Error = InvalidPathMatch;
+
+        fn try_from(rm: api::PathMatch) -> Result<Self, Self::Error> {
+            // TODO parse paths to validate they're valid.
```
hawkw, line 125: do we want to address this now, or is it something we're going to do later?

linkerd/server-policy/src/lib.rs
```diff
@@ -119,46 +125,68 @@ pub mod proto {
...
+                api::proxy_protocol::Kind::Grpc(api::proxy_protocol::Grpc { routes }) => {
+                    Protocol::Grpc(mk_routes!(grpc, authorizations, routes)?)
                 }
 
                 api::proxy_protocol::Kind::Tls(_) => Protocol::Tls(authorizations),
                 api::proxy_protocol::Kind::Opaque(_) => Protocol::Opaque(authorizations),
             };
 
-            let meta = Meta::try_new_with_default(proto.labels, "policy.linkerd.io", "server")?;
+            // TODO use a `Metadata` type
```
hawkw, line 187: the metadata type has landed, so we can address this now, right?

linkerd/http-route/src/http/filter/redirect.rs
```diff
@@ -193,6 +193,84 @@ impl RedirectRequest {
...
+        #[error("invalid location scheme: {0}")]
+        Scheme(#[from] http_types::InvalidScheme),
+
+        #[error("invalid HTTP status code: {0}")]
+        Status(#[from] http::status::InvalidStatusCode),
+
+        #[error("invalid HTTP authority: {0}")]
+        Authority(#[from] http::uri::InvalidUri),
+
+        #[error("invalid HTTP status code: {0}")]
```
hawkw, line 212: should this error maybe indicate that the status was too big to fit in 16 bits, to differentiate it from the other "invalid HTTP status code" message?

hawkw (approved):
this change looks good to me whenever the proxy-api release is ready. i commented on some very minor style nits, none of which are important --- i think all the substantial feedback i've had has been addressed already.

linkerd/http-route/src/http/filter/redirect.rs
olix0r: See TODOs in PR description. I'm not sure it's required.

linkerd/http-route/src/http/match/path.rs
olix0r: See TODOs in PR description. I'm not sure it's required.

linkerd/app/inbound/src/policy/api.rs
```diff
@@ -71,7 +71,11 @@ where
             Ok(rsp.map(|updates| {
                 updates
                     .map(|up| {
-                        let policy = up?.try_into().map_err(|e| {
+                        // FIXME(ver): If we get an invalid policy, we can't
+                        // pretend we didn't get it -- it might mean that the
+                        // server has tried to configure a policy that we don't
+                        // support!
```
hawkw, line 77: does this need an issue?

linkerd/app/inbound/src/policy/http/tests.rs
```diff
@@ -84,8 +84,8 @@ async fn http_route() {
                         networks: vec![std::net::IpAddr::from([192, 168, 3, 3]).into()],
                         meta: Arc::new(Meta::Resource {
                             group: "policy.linkerd.io".into(),
-                            kind: "server".into(),
-                            name: "testsaz".into(),
+                            kind: "AuthorizatoinPolicy".into(),
```
hawkw, line 87: typo:
```suggestion
                            kind: "AuthorizationPolicy".into(),
```

linkerd/http-route/src/http/filter/modify_header.rs
```diff
@@ -1,11 +1,13 @@
 use http::header::{HeaderMap, HeaderName, HeaderValue};
-use std::collections::{BTreeMap, BTreeSet};
 
+// Note that while `set` and `remove` could perhaps be better-modled as a Map
```
hawkw, line 3: typo
```suggestion
// Note that while `set` and `remove` could perhaps be better-modeled as a Map
```

hawkw (approved):

PR #1815, opened 2022-07-07
add shutdown grace period timeout for graceful shutdowns

hawkw:
When a proxy receives a SIGTERM or the shutdown admin endpoint is
called, a graceful shutdown is triggered. This causes the proxy to stop
accepting new connections, wait until all currently open connections
have closed, and then terminate. This means that proxies can potentially
be kept alive indefinitely even after a shutdown has been signalled, if
any open connections do not close (or if a connection closes but the
proxy fails to notice it has closed; see linkerd/linkerd2#8033).

In order to ensure that the proxy terminates when a shutdown is
signalled even if some connections fail to close, this branch introduces
a grace period timeout to the shutdown process. When a graceful shutdown
is initiated, the proxy will still wait to drain open connections, as it
did previously, but if the drain does not complete within the shutdown
grace period, the proxy will terminate "gracelessly".

To configure the shutdown grace period, this PR adds a new
`LINKERD2_PROXY_SHUTDOWN_GRACE_PERIOD` environment variable which can be
set by the proxy injector. The duration is parsed the same way all other
configuration durations are parsed. The default grace period is 2
minutes.

Note that currently, there will always be a timeout, even if the
environment variable is not set, as it defaults to 2 minutes. If we
really want to have a way to ensure a proxy will always wait
indefinitely for a drain to complete, we can change the default, or add
a way to explicitly disable the grace period through configuration...but
I thought it was probably fine to just pick a fairly generous grace
period by default, and always have a timeout.

### DIFF
--- a/linkerd2-proxy/src/main.rs
+++ b/linkerd2-proxy/src/main.rs
@@ -113,15 +113,17 @@ fn main() {
         let drain = app.spawn();
         tokio::select! {
             _ = signal::shutdown() => {
-                info!(grace_period = ?shutdown_grace_period, "Received shutdown signal");
+                info!("Received shutdown signal");
             }
             _ = shutdown_rx.recv() => {
-                info!(grace_period = ?shutdown_grace_period, "Received shutdown via admin interface");
+                info!("Received shutdown via admin interface");
             }
         }
         match time::timeout(shutdown_grace_period, drain.drain()).await {
             Ok(()) => debug!("Shutdown completed gracefully"),
-            Err(_) => warn!("Graceful shutdown did not complete in {shutdown_grace_period:?}, terminating now"),
+            Err(_) => warn!(
+                "Graceful shutdown did not complete in {shutdown_grace_period:?}, terminating now"
+            ),
         }
     });
 }


kleimkuhler (approved):
A timeout always being there seems reasonable 👍

linkerd2-proxy/src/main.rs
```diff
@@ -111,12 +113,15 @@ fn main() {
         let drain = app.spawn();
         tokio::select! {
             _ = signal::shutdown() => {
-                info!("Received shutdown signal");
+                info!(grace_period = ?shutdown_grace_period, "Received shutdown signal");
             }
             _ = shutdown_rx.recv() => {
-                info!("Received shutdown via admin interface");
+                info!(grace_period = ?shutdown_grace_period, "Received shutdown via admin interface");
```
olix0r, line 119: I'd prefer that we either log the grace period here or below in the timeout case, but it seems a bit overzealous to log it in both places. I'd suggest omitting it here since it won't be encountered the vast majority of the time.

olix0r (approved):

PR #1867, opened 2022-08-03
admin: add `/env.json` endpoint

hawkw:
We have a need to see the current environment variables for running
Linkerd proxies: while a pod spec includes most of the proxy's
environment variables, this environment may include references that are
not rendered until the container actually starts.

For example:
```yaml
      - name: _pod_sa
        valueFrom:
          fieldRef:
            apiVersion: v1
            fieldPath: spec.serviceAccountName
      - name: LINKERD2_PROXY_IDENTITY_DIR
        value: /var/run/linkerd/identity/end-entity
      - name: LINKERD2_PROXY_IDENTITY_TRUST_ANCHORS
        valueFrom:
          configMapKeyRef:
            key: ca-bundle.crt
            name: linkerd-identity-trust-roots
```

There's no reliable, non-racey way to know which value a pod is actually
running with except to have the proxy expose its environment values.

See linkerd/linkerd2#9067 for details.

This branch implements the proposal described in [this comment][1]: we
add an `/env.json` endpoint to the proxy's admin server that responds to
GET requests with a JSON object representing that proxy's environment
variables. If the request has query parameters, the response will only
contain the value of the environment variables with those names (or JSON
`null`s if a requested env var name is unset).

Closes linkerd/linkerd2#9067

[1]: https://github.com/linkerd/linkerd2/issues/9067#issuecomment-1204280075


linkerd/app/integration/src/tests/telemetry.rs
```diff
@@ -5,6 +5,7 @@
 
 mod log_stream;
 mod tcp_errors;
+mod env_json;
```
olix0r, line 8: perhaps:

```suggestion
mod env;
```

The JSON-ness of the module seems like a detail, we don't expose for log-streaming, etc.

linkerd/app/admin/src/server.rs
```diff
@@ -82,6 +83,52 @@ impl<M> Admin<M> {
...
+            .uri()
+            .path_and_query()
+            .and_then(http::uri::PathAndQuery::query);
+        let env = if let Some(query) = query {
+            if query.contains('=') {
+                return json::json_error_rsp(
+                    "env.json query parameters may not contain key-value pairs",
+                    StatusCode::BAD_REQUEST,
+                );
+            }
```
hawkw, line 111: @olix0r is this how we want to do the query params, or would we rather do something like `?env_var=FOO&env_var=BAR`?

olix0r: This seems good to me.

olix0r (approved):
lgtm!

PR #1876, opened 2022-08-04
Add authz labels in tap output

adleong:
In order to get more visibility into how live requests are mapped to servers, httproutes, and authorizations, we add the following labels to the RouteLabels that the proxy returns when a request is tapped:

* {srv, route, authz}_{group, kind, name}

The RouteLabels of a tap event were originally intended for metadata about the ServiceProfile route, so it is a bit weird for us to put authorization labels there.  However, it is a convenient place that we can use to return these labels without updating the tap API.  One unfortunate consequence of this is the `rt_` prefix in the CLI output for these labels.  We could consider adding a dedicated structure for authz labels to the tap events in the tap API if we feel that putting these in the RouteLabels is too weird.

Sample tap CLI output:

```console
> linkerd viz tap -n emojivoto deploy/web -o wide | fold -sw 80
req id=30:0 proxy=in  src=10.244.0.69:46034 dst=10.244.0.104:8080 tls=true
:method=GET :authority=web-svc.emojivoto:80 :path=/api/list
src_res=deploy/vote-bot src_ns=emojivoto dst_res=deploy/web dst_ns=emojivoto
rt_authz_group=policy.linkerd.io rt_route_group= rt_route_name=default
rt_route_kind=default rt_srv_group=policy.linkerd.io
rt_authz_kind=authorizationpolicy rt_srv_name=web-http rt_authz_name=web-public
rt_srv_kind=server
```

### DIFF
--- a/linkerd/app/inbound/src/http/router.rs
+++ b/linkerd/app/inbound/src/http/router.rs
@@ -1,6 +1,7 @@
 use crate::{policy, stack_labels, Inbound};
 use linkerd_app_core::{
-    classify, errors, http_tracing, metrics,
+    classify, errors, http_tracing,
+    metrics::{self, RouteAuthzLabels, RouteLabels},
     profiles::{self, DiscoveryRejected},
     proxy::{http, tap},
     svc::{self, ExtractParam, Param},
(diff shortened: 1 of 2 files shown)


linkerd/app/inbound/src/http/router.rs
```diff
@@ -417,9 +418,60 @@ impl tap::Inspect for Logical {
...
+            labels.insert(
+                "authz_name".to_string(),
+                authz_labels.authz.name().to_string(),
+            );
+        }
+        if labels.is_empty() {
+            None
+        } else {
+            Some(std::sync::Arc::new(labels))
+        }
```
olix0r, line 474: We already do this work here:

https://github.com/linkerd/linkerd2-proxy/blob/92d50fcfc5b9b7e0c71049d9250ed1e43fab1da4/linkerd/app/inbound/src/http/router.rs#L250-L276

That is, all of the copying/allocation happens once per stack and not for each request. This also avoids messing with extensions, etc. IIUC, we're already including route labels there...

adleong:
Ah, you're right.  The proxy is already serving these labels but the tap CLI is currently dropping them.

linkerd/app/inbound/src/http/router.rs
```diff
@@ -417,9 +418,60 @@ impl tap::Inspect for Logical {
...
+            .extensions()
             .get::<profiles::http::Route>()
-            .map(|r| r.labels().clone())
+            .iter()
+            .flat_map(|route| route.labels().iter())
+        {
+            labels.insert(k.clone(), v.clone());
+        }
+        if let Some(route_labels) = req.extensions().get::<RouteLabels>() {
+            labels.insert(
```
hawkw, line 431: nit/TIOLI: since we're always adding 6 labels in this branch, we could write
```suggestion
            labels.reserve(6);
            labels.insert(
```
so that we only reallocate once.

alternatively, using `labels.extend(&[...])` instead would also do this.

PR #1901, opened 2022-08-18
Fix the typo in the tests expectation

MOZGIII:


olix0r:
This comment is actually correct as-is:

https://github.com/linkerd/linkerd2-proxy/blob/285d41532255cff0ef324f93b6a7dbbc4f1da6a3/linkerd/app/inbound/src/server.rs#L89-L96

The inner stack for push_accept is the detect stack. It must not be built in this case.

MOZGIII:
Oh, damn I misread the code :D

PR #1914, opened 2022-08-30
build: add shell.nix and minimal flake.nix

lionello:
with build dependencies.

* Tested with `--pure`;
* Fixes running `just build` and `just docker` on clean MacOS with Nix (and Docker desktop.)
* Updated the readme (also mentioning `fetch`, without which `just build` fails.)
* Added a Github action

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,17 @@
+name: nix
+
+on: push
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install Nix
+      uses: cachix/install-nix-action@v12
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
+
+    - name: Build in shell
+      run: nix-shell --pure --run "just fetch build"
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,6 +1,12 @@
 name: nix
 
-on: push
+on:
+  pull_request:
+    paths:
+      - Cargo.lock
+      - "*.nix"
+      - justfile
+      - .github/workflows/nix.yml
 
 jobs:
   build:
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -12,10 +12,10 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
 
     - name: Install Nix
-      uses: cachix/install-nix-action@v12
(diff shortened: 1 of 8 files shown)


lionello:
Closing since I want to add CI before I open the PR.

.github/workflows/nix.yml
```diff
@@ -0,0 +1,23 @@
...
+      - .github/workflows/nix.yml
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install Nix
+      uses: cachix/install-nix-action@v12
```
olix0r, line 18: Please use the full SHA versions of these actions, as we do in other workflows.  This prevents the tag from changing without a PR (which dependabot will do as these actions are updated).

README.md
```diff
@@ -35,10 +35,19 @@ The Linkerd project is hosted by the Cloud Native Computing Foundation
...
 * `just test` -- Runs unit and integration tests on your local system using `cargo`
 * `just docker` -- Builds a Docker container image that can be used for testing.
 
+### Nix
+
+A [`shell.nix`](./shell.nix) (and generic [`flake.nix`](./flake.nix)) is
+provided with the build dependencies. Use this with `nix-shell` (or with
+`nix develop`, respectively) to get a reproducible development environment.
+This is particularly useful in conjuction with [Direnv](https://direnv.net)'s
+[`use flake`](https://direnv.net/man/direnv-stdlib.1.html#codeuse-flake-ltinstallablegtcode).
```
olix0r, line 49: Please move this down below the devcontainer section.

olix0r (commented):
Thanks for the submission, @lionello.

To be honest, I'm not really eager to have to support this, as we already have a reproducible development setup using a devcontainer. (I tried to use Nix for a while, but I had too many problems with it. To each their own, though.)

But @hawkw uses nix and works on this project regularly. As long as I never have to know anything about Nix, I'm open to merging it if Eliza wants to maintain/support it.

flake.lock
```diff
@@ -0,0 +1,40 @@
...
+    "root": {
+      "inputs": {
+        "flake-utils": "flake-utils",
+        "nixpkgs": "nixpkgs"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
```
olix0r, line 40: How is this file maintained? Does this need to change regularly? Should this be omitted from version control?

lionello:
@olix0r I understand completely, and yes, devcontainers take care of the dev experience parts. I opened the PR mostly so other Nix users would have a starting point. It's OK with me if we close this PR and leave it at that. People can still find it if they search for "nix".

flake.lock
lionello: It's created by `nix flake lock` (and updated by `nix flake update`). It should be in source control to have a completely reproducible environment.

PR #1931, opened 2022-09-08
inbound: Restrict the HTTP/1 idle timeout

olix0r:
The inbound proxy may hold idle inbound connections to the application open for (by default) 20s. This can be problematic when a server sets a lesser idle timeout and that idle timeout matches some poll interval (like a Kubernetes probe or metrics scrape).

This change reduces the HTTP/1 connection pool idle timeout to 3s. This should be short enough to avoid this problem in many cases but it is also large enough that connection pooling can help busy services.

In the future, this configuration should be made dynamic based on the proxy's API.


linkerd/app/src/env.rs
```diff
@@ -218,6 +218,9 @@ const ENV_INITIAL_STREAM_WINDOW_SIZE: &str = "LINKERD2_PROXY_HTTP2_INITIAL_STREA
 const ENV_INITIAL_CONNECTION_WINDOW_SIZE: &str =
     "LINKERD2_PROXY_HTTP2_INITIAL_CONNECTION_WINDOW_SIZE";
 
+const ENV_INBOUND_HTTP1_CONNECTION_POOL_IDLE_TIMEOUT: &str =
+    "LINKERD2_PROXY_INBOUD_HTTP1_CONNECTION_POOL_IDLE_TIMEOUT";
```
hawkw, line 222: typo: "inboud"
```suggestion
    "LINKERD2_PROXY_INBOUND_HTTP1_CONNECTION_POOL_IDLE_TIMEOUT";
```

linkerd/app/src/env.rs
```diff
@@ -258,6 +261,13 @@ const DEFAULT_SHUTDOWN_GRACE_PERIOD: Duration = Duration::from_secs(2 * 60);
 const DEFAULT_INBOUND_ROUTER_MAX_IDLE_AGE: Duration = Duration::from_secs(20);
 const DEFAULT_OUTBOUND_ROUTER_MAX_IDLE_AGE: Duration = Duration::from_secs(5);
 
+// XXX This default inbound connection idle timeout should be less than or equal
+// to the server's idle timeout so that we don't try to reuse a connection as it
+// is being timed out of the server.
+//
+// In the future this should be made configurable per-server from the proxy API.
```
hawkw, line 268: Is/should there be an issue for this?

hawkw (changes requested):
this looks good to me except that the name of the environment variable is spelled wrong :)

linkerd/app/src/env.rs
olix0r: discussed in https://github.com/linkerd/linkerd2/issues/9273, we'll create a new issue when it's scoped

hawkw (approved):
looks good to me!

PR #1947, opened 2022-09-27
Fix inbound server label ambiguity

olix0r:
As the proxy discovers policies, it refers to the source of the policy configuration. This comes in two flavors: "default" (synthetic) resources and resources that actually exist in the cluster.

When the proxy reads routes and authorization, the control plane uses protobuf types to differentiate these states; but server resources are described by a generic label map. When the proxy reads this label map, it doesn't properly detect when default/synthetic resources are used and it _always_ indicates that the resource is real.

This can lead to duplicate metrics when, for instance, the proxy initially uses a default server and then obtains a default configuration by discovery.

This change fixes this situation in two ways:

1. We now properly detect and differentiate default server configurations.
2. We update the hash and comparisons functions for the `Meta` type so that an improperly construct resource matches a default (i.e. when inserted into a label map).

Fixes linkerd/linkerd2#9502

### DIFF
--- a/linkerd/server-policy/src/meta.rs
+++ b/linkerd/server-policy/src/meta.rs
@@ -1,6 +1,6 @@
-use std::{borrow::Cow, hash::Hash, sync::Arc};
+use std::{borrow::Cow, sync::Arc};
 
-#[derive(Debug, PartialEq, Eq, Hash)]
+#[derive(Debug, Eq)]
 pub enum Meta {
     Default {
         name: Cow<'static, str>,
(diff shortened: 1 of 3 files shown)


linkerd/server-policy/src/meta.rs
```diff
@@ -16,31 +16,45 @@ pub enum Meta {
...
         match self {
-            Meta::Default { .. } => "",
-            Meta::Resource { group, .. } => group,
+            Self::Default { .. } => "",
+            Self::Resource { group, .. } => group,
         }
     }
 }
 
+impl std::cmp::PartialEq for Meta {
```
hawkw, line 44: nit: it might be nice if there was a comment explaining why this is manually implemented?

linkerd/server-policy/src/meta.rs
```diff
@@ -76,19 +90,26 @@ pub mod proto {
                 .unwrap_or_else(|| default_group.to_string());
 
             if let Some(kind) = labels.remove("kind") {
-                return Ok(Arc::new(Meta::Resource { group, kind, name }));
+                // If the controller set a 'group' and 'kind', but it's just
+                // indicating that we're dealing with a default, then
```
hawkw, line 94: then...?

hawkw (approved):
looks good to me overall, modulo a couple of very tiny comment-related nits.

hawkw (approved):
looks good to me!

PR #1957, opened 2022-10-14
[WIP] reimplement service profile matching with HTTPRoute types

hawkw:
This is a proof of concept of rewriting the internal proxy types used for matching Service Profile routes to use the same route-matching code that's used to match HTTPRoutes.

So far, it seems like this should be *possible*, but the logic for converting from the recursively-nested Service Profile route-matching types to the "flat" HTTPRoute route matching types is likely going to be _quite_ complex: https://github.com/linkerd/linkerd2-proxy/blob/c505931b22d31df158bd542d2fa5b82b011d5f66/linkerd/service-profiles/src/proto.rs#L78-L132

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1507,6 +1507,7 @@ dependencies = [
  "linkerd-dns-name",
  "linkerd-error",
  "linkerd-http-box",
+ "linkerd-http-route",
  "linkerd-proxy-api-resolve",
  "linkerd-stack",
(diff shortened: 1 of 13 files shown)


hawkw:
We've determined not to move forwards with this approach, since there's an impedance mismatch between the two different route-matching syntax. I'm closing this PR in favor of #2029 and friends.

PR #1973, opened 2022-11-04
Allow Unicode-dfs-2016 for unicode-ident

kleimkuhler:
The other unicode packages don't use this license, so an exception should be
okay. This could also just be moved to licenses.allow.

Fixes #1964 checks.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -29,6 +29,10 @@ confidence-threshold = 0.8
 exceptions = [
     { allow = ["Zlib"], name = "adler32", version = "*" },
     { allow = ["ISC", "MIT", "OpenSSL"], name = "ring", version = "*" },
+    # The Unicode-DFS-2016 license is necessary for unicode-ident because they
+    # use data from the unicode tables to generate the tables which are
+    # included in the application. We do not distribute those data files so
+    # this is not a problem for us. See https://github.com/dtolnay/unicode-ident/pull/9/files
     { allow = ["Unicode-DFS-2016"], name = "unicode-ident", version = "*"},
 ]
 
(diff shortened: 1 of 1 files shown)


hawkw:
I'm not a lawyer, but looking at the Unicode-DFS-2016 license (see [here](https://spdx.org/licenses/Unicode-DFS-2016.html)), it looks like this license requires us to reproduce the copyright and permissions notice somewhere in order to use the unicode data files:

> Permission is hereby granted, free of charge, to any person obtaining a copy of the Unicode data files and any associated documentation (the "Data Files") or Unicode software and any associated documentation (the "Software") to deal in the Data Files or Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, and/or sell copies of the Data Files or Software, and to permit persons to whom the Data Files or Software are furnished to do so, provided that either
>
>    (a) this copyright and permission notice appear with all copies of the Data Files or Software, or
>    (b) this copyright and permission notice appear in associated Documentation.

I'm not sure where exactly it would make the most sense for us to provide the Unicode copyright notice?

deny.toml
```diff
@@ -29,6 +29,7 @@ confidence-threshold = 0.8
 exceptions = [
     { allow = ["Zlib"], name = "adler32", version = "*" },
     { allow = ["ISC", "MIT", "OpenSSL"], name = "ring", version = "*" },
+    { allow = ["Unicode-DFS-2016"], name = "unicode-ident", version = "*"},
```
olix0r, line 32:
```suggestion
    # The Unicode-DFS--2016 license is necessary for unicode-ident because they
    # use data from the unicode tables to generate the tables which are
    # included in the application. We do not distribute those data files so
    # this is not a problem for us. See https://github.com/dtolnay/unicode-ident/pull/9/files
    { allow = ["Unicode-DFS-2016"], name = "unicode-ident", version = "*" },
```

hawkw (approved):
lgtm with the addition of the comment!

PR #1992, opened 2022-11-15
add header-based routing using the outbound policy API

hawkw:
This branch implements header-based routing using the new client policy
API added in linkerd/linkerd2-proxy-api#165.

## Overview

In particular, this branch does the following:

#### Client policy discovery:

* Add a new `linkerd-client-policy` crate with internal types
  representing the client policies and code for converting protobuf
  policy messages to those representations
* Add a new `policy` module in `linkerd-app-outbound` with an API client
* Add client policy discovery to the outbound proxy's `push_discover`
  stack.

This means that profile discovery and client policy discovery are:

* Stored in the same cache per original destination IP address.
* Currently always occur for all destinations, although the client policy
  discovery is dropped when no profile is resolved or if the profile is a
  direct endpoint.
* Occur in roughly the same place in the proxy and don't require
  propagating original destination addresses to new parts of the outbound
  stack.

#### Client policy HTTPRoute matching:

* Add a router implementation in `linkerd-client-policy` for recognizing
  an HTTPRoute defined in a client policy
* Add a new stack in the outbound proxy's `push_http_logical` stack that
  is switched to if a client policy is present, rather than the current
  Service Profile-based HTTP logical stack
* Add the HTTPRoute router to the client policy stack

#### Client policy traffic splitting:

* Move the traffic split implementation from `linkerd-service-profiles`
  into `linkerd-client-policy`, and update `linkerd-service-profiles` to
  depend on it from `client-policy`.
* Separate the traffic split middleware into fixed and `Dynamic`
  (updated from discovery) traffic splits, because updates to the set of
  routes in a client policy will be handled by the policy router.
* Build fixed traffic splits for each set of backends defined in an
  HTTPRoute rule in the client policy stack.

## Not Currently Implemented

- [ ] `backendRef`s that are a direct endpoint rather than a logical
  name are currently ignored
- [ ] integration tests for client policy have yet to be written
- [ ] only traffic splitting is currently implemented (no other policies
  can be set)
- [ ] per route metrics?
- [ ] client policy and Service Profile API lookups are currently run
  sequentially, but could be parallelized (with some rewriting of the
  current profile discovery code)


linkerd/app/gateway/src/gateway.rs
```diff
@@ -98,6 +98,7 @@ where
                 profile,
                 protocol: http.version,
                 logical_addr,
+                orig_dst: todo!("eliza: fix this"),
```
hawkw, line 101: i punted on fixing the gateway stuff for now, since this is just a rough sketch.

linkerd/app/outbound/src/http/logical.rs
```diff
@@ -172,15 +186,54 @@ impl<E> Outbound<E> {
...
+                let policy = logical.push_map_target(|(policy, logical): (Policy, Logical)| {
+                    // for now, throw away the policy and continue to the
+                    // logical stack.
+                    drop(policy);
+                    // TODO(eliza): this is where the stack used when a client
+                    // policy is resolved will go...
+                    logical
+                })
+                .instrument(|(policy, _): &(Policy, Logical)| debug_span!("policy", addr = %policy.dst))
+                .check_new_service::<(Policy, Logical), http::Request<_>>();
```
hawkw, line 200: this is where client policy middleware would go...

linkerd/app/outbound/src/policy/api.rs
```diff
@@ -0,0 +1,110 @@
...
+        Box::pin(async move {
+            let rsp = client.watch(target).await?;
+            Ok(rsp.map(|updates| {
+                updates
+                    .map_ok(move |up| {
+                        // If the server returned an invalid client policy, we
+                        // default to using an invalid policy. Currently, this
+                        // doesn't cause an internal error, because we don't have
+                        // filters in client policy rules yet...
+                        // TODO(eliza): this should get an internal error filter...
```
hawkw, line 81: how do we want to handle invalid client policies?

linkerd/app/src/env.rs
```diff
@@ -419,6 +419,81 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
+            connect,
+            buffer_capacity,
+        }
+    });
+
+    // The workload, which is opaque from the proxy's point-of-view, is sent to the
+    // policy controller to support policy discovery.
+    // XXX(eliza): this assumes that the workload identifier env var will be
+    // changed to include the pod name, so that it can be used to resolve client
+    // policies as well as server policies...
```
hawkw, line 485: do we want to use the same workload identifier for client and server policies?

adleong:
> The API response for server policies includes metadata for both the
entire server policy and for the individual routes in that policy.
On the other hand, the client policy API response only includes
metadata for HTTP routes. Presumably we will want to add metadata for
the whole policy message as well?

It looks like in the "inbound" API, there is metadata for each authorization and each httproute.  This makes sense because they correspond to k8s resources (e.g. to AuthorizationPolicy and HttpRoute).  For "outbound", we attach HttpRoute resources to Services directly.  We could have a top level metadata field which describes the Service resource, but this doesn't seem super useful to me.

hawkw:
> We could have a top level metadata field which describes the Service resource, but this doesn't seem super useful to me.

hmm, yeah, i suppose the service would already be recorded in other labels..

Cargo.lock
```diff
@@ -85,6 +85,51 @@ version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
+[[package]]
+name = "axum"
```
olix0r, line 89: Why do we need an axum dependency?

hawkw: no idea, i'll try to figure out where that came from...

hawkw: apparently the `axum` dep is via `tonic`:
<details>

<summary>`cargo tree` output</summary>

```
:; cargo tree -p axum -i
axum v0.5.17
└── tonic v0.8.1
    ├── linkerd-app v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app)
    │   ├── linkerd-app-integration v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/integration)
    │   └── linkerd2-proxy v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd2-proxy)
    ├── linkerd-app-core v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/core)
    │   ├── linkerd-app v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app) (*)
    │   ├── linkerd-app-admin v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/admin)
    │   │   └── linkerd-app v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app) (*)
    │   │   [dev-dependencies]
    │   │   └── linkerd-app-integration v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/integration)
    │   ├── linkerd-app-gateway v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/gateway)
    │   │   └── linkerd-app v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app) (*)
    │   ├── linkerd-app-inbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/inbound)
    │   │   ├── linkerd-app v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app) (*)
    │   │   ├── linkerd-app-admin v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/admin) (*)
    │   │   └── linkerd-app-gateway v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/gateway) (*)
    │   ├── linkerd-app-integration v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/integration)
    │   ├── linkerd-app-outbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/outbound)
    │   │   ├── linkerd-app v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app) (*)
    │   │   └── linkerd-app-gateway v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/gateway) (*)
    │   └── linkerd-app-test v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/test)
    │       └── linkerd-app-integration v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/integration)
    │       [dev-dependencies]
    │       ├── linkerd-app-gateway v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/gateway) (*)
    │       ├── linkerd-app-inbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/inbound) (*)
    │       └── linkerd-app-outbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/outbound) (*)
    ├── linkerd-app-inbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/inbound) (*)
    ├── linkerd-app-integration v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/integration)
    ├── linkerd-app-outbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/outbound) (*)
    ├── linkerd-client-policy v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/client-policy)
    │   └── linkerd-app-outbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/outbound) (*)
    ├── linkerd-opencensus v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/opencensus)
    │   ├── linkerd-app v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app) (*)
    │   └── linkerd-app-core v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/core) (*)
    ├── linkerd-proxy-api-resolve v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/proxy/api-resolve)
    │   ├── linkerd-app-core v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/core) (*)
    │   └── linkerd-service-profiles v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/service-profiles)
    │       └── linkerd-app-core v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/core) (*)
    ├── linkerd-proxy-identity-client v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/proxy/identity-client)
    │   └── linkerd-app-core v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/core) (*)
    ├── linkerd-proxy-tap v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/proxy/tap)
    │   └── linkerd-app-core v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/core) (*)
    ├── linkerd-service-profiles v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/service-profiles) (*)
    ├── linkerd-tonic-watch v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/tonic-watch)
    │   ├── linkerd-app-inbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/inbound) (*)
    │   ├── linkerd-app-outbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/outbound) (*)
    │   └── linkerd-service-profiles v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/service-profiles) (*)
    ├── linkerd2-proxy-api v0.7.0 (https://github.com/linkerd/linkerd2-proxy-api?branch=alex/outbound#9cfb93f3)
    │   ├── linkerd-app-inbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/inbound) (*)
    │   ├── linkerd-app-integration v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/integration)
    │   ├── linkerd-app-outbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/outbound) (*)
    │   ├── linkerd-client-policy v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/client-policy) (*)
    │   ├── linkerd-http-route v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/http-route)
    │   │   ├── linkerd-client-policy v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/client-policy) (*)
    │   │   └── linkerd-server-policy v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/server-policy)
    │   │       ├── linkerd-app-core v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/core) (*)
    │   │       └── linkerd-app-inbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/inbound) (*)
    │   ├── linkerd-policy-core v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/policy-core)
    │   │   ├── linkerd-client-policy v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/client-policy) (*)
    │   │   └── linkerd-server-policy v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/server-policy) (*)
    │   ├── linkerd-proxy-api-resolve v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/proxy/api-resolve) (*)
    │   ├── linkerd-proxy-identity-client v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/proxy/identity-client) (*)
    │   ├── linkerd-proxy-tap v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/proxy/tap) (*)
    │   ├── linkerd-server-policy v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/server-policy) (*)
    │   └── linkerd-service-profiles v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/service-profiles) (*)
    │   [dev-dependencies]
    │   ├── linkerd-proxy-tap v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/proxy/tap) (*)
    │   └── linkerd-service-profiles v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/service-profiles) (*)
    └── opencensus-proto v0.1.0 (/home/eliza/Code/linkerd2-proxy/opencensus-proto)
        └── linkerd-opencensus v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/opencensus) (*)
```

</details>

which is interesting, because i _thought_ all of our `tonic` deps disabling defautl features would avoid this, and this PR doesn't add any new `tonic` deps that don't have `default-features = false` AFAICT. hmm.

adleong:
Food for thought:

If I understand correctly, in the proxy we do a profile lookup in order to get a logical addr (i.e. Service DNS name), but then throw that away and instead do a client policy lookup using the orig dst.  The controller uses that target ip address and does a lookup in a Service index to find the Service name with that IP.  This is redundant.

Would it make sense to move the client policy lookup earlier in the proxy: before the lookup of the logical addr?  I think this helps us move toward a world where the service profile API isn't used at all when a client policy exists.

linkerd/app/src/env.rs
```diff
@@ -441,46 +516,39 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
+            Some(control) => outbound::policy::Config {
+                workload: workload.clone(),
+                control,
+            },
+            None => {
+                // TODO(eliza): should we just disable client policy if there's no
+                // policy service address?
+                error!("{}_ADDR must be set", ENV_POLICY_SVC_BASE);
+                return Err(EnvError::InvalidEnvVar);
+            }
```
hawkw, line 534: should we require that a policy service address be provided, or just disable client policy discovery if there's no policy service addr?

hawkw:
updated the PR description to reflect the current state of this branch

linkerd/client-policy/src/http/router.rs
```diff
@@ -0,0 +1,166 @@
...
+    pub fn layer() -> impl layer::Layer<N, Service = Self> + Clone {
+        layer::mk(Self)
+    }
+}
+
+impl<T, N> NewService<T> for NewServiceRouter<N>
+where
+    T: Param<Option<Receiver>> + Clone,
+    N: NewService<(Option<RoutePolicy>, T)> + Clone,
+{
```
olix0r, line 47: Do we really need to handle these as optional? We should use the stack to enforce that these are present; and *all* requests must have routes that go through this service, so I'd rather not promulgate optionality to the inner stack.

olix0r: I know this is different from how it's handled in service profiles; but I think we want to drive towards a state where there is a route description on all requests.

linkerd/client-policy/src/http/router.rs
```diff
@@ -0,0 +1,166 @@
...
+                    route.name = %policy.meta.name(),
+                    "req.match" = ?req_match,
+                    "Using HTTPRoute service",
+                );
+
+                self.services.get(policy).expect("route must exist").clone()
+            }
+            None => {
+                tracing::trace!("No HTTPRoutes matched");
+                self.default.clone()
```
olix0r, line 145: I'm pretty sure we have to error here in a way that causes a 404 from the error handler.

hawkw: ah, yup, we can change it to do that!

linkerd/client-policy/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+pub struct Receiver {
+    inner: watch::Receiver<ClientPolicy>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ClientPolicy {
+    pub http_routes: Arc<[http::Route]>,
+    pub meta: Arc<Meta>,
+    pub backends: Vec<split::Backend>,
+}
```
olix0r, line 23: Echoing my comments on https://github.com/linkerd/linkerd2-proxy-api/pull/165#discussion_r1042661341, we're going to have to set this up so that HTTP is just one flavor of it.

Also, what does the `Meta` on this resource correspond to?  On the `Server`Policy` side, it indicates the `Server`. What does it mean here?

hawkw: > Echoing my comments on [linkerd/linkerd2-proxy-api#165 (comment)](https://github.com/linkerd/linkerd2-proxy-api/pull/165#discussion_r1042661341), we're going to have to set this up so that HTTP is just one flavor of it.

Yes, my expectation is that the `ClientPolicy` type would grow additional fields for e.g. `grpc_routes` and `tcp_routes` when support for other protocols are added to the `outbound` gRPC API.

> Also, what does the `Meta` on this resource correspond to? On the `ServerPolicy` side, it indicates the `Server`. What does it mean here?

Right now it's un-populated and should probably be removed. It might be desirable to have the name of the Service that the policy corresponds to, but that's already included in other metrics...I think the top-level metadata here ought to go away, at least until we decide how we want metrics for client policy to work?

hawkw: > Echoing my comments on [linkerd/linkerd2-proxy-api#165 (comment)](https://github.com/linkerd/linkerd2-proxy-api/pull/165#discussion_r1042661341), we're going to have to set this up so that HTTP is just one flavor of it.

Yes, my expectation is that the `ClientPolicy` type would grow additional fields for e.g. `grpc_routes` and `tcp_routes` when support for other protocols are added to the `outbound` gRPC API.

> Also, what does the `Meta` on this resource correspond to? On the `ServerPolicy` side, it indicates the `Server`. What does it mean here?

Right now it's un-populated and should probably be removed. It might be desirable to have the name of the Service that the policy corresponds to, but that's already included in other metrics...I think the top-level metadata here ought to go away, at least until we decide how we want metrics for client policy to work?

hawkw: Removed the `meta` field here in 9fe2284a8a86328c925bdccd1ea2c732ee61b766.

linkerd/client-policy/src/http/router.rs
hawkw: this now 404s as of fc32de7a3ea7c57229e5f716730203ae7b93ef71

linkerd/client-policy/src/http/router.rs
hawkw: the route is no longer optional as of fc32de7a3ea7c57229e5f716730203ae7b93ef71; the policy rx is still an `Option`, but i _could_ change this to operate on a separate target type (there's already a TODO about this)...

linkerd/app/outbound/src/http/logical.rs
```diff
@@ -172,15 +173,76 @@ impl<E> Outbound<E> {
...
-                .push(svc::ArcNewService::layer())
+                .push(profiles::http::NewServiceRouter::layer());
+
+            // for now, the client-policy route stack is just a fixed traffic split
+            let policy = concrete
+                .check_new_service::<(ConcreteAddr, Logical), _>()
+                .push_map_target(|(concrete, PolicyRoute { route: _, logical }): (ConcreteAddr, PolicyRoute)| {
+                    (concrete, logical)
+                })
+                .push(policy::split::layer())
```
olix0r, line 184: I may be wrong, but my reading of this means that the policy split creates an inner stack for every backend of every route. Does this mean that multiple routes to a single backend service do not share load balancers/connections? Or where are the load balancers cached if not here?

hawkw: Yes, that's correct; we should probably add a caching layer here so that each concrete address's client stack is shared across multiple routes that share that backend.

When we add more forms of per-route policy, though (e.g. different retry policies, HTTPRoute filters, and per route metrics), we'll want to ensure that the portions of that stack that may differ based on which route is used are before the client stack cache, though.

linkerd/app/outbound/src/http/logical.rs
```diff
@@ -14,6 +14,267 @@ use tracing::debug_span;
...
+                        )
+                        .push(svc::FailFast::layer("HTTPRoute", dispatch_timeout))
+                        .push_spawn_buffer(buffer_capacity),
+                )
+                .check_new_clone::<PolicyRoute>()
+                .push_map_target(|(route, logical)| {
+                    tracing::debug!(?route);
+                    PolicyRoute { route, logical }
+                })
+                .push(policy::http::NewServiceRouter::layer())
```
olix0r, line 154: If the concrete stack is cached, can we omit the cache/buffer around the split so that every route gets its own split?

If so, perhaps it makes sense to pull it all into one layer that is Logical-outside and Concrete-inside... We'll need to consider how/where per-route functionality is added for things like retries/tmeouts.

PR #2057, opened 2022-12-10
release: Produce static binaries

olix0r:
This change adds release targets for static binaries (using musl).

With this change, we retain debug symbols in an alternate file.

This change bumps the devcontainer to v38 to include just-cargo fixes.

### DIFF
--- a/justfile
+++ b/justfile
@@ -12,7 +12,6 @@ export PROTOC_NO_VENDOR := "1"
 # By default we compile in development mode mode because it's faster.
 profile := if env_var_or_default("RELEASE", "") == "" { "debug" } else { "release" }
 toolchain := ""
-_cargo := "just-cargo profile=" + profile + " toolchain=" + toolchain
 
 features := ""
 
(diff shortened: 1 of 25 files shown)


justfile
```diff
@@ -106,27 +111,39 @@ test-dir dir *flags:
...
+    @{{ _cargo }} build --frozen --package=linkerd2-proxy {{ _features }}
+
+_strip:
+    {{ _objcopy }} --only-keep-debug {{ _target_bin }} {{ _target_bin }}.dbg
+    {{ _objcopy }} --strip-unneeded {{ _target_bin }}
+    {{ _objcopy }} --add-gnu-debuglink={{ _target_bin }}.dbg {{ _target_bin }}
+
+_package_bin := _package_dir / "bin" / "linkerd2-proxy"
+
+# XXX {aarch64,arm}-musl builds do not enable PIE, so we can't use
```
hawkw, line 125: so we can't use...what? is this comment unfinished?

hawkw (approved):
This looks good to me!

We may want to consider switching the `rust-toolchain` file to use the [TOML format](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file), so that we can define which targets we need and have `rustup` install them automatically (it can also automagically install `llvm-tools-preview`). This may be less important given that most folks are using the devcontainer, which also has all the appropriate Rust toolchains in it, but it might be worth doing regardless?

olix0r:
> We may want to consider switching the rust-toolchain file to use the [TOML format](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file), so that we can define which targets we need and have rustup install them automatically (it can also automagically install llvm-tools-preview). This may be less important given that most folks are using the devcontainer, which also has all the appropriate Rust toolchains in it, but it might be worth doing regardless?

Let's defer this for now.

1. I don't think it's important/valuable to fetch alternate targets. This is only needed for release, and is wasteful during development.
2. The rust (dev) container provides a stable set of llvm tooling. If we're going to include it, it should be done there so we don't have skew across environments.

PR #2078, opened 2022-12-20
config: Decouple HTTP and TCP buffering config

olix0r:
Proxies may buffer TCP connections while waiting for policy discovery
and may buffer HTTP requests while waiting for a shared resource (like a
load balancer).

Previously, a single configuration was used to configure both TCP and
HTTP buffers. This change decouples these configurations in preparation
for allowing balancer configuration to be configured by the control
plane.

Furthermore, this change updates the stack builder to always construct
buffers with failfast and spawnready to (1) ensure that all buffers
enforce the proper backpressure and load shedding semantics and (2) to
reduce boilerplate.

This change updates the proxy's environment configuration as follows:

* Removed `LINKERD2_PROXY_INBOUND_DISPATCH_TIMEOUT`
* Removed `LINKERD2_PROXY_OUTBOUND_DISPATCH_TIMEOUT`
* Added `LINKERD2_PROXY_INBOUND_DISCOVERY_IDLE_TIMEOUT`
* Added `LINKERD2_PROXY_OUTBOUND_DISCOVERY_IDLE_TIMEOUT`
* Removed `LINKERD2_PROXY_BUFFER_CAPACITY`
* Removed `LINKERD2_PROXY_INBOUND_ROUTER_MAX_IDLE_AGE`
* Removed `LINKERD2_PROXY_OUTBOUND_ROUTER_MAX_IDLE_AGE`
* Added `LINKERD2_PROXY_INBOUND_HTTP_QUEUE_CAPACITY`
* Added `LINKERD2_PROXY_INBOUND_HTTP_FAILFAST_TIMEOUT`
* Added `LINKERD2_PROXY_OUTBOUND_TCP_QUEUE_CAPACITY`
* Added `LINKERD2_PROXY_OUTBOUND_TCP_FAILFAST_TIMEOUT`
* Added `LINKERD2_PROXY_OUTBOUND_HTTP_QUEUE_CAPACITY`
* Added `LINKERD2_PROXY_OUTBOUND_HTTP_FAILFAST_TIMEOUT`
* Added `LINKERD2_PROXY_OUTBOUND_HTTP1_CONNECTION_POOL_IDLE_TIMEOUT`

### DIFF
--- a/linkerd/app/core/src/svc.rs
+++ b/linkerd/app/core/src/svc.rs
@@ -195,6 +195,7 @@ impl<S> Stack<S> {
         self.push(http::insert::NewResponseInsert::layer())
     }
 
+    /// Buffers requests in an mpsc, spawning the inner service onto a dedicated task.
     pub fn push_buffer_on_service<Req>(
         self,
         name: &'static str,
--- a/linkerd/app/core/src/svc.rs
+++ b/linkerd/app/core/src/svc.rs
@@ -11,12 +11,14 @@ pub use linkerd_stack::{
     NewRouter, NewService, Param, Predicate, UnwrapOr,
 };
 pub use linkerd_stack_tracing::{GetSpan, NewInstrument, NewInstrumentLayer};
+use stack::OnService;
 use std::{
+    marker::PhantomData,
     task::{Context, Poll},
     time::Duration,
 };
 use tower::{
-    buffer::{Buffer as TowerBuffer, BufferLayer},
+    buffer::Buffer as TowerBuffer,
     layer::util::{Identity, Stack as Pair},
     make::MakeService,
 };
@@ -29,6 +31,13 @@ pub use tower::{
 pub struct AlwaysReconnect(ExponentialBackoff);
 
 pub type Buffer<Req, Rsp, E> = TowerBuffer<BoxService<Req, Rsp, E>, Req>;
+
+pub struct BufferLayer<Req> {
+    name: &'static str,
+    capacity: usize,
+    failfast_timeout: Duration,
+    _marker: PhantomData<fn(Req)>,
+}
 
 pub type BoxHttp<B = http::BoxBody> =
     BoxService<http::Request<B>, http::Response<http::BoxBody>, Error>;
@@ -80,15 +89,16 @@ impl<L> Layers<L> {
     }
 
     /// Buffers requests in an mpsc, spawning the inner service onto a dedicated task.
-    pub fn push_spawn_buffer<Req>(
+    pub fn push_buffer<Req>(
         self,
+        name: &'static str,
         capacity: usize,
-    ) -> Layers<Pair<Pair<L, BoxServiceLayer<Req>>, BufferLayer<Req>>>
+        failfast_timeout: Duration,
+    ) -> Layers<Pair<L, BufferLayer<Req>>>
     where
         Req: Send + 'static,
     {
-        self.push(BoxServiceLayer::new())
-            .push(BufferLayer::new(capacity))
+        self.push(buffer(name, capacity, failfast_timeout))
     }
 
     pub fn push_on_service<U>(self, layer: U) -> Layers<Pair<L, stack::OnServiceLayer<U>>> {
@@ -152,22 +162,6 @@ impl<S> Stack<S> {
         self.push(NewReconnect::layer(AlwaysReconnect(backoff)))
     }
 
-    /// Buffer requests when when the next layer is out of capacity.
-    pub fn spawn_buffer<Req, Rsp>(
-        self,
-        capacity: usize,
-    ) -> Stack<Buffer<Req, S::Response, S::Error>>
-    where
-        Req: Send + 'static,
-        S: Service<Req> + Send + 'static,
-        S::Response: Send + 'static,
-        S::Error: Into<Error> + Send + Sync + 'static,
-        S::Future: Send,
-    {
-        self.push(BoxServiceLayer::new())
-            .push(BufferLayer::new(capacity))
-    }
-
     /// Assuming `S` implements `NewService` or `MakeService`, applies the given
     /// `L`-typed layer on each service produced by `S`.
     pub fn push_on_service<L: Clone>(self, layer: L) -> Stack<stack::OnService<L, S>> {
@@ -200,6 +194,18 @@ impl<S> Stack<S> {
         self,
     ) -> Stack<http::insert::NewResponseInsert<P, S>> {
         self.push(http::insert::NewResponseInsert::layer())
+    }
+
+    pub fn push_buffer_on_service<Req>(
+        self,
+        name: &'static str,
+        capacity: usize,
+        failfast_timeout: Duration,
+    ) -> Stack<OnService<BufferLayer<Req>, S>>
+    where
+        Req: Send + 'static,
+    {
+        self.push_on_service(buffer(name, capacity, failfast_timeout))
     }
 
     pub fn push_cache<T>(self, idle: Duration) -> Stack<cache::NewCachedService<T, S>>
@@ -379,5 +385,49 @@ impl<E: Into<Error>> Recover<E> for AlwaysReconnect {
 
     fn recover(&self, _: E) -> Result<Self::Backoff, E> {
         Ok(self.0.stream())
+    }
+}
+
+// === impl BufferLayer ===
+
+fn buffer<Req>(
+    name: &'static str,
+    capacity: usize,
+    failfast_timeout: Duration,
+) -> BufferLayer<Req> {
+    BufferLayer {
+        name,
+        capacity,
+        failfast_timeout,
+        _marker: PhantomData,
+    }
+}
+
+impl<Req, S> Layer<S> for BufferLayer<Req>
+where
+    Req: Send + 'static,
+    S: Service<Req, Error = Error> + Send + 'static,
+    S::Future: Send,
+{
+    type Service = Buffer<Req, S::Response, Error>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        Buffer::new(
+            BoxService::new(
+                FailFast::layer(self.name, self.failfast_timeout).layer(SpawnReady::new(inner)),
+            ),
+            self.capacity,
+        )
+    }
+}
+
+impl<Req> Clone for BufferLayer<Req> {
+    fn clone(&self) -> Self {
+        Self {
+            capacity: self.capacity,
+            name: self.name,
+            failfast_timeout: self.failfast_timeout,
+            _marker: self._marker,
+        }
     }
 }
--- a/linkerd/app/core/src/svc.rs
+++ b/linkerd/app/core/src/svc.rs
@@ -1,7 +1,7 @@
 // Possibly unused, but useful during development.
 
 pub use crate::proxy::http;
-use crate::{cache, Error};
+use crate::{cache, config::BufferConfig, Error};
 use linkerd_error::Recover;
 use linkerd_exp_backoff::{ExponentialBackoff, ExponentialBackoffStream};
 pub use linkerd_reconnect::NewReconnect;
@@ -92,13 +92,12 @@ impl<L> Layers<L> {
     pub fn push_buffer<Req>(
         self,
         name: &'static str,
-        capacity: usize,
-        failfast_timeout: Duration,
+        config: &BufferConfig,
     ) -> Layers<Pair<L, BufferLayer<Req>>>
     where
         Req: Send + 'static,
     {
-        self.push(buffer(name, capacity, failfast_timeout))
+        self.push(buffer(name, config))
     }
 
     pub fn push_on_service<U>(self, layer: U) -> Layers<Pair<L, stack::OnServiceLayer<U>>> {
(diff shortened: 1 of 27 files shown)


linkerd/app/gateway/src/lib.rs
```diff
@@ -158,11 +152,13 @@ where
                         .stack
                         .layer(metrics::StackLabels::inbound("tcp", "gateway")),
                 )
-                .push(svc::layer::mk(svc::SpawnReady::new))
-                .push(svc::FailFast::layer("TCP Gateway", dispatch_timeout))
-                .push_spawn_buffer(buffer_capacity),
+                .push_buffer(
+                    "TCP Gatweay",
+                    inbound_config.tcp_server_buffer.capacity,
+                    inbound_config.tcp_server_buffer.failfast_timeout,
```
hawkw, line 158: @olix0r wdyt of having this take a `BufferConfig` rather than having to explicitly do this?

linkerd/app/inbound/src/http/server.rs
```diff
@@ -58,15 +56,15 @@ impl<H> Inbound<H> {
...
-                        .push(svc::FailFast::layer("HTTP Server", dispatch_timeout)),
+                        .push(http::orig_proto::Downgrade::layer()),
+                    // Limit the number of in-flight requests. When the proxy is
+                    // at capacity, go into failfast after a dispatch timeout.
+                    // Note that the inner service _always_ returns ready (due
+                    // to `NewRouter`) and the concurrency limit need not be
+                    // driven outside of the request path, so there's no need
+                    // for SpawnReady
+                    //.push(svc::ConcurrencyLimitLayer::new(max_in_flight_requests))
+                    //.push(svc::FailFast::layer("HTTP Server", dispatch_timeout)),
```
olix0r, line 67: FIXME

linkerd/app/core/src/svc.rs
```diff
@@ -202,6 +195,17 @@ impl<S> Stack<S> {
         self.push(http::insert::NewResponseInsert::layer())
     }
 
+    pub fn push_buffer_on_service<Req>(
```
hawkw, line 198: tiny nit, unimportant: `push_buffer` has a doc comment, should we put one on this too? it's pretty obvious what it does from the name if the reader understands what `push_buffer` does, so idk if it's important...

linkerd/app/inbound/src/lib.rs
```diff
@@ -39,8 +39,18 @@ pub struct Config {
     pub allow_discovery: NameMatch,
     pub proxy: ProxyConfig,
     pub policy: policy::Config,
-    pub profile_idle_timeout: Duration,
     pub allowed_ips: transport::AllowIps,
+
+    /// Configures how long the proxy will wait for a profile response for
+    /// skipping route information.
```
hawkw, line 45: tiny nit: this wording feels a little unclear?
```suggestion
    /// Configures how long the proxy will wait for a profile response. If no
    /// profile response is received before this timeout elapses, route information
    /// is skipped.
```
or even just
```suggestion
    /// Configures how long the proxy will wait for a profile response before
    /// skipping route information.
```

linkerd/app/outbound/src/http/logical.rs
```diff
@@ -18,92 +24,127 @@ impl<N> Outbound<N> {
...
                 )
-                .push_cache(cache_max_idle_age);
+                // TODO(ver) this should not be a generalized time-based evicting cache.
+                .push_cache(*discovery_idle_timeout);
 
             // If there's no route, use the logical service directly; otherwise
             // use the per-route stack.
-            logical
-                .clone()
+            let route = (split.clone())
```
hawkw, line 67: huh, what are the parens doing here? AFAICT just writing `split.clone()` would be equivalent?

hawkw: (this isn't a big deal; i was just kinda curious about it)

linkerd/app/outbound/src/lib.rs
```diff
@@ -49,9 +49,27 @@ const EWMA_DECAY: Duration = Duration::from_secs(10);
...
     pub allow_discovery: AddrMatch,
 
+    pub proxy: ProxyConfig,
+
+    /// Configures the duration the proxy will retain idle stacks (with no
+    /// active connections) for an outbound address. When an idle stack is
+    /// dropped, all cached service discovery information is dropped.
+    pub discovery_idle_timeout: Duration,
+
+    /// Configures how connections are buffered *for each outbound address*.
```
hawkw, line 61: nit:
```suggestion
    /// Configures how many connections are buffered *for each outbound address*.
```

linkerd/app/outbound/src/lib.rs
```diff
@@ -49,9 +49,27 @@ const EWMA_DECAY: Duration = Duration::from_secs(10);
...
+    ///
+    /// A buffer capacity of 100 eans that 100 connections may be buffered for
```
hawkw, line 63: typo:
```suggestion
    /// A buffer capacity of 100 means that 100 connections may be buffered for
```

linkerd/app/outbound/src/lib.rs
```diff
@@ -49,9 +49,27 @@ const EWMA_DECAY: Duration = Duration::from_secs(10);
...
+    /// each IP:port to which an application attempts to connect.
+    pub tcp_connection_buffer: BufferConfig,
+
+    /// Configures how HTTP requests are buffered *for each outbound address*.
+    ///
+    /// A buffer capacity of 100 eans that 100 connections may be buffered for
+    /// each IP:port to which an application attempts to connect.
```
hawkw, line 70: typos + fixing the slightly wrong copypasta:
```suggestion
    /// Configures how many HTTP requests are buffered *for each outbound address*.
    ///
    /// A buffer capacity of 100 means that 100 requests may be buffered for
    /// each IP:port to which an application has opened an outbound TCP connection.
```

linkerd/app/src/env.rs
```diff
@@ -96,8 +96,13 @@ pub const ENV_METRICS_RETAIN_IDLE: &str = "LINKERD2_PROXY_METRICS_RETAIN_IDLE";
 
 const ENV_INGRESS_MODE: &str = "LINKERD2_PROXY_INGRESS_MODE";
 
-const ENV_INBOUND_DISPATCH_TIMEOUT: &str = "LINKERD2_PROXY_INBOUND_DISPATCH_TIMEOUT";
-const ENV_OUTBOUND_DISPATCH_TIMEOUT: &str = "LINKERD2_PROXY_OUTBOUND_DISPATCH_TIMEOUT";
+const ENV_INBOUND_HTTP_BUFFER_CAPACITY: &str = "LINKERD2_PROXY_INBOUND_HTTP_BUFFER_CAPACITY";
+const ENV_INBOUND_HTTP_FAILFAST_TIMEOUT: &str = "LINKERD2_PROXY_INBOUND_HTTP_FAILFAST_TIMEOUT";
+
+const ENV_OUTBOUND_TCP_BUFFER_CAPACITY: &str = "LINKERD2_PROXY_OUTBOUND_TCP_BUFFER_CAPACITY";
```
hawkw, line 102: nit/tioli: not sure about calling this "TCP_BUFFER_CAPACITY" --- it seems like it could be misinterpreted to mean the capacity of the TCP send/recv buffers? maybe we should be calling these `QUEUE_CAPACITY` instead? that would be using a different name than `tower::buffer`, but it seems less easy to mix up with other "TCP buffer" configurations...

hawkw: or even "TCP_CONNECT_BUFFER" or something, maybe?

linkerd/app/src/env.rs
```diff
@@ -248,26 +261,34 @@ const DEFAULT_INITIAL_CONNECTION_WINDOW_SIZE: u32 = 1048576; // 1MB ~ 16 streams
...
-//
-// On the inbound side, we want to be a bit more permissive so that periodic, as
-// the number of endpoints should generally be pretty constrained.
-const DEFAULT_INBOUND_ROUTER_MAX_IDLE_AGE: Duration = Duration::from_secs(20);
-const DEFAULT_OUTBOUND_ROUTER_MAX_IDLE_AGE: Duration = Duration::from_secs(5);
+// the application communicates with many destinations.
+const ENV_OUTBOUND_DISCOVERY_IDLE_TIMEOUT: &str = "LINKERD2_PROXY_OUTBOUND_DISCOVERY_IDLE_TIMEOUT";
+const DEFAULT_OUTBOUND_DISCOVERY_IDLE_TIMEOUT: Duration = Duration::from_secs(5);
+
+// On the inbound side, we may lookup per-port policy or per-service profile
```
hawkw, line 275: tiny nit: should this be
```suggestion
// On the inbound side, we may lookup per-port policy or per-service-profile
```

linkerd/stack/src/switch_ready.rs
```diff
@@ -78,7 +78,12 @@ impl<A, B> SwitchReady<A, B> {
     /// the `secondary` service is used until the primary service becomes ready again.
     pub fn new(primary: A, secondary: B, switch_after: time::Duration) -> Self {
         Self {
-            primary,
+            // The primary service is wrapped in a `SpawnReady` so that it can
+            // still become ready even when we've reverted to using the
+            // secondary service.
+            primary: SpawnReady::new(primary),
+            // The secondary service is not wrapped because we don't really care
+            // about driving it to readiness unless the primary has timed out.
```
hawkw, line 86: :thumbsup:

hawkw (approved):
overall, this looks good to me! i had some minor nits about naming and documentation, but other than that, this looks great.

are we going to need to make a proxy-injector change to set the new env vars, as well?

linkerd/app/outbound/src/lib.rs
olix0r: It's not just how many, it's also for how long. So I think the original is correct... it configures the buffering behavior.

linkerd/app/outbound/src/http/logical.rs
olix0r: yeah but it will be

```rust
let route = split
    .clone()
    ...
```

so this is just tricking rustfmt. probably not worth it.

linkerd/app/outbound/src/lib.rs
hawkw: mmm, okay, yeah, nevermind.

linkerd/app/outbound/src/http/logical.rs
hawkw: no worries, i was just wondering if it was load-bearing!

linkerd/app/src/env.rs
olix0r: No, we may lookup per-service serviceprofiles.

linkerd/app/src/env.rs
olix0r: i like queue.

olix0r:
> are we going to need to make a proxy-injector change to set the new env vars, as well?

The proxy injector doesn't (yet) provide configuration related to the changed values:

```:; k get po -n emojivoto web-6c5b7bb685-x6688 -o yaml |grep PROXY_
    - name: LINKERD2_PROXY_LOG
    - name: LINKERD2_PROXY_LOG_FORMAT
    - name: LINKERD2_PROXY_DESTINATION_SVC_ADDR
    - name: LINKERD2_PROXY_DESTINATION_PROFILE_NETWORKS
    - name: LINKERD2_PROXY_POLICY_SVC_ADDR
    - name: LINKERD2_PROXY_POLICY_WORKLOAD
    - name: LINKERD2_PROXY_INBOUND_DEFAULT_POLICY
    - name: LINKERD2_PROXY_POLICY_CLUSTER_NETWORKS
    - name: LINKERD2_PROXY_INBOUND_CONNECT_TIMEOUT
    - name: LINKERD2_PROXY_OUTBOUND_CONNECT_TIMEOUT
    - name: LINKERD2_PROXY_CONTROL_LISTEN_ADDR
    - name: LINKERD2_PROXY_ADMIN_LISTEN_ADDR
    - name: LINKERD2_PROXY_OUTBOUND_LISTEN_ADDR
    - name: LINKERD2_PROXY_INBOUND_LISTEN_ADDR
    - name: LINKERD2_PROXY_INBOUND_IPS
    - name: LINKERD2_PROXY_INBOUND_PORTS
    - name: LINKERD2_PROXY_DESTINATION_PROFILE_SUFFIXES
    - name: LINKERD2_PROXY_INBOUND_ACCEPT_KEEPALIVE
    - name: LINKERD2_PROXY_OUTBOUND_CONNECT_KEEPALIVE
    - name: LINKERD2_PROXY_INBOUND_PORTS_DISABLE_PROTOCOL_DETECTION
    - name: LINKERD2_PROXY_DESTINATION_CONTEXT
    - name: LINKERD2_PROXY_IDENTITY_DIR
    - name: LINKERD2_PROXY_IDENTITY_TRUST_ANCHORS
    - name: LINKERD2_PROXY_IDENTITY_TOKEN_FILE
    - name: LINKERD2_PROXY_IDENTITY_SVC_ADDR
    - name: LINKERD2_PROXY_IDENTITY_LOCAL_NAME
    - name: LINKERD2_PROXY_IDENTITY_SVC_NAME
    - name: LINKERD2_PROXY_DESTINATION_SVC_```

PR #2079, opened 2022-12-21
remove fixed inbound policy mode

hawkw:
Currently, the inbound proxy is capable of operating in two modes: a
fixed mode, where no policy controller address is provided, and inbound
port policies are configured by the environment variables
`LINKERD2_PROXY_INBOUND_PORTS_DISABLE_PROTOCOL_DETECTION`,
`LINKERD2_PROXY_INBOUND_PORTS_REQUIRE_IDENTITY`, and
`LINKERD2_PROXY_INBOUND_PORTS_REQUIRE_TLS`; and a discovery-based mode,
where port policies are discovered from the policy controller. In
practice, the fixed mode is never actually used, so we probably ought to
remove it entirely in 2.13.

This branch removes the fixed mode for inbound port policy. Starting the
proxy without the `LINKERD2_PROXY_POLICY_SVC_ADDR` and
`LINKERD2_PROXY_POLICY_SVC_IDENTITY` environment variables set is now an
error. If the `LINKERD2_INBOUND_PORTS_REQUIRE_IDENTITY` or
`LINKERD2_PROXY_INBOUND_PORTS_REQUIRE_TLS` environment variables are
present, the proxy will now log a warning and ignore their values. The
variable is still supported, however, in order to allow the proxy to
avoid performing policy controller lookups when configured with a large
list of opaque ports.

The `Store::spawn_fixed` method on `linkerd_app_inbound::policy::Store`
is no longer used when running the proxy normally, but it is retained
for testing purposes, and has been made test-only.

Finally, it was necessary to add a mock implementation of the policy
controller for integration tests to pass, since a policy controller is
now required. This is based on the implementation of a mock policy
controller added in #1992, but without the outbound policy API. PR #1992
should be able to rebase cleanly once this branch merges.

Closes #2076

### DIFF
--- a/linkerd/app/inbound/src/policy/config.rs
+++ b/linkerd/app/inbound/src/policy/config.rs
@@ -5,7 +5,10 @@ use tokio::time::Duration;
 
 /// Configures inbound policies.
 ///
-/// The proxy usually watches dynamic policies from the control plane.
+/// Most policies are watched dynamically from the control plane. A default
+/// policy is used when the controller does not provide a policy for a given
+/// port. In addition, a pre-configured list of opaque ports can be provided
+/// to avoid unnecessary policy lookups for ports which will always be opaque.
 #[derive(Clone, Debug)]
 pub struct Config {
     pub control: control::Config,
--- a/linkerd/app/inbound/src/policy/config.rs
+++ b/linkerd/app/inbound/src/policy/config.rs
@@ -1,27 +1,25 @@
 use super::{api::Api, DefaultPolicy, GetPolicy, Protocol, ServerPolicy, Store};
 use linkerd_app_core::{control, dns, identity, metrics, svc::NewService};
-use std::collections::{HashMap, HashSet};
+use std::collections::HashSet;
 use tokio::time::Duration;
 
 /// Configures inbound policies.
 ///
-/// The proxy usually watches dynamic policies from the control plane, though it can also use
-/// 'fixed' policies configured at startup.
+/// The proxy usually watches dynamic policies from the control plane.
 #[derive(Clone, Debug)]
-#[allow(clippy::large_enum_variant)]
-pub enum Config {
-    Discover {
-        control: control::Config,
-        workload: String,
-        default: DefaultPolicy,
-        cache_max_idle_age: Duration,
-        ports: HashSet<u16>,
-    },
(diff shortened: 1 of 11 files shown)


hawkw:
fuzzer build failure looks like maybe a nightly regression? https://github.com/linkerd/linkerd2-proxy/actions/runs/3752888848/jobs/6375571794

kleimkuhler (approved):
Looks good!

linkerd/app/inbound/src/policy/config.rs
```diff
@@ -1,27 +1,25 @@
...
 use linkerd_app_core::{control, dns, identity, metrics, svc::NewService};
-use std::collections::{HashMap, HashSet};
+use std::collections::HashSet;
 use tokio::time::Duration;
 
 /// Configures inbound policies.
 ///
-/// The proxy usually watches dynamic policies from the control plane, though it can also use
-/// 'fixed' policies configured at startup.
+/// The proxy usually watches dynamic policies from the control plane.
```
hawkw, line 8: update slightly wrong comment
```suggestion
/// Most policies are watched dynamically from the control plane. A default
/// policy is used when the controller does not provide a policy for a given
/// port. In addition, a pre-configured list of opaque ports can be provided
/// to avoid unnecessary policy lookups for ports which will always be opaque.
```

linkerd/app/inbound/src/policy/config.rs
```diff
@@ -1,27 +1,36 @@
...
 use tokio::time::Duration;
 
 /// Configures inbound policies.
 ///
-/// The proxy usually watches dynamic policies from the control plane, though it can also use
-/// 'fixed' policies configured at startup.
+/// Most policies are watched dynamically from the control plane. A default
+/// policy is used when the controller does not provide a policy for a given
+/// port. In addition, a pre-configured list of opaque ports can be provided
+/// to avoid unnecessary policy lookups for ports which will always be opaque.
```
olix0r, line 11: I don't think this is correct.

```suggestion
/// to avoid unnecessary policy lookups for ports which will always be opaque.
```

We still lookup policies for these... What we configure here is the _default_ behavior, i.e. on the first request before the resolution is complete.

We don't currently block inbound connections on a policy existing. That could change in the future, I guess...

PR #2080, opened 2022-12-21
allow port ranges in opaque ports environment variable

hawkw:
Depends on #2079

Currently, the `LINKERD2_PROXY_INBOUND_PORTS_DISABLE_PROTOCOL_DETECTION`
environment variable accepts a comma-delimited list of individual port
numbers. Each of these ports is then stored as a separate cache entry in
the port policy cache to indicate that the port is opaque.

Unfortunately, this does not work well when a very large number of ports
are marked as opaque, because the proxy-injector will generate a massive
value for that environment variable, listing every individual port. This
can cause issues when this manifest becomes larger than the Kubernetes
API can reasonably handle. In addition, huge numbers of ports will all
be kept in memory as a separate cache entry by the proxy, increasing
proxy memory usage. See linkerd/linkerd2#9803 for details.

This branch changes the proxy so that the
variable may contain a list of individual port numbers *and* port
ranges, which are specified as `<low>-<high>`.

Opaque ports are now stored using a `RangeInclusiveSet` from the
`rangemap` crate, rather than by creating individual cache entries for
each port. This means we are no longer storing a separate entry in the
cache for every port in a range, reducing memory consumption when there
are very large ranges of opaque ports.

This is the proxy half of the work towards resolving
linkerd/linkerd2#9803. Once this branch lands, we'll also have to change
the proxy-injector so that it no longer handles opaque port ranges by
generating a list of all the individual ports in the range, and simply
forwards those ranges as they were specified when generating the
environment variable.

### DIFF
--- a/linkerd/app/inbound/src/policy/store.rs
+++ b/linkerd/app/inbound/src/policy/store.rs
@@ -1,12 +1,14 @@
 use super::{api, config, AllowPolicy, DefaultPolicy, GetPolicy};
 use linkerd_app_core::{proxy::http, transport::OrigDstAddr, Error};
-use linkerd_cache::Cache;
+use linkerd_cache::{Cache, Cached};
 pub use linkerd_server_policy::{
     authz::Suffix, Authentication, Authorization, Protocol, ServerPolicy,
 };
+use rangemap::RangeInclusiveSet;
 use std::{
     collections::HashSet,
     hash::{BuildHasherDefault, Hasher},
+    sync::Arc,
 };
 use tokio::{sync::watch, time::Duration};
 use tracing::info_span;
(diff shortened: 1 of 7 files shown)


linkerd/app/inbound/src/policy/store.rs
```diff
@@ -82,26 +74,35 @@ impl<S> Store<S> {
...
                     .in_scope(|| discover.spawn_with_init(port, default.into()));
                 (port, rx)
             });
-            Cache::with_permanent_from_iter(idle_timeout, rxs.chain(opaque))
+            Cache::with_permanent_from_iter(idle_timeout, rxs)
         };
+        Self::new(cache, default, Some(discover), opaque_ports)
+    }
+
+    fn new(
```
olix0r, line 89: Prefer to make it clear it spawns things.

```suggestion
    fn spawn(
```

PR #2085, opened 2022-12-22
Introduce a 'distribute' stack module

olix0r:
This change introduces a new, currently unused, stack module that generalizes the functionality in `service_profiles::split` in anticipation of upcoming request routing functionality.

Unlike `split` the `Distribute` service is designed to be used with cloneable inner stacks. It does **not** handle dynamic updates. Updating is to be handled centrally by rebuilding stacks rather than updating them internally.

Furthermore, the distribute module now supports a variety of distribution strategies over a shared set of inner services. This will allow each route to change its load distribution strategy. The strategies include:

* Empty: causes requests the route to be unavailable
* First Available: selects the first available inner service, i.e. to implement failover.
* Random Available: randomly selects over a weighted set of inner services. Unavailable services are ignored.


linkerd/distribute/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+
+#[derive(Clone, Debug)]
+pub struct NewDistribute<K, S> {
+    backends: Arc<IndexMap<K, S>>,
+}
+
+#[derive(Debug)]
+pub struct Distribute<K, S> {
+    backends: Arc<IndexMap<K, S>>,
+    ready: Option<S>,
```
hawkw, line 28: naming nit, take it or leave it: i might have called this `next`, to more clearly indicate that it's the service that will be called by the next `call`? alternatively, a comment here might make it clearer what this is for...

linkerd/distribute/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+}
+
+#[derive(Debug, PartialEq)]
+pub struct WeightedKeys<K> {
+    keys: Vec<K>,
+    index: WeightedIndex<u32>,
+}
+
+#[derive(Debug)]
+enum Inner<K> {
```
hawkw, line 53: naming nit, take it or leave it: if we wanted to give this a more descriptive name, what do you think of `Selection`?

linkerd/distribute/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+
+            Inner::FirstAvailable { ref keys } => {
+                for key in keys.iter() {
+                    let svc = self
+                        .backends
+                        .get(key)
+                        .expect("distributions must not reference unknown backends");
+                    let mut svc = svc.clone();
+                    if svc.poll_ready(cx)?.is_ready() {
+                        self.ready = Some(svc);
```
hawkw, line 138: how will this interact with `Service`s where cloning clears readiness state? AFAICT, if the inner service's readiness is per-clone, and we return `Poll::Pending` because no inner service is ready, the task will not be woken by one of those inner services becoming ready, since the clones are all dropped immediately after polling them.

would it make more sense to use `get_mut` here, and poll the service instance stored in `self.backends`, rather than cloning it? we could change `self.ready` to store the _key_ [^1] of the ready service, rather than a clone of that service, and have `call` look up the service for that key.

[^1]: since the services live in an `IndexMap`, it would be a bit more efficient to use [`get_full_mut`](https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html#method.get_full_mut) and store the ready service's index rather than its key --- this is morally equivalent but has the advantage of not having to hash the key, and doesn't clone the key if the keys are expensive to clone (e.g. strings)...

linkerd/distribute/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+                while polled_idxs.len() != keys.keys.len() {
+                    let (idx, _, svc) = self
+                        .backends
+                        .get_full(keys.next(rng))
+                        .expect("distributions must not reference unknown backends");
+                    if polled_idxs.insert(idx) {
+                        // The index was not already polled, so poll it.
+                        let mut svc = svc.clone();
+                        if svc.poll_ready(cx)?.is_ready() {
+                            self.ready = Some(svc);
```
hawkw, line 163: same comment as in `FirstAvailable` regarding services with per-clone readiness state.

linkerd/distribute/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+}
+
+impl<K> Clone for Inner<K> {
+    fn clone(&self) -> Self {
+        match self {
+            Self::Empty => Inner::Empty,
+            Self::FirstAvailable { keys } => Self::FirstAvailable { keys: keys.clone() },
+            Self::RandomAvailable { keys, .. } => Self::RandomAvailable {
+                keys: keys.clone(),
+                polled_idxs: HashSet::with_capacity(keys.keys.len()),
```
hawkw, line 244: this could potentially be replaced with a bitset, which could probably avoid allocating in the case where the number of keys is <= 64 and a single integer could be used to represent all the keys. but that's fine to save for a future branch.

linkerd/distribute/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+};
+
+#[derive(Clone, Debug)]
+pub struct NewDistribute<K, S> {
+    backends: Arc<IndexMap<K, S>>,
+}
+
+#[derive(Debug)]
+pub struct Distribute<K, S> {
+    backends: Arc<IndexMap<K, S>>,
```
hawkw, line 27: i don't _think_ i actually see anywhere where the ordered nature of the `IndexMap` is used, is that correct?

(but see my comment later where i discuss a way where we could use `IndexMap` to our advantage if handling services that don't share readiness across clones is desirable...)

hawkw (commented):
overall, this looks great! i only noticed one potential issue. besides that, most of my comments were naming nits based on stuff that seemed unclear to some folks while we were walking through this code earlier.

also, it might be nice to mention somewhere in the doc comments that the backend list is intended to be shared across multiple distributions with different subsets of the backends, since that's an important detail that may not be immediately obvious...

linkerd/distribute/src/lib.rs
olix0r: that's an excellent point. We can't use get_*_mut currently because the index is arc'd to make cloning cheap. I think it's probably a premature optimization, though; so it would be better to clone the indexmap. (we expect the number of backends to be fairly small).

A prior version stored `ready_idx` instead of `ready`. I agree it probably makes sense to go back to that.

linkerd/distribute/src/lib.rs
olix0r: yeah, this was a remnant of storing indices instead of service clones...

linkerd/distribute/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+    backends: IndexMap<K, S>,
+}
+
+/// A service that distributes requests over a set of backends.
+#[derive(Debug)]
+pub struct Distribute<K, S> {
+    backends: IndexMap<K, S>,
+    selection: Selection<K>,
+
+    /// Stores the index of the backend that was has been polled to ready. The
```
hawkw, line 32: tiny typo here:
```suggestion
    /// Stores the index of the backend that has been polled to ready. The
```

linkerd/distribute/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+#[derive(Clone, Debug, PartialEq)]
+pub enum Distribution<K> {
+    /// A distribution that has no backends, and therefore never becomes ready.
+    Empty,
+
+    /// A distribution that uses the first available backend in an ordered list.
+    FirstAvailable(Arc<[K]>),
+
+    /// A distribution that uses the first available backend when randomly
+    /// selecting over a weighted distributino of backends.
```
hawkw, line 48: tiny typo here
```suggestion
    /// selecting over a weighted distribution of backends.
```

hawkw (commented):
noticed a couple little typos, but other than that, this is looking good IMO!

linkerd/distribute/src/params.rs
```diff
@@ -0,0 +1,80 @@
+use rand::distributions::{WeightedError, WeightedIndex};
+use std::sync::Arc;
+
+/// A parameter type that configures how a [`Distribute`] should behave.
```
hawkw, line 4: tiny nit: docs link is broken
```suggestion
/// A parameter type that configures how a [`Distribute`] should behave.
///
/// [`Distribute`]: crate::service::Distribute
```

linkerd/distribute/src/service.rs
```diff
@@ -0,0 +1,317 @@
...
+            // Choose a random index (via the weighted distribution) to try to
+            // poll the backend. Continue selecting endpoints until we find one
+            // that is ready or we've tried all backends in the distribution.
+            Selection::RandomAvailable {
+                ref keys,
+                ref mut rng,
+            } => {
+                // Clone the weighted index so that we can zero out the weights
+                // as pending services are polled.
+                let mut index = keys.index().clone();
```
hawkw, line 93: i wonder if we could reduce allocations slightly by cloning the `WeightedIndex` once per `Distribute` instance and resetting the zeroed weights every time we select a ready index, rather than cloning it once per `poll_ready`? this may be a premature optimization but it could be worth considering at some point...

linkerd/distribute/src/service.rs
```diff
@@ -0,0 +1,317 @@
...
+            Selection::Empty => {
+                tracing::debug!("empty distribution will never become ready");
+            }
+
+            Selection::FirstAvailable => {
+                for (idx, svc) in self.backends.values_mut().enumerate() {
+                    if svc.poll_ready(cx)?.is_ready() {
+                        self.ready_idx = Some(idx);
+                        return Poll::Ready(Ok(()));
+                    }
```
hawkw, line 80: this is probably not worth worrying about for now, but it occurs to me that it might be possible to have a scheme where, if we've polled all the backends and none are ready, when we get woken up by one becoming ready, we could know which index became ready rather than having to poll them all in order again by wrapping the `Context`'s waker with one that knows the index and tells us which one became ready. this would avoid a second O(_n_) polling loop.

probably not worth worrying about right now, but could be an interesting optimization for the future if we care about this.

linkerd/distribute/src/service.rs
```diff
@@ -0,0 +1,317 @@
...
+                            tracing::error!(%error, "unexpected error updating weights; giving up");
+                            break;
+                        }
+                    }
+                }
+            }
+        }
+
+        debug_assert!(self.ready_idx.is_none());
+        Poll::Pending
```
hawkw, line 125: nit: should we log something like
```suggestion
        tracing::trace!("no ready services in distribution");
        Poll::Pending
```
here?

linkerd/distribute/src/service.rs
```diff
@@ -0,0 +1,317 @@
...
+        assert_ready_ok!(call.poll());
+    }
+
+    #[test]
+    fn random_available_follows_availability() {
+        let (mock0, mut handle0) = mock::pair();
+        let (mock1, mut handle1) = mock::pair();
+        let mut dist_svc = mock::Spawn::new(Distribute::new(
+            vec![("foo", mock0), ("bar", mock1)].into_iter().collect(),
+            Distribution::random_available([("foo", 1), ("bar", 99999)]).unwrap(),
```
hawkw, line 302: i noticed all the `random_available` tests have two services...should we be testing cases with >2 services?

hawkw (approved):
This looks good to me overall! I left a couple more comments, but no blocking concerns.

If you like, it might be worth trying to change the `RandomAvailable` selection to reuse the same `WeightedIndex` clone for the whole lifetime of the service? But, we can also do that in a follow-up.

linkerd/distribute/src/service.rs
olix0r: it's not particularly easy to reset the weights... i think it's only worth yakshaving if we can identify problems with the easier approach

linkerd/distribute/src/service.rs
olix0r: Maybe? what cases do you think we're not covering in the above logic? It's already a little sketchy writing tests overweighted random distributions (this test will fail 0.00001% of the time).

linkerd/distribute/src/service.rs
olix0r: https://github.com/linkerd/linkerd2/issues/9892

olix0r: Yeah, it's complicated enough that I don't think it's worth it presently. We're not expecting this list to be large in the vast majority of situations. Clarity/correctness is far more important than shaving off a small allocation.

PR #2086, opened 2022-12-23
update logical router in the background task

hawkw:
This branch adds an implementation to the `todo!`'d background task for updating the logical router's shared state. The background task updates the set of backends, distribution, and routes when the profile watch changes. Code for actually generating different parts of the stack is split out into functions so that it can be shared between the `new_service` impl that generates the initial state of the router, and the background task itself.

It might be potentially worth investigating a way to avoid rebuilding parts of the profile when only some of the profile changes (e.g. if the set of targets remains constant but the set of routes changes, only rebuild the routes, and so on). However, I punted on this for now as it could potentially introduce a lot of complexity.

I'll address some of the other TODOs in `ver/concretemas` in subsequent PRs.


linkerd/app/outbound/src/http/logical/router.rs
```diff
@@ -85,23 +87,108 @@ where
...
+            async move {
+                while profiles.changed().await.is_ok() {
+                    let profile = profiles.borrow_and_update();
+                    // Update `backends`.
+                    // TODO(eliza): can we avoid doing this if the backends haven't
+                    // changed at all (e.g., handle updates that only change the
+                    // routes?)
+                    tracing::debug!(backends = profile.targets.len(), "Updating backends");
+                    backends.clear();
+                    Self::mk_backends(&mut backends, &new_backend, &profile.targets, &target);
```
olix0r, line 116: We need to preserve backends that are unchanged, otherwise we'll end up dropping all balancer state on updates.

hawkw: ahh, yup, you're right. will do!

linkerd/app/outbound/src/tcp/logical/router.rs
```diff
@@ -133,6 +180,15 @@ impl<Req, S: Service<Req> + Clone> Service<Req> for Router<S> {
...
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), S::Error>> {
+        // does the current distribution need to be updated?
+        if self
+            .rx
+            .has_changed()
+            .expect("router background task terminated unexpectedly!")
+        {
+            self.current = self.rx.borrow().clone();
+        }
+
```
olix0r, line 191: This is definitely better; though, I think it means we should remove the update in `call`?

linkerd/app/outbound/src/tcp/logical/router.rs
```diff
@@ -60,23 +61,91 @@ where
...
-        for t in profile.targets.iter() {
+        Self::mk_backends(&mut backends, &self.0, &profile.targets, &target);
+
+        // Create a stack that can distribute requests to the backends.
+        let distribute = Self::mk_distribute(&backends, &profile.targets, &target);
+        let (tx, rx) = watch::channel(distribute.clone());
+        drop(profile);
+
+        let new_backend = self.0.clone();
+        let span = tracing::debug_span!("router", target = %Param::<profiles::LogicalAddr>::param(&target));
```
olix0r, line 72: it's fine to add a router span, but the target should already be on the "logical" parent span

linkerd/app/outbound/src/http/logical/router.rs
```diff
@@ -85,23 +87,108 @@ where
...
+        let (tx, rx) = watch::channel(Shared {
+            matches: profile.http_routes.clone(),
+            routes,
+        });
+
+        drop(profile);
+
+        let new_backend = self.new_backend.clone();
+        let route_layer = self.route_layer.clone();
+        let span = tracing::debug_span!("router", target = %Param::<profiles::LogicalAddr>::param(&target));
```
olix0r, line 105: target is already on parent span

PR #2091, opened 2022-12-23
rewrite `FailFast` to drain buffers without becoming ready

hawkw:
The `FailFast` middleware limits the period of time an inner `Service` may return `Poll::Pending` from its `poll_ready` method without becoming ready. If inner service does not become ready by the end of the failfast timeout, the `FailFast` middleware becomes ready and immediately accepts but fails all requests until the inner service once again becomes ready. This allows `Buffer` around an inner service to be drained when the buffered service has become unavailable.

However, an unfortunate consequence of the current `FailFast` design is that it advertises readiness up the stack _past_ a `Buffer` that wraps it. Once the service has entered the failfast state, it will return `Poll::Ready(Ok(()))` for all `poll_ready` calls until the inner service exits failfast. Since a `Buffer` will accept new requests as long as it has queue capacity, without polling the inner service (it is instead polled by the buffer's background worker task), the buffer will continue accepting new requests and immediately failing them as long as the service inside it is in a failfast state. This is not desirable, as it means that a higher level traffic distributor or load balancer will see a `Buffer<FailFast<S>>` as being ready to recieve new requests even when the `S` service is not ready and the `FailFast` middleware will simply fail those requests. It would be preferable for the `Buffer<FailFast<S>>` to return `Poll::Pending` from `poll_ready` when in the failfast state, so that *new* requests are not dispatched to it, but still proactively drain requests already dispatched into the buffer.

This branch changes the `FailFast` implementation to consist a _pair_ of middleware, `FailFast` and `Advertise`. The `FailFast` middleware behaves similarly to how it did previously, but is modified so that the inner service's readiness state is shared with the `Advertise` middleware. The `Advertise` service will advertise the inner service's _actual_ readiness in `poll_ready` (e.g. return `Poll::Pending` while in failfast), while the `FailFast` service will return `Poll::Ready(Ok(()))` and failfast all requests it recieves.

This way, a `Buffer` (or other middleware) can be placed in between an inner `FailFast` service and an outer `Advertise` service. The `Buffer`'s worker task will see the `Poll::Ready(Ok(()))` returned by the `FailFast` middleware, and drain its queue, while an outer traffic distributor or load balancer will see the `Poll::Pending` returned by `Advertise` (reflecting the actual readiness of the inner service), and dispatch traffic elsewhere.

In order to implement this, it was necessary to change the `FailFast` middleware to drive the innermost service to readiness on a background task when it is in the `FailFast` state. This is because a `Buffer`'s worker only polls the inner service's readiness when it has requests to dispatch, rather than in the `Buffer` service's `poll_ready`. Therefore, it's necessary to ensure we proactively drive the readiness of the inner service while in the failfast state, as the `FailFast` service itself will not be polled again once the buffer's queue has drained. Currently, `Buffer`s are always paired with a `tower::spawn_ready::SpawnReady` middleware, which does something analogous, but unconditionally on _all_ `Pending` calls to `poll_ready`. The new implementation has the side benefit of obviating the need for a separate `SpawnReady` middleware, and is a bit more efficient about spawning tasks, since a new task is only spawned when *in the failfast state*, rather than any time the inner service returns `Poll::Pending`.

Depends on #2078


linkerd/stack/src/failfast.rs
```diff
@@ -43,112 +105,192 @@ pub enum ResponseFuture<F> {
...
+                            let _ = inner.ready().await;
+                            info!("{scope} service has recovered");
+                            shared.exit_failfast();
+                            inner
+                        }));
+                    }
+                },
+                State::FailFast(mut task) => {
+                    if let Poll::Ready(res) = task.poll_unpin(cx) {
+                        // the service became ready in the background, exit failfast.
```
olix0r, line 273: Nit/consistency

```suggestion
                        // The service became ready in the background, exit failfast.
```

linkerd/stack/src/failfast.rs
```diff
@@ -168,19 +310,39 @@ where
     }
 }
 
+// === impl Shared ===
+impl Shared {
+    fn exit_failfast(&self) {
+        // the load part of this operation can be `Relaxed`, because this task
```
olix0r, line 316: nit/consistency

```suggestion
        // The load part of this operation can be `Relaxed` because this task
```

linkerd/stack/src/failfast.rs
```diff
@@ -31,10 +85,18 @@ pub struct FailFastError {
...
-enum State {
-    Open,
-    Waiting,
-    FailFast,
+enum State<S> {
+    Open(S),
+    Waiting(S),
+    FailFast(task::JoinHandle<S>),
+    /// Empty state used only for transitions.
+    Empty,
```
olix0r, line 93: wdyt about calling this `Invalid`?

hawkw: sure, fine with me!

linkerd/stack/src/failfast.rs
```diff
@@ -43,112 +105,192 @@ pub enum ResponseFuture<F> {
...
+                trace!("Advertise: {scope} service became ready");
+                self.is_waiting = false;
+            } else if self.shared.in_failfast.load(Ordering::Acquire) {
+                // We are in failfast. start waiting for the inner service to
+                // exit failfast.
+                trace!("Advertise: {scope} service in failfast, waiting for readiness",);
+                let shared = self.shared.clone();
+                self.waiting.set(async move {
+                    shared.notify.notified().await;
+                    trace!("Advertise: {scope} service has become ready");
```
olix0r, line 155: We should avoid this scope stuff, imo. We only have it here so that the failfast error messages include some information about which failfast we hit. Better relevant scope information is already going to be in the trace context.

I'd really like to get rid of the `scope` thing entirely`. I think we have two options: (1) At the top of each stack we put a layer that wraps errors with stack context; or (2) we make failfast generic over the type of error it returns. I'd lean to the former...

Anyway, for now, let's omit scope from the logs.

linkerd/stack/src/failfast.rs
```diff
@@ -43,112 +105,192 @@ pub enum ResponseFuture<F> {
...
     type Future = ResponseFuture<S::Future>;
 
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        match self.inner.poll_ready(cx) {
-            // If the inner service is not ready, go into failfast after `max_unavailable`.
-            Poll::Pending => loop {
-                self.state = match self.state {
-                    // The inner service just transitioned to NotReady, so initiate a new timeout.
-                    State::Open => {
+        let scope = self.scope;
```
olix0r, line 223: etc

linkerd/stack/src/failfast.rs
hawkw: okay, i can remove the field entirely here if you like?

linkerd/stack/src/failfast.rs
```diff
@@ -43,112 +105,192 @@ pub enum ResponseFuture<F> {
...
-                    }
 
-                    // Admit requests and fail them immediately.
-                    State::FailFast => {
+                        warn!("{scope} entering failfast after {:?}", self.max_unavailable);
+                        self.shared.in_failfast.store(true, Ordering::Release);
+
+                        let shared = self.shared.clone();
+                        self.state = State::FailFast(task::spawn(async move {
+                            let _ = inner.ready().await;
```
olix0r, line 264: Hmmm. Don't we need to propagate errors here?

linkerd/stack/src/failfast.rs
olix0r: Yeah I think that's probably good. maybe leave a TODO in the layer about error annotations?

linkerd/stack/src/failfast.rs
hawkw: all this task is doing is notifying the `FailFast` service when the inner service becomes ready. when `FailFast` is notified, it will poll `inner` again. my assumption was that if poll_ready errors once, subsequent calls will also error.

we could change the task to propagate the error and not poll inner immediately after the `FailFast` is woken, but we would still have to return the inner service so that subsequent poll_readys can also get their own errors.

hawkw: okay, it returns the original error as of 990bf54219b31bed970e01addd02029ab7589acb

hawkw: 990bf54219b31bed970e01addd02029ab7589acb changes this to eagerly propagate the original error, just to be safe.

hawkw:
Okay, the failure in the `telemetry::tcp_errors::inbound_direct_success` test actually does seem to be real, trying to figure that out.

hawkw:
> Okay, the failure in the `telemetry::tcp_errors::inbound_direct_success` test actually does seem to be real, trying to figure that out.

Ah, it looks like that test relies on a service which is in failfast because it can't be resolved from the control plane actually failing a request, rather than being `Pending` and waiting for it to become un-pending. I guess we still need to figure out how we want to actually handle the case where no balancers are available for a request...

hawkw:
Turns out that the metrics test is the only one where we exercise what the proxy does when the `destination` service returns a response with `grpc-status: unavailable` (which is to retry forever, apparently); the `outbound_times_out` tests only exercise what happens when the controller accepts the connection and just doesn't reply to the request...

linkerd/stack/src/failfast.rs
```diff
@@ -2,153 +2,300 @@
...
+
+impl<S> FailFast<S> {
+    /// Returns a layer for producing a `FailFast` pair wrapping an inner
+    /// layer's service.
+    ///
+    /// When the service is in fail-fast, the inner layer's service will see the
+    /// fail-fast errors produced by _its_ inner service. However, the outer
+    /// failfast service, which wraps the inner layer's service, will return
+    /// `Poll::Pending` while the innermost service is unavailable.
+    pub fn wrap_layer<L>(
```
olix0r, line 174:
```suggestion
    pub fn layer_gated<L>(
```

linkerd/stack/src/failfast.rs
```diff
@@ -2,153 +2,300 @@
...
-                    }
 
-                    // Admit requests and fail them immediately.
-                    State::FailFast => {
-                        debug!("{} in failfast", self.scope);
+                        warn!("Service entering failfast after {:?}", self.max_unavailable);
+                        self.shared.in_failfast.store(true, Ordering::Release);
+
+                        let shared = self.shared.clone();
+                        self.state = State::FailFast(task::spawn(async move {
```
olix0r, line 254: We need to abort this task on drop like `SpawnReady` does.

linkerd/stack/src/failfast.rs
```diff
@@ -2,153 +2,300 @@
...
+                            let res = inner.ready().await;
+                            shared.exit_failfast();
```
olix0r, line 256:
```suggestion
                            // Notify the paired `Gate` instances to begin
                            // advertising readiness so that the failfast
                            // service can advance. 
                            shared.exit_failfast();
```

olix0r, line 256:
```suggestion
                            // Notify the paired `Gate` instances to begin
                            // advertising readiness so that the failfast
                            // service can advance. 
                            shared.exit_failfast();
```

olix0r, line 256:
```suggestion
                            // Notify the paired `Gate` instances to begin
                            // advertising readiness so that the failfast
                            // service can advance. 
                            shared.exit_failfast();
```

olix0r, line 256:
```suggestion
                            // Notify the paired `Gate` instances to begin
                            // advertising readiness so that the failfast
                            // service can advance. 
                            shared.exit_failfast();
```

linkerd/stack/src/failfast.rs
hawkw: done in [6d3eb56](https://github.com/linkerd/linkerd2-proxy/pull/2091/commits/6d3eb56d84828cd975a7f13e2e0ec6429a1cda5b)

linkerd/stack/src/failfast.rs
```diff
@@ -163,24 +340,45 @@ where
...
+
+impl Shared {
+    fn exit_failfast(&self) {
+        // The load part of this operation can be `Relaxed` because this task
+        // is the only place where the the value is ever set.
+        if self
+            .in_failfast
+            .compare_exchange(true, false, Ordering::Release, Ordering::Relaxed)
+            .is_ok()
+        {
```
olix0r, line 358: This seems simpler and equivalent, right?

```suggestion
        if self.in_failfast.swap(false, Ordering::Release) {
            self.notify.notify_waiters();
        }
```

hawkw: ah, yeah, it is, good suggestion!

linkerd/stack/src/loadshed.rs
```diff
@@ -0,0 +1,64 @@
...
+use linkerd_error::Error;
+use std::task::{Context, Poll};
+use thiserror::Error;
+use tower::Service;
+
+/// A middleware that sheds load when the inner `Service` isn't ready.
+#[derive(Clone, Debug)]
+pub struct LoadShed<S> {
+    inner: S,
+    open: bool,
```
olix0r, line 12: How does `open` work wrt clone? Should we set open on clone?

linkerd/stack/src/loadshed.rs
```diff
@@ -0,0 +1,85 @@
+use super::layer;
+use futures::{future, TryFutureExt};
+use linkerd_error::Error;
+use std::task::{Context, Poll};
+use thiserror::Error;
+use tower::Service;
+
+/// A middleware that sheds load when the inner `Service` isn't ready.
+#[derive(Debug)]
+pub struct LoadShed<S> {
```
olix0r, line 10: We should add some basic tests for this module

hawkw: done [cf8e6d8](https://github.com/linkerd/linkerd2-proxy/pull/2091/commits/cf8e6d85d58ab3748c3fe88b27716dfd09088f79)!

olix0r (approved):
excelente!

PR #2095, opened 2022-12-28
Update routers to support per-request backend distributions

olix0r:
The outbound logical stacks use a single `profiles::split` for all
requests. Therefore, request routing can't influence a request's backend
selection.

This change replaces the service-profile-specific router and split
implementations in favor of the newer, more general router (https://github.com/linkerd/linkerd2-proxy/commit/577ad5285a55c344beba066b040c11bd1d828f49) and
distribute (https://github.com/linkerd/linkerd2-proxy/commit/47678e6cfb35aaac8ab31d3bab53944542929155) modules. Unlike the service-profile-router, the
generic router is lazy and only creates route services as a matching
request is found.

Buffers are moved into the concrete stack so that balancers may be
shared by more than one route.

This change also updates trace contexts so that we include
logical/endpoint metadata as early as possible.


linkerd/app/outbound/src/tcp/logical.rs
```diff
@@ -1,3 +1,8 @@
+mod router;
+#[cfg(test)]
+mod tests;
```
olix0r, line 3: https://github.com/linkerd/linkerd2-proxy/pull/2096

linkerd/stack/src/lazy.rs
```diff
@@ -0,0 +1,107 @@
...
+                Shared::Service(svc) => {
+                    self.inner = Some(svc.clone());
+                    Shared::Service(svc)
+                }
+
+                Shared::Invalid => unreachable!(),
+            };
+
+            debug_assert!(self.inner.is_some());
+        }
```
hawkw, line 84: nit, take it or leave it: seems like we could get rid of the `loop`/`debug_assert!` and just write
```suggestion
        if let Some(inner) = self.inner.as_mut() {
            return inner.poll_ready(cx);
        }

        let mut shared = self.shared.lock();
        *shared = match std::mem::take(&mut *shared) {
            Shared::Uninit { inner, target } => {
                let svc = inner.new_service(target);
                self.inner = Some(svc.clone());
                Shared::Service(svc)
            }

            Shared::Service(svc) => {
                self.inner = Some(svc.clone());
                Shared::Service(svc)
            }

            Shared::Invalid => unreachable!(),
        };

        self.inner.as_mut().unwrap().poll_ready(cx)
```
since, AFAICT, the only thing the loop will ever iterate is the `if let Some(inner)` part, and we could just expect that `self.inner` is `Some` if we've just done the state transition?

linkerd/app/outbound/src/http/logical/router.rs
```diff
@@ -0,0 +1,290 @@
...
+where
+    S: Service<http::Request<B>> + Clone,
+{
+    type Response = S::Response;
+    type Error = S::Error;
+    type Future = Oneshot<S, http::Request<B>>;
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), S::Error>> {
+        // TODO(ver) figure out how backpressure should work here. Ideally, we
+        // should only advertise readiness when at least one backend is ready.
```
hawkw, line 129: we could maybe wrap the backends in a `SpawnReady` type thing that also has a shared count of how many backends are ready?

alternatively, though, would it make sense for the router itself to shed load if none of its backends are ready?

linkerd/app/outbound/src/http/logical.rs
```diff
@@ -14,101 +13,87 @@ struct ProfileRoute {
...
-                .push(profiles::split::layer())
-                .push_on_service(
+            // If there's no route, use the logical service directly; otherwise
+            // use the per-route stack.
+            concrete
+                .push(NewRouter::<_, _, Concrete>::layer(
                     svc::layers()
+                        // Maintain a per-route distributor over concrete
+                        // backends from the (above) concrete cache.
+                        .push(http::insert::NewInsert::<ProfileRoute, _>::layer())
```
hawkw, line 50: it's not _immediately_ obvious how the `NewInsert` layer is related to maintaining a per-route distributor, can we update the comment to explain that?

linkerd/app/outbound/src/http/logical/router.rs
olix0r: > would it make sense for the router itself to shed load if none of its backends are ready?

In the current setup, yes; generally, no. We could imagine a scenario where there are multiple levels of routing (e.g. applying routes on both logical and concrete services) such that we may want to be able to failover between routers. Or we could have a reason to put a buffer above the router to absorb some amount of backpressure...

But that's not a hard requirement. We could shed load here if it makes sense. But it also seems like it's not really the router's concern.

olix0r: > we could maybe wrap the backends in a SpawnReady type thing that also has a shared count of how many backends are ready?

Yeah, we can probably wrap each backend with a service that spawnready's and publishes the inner service's state (i.e. to an AtomicBool + Notify) so we can peak at recent backend states without having to do any polling here except to be notified of updates.

The caveat is that we'll only really know for backends that have been used and we'll only be able to detect unavailable servers as they are used by the router... but that's probably the right compromise.

None of this seems like a blocker.

olix0r: That sounds a lot like what `failfast::Gate` is already doing; but it would probably be a layering headache to reuse that directly.

hawkw:
somewhere between f7a48af4a65a07867adc0031e8f3e2cabf37dd9b and now, we stopped load-shedding when a resolution takes forever (again); this broke the `telemetry::tcp_errors::inbound_direct_success` test (again).

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -59,51 +59,44 @@ impl<N> Outbound<N> {
...
                         .push(
                             rt.metrics
                                 .proxy
                                 .stack
-                                .layer(stack_labels("http", "balancer")),
+                                .layer(stack_labels("http", "concrete")),
                         )
-                        .push(http::BoxResponse::layer()),
+                        .push_buffer::<http::Request<http::BoxBody>>("HTTP", http_request_buffer)
+                        .push(svc::LoadShed::layer()),
```
olix0r, line 97:
```suggestion
```

right?

linkerd/app/outbound/src/http/logical.rs
```diff
@@ -1,143 +1,264 @@
...
+                .push(svc::NewSpawnWatch::<Profile, _>::layer_into::<Params>());
+
+            // Caches each logical stack so that it can be reused across
+            // per-connection HTTP server stacks (i.e. created by the
+            // `DetectService`).
+            //
+            // TODO(ver) Update the detection stack so this dynamic caching can
+            // be removed.
+            //
+            // XXX(ver) This cache key includes the HTTP version. Should it?
```
olix0r, line 124: y

linkerd/app/outbound/src/tcp/concrete.rs
```diff
@@ -60,24 +68,35 @@ impl<C> Outbound<C> {
             connect
                 .push(svc::stack::WithoutConnectionMetadata::layer())
                 .push_make_thunk()
-                .instrument(|t: &Endpoint| debug_span!("endpoint", addr = %t.addr))
+                .push_on_service(
+                    rt.metrics
+                        .proxy
+                        .stack
+                        .layer(stack_labels("http", "endpoint")),
```
olix0r, line 75:
```suggestion
                        .layer(stack_labels("tcp", "endpoint")),
```

linkerd/app/outbound/src/http/logical.rs
```diff
@@ -1,143 +1,270 @@
...
+                .push(BackendCache::layer())
+                // Lazily cache a service for each `RouteParams`
+                // returned from the `SelectRoute` impl.
+                .push_on_service(route)
+                .push(svc::NewOneshotRoute::<Params, _, _>::layer_cached());
+
+            // For each `Logical` target, watch its `Profile`, rebuilding a
+            // router stack.
+            concrete
+                // Share the concrete stack each router stack.
```
olix0r, line 125:
```suggestion
                // Share the concrete stack with each router stack.
```

linkerd/app/outbound/src/tcp/logical.rs
```diff
@@ -17,40 +50,116 @@ impl<N> Outbound<N> {
...
-                        // target.
-                        .push_buffer("TCP Logical", tcp_connection_buffer),
-                )
-                // TODO(ver) Can we replace this evicting cache? The detect
-                // stack would have to hold/reuse inner stacks.
-                .push_idle_cache(*discovery_idle_timeout)
-                .check_new_service::<Logical, I>()
-                .instrument(|_: &Logical| debug_span!("tcp"))
-                .check_new_service::<Logical, I>()
+                // Share the concrete stack each router stack.
```
olix0r, line 79:
```suggestion
                // Share the concrete stack with each router stack.
```

linkerd/app/outbound/src/http/detect.rs
```diff
@@ -31,6 +43,8 @@ impl<N> Outbound<N> {
         self.map_stack(|config, rt, tcp| {
             let ServerConfig { h2_settings, .. } = config.proxy.server;
 
+            let tcp = tcp.instrument(|_: &_| debug_span!("opaque"));
```
hawkw, line 46: nit: looks like the `opaque` span is being added to the TCP stack that's used if detection is disabled, _or_ if HTTP is not detected...i feel like it might be clearer if there were separate "opaque" and "tcp" spans to differentiate between "no HTTP detected" and "HTTP detection skipped" cases?

linkerd/app/outbound/src/http/logical.rs
```diff
@@ -1,143 +1,270 @@
...
+
 impl<N> Outbound<N> {
-    pub fn push_http_logical<NSvc>(self) -> Outbound<svc::ArcNewHttp<Logical>>
+    /// Builds a `NewService` that produces a router service for each logical
+    /// target.
+    ///
+    /// The router uses discovery information (provided on the target) to
+    /// support per-request routing over a set of concrete inner services.
+    /// Only available inner services are used for routing. When there are no
+    /// available backends, requests are failed with a [`svc::stack::LoadShedError`].
```
hawkw, line 43: nit, not a big deal: if we're gonna document the error types here, should we maybe also state that the `NoRoute` error is returned if there's no route that matches a request?

linkerd/app/outbound/src/http/logical.rs
```diff
@@ -1,143 +1,270 @@
...
+                .push_new_clone()
+                // Rebuild this router stack every time the profile changes.
+                .push_on_service(router)
+                .push(svc::NewSpawnWatch::<Profile, _>::layer_into::<Params>())
+                // Add l5d-dst-canonical header to requests.
+                //
+                // TODO(ver) move this into the endpoint stack so that we can only
+                // set this on meshed connections.
+                //
+                // TODO(ver) do we need to strip headers here?
```
hawkw, line 135: do we need to figure this out before merging? the previous comment implied that we did this but i'm not sure if that was actually true...

linkerd/app/outbound/src/http/server.rs
```diff
@@ -12,6 +12,13 @@ pub(crate) struct ServerRescue {
 }
 
 impl<N> Outbound<N> {
+    ///  Builds a [`svc::NewService`] stack that prepares HTTP requests to be
+    ///  proxied.
```
hawkw, line 16: tiny style nit:
```suggestion
    /// Builds a [`svc::NewService`] stack that prepares HTTP requests to be
    /// proxied.
```

linkerd/app/outbound/src/tcp/logical.rs
```diff
@@ -1,12 +1,45 @@
...
 
 impl<N> Outbound<N> {
+    /// Builds a `NewService` that produces a router service for each logical
+    /// target.
+    ///
+    /// The router uses discovery information (provided on the target) to
+    /// support per-request connection routing over a set of concrete inner
+    /// services. Only available inner services are used for routing. When
+    /// there are no available backends, requests are failed with a
+    /// [`svc::stack::LoadShedError`].
```
hawkw, line 40: nit, take it or leave it: if we're documenting the errors this returns, should we mention that `NoRoute` errors are returned when no route matches?

linkerd/app/outbound/src/tcp/logical.rs
```diff
@@ -17,40 +50,116 @@ impl<N> Outbound<N> {
...
+                        logical: logical.clone(),
+                        resolve: ConcreteAddr(addr),
+                    };
+                    (concrete, weight)
+                },
+            ))
+            .expect("distribution must be valid");
+
+            (backends, distribution)
+        };
```
hawkw, line 123: it occurs to me that the code for generating the distribution is identical across the TCP and HTTP routeparams and could probably be shared. not a big deal though.

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -9,20 +9,30 @@ use linkerd_app_core::{
...
-                http::Request<http::BoxBody>,
-                Response = http::Response<http::BoxBody>,
-                Error = Error,
-                Future = impl Send,
-            >,
+                    http::Request<http::BoxBody>,
+                    Response = http::Response<http::BoxBody>,
+                    Error = Error,
+                    Future = impl Send,
+                > + Clone,
```
hawkw, line 35: tiny nit: is this indentation rustfmt-approved? it _seems_ off to me

hawkw (approved):
i'm really excited to move forward with this, and it looks great overall! i left some last-minute style nits and a couple questions about TODOs.

linkerd/app/outbound/src/http/concrete.rs
olix0r: Yeah, that's all rustfmt's doing.

linkerd/app/outbound/src/tcp/logical.rs
olix0r: yeah, agreed. let's see where the config goes before we try to unify things. I'm not sure we want a single `Concrete` type, for instance...

linkerd/app/outbound/src/http/logical.rs
olix0r: Right. The previous comment is wrong, I think. Needs revisiting before stable but not a blocker for now.

linkerd/app/outbound/src/http/detect.rs
olix0r: i'd rather not vary that at the span level. we have events on the parent that indicate any detection timeouts.

PR #2100, opened 2022-12-29
wip: Watch stack module

olix0r:


linkerd/stack/src/watch.rs
```diff
@@ -0,0 +1,96 @@
+use crate::{NewService, Param, Service};
+use std::task::{Context, Poll};
+use tokio::sync::watch;
+use tracing::Instrument;
+
+pub trait UpdateWatch<T> {
+    type Service: Clone + Default;
```
hawkw, line 7: i put the `Clone` and `Default` bounds here because we'll get a much more obvious type error if they're missing

hawkw: we probably should also put the `Send` + `Sync` bounds here since this trait is only used by `NewSpawnWatch` anyway...

linkerd/app/outbound/src/http/logical.rs
```diff
@@ -40,45 +40,45 @@ impl<N> Outbound<N> {
...
-                        )
-                        // Sets the per-route response classifier as a request
-                        // extension.
-                        .push(classify::NewClassify::layer())
-                        .push_on_service(http::BoxResponse::layer())
-                        .push_map_target(|(route, logical)| ProfileRoute { logical, route })
-                        .push(svc::NewLazy::layer()),
+            let route = svc::layers()
+                // Maintain a per-route distributor over concrete
+                // backends from the (above) concrete cache.
```
hawkw, line 45: nit: is the cache still above this (in the file)? does the comment need to be updated?

linkerd/distribute/src/service.rs
```diff
@@ -150,6 +150,17 @@ impl<K: Clone, S: Clone> Clone for Distribute<K, S> {
     }
 }
 
+impl<K, S> Default for Distribute<K, S> {
+    /// Returns an empty distribution. This distribution will never become ready/
```
hawkw, line 154: whoops, typo on my part:
```suggestion
    /// Returns an empty distribution. This distribution will never become ready.
```

linkerd/stack/src/watch.rs
```diff
@@ -0,0 +1,95 @@
...
+use tracing::Instrument;
+
+pub trait UpdateWatch<T>: Send + Sync + 'static {
+    type Service: Clone + Default + Send + Sync + 'static;
+
+    fn update(&mut self, target: &T) -> Option<Self::Service>;
+}
+
+#[derive(Clone, Debug)]
+pub struct NewSpawnWatch<P, N> {
```
hawkw, line 13: probably worth a comment somewhere in the file explaining what this does?

linkerd/stack/src/watch.rs
```diff
@@ -0,0 +1,95 @@
...
+use tokio::sync::watch;
+use tracing::Instrument;
+
+pub trait UpdateWatch<T>: Send + Sync + 'static {
+    type Service: Clone + Default + Send + Sync + 'static;
+
+    fn update(&mut self, target: &T) -> Option<Self::Service>;
+}
+
+#[derive(Clone, Debug)]
```
hawkw, line 12: this will only implement `Clone`/`Debug` if `P` is `Clone`/`Debug`, because of the `PhantomData`...should we have a manual impl?

linkerd/stack/src/watch.rs
```diff
@@ -0,0 +1,95 @@
...
+impl<Req, S> Service<Req> for SpawnWatch<S>
+where
+    S: Service<Req> + Clone,
+{
+    type Response = S::Response;
+    type Error = S::Error;
+    type Future = S::Future;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        if matches!(self.rx.has_changed(), Ok(true)) {
```
hawkw, line 85: should `poll_ready` error if the background task is dead? AFAICT it could disappear on us abruptly if the underlying target watch closes, and wouldn't be recoverable (we would need to start a new resolution)...

hawkw (approved):
it's neat that the watch logic can be factored out, this looks good to me! i left a few minor nits/questions, but this looks good overall!

linkerd/stack/src/watch.rs
olix0r: I don't think there's any need to error--we have a handle to the service we can continue using.

linkerd/stack/src/watch.rs
olix0r: yeah probably

PR #2101, opened 2022-12-29
stack: add `SpawnWatch` middleware

hawkw:
This branch adds an implementation of a middleware that watches a `watch::Receiver` of targets and builds a new service when the watch is updated. The current state of the inner service is broadcast over a second watch to any clones of the `SpawnWatch` service.

This is the same code added in PR #2100, but rebased onto `main`, with documentation and tests added.

### DIFF
--- a/linkerd/stack/src/watch.rs
+++ b/linkerd/stack/src/watch.rs
@@ -11,8 +11,14 @@ pub trait UpdateWatch<T>: Send + 'static {
 
     /// Update the inner service with a new `target` value.
     ///
-    /// If this method returns `None`, the `Default` implementation for
-    /// [`Self::Service`] is used instead.
+    /// If the inner service has not changed as a result of the update, this
+    /// method may return `None`. If `None` is returned, no update will be
+    /// published to the [`SpawnWatch`] service.
+    ///
+    /// This method is also called initially when the first [`SpawnWatch`]
+    /// service is constructed. In this case, it should generally not return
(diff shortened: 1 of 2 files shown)


linkerd/stack/src/watch.rs
```diff
@@ -0,0 +1,225 @@
+use crate::{NewService, Param, Service};
+use std::task::{Context, Poll};
+use tokio::sync::watch;
+use tracing::Instrument;
+
+/// Defines how to update a [`Service`] in the background as a [`watch::Reciever`]
```
olix0r, line 6:
```suggestion
/// Defines how to update a [`Service`] in the background as a [`watch::Receiver`]
```

linkerd/stack/src/watch.rs
```diff
@@ -0,0 +1,225 @@
...
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        if dbg!(matches!(self.rx.has_changed(), Ok(true))) {
+            self.inner = self.rx.borrow_and_update().clone();
+        }
+
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Req) -> Self::Future {
```
olix0r, line 112:
```suggestion
    #[inline]
    fn call(&mut self, req: Req) -> Self::Future {
```

olix0r (approved):

PR #2108, opened 2022-12-30
unify HTTP and TCP logical routers

hawkw:
Currently, as of #2095, separate implementations of the outbound logical
router exist for HTTP and raw TCP stacks. This is kind of unfortunate,
as these contain a lot of duplicated code.

This branch builds on top of PR #2107 unifies these two implementations
into a single logical router type that's completely generic over how
routes are represented and matched. In the `linkerd-service-profiles`
crate, we introduce a new `tcp` module with TCP `tcp::RequestMatch` and
`tcp::Route` types. The intention is that when Linkerd eventually grows
support for the Gateway API TCPRoute CRD, a `Profile` (which I suspect
subsequent branches will rename to `Policy`) will contain a list of TCP
matchers and routes, but currently, every `Profile` has only a single
TCP route. The `tcp::Route` type currently only contains the list of
targets for traffic splitting.

This change also continues setting up the `Profile` type from
`linkerd-service-profiles` to become a generic client policy
representation that can be produced from either a watch on the
ServiceProfiles proxy API watch *or* a watch on the new client policy
Outbound proxy API. This was started in #2107 by moving the list of
backends from the top-level profile onto the individual routes in that
profile. In this PR, I've added default HTTP and TCP routes which are
generated when the profile comes from a ServiceProfiles API resolution,
so that the router will no longer fail when a request doesn't match any
of the routes in the ServiceProfile. This will allow future client
policy work to implement the Gateway API's required behavior of
returning a 404 error when no HTTPRoute matches a request, without
changing the behavior of existing ServiceProfiles.

Finally, I found a small bug in the logical router implementation that
resulted in most tests failing with a "distribution references unknown
backend" error. The `UpdateWatch::update` implementation returns early
if the list of backends stored in the `Update` type is the same as the
list of backend names in the profile...but this is incorrect when the
list of backends is empty, because we must still generate a default
backend for the route's logical address. Commit
ccfcc364a43bca52dbed784b48ecf22542f36975 changes the `UpdateWatch`
implementation to only return early here when the set of targets is not
empty.

Depends on #2107


olix0r (approved):

linkerd/app/outbound/src/logical/router.rs
```diff
@@ -207,24 +207,22 @@ where
...
-        if target_addrs
-            .iter()
-            .all(|addr| self.backends.contains_key(addr))
-            // Don't return early if we still need to populate the default backend!
-            && !target_addrs.is_empty()
-        {
+        // We just removed all backends that aren't in the new addrs, so we
+        // we can skip further processing by comparing their lengths.
+        debug_assert!(addrs.len() >= self.backends.len());
+        if addrs.len() == self.backends.len() {
```
hawkw, line 220: this isn't quite right, i think; the tests i fixed in ccfcc364a43bca52dbed784b48ecf22542f36975 are failing again; i think because the initial route set needs to be generated and we'll indicate that it doesn't?

hawkw: oh nevermind, the test failure is in a stack test and it's only failing because the stack test doesn't make a profile that matches what we would generate from proto anymore

PR #2110, opened 2022-12-31
refactor to prep for client policy integration

hawkw:
This branch sets up the changes in PR #2095 for integration with the
client-policy API. The intended design is to have a single client policy
watch type which, depending on the target service, may be updated by
ServiceProfile or outbound Service discovery watches. This branch does
not actually integrate with the client policy API, but makes setup
changes to prepare for that integration.

This branch makes the following changes:

- use client policy terminology (rename `targets` to `backends`)
- add a new `linkerd-client-policy` crate
- moves the `Profile`, `Receiver`, `ReceiverStream`, and `LookupAddr`
  types from `linkerd-service-profiles` to `linkerd-client-policy`, and
  renames `Profile` to `Policy`
- redesigns the `MatchRoute` trait used in the logical router from a
  trait that is implemented for a single route matcher to one that is
  implemented for an entire route set
- makes the `Policy` (nee `Profile`) type generic over a set of HTTP
  routes and TCP routes implementing the new `MatchRoute`

Currently, all policy watches are hardcoded to use only the
ServiceProfiles route matching types. However, in a subsequent change,
separate route-matching types for HTTPRoute will be added, and the
`Policy` types watched by the router will store an enum of either the
HTTPRoute route-matching type or the ServiceProfile route matching type,
depending on what kind of policy was discovered for that target. This
allows us to vary the behavior for how routes are selected
(ServiceProfile-style first matching route selection vs HTTPRoute-style
most specific route selection) based on where the routes came from,
without the logical router having to be aware of the specific route
kinds.

This depends on #2095, but if that PR is going to merge soon, this PR
should probably wait until #2095 merges and merge separately along with
other client-policy work.


olix0r:
As this basically adds all of the policy types, I'd prefer to hold off on this until we can actually move the router changes through. I have some specific thoughts on how to do that, and it probably has implications for this work...

hawkw:
mmk, i'll close this branch for now, then.

PR #2115, opened 2023-01-03
wip: refactor router

olix0r:
* Extract generic linkerd-router crate that deals with managing stacks and dispatching requests over a backend.
* Update `SpawnWatch` to use `NewService` instead of `Update` -- no need for a new trait.
* Add `CacheNewDistribute` that handles reusing backends for a `NewDistribute` (removing need for `Update`).
* The outbound crate should deal with all of 'domain' details -- target types, route types, etc.
* Using tuple-targets like `(Route, T)` adds more complexity than we want. It's far simpler to duplicate information (i.e. include `T` in all route targets).
* The `profile` crate should only deal with describing validated profile updates. The `policy` crate will deal only with the policy API types. E.g., we shouldn't have any of the tcp route info on profiles.
* The `outbound` crate provides its own target types that are used to configure the router(s). These types are derived from the profile/policy watches.
* We probably want to stop providing a `Receiver<Profile>` to the outbound stack and instead provide a watch of an outbound-specific type. The discovery stack would handle (1) discovery from all data sources we need to build that target (ideally in parallel) and (2) it spawns a task that processes updates from the discovery sources, reconciles those differences, and provides the types we need to proceed.

There are still a bunch of problems with this branch. The SpawnWatch/Cache changes may be worth trying to merge back to concretemas in isolation...


linkerd/addr/src/lib.rs
```diff
@@ -16,7 +17,7 @@ pub enum Addr {
 
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
 pub struct NameAddr {
-    name: Name,
+    name: Arc<Name>,
```
hawkw, line 20: nit, take it or leave it: this change could probably land on `main` in a totally separate branch?

linkerd/distribute/src/cache.rs
```diff
@@ -0,0 +1,73 @@
...
+
+/// A [`NewService`] that produces [`NewDistribute`]s using a shared cache of
+/// backends.
+#[derive(Debug)]
+pub struct CacheNewDistribute<K, N, S>(Arc<Inner<K, N, S>>);
+
+#[derive(Debug)]
+pub struct Inner<K, N, S> {
+    new_backend: N,
+    backends: Mutex<ahash::AHashMap<K, S>>,
```
hawkw, line 14: AFAICT, the intention is to use this as the `NewService` passed to a `SpawnWatch` middleware, in which case, this mutex is never actually contended?

linkerd/distribute/src/cache.rs
```diff
@@ -0,0 +1,73 @@
...
+use parking_lot::Mutex;
+use std::{fmt::Debug, hash::Hash, sync::Arc};
+
+/// A [`NewService`] that produces [`NewDistribute`]s using a shared cache of
+/// backends.
+#[derive(Debug)]
+pub struct CacheNewDistribute<K, N, S>(Arc<Inner<K, N, S>>);
+
+#[derive(Debug)]
+pub struct Inner<K, N, S> {
```
hawkw, line 12: AFAICT this doesn't actually need to be `pub`? and it looks like a private implementation detail...
```suggestion
struct Inner<K, N, S> {
```

linkerd/distribute/src/params.rs
```diff
@@ -53,10 +69,11 @@ impl<K> Distribution<K> {
         if keys.len() < 2 {
             return Ok(Self::first_available(keys));
         }
-        let index = WeightedIndex::new(weights)?;
+        // Error if the distribution is invalid.
+        let _index = WeightedIndex::new(weights.iter().copied())?;
```
hawkw, line 73: ?

linkerd/router/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+use tracing::{debug, error};
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct RouteKeys<K>(Arc<[K]>)
+where
+    K: Eq + Hash + Clone;
+
+pub trait SelectRoute<Req> {
+    type Key: Eq + Hash + Clone + Debug + Send + Sync + 'static;
+    type Error: std::error::Error + Send + Sync + 'static;
```
hawkw, line 21: making this trait responsible for returning the error seems like a good call :+1:

linkerd/router/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+/// Each request is matched against the route table and routed to the
+/// appropriate
+#[derive(Clone, Debug)]
+pub struct Route<T, K, S> {
+    params: T,
+    routes: Arc<AHashMap<K, S>>,
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("unknown route: {0:?}")]
```
hawkw, line 53: if the `SelectRoute` selects a key that's not in `self.routes`, that's almost certainly a bug in the `SelectRoute` impl...should this  error message maybe indicate clearly that this is an internal bug in Linkerd?

linkerd/distribute/src/cache.rs
olix0r: yyyes though i think there's actually somewhat of a bug right now, since this cache is currently not scoped to the logical target

linkerd/router/src/lib.rs
olix0r: I don't think it's necessarily a 'bug' from the crate's point of view. It doesn't necessarily require that `RouteKeys` is comprehensive for all possible route values. Though, in our specific usage we could have special error handling to that effect.

linkerd/router/src/lib.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    fn new_service(&self, target: T) -> Self::Service {
+        let inner = self.inner.new_service(target.clone());
+        let new_route = self.route_layer.layer(inner);
+
+        let RouteKeys(keys) = target.param();
+        let routes = keys
+            .iter()
+            .map(|key| (key.clone(), new_route.new_service(key.clone())))
+            .collect();
```
olix0r, line 88: We could potentially eliminate this and instead lazily cache inner services as they're referenced.

linkerd/stack/src/watch.rs
```diff
@@ -30,10 +11,23 @@ pub trait UpdateWatch<T>: Send + 'static {
...
 }
 
+#[derive(Clone, Debug)]
+pub struct NewWatchInto<P, N> {
+    inner: N,
+    _marker: std::marker::PhantomData<fn() -> P>,
+}
+
+#[derive(Clone, Debug)]
+pub struct NewInto<T, P, N> {
```
hawkw, line 25: naming quibble, take it or leave it: this looks like it specifically handles building a `P` from a `(T, U)` tuple...maybe `NewFromTuple` or `NewFromPair` (or something) would be a clearer name?

hawkw: also, very minor nit, but would it make sense for this to live in `stack` as its own middleware, since it's not really specific to watch? we might not need to use it anywhere else, so it's not a huge deal either way...

linkerd/router/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+    type Key: Eq + Hash + Clone + Debug + Send + Sync + 'static;
+    type Error: std::error::Error + Send + Sync + 'static;
+
+    /// Given a a request, returns the key matching this request.
+    ///
+    /// If no route matches the request, this method returns an error.
+    fn select<'r>(&self, req: &'r Req) -> Result<&Self::Key, Self::Error>;
+}
+
+/// A [`NewService`] that lazy builds stacks for each `K`-typed key.
```
hawkw, line 28: nit: technically it's the `Route` that lazily builds stacks for each `K`-typed key; this just builds `Route`s...

linkerd/distribute/src/cache.rs
```diff
@@ -0,0 +1,73 @@
...
+
+        NewDistribute::from(cache.clone())
+    }
+}
+
+// impl<K, N, S> Clone for CacheNewDistribute<K, N, S> {
+//     fn clone(&self) -> Self {
+//         Self(self.0.clone())
+//     }
+// }
```
hawkw, line 73: is this needed, or should it be deleted?

linkerd/app/outbound/src/tcp/logical.rs
```diff
@@ -122,39 +175,74 @@ impl<C> Outbound<C> {
...
+
+        Self {
+            logical,
+            backends,
+            route,
         }
+    }
+}
+
+// === impl Params ===
```
hawkw, line 219: tiny nit: there's already an `impl Params` comment above the `From` impl
```suggestion
```

linkerd/app/outbound/src/tcp/logical.rs
```diff
@@ -122,39 +175,74 @@ impl<C> Outbound<C> {
     }
 }
 
-// === impl ProfileRoute ===
+// === impl Params ===
+
+impl From<(Profile, Logical)> for Params {
```
hawkw, line 180: nit: _maaaaaybe_ worth a comment saying where this gets impl gets consumed? not a huge deal though.

linkerd/app/outbound/src/http/logical.rs
```diff
@@ -99,53 +133,110 @@ impl<N> Outbound<N> {
     }
 }
 
-// === impl ProfileRoute ===
+// === impl Params ===
 
-impl svc::Param<profiles::http::Route> for ProfileRoute {
-    fn param(&self) -> profiles::http::Route {
-        self.route.clone()
+impl From<(Profile, Logical)> for Params {
```
hawkw, line 138: nit: _maaaaaybe_ worth a comment saying where this gets impl gets consumed? not a huge deal though.

hawkw (approved):
overall, this feels like a good direction!

i mostly left some minor nitpicky comments on docs and naming; the change looks good to me overall.

i'd be interested in investigating whether there's a meaningful advantage from using a concurrent hashmap as the routing table, but i think that makes sense to do later after these changes land.

linkerd/router/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+
+/// The [`Service`] constructed by [`NewRoute`].
+///
+/// Each request is matched against the route table and routed to the
+/// appropriate
+#[derive(Clone, Debug)]
+pub struct Route<T, K, N, S> {
+    params: T,
+    new_route: N,
+    routes: Arc<Mutex<AHashMap<K, S>>>,
```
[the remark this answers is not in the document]
olix0r: I'm not particularly keen on dashmap. Let's keep it simple for now and revisit if it shows up in flamegraphs.  Contention is really a factor of the number of cores given to Linkerd. As long as we expect core allocations to be fairly low, I'd be astonished if this sort of thing shows up at all.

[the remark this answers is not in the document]
hawkw: yeah, that was my thinking as well; i think we should merge this as-is, and maybe do some load testing against the proxy refactor later.

linkerd/stack/src/watch.rs
olix0r: it's already in stack, but for now this feels specific enough to commingle.

linkerd/app/core/src/svc.rs
```diff
@@ -228,13 +230,13 @@ impl<S> Stack<S> {
         self.push_on_service(buffer(name, config))
     }
 
-    pub fn push_cache<T>(self, idle: Duration) -> Stack<cache::NewCachedService<T, S>>
+    pub fn push_cache<T>(self, idle: Duration) -> Stack<idle_cache::NewIdleCached<T, S>>
```
hawkw, line 233: nit, take it or leave it: should this maybe be called `push_idle_cache`? that way it's more obvious which cache implementation is being pushed...

linkerd/idle-cache/src/lib.rs
```diff
@@ -19,10 +19,11 @@ use tracing::{debug, instrument, trace};
 
 mod new_service;
 
-pub use new_service::NewCachedService;
+pub use new_service::NewIdleCached;
 
+/// A `NewService` that
```
hawkw, line 24: a `NewService` that...?

linkerd/stack/src/watch.rs
```diff
@@ -48,31 +43,45 @@ pub struct SpawnWatch<S> {
...
+            inner,
             _marker: std::marker::PhantomData,
         }
     }
+
+    pub fn layer() -> impl tower::layer::Layer<N, Service = Self> + Clone {
+        crate::layer::mk(Self::new)
+    }
+
+    pub fn layer_into<T>(
```
hawkw, line 57: maybe `layer_from_tuple` or something because the name of the inner service changed?

hawkw (approved):

linkerd/stack/src/watch.rs
olix0r: i think _into_ reads more clearly here

PR #2122, opened 2023-01-05
distribute: Add a backend cache

olix0r:
`NewDistribute` will be used to implement per-route traffic distribution policies so that, for instance, each request route may use a different traffic split. In this setup, we want each instance--re-constructed as configuration changes--to reuse a common set of concrete services so that load balancer state need not be lost when a policy changes.

This change adds a `distribute::BackendCache` module that produces a `NewDistribute` for an updated, cached set of backends.

### DIFF
--- a/linkerd/distribute/src/cache.rs
+++ b/linkerd/distribute/src/cache.rs
@@ -1,0 +1,76 @@
+use super::{params, NewDistribute};
+use linkerd_stack::{layer, NewService, Param};
+use parking_lot::Mutex;
+use std::{fmt::Debug, hash::Hash, sync::Arc};
+
+/// A [`NewService`] that produces [`NewDistribute`]s using a shared cache of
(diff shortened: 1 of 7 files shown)


linkerd/distribute/src/cache.rs
```diff
@@ -0,0 +1,76 @@
...
+                // Skip rebuilding targets we already have a stack for.
+                if cache.contains_key(addr) {
+                    continue;
+                }
+
+                let backend = self.0.new_backend.new_service(addr.clone());
+                cache.insert(addr.clone(), backend);
+            }
+        }
+
```
hawkw, line 67: i wonder if it's worth logging the cache change (at least the number of services added/removed) at `trace`/`debug`? not a blocker.

hawkw (approved):
looks good to me!

PR #2123, opened 2023-01-05
stack: Eliminate the `UpdateWatch` trait

olix0r:
We've learned that there's no need for a specialized `UpdateWatch` trait. Instead, this change uses a hierarchy of `NewService` instances. The first `NewService` builds an inner stack for each watch, so that we can isolate per-watch scope; and the second builds an inner stack each time the watch updates.

The `NewSpawnWatch::layer_into` builder alters the update-level target type by combining the watch-level and update-level target types.

This is intended to combine with #2122.

### DIFF
--- a/linkerd/stack/src/watch.rs
+++ b/linkerd/stack/src/watch.rs
@@ -3,35 +3,34 @@ use std::task::{Context, Poll};
 use tokio::sync::watch;
 use tracing::Instrument;
 
-/// Defines how to update a [`Service`] in the background as a [`watch::Receiver`].
-/// containing a `T-`typed target is updated.
-pub trait UpdateWatch<T>: Send + 'static {
-    /// The type of service built based on the watched target.
-    type Service: Clone + Default + Send + Sync + 'static;
-
-    /// Update the inner service with a new `target` value.
-    ///
-    /// If the inner service has not changed as a result of the update, this
-    /// method may return `None`. If `None` is returned, no update will be
-    /// published to the [`SpawnWatch`] service.
-    ///
-    /// This method is also called initially when the first [`SpawnWatch`]
-    /// service is constructed. In this case, it should generally not return
-    /// `None`, but if it does, the [`Default`] value of the [`Self::Service`]
-    /// type is used.
-    fn update(&mut self, target: &T) -> Option<Self::Service>;
-}
-
-/// Builds [`SpawnWatch`] services, where an inner [`Service`] is updated
-/// dynamically by a background task as a target wrapped in a
-/// [`watch::Receiver`] changes.
+/// Builds [`SpawnWatch`] services from a target that produces a `P`-typed
+/// [`watch::Receiver`].
 ///
-/// `N` is a [`NewService`] which produces types implementing [`UpdateWatch`],
-/// which define the logic for how to update the inner [`Service`].
+/// A background task is spawned that updates the inner service when the watch
+/// value changes.
 #[derive(Clone, Debug)]
 pub struct NewSpawnWatch<P, N> {
-    new_update: N,
+    inner: N,
     _marker: std::marker::PhantomData<fn(P)>,
+}
+
+/// Builds `NewFromTuple` instances within a `NewSpawnWatchStack`.
+///
+/// When we use a `T`-typed target to produce a `U`-typed `watch::Receiver`,
+/// this module converts the `U`-typed target value to a `P`-typed target for
+/// the inner stack. `P` must implement `From<(U, T)>`.
+#[derive(Clone, Debug)]
+pub struct NewWatchedFromTuple<P, N> {
+    inner: N,
+    _marker: std::marker::PhantomData<fn() -> P>,
+}
+
+/// Builds services by passing `P` typed values to an `N`-typed inner stack.
+#[derive(Clone, Debug)]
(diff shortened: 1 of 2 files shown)


linkerd/stack/src/watch.rs
```diff
@@ -3,37 +3,36 @@ use std::task::{Context, Poll};
...
+/// A background task is spawned that updates the inner service when the watch
+/// value changes.
 #[derive(Clone, Debug)]
 pub struct NewSpawnWatch<P, N> {
-    new_update: N,
+    inner: N,
     _marker: std::marker::PhantomData<fn(P)>,
 }
 
+/// Builds `NewFromTuple` instances within a `NewSpawnWatchStack`.
```
hawkw, line 17: I think you meant:
```suggestion
/// Builds `NewFromTuple` instances within a `NewSpawnWatch` stack.
```

linkerd/stack/src/watch.rs
```diff
@@ -3,37 +3,36 @@ use std::task::{Context, Poll};
...
+/// this module converts the `U`-typed target value to a `P`-typed target for
+/// the inner stack. `P` must implement `From<(U, T)>`.
+#[derive(Clone, Debug)]
+pub struct NewWatchedFromTuple<P, N> {
+    inner: N,
+    _marker: std::marker::PhantomData<fn() -> P>,
+}
+
+/// Builds services by passing `P` typed values to an `N`-typed inner stack.
+#[derive(Clone, Debug)]
```
hawkw, line 29: these `Clone` and `Debug` impls will require that `P` be `Clone`/`Debug` --- do we need manual impls, or is that not a problem here?

linkerd/stack/src/watch.rs
```diff
@@ -3,37 +3,36 @@ use std::task::{Context, Poll};
...
 ///
-/// `N` is a [`NewService`] which produces types implementing [`UpdateWatch`],
-/// which define the logic for how to update the inner [`Service`].
+/// A background task is spawned that updates the inner service when the watch
+/// value changes.
 #[derive(Clone, Debug)]
 pub struct NewSpawnWatch<P, N> {
-    new_update: N,
+    inner: N,
     _marker: std::marker::PhantomData<fn(P)>,
```
hawkw, line 14: nit, not important: this file has enough `PhantomData`'s in it that we may as well import it?

linkerd/stack/src/watch.rs
```diff
@@ -48,31 +47,50 @@ pub struct SpawnWatch<S> {
...
         }
     }
+
+    pub fn layer() -> impl tower::layer::Layer<N, Service = Self> + Clone {
+        crate::layer::mk(Self::new)
+    }
+
+    // Creates a new `NewSpawnWatch` layer that transforms the target type to
+    // `T` for the inner stack.
+    pub fn layer_into<T>(
```
hawkw, line 63: naming nit, not a big deal: i might prefer to call this `layer_from_tuple`, or something, since the inner `NewService` is `NewWatchedFromTuple` rather than `NewWatchedInto`...

linkerd/stack/src/watch.rs
```diff
@@ -48,31 +47,50 @@ pub struct SpawnWatch<S> {
...
+    ) -> impl tower::layer::Layer<N, Service = NewSpawnWatch<P, NewWatchedFromTuple<T, N>>> + Clone
```
hawkw, line 64: this is probably the weirdest rustfmt decision i've seen in a while, although i guess i get why it did that (not wanting to line break inside the return type)... :woman_shrugging:  probably not worth trying to make the type shorter just to avoid this...

linkerd/stack/src/watch.rs
```diff
@@ -48,31 +47,50 @@ pub struct SpawnWatch<S> {
...
             _marker: std::marker::PhantomData,
         }
     }
+
+    pub fn layer() -> impl tower::layer::Layer<N, Service = Self> + Clone {
+        crate::layer::mk(Self::new)
+    }
+
+    // Creates a new `NewSpawnWatch` layer that transforms the target type to
+    // `T` for the inner stack.
```
hawkw, line 62: tiny nit: looks like these were supposed to be doc comments?
```suggestion
    /// Creates a new `NewSpawnWatch` layer that transforms the target type to
    /// `T` for the inner stack.
```

hawkw (approved):
looks good to me --- i commented on some very minor nits.

linkerd/stack/src/watch.rs
olix0r: I think practically we always expect targets are clone but you are right that it would be more correcter to manual impl

linkerd/stack/src/watch.rs
olix0r: yeah, but into is much clearer at the callsite, in my opinion. The caller doesn't care which middleware is used to achieve the into.

olix0r: I suppose _from_targets_ could work. The _tuple_ is kind of a non-sequitur from the caller's point of view, though.

hawkw: fair enough, i'm fine with whatever you think is clearer :+1:

hawkw (approved):

PR #2126, opened 2023-01-05
addr: reference-count `NameAddr`s

hawkw:
A `NameAddr` consists of a DNS `Name` (a `String`) and a port number. The `Name` string is never mutated once the `NameAddr` is constructed, and `NameAddr`s are often cloned (and will be cloned more once #2095 is merged). Therefore, it makes sense to reference count the `Name` component of `NameAddr`s.

This branch wraps the `Name` component of `NameAddr`s in an `Arc` so that they can be cloned without performing a bytewise copy of the `String`.

This change was factored out from PR #2095 to land separately.

### DIFF
--- a/linkerd/addr/src/lib.rs
+++ b/linkerd/addr/src/lib.rs
@@ -5,6 +5,7 @@ use std::{
     fmt,
     net::{IpAddr, SocketAddr},
     str::FromStr,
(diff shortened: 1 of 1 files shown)


hawkw:
whoops, #2121 already did this.

PR #2128, opened 2023-01-06
Support router-scoped caches

olix0r:
`stack::NewRouter` reuses its inner stack for all `Router` instances. This makes it difficult to implement a cache that is scoped just to the router (and not shared across all routers).

This change:

* Removes the `linkerd_stack::router` module;
* Adds the `linkerd-router` crate;
* Adds an `ahash` cache implementation to the router crate with a convenience layer;
* The `RecognizeRoute` is now called `SelectRoute`;
* `Router` is now called `OneshotRoute`;
* `NewOneshotRoute` now uses `ExtractParam` to acquire a `SelectRoute`;
* A `NewCloneService` module is added to support lifting a stack to be shared across routes (i.e. to preserve the prior behavior).


linkerd/app/gateway/src/lib.rs
```diff
@@ -203,7 +203,10 @@ where
         .with_stack(
             // A router is needed so that we use each request's HTTP version
             // (i.e. after server-side orig-proto downgrading).
-            http.push(svc::NewRouter::layer(|(_, target)| RouteHttp(target)))
+            http.push_new_clone()
+                .push(svc::NewOneshotRoute::layer(
```
olix0r, line 207: needs a loadshed

linkerd/router/src/cache.rs
```diff
@@ -0,0 +1,89 @@
+use crate::NewService;
+use ahash::AHashMap;
+use parking_lot::Mutex;
+use std::{fmt::Debug, hash::Hash, marker::PhantomData, sync::Arc};
+
+/// A [`NewService`] that lazy builds stacks for each `K`-typed key.
```
hawkw, line 6: isn't this a `NewService` that produces `Cache` `NewService`s, rather than actually lazily building stacks for `K`s?

linkerd/router/src/cache.rs
```diff
@@ -0,0 +1,89 @@
...
+#[derive(Debug)]
+pub struct NewCache<K, N> {
+    inner: N,
+    _marker: PhantomData<fn(K)>,
+}
+
+/// A [`NewService`] that lazily builds an inner `S`-typed service for each
+/// `K`-typed target.
```
hawkw, line 14: should maybe explicitly state that the `S`-typed service is only built once per `K` and then cloned?

hawkw: maybe also worth documenting that this cache never evicts?

linkerd/router/src/cache.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl<T, K, N, M> NewService<T> for NewCache<K, N>
+where
+    T: Clone,
+    N: NewService<T, Service = M>,
+    M: NewService<K>,
+{
+    type Service = Cache<K, M, M::Service>;
+
+    fn new_service(&self, target: T) -> Self::Service {
```
hawkw, line 40: nit: should maybe inline?
```suggestion
    #[inline]
    fn new_service(&self, target: T) -> Self::Service {
```

linkerd/router/src/cache.rs
```diff
@@ -0,0 +1,89 @@
...
+    K: Eq + Hash + Clone,
+    N: NewService<K>,
+    N::Service: Clone,
+{
+    type Service = N::Service;
+
+    fn new_service(&self, key: K) -> Self::Service {
+        self.services
+            .lock()
+            .entry(key.clone())
```
hawkw, line 76: could be worth noting in the docs that `Cache`'s `NewService` impl will always clone the target type, even if there's already a cached service for it...

linkerd/router/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// Given a a request, returns the key matching this request.
+    ///
+    /// If no route matches the request, this method returns an error.
+    fn select(&self, req: &Req) -> Result<Self::Key, Self::Error>;
+}
+
+/// A [`NewService`] that builds `Route` services for targets that implement
+/// [`SelectRoute`].
+#[derive(Debug)]
+pub struct NewOneshotRoute<Sel, X, N> {
```
hawkw, line 31: good call on naming here :+1:

linkerd/router/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub fn layer_cached<K>(
+        extract: X,
+    ) -> impl layer::Layer<N, Service = NewOneshotRoute<Sel, X, NewCache<K, N>>> + Clone {
+        layer::mk(move |inner: N| NewOneshotRoute::new(extract.clone(), NewCache::new(inner)))
+    }
+}
+
+impl<T, Sel, X, N> NewService<T> for NewOneshotRoute<Sel, X, N>
+where
+    X: ExtractParam<Sel, T>,
```
hawkw, line 72: what is the difference between using `ExtractParam` here and just requiring that `T: Param<Sel>`? my guess is that it's to allow disambiguating when the target is `Param` of multiple types that implement `SelectRoute`?

should we add a comment explaining what the `ExtractParam` impl is for?

hawkw (approved):
Looks great, I commented on a few nits, but no blockers.

Do we want to add the required loadshedding layers in this PR, or is that intended for a follow-up?

linkerd/router/src/lib.rs
olix0r: i think ideally we wouldn't need extractparam but there are some places now where we use general target types (like `http::Accept`) where we don't want to implement `SelectRoute` on that target type. ideally, we'd use different target types, but that's a bigger change i think.

`ExtractParam` lets us use either approach -- either by providing an extractor or by passing `()` to use `T: Param<Sel>`

olix0r: actually that may not be entirely true... I'll take a closer look

olix0r: simplified the interface without changing the extract behavior -- the current uses would require more target mapping than i'd like. The extract stuff is more general than Param and can describe that behavior as well

PR #2142, opened 2023-01-09
Parameterize the load balancer stack

olix0r:
The load balancer/discovery modules in the concrete stacks use
hardcoded ewma parameters.

In preparation for making these parameters discoverable via the
API, this change modifies the balancer to accept this configuration
as a target param.

Furthermore, this change cleans up the balancer API to simplify the
way that we convert `Resolve` resolutions to `Discover` streams for
the balancer: the discovery buffer is spawned immediately instead
of using a `MakeService`. In practice, we flatten the 'make' future
into the discovery stream in the readycache, anyway; so there's no
functional difference.


linkerd/proxy/tcp/src/balance.rs
```diff
@@ -1,28 +1,86 @@
...
+    _marker: PhantomData<fn(I)>,
+}
+
+pub type Balance<I, S> = p2c::Balance<PeakEwmaDiscover<discover::Buffer<S>, CompleteOnResponse>, I>;
+
+// === impl NewBalance ===
+
+impl<I, R, N> NewBalance<I, R, N> {
+    // FIXME(ver)
+    const CAPACITY: usize = 1_000;
```
hawkw, line 33: can/should this also be on `EwmaConfig`? it's not technically a EWMA parameter, and it's tuning the resolution config rather than the balancer...and i'm not sure if we'd want this to be part of the discovered configuration anyway?

linkerd/proxy/tcp/src/balance.rs
```diff
@@ -1,28 +1,86 @@
...
+
+#[derive(Clone, Debug)]
+pub struct EwmaConfig {
+    pub default_rtt: Duration,
+    pub decay: Duration,
+}
+
+/// Configures a stack to resolve `T` typed targets to balance requests over
+/// `M`-typed endpoint stacks.
+#[derive(Debug)]
```
hawkw, line 20: nit, may not be important: derived `Debug` impl requires that `I: Debug`, which it shouldn't actually need...

linkerd/proxy/resolve/src/recover.rs
```diff
@@ -84,8 +84,8 @@ where
     type Future = ResolveFuture<T, E, R>;
 
     #[inline]
-    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        self.resolve.poll_ready(cx).map_err(Into::into)
+    fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
```
hawkw, line 88: maybe worth a comment here explaining why this is unconditionally ready?

linkerd/proxy/resolve/src/map_endpoint.rs
```diff
@@ -60,8 +60,8 @@ where
     type Future = ResolveFuture<T, R::Future, M>;
 
     #[inline]
-    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        self.resolve.poll_ready(cx)
+    fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
```
hawkw, line 64: maybe worth a comment explaining why this is unconditionally ready?

linkerd/proxy/http/src/balance.rs
```diff
@@ -1,59 +1,88 @@
...
+    type Service = Balance<ReqB, S>;
 
-    fn layer(&self, discover: D) -> Self::Service {
-        let instrument = PendingUntilFirstData::default();
-        let loaded = PeakEwmaDiscover::new(discover, self.default_rtt, self.decay, instrument);
-        Balance::from_rng(loaded, &mut thread_rng()).expect("RNG must be valid")
+    fn new_service(&self, target: T) -> Self::Service {
+        let EwmaConfig { default_rtt, decay } = target.param();
+        let disco = NewSpawnDiscover::new(Self::CAPACITY, self.resolve.clone(), self.inner.clone())
+            .new_service(target);
```
hawkw, line 72: style nit, take it or leave it: since the `NewSpawnDiscover` is only used once in the `NewService` impl for `the NewBalance` types...does this really need to be a whole `NewService` impl? couldn't we just have a function taking those arguments and returning a `discover::Buffer`?

or, is there an intent to compose the `NewSpawnDiscover` `NewService` with other `NewService`s later?

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -69,24 +52,17 @@ impl<N> Outbound<N> {
...
-                .push_on_service(http::balance::layer(
-                    crate::EWMA_DEFAULT_RTT,
-                    crate::EWMA_DECAY,
-                ))
-                .check_make_service::<Concrete, http::Request<http::BoxBody>>()
-                .push(svc::MapErr::layer(Into::into))
+                .check_new_service::<Endpoint, http::Request<_>>()
+                .push_new_clone()
+                .check_new_new::<Concrete, Endpoint>()
+                .check_new_new_service::<Concrete, Endpoint, http::Request<_>>()
```
hawkw, line 58: nit: isn't the `check_new_new` here duplicating the type assertion made by the `check_new_new_service` on the next line?

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -69,24 +52,17 @@ impl<N> Outbound<N> {
...
+                .push(endpoint::NewFromMetadata::layer(config.inbound_ips.clone()))
+                .check_new_new_service::<Concrete, (_, _), http::Request<_>>()
+                .push(http::NewBalance::layer(resolve))
+                .check_clone()
+                .check_new::<Concrete>()
+                .check_new_service::<Concrete, http::Request<_>>()
```
hawkw, line 64: similarly, isn't this equivalent to just
```suggestion
                .check_new_service::<Concrete, http::Request<_>>()
```

hawkw (commented):
overall, this makes sense! commented on some minor things.

linkerd/proxy/tcp/src/balance.rs
```diff
@@ -1,28 +1,86 @@
...
-            PeakEwmaDiscover::new(discover, default_rtt, decay, CompleteOnResponse::default());
-        Balance::from_rng(loaded, &mut thread_rng()).expect("RNG must be valid")
-    })
+    type Service = Balance<I, S>;
+
+    fn new_service(&self, target: T) -> Self::Service {
+        let EwmaConfig { default_rtt, decay } = target.param();
+        let disco = NewSpawnDiscover::new(Self::CAPACITY, self.resolve.clone(), self.inner.clone())
+            .new_service(target);
+        let disco = PeakEwmaDiscover::new(disco, default_rtt, decay, CompleteOnResponse::default());
```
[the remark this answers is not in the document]
olix0r: yep! noted in the TODOs above

linkerd/app/core/src/control.rs
```diff
@@ -167,41 +185,25 @@ mod resolve {
...
-        svc::layer::mk(move |endpoint| {
-            discover::resolve(
-                endpoint,
-                map_endpoint::Resolve::new(
-                    IntoTarget(()),
-                    recover::Resolve::new(recover.clone(), DnsResolve::new(dns.clone())),
-                ),
-            )
-        })
+        map_endpoint::Resolve::new(
```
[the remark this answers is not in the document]
olix0r: yes, should probably do that here so we can kill map_endpoint.

linkerd/app/outbound/src/http/concrete.rs
olix0r: i'm going to clean this up before review (see TODOs in description), but this differentiates whether a compile error is on the new or the service.

linkerd/app/outbound/src/http/concrete.rs
olix0r: differentiates if the compile error is on the service.

linkerd/proxy/http/src/balance.rs
olix0r: really, the goal is to have our modules expose stacks so they don't dictate how they're called--it makes them composable. agree that the current usage is a bit strained.

linkerd/proxy/tcp/src/balance.rs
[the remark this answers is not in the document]
hawkw: oh yup missed that one

linkerd/proxy/http/src/balance.rs
```diff
@@ -1,59 +1,9 @@
...
-#[derive(Debug)]
-pub struct Layer<A, B> {
-    decay: Duration,
-    default_rtt: Duration,
-    _marker: PhantomData<fn(A) -> B>,
-}
+pub type Body<B> = PendingUntilFirstDataBody<balance::Handle, B>;
 
-// === impl Layer ===
+pub type EwmaConfig = balance::EwmaConfig;
```
hawkw, line 6: what's up with the type alias? is there a plan for these types to differ based on protocol?

hawkw (approved):
looks great! i left comments on a handful of minor nits, but no blockers!

linkerd/proxy/balance/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+#[derive(Copy, Clone, Debug, Default)]
+pub struct EwmaConfig {
+    pub default_rtt: Duration,
+    pub decay: Duration,
+}
+
+/// Configures a stack to resolve targets to balance requests over `N`-typed
+/// endpoint stacks.
+#[derive(Debug)]
+pub struct NewBalancePeakEwma<C, Req, R, N> {
```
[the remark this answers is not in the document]
olix0r: i think of it more as 'balance (with) peak ewma'

linkerd/proxy/balance/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+
+[dependencies]
+futures = { version = "0.3", default-features = false }
+futures-util = "0.3"
+indexmap = "1"
+linkerd-error = { path = "../../error" }
+linkerd-proxy-core = { path = "../../proxy/core" }
+linkerd-stack = { path = "../../stack" }
+pin-project = "1"
+rand = "0.8"
```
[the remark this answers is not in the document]
olix0r: this is what we're using elsewhere at the moment. limiting it doesn't seem to help us (so a dep must pull in the defaults).

linkerd/proxy/http/src/balance.rs
olix0r: eh, it's basically just a `pub use` for the moment so callers don't need to know about the balance crate explicitly.

PR #2143, opened 2023-01-09
Disable musl in release build

olix0r:
Our cross-compiled musl builds do not currently compile (OOM). To unblock releases, this change disables these builds for now.


hawkw (approved):
fine for now if it fixes the release!

PR #2146, opened 2023-01-10
implement outbound policy

hawkw:


linkerd/service-profiles/src/lib.rs
```diff
@@ -25,6 +24,7 @@ pub mod http;
 mod proto;
 
 pub use self::client::Client;
+pub use linkerd_client_policy::{LogicalAddr, LookupAddr};
```
olix0r, line 27: I'd rather keep these things completely decoupled. Let's assume the logicaladdr and lookup types aren't going to be the same.

This PR should ~basically require no changes to anything in service-profiles (Though I think we'll need to cleanup some of the profile/discover stuff as a part of this work).

linkerd/client-policy/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub addr: Addr,
+}
+
+/// A profile lookup target.
+#[derive(Clone, Hash, Eq, PartialEq)]
+pub struct LookupAddr(pub Addr);
+
+/// A bound logical service address
+#[derive(Clone, Hash, Eq, PartialEq)]
+pub struct LogicalAddr(pub NameAddr);
```
olix0r, line 50: We should be using Meta/ref types for anything related to addrs in policy-land

linkerd/client-policy/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+    inner: watch::Receiver<Policy>,
+}
+
+// TODO(eliza): this is looking a lot like the `Profile` type...could be made
+// generic eventually...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Policy {
+    pub addr: Option<LogicalAddr>,
+    pub http_routes: Arc<[http::Route]>,
+    pub opaque_protocol: bool,
```
olix0r, line 28: How does this fit into the policy API in its current form? Where does this configuration come from?

linkerd/client-policy/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+use linkerd_proxy_api_resolve as resolve;
+use std::{fmt, net::SocketAddr, str::FromStr, sync::Arc};
+use tokio::sync::watch;
+
+type Backends = Arc<[Backend]>;
+
+#[derive(Clone, Debug)]
+pub struct Receiver {
+    inner: watch::Receiver<Policy>,
+}
```
olix0r, line 20: let's do without this stuff for now and focus on just the `Policy` type for the time being. (I think the profiles::Receiver type isn't so useful in general)

linkerd/client-policy/src/lib.rs
hawkw: there's a proto module converting this...

linkerd/client-policy/src/lib.rs
hawkw: yeah, i thought about getting rid of both of them

linkerd/client-policy/src/lib.rs
hawkw: the logical address is coming from the protobuf's `backend` field; turning that into a meta type will need a change in proxy-api as well.

olix0r: Yep. ~Everything about the proxy api draft is on the table to change :) This is why I want to work 'out' from the logical stacks -- let the proxy's needs dictate the API and not vice-versa.

hawkw: yeah, that makes sense.

PR #2154, opened 2023-01-12
annotate proxy errors with target metadata

hawkw:
In order to improve the proxy's debuggability, it's desirable that the
errors logged by the proxy and/or returned to HTTP clients include
relevant information about where in the proxy that error occurred. 
Currently, buffer layers are constructed with a string "name" parameter
that describes which part of the stack is behind that buffer, so that
the fail-fast errors returned by those buffers can contain the stack
layer's name. While this provides some useful context, it isn't a
complete solution, as this context is only added to fail-fast errors,
and not other errors such as I/O errors, Additionally, these names only
indicate where in the stack the error occurred, and don't provide
metadata describing *which instance* of that stack layer produced the
error --- e.g., there is no context describing the client address,
logical address, concrete address, or endpoint (as appropriate) that
produced the error.

Instead, we should annotate *all* errors with the following information:

- The name of the stack layer where the error occurred
- Target metadata including the logical, concrete, and endpoint
  addresses (as appropriate)

This branch introduces a new `AnnotateError` middleware in
`linkerd-stack`. This middleware consists of a `NewService` which
extracts a `Param` from the target and formats it into a
reference-counted `Arc<str>`, and the `Service` produced by that
`NewService`, which wraps any errors returned by the inner `Service`
with that context. An `ExtractParam` implementation can also be provided
to configure how the context is extracted from the target. In addition,
the `annotate_error` modules contains helpers for formatting a `Param`
provided by the target plus a string name describing a stack layer.

Furthermore, this branch removes the existing name parameter from buffer
layers, and replaces them with the `AnnotateError` middleware, which is
used to add target metadata *and* a string name describing the stack
layer to all errors. Additionally, the various `HttpRescue`
implementations have been changed slightly, so that when a specific
cause is found for an error, the entire error chain is still formatted
in the HTTP response `l5d-error` header, rather than _just_ the cause.
This ensures that all context added to an error is included in the
formatted message, rather than just the root cause (which lacks target
metadata).

### DIFF
--- a/linkerd/app/inbound/src/http/tests.rs
+++ b/linkerd/app/inbound/src/http/tests.rs
@@ -341,7 +341,7 @@ async fn h2_response_meshed_error_header() {
         .headers()
         .get(L5D_PROXY_ERROR)
         .expect("response did not contain L5D_PROXY_ERROR header");
-    assert_eq!(message, "service in fail-fast");
+    assert_eq!(message, "HTTP logical (127.0.0.1:80): service in fail-fast");
 
     // Drop the client and discard the result of awaiting the proxy background
     // task. The result is discarded because it hits an error that is related
@@ -423,7 +423,7 @@ async fn grpc_meshed_response_error_header() {
         .headers()
         .get(L5D_PROXY_ERROR)
         .expect("response did not contain L5D_PROXY_ERROR header");
-    assert_eq!(message, "service in fail-fast");
+    assert_eq!(message, "HTTP logical (127.0.0.1:80): service in fail-fast");
 
     // Drop the client and discard the result of awaiting the proxy background
     // task. The result is discarded because it hits an error that is related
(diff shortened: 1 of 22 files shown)


linkerd/app/inbound/src/http/tests.rs
```diff
@@ -341,7 +341,7 @@ async fn h2_response_meshed_error_header() {
         .headers()
         .get(L5D_PROXY_ERROR)
         .expect("response did not contain L5D_PROXY_ERROR header");
-    assert_eq!(message, "service in fail-fast");
+    assert_eq!(message, "HTTP logical (127.0.0.1:80): service in fail-fast");
```
hawkw, line 344: i'm open to bikeshedding this format, but it seemed reasonable to me....

olix0r (commented):
Is it possible to, instead of using this "naming" strategy, allow each stack to produce entirely different error _type_? This will make error handling/inspection simpler, I think. It also means we're not beholden to the error format of `<name> (<context>)`.

What if the `NewAnnotateError` stack takes an `ExtractParam` that produces, something like `Annotate<Fn(E) -> Error>` (i.e. a marker type wrapping an error converter). Then we could perhaps even use `MapErr` directly (i.e. replacing the AnnotateError service).

Then, each stack could define its own error _type_ and even do error inspection themselves.

hawkw:
Closing this branch out, as I'm working on a new version on top of #2158.

PR #2158, opened 2023-01-13
stack: add `AnnotateError` middleware

hawkw:
This branch introduces a new `AnnotateError` middleware in `linkerd-stack`, for adding context to errors returned by a wrapped `Service`.

The `annotate_error` module contains a new trait, `ErrorContext`, which is implemented by types that provide context to `Error`s by wrapping them with a new `Error` type, and the `AnnotateError` middleware itself. The middleware consists of a `NewService` which extracts a `Param` implementing `ErrorContext` and produces a `Service`, which wraps any errors returned by the inner `Service` using that `ErrorContext`.

Additionally, `NewAnnotateError` may be constructed with an `ExtractParam` implementation that configures how an `ErrorContext` implementation is produced from a target type, when the `Param`s themselves do not implement `ErrorContext`. Finally, a pre-made implementation of `ErrorContext`, called `FromTarget`, that produces new errors when the error type implements `From<(&T, Error)>` is provided. In this case, the error type can implement `From` with the appropriate target type, and the `annotate_error::layer_from_target` constructor can be used, without requiring an `ExtractParam` implementation to be written for each stack.

This change was extracted from PR #2154.

### DIFF
--- a/linkerd/stack/src/annotate_error.rs
+++ b/linkerd/stack/src/annotate_error.rs
@@ -62,31 +62,33 @@ type NewFromTarget<N, T, E> = NewAnnotateError<N, FromTarget<T, E>, fn(&T) -> Fr
 
 // === impl NewAnnotateError ===
 
-/// Returns a `Layer` that produces [`NewAnnotateError`] middleware which
-/// annotate errors by constructing an `E`-typed error that implements
-/// `From<(&T, Error)>`, where `T` is a stack target.
-///
-/// `T` must implement [`Clone`]
-pub fn layer_from_target<E, T, N>() -> impl layer::Layer<N, Service = NewFromTarget<N, T, E>> + Clone
-where
-    T: Clone,
-    E: for<'a> From<(&'a T, linkerd_error::Error)>,
-    E: Error + Send + Sync + 'static,
-{
-    NewAnnotateError::layer_with(
-        (|target: &T| FromTarget {
-            target: target.clone(),
-            _err: PhantomData,
-        }) as fn(&T) -> FromTarget<T, E>,
-    )
-}
-
 impl<N, C: ErrorContext> NewAnnotateError<N, C> {
     /// Returns a `Layer` that produces [`NewAnnotateError`] middleware which
     /// annotate errors using an `C`-typed [`Param`](super::Param) type's
     /// [`ErrorContext`] implementation.
     pub fn layer() -> impl layer::Layer<N, Service = Self> + Clone {
         NewAnnotateError::layer_with(())
+    }
+}
+
+impl<N, T, E> NewFromTarget<N, T, E> {
+    /// Returns a `Layer` that produces [`NewAnnotateError`] middleware which
+    /// annotate errors by constructing an `E`-typed error that implements
+    /// `From<(&T, Error)>`, where `T` is a stack target.
+    ///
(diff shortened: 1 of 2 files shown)


linkerd/stack/src/annotate_error.rs
```diff
@@ -0,0 +1,227 @@
...
+type NewFromTarget<N, T, E> = NewAnnotateError<N, FromTarget<T, E>, fn(&T) -> FromTarget<T, E>>;
+
+// === impl NewAnnotateError ===
+
+/// Returns a `Layer` that produces [`NewAnnotateError`] middleware which
+/// annotate errors by constructing an `E`-typed error that implements
+/// `From<(&T, Error)>`, where `T` is a stack target.
+///
+/// `T` must implement [`Clone`]
+pub fn layer_from_target<E, T, N>() -> impl layer::Layer<N, Service = NewFromTarget<N, T, E>> + Clone
```
olix0r, line 70: Is there any reason this can't be hung off of NewAnnotateError, too? Seems like it should be doable... I'd expect that to slightly simplify imports for callers.

linkerd/stack/src/annotate_error.rs
```diff
@@ -0,0 +1,227 @@
...
+    marker::PhantomData,
+    pin::Pin,
+    sync::Arc,
+    task::{Context, Poll},
+};
+
+/// A [`NewService`] that extracts an [`ErrorContext`] implementation from its target,
+/// and produces an [`AnnotateError`] middleware which wraps an inner [`Service`]'s
+/// error type with that [`ErrorContext`].
+pub struct NewAnnotateError<N, C, X = ()> {
```
olix0r, line 15: I have a slight bias to keeping `N` as the last param... In our stack the inner type can be very very large, so i try to keep the contextual types first so they're easier to spot in compile errors... e.g. `NewAnnotateError<MyErrorContext, (), Some<Very<Deep<Type<Signature<...>>>>>>`

That may be worth

hawkw: hmm, good point --- i put the `ExtractParam` type last so that it could default to `()`, but I don't think the default is actually that important. I'll change it.

linkerd/stack/src/annotate_error.rs
hawkw: i thought it wouldn't work because the impl has to be generic over `T`, which isn't one of `NewAnnotateError`'s type params, but it actually works fine. I changed it.

olix0r (approved):
🎉

linkerd/stack/src/annotate_error.rs
hawkw: oh, for the record, i remembered why this was a free function: it's to avoid this: https://github.com/linkerd/linkerd2-proxy/blob/d05b24adb01dce3256a0d69f496fa52dd83b515f/linkerd/app/outbound/src/http/endpoint.rs#L67-L71

ideally, we would want the invocation of `layer_from_target` to be generic over the `Error` type that's being made, rather than the (more complex) type of the _machinery for making that error_, so that we wouldn't have to write out the whole `FromTarget<T, E>` thing and could just write the name of the error type. But when the layer is constructed by an associated function on the `NewService` impl, it's not actually generic over the error type.

not a big deal, just figured i'd mention it.

PR #2162, opened 2023-01-17
add target metadata to error contexts

hawkw:
In order to improve the proxy's debuggability, it's desirable that the
errors logged by the proxy and/or returned to HTTP clients include
relevant information about where in the proxy that error occurred.
Currently, buffer layers are constructed with a string "name" parameter
that describes which part of the stack is behind that buffer, so that
the fail-fast errors returned by those buffers can contain the stack
layer's name. While this provides some useful context, it isn't a
complete solution, as this context is only added to fail-fast errors,
and not other errors such as I/O errors, Additionally, these names only
indicate where in the stack the error occurred, and don't provide
metadata describing *which instance* of that stack layer produced the
error --- e.g., there is no context describing the client address,
logical address, concrete address, or endpoint (as appropriate) that
produced the error.

Instead, we should annotate *all* errors with the following information:

- The name of the stack layer where the error occurred
- Target metadata including the logical, concrete, and endpoint
  addresses (as appropriate)

This branch uses the `AnnotateError` middleware added in PR #2158 to add
layers to the proxy which annotate errors with metadata extracted from targets.
Furthermore, this branch removes the existing name parameter from buffer
layers, and replaces them with the `AnnotateError` middleware, which is
used to add target metadata *and* a string name describing the stack
layer to all errors. Additionally, the various `HttpRescue`
implementations have been changed slightly, so that when a specific
cause is found for an error, the entire error chain is still formatted
in the HTTP response `l5d-error` header, rather than _just_ the cause.
This ensures that all context added to an error is included in the
formatted message, rather than just the root cause (which lacks target
metadata).


olix0r:
@hawkw i'm ambivalent about how to resolve the test issue. It seems okay to keep the string matching. But it's also not clear if we actually care that the failfast error message is present, do we? We just care that the header isn't empty. IDK...

hawkw:
yeah...if we care about the presence of the failfast message, we could change the tests to assert that the string contains it, rather than making an exact assertion, so they the tests don't care if the error contexts change a little? but i don't know if the presence of the failfast message is super important, either? i guess it's nice to be checking that the header value isn't a totally unrelated error, maybe?

olix0r (approved):

PR #2167, opened 2023-01-18
metrics: add response classification to latency metrics

hawkw:
Currently, the documentation for proxy metrics [states][1] that all
`response_`-prefixed metrics are labeled with response classifications.
However, the `response_latency_ms` histograms lack this label --- they
are labeled with status codes, but not with response classifications or
gRPC status labels.

This is because response latencies are calculated when the last `DATA`
frame of a response body is received, rather than when a `TRAILERS`
frame is received or the body stream ends (depending on whether the
response includes trailers). Response classifications, however, cannot
be determined until a `TRAILERS` frame is received or the body stream
terminates, as a `TRAILERS` frame may contain a `grpc-status` trailer
that influences response classification.

This branch adds response classification labels to the
`response_latency_ms` histogram metrics. In order to do so, it was
necessary to change how response latencies are recorded slightly. Now,
the `ResponseBody` type still _calculates_ the latency when the last
`DATA` frame of a response is received, but the latency is not actually
recorded in the histogram until a response classification is recorded.
This way, the response latency histogram can be part of the
`ClassMetrics` structure rather than the `StatusMetrics` structure.

Now that response latency metrics are labeled with a classification as
well as a status code, it is no longer necessary to maintain a hashmap
of metrics by status code which contains hashmaps of metrics by
classification, since all response metrics are now labeled with both a
status and a classification. Therefore, we can combine these two nested
hashmaps into a single map keyed by `(Option<Status>, Classification)`.
This makes the code a bit simpler and may be slightly more efficient due
to fewer pointer dereferences to traverse nested maps.

Fixes linkerd/linkerd2#9737

[1]: https://linkerd.io/2.12/reference/proxy-metrics/#response-labels

### DIFF
--- a/linkerd/app/integration/src/tests/telemetry.rs
+++ b/linkerd/app/integration/src/tests/telemetry.rs
@@ -440,7 +440,9 @@ where
 
     // assert the >=1000ms bucket is incremented by our request with 500ms
     // extra latency.
-    let labels = labels.label("status_code", 200);
+    let labels = labels
+        .label("status_code", 200)
+        .label("classification", "success");
     let mut bucket_1000 = labels
         .clone()
         .metric("response_latency_ms_bucket")
--- a/linkerd/app/integration/src/tests/transparency.rs
+++ b/linkerd/app/integration/src/tests/transparency.rs
@@ -361,6 +361,7 @@ macro_rules! http1_tests {
             let srv = server::http1()
                 .route_fn("/", |req| {
                     assert!(!req.headers().contains_key("x-foo-bar"));
+                    tracing::info!("request headers: {:?}", req.headers());
                     Response::builder()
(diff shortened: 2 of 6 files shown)


hawkw:
Wait, how does this change prevent latencies from being reported for streaming responses? The latency should still be reported when the stream ends, which is the current behavior as well.

hawkw:
After talking to @olix0r about it offline, I agree that this solution isn't workable. Since we measure latency based on time to first byte, but only classify responses based on trailers, this change would prevent latencies for streaming requests from being reported for much longer than they are currently. Instead, we should probably change how response classification works so that non-gRPC HTTP responses are classified based on the status code, allowing the classification to be recorded on the latency metrics.

PR #2168, opened 2023-01-19
add meshtls-fips feature flag

zaharidichev:


linkerd2-proxy/src/main.rs
```diff
@@ -10,6 +10,10 @@ compile_error!(
     "at least one of the following TLS implementations must be enabled: 'meshtls-boring', 'meshtls-rustls'"
 );
 
+// Emit a compile-time error if FIPS mode is enabled but meshtls-boring is not
+#[cfg(all(not(feature = "meshtls-boring"), all(feature = "meshtls-fips")))]
+compile_error!("'meshtls-boring' needs to be enabled to support FIPS mode");
```
hawkw, line 15: it seems like an alternative solution would just be to have `meshtls-fips` enable the `meshtls-boring` feature?

the question is whether we would want the `meshtls-fips` feature flag to eventually be able to enable FIPS support with other (non-BoringSSL) TLS implementations (in which case, `meshtls-fips` should not enable the `meshtls-boring` crate), or whether it's specifically configuring the BoringSSL TLS implementation...

linkerd/meshtls/Cargo.toml
```diff
@@ -9,6 +9,7 @@ publish = false
 [features]
 rustls = ["linkerd-meshtls-rustls", "__has_any_tls_impls"]
 boring = ["linkerd-meshtls-boring", "__has_any_tls_impls"]
+fips = ["linkerd-meshtls-boring/fips"]
```
hawkw, line 12: this feature flag will always enable the optional dependency on `linkerd-meshtls-boring` if it's enabled, even if the `boring` feature flag is not enabled. is that the intended behavior?

we could use a `?` feature (see [here](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features) for details) to only have the `meshtls` crate's `fips` feature enable the `linkerd-meshtls-boring` crate's `fips` feature _if_ some other feature flag (e.g. this crate's `boring` feature) enables the `linkerd-meshtls-boring` dependency):
```suggestion
fips = ["?linkerd-meshtls-boring/fips"]
```
i believe that would be the intended behavior here.

alternatively, if we wanted to have the `fips` feature always enable the dependency on `linkerd-meshtls-boring`, we may as well just have it enable this crate's `boring` feature flag, so that it automatically enables BoringSSL. this would make the feature flag combinations a bit simpler, but (as I discussed in my other comment) it would preclude the possibility to make this crate's `fips` feature enable FIPS support for _any_ TLS implementation that's enabled, in the future...

hawkw (changes requested):
I left a couple of comments discussing the feature flagging approach. I think we need to decide whether the `linkerd-meshtls` crate's `fips` feature should  specifically enable BoringSSL-based `meshtls` implementation, or whether it's intended to (potentially) enable FIPS compliance in any `meshtls` implementation (which is what it seems like is the intention here).

If the intent is to make the `fips` feature decoupled from the feature flags for enabling specific `meshtls` implementations, we need to change the way it enables the `linkerd-meshtls-boring` crate's `fips` feature to be a weak feature, so that the dependency on that crate is not also enabled. I left an inline comment discussing that.

Hopefully my comments make sense --- the interaction between feature flags and dependencies is a little bit complex, but I tried to explain it as best as I could. Let me know if you have any questions!

zaharidichev:
@hawkw thanks a lot. This was a really great explanation!

hawkw (approved):
this looks correct to me! i imagine @olix0r will also want to give it a review, though.

linkerd2-proxy/Cargo.toml
```diff
@@ -12,6 +12,7 @@ default = ["multicore", "meshtls-rustls"]
 multicore = ["tokio/rt-multi-thread", "num_cpus"]
 meshtls-boring = ["linkerd-meshtls/boring"]
 meshtls-rustls = ["linkerd-meshtls/rustls"]
+meshtls-fips = ["linkerd-meshtls/fips"]
```
olix0r, line 15: sorry, could we call this `meshtls/boring-fips` (and `linkerd-meshtls/boring-fips`)?

olix0r (approved):

PR #2173, opened 2023-01-20
configure buffers from target `Param`s

hawkw:
Buffer services in the proxy are constructed with a queue capacity and a
failfast timeout. Currently, these parameters are configured based on
the proxy's config structs (parsed from environment variables), and each
buffer `Layer` is created with a fixed buffer configuration depending on
which stack it is used in, so all buffers constructed by that stack will
have the same configuration. In the future, we would like to allow these
parameters to be configured per-target by the control plane.

In order to support configuring buffer parameters dynamically, this
branch introduces a `NewQueue` type, which implements `NewService` and
produces `Buffer` + `FailFast` services configured using a `QueueConfig`
`Param` extracted from the target. This will allow target types to
implement `Param<QueueConfig>` to dynamically configure the buffer
parameters.

In order to continue supporting the current state of the world (i.e.,
fixed buffer parameters) without adding them to every target type,
`NewQueue`'s `NewService` implementation is generic over an
`ExtractParam` implementation to control how the `QueueConfig` params
are extracted from targets. Convenience constructors are provided for
`()` as the `ExtractParam` type, expecting that `T: Param<QueueConfig>`,
and for `CloneParam` as the `ExtractParam` type, to clone the same set
of params every time a buffer is constructed. Currently, all `NewQueue`
layers constructed in the proxy use `CloneParam`, to continue
implementing the existing fixed configuration behavior.

## Notes

* This branch also removes the `name` field passed when constructing
  buffers. The name is no longer passed to `FailFast` and doesn't
  currently do anything, but is still part of the `push_buffer` methods.
  Since this PR changed how buffers are constructed anyway, it didn't
  make sense to continue taking a `name` field. This may cause a small
  merge conflict with #2162, but it shouldn't be a huge deal.

* *A note on naming*: I decided to name the `NewService` type `NewQueue`
  rather than `NewBuffer`, and renamed the configuration params from
  `BufferConfig` to `QueueConfig`. I thought it was clearer to refer to
  these services as _queues_, rather than as _buffers_ (despite the fact
  that Tower calls the service `Buffer`), because the naming
  `BufferConfig` could be interpreted as configuring the byte buffers
  that bytes from the network are read into, especially for readers who
  aren't familiar with Tower's naming scheme. Also, @olix0r's draft PR
  #2151 named the client policy configuration type for buffer layer
  configs `Queue` rather than `Buffer`, so I thought I'd go with that
  naming here as well.

  However, I'm not super attached to this naming scheme, and am open to
  bikeshed over it. I can rename the uses of "queue" to "buffer" if
  consistency with Tower's naming scheme is important to us.

### DIFF
--- a/linkerd/stack/src/queue.rs
+++ b/linkerd/stack/src/queue.rs
@@ -1,7 +1,7 @@
 use crate::{
     failfast::{self, FailFast},
     layer::{self, Layer},
-    BoxService, CloneParam, ExtractParam, NewService, Param, Service,
+    BoxService, CloneParam, ExtractParam, NewService, Service,
 };
 use linkerd_error::Error;
 use std::{fmt, marker::PhantomData, time::Duration};
@@ -18,7 +18,7 @@ pub struct QueueConfig {
     pub failfast_timeout: Duration,
 }
 
-pub struct NewQueue<N, Req, X = ()> {
+pub struct NewQueue<X, Req, N> {
     inner: N,
     extract: X,
     _req: PhantomData<fn(Req)>,
@@ -28,7 +28,7 @@ pub type Queue<Req, Rsp> = failfast::Gate<Buffer<BoxService<Req, Rsp, Error>, Re
 
 // === impl NewQueue ===
 
-impl<T, N, X, Req> NewService<T> for NewQueue<N, Req, X>
+impl<T, X, Req, N> NewService<T> for NewQueue<X, Req, N>
 where
     Req: Send + 'static,
     X: ExtractParam<QueueConfig, T>,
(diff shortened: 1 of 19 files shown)


linkerd/stack/src/queue.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// The number of requests (or connections, depending on the context) that
+    /// may be buffered
+    pub capacity: usize,
+
+    /// The maximum amount of time a request may be buffered before failfast
+    /// errors are emitted.
+    pub failfast_timeout: Duration,
+}
+
+pub struct NewQueue<N, Req, X = ()> {
```
olix0r, line 21: I generally prefer to keep the service types last (for type readability in debug output). I don't think the default parameter is really worth it in this case.

linkerd/stack/src/queue.rs
```diff
@@ -0,0 +1,117 @@
...
+            capacity,
+            failfast_timeout,
+        } = self.extract.extract_param(&target);
+        let buf = layer::mk(move |inner| Buffer::new(BoxService::new(inner), capacity));
+        let buf = FailFast::layer_gated(failfast_timeout, buf);
+        buf.layer(self.inner.new_service(target))
+    }
+}
+
+impl<T: Param<QueueConfig>, Req> NewQueue<T, Req> {
```
olix0r, line 54: This type signature doesn't look right to me. `T` here is the NewService type. We don't want any type constraints here. In general, we should use `N` for the primary inner new service.

linkerd/stack/src/queue.rs
```diff
@@ -0,0 +1,117 @@
...
+    #[inline]
+    #[must_use]
+    pub fn layer_fixed(params: QueueConfig) -> impl Layer<T, Service = Self> + Clone {
+        Self::layer_with(CloneParam(params))
+    }
+}
+
+impl<T, Req, X> NewQueue<T, Req, X>
+where
+    X: ExtractParam<QueueConfig, T> + Clone,
```
olix0r, line 76: Similarly

olix0r (commented):
I'll push fixes for these suggestions.

olix0r (approved):

PR #2174, opened 2023-01-21
Ver/eliza/error context

olix0r:
`AnnotateError` is really, really similar to `MapErr`. This change
merges the implementations and cleans up some of the caller type
boilerplate.

This also includes some changes to error formatting.

I've removed the http version param constraints from Endpoint -- we should have that information elsewhere on the stack.

### DIFF
--- a/linkerd/app/outbound/src/endpoint.rs
+++ b/linkerd/app/outbound/src/endpoint.rs
@@ -293,11 +293,7 @@ impl<S> Outbound<S> {
 
         opaque.push_detect_http(http).map_stack(|_, _, stk| {
             stk.instrument(|e: &tcp::Endpoint| info_span!("forward", endpoint = %e.addr))
-                .push(svc::NewAnnotateError::<
-                    svc::annotate_error::FromTarget<_, ForwardError>,
-                    _,
-                    _,
-                >::layer_from_target())
+                .push(svc::map_err::layer_new_from_target::<ForwardError, _, _>())
                 .push_on_service(svc::MapErr::layer(Error::from))
                 .push_on_service(svc::BoxService::layer())
                 .push(svc::ArcNewService::layer())
--- a/linkerd/app/outbound/src/endpoint.rs
+++ b/linkerd/app/outbound/src/endpoint.rs
@@ -293,7 +293,7 @@ impl<S> Outbound<S> {
 
         opaque.push_detect_http(http).map_stack(|_, _, stk| {
             stk.instrument(|e: &tcp::Endpoint| info_span!("forward", endpoint = %e.addr))
-                .push(svc::map_err::layer_new_from_target::<ForwardError, _, _>())
+                .push(svc::NewMapErr::layer_from_target::<ForwardError, _>())
                 .push_on_service(svc::MapErr::layer(Error::from))
                 .push_on_service(svc::BoxService::layer())
                 .push(svc::ArcNewService::layer())
--- a/linkerd/app/outbound/src/endpoint.rs
+++ b/linkerd/app/outbound/src/endpoint.rs
@@ -39,15 +39,15 @@ pub struct FromMetadata<P, N> {
 }
 
 #[derive(Debug, thiserror::Error)]
+#[error("endpoint {addr}: {source}")]
 pub struct EndpointError {
     addr: Remote<ServerAddr>,
     #[source]
     source: Error,
-    protocol: Option<http::Version>,
 }
 
 #[derive(Debug, thiserror::Error)]
-#[error("forward ({}): {}", addr, source)]
+#[error("forwarding to {addr}: {source}")]
 pub struct ForwardError {
     addr: Remote<ServerAddr>,
     #[source]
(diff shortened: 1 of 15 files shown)


linkerd/app/outbound/src/endpoint.rs
```diff
@@ -319,30 +303,13 @@ impl<S> Outbound<S> {
...
-        let Self {
-            addr,
-            protocol,
-            source,
-        } = self;
-        let protocol = match protocol {
-            Some(http::Version::Http1) => "HTTP/1",
-            Some(http::Version::H2) => "HTTP/2",
-            None => "opaque",
-        };
```
hawkw, line 333: the reason that I included protocols here is because the `EndpointError` may be handled by a `ClientRescue` layer without being wrapped in other stacks' higher-level error contexts, so i thought it was desirable to include that information here. i'm fine with removing it for simplicity's sake, though.

linkerd/stack/src/map_err.rs
```diff
@@ -1,58 +1,263 @@
...
+            inner,
+            extract,
+            _cx: PhantomData,
+        }
+    }
+}
+
+impl<W, X: Clone, N> NewMapErr<W, X, N> {
+    /// A `NewService` layer that extracts a `WrapErr` implementation via the
+    /// `X`-typex `ExtractParam`.
```
hawkw, line 82: typo:
```suggestion
    /// `X`-typed `ExtractParam`.
```

linkerd/stack/src/map_err.rs
```diff
@@ -1,58 +1,263 @@
...
+impl<W, S> MapErr<W, S> {
+    pub fn new(inner: S, wrap: W) -> Self {
+        Self { inner, wrap }
+    }
+
+    pub fn layer(wrap: W) -> impl super::layer::Layer<S, Service = Self> + Clone
+    where
+        W: Clone,
+    {
+        super::layer::mk(move |inner| Self::new(inner, wrap.clone()))
```
hawkw, line 169: take it or leave it: it seems like `NewMapErr` could also have this constructor, using `CloneParam`? perhaps that should be the default `layer()` constructor to keep the "classic" `NewMapErr::layer(|error| { ... })` behavior, and the `ExtractParam` impls could be `layer_param` or something?

OTTOH we don't currently use a fn constructor with `NewMapErr`, so maybe it's not important...

linkerd/stack/src/map_err.rs
```diff
@@ -1,58 +1,263 @@
...
+        OutE::from((&self.target, error.into()))
+    }
+}
+
+impl<T: Clone, E> Clone for WrapFromTarget<T, E> {
+    fn clone(&self) -> Self {
+        Self {
+            target: self.target.clone(),
+            _err: PhantomData,
+        }
```
hawkw, line 261: nit: should probably have a `Debug` impl as well, since otherwise it's going to break the `NewService`'s `Debug` impl?

hawkw (approved):
overall, this looks good --- i think unifying `AnnotateError` and `MapErr` makes sense. i left a few minor nits and suggestions, but this seems good to me overall.

linkerd/stack/src/map_err.rs
```diff
@@ -1,58 +1,263 @@
...
+    T: Clone,
+    WrapFromTarget<T, E>: WrapErr<Error> + Clone,
+{
+    NewMapErr::layer_with(
+        (|t: &T| WrapFromTarget {
+            target: t.clone(),
+            _err: PhantomData,
+        }) as ExtractWrapFromTarget<T, E>,
+    )
+}
```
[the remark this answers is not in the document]
olix0r: i think i tried it and the compiler complained that it couldn't infer the original types.

linkerd/app/outbound/src/endpoint.rs
olix0r: ooooh. hmm. okay. I need to refresh on that. In any case, the target types are going to change, so we can keep it simple for now and then polish as it settles.

linkerd/stack/src/map_err.rs
olix0r: You can still use `MapErr::layer` for the original behavior afaiu.

linkerd/stack/src/map_err.rs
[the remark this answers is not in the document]
hawkw: oh well! nevermind then.

linkerd/stack/src/map_err.rs
```diff
@@ -1,58 +1,263 @@
...
 /// Like `tower::util::MapErr`, but with an implementation of `Proxy`.
 #[derive(Clone, Debug)]
-pub struct MapErr<S, F> {
+pub struct MapErr<W, S> {
     inner: S,
-    f: F,
+    wrap: W,
+}
+
+/// Futures returned by [`AnnotateError`].
```
hawkw, line 52: docs need to be updated here
```suggestion
/// Futures returned by [`MapErr`].
```

linkerd/stack/src/map_err.rs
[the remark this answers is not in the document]
olix0r: found a way around this problem

linkerd/stack/src/map_err.rs
```diff
@@ -95,21 +95,26 @@ impl<W, N> NewMapErr<W, (), N> {
...
-    WrapFromTarget<T, E>: WrapErr<Error> + Clone,
-{
-    NewMapErr::layer_with(
-        (|t: &T| WrapFromTarget {
-            target: t.clone(),
-            _err: PhantomData,
-        }) as ExtractWrapFromTarget<T, E>,
-    )
+// We don't actually care about the `()` types here, but they help avoid
+// inference problems.
```
hawkw, line 99: smart!

PR #2181, opened 2023-01-24
Error type suggestions for #2162

olix0r:
* Move HTTP version logging into the `Respond` handler (since it already includes all of this information). This also let's us be explicit about gRPC. This removes needing to wire it through elsewhere.
* All error types are defined in the module in which they are used, even when they exactly match error types elsewhere. As we move forward with the upcoming changes, we want each stack to be as independent as possible.
* `http::Version` now implements `Debug` with full "HTTP/1" or "HTTP/2" strings. We change uses of `Display` to `Debug`.
* Remove `RouteError` types for now. They're not really critical for debugging. We may restore them later as those stacks evolve.

### DIFF
--- a/linkerd/app/inbound/src/http/router.rs
+++ b/linkerd/app/inbound/src/http/router.rs
@@ -244,10 +244,7 @@ impl<C> Inbound<C> {
                 .check_new_service::<Logical, http::Request<http::BoxBody>>()
                 .instrument(|t: &Logical| {
                     let name = t.logical.as_ref().map(tracing::field::display);
-                    match t.http {
-                        http::Version::H2 => debug_span!("http2", name),
-                        http::Version::Http1 => debug_span!("http1", name),
-                    }
+                    debug_span!("http", name)
                 })
                 // Routes each request to a target, obtains a service for that target, and
                 // dispatches the request.
--- a/linkerd/app/inbound/src/http/router.rs
+++ b/linkerd/app/inbound/src/http/router.rs
@@ -7,7 +7,7 @@ use linkerd_app_core::{
     transport::{self, ClientAddr, Remote, ServerAddr},
     Error, Infallible, NameAddr, Result,
 };
-use std::{fmt, net::SocketAddr, sync::Arc};
+use std::{fmt, net::SocketAddr};
 use tracing::{debug, debug_span};
 
 /// Describes an HTTP client target.
@@ -60,15 +60,6 @@ struct ClientRescue;
 struct LogicalError {
     logical: Option<NameAddr>,
     addr: Remote<ServerAddr>,
-    http: http::Version,
-    #[source]
-    source: Error,
-}
-
-#[derive(Debug, thiserror::Error)]
-#[error("route ({:?}): {}", .route_labels, .source)]
-struct RouteError {
-    route_labels: Arc<std::collections::BTreeMap<String, String>>,
     #[source]
     source: Error,
 }
@@ -177,7 +168,6 @@ impl<C> Inbound<C> {
                         .push_on_service(http::BoxResponse::layer())
                         .push(classify::NewClassify::layer())
                         .push_http_insert_target::<profiles::http::Route>()
-                        .push(svc::NewMapErr::layer_from_target::<RouteError, _>())
                         .push_map_target(|(route, profile)| ProfileRoute { route, profile })
                         .push_on_service(svc::MapErr::layer(Error::from))
                         .into_inner(),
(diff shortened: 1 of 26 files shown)


olix0r:
Some of this is split out (against main) in https://github.com/linkerd/linkerd2-proxy/pull/2182.

linkerd/app/inbound/src/http/router.rs
```diff
@@ -60,15 +60,6 @@ struct ClientRescue;
...
     logical: Option<NameAddr>,
     addr: Remote<ServerAddr>,
-    http: http::Version,
-    #[source]
-    source: Error,
-}
-
-#[derive(Debug, thiserror::Error)]
-#[error("route ({:?}): {}", .route_labels, .source)]
-struct RouteError {
```
hawkw, line 70: I only added this error type because the ticket for error contexts listed "logical/route/concrete/endpoint" as the desired contexts to include in errors...do we not want to add a route error context after all, or is that still something we care about?

linkerd/app/outbound/src/tcp/concrete.rs
```diff
@@ -1,13 +1,25 @@
...
 use std::time;
 use tracing::info_span;
 
+#[derive(Debug, thiserror::Error)]
+#[error("concrete service {addr}: {source}")]
+pub struct ConcreteError {
+    addr: ConcreteAddr,
+    #[source]
+    source: Error,
+}
```
hawkw, line 21: what's the motivation for having separate versions of `ConcreteError`, `LogicalError`, and `EndpointError` for the HTTP and TCP stacks when they are now identical?

linkerd/proxy/http/src/version.rs
```diff
@@ -21,12 +21,11 @@ impl std::convert::TryFrom<http::Version> for Version {
...
 
-// A convenience for tracing contexts.
-impl std::fmt::Display for Version {
+impl std::fmt::Debug for Version {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            Self::Http1 => write!(f, "1.x"),
-            Self::H2 => write!(f, "h2"),
+            Self::Http1 => write!(f, "HTTP/1"),
+            Self::H2 => write!(f, "HTTP/2"),
```
hawkw, line 28: _technically_ this feels more correct as a `Display` impl to me, since I feel like the general idiom for `Debug` is to mimic the way the type appears in Rust source code...but if this was the `Display` impl, we would just basically never use the `Debug` impl, and always using this formatting seems fine to me, so it's not a big deal. probably better to just use this formatting consistently.

hawkw (approved):
seems fine to me! i left a couple minor comments, but no blockers.

linkerd/proxy/http/src/version.rs
olix0r: FWIW this is what the `http` crate's `Version` does -- i'm assuming because you'll never see `HTTP/2` anywhere in a request stream...

linkerd/app/outbound/src/tcp/concrete.rs
olix0r: > All error types are defined in the module in which they are used, even when they exactly match error types elsewhere. As we move forward with the upcoming changes, we want each stack to be as independent as possible.

As I mentioned when we did a read-through of https://github.com/linkerd/linkerd2-proxy/tree/ver/policy-client-sketch, the current strategy (to support modularity, refactoring, etc) is to focus on decoupling over reuse. We'll be doing the same type of decoupling with target types.

If they really are common, we should _also_ split out the common parts of the stack impl so that a single module can hold it all.

Basically: we should drive towards making each stack module as self-contained as possible.

linkerd/app/inbound/src/http/router.rs
olix0r: Gotcha. There probably wasn't much thought behind that description.

> Remove RouteError types for now. They're not really critical for debugging. We may restore them later as those stacks evolve.

I think the general service/endpoint coordinates are more sufficient for what we really need for diagnostics/error reporting. I want to hold off on doing route-specific things until we see how those types start to settle. I suspect we may want to omit them in the short term--they'll add a ton of verbosity to already-busy error messages. This could change as routes need to actually fail requests with errors, which I don't believe is really possible (other than load shedding on service availability).

hawkw: cool, sounds good to me!

PR #2182, opened 2023-01-24
Update HTTP error responder to log version info

olix0r:
The HTTP error responder already knows whether a request is considered HTTP/1, HTTP/2, or gRPC.

This change updates these log messages to include the relevant protocol information for each failed request.

Furthermore, this change updates the `http::Version` enum to implement `Debug` with full HTTP version strings like "HTTP/1" and "HTTP/2". We then remove the version for debug spans, as it's needlessly verbose for most uses. When debugging, we can identify the version based on specific events in a span. It need not exist on every event on the span.

### DIFF
--- a/linkerd/app/core/src/errors/respond.rs
+++ b/linkerd/app/core/src/errors/respond.rs
@@ -385,7 +385,12 @@ where
         };
 
         let rsp = info_span!("rescue", client.addr = %self.client_addr()).in_scope(|| {
-            tracing::info!(error, "Request failed");
+            if self.is_grpc {
+                let version = self.version;
+                tracing::info!(error, "{version:?} request failed",);
+            } else {
+                tracing::info!(error, "gRPC request failed");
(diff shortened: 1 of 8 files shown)


linkerd/app/core/src/errors/respond.rs
```diff
@@ -385,7 +385,12 @@ where
         };
 
         let rsp = info_span!("rescue", client.addr = %self.client_addr()).in_scope(|| {
-            tracing::info!(error, "Request failed");
+            if self.is_grpc {
```
kleimkuhler, line 388: Looks like the log lines indicate we want to check if this is _not_ gRPC:

```suggestion
            if !self.is_grpc {
```

olix0r: Thank you!

kleimkuhler (approved):

PR #2186, opened 2023-01-25
inbound: connections wait for ServerPolicy discovery

hawkw:
Currently, when the inbound proxy gets a connection to a port it has not
previously configured, it starts doing ServerPolicy discovery for that
port but then uses the default policy until the configuration is synced
from the control plane. This behavior is fairly complex to reason about
and surprising for users. Therefore, we would like to remove this
defaulting behavior from the inbound proxy.

This branch changes how the inbound proxy performs policy discovery.
Rather than spawning the lookup in a background task and using a default
policy until a policy is discovered, the inbound proxy will now wait
until the policy for a port is discovered before continuing to process a
connection on that port. In terms of the inbound stack, this means that
policy discovery is now a `MakeService` rather than a `NewService`. The
cache is still used when a policy discovery watch has already been
started.

This branch does not *completely* remove all default policy
configuration from the proxy. The default policy environment variables
are still used when the proxy is configured without a control plane
policy controller address. However, this mode will be removed in a
subsequent PR. I thought it was preferable to limit the scope of this
change to just making inbound connection handling wait for policy
discovery, and remove the defaulting behavior completely as a separate
change, so that the individual changes are easier to review. However, if
we'd prefer to do this atomically, I can make that change as part of
this branch as well.


linkerd/app/inbound/src/direct.rs
```diff
@@ -80,6 +80,12 @@ struct TlsParams {
     identity: identity::Server,
 }
 
+#[derive(Debug, Clone)]
+struct WithTransportHeader {
```
hawkw, line 84: this is kinda ugly and i would like to figure out something nicer

linkerd/app/inbound/src/policy.rs
```diff
@@ -38,9 +41,19 @@ pub struct ServerUnauthorized {
     server: Arc<Meta>,
 }
 
+/// Returns the traffic policy configured for the destination address.
 pub trait GetPolicy {
-    // Returns the traffic policy configured for the destination address.
-    fn get_policy(&self, dst: OrigDstAddr) -> AllowPolicy;
+    type Error: Into<Error>;
```
olix0r, line 46: What are the tradeoffs between including an Error type attribute and hardcoding the `Error` type. It seems like it could be unnecessary boilerplate for non-library use cases.

linkerd/app/inbound/src/policy/store.rs
```diff
@@ -112,29 +113,44 @@ where
     S::ResponseBody:
         http::HttpBody<Data = tonic::codegen::Bytes, Error = Error> + Default + Send + 'static,
 {
-    fn get_policy(&self, dst: OrigDstAddr) -> AllowPolicy {
+    type Future = future::Either<
+        future::Ready<Result<AllowPolicy, Self::Error>>,
+        future::BoxFuture<'static, Result<AllowPolicy, S::Error>>,
+    >;
+    type Error = S::Error;
+    fn get_policy(&self, dst: OrigDstAddr) -> Self::Future {
```
olix0r, line 121:
```suggestion

    fn get_policy(&self, dst: OrigDstAddr) -> Self::Future {
```

linkerd/app/inbound/src/accept.rs
```diff
@@ -46,6 +49,19 @@ impl<N> Inbound<N> {
...
+                    tracing::debug!(policy = ?&*policy.borrow(), "Accepted");
+                    Accept {
+                        client_addr: t.param(),
+                        orig_dst_addr: t.param(),
+                        policy,
+                    }
+                })
+                .push(policy::Discover::layer(policies))
+                .push(svc::MapErr::layer(Into::into as fn(_) -> Error))
+                .into_new_service()
```
olix0r, line 63: How does backpressure work on this service? Is this discovery cached/reused?

linkerd/app/inbound/src/accept.rs
```diff
@@ -46,6 +49,19 @@ impl<N> Inbound<N> {
     {
         self.map_stack(|cfg, rt, accept| {
             accept
+                .push_on_service(svc::MapErr::layer(Into::into as fn(_) -> Error))
```
olix0r, line 52:
```suggestion
                .push_on_service(svc::MapErr::layer_boxed())
```

linkerd/app/inbound/src/accept.rs
```diff
@@ -46,6 +49,19 @@ impl<N> Inbound<N> {
...
+                .push_map_target(|(policy, t): (AllowPolicy, T)| {
+                    tracing::debug!(policy = ?&*policy.borrow(), "Accepted");
+                    Accept {
+                        client_addr: t.param(),
+                        orig_dst_addr: t.param(),
+                        policy,
+                    }
+                })
+                .push(policy::Discover::layer(policies))
+                .push(svc::MapErr::layer(Into::into as fn(_) -> Error))
```
olix0r, line 62:
```suggestion
                .push(svc::MapErr::layer_boxed())
```

linkerd/app/inbound/src/direct.rs
```diff
@@ -211,6 +220,10 @@ impl<N> Inbound<N> {
                         )
                         .into_inner(),
                 )
+                .push(policy::Discover::layer(policies))
+                .push(svc::MapErr::layer(Into::into))
```
olix0r, line 224:
```suggestion
                .push(svc::MapErr::layer_boxed())
```

linkerd/app/inbound/src/policy.rs
hawkw: this was lifted from `GetProfile` for consistency. it's probably better off just always turning the error into an `Error` though.

linkerd/app/inbound/src/accept.rs
hawkw: `policies` here is a `Store`, which is (still) an idle cache of resolutions. `Store` will still still cache and reuse the discovery the way it did previously; the only difference in `Store`'s behavior is that we now return a future that waits for the _initial_ watch to start, rather than returing the default policy if the watch has not yet started: https://github.com/linkerd/linkerd2-proxy/blob/d64d41639649f3d6ec27b550e4099120cb064ae5/linkerd/app/inbound/src/policy/store.rs#L121-L154

The `Discover` service doesn't advertise meaningful backpressure because the source of backpressure would be the tonic client, and tonic clients don't advertise backpressure...

hawkw:
Looking into potentially changing the `ServerPolicy` code to use the `NewCachedDiscover` type added in #2195, rather than implementing its own caching, although this would require figuring out a nice way to proactively start watches for pre-configured inbound ports. Will figure that out after removing the defaulting behavior.

linkerd/app/inbound/src/server.rs
```diff
@@ -44,6 +44,8 @@ impl Inbound<()> {
         GSvc::Error: Into<Error>,
         GSvc::Future: Send,
         P: profiles::GetProfile<Error = Error>,
+        O: policy::GetPolicy + Clone + Send + Sync + 'static,
+        O::Future: Unpin + Send,
```
olix0r, line 48: We probably want to move these up onto the trait alias.

linkerd/app/admin/src/stack.rs
```diff
@@ -85,13 +89,15 @@ impl Config {
         R: FmtMetrics + Clone + Send + Sync + Unpin + 'static,
         B: Bind<ServerConfig>,
         B::Addrs: svc::Param<Remote<ClientAddr>> + svc::Param<Local<ServerAddr>>,
+        P: inbound::policy::GetPolicy + Send + Sync + 'static,
+        P::Future: Send,
     {
         let (listen_addr, listen) = bind.bind(&self.server)?;
 
+        let (ready, latch) = crate::server::Readiness::new();
```
olix0r, line 97: Did this need to move? It seemed a little clearer where it was.

hawkw: i think this moved because at one point, i had changed the function to spawn rather than being `async fn` and half of the readiness latch needed to be outside the async block. it can move back now since that's no longer needed.

linkerd/app/inbound/src/policy.rs
```diff
@@ -38,9 +41,11 @@ pub struct ServerUnauthorized {
...
 
-pub trait GetPolicy {
-    // Returns the traffic policy configured for the destination address.
-    fn get_policy(&self, dst: OrigDstAddr) -> AllowPolicy;
+/// Returns the traffic policy configured for the destination address.
+pub trait GetPolicy: Clone + Send + Sync + 'static {
+    type Future: Future<Output = Result<AllowPolicy, Error>> + Unpin + Send;
+
+    fn get_policy(&self, target: OrigDstAddr) -> Self::Future;
 }
```
olix0r, line 49: We should provide a blanket impl for all `Service<OrigDstAddr>` -- this lets us use middlewares and `svc::mk` etc.

linkerd/app/inbound/src/direct.rs
```diff
@@ -484,3 +485,21 @@ impl<T> InsertParam<tls::ConditionalServerTls, T> for TlsParams {
...
+            // target's policy (rather than the inbound proxy's address).
+            return OrigDstAddr((self.client.local_addr.ip(), self.header.port).into());
+        }
+
+        // When the transport header provides an alternate target, the
+        // connection is a gateway connection. We check the _gateway
+        // address's_ policy (rather than the target address).
+        self.client.local_addr
+    }
+}
```
olix0r, line 505: This is a bit of a red flag to me... it indicates that we're using the `OrigDstAddr` type incorrectly.

* we should have a newtype that clearly indicates what we're using it for.
* then this impl sorts out whether the orig dst is used or something else.

Also I'm having trouble parsing this. 
```suggestion
impl svc::Param<policy::LookupAddr> for Discover {
    fn param(&self) -> policy::LookupAddr {
        // When the transport header provides an alternate target, the
        // connection is a gateway connection. We use the `OrigDstAddr` to lookup policies.
        if self.header.name.is_some() {
            return policy::LookupAddr(*self.client.local_addr);
        }
        
        // Otherwise, use the port override from the header so we. 
        policy::LookupAddr((self.client.local_addr.ip(), self.header.port).into());
    }
}
```

hawkw: Assuming we never want to turn a `LookupAddr` back into an `OrigDstAddr` (which...kind of defeats the purpose of the separate type, since a `LookupAddr` may not have been an actual original destination), this ends up being kind of a large change --- the address passed to `get_policy` ends up being propagated through a lot of different server policy code. I'm happy to change it, but it's gonna be a much bigger diff.

olix0r: OK let me take a quick look at the current state of things.

hawkw: in particular, it means that all the authz checks would be performed on a `LookupAddr`, which..._also_ smells weird to me. Although, I guess, at that point it could just be turned back into an un-newtyped `SocketAddr`...:woman_shrugging:

olix0r: OK we can fix this newtype a followup.

hawkw: well, i went ahead and did it (1252ca721e027836a516f5d11b5e6cef75fd4f2a), but am happy to back it out if you prefer?

hawkw: as an aside, it occurs to me that this logic could have been an `ExtractParam` impl and that the `Discover` target _could_ go away...not sure if that's nicer in this case though :woman_shrugging:

olix0r: eh, let's keep your change for a new branch and [3a2e376](https://github.com/linkerd/linkerd2-proxy/pull/2186/commits/3a2e376cf9bcffdbea17233f76eeb58ba3cfcc8e) + blanket impl?

After some poking it at, i don't know that we really want a `LookupAddr` type to be ~everywhere in the inbound stack. And it's not really correct for the inbound stack to rely on `OrigDstAddr` either, since that's not really accurate. What we want is:

* `OrigDstAddr` is used **only** to indicate the address returned by the socket option lookup.
* `ServerAddr` is used to refer to the address an inbound connection targets.
  * When we know `ServerAddr` is not the proxy's inbound server, we use `Remote<ServerAddr>`
  * If it's useful, we can describe direct connections targeting a `Local<ServerAddr>`
* The `LookupAddr` type is probably still use as a param marker to indicate when the param controls. It should _probably_ be a `LookupAddr(ServerAddr)`.

hawkw: mmkay, that makes sense — i'll make that change in a separate branch.

hawkw: okay, put everything back the way it was, and addressed the other review feedback. considered adding a `GetPolicy::into_service` as well, but figured we could add that if we actually need it. i'll make a separate PR for the address param changes.

linkerd/app/inbound/src/policy.rs
```diff
@@ -75,6 +80,28 @@ pub enum Routes {
...
+// === impl GetPolicy ===
+
+impl<S> GetPolicy for S
+where
+    S: tower::Service<OrigDstAddr, Response = AllowPolicy, Error = Error>
+        + Clone
+        + Send
+        + Sync
+        + Unpin
+        + 'static,
```
olix0r, line 92:
```suggestion
    S: tower::Service<OrigDstAddr, Response = AllowPolicy, Error = Error>,
    S: Clone + Send + Sync + Unpin + 'static,
```

linkerd/app/inbound/src/direct.rs
hawkw: pulled the `LookupAddr` change out into a separate PR  #2264

linkerd/idle-cache/src/lib.rs
```diff
@@ -134,6 +134,13 @@ where
         Some(cached)
     }
 
+    pub fn insert(&self, key: K, value: V) -> Cached<V>
+    where
+        V: Clone,
+    {
+        self.insert_with(key, move |_| value)
+    }
+
```
olix0r, line 143: Can you explain this part of the change a little bit more? I don't understand why the existing API doesn't work for our needs:

* `insert_with` is the same as `get_or_insert_with`, right? I guess it avoids the read lock? But it's it's effectively the same: if the value exists it returns the existing value. It gets.
* Do we need a `get_or_insert()` that calls `get_or_insert_with` internally?
* I'd generally expect `insert` to return an option holding the prior value in the occupied case...

linkerd/app/inbound/src/policy/store.rs
```diff
@@ -104,37 +106,58 @@ impl<S> Store<S> {
     }
 }
 
-impl<S> GetPolicy for Store<S>
+impl<S> svc::Service<OrigDstAddr> for Store<S>
```
hawkw, line 109: good call on just making this a service impl

olix0r (approved):

PR #2190, opened 2023-01-26
cache discovery based on the discovered target

hawkw:
Currently, caching of discovery watches is at the level of the
discovered `Service`s. These caches are keyed by the stack target that
constructed that `Service`. In some cases, tese stack targets contain
more data than just the destination address that a profile is discovered
for, such as in the gateway stack, where targets also include client
info. This means that we may perform duplicate discovery lookups for the
same destination address when other parts of the key differ, and these
separate lookups may be evicted separately as well.

Instead, we should have a single cache for all watches for a particular
type of discovery (e.g. profiles, client policies, server policies,
etc). These watches should only be subject to idle eviction from the
cache only when no service is using that watch.

This branch introduces a new `Stack::push_discovery_cache` function
which, when given a `Service` stack representing a discovery client,
returns a new `Service` which caches the discovery results from that
client service. The existing ServiceProfile client has been wrapped in a
discovery caching stack. The same stack functions may be useful for
future client and server policy watches as well.

### DIFF
--- a/linkerd/app/src/lib.rs
+++ b/linkerd/app/src/lib.rs
@@ -18,7 +18,8 @@ use linkerd_app_core::{
     control::ControlAddr,
     dns, drain,
     metrics::FmtMetrics,
-    svc::Param,
+    profiles,
+    svc::{self, Param},
     telemetry,
     transport::{listen::Bind, ClientAddr, Local, OrigDstAddr, Remote, ServerAddr},
     Error, ProxyRuntime,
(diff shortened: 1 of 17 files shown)


linkerd/app/src/lib.rs
```diff
@@ -142,6 +143,15 @@ impl Config {
             info_span!("dst").in_scope(|| dst.build(dns, metrics, identity.receiver().new_client()))
         }?;
 
+        let profiles = svc::stack(profiles::GetProfile::into_service(dst.profiles))
+            // TODO(eliza): these configs should not be inbound/outbound-specific,
+            // and should also have names that remotely reflect what they configure...
```
hawkw, line 148: i thought it was better to make the environment variable change in a separate PR, since it would presumably also require a proxy-injector change. but i could be convinced to do it here, if we'd prefer not to let this TODO make it to `main`?

hawkw:
Closed by #2195

PR #2194, opened 2023-01-27
 cache discovery based on the discovered destination address

hawkw:
Currently, caching of discovery watches is keyed by the stack target
that the watch was started for. In some cases, these stack targets
contain more data than just the destination address that a profile is
discovered for, such as in the gateway stack, where targets also include
client info. This means that we may perform duplicate discovery lookups
for the same destination address when other parts of the key differ, and
these separate lookups may be evicted separately as well.

Instead, we should have a single cache for all watches for a particular
type of discovery (e.g. profiles, client policies, server policies,
etc). These watches should only be subject to idle eviction from the
cache only when no service is using that watch.

This branch introduces a new `Stack::push_discovery_cache` function
which, when given a `Service` stack representing a discovery client and
layered onto an inner stack that consumes the discovery result, produces
a `NewService` which caches the discovery results from the discovery
client service. Each service constructed by the discovery cache holds
the `Cached` key for the discovery watch for its lifetime, ensuring that
cached discovery watches are only subject to eviction when no services
are using them. The existing ServiceProfile client has been wrapped in a
discovery caching stack. The same stack functions may be useful for
future client and server policy watches as well.

Closes #2190


hawkw:
there appears to be some kind of issue around failfast, trying to figure that out

hawkw:
The issue appears to be due to introducing a buffer around the individual resolution services, the `SwitchReady` that switches over to the default profile when the inbound HTTP stack's profile client isn't ready no longer triggers, because the buffer is advertising readiness when the inner service hasn't actually been created yet. Then, the inner service starts a resolution, which never becomes ready, and the requests stuck in the buffer time out from failfast. The gate outside of the buffer would advertise that it isn't ready, but the requests in the test are already stuck in the buffer.

Hmmm...

hawkw:
It would be really nice if we didn't need a buffer there...

hawkw:
well, 21e9e06448a313b606045eb8c1a06912653f0e60 fixes the tests by making `ThunkCloneResponse` implement `Clone` so that the buffer around it can be removed, and `SwitchReady` still works for the disco stack. I'm not sure if I _love_ the solution in that commit though...

linkerd/stack/src/spawn_clone_response.rs
```diff
@@ -0,0 +1,130 @@
...
+    sync::Arc,
+    task::{Context, Poll},
+};
+use tokio::sync::watch;
+
+/// A `NewService<T>` that produces a [`ThunkCloneResponse`] over an inner
+/// `Service<T>`, calling that service a single time in a background task and
+/// producing `Service<()>`s that clone the response of that call.
+#[derive(Debug, Clone)]
+pub struct SpawnThunkCloneResponse<S> {
```
hawkw, line 17: another name alternative could be `SpawnMemoize` (`NewService`) and `Memoized` (`Service`)?

linkerd/app/core/src/svc.rs
```diff
@@ -233,6 +235,47 @@ impl<S> Stack<S> {
...
+        D: Service<K> + Clone + Send + Sync + 'static,
+        D::Error: Into<Error> + Send + 'static,
+        D::Future: Send + 'static,
+        D::Response: Clone + Send + Sync + 'static,
+        S: NewService<(D::Response, T)> + Clone,
+    {
+        let discover = stack(discover)
+            .push(stack::SpawnThunkCloneResponse::layer())
+            .check_new_service::<K, ()>();
+        self.push(crate::disco_cache::NewDiscoveryCache::layer(discover, idle))
```
olix0r, line 275: Can we use `NewDiscoveryCache` without `SpawnThunkCloseResponse`? If not, it's probably simpler to move it into the disco_cache module. Specifically it eliminates the awful type mess.

linkerd/service-profiles/src/default.rs
```diff
@@ -9,7 +9,7 @@ use tracing::debug;
 pub struct RecoverDefault<S>(S);
 
 impl<S> RecoverDefault<S> {
-    pub(crate) fn new(inner: S) -> Self {
+    pub fn new(inner: S) -> Self {
```
olix0r, line 12: This shouldn't be necessary.

linkerd/app/gateway/src/lib.rs
```diff
@@ -81,7 +81,7 @@ where
 {
     let inbound_config = inbound.config().clone();
     let local_id = identity::LocalId(inbound.identity().name().clone());
-
+    let profiles = profiles::RecoverDefault::new(profiles.into_service());
```
olix0r, line 84: In the other branch I've moved this stuff up to where the profile client is built.

linkerd/app/core/src/svc.rs
```diff
@@ -233,6 +235,47 @@ impl<S> Stack<S> {
...
+        >,
+    >
+    where
+        T: Param<K> + Clone,
+        K: Clone + fmt::Debug + Eq + Hash + Send + Sync + 'static,
+        D: Service<K> + Clone + Send + Sync + 'static,
+        D::Error: Into<Error> + Send + 'static,
+        D::Future: Send + 'static,
+        D::Response: Clone + Send + Sync + 'static,
+        S: NewService<(D::Response, T)> + Clone,
```
olix0r, line 270: I think in general we want a double-new layer here. We can always lift a double to a single but we can't go the other way.

olix0r (commented):
This looks really good!

I worked on something very similar in parallel.

* https://github.com/linkerd/linkerd2-proxy/pull/2196 proposes adding a queue variant that doesn't include failfast/timeouts. This is basically in place of adding a spawning layer.
* https://github.com/linkerd/linkerd2-proxy/pull/2195/ is a `disco_cache` impl. It uses `FutureService` instead of hand-rolling. It's very similar to what you did here, though, otherwise.

Modulo some small organizational differences, I think the main question between our branches is whether we should be using a buffer or a watch. I have a small bias towards the buffer, since it's more of a known quantity, but either approach could be workable.

linkerd/stack/src/spawn_clone_response.rs
```diff
@@ -0,0 +1,130 @@
...
+    S::Future: Send,
+    S::Response: Clone + Send + Sync + 'static,
+{
+    type Service = ThunkCloneResponse<S::Response>;
+
+    fn new_service(&self, target: T) -> Self::Service {
+        let (tx, rx) = watch::channel(None);
+        let service = self.service.clone();
+        tokio::spawn(async move {
+            let rsp = service.oneshot(target).await.map_err(|error| {
```
olix0r, line 51: This needs to select against the tx closing...

linkerd/stack/src/spawn_clone_response.rs
```diff
@@ -0,0 +1,130 @@
...
+
+            if let Some(ref mut waiting) = self.waiting {
+                futures::ready!(waiting.poll_unpin(cx));
+            } else {
+                let mut rx = self.rx.clone();
+                self.waiting = Some(Box::pin(async move {
+                    let _ = rx.changed().await;
+                }))
+            }
+        }
```
olix0r, line 89: This will only ever go from pending to ready once, correct? If so, it probably fits better as a `FutureService`? Oh, except that can't be clone...

olix0r:
closing in favor of #2195

PR #2195, opened 2023-01-28
Cache discovery results independently of proxy stacks 

olix0r:
The current discovery module resolves profiles from the control plane
and injects the response into the inner stack. So, in order to cache
this result, the entire stack must be cached. This prevents us from
interacting with discovery in any stack that includes
higher-cardinality, peer-specific metadata. This limitation is
unnecessary and cumbersome to work around.

This change adds an `app::core::NewCachedDiscover` module that
internally manages resolutions independently of the proxy stack.

This change does **not** remove any other existing caches, though it's
expected that several of these caches are no longer necessary.

Depends on #2196


linkerd/idle-cache/src/lib.rs
```diff
@@ -329,6 +339,15 @@ impl<V> Cached<V> {
     }
 }
 
+impl<T> Cached<T> {
+    pub fn clone_with<U>(&self, inner: U) -> Cached<U> {
```
hawkw, line 343: maybe worth a comment explaining that this creates a new `Cached<U>` that's tied to the eviction lifetime of `self`?

linkerd/app/core/src/disco_cache.rs
```diff
@@ -0,0 +1,155 @@
...
+{
+    // NewService<K, Service<(), D::Response>>
+    cache: NewIdleCached<K, NewQueueThunk<NewDiscoverThunk<D>>>,
+
+    // NewService<D::Response>
+    inner: N,
+}
+
+/// The future that drives discovery to build an new inner service wrapped
+/// in the [`Cached`] decorator from the discovery lookup.
```
hawkw, line 28: nit, maybe worth making this slightly more explicit:
```suggestion
/// in the [`Cached`] decorator from the discovery lookup, so that the lookup is not
/// considered idle while services built from it exist.
```
or something?

hawkw (approved):
this looks good to me --- we talked through most of the details offline, so this isn't a particularly lengthy review. 

i'm still a little bit iffy of having every connection go through the queue even once the discovery watch has started, but i don't see any major problems with it. clever use of existing primitives here!

PR #2196, opened 2023-01-28
Add a queue that does not enforce failfast timeouts

olix0r:
While most of our data-path queues need some sort of failfast behavior, the queues used to request & synchronize discovery information do not need to enforce timeouts locally.

This change updates the `stack::queue` module to support two variants (via type aliases and builders): _with_ and _without_ timeouts.

This change introduces new param types for capacity and timeout configuration. The `QueueConfig` type is moved back into the `app::core::config` module. Furthermore, all config field names are updated to replace "buffer" with "queue".


linkerd/stack/src/queue.rs
```diff
@@ -1,93 +1,153 @@
...
+pub type QueueWithTimeout<Req, Rsp, E = Error> =
+    failfast::Gate<Buffer<BoxService<Req, Rsp, E>, Req>>;
+
+/// A marker type to indicate a queue does not apply a timeout.
+#[derive(Debug)]
+pub enum WithoutTimeout {}
+
+pub type NewQueueWithoutTimeout<X, Req, N> = NewQueue<WithoutTimeout, X, Req, N>;
+
+pub type QueueWithoutTimeout<Req, Rsp, E = Error> = Buffer<BoxService<Req, Rsp, E>, Req>;
```
hawkw, line 40: what's the advantage of using the marker types versus just having separate `NewQueue`/`NewQueueWithTimeout` structs?

if `NewQueueWithTimeout` and `NewQueueWithoutTimeout` were just separate structs rather than using the marker, we could invoke the constructors as `NewQueueWithTimeout::layer`/`NewQueue::layer`...

linkerd/stack/src/queue.rs
```diff
@@ -1,93 +1,153 @@
...
+        Self {
+            inner,
+            extract,
+            _req: PhantomData,
+        }
+    }
+
     /// Returns a [`Layer`] that constructs new [`failfast::Gate`]d [`Buffer`]s
     /// using an `X`-typed [`ExtractParam`] implementation to extract
     /// [`QueueConfig`] from a `T`-typed target.
```
hawkw, line 96: this comment is no longer correct, as this layer no longer constructs buffers with the failfast gate.

linkerd/stack/src/lib.rs
```diff
@@ -61,6 +64,11 @@ pub use tower::{
     Service,
 };
 
+pub type BoxFutureService<S, E = linkerd_error::Error> = FutureService<
+    std::pin::Pin<Box<dyn std::future::Future<Output = Result<S, E>> + Send + 'static>>,
+    S,
+>;
```
hawkw, line 70: nit: did this sneak in from another PR accidentally? i don't see where it's used here...

hawkw (approved):
overall, this looks good to me, and i like the use of separate params for `Capacity` and `Timeout`s. i'm a little unsure if the use of marker types for `NewQueue` actually makes this code any simpler, or if we'd be better off with two separate `NewService` impls, but i don't mind the current approach! :+1:

linkerd/stack/src/lib.rs
olix0r: yeah. and we're not even using it in the other branches anymore. I'll kill it.

linkerd/stack/src/queue.rs
olix0r: yeah agreed. I've left `NewQueue` to _include_ the timeout, since that's generally the behavior we want. We should be explicit when that safety measure isn't in place.

olix0r:
I agree that marker types aren't really worth it here.

PR #2197, opened 2023-01-28
Rename MakeThunk to NewThunk

olix0r:
We don't use `MakeThunk` as a `MakeService` anywhere, so it's clearer to call it `NewThunk`.

This change also adds a `ThunkClone` service that clones a single response for each call.


olix0r:
trivial. merging.

PR #2198, opened 2023-01-29
gateway: Unify discovery for HTTP & opaque stacks

olix0r:
In its normal sidecar configuration, the outbound stack does discovery for each target ip:port. The result of this discovery will influences protocol detection, etc.

In the gateway configuration, however, discovery is pushed into each distinct stack so that lookups for HTTP services are deferred until we've terminated the connection and wired it into the per-request outbound stack.

This difference was primarily motivated by a limitation of the way that discovery information is cached: because the outermost target includes per-client metadata, we can't cache a discovery stack based on that target type. #2195 removes this limitation so that we can update the gateway stack to look more like a typical outbound stack (which will allow for easier code reuse as new discovery types are introduced).

This change:

1. Splits the gateway stack into smaller functions for readability.
2. Extracts discovery from the HTTP and opaque stacks.
3. Renames gateway types for clarity.


linkerd/app/gateway/src/lib.rs
```diff
@@ -77,9 +84,158 @@ where
...
+    O: Clone + Send + Sync + Unpin + 'static,
+    O: svc::MakeConnection<outbound::tcp::Connect, Metadata = Local<ClientAddr>, Error = io::Error>,
+    O::Connection: Send + Unpin,
+    O::Future: Send + Unpin + 'static,
+    R: Clone + Send + Sync + Unpin + 'static,
+    R: Resolve<outbound::tcp::Concrete, Endpoint = Metadata, Error = Error>,
+    <R as Resolve<outbound::tcp::Concrete>>::Resolution: Send,
+    <R as Resolve<outbound::tcp::Concrete>>::Future: Send + Unpin,
+{
     // For each gatewayed connection that is *not* HTTP, use the target from the
```
hawkw, line 239: nit, TIOLI: should this comment maybe be on the `new_opaque` function now?

linkerd/app/gateway/src/lib.rs
```diff
@@ -77,9 +84,158 @@ where
...
+        let stk = new_http(local_id, outbound.clone(), resolve);
+        inbound
+            .clone()
+            .with_stack(
+                svc::stack(stk)
+                    .push_on_service(svc::LoadShed::layer())
+                    .lift_new()
+                    .push(svc::NewOneshotRoute::layer_via(
+                        |(_permit, http): &(_, HttpWithClient)| ByRequestVersion(http.http.clone()),
+                    ))
```
hawkw, line 107: neat use of extractparam here!

linkerd/app/gateway/src/lib.rs
```diff
@@ -23,30 +22,37 @@ use linkerd_app_inbound::{
...
-    policy: policy::AllowPolicy,
+    inbound_policy: policy::AllowPolicy,
 }
 
 #[derive(Clone, Debug)]
-struct RouteHttp<T>(T);
-
-#[derive(Clone, Debug, PartialEq, Eq, Hash)]
-struct HttpTarget {
+struct Http {
```
hawkw, line 39: it could _maybe_ be worth adding comments to these structs explaining their relationship, for future readers? Something like "`HttpWithClient` is the target for the HTTP server stack, and then the client info is removed from the target and added to a request extension, and services are cached based on the `Http` target"?

hawkw (approved):
this looks good to me!

linkerd/app/gateway/src/lib.rs
```diff
@@ -28,20 +28,27 @@ pub struct Config {
...
+struct InboundHttp {
     client: ClientInfo,
     inbound_policy: policy::AllowPolicy,
+    outbound: OutboundHttp,
 }
 
+/// A target type describing outbound HTTP traffic.
 #[derive(Clone, Debug)]
-struct Http {
+struct OutboundHttp {
```
hawkw, line 43: :+1: on the naming change here

hawkw (approved):

PR #2199, opened 2023-01-30
core: Rename Stack utilities for clarity

olix0r:
`Stack::push_request_filter` may apply to either `Service` or `NewService` stacks. The use of "request" makes it sound like it only applies to services. This change renames it to `Stack::push_filter`.

`Stack::push_idle_cache` is renamed to `Stack::push_new_idle_cached` to reflect that the stack is a `NewService`.


hawkw (approved):
thanks for pulling this out!

PR #2204, opened 2023-02-01
inbound: remove default policies

hawkw:
## Motivation

Currently, the inbound proxy has a notion of default ServerPolicies,
which are configured through an environment variable which selects the
default policy, and other environment variables that set properties for
lists of ports (requiring TLS, requiring mesh identity, and skipping
protocol detection). These are used in two cases:

 - When the proxy is configured without a policy controller address. In
   this case, the default policy is always used. In practice, this only
   occurs in the proxy's own integration tests.
 - When the proxy *is* configured with a policy controller, default
   policies are used when a ServerPolicy has yet to be resolved from the
   control plane for a specific port.

The default ServerPolicies, and the logic for determining the default
policy for a given port based on multiple environment variables,
introduces complexity into the proxy that make its behavior harder to
reason about. In addition, the proxy injector must currently inject the
proxy with an environment variable containing every opaque port, which
can cause issues when large port ranges are marked as opaque. See 
linkerd/linkerd2#9803 for details.

## Solution

In order to simplify the proxy's configuration surface area, this branch
removes the default policy configuration.

PR #2186 changed the inbound proxy to wait for policies to be resolved
before handling connections on those ports, rather than using the
default policy until a result is returned by the policy controller, so
the default policy can be removed in that case.

In addition, this branch removes the mode of operation where no policy
controller address is configured, removing the other case where default
policies are used. The policy-controller-less mode is only used for the
proxy's integration tests, and maintaining it introduces additional
complexity. Since the policy controller is now required in all cases, it
was necessary to add a mock policy controller implementation for use in
integration tests which previously relied on default policies.

The `LINKERD2_PROXY_INBOUND_PORTS` environment variable, which
configures the set of ports to eagerly look up policies for, and which
are never evicted from the cache once looked up, is still honored as it
was previously.

## Implications

Critically, this change now means that the policy controller is now
*required* for the inbound proxy to operate. The proxy will not evict
the pre-configured list of inbound ports from its cache, so connections
to those ports *will* still function if communication to the policy
controller is interrupted. However, ports not in
`LINKERD2_PROXY_INBOUND_PORTS` may be evicted from the cache if they
exceed the idle timeout, and subsequent connections on those ports will
fail if the policy controller cannot be reached. Furthermore, the proxy
will need to communicate with the policy controller on startup to
populate the pre-configured list of inbound ports.

This PR will allow changing the proxy injector to avoid injecting
proxies with a list of every opaque port, since opaque ports are now
always configured by the policy controller, which can handle port
ranges, and the proxy no longer needs to be aware of every individual
port in a range until it actually receives a connection on that port.

Depends on #2186
Closes #2076
Closes #2079
Closes #2080


hawkw:
Rebased now that #2186 has merged

olix0r (approved):

PR #2208, opened 2023-02-02
outbound: Make stacks generic over target types

olix0r:
We want to change the shape of the outbound stack--removing
`switch_logical` so that forwarding/balancing decisions are made in
concrete stacks. We also want to move away from using profiles
exclusively, adding new discovery types. This is complicated by our use
of fixed target types because it's difficult to make incremental
changes. Furthermore, a stack's requirements are not clearly documented.

This change addresses this as follows:

* The logical, concrete, and endpoint stacks are generic over their
  input types. Each stack enumerates required parameters from these
  targets.
* Each of these stacks declare module-local types that are passed to
  inner services. These types must expose params to satisfy all of the
  requirements of inner stacks.
* `tcp::concrete` and `tcp::logical` have been renamed to
  `opaque::concrete` and `opaque::logical` to reflect that they are an
  alternative to the http stack. The HTTP stack continues to incorporate
  the `tcp::endpoint` stack. The stack helpers have been renamed
  accordingly.

It is intentional that some of the new target types are effectively identical
to those in other stacks. For now, decoupling is a higher priority than
code reuse.


linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -19,6 +28,22 @@ pub struct ConcreteError {
...
+pub struct NewEndpoint<N> {
+    inner: N,
+    inbound_ips: IpSet,
+}
+
+pub type IpSet = Arc<AHashSet<IpAddr>>;
+
 impl<N> Outbound<N> {
     /// Builds a [`svc::NewService`] stack that builds buffered HTTP load
     /// balancer services for [`Concrete`] targets.
```
hawkw, line 49: comment should probably be updated since this no longer operates specifically on the `Concrete` type

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -29,12 +54,12 @@ impl<N> Outbound<N> {
     /// services.
     //
     // TODO(ver) make the outer target type generic/parameterized.
```
hawkw, line 56: looks like this TODO has been addressed :)
```suggestion
```

linkerd/app/outbound/src/ingress.rs
```diff
@@ -267,8 +242,126 @@ impl Outbound<svc::ArcNewHttp<http::Endpoint>> {
...
+impl svc::Param<http::client::Settings> for Http<OrigDstAddr> {
+    fn param(&self) -> http::client::Settings {
+        match self.param() {
+            http::Version::H2 => http::client::Settings::H2,
+            http::Version::Http1 => http::client::Settings::Http1,
+        }
+    }
+}
+
+// TODO(ver) move this into the endpoint stack?
```
hawkw, line 333: is this comment still relevant here? looks like it was pasted from the earlier code?

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -77,31 +109,241 @@ impl<N> Outbound<N> {
...
+            http::Version::H2 => client::Settings::H2,
+            http::Version::Http1 => match self.metadata.protocol_hint() {
+                ProtocolHint::Unknown => client::Settings::Http1,
+                ProtocolHint::Http2 => client::Settings::OrigProtoUpgrade,
+            },
+        }
+    }
+}
+
+// TODO(ver) move this into the endpoint stack?
```
hawkw, line 281: is this comment still relevant here? looks like it was pasted from the earlier code?

linkerd/app/outbound/src/opaque/concrete.rs
```diff
@@ -0,0 +1,281 @@
...
+impl<C> Outbound<C> {
+    /// Builds a [`svc::NewService`] stack that builds buffered opaque TCP load
+    /// balancer services for [`Concrete`] targets.
+    ///
+    /// When a balancer has no available inner services, it goes into
+    /// 'failfast'. While in failfast, buffered requests are failed and the
+    /// service becomes unavailable so callers may choose alternate concrete
+    /// services.
+    //
+    // TODO(ver) make the outer target type generic/parameterized.
```
hawkw, line 59: i think we've done that in this PR :)
```suggestion
```

linkerd/app/outbound/src/opaque/concrete.rs
```diff
@@ -0,0 +1,281 @@
...
+    inbound_ips: IpSet,
+}
+
+pub type IpSet = Arc<AHashSet<IpAddr>>;
+
+// === impl Outbound ===
+
+impl<C> Outbound<C> {
+    /// Builds a [`svc::NewService`] stack that builds buffered opaque TCP load
+    /// balancer services for [`Concrete`] targets.
```
hawkw, line 52: comment could be updated since this is now generic?

linkerd/app/outbound/src/opaque/forward.rs
```diff
@@ -0,0 +1,108 @@
...
+where
+    T: svc::Param<Remote<ServerAddr>>,
+{
+    fn from((target, source): (&T, Error)) -> Self {
+        Self {
+            addr: target.param(),
+            source,
+        }
+    }
+}
```
hawkw, line 108: nit, not important: normally i would expect the test module to be at the very end of the file, with impls for types defined in the main module before the tests?

hawkw (approved):
overall, this is pretty straightforward, and the change looks good to me!

i commented on a few minor nits, mostly comments that look like they ought to be updated. other than that, this seems good!

olix0r:
superseded by https://github.com/linkerd/linkerd2-proxy/pull/2210

PR #2209, opened 2023-02-02
handle `Opaque` protocol hints on endpoints

hawkw:
Currently, when the outbound proxy makes a direct connection prefixed
with a `TransportHeader` in order to send HTTP traffic, it will always
send a `SessionProtocol` hint with the HTTP version as part of the
header. This instructs the inbound proxy to use that protocol, even if
the target port has a ServerPolicy that marks that port as opaque, which
can result in incorrect handling of that connection. See
linkerd/linkerd2#9888 for details.

In order to prevent this, linkerd/linkerd2-proxy-api#197 adds a new
`ProtocolHint` value to the protobuf endpoint metadata message. This
will allow the Destination controller to explicitly indicate to the
outbound proxy that a given endpoint is known to handle all connections
to a port as an opaque TCP stream, and that the proxy should not perform
a protocol upgrade or send a `SessionProtocol` in the transport header.
This branch updates the proxy to handle this new hint value, and adds
tests that the outbound proxy behaves as expected.

Once the Destination controller is updated to send this `ProtocolHint`
for ports that are marked as opaque on the inbound side, this will fix
linkerd/linkerd2#9888.

This branch is currently a draft because the `linkerd2-proxy-api`
changes have not been published yet. Once a new version of the proxy API
is published, we'll update this branch to depend on that, and this can
be merged.


linkerd/app/integration/src/tests/direct.rs
```diff
@@ -0,0 +1,134 @@
...
+            }
+            .into_update(srv_addr),
+        );
+        let (proxy, profile) = mk_outbound(srv_addr, ctrl, out_identity).await;
+        (proxy, profile, dst)
+    };
+
+    let client = client::http1(outbound.outbound, "opaque.test.svc.cluster.local");
+
+    assert_eq!(client.get("/").await, "hello");
```
mateiidavid, line 93: do you think it's worth adding an assert here to check the status code? I'm thinking it might give some visibility if the should ever fail in the future? Then again, I think what's in here is already good 🤷🏻

mateiidavid (approved):
Nice, it looks good to me.

linkerd/app/integration/src/tests/direct.rs
hawkw: The `client.get(...)` method panics if the status isn't 200 OK. We would use `client.request(...)` to not fail on other status codes. So, this already tests that it's OK. :)

linkerd/app/integration/src/tests/direct.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+/// Reproduces linkerd/linkerd2#9888. A proxy receives HTTP traffic direct
+/// traffic with a transport header, and the port is also in the
```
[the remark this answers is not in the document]
hawkw: ah, yup, that's a typo, good catch!

Cargo.toml
```diff
@@ -83,4 +83,4 @@ lto = true
 [patch.crates-io]
 webpki = { git = "https://github.com/linkerd/webpki", branch = "cert-dns-names-0.22" }
 boring = { git = "https://github.com/cloudflare/boring" }
-tokio-boring = { git = "https://github.com/cloudflare/boring" }
+tokio-boring = { git = "https://github.com/cloudflare/boring" }
```
olix0r, line 86: What setting do you have that does this? Can you change it for this repo?

linkerd/proxy/api-resolve/src/metadata.rs
```diff
@@ -31,6 +31,9 @@ pub enum ProtocolHint {
     Unknown,
     /// The destination can receive HTTP2 messages.
     Http2,
+    /// The destination will handle all connections as opaque TCP, so do not
+    /// send a session protocol in a tagged transport connection.
```
olix0r, line 35: This is a downstream implication of this hint 

```suggestion
    /// The destination will handle traffic as opaque, regardless of
    /// the local proxy's handling of the traffic.
```

The details of tagged transport are out of scope here. We should be clear about what the setting means, not enumerate all of the behaviors that might rely on this setting.

olix0r (commented):
This is going to conflict with the other changes i have in flight so I'm going to have to be a little careful about merging this.

How can we make the change smaller? For instance, there's a lot of test infra change in this PR that i haven't really parsed.

linkerd/app/outbound/src/tcp/opaque_transport.rs
```diff
@@ -162,6 +167,32 @@ mod test {
         )
     }
 
+    fn http_ep(metadata: Metadata, version: http::Version) -> crate::http::Connect {
+        let e = crate::http::Endpoint::from((version, ep(metadata)));
+        crate::http::Connect::from((version, e))
+    }
```
olix0r, line 173: This stuff is going to conflict with other work, especially. Can you decouple the tests from the specific target types used in the stack?

hawkw:
> For instance, there's a lot of test infra change in this PR that i haven't really parsed.

We can remove the integration test changes if we don't care about having integration tests for this case; I wrote that initially while trying to reproduce the issue. If it's more desirable to have a smaller branch, I can back that out for now and we can consider merging those changes separately (or not at all)?

Cargo.toml
hawkw: Huh, I dunno what happened here --- my VS Code *claims* it's configured not to mess with trailing whitespace...weird. Will fix.

linkerd/app/outbound/src/tcp/opaque_transport.rs
hawkw: Okay, 755767627b7b3894a96b0c29b654d3c014f2c4ea changes these tests so that they no longer use the actual target types --- is that sufficient?

IMO, this is somewhat unfortunate, as the target types' `Param` impls were part of the code being tested here, and now, we need to ensure that the targets used in the tests actually behave the same as the real targets. But, hopefully that reduces the conflict with other changes?

olix0r:
> We can remove the integration test changes if we don't care about having integration tests for this case

My point was kind of the opposite: there are probably some parts of this that are relatively easy/uncontroversial to merge that don't actually touch the outbound stack. The outbound stack changes need to be considered in the scope of other changes, while much of the other changes (proxy-api bump, test infra stuff, etc) do not.

linkerd/app/outbound/src/tcp/opaque_transport.rs
olix0r: Look at what's coming in 
https://github.com/linkerd/linkerd2-proxy/pull/2210/files#diff-397930556d70c6560dae17b17cf26153f2b1398664e6e7b46112d5135a1ec4bb -- we don't have any reason to use `Metadata` at all.

> IMO, this is somewhat unfortunate, as the target types' Param impls were part of the code being tested here,

This decoupling is intended to force this conversation. Why are these tests coupled to the caller context? Do we need tests in the caller module(s) that exercise how they set their parameters? Or does this indicate that the params are not factored properly so that the inputs to this stack are too fine grain?

hawkw:
> > We can remove the integration test changes if we don't care about having integration tests for this case
> 
> My point was kind of the opposite: there are probably some parts of this that are relatively easy/uncontroversial to merge that don't actually touch the outbound stack. The outbound stack changes need to be considered in the scope of other changes, while much of the other changes (proxy-api bump, test infra stuff, etc) do not.

Ooooh, gotcha. I'll pull that stuff out, then.

olix0r:
https://github.com/linkerd/linkerd2-proxy/pull/2226/ includes some of those test changes

hawkw:
Okay, I've opened a new PR for this change (#2237), which is updated to track the latest changes on `main`. I'm going to close this branch in favor of that one.

PR #2210, opened 2023-02-03
outbound: Refactor stack target types

olix0r:
To support a new configuration API, we want to update the outbound stack
to compose the following stacks:

* Routing discovery;
* Protocol switching;
* Protocol-specific logical routing;
* Concrete forwarding / load balancing;
* Endpoint dispatching.

To do this, we update the existing stacks so that:

* each stack accepts generic input targets (with enumerated `Param`
  requirements); and
* each module defines all necessary internal target types, including
  those passed to inner stacks.

This decouples stack configuration from the specifics of discovery:
tests can satisfy stack parameters without having to stub out all of the
discovery types and we are setup to change discovery implementation
details moving forward.

This requires updating the gateway and ingress stacks in addition to the
(default) sidecar stack.

Generally, we prefer duplicating similar target types to coupling
modules. In followup changes, we can introduce macros to reduce `Param`
boilerplate if necessary.

Changes include:

* A new `Gateway` builder type (analogous to `Inbound` and `Outbound`);
* A new `Gateway::server` stack handles discovery and protocol
  switching;
* Service cacheing is performed for the http and opaq stacks (to
  preserve load balancer state, etc) using only the information needed
  for those stacks;
* Concrete stacks provide the only data-plane queue. This enables
  outer stacks to be cloneable. This also means that the concrete stacks
  are the only stacks with stack metrics;
* The `switch_logical` module is removed;
* The `http::detect` module is moved into the `protocol` stack, which
  now supports discovery-hinted protocols to bypass detection.
* The `logical` and `endpoint` modules are removed in favor of
  protocol-specific modules.


linkerd/app/outbound/src/ingress.rs
```diff
@@ -176,49 +97,24 @@ impl<S> Outbound<S> {
...
                 // are not.
-                let new_http = ingress_override
-                    .push_switch(
-                        |Http { target, version }: Http<Target>| -> Result<_, Infallible> {
-                            Ok(match target {
-                                Target::Override(target) => svc::Either::A(Http { target, version }),
-                                Target::Forward(target) => svc::Either::B(Http { target, version }),
-                            })
-                        },
-                        http_endpoint.into_inner(),
```
olix0r, line 187: TODO: come back to this part of the change...

linkerd/app/outbound/src/lib.rs
```diff
@@ -137,10 +142,7 @@ impl<S> Outbound<S> {
         self.stack.into_inner()
     }
 
-    fn no_tls_reason(&self) -> tls::NoClientTls {
-        tls::NoClientTls::NotProvidedByServiceDiscovery
-    }
-
+    // Wraps the inner `S`-typed stack in the given `L`-typed [`Layer`].
```
hawkw, line 145: tiny nit: 
```suggestion
    /// Wraps the inner `S`-typed stack in the given `L`-typed [`Layer`].
```

linkerd/app/outbound/src/opaq/concrete.rs
```diff
@@ -1,109 +1,301 @@
...
     /// Builds a [`svc::NewService`] stack that builds buffered opaque TCP load
-    /// balancer services for `Concrete` targets.
+    /// balancer services for `T`-typed concrete targets.
     ///
     /// When a balancer has no available inner services, it goes into
     /// 'failfast'. While in failfast, buffered requests are failed and the
     /// service becomes unavailable so callers may choose alternate concrete
     /// services.
     //
     // TODO(ver) make the outer target type generic/parameterized.
```
hawkw, line 63: this comment can be removed now since this PR does that :)
```suggestion
```

linkerd/app/outbound/src/protocol.rs
```diff
@@ -0,0 +1,142 @@
...
+                            Protocol::Http2 => Ok(svc::Either::A(svc::Either::A(Http {
+                                version: http::Version::H2,
+                                parent,
+                            }))),
+                            Protocol::Opaque => Ok(svc::Either::A(svc::Either::B(parent))),
+                            Protocol::Detect => Ok(svc::Either::B(parent)),
+                        }
+                    },
+                    detect.into_inner(),
+                )
```
hawkw, line 105: the nested switch here (and corresponding two levels of nested of `Either`s) feels a bit complex, it might be nice if there was a comment explaining what the inner and outer `Either`s control at the top of the match or something?

linkerd/app/outbound/src/protocol.rs
```diff
@@ -0,0 +1,142 @@
...
+        let http = self.with_stack(http).map_stack(|config, rt, stk| {
+            stk.push_on_service(http::BoxRequest::layer())
+                .unlift_new()
+                .push(http::NewServeHttp::layer(
+                    config.proxy.server.h2_settings,
+                    rt.drain.clone(),
+                ))
+        });
+
+        // The inner stacks are built for each connection and not cached.
```
hawkw, line 66: tiny nit: maybe this comment should move to the top of the function, because it also applies to `opaq` and `http`?

linkerd/app/outbound/src/opaq/logical.rs
```diff
@@ -52,20 +75,25 @@ impl<N> Outbound<N> {
     /// [`svc::stack::LoadShedError`].
     ///
     // TODO(ver) make the outer target type generic/parameterized.
```
hawkw, line 77: comment can be removed since this branch does that :)
```suggestion
```

linkerd/app/outbound/src/opaq/logical.rs
```diff
@@ -1,43 +1,66 @@
...
 
-#[cfg(tests)]
+#[cfg(test)]
 mod tests;
 
+#[derive(Clone, Debug)]
+pub enum Logical {
+    Route(NameAddr, profiles::Receiver),
+    Forward(Remote<ServerAddr>, Metadata),
+}
```
hawkw, line 22: naming bikeshed (significance: silly, feel free to ignore): are both variants of this actually "logical" addresses? the `Forward` case represents a concrete address, and it's really only the `Route` case that's a logical address, so it's really the `Routable` target type that's a logical target. but...i dunno what we would call this type instead... :woman_shrugging:

linkerd/app/outbound/src/opaq/logical.rs
```diff
@@ -84,31 +112,64 @@ impl<N> Outbound<N> {
                 // Lazily cache a service for each `RouteParams`
                 // returned from the `SelectRoute` impl.
                 .push_on_service(route)
-                .push(svc::NewOneshotRoute::<Params, _, _>::layer_cached());
+                .push(svc::NewOneshotRoute::<Params<T>, _, _>::layer_cached());
 
             // For each `Logical` target, watch its `Profile`, maintaining a
             // cache of all concrete services used by the router.
             concrete
```
hawkw, line 119: this comment should maybe be updated since we're not doing this for each `Logical` target, but only for the `Routable` ones, and the stack we build here now contains a switch?

linkerd/app/outbound/src/opaq/concrete.rs
hawkw, line 63: i feel like this doc comment should maybe be updated to also indicate that this is the stack that determines whether to build a TCP load-balancer stack or forward directly to an endpoint?

linkerd/app/outbound/src/http.rs
```diff
@@ -1,164 +1,142 @@
...
+    {
+        self.push_tcp_endpoint()
+            .push_http_endpoint()
+            .push_http_concrete(resolve)
+            .push_http_logical()
+            .push_http_server()
+            .map_stack(move |config, _, stk| {
+                stk.push_new_idle_cached(config.discovery_idle_timeout)
+                    // Use a dedicated target type to configure parameters for the
+                    // HTTP stack. It also helps narrow the cache key.
```
hawkw, line 80: nit, maybe:
```suggestion
                    // Use a dedicated target type to configure parameters for the
                    // HTTP stack. It also helps narrow the cache key to just the
                    // logical target and HTTP version, discarding any other target
                    // information.
```

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -1,104 +1,368 @@
...
+
 impl<N> Outbound<N> {
     /// Builds a [`svc::NewService`] stack that builds buffered HTTP load
-    /// balancer services for [`Concrete`] targets.
+    /// balancer services for `T`-typed concrete targets.
     ///
     /// When a balancer has no available inner services, it goes into
     /// 'failfast'. While in failfast, buffered requests are failed and the
     /// service becomes unavailable so callers may choose alternate concrete
     /// services.
```
hawkw, line 63: should this comment maybe be updated to indicate that this stack also now determines whether to build a load balancer or forward directly to an endpoint?

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -1,104 +1,368 @@
...
                     Future = impl Send,
                 > + Clone,
         >,
     >
     where
-        N: svc::NewService<Endpoint, Service = NSvc> + Clone + Send + Sync + 'static,
+        // Concrete target type.
+        T: svc::Param<Dispatch>,
+        // T: svc::Param<svc::queue::Capacity>,
+        // T: svc::Param<svc::queue::Timeout>,
```
hawkw, line 82: these should probably be removed, right?
```suggestion
```

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -1,104 +1,368 @@
...
+                        tracing::trace!(%addr, ?metadata, ?target, "Resolved endpoint");
+                        let is_local = inbound_ips_.contains(&addr.ip());
+                        Endpoint {
+                            addr: Remote(ServerAddr(addr)),
+                            metadata,
+                            is_local,
+                            parent: target.parent,
+                        }
+                    },
+                )
```
hawkw, line 133: silly style nit, not important: we could avoid having to write `inbound_ips_` by writing this instead:
```suggestion
            let balance = endpoint
                .push_map_target({
                    let inbound_ips = inbound_ips.clone();
                    move |((addr, metadata), target): ((SocketAddr, Metadata), Balance<T>)| {
                        tracing::trace!(%addr, ?metadata, ?target, "Resolved endpoint");
                        let is_local = inbound_ips.contains(&addr.ip());
                        Endpoint {
                            addr: Remote(ServerAddr(addr)),
                            metadata,
                            is_local,
                            parent: target.parent,
                        }
                    }
                })
```
up to you whether that's nicer, though...

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -1,104 +1,368 @@
...
+                .lift_new_with_target()
                 .push(http::NewBalancePeakEwma::layer(resolve))
                 // Drives the initial resolution via the service's readiness.
```
hawkw, line 136: this comment feels like it's in the wrong place --- i don't think the `MapErr` layer is responsible for driving resolutions. this should probably either be moved or deleted?

linkerd/app/outbound/src/http.rs
```diff
@@ -1,164 +1,142 @@
...
-    fn from(CanonicalDstHeader(dst): CanonicalDstHeader) -> HeaderPair {
-        HeaderPair(
-            HeaderName::from_static(CANONICAL_DST_HEADER),
-            HeaderValue::from_str(&dst.to_string()).expect("addr must be a valid header"),
-        )
+// === impl Outbound ===
+
+impl<C> Outbound<C> {
+    /// Builds a stack that handles protocol detection as well as routing and
+    /// load balancing for a single logical destination.
```
hawkw, line 41: copyedit nit, not important: maybe
```suggestion
    /// Builds a stack that handles protocol detection, routing, and
    /// load balancing for a single logical destination.
```

hawkw (approved):
Despite the size of the change, this branch ended up being surprisingly straightforward to review --- this makes a lot of sense! I commented on a lot of fairly small nits, but the overall approach is great.

linkerd/app/outbound/src/opaq/logical.rs
olix0r: I had initially called it `Target`, but that seemed too overloaded. I think the
name will change as we introduce real config types...

PR #2212, opened 2023-02-07
rename `OpaqueTransport` to `TaggedTransport`

hawkw:
Depends on #2209 

The "opaque transport" feature is somewhat confusingly named. This
feature refers to a system where the outbound proxy makes a direct
connection to the inbound proxy's inbound port, and sends a
`TransportHeader` containing the actual target port, along with an
optional authority and protocol hint. This feature can be used to
transport actually opaque protocols, but it is *also* used by
multicluster gateways to transport HTTP traffic non-opaquely. Therefore,
the name "opaque transport" is somewhat confusing here.

This branch renames the "opaque transport" middleware to "tagged
transport", which is (hopefully) somewhat clearer, as it no longer
suggests that this form of tagging inherently means that no
protocol-aware behavior is performed.

### DIFF
--- a/linkerd/app/outbound/src/endpoint.rs
+++ b/linkerd/app/outbound/src/endpoint.rs
@@ -113,11 +113,11 @@ impl<P> svc::Param<Option<http::detect::Skip>> for Endpoint<P> {
     }
 }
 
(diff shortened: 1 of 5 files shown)


olix0r:
This PR is stale, but this change is included in #2224

PR #2221, opened 2023-02-13
Use a cascading stack with protocol detection

olix0r:
The protocol detection stack does not easily allow for its inner stack to hold any per-target state because it commingles the (async) detection result with the original target.

This change updates the detection module to use a cascading inner stack so that the caller may determine how to use these targets. All implementations lift the original targets to preserve the prior behavior.

### DIFF
--- a/linkerd/detect/src/lib.rs
+++ b/linkerd/detect/src/lib.rs
@@ -9,6 +9,7 @@ use std::{
     fmt,
     future::Future,
     pin::Pin,
+    result::Result as StdResult,
     task::{Context, Poll},
 };
 use thiserror::Error;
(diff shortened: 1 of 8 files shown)


linkerd/detect/src/lib.rs
```diff
@@ -147,19 +149,21 @@ where
...
             };
 
             trace!("Dispatching connection");
-            let svc = inner.new_service((detected, target));
-            let mut svc = svc.ready_oneshot().await.map_err(Into::into)?;
-            svc.call(io::PrefixedIo::new(buf.freeze(), io))
-                .await
-                .map_err(Into::into)?;
+            let svc = inner.new_service(detected);
+            let mut svc = svc.ready_oneshot().await?;
```
hawkw, line 165: looks like removing the `map_err(Into::into)` broke compilation here: https://github.com/linkerd/linkerd2-proxy/actions/runs/4166843841/jobs/7211692762

PR #2224, opened 2023-02-13
outbound: Split out separate 'opaq' modules

olix0r:
We differentiate "opaque" and "tcp" stacks because some TCP stacks may
also apply for HTTP traffic. We shorten this in code as "opaq" because
it's unambiguous and the same length as "http"

* Add an `outbound::opaq` module for opaque stacks. ("opaq").
* Rename `push_tcp_{logical,concrete,forward}` to `push_opaq_*`
* Rename `OpaqueTransport` to `TaggedTransport` (Closes #2212).

### DIFF
--- a/linkerd/app/outbound/src/tcp/endpoint.rs
+++ b/linkerd/app/outbound/src/tcp/endpoint.rs
@@ -1,0 +1,57 @@
+use super::{opaque_transport::OpaqueTransport, *};
+use crate::{ConnectMeta, Outbound};
+use linkerd_app_core::{
+    io,
+    proxy::http,
+    svc, tls,
+    transport::{self, ClientAddr, Local, Remote, ServerAddr},
+    transport_header::SessionProtocol,
+    Error,
+};
+
+impl<C> Outbound<C> {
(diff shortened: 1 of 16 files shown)


linkerd/app/outbound/src/tcp/endpoint.rs
```diff
@@ -0,0 +1,57 @@
...
+        C: svc::MakeConnection<Connect, Metadata = Local<ClientAddr>, Error = io::Error>,
+        C: Clone + Send + 'static,
+        C::Connection: Send + Unpin,
+        C::Metadata: Send + Unpin,
+        C::Future: Send + 'static,
+    {
+        self.map_stack(|config, rt, connect| {
+            connect
+                // Initiates mTLS if the target is configured with identity. The
+                // endpoint configures ALPN when there is an opaque transport hint OR
```
hawkw, line 42: nit: should we say
```suggestion
                // endpoint configures ALPN when there is a tagged transport hint OR
```

hawkw (approved):
looks good to me --- differentiating between explicitly opaque and generic TCP is a good idea. i commented on one small comment i might change.

PR #2226, opened 2023-02-14
outbound: Lint stack target types

olix0r:
To support new discovery types, we want to changing how stacks are composed so that target types are domain-specific and configuration is instrumented via target wrappers.

This change sets up stack changes as follows:

* `Layers::push_flatten_new` is a utility for flatten "cascaded" `NewService` types by providing a target for child `NewService` impls.
* Target/Param constraints and impls are made uniform.
* In tests, we provide new test-specific target types.
* Configuration types are made ready to be stored in a cache.


olix0r:
trivial.

PR #2227, opened 2023-02-14
outbound: Split sidecar and ingress stack modules

olix0r:
In preparation for new target types, this change splits the ingress and proxy stack builders into ingress and sidecar modules, respectively.


olix0r:
trivial: simple move.

PR #2231, opened 2023-02-14
test: Disable tap::rejects_incorrect_identity_when_identity_is_expected

olix0r:
The `tap::rejects_incorrect_identity_when_identity_is_expected` test passes incorrectly. This change updates the test to assert the expected condition; and then the test is skipped (like the other tap integration tests). These tests all need discovery configuration to operate as intended.

This change also updates the devcontainer.json spec with an increased memory limit. This is necessary to run tests.


hawkw (approved):
makes sense, let's fix this later.

PR #2236, opened 2023-02-15
Add client-policy types

olix0r:
This change introduces the client policy types that will be used to configure outbound proxies. These types are analogous to those used for server policy in inbound proxies. They support integration with the Kubernetes Gateway API.

These types are relatively sketchy at the moment. They will change as they are integrated into the outbound proxy.

### DIFF
--- a/linkerd/proxy/client-policy/src/grpc.rs
+++ b/linkerd/proxy/client-policy/src/grpc.rs
@@ -1,0 +1,51 @@
+use linkerd_http_route::{grpc, http};
+use std::sync::Arc;
+
+pub use linkerd_http_route::grpc::{filter, r#match, RouteMatch};
+
+pub type Policy = crate::RoutePolicy<Filter>;
+pub type Route = grpc::Route<Policy>;
+pub type Rule = grpc::Rule<Policy>;
+
+// TODO HTTP2 settings
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Grpc {
+    pub routes: Arc<[Route]>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum Filter {
+    InjectFailure(filter::InjectFailure),
+    RequestHeaders(http::filter::ModifyHeader),
+    InternalError(&'static str),
+}
+
+#[inline]
+pub fn find<'r, B>(
+    routes: &'r [Route],
+    req: &::http::Request<B>,
+) -> Option<(RouteMatch, &'r Policy)> {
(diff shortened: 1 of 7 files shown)


linkerd/proxy/client-policy/src/grpc.rs
```diff
@@ -0,0 +1,51 @@
...
+                distribution,
+            },
+        }],
+    }
+}
+
+impl Default for Grpc {
+    fn default() -> Self {
+        Self {
+            routes: Arc::new([]),
```
hawkw, line 48: nit, maybe worth addressing later: it would be nice if these all held clones of the same `Arc`, instead of allocating a new one for each empty set of routes?

linkerd/proxy/client-policy/src/grpc.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+pub fn default(distribution: crate::RouteDistribution<Filter>) -> Route {
+    Route {
+        hosts: vec![],
+        rules: vec![Rule {
+            matches: vec![],
+            policy: Policy {
+                meta: crate::Meta::new_default("default"),
+                filters: Arc::new([]),
```
hawkw, line 38: nit, maybe worth addressing later: it would be nice if these all held clones of the same `Arc`, instead of allocating a new one for each empty set of filters?

linkerd/proxy/client-policy/src/lib.rs
```diff
@@ -0,0 +1,171 @@
...
+    RandomAvailable(Arc<[(RouteBackend<T>, u32)]>),
+}
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct RouteBackend<T> {
+    pub filters: Arc<[T]>,
+    pub backend: Backend,
+}
+
+// TODO(ver) how does configuration like failure accrual fit in here? What about
```
hawkw, line 77: what about...?

linkerd/proxy/client-policy/src/http.rs
```diff
@@ -0,0 +1,66 @@
...
+        Self {
+            routes: Arc::new([]),
+        }
+    }
+}
+
+impl Default for Http2 {
+    fn default() -> Self {
+        Self {
+            routes: Arc::new([]),
```
hawkw, line 63: would be nice for the empty `Arc`s to be singletons here as well...

hawkw (approved):
this looks like a good start! obviously there's room to improve stuff as we actually build out the client policy code, but this is about the skeleton i would expect.

PR #2240, opened 2023-02-16
outbound: Fix ingress-mode proxy

olix0r:
In 91bbbd4, we added a `LoadShed` onto the ingress proxy. This causes errors while discovery is pending.

This change removes the ingress `LoadShed` layer and documents the backpressure and error handling semantics of the ingress stack. This change also adds a descriptive error wrapper onto the inbound server stack.

### DIFF
--- a/linkerd/app/outbound/src/ingress.rs
+++ b/linkerd/app/outbound/src/ingress.rs
@@ -94,8 +94,8 @@ impl<N> Outbound<N> {
     /// outbound traffic is HTTP and HTTP detection is **always** performed. If
     /// HTTP detection fails, we revert to using the provided `fallback` stack.
     ///
-    /// An inner stack is service for each HTTP request. This stack must handle
-    /// its own caching.
+    /// The inner stack is used to create a service for each HTTP request. This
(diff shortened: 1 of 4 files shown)


linkerd/app/outbound/src/ingress.rs
```diff
@@ -93,12 +93,9 @@ impl<N> Outbound<N> {
...
     /// outbound traffic is HTTP and HTTP detection is **always** performed. If
     /// HTTP detection fails, we revert to using the provided `fallback` stack.
-    //
-    // Profile-based stacks are cached so that they can be reused across
-    // multiple requests to the same logical destination (even if the
-    // connections target individual endpoints in a service). No other caching
-    // is employed here: per-endpoint stacks are uncached, and fallback stacks
-    // are expected to be cached elsewhere, if necessary.
+    ///
+    /// An inner stack is service for each HTTP request. This stack must handle
```
hawkw, line 97: "an inner stack is service"?

hawkw (approved):
the change looks correct to me. not sure if i totally understand the tests?

PR #2250, opened 2023-02-20
outbound: Introduce a new outbound-route watch type

olix0r:
To support new outbound route types, this change adds an `outbound::http::Routes` type that is an enum of available logical router configurations. Accordingly, the Profile-specific routing stack is moved to a new `http::logical::profile` sub-module.

This sets the stage to further adapt routing configurations.

### DIFF
--- a/linkerd/app/outbound/src/http.rs
+++ b/linkerd/app/outbound/src/http.rs
@@ -25,7 +25,7 @@ mod retry;
 mod server;
 mod strip_proxy_error;
 
-pub use self::logical::{LogicalAddr, ProfileRoutes, Routes};
+pub use self::logical::{profile, LogicalAddr, Routes};
 pub(crate) use self::require_id_header::IdentityRequired;
 pub use linkerd_app_core::proxy::http::{self as http, *};
 
--- a/linkerd/app/outbound/src/http.rs
+++ b/linkerd/app/outbound/src/http.rs
@@ -9,9 +9,12 @@ use linkerd_app_core::{
         api_resolve::{ConcreteAddr, Metadata},
         core::Resolve,
     },
-    svc, Error,
+    svc,
+    transport::addrs::*,
+    Error,
 };
 use std::{fmt::Debug, hash::Hash};
+use tokio::sync::watch;
 
 pub mod concrete;
 mod endpoint;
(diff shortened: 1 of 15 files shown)


linkerd/app/outbound/src/http.rs
```diff
@@ -22,14 +25,46 @@ mod retry;
...
+        let routes = (mk)(&*profile_rx.borrow());
+        watch::channel(routes)
+    };
+
+    tokio::spawn(async move {
+        loop {
+            tokio::select! {
+                biased;
+                _ = tx.closed() => return,
+                _ = profile_rx.changed() => {
```
hawkw, line 49: what, if anything, should this task do if `changed()` returns an error (indicating that the sender of that watch has been dropped)?

linkerd/app/outbound/src/http/logical/profile.rs
```diff
@@ -0,0 +1,308 @@
...
+
+// === impl Params ===
+
+// Wraps a `NewService`--instantiated once per logical target--that caches a set
+// of concrete services so that, as the watch provides new `Params`, we can
+// reuse inner services.
+impl<T> Params<T>
+where
+    T: Clone + Debug + Eq + Hash + Send + Sync + 'static,
+{
```
hawkw, line 49: tiny nit: was this intended to be a doc comment on `layer`?
```suggestion
impl<T> Params<T>
where
    T: Clone + Debug + Eq + Hash + Send + Sync + 'static,
{
    /// Wraps a `NewService`--instantiated once per logical target--that caches
    /// a set of concrete services so that, as the watch provides new `Params`,
    /// we can reuse inner services.
```

linkerd/app/outbound/src/http/logical/profile.rs
```diff
@@ -0,0 +1,308 @@
...
+                // Records per-route metrics.
+                .push(
+                    metrics
+                        .http_profile_route
+                        .to_layer::<classify::Response, _, RouteParams<T>>(),
+                )
+                // Sets the per-route response classifier as a request
+                // extension.
+                .push(classify::NewClassify::layer())
+                // TODO(ver) .push(svc::NewMapErr::layer_from_target::<RouteError, _>())
```
hawkw, line 266: ?

linkerd/app/outbound/src/ingress.rs
```diff
@@ -252,69 +259,132 @@ impl Param<profiles::LookupAddr> for Http<RequestTarget> {
...
-                    .ok_or_else(|| ProfileRequired(addr.clone()))?;
+            (RequestTarget::Named(addr), profile) => {
+                let profile = profile.ok_or_else(|| ProfileRequired(addr.clone()))?;
+                let mut route =
+                    mk_route(&*profile.borrow()).ok_or_else(|| ProfileRequired(addr.clone()))?;
+                let routes = http::spawn_routes(profile, move |p: &profiles::Profile| {
+                    if let Some(r) = mk_route(p) {
+                        route = r;
+                    }
+                    route.clone()
```
hawkw, line 298: would it make sense for the closure passed to `spawn_routes` to return an `Option`, and only update the watch if the `Option` is `Some`? this looks like we will spuriously update the watch with the existing state if `mk_route` returns `None`. i don't think this is a major issue, but it's not strictly necessary...

linkerd/app/outbound/src/ingress.rs
```diff
@@ -252,69 +259,132 @@ impl Param<profiles::LookupAddr> for Http<RequestTarget> {
...
+    }
+
+    if let Some((addr, metadata)) = endpoint.clone() {
+        return Some(http::Routes::Endpoint(Remote(ServerAddr(addr)), metadata));
+    }
+
+    None
+}
+
+mod foo {}
```
hawkw, line 370: ?

hawkw (approved):
looks good to me overall --- i like this direction! i commented on a few small nits

PR #2251, opened 2023-02-20
outbound: Add a policy router

olix0r:
To support per-request backend routing, this change adds a policy router to the outbound stack. No configurations currently cause the policy router to be used, but this change updates the configuration types to support HTTPRoute routing.

---

- [x] https://github.com/linkerd/linkerd2-proxy/pull/2250


linkerd/app/outbound/src/http/logical/policy.rs
```diff
@@ -0,0 +1,396 @@
...
+}
+
+// === impl Params ===
+
+impl<T> Params<T>
+where
+    // Parent target type.
+    T: Clone + Debug + Eq + Hash + Send + Sync + 'static,
+{
+    /// HTTP or gRPC policy routing.
```
hawkw, line 79: nit, TIOLI: maybe
```suggestion
    /// Wraps an HTTP `NewService` with HTTP or gRPC policy routing layers.
```
or something?

linkerd/app/outbound/src/http/logical/policy.rs
```diff
@@ -0,0 +1,396 @@
...
+                > + Clone,
+        >,
+    > + Clone
+    where
+        // Assert that we can route the provided `Params`.
+        RouterParams<T, M, F>: svc::router::SelectRoute<
+            http::Request<http::BoxBody>,
+            Key = RouteParams<T, F>,
+            Error = NoRoute,
+        >,
```
hawkw, line 180: tiny nit: is there a reason this bound is on the `layer` function rather than on the `impl`? it doesn't bound the type params that exist only for the function, afaict?

linkerd/app/outbound/src/http/logical/policy.rs
```diff
@@ -0,0 +1,396 @@
...
+        };
+
+        let routes = routes
+            .iter()
+            .map(|route| route::Route {
+                hosts: route.hosts.clone(),
+                rules: route
+                    .rules
+                    .iter()
+                    .cloned()
```
hawkw, line 277: almost left a comment that this clones `distribution` unnecessarily: it's only `meta` and `filters` that are cloned into the `RouteParams`, so we could iterate by reference and have `mk_policy` take an `&RoutePolicy`, and clone just `meta` and `filters`. but, the distribution is also `arc`ed so this doesn't actually matter that much...

linkerd/app/outbound/src/http/logical/policy/tests.rs
```diff
@@ -0,0 +1,114 @@
...
+use linkerd_app_core::{
+    svc::NewService,
+    svc::{Layer, ServiceExt},
+    trace,
+};
+use std::{net::SocketAddr, sync::Arc};
+use tokio::time;
+
+#[tokio::test(flavor = "current_thread")]
+async fn header_based_route() {
```
hawkw, line 11: very cool!

linkerd/app/outbound/src/http/logical/policy.rs
```diff
@@ -0,0 +1,396 @@
...
+    T: Clone + Debug + Eq + Hash + Send + Sync + 'static,
+    // Request matcher.
+    M: Clone + Send + Sync + 'static,
+    // Request filter.
+    F: Eq + Hash,
+    F: Clone + Send + Sync + 'static,
+{
+    // Wraps a `NewService`--instantiated once per logical target--that caches a set
+    // of concrete services so that, as the watch provides new `Params`, we can
+    // reuse inner services.
```
hawkw, line 161: tiny nit, not a blocker: if this is a doc comment, it will be visible if we build RustDoc with `--document-private-items` --- so it may be slightly preferable to use a real doc comment even though the function is public.
```suggestion
    /// Wraps a `NewService`--instantiated once per logical target--that caches a set
    /// of concrete services so that, as the watch provides new `Params`, we can
    /// reuse inner services.
```

hawkw (approved):
this looks good to me overall! i left some very minor comments.

PR #2265, opened 2023-02-24
outbound: discover client policies from the `OutboundPolicy` API

hawkw:
This branch changes the outbound proxy to resolve client policies using
the `OutboundPolicy` proxy-api service (added in
linkerd/linkerd2-proxy-api#165). The proxy will preferentially use an
`OutboundPolicy` resolution over a `ServiceProfile` resolution in the
case that both are resolved and the `OutboundPolicy` is not a default
policy. If the `OutboundPolicy` _is_ a default policy, it is used only
when no `ServiceProfile` is resolved. The ingress-mode proxy also 
performs `OutboundPolicy` resolutions.

This branch also adds an implementation of the `OutboundPolicy` service
to the mock policy controller in `linkerd-app-integration`, and some
integration tests for client policies (including header-based routing).

Some notes:

 + The policy controller address, policy controller name, and workload
   identifier environment variables are now mandatory, since the proxy
   will always attempt to resolve outbound client policies.
 + Currently, targets constructed from an `OutboundPolicy` will report
   the original destination address as their `LogicalAddr`, rather than
   an authority. In the future, we will change the way metrics, tap, and
   error labels work to obviate the need for `LogicalAddr`s, and use
   structured metadata here instead.
 + We probably ought to add additional tests in follow-up branches.
 + I don't *love* that the `TargetAddr` type, which is the target
   address for both `ServiceProfile` and `OutboundPolicy` resolutions,
   has to live in the `linkerd-client-policy` crate. But unfortunately,
   it cannot be defined in `linkerd-app-outbound`, as the direct stack in
   `linkerd-app-inbound` must also implement `Param<TargetAddr>`.
 + The gateway doesn't currently resolve client policies, but I'll have a
   follow-up PR for that ready shortly.


hawkw:
pushed test docker image `mycoliza/l2-proxy:eliza.policy-discovery.744bd98c7`

hawkw:
trying to figure out why the k3d install test is failing...

hawkw:
> trying to figure out why the k3d install test is failing...

figured this out, the proxy-injector wasn't coming up because the proxy was stuck infinitely retrying when the policy controller built from edge was continually replying with `Unimplemented` to attempts to lookup outbound policy. 3c0e5787a8089da5bf4248c9711087995a24297f changes the proxy to not recover when the lookup returns `Unimplemented`, since this (presumably) indicates we are talking to an older control plane version.

this does give rise to some broader questions about the retry behavior, though. currently, proxies will retry both policies and service-profile resolutions infinitely (with an exponential backoff), bounded only by the request timeout. the way we currently join the client policy and service profile resolutions means that we wait for both to either succeed or error before continuing with the request. this means that if one resolution is retrying endlessly and the other one has succeeded, we will not continue processing the req until both either succeed or fail. 

i wonder if we want separate timeouts for both resolutions, or to move the retry logic up a level so that we only retry when both resolutions fail, or something?

linkerd/app/outbound/src/sidecar.rs
```diff
@@ -31,22 +32,49 @@ struct HttpSidecar {
...
+                // TODO(eliza): we already recover on errors elsewhere in the
+                // stack...this is kinda unfortunate...
+                let profile = profile.unwrap_or_else(|error| {
+                    tracing::warn!(%error, "Failed to resolve profile");
+                    None
+                });
+                let policy = policy.unwrap_or_else(|error| {
+                    tracing::warn!(%error, "Failed to resolve client policy");
+                    None
+                });
```
olix0r, line 73: I don't think policies should be optional. Errors should be propagated.

hawkw: If the error is `NotFound`, the policy controller is telling us that the thing we looked up is not something it knows about (i.e. not a Service IP, currently). So that case, at least, _should_ produce a `None`. I think that logic will have to be pushed down into the discovery service though, so that any error that makes it this far can bubble up.

olix0r: I don't think we should rely on that `NotFound` behavior. The controller should always provide a route if the proxy should route. Otherwise the proxy should fail.

hawkw: Okay, that means we'll want to change the current controller behavior to return a `Forward` backend for any IP it doesn't know about, then (cc @adleong)

linkerd/app/inbound/src/direct.rs
```diff
@@ -449,6 +449,12 @@ impl Param<profiles::LookupAddr> for GatewayTransportHeader {
     }
 }
 
+impl Param<linkerd_proxy_client_policy::TargetAddr> for GatewayTransportHeader {
+    fn param(&self) -> linkerd_proxy_client_policy::TargetAddr {
+        linkerd_proxy_client_policy::TargetAddr(self.target.clone().into())
+    }
+}
+
```
olix0r, line 457: This should be in the gateway and not in inbound. The inbound proxy doesn't care how client policies are looked up from these headers.

linkerd/app/inbound/src/policy/api.rs
```diff
@@ -34,7 +34,7 @@ where
     S::ResponseBody:
         http::HttpBody<Data = tonic::codegen::Bytes, Error = Error> + Default + Send + 'static,
 {
-    pub(super) fn new(workload: String, detect_timeout: time::Duration, client: S) -> Self {
+    pub(super) fn new(workload: Arc<str>, detect_timeout: time::Duration, client: S) -> Self {
```
olix0r, line 37: This doesn't belong in this PR.

linkerd/app/inbound/src/policy/config.rs
```diff
@@ -31,12 +28,19 @@ pub enum Config {
 // === impl Config ===
 
 impl Config {
-    pub(crate) fn build(
+    pub(crate) fn build<C>(
         self,
-        dns: dns::Resolver,
-        metrics: metrics::ControlHttp,
```
olix0r, line 37: Can the inbound change be made independently?

linkerd/proxy/client-policy/src/lib.rs
```diff
@@ -10,6 +11,10 @@ pub mod opaq;
 pub use linkerd_http_route as route;
 pub use linkerd_proxy_api_resolve::Metadata as EndpointMetadata;
 
+/// A target address for outbound policy discovery.
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct TargetAddr(pub Addr);
+
```
olix0r, line 17: I don't think this should exist. It's not used anywhere in this crate.

linkerd/proxy/client-policy/Cargo.toml
```diff
@@ -22,6 +22,7 @@ http = "0.2"
 linkerd2-proxy-api = { version = "0.8", optional = true, features = [
     "outbound",
 ] }
+linkerd-addr = { path = "../../addr" }
```
olix0r, line 25:
```suggestion
```

linkerd/app/outbound/src/discover.rs
```diff
@@ -1,9 +1,10 @@
-use crate::Outbound;
+use crate::{policy, Outbound};
 use linkerd_app_core::{
     profiles,
     svc::{self, stack::Param},
     Error,
 };
+pub use policy::TargetAddr;
```
olix0r, line 7: Instead, let's define a `pub struct DiscoverAddr(pub Addr)` and reexport it from the parent so the gateway can provide it is `outbound::DiscoverAddr`

olix0r (changes requested):

linkerd/app/outbound/src/discover.rs
```diff
@@ -53,11 +54,145 @@ impl<N> Outbound<N> {
...
+                tokio::spawn(async move { tx.closed().await });
+                return Ok((None, policy_rx));
+            }
+
+            // Otherwise, errors resolving the policy are considered fatal.
+            Err(policy_error)
+        })
+    })
+}
+
```
olix0r, line 173: I think this is a little easier to reason about if we split it up into some smaller chunks: https://github.com/linkerd/linkerd2-proxy/pull/2310

olix0r (approved):

linkerd/proxy/client-policy/src/lib.rs
```diff
@@ -191,6 +200,10 @@ impl Meta {
             Self::Resource { section, .. } => section.as_deref().unwrap_or(""),
         }
     }
+
+    pub fn is_default(&self) -> bool {
+        matches!(self, Self::Default { .. })
+    }
```
olix0r, line 206: Unused now?

```suggestion
```

linkerd/app/outbound/src/discover/tests.rs
```diff
@@ -35,13 +35,19 @@ async fn errors_propagate() {
         })
     };
 
-    let profiles = support::profile::resolver().profile(addr, profiles::Profile::default());
+    let discover = {
+        let profiles = support::profile::resolver().profile(addr, profiles::Profile::default());
+        svc::mk(move |OrigDstAddr(addr)| {
+            // TODO(eliza): policy!
+            profiles.clone().oneshot(profiles::LookupAddr(addr.into()))
```
olix0r, line 42: does this test care about profiles or policies at all, really? What are we testing?

linkerd/app/outbound/src/sidecar.rs
```diff
@@ -74,15 +90,92 @@ impl Outbound<()> {
...
-                        http::spawn_routes(rx, init, move |profile: &profiles::Profile| {
-                            Some(mk_routes(profile))
+
+                let mk_policy_routes = move |policy: &policy::ClientPolicy| {
+                    // A `ClientPolicy` does not have a single known logical
+                    // address, as it may consist of multiple distinct backends.
+                    // Therefore, just use the original destination address for now.
+                    // TODO(eliza): eventually, remove the logical addr field on
+                    // the policy route types when we use metadata for those
+                    // labels instead.
```
olix0r, line 100: Is this comment accurate? I don't understand it.

linkerd/app/outbound/src/sidecar.rs
```diff
@@ -74,15 +90,92 @@ impl Outbound<()> {
...
+                    let addr = orig_dst.0.into();
```
olix0r, line 101: Should we extract above instead? I.e. `let OrigDstAddr(orig_dst) = (*param).param();` so we don't have to do awkward `.0` stuff?

linkerd/app/outbound/src/discover/tests.rs
hawkw: i don't think this test actually cares besides that it needs something it can construct an inner stack from. if we remove the`From<(Option<profile::Receiver>, T)>` impl for `Discovery<T>` (making both a policy rx mandatory), we will need to return both a policy and profile rx here as well.

i believe this test _is_ testing the behavior of the `push_discover` stack, so we can't just remove that entirely.

linkerd/app/outbound/src/sidecar.rs
hawkw: The `HttpRoutes` types need to be constructed with a `LogicalAddr`, which is used for metrics labels etc. Unlike ServiceProfiles, we don't have a single authority to use for that logical address; the policy target is instead identified by its structured metadata. Currently, we use the original destination address for the `LogicalAddr`, but (as we discussed previously), we would like to remove the need to have `Param<LogicalAddr>` impls for the route types and generate metric labels and stuff based on the `Metadata` instead.

Does that make sense? I can try to reword the comment to make this clearer.

linkerd/app/outbound/src/sidecar.rs
```diff
@@ -34,19 +35,34 @@ struct HttpSidecar {
...
+        let discover = {
+            let detect_timeout = self.config.proxy.detect_protocol_timeout;
+
+            // TODO(eliza): opaque should have a different queue config...
+            let queue = self.config.http_request_queue;
+            let default_queue = policy::Queue {
+                capacity: queue.capacity,
+                failfast_timeout: queue.failfast_timeout,
+            };
+            discover::resolver(profiles, policies, default_queue, detect_timeout)
```
olix0r, line 63: Mmmm... Is this an HTTP request queue in any of the cases? Isn't this always a tcp connection queue?

linkerd/app/outbound/src/sidecar.rs
hawkw: elsewhere we also use that addr as an `OrigDstAddr`, so we would need to
re-wrap it in `OrigDstAddr` in those places. which is probably fine?

alternatively, we could just change this into 
```suggestion
                    let OrigDstAddr(addr) = orig_dst;
```
which is equivalent but IMO less ugly...

linkerd/app/outbound/src/sidecar.rs
hawkw: whoops i think you're right. fixed.

linkerd/app/outbound/src/sidecar.rs
hawkw: tried to reword it in 584fd7703821d05de0677056c5f9c8f6a711a2e3, LMKWYT!

linkerd/app/outbound/src/sidecar.rs
```diff
@@ -74,15 +90,92 @@ impl Outbound<()> {
...
+                        policy::Protocol::Grpc(ref grpc) => http::Routes::Policy(
+                            http::policy::Routes::Grpc(http::policy::GrpcRoutes {
+                                addr,
+                                backends: policy.backends.clone(),
+                                routes: grpc.routes.clone(),
+                            }),
+                        ),
+                        // TODO(eliza): if the policy's protocol is opaque, but we
+                        // are handling traffic as HTTP...that's obviously wrong.
+                        // should we panic or just fail traffic here?
```
olix0r, line 151: Let's get clearer about this condition. When would we handle traffic as HTTP but have policy protocol as opaque? Does this indicate a race of some sort? Or is this actually unreachable? It would be helpful to be clearer about this.

linkerd/app/outbound/src/sidecar.rs
```diff
@@ -74,15 +89,105 @@ impl Outbound<()> {
...
+                    // Unlike ServiceProfiles, OutboundPolicies do not contain a
+                    // single authority which can be used as a logical address.
+                    // Instead, the target of the policy is identified by
+                    // structured `policy::Meta` metadata. Since we don't have
+                    // an authority here, we use the original dest address as
+                    // the logical address for the time being.
+                    //
+                    // TODO(eliza): eventually, metrics and tap labels should be
+                    // generated from the policy's structured metadata, rather
+                    // than an address.
```
olix0r, line 108:
```suggestion
                    // We used the `OrigDstAddr` to lookup the policy.
                    // Attach it to routes so it can use that metadata.
```

linkerd/app/outbound/src/sidecar.rs
hawkw: I *believe* that if the policy's protocol was opaque, we won't build an HTTP stack, and won't be in this code at all. We should maybe have tests to ensure that this is the case, but changing the policy protocol should tear the whole stack down and rebuild it...

linkerd/app/outbound/src/sidecar.rs
olix0r: basically, i think this is mostly a distraction from what this code is doing.

PR #2267, opened 2023-02-25
Revert inbound policy discovery changes

olix0r:
The changes--specifically those in 93b06e69--prevent control plane boot-strapping: the identity controller is unable to serve requests because its proxy can't contact the destination pod for policy because the destination pod doesn't have identity yet.

Ultimately, we probably want to change Linkerd's control plane deployment topology to avoid these bootstrapping issues. In the meantime, we'll need to revisit our approach to these changes.

* Revert 89ee318e inbound: Introduce a `policy::LookupAddr` type (#2264)
* Revert 93b06e69 inbound: Remove default policies (#2204)
* Revert c186d88a inbound: connections wait for ServerPolicy discovery (#2186)


hawkw (approved):
Well...that's unfortunate. I agree that reverting is a reasonable solution for now.

PR #2268, opened 2023-02-25
ci: Add a linkerd install workflow

olix0r:
We don't currently test that proxy changes are at least usable in a
basic Linkerd installation. This adds a CI workflow to test all Linkerd
code changes against a real k8s cluster with the control plane from the
most recent edge release.

PR #2275, opened 2023-02-27
allow `opaque_hidden_inferred_bound` warning on nightly

hawkw:
Currently, our nightly builds are failing due to the new warning `opaque_hidden_inferred_bound`, which triggers when an opaque type (`impl Trait`) in an associated type position does not explicitly include the associated type's trait bounds (e.g. returning a `Service<Future = impl Send, ...>`) or similar.

Unfortunately, we cannot simply change our code to make the trait bound's type explicit, as changing `impl Send` to
`impl Future<...> + Send` in this position results in a surprising error which I don't think is correct:

```
error[E0277]: `impl std::marker::Send` is not a future
  --> linkerd/app/outbound/src/http/logical.rs:97:30
   |
97 |                     Future = impl Future<Output = Result<http::Response<http::BoxBody>, Error>> + Send,
   |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `impl std::marker::Send` is not a future
   |
   = help: the trait `futures::Future` is not implemented for `impl std::marker::Send`
   = note: impl std::marker::Send must be a future or must implement `IntoFuture` to be awaited
   = note: required for `stack::map_err::ResponseFuture<(), impl std::marker::Send>` to implement `futures::Future`

For more information about this error, try `rustc --explain E0277`.
```

See https://github.com/linkerd/linkerd2-proxy/pull/2268#issuecomment-1446738668 for details.

This should probably be reported on the Rust issue tracker, since a warning that's (apparently) impossible to fix seems not great. However, for now, we can simply allow this warning for our nightly builds.

This should fix CI.

### DIFF
--- a/.github/workflows/fuzzers.yml
+++ b/.github/workflows/fuzzers.yml
@@ -17,7 +17,7 @@ env:
   CARGO_INCREMENTAL: 0
   CARGO_NET_RETRY: 10
   RUST_BACKTRACE: short
-  RUSTFLAGS: "-D warnings -A deprecated"
+  RUSTFLAGS: "-D warnings -A deprecated -A opaque_hidden_inferred_bound"
   RUSTUP_MAX_RETRIES: 10
 
 permissions:
--- a/.github/workflows/fuzzers.yml
+++ b/.github/workflows/fuzzers.yml
@@ -17,7 +17,7 @@ env:
   CARGO_INCREMENTAL: 0
   CARGO_NET_RETRY: 10
   RUST_BACKTRACE: short
-  RUSTFLAGS: "-D warnings -A deprecated -A opaque_hidden_inferred_bound"
+  RUSTFLAGS: "-D warnings -A deprecated"
   RUSTUP_MAX_RETRIES: 10
 
 permissions:
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -13,7 +13,7 @@ on:
 env:
   CARGO_INCREMENTAL: 0
   CARGO_NET_RETRY: 10
-  RUSTFLAGS: "-D warnings"
+  RUSTFLAGS: "-D warnings -A opaque_hidden_inferred_bound"
   RUSTUP_MAX_RETRIES: 10
 
 permissions:


mateiidavid:
> Unfortunately, we cannot simply change our code to make the trait bound's type explicit, as changing impl Send to
impl Future<...> + Send in this position results in a surprising error which I don't think is correct:

Am I understanding this right? So basically, we have a `Send` bound on the associated type, and if we use an associated type that uses `impl Future` (so whatever opaque type that implements Future), and we add that this type has a `Send` bound, it won't work? 

Also, just curious but how come we use an opaque type (`impl Send`) instead of just specifying a bound `Send` on it for more flexibility? Sorry bit besides the point, but while I'm here... 🤷🏻

olix0r:
FYI, i believe this change is also included in Rust 1.66; so we may need a more robust change than changing the nightly workflow.

hawkw:
@mateiidavid 
> > Unfortunately, we cannot simply change our code to make the trait bound's type explicit, as changing impl Send to
> > impl Future<...> + Send in this position results in a surprising error which I don't think is correct:
> 
> Am I understanding this right? So basically, we have a `Send` bound on the associated type, and if we use an associated type that uses `impl Future` (so whatever opaque type that implements Future), and we add that this type has a `Send` bound, it won't work?

I think what you're describing here is kind of backwards: we have an associated type where the trait bounds the associated type with `Future`:
```rust
pub trait Service {
     type Future: Future<Output = Result<Self::Response, Self::Error>>;
    // ...
}
```
and here, we are returning a `Service` and saying that the `Future` type is `impl Send`, to indicate that the future is also `Send`:
```rust
fn stuff() -> impl Service<
    Future = impl Send,
    // ...
> {
    // ...
}
```
Since `Service` bounds its `Future` associated type with the `Future` trait, the `impl Service`'s `Future` associated type is already required to implement `Future`. We don't need to return `impl Future` there, although the `opaque_hidden_inferred_bound` lint asks us to add it (presumably for explicitness)

> Also, just curious but how come we use an opaque type (`impl Send`) instead of just specifying a bound `Send` on it for more flexibility? Sorry bit besides the point, but while I'm here... 🤷🏻

We use an opaque type in _return type position_ --- trait bounds constrain generic type parameters, which are (generally) inputs to a function[^1]; we cannot add trait bounds to a return type that a function determines on its own. Instead we return opaque types (`impl Trait`s) to say "this function returns 'something' which implements these traits".

[^1] a function can _return_ one of its type parameters, but it isn't responsible for _determining what type_ that value is --- type parameters are types determined by the caller, while opaque types (`impl Trait`) are determined by the returner. If that makes sense...

olix0r (approved):

PR #2279, opened 2023-02-28
stack: Make `failfast::Gate` general purpose

olix0r:
`failfast::Gate` is a stack middleware that supports changing its readiness based on some external signal (the failfast state changing). Nothing about the gate's behavior is specific to failfast, though; and this type of gating is general useful on its own.

To setup further reuse, this change moves `stack::failfast::gate` to `stack::gate`. The gate is now tested independently from failfast. Failfast continues to maintain an (optional) `gate::Tx` so that it can control a gate.


linkerd/stack/src/gate.rs
```diff
@@ -0,0 +1,261 @@
...
+}
+
+impl Drop for Rx {
+    fn drop(&mut self) {
+        let Rx(_, handle) = self;
+        if Arc::strong_count(handle) == 1 {
+            self.0.closed.notify_waiters();
+        }
+    }
+}
```
hawkw, line 81: take it or leave it: it occurrs to me that if, rather than two `Notify`s and an `Arc<()>`, we used a single `watch<bool>` channel we could get this behavior for free --- to notify the gate, a sender would send a `true` or `false` to indicate closure, and when all the senders are dropped, the watch receiver's `closed` future would complete.

i think the current approach works, too, but we could maybe simplify the implementation a bit this way?

hawkw (commented):
factoring this out makes sense to me. i left a comment on how i think we could potentially make the implementation a lot simpler by using a `tokio::sync::watch` channel rather than a pair of `Notify`s, but i don't think that's a hard blocker.

linkerd/stack/src/gate.rs
olix0r: `watch::Sender` isn't `Clone`. Our `Tx` is.

PR #2284, opened 2023-03-01
outbound: Decouple backend caching from request distribution

olix0r:
The `BackendCache` module currently produces `NewDistribute` instances. This does not allow any other layers to be insertedb between the distribution decision and the caching decision.

To support per-route-backend filters, we need to separate these modules.

This change updates `BackendCache` as `NewBackendCache` and adds a basic service to extract targets from the cache.

`NewDistribute` is updated so that it builds new inner services instead of using the cache directly. The distribution layers are moved up into the route stack to prepare for additional filters being applied.

### DIFF
--- a/linkerd/distribute/src/cache.rs
+++ b/linkerd/distribute/src/cache.rs
@@ -1,51 +1,63 @@
-use super::{params, NewDistribute};
-use linkerd_stack::{layer, NewService, Param};
+use super::params;
+use linkerd_stack::{layer, ExtractParam, NewService};
 use parking_lot::Mutex;
 use std::{fmt::Debug, hash::Hash, sync::Arc};
 
-/// A [`NewService`] that produces [`NewDistribute`]s using a shared cache of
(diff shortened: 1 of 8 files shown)


linkerd/distribute/src/cache.rs
```diff
@@ -1,51 +1,63 @@
-use super::{params, NewDistribute};
-use linkerd_stack::{layer, NewService, Param};
+use super::params;
+use linkerd_stack::{layer, ExtractParam, NewService};
 use parking_lot::Mutex;
 use std::{fmt::Debug, hash::Hash, sync::Arc};
 
-/// A [`NewService`] that produces [`NewDistribute`]s using a shared cache of
+/// A [`NewService`] that produces [`BackendCache`]s using a shared cache of
```
hawkw, line 6: this docs link is emits a warning since `BackendCache` is not publicly re-exported.

hawkw (approved):
cool, looks good to me!

PR #2288, opened 2023-03-03
integration: add test policy controller

hawkw:
Currently, the proxy integration tests run without a policy controller, and instead use environment variables to configure default inbound server policies. However, we do not intend to add a similar environment based defaulting mechanism for outbound client policies. Therefore, when the proxy starts consuming the outbound client policy API, we will no longer be able to run the integration tests without a policy controller.

This branch adds an implementation of a mock policy controller for the integration tests. This was factored out from PR #2265, which implements outbound policy discovery and therefore requires a policy controller.

Note that the mock policy controller added in #2265 serves both the inbound and outbound APIs, while this PR implements only the inbound API. This is because the a version of `linkerd2-proxy-api` that includes the outbound policy API has not yet been released, and I wanted to be able to merge this branch without waiting for a proxy API release. However, the mock policy controller implementation here was taken directly from #2265 with the outbound half removed. Some aspects of the implementation may seem somewhat more complex than necessary to implement only the inbound policy API, but I thought it was nicer to have a smaller diff when the outbound half is added in the near future.

One small complexity here is that the inbound proxy will use default policies when it has not yet resolved. This means there is a potential for racy behavior in tests when a request is sent to an inbound proxy, if it has not yet started a watch on the policy for a port (or is in the process of starting that watch). This can result in flaky test failures: for example, in metrics tests, the name of the policy can change between requests if one request is handled with the default and the next request is handled with one resolved from the mock policy controller, resulting in two sets of labels rather than one. To avoid the test proxy flapping between the default policy and one resolved from the mock policy controller, I've also changed the proxy test harness so that any configured inbound ports in the test are added to
`LINKERD2_PROXY_INBOUND_PORTS`, so that the proxy will eagerly resolve them. The test harness now waits for the mock policy controller to tell it that the proxy has started watches on those ports before actually running the test code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -90,6 +90,52 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
+name = "axum"
+version = "0.6.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8582122b8edba2af43eaf6b80dbfd33f421b5a0eb3a3113d21bc096ac5b44faf"
+dependencies = [
+ "async-trait",
+ "axum-core",
+ "bitflags",
+ "bytes",
+ "futures-util",
+ "http",
+ "http-body",
+ "hyper",
+ "itoa",
+ "matchit",
+ "memchr",
+ "mime",
+ "percent-encoding",
+ "pin-project-lite",
+ "rustversion",
+ "serde",
+ "sync_wrapper",
+ "tower",
+ "tower-http",
+ "tower-layer",
+ "tower-service",
+]
+
+[[package]]
+name = "axum-core"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b2f958c80c248b34b9a877a643811be8dbca03ca5ba827f2b63baf3a81e5fc4e"
+dependencies = [
+ "async-trait",
+ "bytes",
+ "futures-util",
+ "http",
+ "http-body",
+ "mime",
+ "rustversion",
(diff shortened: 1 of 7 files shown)


Cargo.lock
```diff
@@ -89,6 +89,52 @@ version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
+[[package]]
+name = "axum"
```
olix0r, line 93: is this required?

hawkw: it will be when we add the outbound API to the mock controller. using `tonic`'s "transport" feature is necessary to serve multiple gRPC APIs on the same port, and "transport" depends on `axum`, because it uses `axum`'s request router to implement this. 

we could avoid the dependency in this PR and only add it in the subsequent PR that adds the outbound API to the mock controller. however, i decided not to do that, because this would make the diff in the subsequent PR bigger --- it would change how the server is started. this way, we would just add one line to add the outbound API service to the server builder, which i thought was a neater diff. up to you though.

olix0r: Do we need all of axum's transitive dependencies? It's not a blocker, but it's not awesome to have to take on a bunch of new deps for what boils down to a match statement.

hawkw: i can replace this with a handwritten router if that's preferable!

olix0r: It depends, but if we're in the ballpark of <100LOC, i'd say that's preferable to the `transport` feature. Same goes for the policy controller, for that matter.

olix0r (approved):

Cargo.lock
olix0r: Y'know what... we can always tear it out later if it's a pain. #FISI.

PR #2291, opened 2023-03-06
outbound: change `push_discover` to take a `Service`

hawkw:
Currently, the `push_discover` method in `linkerd_app_outbound` takes a
`GetProfile` implementation, which is used to discover ServiceProfiles.
In order to change this function to discover both ServiceProfiles and
OutboundPolicies, PR #2265 will change this function to take an
arbitrary `Service` that may discover ServiceProfiles, OutboundPolicies,
or both. In order to make reviewing the change adding client policy
discovery easier, the change to `push_discover` was factored out into
this PR.

Since `push_discover` now takes an arbitrary discovery service, and
since we only want the configured profile discovery allowlist to apply
to ServiceProfiles (and not to OutboundPolicies), this change also pulls
out the filtering of addresses not in the allowlist from a layer pushed
in `push_discover` into a middleware which is added to the profile
discovery `Service`.

### DIFF
--- a/linkerd/app/gateway/src/lib.rs
+++ b/linkerd/app/gateway/src/lib.rs
@@ -2,7 +2,6 @@
 #![forbid(unsafe_code)]
 
 use linkerd_app_core::{
-    addr::NameMatch,
     io, profiles,
     proxy::{
         api_resolve::{ConcreteAddr, Metadata},
@@ -11,7 +10,7 @@ use linkerd_app_core::{
     svc, tls,
     transport::addrs::*,
     transport_header::SessionProtocol,
-    Error,
+    Error, NameMatch,
 };
 use linkerd_app_inbound::{self as inbound, GatewayAddr, Inbound};
 use linkerd_app_outbound::Outbound;
--- a/linkerd/app/gateway/src/lib.rs
+++ b/linkerd/app/gateway/src/lib.rs
@@ -2,6 +2,7 @@
 #![forbid(unsafe_code)]
 
 use linkerd_app_core::{
+    addr::NameMatch,
     io, profiles,
     proxy::{
         api_resolve::{ConcreteAddr, Metadata},
@@ -10,7 +11,7 @@ use linkerd_app_core::{
     svc, tls,
     transport::addrs::*,
     transport_header::SessionProtocol,
-    Error, NameMatch,
+    Error,
 };
 use linkerd_app_inbound::{self as inbound, GatewayAddr, Inbound};
 use linkerd_app_outbound::Outbound;
--- a/linkerd/app/outbound/src/discover.rs
+++ b/linkerd/app/outbound/src/discover.rs
@@ -1,12 +1,14 @@
 use crate::Outbound;
+use futures::future;
 use linkerd_app_core::{
     profiles,
     svc::{self, stack::Param},
-    Error, Infallible,
+    Addr, AddrMatch, Error,
 };
 use std::{
     hash::{Hash, Hasher},
     ops::Deref,
+    task::{Context, Poll},
 };
 use tokio::sync::watch;
 use tracing::debug;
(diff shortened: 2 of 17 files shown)


linkerd/app/outbound/src/discover.rs
```diff
@@ -21,11 +23,18 @@ pub struct Discovery<T> {
     profile: Option<profiles::Receiver>,
 }
 
+/// Wraps a discovery service with an allowlist of addresses.
+#[derive(Clone, Debug)]
+pub struct WithAllowlist<S> {
+    allow_discovery: AddrMatch,
+    inner: S,
+}
```
olix0r, line 31: This strikes me as something that shouldn't be in this module. It's not referenced anywhere in here. Does this even carry weight as a general middlewear?

olix0r: Err, it's not exactly exactly a push switch i guess. It just seems in the uncanny valley between reusable/generic and special-purpose.

linkerd/app/outbound/src/discover/tests.rs
```diff
@@ -190,14 +185,21 @@ async fn no_profiles_when_outside_search_nets() {
...
         cfg.allow_discovery =
             AddrMatch::new(None, Some(IpNet::from(IpAddr::from([192, 0, 2, 66]))));
         cfg
     };
+    let profiles = {
+        // XXX we should assert that the resolver isn't even invoked, but the mocked resolver
+        // doesn't support that right now. So, instead, we return a profile for resolutions to
+        // and assert (below) that no profile is provided.
+        let resolver = support::profile::resolver().profile(addr, profiles::Profile::default());
+        super::WithAllowlist::new(resolver, cfg.allow_discovery.clone())
```
olix0r, line 200: Why do we need the allowlist here. What are we testing with it?

linkerd/app/outbound/src/lib.rs
```diff
@@ -30,7 +30,7 @@ use std::{
     time::Duration,
 };
 
-mod discover;
+pub mod discover;
```
olix0r, line 33: This isn't exactly a blocker but it gives me pause. What outbound-specific discovery logic are we trying to expose?

olix0r (changes requested):
I think the best course of action is probably to move the allowlist stuff directly into the service profile crate. It's not an outbound-specific concern. There's not enough value in it being generic/decoupled from its use case. It's fine.

We shouldn't have to care about the type in tests, though; and it should just be part of how a profile service is built for each proxy variant (including inbound!)

linkerd/app/outbound/src/lib.rs
hawkw: this was made public as the `gateway` crate also needs to construct the `WithAllowlist` type with its own allowlist. moving that code into `linkerd-proxy-service-profiles` will obviate the need for that!

linkerd/app/outbound/src/discover.rs
hawkw: ah, the reason it's not currently in `linkerd2-proxy-service-profiles` is that the `AddrMatch` stuff is defined in `app-core`, which depends on `service-profiles`, so we would need to move the `AddrMatch` types as well. which i'm willing to do.

linkerd/app/outbound/src/discover/tests.rs
hawkw: the only thing that this test is testing is...whether the allowlist works. i just removed the test.

linkerd/app/gateway/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 #![forbid(unsafe_code)]
 
 use linkerd_app_core::{
+    addr::NameMatch,
```
olix0r, line 5: Can we preserve the reexports so we don't have to change all of these in this PR? sorry.

linkerd/service-profiles/src/allowlist.rs
```diff
@@ -0,0 +1,46 @@
...
+where
+    G: GetProfile,
+    G::Error: Send,
+{
+    type Future = future::Either<future::Ready<Result<Option<Receiver>, Self::Error>>, G::Future>;
+    type Error = G::Error;
+
+    fn get_profile(&mut self, addr: LookupAddr) -> Self::Future {
+        // TODO(ver) Should this allowance be parameterized by
+        // the target type?
```
olix0r, line 33:
```suggestion
```

Question answered

olix0r (approved):
lgtm. Would generally prefer that we continue to export the Match types from core (as we do for the Addr types).

hawkw:
sounds good, i'll put it back

PR #2294, opened 2023-03-07
client-policy: Simplify route backend collection.

olix0r:
There's some pretty obtuse code to render route iterators. This change avoids this by passing around a hash set to collect backends into. This is sufficient, especially, as this is all handled internally to the crate.


hawkw (approved):
oh, yeah, that's definitely nicer, good call!

PR #2308, opened 2023-03-10
Simplify outbound discovery target types

olix0r:
By avoiding a dedicated param value type, we can more flexibly instrument discovery!


hawkw (approved):
making the `push_discover` stack generic over an arbitrary key is a great solution, much nicer than the stuff i was trying to do with `ExtractParam`. looks good to me!

PR #2310, opened 2023-03-11
split out synthetic policies for readability

olix0r:


linkerd/app/outbound/src/discover.rs
```diff
@@ -68,109 +69,109 @@ pub(crate) fn resolver(
...
-                    });
-                    let policy_from_profile = move |profile: &profiles::Profile| {
-                        let dispatcher = if let Some((addr, meta)) = profile.endpoint.clone() {
-                            policy::BackendDispatcher::Forward(addr, meta)
-                        } else if let Some(profiles::LogicalAddr(ref addr)) = profile.addr {
-                            policy::BackendDispatcher::BalanceP2c(
-                                ewma,
-                                policy::EndpointDiscovery::DestinationGet {
-                                    path: addr.to_string(),
-                                },
```
olix0r, line 112: This case isn't meaningful because we expect the policy control to always return for anything we could balance over.

linkerd/app/outbound/src/discover.rs
```diff
@@ -68,109 +72,181 @@ pub(crate) fn resolver(
...
+    let policy = synthesize_forward_policy(detect_timeout, queue, forward);
+    tracing::debug!(?policy, "Synthesizing policy");
+    let (tx, rx) = watch::channel(policy);
+    tokio::spawn(async move {
+        tx.closed().await;
+    });
+    rx
+}
+
+fn synthesize_forward_policy(
```
olix0r, line 185: I moved this out of `client-policy` because it's basically a hardcoded configuration. I wouldn't want to rely on it anywhere else but here.

linkerd/app/outbound/src/discover.rs
olix0r: It turns out this was load-bearing for integration tests: when profiles provided a logical name, we would proceed with balancing/discovery on that name. But this isn't how we expect things to work in practice: We want to rely on the policy controller to provide service information. If it doesn't know about a service, we expect to fallback to a forward stack.

linkerd/app/outbound/src/discover.rs
```diff
@@ -68,109 +72,186 @@ pub(crate) fn resolver(
...
+    name: impl ToString,
```
hawkw, line 190: nit, TIOLI: i feel like this could maybe take an `Arc<Meta>` rather than an `impl ToString` --- we use hard-coded strings ("endpoint" and "fallback") for the name, so that could be a singleton rather than creating a new `Arc<Meta>` for every synthesized policy? not a blocker.

linkerd/app/outbound/src/discover.rs
```diff
@@ -68,109 +72,186 @@ pub(crate) fn resolver(
...
+            Ok((None, policy))
         })
     })
 }
 
+#[inline]
+fn is_not_found(e: &Error) -> bool {
+    e.downcast_ref::<tonic::Status>()
+        .map(|s| s.code() == tonic::Code::NotFound)
+        .unwrap_or(false)
```
hawkw, line 129: currently the error is always a `Status` at the top level, but it occurs to me that this is a bit brittle in that the error might get wrapped...maybe this should traverse a whole source chain if there is one? we can address separately though.

linkerd/app/outbound/src/discover.rs
```diff
@@ -68,109 +72,186 @@ pub(crate) fn resolver(
...
+        // TODO(ver) We should be able to figure outc resource coordinates for
+        // the endpoint?
+        synthesize_forward_policy("endpoint", detect_timeout, queue, addr, meta)
+    };
+
+    let policy = mk(&*profile.borrow_and_update());
+    tracing::debug!(?policy, profile = ?*profile.borrow(), "Synthesizing policy from profile");
+    let (tx, rx) = watch::channel(policy);
+    tokio::spawn(async move {
+        let mut profile = profile;
```
hawkw, line 152: is this necessary? `profile` is already `mut` in the function declaration...

linkerd/app/outbound/src/discover.rs
```diff
@@ -68,109 +72,186 @@ pub(crate) fn resolver(
...
+                return;
+            };
+            let policy = mk(&*profile.borrow());
+            tracing::debug!(?policy, "Profile updated; synthesizing policy");
+            if tx.send(policy).is_err() {
+                tracing::debug!("Policy watch closed, terminating");
+                return;
+            }
+        }
+    });
```
hawkw, line 165: occurred to me that this future maybe ought to have a span indicating the orig_dst or something, since otherwise, the logs here are not going to tell us _which_ policy watch any of this happened to...i can address that after merging though.

hawkw (approved):
this looks great, thanks for taking the time to fix up the tests (i didn't want to do that). i had some very minor nits to pick but i can address some of them myself after merging, as well.

linkerd/app/outbound/src/discover.rs
olix0r: keep in mind this is going to happen at most once per connection (and less frequently than that in practice).

linkerd/app/outbound/src/discover.rs
olix0r: oh yeah we should use the `errors` helper here probably but it's fine for now.

linkerd/app/outbound/src/discover.rs
olix0r: this was probably here first. can be removed.

linkerd/app/outbound/src/discover.rs
olix0r: we already know it on the parent span

olix0r: oh but we do want `in_current_span` here

PR #2311, opened 2023-03-11
Clarify sidecar routing behavior

olix0r:


linkerd/app/outbound/src/ingress.rs
```diff
@@ -319,7 +319,9 @@ impl TryFrom<Discovery<Http<RequestTarget>>> for Http<Logical> {
                     }
 
                     // If the client policy is default, try the service profile first
```
hawkw, line 321: comment should maybe be updated, since the default-ness of the policy is not directly the condition here.

linkerd/app/outbound/src/sidecar.rs
```diff
@@ -288,12 +175,120 @@ impl std::hash::Hash for Sidecar {
...
+                            Some(Self::mk_profile_routes(addr.clone(), profile))
+                        });
+                    return HttpSidecar {
+                        orig_dst,
+                        version,
+                        routes,
+                    };
+                }
+            }
+        }
```
hawkw: this feels needlessly convoluted --- `addr` is only `Some` if `profile.has_routes_or_targets()`, why do we need the `Option`? can't we just write
```suggestion
            // Only use service profiles if there are novel routes/target
            // overrides.
            if profile.borrow().has_routes_or_targets() {
                let addr = profile.borrow().addr.clone()
                tracing::debug!("Using ServiceProfile");
                let init = Self::mk_profile_routes(addr.clone(), &*profile.borrow_and_update());
                let routes =
                    http::spawn_routes(profile, init, move |profile: &profiles::Profile| {
                        Some(Self::mk_profile_routes(addr.clone(), profile))
                    });
                return HttpSidecar {
                    orig_dst,
                    version,
                    routes,
                };
            }
```

hawkw: oh i see that you just fixed that, whoops.

linkerd/app/outbound/src/sidecar.rs
```diff
@@ -288,12 +175,118 @@ impl std::hash::Hash for Sidecar {
...
+
+        Some(http::Routes::Policy(http::policy::Routes::Http(
+            http::policy::HttpRoutes {
+                addr: orig_dst.into(),
+                routes,
+                backends: policy.backends.clone(),
+            },
+        )))
+    }
+    fn mk_profile_routes(addr: profiles::LogicalAddr, profile: &profiles::Profile) -> http::Routes {
```
hawkw, line 271: tiny nit:
```suggestion
    }
    
    fn mk_profile_routes(addr: profiles::LogicalAddr, profile: &profiles::Profile) -> http::Routes {
```

linkerd/app/outbound/src/sidecar.rs
```diff
@@ -288,12 +175,118 @@ impl std::hash::Hash for Sidecar {
...
+                let routes =
+                    http::spawn_routes(profile, init, move |profile: &profiles::Profile| {
+                        Some(Self::mk_profile_routes(addr.clone(), profile))
+                    });
+                return HttpSidecar {
+                    orig_dst,
+                    version,
+                    routes,
+                };
+            }
```
hawkw, line 207: this feels needlessly convoluted --- `addr` is only `Some` if `profile.has_routes_or_targets()`, why do we need the `Option`? can't we just write
```suggestion
            // Only use service profiles if there are novel routes/target
            // overrides.
            if profile.borrow().has_routes_or_targets() {
                let addr = profile.borrow().addr.clone()
                tracing::debug!("Using ServiceProfile");
                let init = Self::mk_profile_routes(addr.clone(), &*profile.borrow_and_update());
                let routes =
                    http::spawn_routes(profile, init, move |profile: &profiles::Profile| {
                        Some(Self::mk_profile_routes(addr.clone(), profile))
                    });
                return HttpSidecar {
                    orig_dst,
                    version,
                    routes,
                };
            }
```

linkerd/app/outbound/src/sidecar.rs
olix0r: basically, this makes sure that we have a logicaladdr that we can reuse no matter the state of the profile.

PR #2312, opened 2023-03-11
just: Fix docker tag formatting

olix0r:
The justfile does not properly escape tags with multiple slashes.

This change updates the sed command to use a global replace.


stevej (commented):
LGTM. 👍

PR #2314, opened 2023-03-13
test: add mock client policy resolver

hawkw:
Currently, `linkerd-app-outbound`'s stack tests construct a discovery service for `push_discover` which only discovers service profiles, and does not resolve client policies. This works fine currently, as `Discovery` can be constructed with only a profile receiver, as the client policy receiver is optional.

However, when we add client policy discovery to the gateway, we will be able to make the policy receiver no longer optional, as a client policy will always be discovered. In that case, these tests will no longer compile. Therefore, this branch adds an implementation of a mock client policy resolver in `linkerd-test`, and updates these stack tests to actually use it.

This was factored out of the change that actually adds policy discovery to the gateway, and is a prerequisite for that change.


linkerd/app/test/src/resolver.rs
```diff
@@ -29,6 +32,16 @@ pub type Dst<E> = Resolver<NameAddr, DstReceiver<E>>;
 
 pub type Profiles = Resolver<Addr, Option<profiles::Receiver>>;
 
+#[cfg(feature = "client-policy")]
+pub type ClientPolicies = Resolver<Addr, watch::Receiver<client_policy::ClientPolicy>>;
```
olix0r, line 36: Is it worth sticking this stuff into a new file/submodule? We can minimize cfg blocks to the top level module

hawkw: yeah, i also thought about this, didn't do it because there were just a couple impl blocks, but then it rapidly grew in size...

PR #2315, opened 2023-03-13
gateway: discover client policies from the policy controller

hawkw:
Depends on #2314.

Currently, the gateway stack does not resolve client policies, and only
uses service profiles for client-side policy discovery. This branch
updates the gateway proxy to also perform client policy discovery,
similarly to the outbound proxy.

Because all outbound stacks, including the one constructed by the
gateway, will now discover client policies, the client policy `Receiver`
in the `Discovery` type, which is currently an `Option`, can now be made
non-optional. This simplifies some code a little bit and removes some
`expect`s.

### DIFF
--- a/linkerd/app/gateway/src/server.rs
+++ b/linkerd/app/gateway/src/server.rs
@@ -1,5 +1,5 @@
 use crate::Gateway;
-use futures::FutureExt;
+use futures::TryFutureExt;
 use linkerd_app_core::{
     errors, io, profiles, proxy::http, svc, tls, transport::addrs::*,
     transport_header::SessionProtocol, Addr, Error,
@@ -94,21 +94,21 @@ impl Gateway {
 
                 // TODO(eliza): we should probably also add the allowlist to
                 // policy resolution...
-            let policy = policies
-                .get_policy(addr.into())
-                .map_err(|e| {
-                    // If the policy controller returned `NotFound`, indicating
-                    // that it doesn't have a policy for this addr, then we
-                    // can't gateway this address.
-                    if is_not_found(&e) {
-                        GatewayDomainInvalid.into()
-                    } else {
-                        e
-                    }
-                })
-                .instrument(tracing::debug_span!("policy"));
+                let policy = policies
+                    .get_policy(addr.into())
+                    .map_err(|e| {
+                        // If the policy controller returned `NotFound`, indicating
+                        // that it doesn't have a policy for this addr, then we
+                        // can't gateway this address.
+                        if is_not_found(&e) {
+                            GatewayDomainInvalid.into()
+                        } else {
+                            e
+                        }
+                    })
+                    .instrument(tracing::debug_span!("policy"));
 
-            future::Either::Right(future::try_join(profile, policy))
+                future::Either::Right(future::try_join(profile, policy))
             })
         };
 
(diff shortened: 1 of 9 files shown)


linkerd/app/gateway/src/server.rs
```diff
@@ -75,20 +76,49 @@ impl Gateway {
...
+                    return future::Either::Left(future::err(GatewayDomainInvalid.into()));
+                }
+
+                let profile = profiles
+                    .clone()
+                    .get_profile(profiles::LookupAddr(addr.clone().into()))
+                    .instrument(tracing::debug_span!("profiles"));
+
+                // TODO(eliza): we should probably also add the allowlist to
+                // policy resolution...
```
olix0r, line 96: This comment isn't relevant here. Is it? We just applied an allowlist above.

linkerd/app/gateway/src/server.rs
```diff
@@ -75,20 +76,49 @@ impl Gateway {
...
+                        // If the policy controller returned `NotFound`,
+                        // indicating that it doesn't have a policy for this
+                        // addr, then we can't gateway this address.
+                        Err(e) if is_not_found(&e) => return Err(GatewayDomainInvalid.into()),
+                        Err(e) => return Err(e),
+                    };
+
+                    Ok((profile?, policy))
+                });
+                future::Either::Right(f)
```
olix0r, line 114: Is this clearer as:

```suggestion
            let policy = policies
                .get_policy(addr.into())
                .map_err(|e| {
                    // If the policy controller returned `NotFound`, indicating
                    // that it doesn't have a policy for this addr, then we
                    // can't gateway this address.
                    if is_not_found(&e) {
                        GatewayDomainInvalid.into()
                    } else {
                        e
                    }
                })
                .instrument(tracing::debug_span!("policy"));

            future::Either::Right(future::try_join(profile, policy))
```

linkerd/app/gateway/src/server.rs
```diff
@@ -75,20 +76,49 @@ impl Gateway {
...
-            self.outbound
-                .with_stack(protocol)
-                .push_discover(svc::mk(move |GatewayAddr(addr)| {
-                    profiles.get_profile(profiles::LookupAddr(addr.into()))
-                }))
-                .into_stack()
+            use futures::future;
+
+            let allowlist = self.config.allow_discovery.clone();
+            svc::mk(move |GatewayAddr(addr)| {
```
olix0r, line 82: I'd generally prefer to split this into a function like
```rust
    fn discovery(
        &self,
        profiles: impl profiles::GetProfile,
        policies: impl outbound::policy::GetPolicy,
    ) -> impl svc::Service<
        GatewayAddr,
        Response = (
            Option<profiles::Receiver>,
            watch::Receiver<outbound::policy::ClientPolicy>,
        ),
        Error = Error,
        Future = impl Send + Unpin,
    > + Clone {
        let allowlist = self.config.allow_discovery.clone();
        svc::mk(move |GatewayAddr(addr)| {
          ...
        })
    }
```

So that we can more clearly delineate "discovery stuff" from stack building stuff

olix0r (commented):
looks good!

linkerd/app/gateway/src/server.rs
hawkw: oh, okay, i actually did that initially but then undid it because i wasn't sure if it was worth it. can put it back like that.

linkerd/app/gateway/src/server.rs
hawkw: yup, meant to remove this

olix0r (approved):

PR #2317, opened 2023-03-13
outbound: Eagerly cancel synthesized profile task

olix0r:
The outbound::discover module spawns a task to watch profiles for updates, but this task does not stop when the associated receiver is dropped.

This changes the task to exit when the policy channel is closed.

### DIFF
--- a/linkerd/app/outbound/src/discover.rs
+++ b/linkerd/app/outbound/src/discover.rs
@@ -159,9 +159,18 @@ fn spawn_synthesized_profile_policy(
     tokio::spawn(
         async move {
             loop {
-                if profile.changed().await.is_err() {
-                    tracing::debug!("Profile watch closed, terminating");
-                    return;
+                tokio::select! {
+                    biased;
+                    _ = tx.closed() => {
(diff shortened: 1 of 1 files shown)


linkerd/app/outbound/src/discover.rs
```diff
@@ -159,9 +159,18 @@ fn spawn_synthesized_profile_policy(
     tokio::spawn(
         async move {
             loop {
-                if profile.changed().await.is_err() {
-                    tracing::debug!("Profile watch closed, terminating");
-                    return;
+                tokio::select! {
+                    biased;
```
hawkw, line 163: does this actually need to be biased? AFAICT, it doesn't actually matter whether we poll the closed future or the profile rx first, as either way, we will end up handling the condition --- the important part is just that we are notified by either.

hawkw (approved):
looks good to me, although i don't think `biased` is actually needed here.

linkerd/app/outbound/src/discover.rs
olix0r: yeah that's right. we'll get it in post

PR #2324, opened 2023-03-15
Configure response classifiers from ClientPolicy routes

olix0r:
* 408457ad Carve out space in the policy API for a "failure policy"
* 2e165e05 classify: Replace SuccessOrFailure with Result
* 6af16e29 Update classifier to reuse clientpolicy response status/code types
* 60a61d96 Configure request classification from client policy routes

### DIFF
--- a/linkerd/app/core/src/classify.rs
+++ b/linkerd/app/core/src/classify.rs
@@ -1,6 +1,7 @@
 use crate::profiles;
 use linkerd_error::Error;
 use linkerd_http_classify as classify;
+use linkerd_proxy_client_policy as client_policy;
 use linkerd_proxy_http::{HasH2Reason, ResponseTimeoutError};
 use std::borrow::Cow;
 use tonic as grpc;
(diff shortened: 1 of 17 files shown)


linkerd/app/core/src/classify.rs
```diff
@@ -127,19 +123,36 @@ impl classify::ClassifyResponse for Response {
...
-            Response::Grpc => grpc_class(rsp.headers())
-                .map(|c| Eos::Grpc(GrpcEos::NoBody(c)))
-                .unwrap_or(Eos::Grpc(GrpcEos::Open)),
-            Response::Profile(ref classes) => Self::match_class(rsp, classes.as_ref())
-                .map(Eos::Profile)
+            Self::Http(statuses) => Eos::Class(Class::Http(if statuses.contains(status) {
+                Err(status)
+            } else {
+                Ok(status)
+            })),
```
hawkw, line 132: nit, take it or leave it: would it make sense for `http::StatusRanges` and `grpc::Codes` to have a function that takes a status/code and returns a `Result<Status>`/`Result<Code>`, rather than just a `bool`? we repeat the if statement here in a few places...

linkerd/app/outbound/src/http/logical/policy/route.rs
```diff
@@ -22,26 +22,35 @@ pub(crate) struct Matched<M, P> {
...
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
-pub(crate) struct Route<T, F> {
+pub(crate) struct Route<T, F, E> {
     pub(super) parent: T,
     pub(super) addr: Addr,
     pub(super) meta: Arc<policy::Meta>,
     pub(super) filters: Arc<[F]>,
     pub(super) distribution: BackendDistribution<T, F>,
+    pub(super) failure_policy: E,
```
hawkw, line 31: based on the naming --- is this just a response classifier, or is it intended to also eventually represent actions taken on failures (e.g. updating a failure counter, etc)?

hawkw (approved):
overall this is quite straightforward, and looks good to me!

PR #2333, opened 2023-03-15
gateway: synthesize ClientPolicies when the controller returns `NotFound`

hawkw:
Depends on #2343 

Both outbound and gateway proxies now resolve client policies from the
OutboundPolicies API. When the outbound proxy attempts to discover a
policy and the policy controller returns `NotFound`, it synthesizes a
default policy from the discovered ServiceProfile. However, when the
gateway proxy receives a `NotFound`, it will currently fail the
connection, based on the assumption that only valid cluster DNS names
are gatewayed (and not arbitrary IPs that might be forwards).
Unfortunately, this is not quite true. Gateway proxies may attempt to
discover cluster DNS names that are *Pod* DNS names, rather than Service
DNS names, and the policy controller will return `NotFound` for those
names.

This branch therefore changes the gateway proxy to also synthesize
default ClientPolicies based on the ServiceProfile when receiving a
`NotFound` status. Some of the code for synthesizing a client policy
from a ServiceProfile that's currently used in the outbound proxy was
factored out so that it could be reused here.


linkerd/app/gateway/src/discover.rs
```diff
@@ -0,0 +1,111 @@
...
+                                        addr,
+                                        meta,
+                                    ),
+                                    None => {
+                                        // TODO(eliza): should we synthesize a policy from the nameaddr?
+                                        tracing::debug!(
+                                            "Gateway ServiceProfile does not contain an endpoint"
+                                        );
+                                        ClientPolicy::invalid(detect_timeout)
+                                    }
```
hawkw, line 98: If the policy controller did not return a policy for the discovery target, it was probably a pod name, and therefore *should* be a forward...so it's *probably* fine to synthesize an invalid policy in the case where there's no forward endpoint, I think? Otherwise, I'm not really sure what we could do here that would be coherent...

linkerd/proxy/client-policy/src/lib.rs
```diff
@@ -157,6 +157,60 @@ impl ClientPolicy {
...
         }
     }
+
+    pub fn synthesize_forward(
+        meta: &Arc<Meta>,
+        timeout: time::Duration,
+        queue: Queue,
+        addr: SocketAddr,
+        metadata: EndpointMetadata,
+    ) -> Self {
```
olix0r, line 167: This feels like configuration to me, so I'd rather keep this in `app` and not in the library.

linkerd/app/gateway/src/discover.rs
olix0r, line 1: It's a bit hard to see how the logic changed in this diff... Is it possible to do a setup that moves the old logic first?

linkerd/app/outbound/src/lib.rs
```diff
@@ -32,7 +32,7 @@ use std::{
     time::Duration,
 };
 
-mod discover;
+pub mod discover;
```
olix0r, line 35: I'd prefer that we reexport the things we need than reexport the module.

```suggestion
mod discover;
```

linkerd/proxy/client-policy/src/lib.rs
olix0r: Specifically: I want to make sure that we centralize all of the logic that creates default routes that allow traffic. All of that code needs to go away, ultimately. It should be closer to where it's needed/used.

linkerd/app/gateway/src/discover.rs
hawkw: opened a setup PR for this: #2343

linkerd/proxy/client-policy/src/lib.rs
hawkw: would you prefer this in the `linkerd-app-outbound` crate, or should it go in a (new) `linkerd_app_core::client_policy` module (or similar)?

olix0r: i think it's fine for it to be where it was alongside the other synthesize function; and then re-export them both from outbound for now.

linkerd/app/outbound/src/discover.rs
```diff
@@ -273,6 +247,27 @@ fn synthesize_forward_policy(
...
+
+    let policy =
+        synthesize_forward_policy(&META, detect_timeout, queue, orig_dst, Default::default());
+    tracing::debug!(?policy, "Synthesizing policy");
+    let (tx, rx) = watch::channel(policy);
+    tokio::spawn(async move {
+        tx.closed().await;
+    });
+    rx
+}
```
hawkw, line 269: this just moved around a bit because i thought it was nicer for the public functions to come first in the module, but i can definitely put it back...

olix0r (approved):

PR #2341, opened 2023-03-16
reuse `sleep` future in `ConsecutiveFailureGate`

hawkw:
This is a suggestion for PR #2334. This change allows the `sleep` future
to be reused for every backoff without reallocating. The previous code
attempts to use the same allocation for sleeps using a
`ReusableBoxFuture`, but it doesn't actually achieve this, since when
the sleep is replaced with a differently-sized `pending` future, the
`ReusableBoxFuture` will reallocate. See here for details:
https://github.com/linkerd/linkerd2-proxy/pull/2334#discussion_r1139294331


olix0r:
thanks, this is great. i've pulled this into the update.

PR #2342, opened 2023-03-16
gate: Support admitting a limited number of requests

olix0r:
To implement circuit breakers, we want to use the gate to control whether or not a service is available to callers. When a service is coming out of a failed state, we may want to admit a limited number of requests through before going back to an unfettered state.

To support this, this change updates the `Gate` service to be backed by a `tokio::sync::watch` channel so that each side of the gate channel can communicate state updates.

`gate::Tx::closed` is renamed to `gate::Tx::lost` to avoid ambiguity.

### DIFF
--- a/linkerd/stack/src/gate.rs
+++ b/linkerd/stack/src/gate.rs
@@ -4,7 +4,7 @@ use std::{
     sync::Arc,
     task::{Context, Poll},
 };
-use tokio::sync::{watch, OwnedSemaphorePermit, Semaphore, TryAcquireError};
+use tokio::sync::{watch, OwnedSemaphorePermit, Semaphore};
 use tokio_util::sync::ReusableBoxFuture;
 use tracing::debug;
 
@@ -14,7 +14,7 @@ pub struct Gate<S> {
     rx: Rx,
     acquire: ReusableBoxFuture<'static, Option<OwnedSemaphorePermit>>,
     acquiring: bool,
-    permit: Poll<Permit>,
+    permit: Poll<Option<OwnedSemaphorePermit>>,
 }
 
 /// Observes gate state changes.
@@ -44,8 +44,6 @@ pub fn channel() -> (Tx, Rx) {
     let (tx, rx) = watch::channel(State::Open);
     (Tx(Arc::new(tx)), Rx(rx))
 }
-
-struct Permit(Option<OwnedSemaphorePermit>);
 
 // === impl Rx ===
 
@@ -80,15 +78,18 @@ impl Rx {
                 State::Open => return None,
                 State::Limited(sem) => match sem.acquire_owned().await {
                     Ok(permit) => return Some(permit),
-                    Err(_) => {
-                        debug!("Gate closed");
-                        return None;
+                    Err(_closed) => {
+                        // When the semaphore is closed, continue waiting for
+                        // the state to change.
+                        debug!("Semaphore closed");
                     }
                 },
                 State::Shut => {}
             }
 
             if let Err(e) = self.0.changed().await {
+                // If the `Tx` is dropped, then no further state changes can
+                // occur, so we simply remain in an unavilable state.
                 debug!("Gate closed: {:?}", e);
                 futures::future::pending::<()>().await;
             }
@@ -111,7 +112,7 @@ impl Tx {
         }
     }
 
-    /// Opens the gate.
+    /// Limits the gate with the provided semaphore.
     pub fn limit(&self, sem: Arc<Semaphore>) {
         if self.0.send(State::Limited(sem)).is_ok() {
             debug!("Gate limited");
@@ -165,15 +166,16 @@ where
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
         let permit = ready!(self.poll_acquire(cx));
         ready!(self.inner.poll_ready(cx))?;
-        self.permit = Poll::Ready(Permit(permit));
+        self.permit = Poll::Ready(permit);
         Poll::Ready(Ok(()))
     }
 
     #[inline]
     fn call(&mut self, req: Req) -> Self::Future {
-        self.permit = match std::mem::replace(&mut self.permit, Poll::Pending) {
+        match std::mem::replace(&mut self.permit, Poll::Pending) {
             Poll::Pending => panic!("poll_ready must be called first"),
-            Poll::Ready(..) => Poll::Pending,
+            Poll::Ready(Some(permit)) => permit.forget(),
+            Poll::Ready(None) => {}
         };
 
         self.inner.call(req)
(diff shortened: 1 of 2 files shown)


linkerd/stack/src/gate.rs
```diff
@@ -1,81 +1,96 @@
...
-        open: AtomicBool::new(true),
-        notify: Notify::new(),
-        closed: Notify::new(),
-    });
-    (Tx(shared.clone()), Rx(shared, Arc::new(())))
+    let (tx, rx) = watch::channel(State::Open);
+    (Tx(Arc::new(tx)), Rx(rx))
 }
 
+struct Permit(Option<OwnedSemaphorePermit>);
```
hawkw, line 47: CI build is failing because this is private but returned by `pub` functions
```suggestion
pub struct Permit(Option<OwnedSemaphorePermit>);
```

hawkw:
integration failure seems flaky, we should maybe do something about that: https://github.com/linkerd/linkerd2-proxy/actions/runs/4444150568/jobs/7802087501?pr=2342#step:8:333

linkerd/stack/src/gate.rs
```diff
@@ -1,81 +1,97 @@
...
+
 // === impl Rx ===
 
 impl Rx {
     /// Indicates whether the gate is open.
+    pub fn state(&self) -> State {
+        self.0.borrow().clone()
+    }
+
```
hawkw, line 57: tiny comment nit, i think the comment "Indicates whether the gate is open" is meant to be on `is_open`:
```suggestion
    pub fn state(&self) -> State {
        self.0.borrow().clone()
    }
    
    /// Indicates whether the gate is open.
```

linkerd/stack/src/gate.rs
```diff
@@ -84,27 +100,28 @@ impl Drop for Rx {
...
     /// Opens the gate.
     pub fn open(&self) {
-        if !self.0.open.swap(true, std::sync::atomic::Ordering::Release) {
+        if self.0.send(State::Open).is_ok() {
             debug!("Gate opened");
-            self.0.notify.notify_waiters();
+        }
+    }
+
+    /// Opens the gate.
```
hawkw, line 114: i don't think this comment is quite right --- `open` is the function that opens the gate.

linkerd/stack/src/gate.rs
```diff
@@ -145,117 +163,210 @@ where
...
+        self.permit = Poll::Ready(Permit(permit));
+        Poll::Ready(Ok(()))
+    }
+
+    #[inline]
+    fn call(&mut self, req: Req) -> Self::Future {
+        self.permit = match std::mem::replace(&mut self.permit, Poll::Pending) {
+            Poll::Pending => panic!("poll_ready must be called first"),
+            Poll::Ready(..) => Poll::Pending,
+        };
```
hawkw, line 177: not sure if i understand the match and reassignment to `self.permit` here --- the `mem::replace` call should already have set `self.permit` to `Poll::Pending`. isn't this just
```suggestion
        if std::mem::replace(&mut self.permit, Poll::Pending).is_pending() {
            panic!("poll_ready must be called first");
        }
```
or even
```suggestion
        let permit = std::mem::replace(&mut self.permit, Poll::Pending);
        assert!(permit.is_ready(), "poll_ready must be called first");
```

linkerd/stack/src/gate.rs
```diff
@@ -145,117 +163,210 @@ where
...
+    }
+}
+
+impl Drop for Permit {
+    fn drop(&mut self) {
+        // Permits are forgotten so the `Tx` controller can decide when to allow
+        // more requests.
+        if let Some(p) = self.0.take() {
+            p.forget();
+        }
```
hawkw, line 237: nit, TIOLI: this seems needlessly complex --- instead of having the default `Drop` behavior of a `Permit` be to forget the inner semaphore permit, and having a `Drop` impl for `Gate` that releases the permit if it's held, can't we just remove all the `Drop` impls and `Permit::relinquish`, keep the existing `Drop` behavior, and have `Gate::call` explicitly call `forget` on the inner semaphore permit? that would, AFAICT, accomplish the same semantics, but much simpler.

linkerd/stack/src/gate.rs
```diff
@@ -145,117 +163,210 @@ where
...
+            match self.rx.state() {
+                State::Open => return Poll::Ready(None),
+                State::Shut => {}
+                State::Limited(sem) => match sem.try_acquire_owned() {
+                    Ok(permit) => return Poll::Ready(Some(permit)),
+                    Err(TryAcquireError::NoPermits) => {}
+                    Err(TryAcquireError::Closed) => {
+                        tracing::warn!("Gate closed");
+                        return Poll::Pending;
+                    }
```
hawkw, line 198: AFAICT --- and I may be misreading, as this is a bit subtle --- it *looks* like this is a potential task leak. when the semaphore is closed (the other side of the gate is dropped), `poll_acquire` will return `Poll::Pending`  without ever having passed the `Context` to something that might wake it. If `poll_acquire` returns `Pending`, the `poll_ready` impl for `Gate` will also return `Pending` , and we've never registered our waker with something that might wake the task up again. I believe this will leak the task into the runtime as it will never terminate.

I think that if we're in the `Limited` state but the semaphore is dropped, we need to do one of two things, depending on what we want the semantics of dropping the semaphore to be:
1. return an error from `poll_acquire` that is then returned in the `poll_ready` impl for `Gate`, marking the service as permanently dead, or
2. don't return here, and continue to try re-starting the `acquire()` future, so that we get woken if the state inside the watch ever changes again.

The correct behavior depends on when the tx side will actually drop the semaphore while the state is limited.

linkerd/stack/src/gate.rs
```diff
@@ -1,81 +1,97 @@
...
 
-impl Drop for Rx {
-    fn drop(&mut self) {
-        let Rx(_, handle) = self;
-        if Arc::strong_count(handle) == 1 {
-            self.0.closed.notify_waiters();
+            if let Err(e) = self.0.changed().await {
+                debug!("Gate closed: {:?}", e);
+                futures::future::pending::<()>().await;
+            }
```
hawkw, line 94: if the `Tx` side of the gate is dropped, is it actually correct for this future to be pending forever? This will result in the `Gate` never opening, but it won't error out...should we have `poll_ready` return an error in this case?

hawkw (changes requested):
this is cool, and the approach overall makes sense. however, i have some concerns about what happens when the `Tx` side of the gate is dropped, and when the `Semaphore` created in the `Limited` state is dropped --- it seems like the behavior in those cases has the potential to create a task leak. it's possible i'm wrong about this, but if i am, it would be reassuring to see tests and/or comments that clarify why this isn't an issue.

linkerd/stack/src/gate.rs
olix0r: This could only be triggered if the other side explicitly called [`close](https://docs.rs/tokio/1.26.0/tokio/sync/struct.Semaphore.html#method.close). We probably should at least be explicit about how that case is handled.

I don't understand how this is a _task leak_, though. Whether this service is held as pending or not is entirely up to the holder. We have no idea why the holder of the `Tx` dropped the handle, but I would expect the gate to remain in whatever state it was in when the handle is lost.

olix0r: Also, just because the semaphore is closed doesn't indicate that the gate won't be updated.  So we should _probably_ go back to waiting for the rx state to change in this case.

linkerd/stack/src/gate.rs
olix0r: No, I don't think so.

linkerd/stack/src/gate.rs
hawkw: Yeah, I think going back to waiting for the state to change seems correct, because the semaphore closing doesn't mean the `Tx` itself has been dropped.

linkerd/stack/src/gate.rs
```diff
@@ -145,117 +165,183 @@ where
...
+    {
+        if !self.acquiring {
+            match self.rx.state() {
+                State::Open => return Poll::Ready(None),
+                State::Limited(sem) => {
+                    if let Ok(permit) = sem.try_acquire_owned() {
+                        return Poll::Ready(Some(permit));
+                    }
+                    // If the semaphore is closed or at capacity, wait until
+                    // something changes.
```
hawkw, line 199: :thumbsup: this looks right to me

hawkw (approved):
okay, this looks good to me --- thanks for clarifying the behavior around close conditions!

PR #2344, opened 2023-03-19
outbound: Fix incorrect l5d-proxy-connection logs

olix0r:
We've received reports of proxies inexplicibly emiting log lines including "Received unmeshed response with l5d-proxy-connection set". These messages may arise when the endpoint stack returns a synthesized response.

Furthermore, we have a note in the code explaining that this connection closure logic should not apply to load balanced requests, though it currently is.

This change fixes both of these issues:

* The HTTP error responding module now inserts a `Synthesized` marker to responses so that other modules can determine when a response was generated locally.
* The `proxy_connection_close` module is updated so that we ignore responses with this extension set.
* Furthermore, it does not apply the `Connection: close` behavior unless the new `Closable` marker extension is present.
* The marker is only set when using a forwarding backend.


linkerd/app/outbound/src/http/handle_proxy_error_headers.rs
```diff
@@ -0,0 +1,253 @@
...
+impl<N> NewHandleProxyErrorHeaders<(), N> {
+    pub fn layer() -> impl svc::layer::Layer<N, Service = Self> + Clone {
+        Self::layer_via(())
+    }
+}
+
+impl<T, X, N> svc::NewService<T> for NewHandleProxyErrorHeaders<X, N>
+where
+    X: svc::ExtractParam<Closable, T>,
+    X: svc::ExtractParam<tls::ConditionalClientTls, T>,
```
hawkw, line 61: do we really want to use the same `ExtractParam` impl to extract both of these? i realize this probably doesn't actually matter as it looks like in practice, the `ExtractParam` impl is currently always `()`, and this is just here for future-proofing...

linkerd/app/outbound/src/http/handle_proxy_error_headers.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    if rsp.version() == http::Version::HTTP_11 {
+        // If the response is HTTP/1.1, we need to send a Connection: close
+        // header to tell the application this connection is being closed.
+        rsp.headers_mut().insert(
+            http::header::CONNECTION,
+            http::HeaderValue::from_static("close"),
+        );
+    }
```
hawkw, line 157: nit, take it or leave it: can we do something like this?
```suggestion
    const CLOSE: http::HeaderValue = http::HeaderValue::from_static("close");
    
    if hdr != Some(CLOSE) {
        return false;
    }

    if rsp.version() == http::Version::HTTP_11 {
        // If the response is HTTP/1.1, we need to send a Connection: close
        // header to tell the application this connection is being closed.
        rsp.headers_mut().insert(
            http::header::CONNECTION,
            CLOSE,
        );
    }
```

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -129,6 +130,10 @@ impl<N> Outbound<N> {
                             metadata,
                             is_local,
                             parent: target.parent,
+                            // We don't close server-side connections when we
+                            // get `l5d-proxy-connect: close` response headers
```
hawkw, line 134: typo
```suggestion
                            // get `l5d-proxy-connection: close` response headers
```

hawkw (approved):
:+1:

linkerd/app/outbound/src/http/handle_proxy_error_headers.rs
olix0r: no, but we can use a static to achieve the same thing :)

linkerd/app/outbound/src/http/handle_proxy_error_headers.rs
olix0r: Yeah, we can change it later if it needs to be changed.

PR #2352, opened 2023-03-27
client-policy: accept backends with empty `dispatcher` fields

hawkw:
When a `backendRef` in an HTTPRoute references a valid Service name
which doesn't currently exist in the cluster, the policy controller
constructs a synthesized backend with a `failureInjector` filter, so
that traffic sent to the invalid backend will fail with a descriptive
error message, and any other backends in the route will still work
normally. Because the backend doesn't exist, there is no reasonable
value for the policy controller to set as the backend's `dispatcher`
field in the OutboundPolicy proto.

Unfortunately, however, the proxy currently rejects any OutboundPolicy
which has a backend without a populated `dispatcher` field. This means
that when an OutboundPolicy has one backend that has no `dispatcher`,
the proxy will synthesize an invalid policy that fails all traffic,
rather than one that only fails traffic sent to that backend.

This branch changes the proxy so that backends without a `dispatcher`
are no longer rejected by the client policy protobuf conversion code.
Instead, a new `Dispatcher::Null` variant is used to represent a backend
which does not exist, so that a functional `ClientPolicy` can still be
constructed when one or more backends lack a dispatcher. When
constructing a stack for a `Concrete` target with a `Null` dispatcher,
the proxy builds a service that fails all requests. In practice, these
errors will not be observed when the policy controller returns a backend
for an invalid Service name, as the `failureInjector` filter's error
should be produced first, but constructing a service is necessary for
the stack to function correctly.

Closes linkerd/linkerd2#10620

### DIFF
--- a/linkerd/app/outbound/src/http/concrete.rs
+++ b/linkerd/app/outbound/src/http/concrete.rs
@@ -23,6 +23,7 @@ use tracing::info_span;
 pub enum Dispatch {
     Balance(NameAddr, balance::EwmaConfig),
     Forward(Remote<ServerAddr>, Metadata),
+    Null,
 }
 
 /// Wraps errors encountered in this module.
@@ -166,6 +167,7 @@ impl<N> Outbound<N> {
                                     close_server_connection_on_remote_proxy_error: true,
                                 }
                             }),
+                            Dispatch::Null => todo!("eliza: add a second switch here :("),
                         })
                     },
                     forward.into_inner(),
--- a/linkerd/app/outbound/src/http/concrete.rs
+++ b/linkerd/app/outbound/src/http/concrete.rs
@@ -15,7 +15,7 @@ use linkerd_app_core::{
     transport::{self, addrs::*},
     Error, Infallible, NameAddr,
 };
-use std::{fmt::Debug, net::SocketAddr};
+use std::{fmt::Debug, net::SocketAddr, sync::Arc};
 use tracing::info_span;
 
 /// Parameter configuring dispatcher behavior.
@@ -23,13 +23,13 @@ use tracing::info_span;
 pub enum Dispatch {
     Balance(NameAddr, balance::EwmaConfig),
     Forward(Remote<ServerAddr>, Metadata),
-    Null,
+    Fail { message: Arc<str> },
 }
 
-/// A backend dispatcher was null.
+/// A backend dispatcher explicitly fails all requests.
 #[derive(Debug, thiserror::Error)]
-#[error("no backend dispatcher configured")]
-pub struct NullDispatcher(());
+#[error("{0}")]
+pub struct DispatcherFailed(Arc<str>);
 
 /// Wraps errors encountered in this module.
 #[derive(Debug, thiserror::Error)]
@@ -155,9 +155,10 @@ impl<N> Outbound<N> {
                 )
                 .instrument(|t: &Balance<T>| info_span!("balance", addr = %t.addr));
 
-            let null = svc::ArcNewService::new(|_: ()| {
-                svc::mk(|_| async move {
-                    Err::<http::Response<http::BoxBody>, NullDispatcher>(NullDispatcher(()))
+            let fail = svc::ArcNewService::new(|message: Arc<str>| {
+                svc::mk(move |_| {
+                    let error = DispatcherFailed(message.clone());
+                    futures::future::ready(Err(error))
                 })
             });
             balance
@@ -178,10 +179,10 @@ impl<N> Outbound<N> {
                                     close_server_connection_on_remote_proxy_error: true,
                                 }
                             })),
-                            Dispatch::Null => svc::Either::B(()),
+                            Dispatch::Fail { message } => svc::Either::B(message),
                         })
                     },
-                    null,
+                    fail,
                 )
                 .push(svc::NewQueue::layer_via(config.http_request_queue))
                 .push(svc::ArcNewService::layer())
--- a/linkerd/app/outbound/src/http/concrete.rs
+++ b/linkerd/app/outbound/src/http/concrete.rs
@@ -26,6 +26,11 @@ pub enum Dispatch {
     Null,
 }
 
+/// A backend dispatcher was null.
+#[derive(Debug, thiserror::Error)]
+#[error("no backend dispatcher configured")]
+pub struct NullDispatcher(());
+
 /// Wraps errors encountered in this module.
 #[derive(Debug, thiserror::Error)]
 #[error("concrete service {addr}: {source}")]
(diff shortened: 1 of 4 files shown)


linkerd/app/outbound/src/http/logical/policy/router.rs
```diff
@@ -145,6 +145,7 @@ where
             policy::BackendDispatcher::Forward(addr, ref metadata) => mk_concrete(
                 concrete::Dispatch::Forward(Remote(ServerAddr(addr)), metadata.clone()),
             ),
+            policy::BackendDispatcher::Null => mk_concrete(concrete::Dispatch::Null),
```
mateiidavid, line 148: Do you think it's worth mentioning that building a concrete target is required for the stack to work correctly but that these requests will fail before? Ah, might be a lot of information to just put in a comment 🤷🏻

mateiidavid (approved):
LGTM. The idea sounds great conceptually too. Bit unfortunate that we still need to treat it as a Concrete address, I guess? Would it matter if the route updated later on to contain a valid backend? I guess the stacks here would just be torn down.

hawkw:
> Would it matter if the route updated later on to contain a valid backend? I guess the stacks here would just be torn down.

Yeah, that's correct. All of this would be re-built if the policy changed.

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -23,8 +23,14 @@ use tracing::info_span;
 pub enum Dispatch {
     Balance(NameAddr, balance::EwmaConfig),
     Forward(Remote<ServerAddr>, Metadata),
+    Null,
 }
 
+/// A backend dispatcher was null.
+#[derive(Debug, thiserror::Error)]
+#[error("no backend dispatcher configured")]
+pub struct NullDispatcher(());
```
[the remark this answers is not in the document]
hawkw: It could be. The inner `()` field prevents it from being constructed outside this module. Which, in this case, doesn't really matter --- it's just a bit of a habit from what I would do if this was a public library API.

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -149,14 +155,20 @@ impl<N> Outbound<N> {
                 )
                 .instrument(|t: &Balance<T>| info_span!("balance", addr = %t.addr));
 
+            let null = svc::ArcNewService::new(|_: ()| {
+                svc::mk(|_| async move {
+                    Err::<http::Response<http::BoxBody>, NullDispatcher>(NullDispatcher(()))
+                })
+            });
```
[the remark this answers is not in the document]
hawkw: Yup, that's right! This pattern occurs a couple other places in the proxy where we need a switch with three possible branches.

olix0r, line 162:
```suggestion
            let null = svc::Fail::<http::Response<http::BoxBody>, NullDispatcher>::default();
```

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -23,8 +23,14 @@ use tracing::info_span;
 pub enum Dispatch {
     Balance(NameAddr, balance::EwmaConfig),
     Forward(Remote<ServerAddr>, Metadata),
+    Null,
 }
 
+/// A backend dispatcher was null.
+#[derive(Debug, thiserror::Error)]
```
olix0r, line 30:
```suggestion
#[derive(Debug, Default, thiserror::Error)]
```

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -23,8 +23,14 @@ use tracing::info_span;
...
+#[error("no backend dispatcher configured")]
```
olix0r, line 31: Is this error sufficiently informative when it shows up in logs and users ask about it? "backend dispatchers" are a bit of an internal concept.

Can this explanation be provided by the proxy alone? Or should the control plane provide this explanation?

olix0r (changes requested):
In general, _null_ is a little vague. For instance, I would expect a null service to never become ready, not to produce a specific error type. It seems to me that what we're _really_ trying to express is that we want all requests to this service to fail.

Would this be more clearly expressed by something like `Dispatch::Fail { message: String }`? or even a type that holds an error type that is capable of configuring response codes, etc?

My main concern with null is that leaves us in a place where we can only communicate somewhat obtuse error messages to users. We should figure out a way so that we can be more specific about what specific error we're dealing with.

OR we should simply have this backend never become ready, but I suspect that won't work well for us.

linkerd/app/outbound/src/http/concrete.rs
hawkw: > Would this be more clearly expressed by something like `Dispatch::Fail { message: String }`? or even a type that holds an error type that is capable of configuring response codes, etc?

I also didn't like how the approach in this PR results in two different error paths, only one of which is descriptive, and also considered something like what you're describing. The reasoning behind the current approach is that the controller is currently communicating the desired error to the proxy using a `failureInjector` filter on the backend. The failure injector filter should inject the nicer error  before we actually hit the explicitly-failing backend stack, so the more descriptive error _is_ the one observed by the user, but the presence of the second error path definitely does feel unfortunate.

If we wanted to have an explicitly failing backend that fails with the error configured by the control plane (rather than an explicitly failing backend with a generic error message that _also_ has a `failureInjector` filter), it seems like there's one of two approaches:
1.  Special-case empty backends with `failureInjector` filters to return a `Dispatch::Fail` variant with the message and status code from the failure injector filter. This would avoid the use of the null backend with a generic error message, but it felt a little weird to synthesize something in the proxy that wasn't quite the same as the policy from the controller.
2. Introduce a `Dispatch::Fail`-like variant to the proxy API as well, and have the policy controller send that in this case, rather than a backend with a `None` dispatcher and a `failureInjector` filter. That would, in many ways, be the cleanest solution (IMO) but also adds more stuff to the proxy API to represent a configuration that _can_ be represented (although maybe less cleanly) with the existing API.

I went with the approach in this branch because it felt like the least intrusive solution, but would be happy to try one of the other approaches I discussed here if you think it would be better! I don't love this approach either...

olix0r: But the null backend fails with an error, right? Are these approaches really mutually exclusive?

My main argument is: I think it's surprising for something called a null backend to poll ready at all. If this backend is really unreachable, let's call it Unreachable and make it panic; or if it's going to return some sort of error, let's make that behavior explicit and part of how the backend is actually configured.

I don't much care whether filters or used or not. I just don't really like the semantics around the null backend.

hawkw: Oh, okay, got it.

olix0r (approved):

PR #2353, opened 2023-03-27
add `NewClassifyGate` and `BroadcastClassification` middleware

hawkw:
This branch introduces middleware for opening and closing a `Gate` middleware based on HTTP response classification. The `BroadcastClassification` service applies an HTTP response classifier to each request, and sends the classification over a channel. The `NewClassifyGate` middleware constructs stacks containing `BroadcastClassification` and `Gate` middleware. In conjunction with a _failure accrual policy_, these middleware allow response classification results to determine whether a service stack is available, implementing _request-level HTTP circuit breaking_.

Failure accrual policies are determined by the `ExtractParam` implementation provided to the `NewClassifyGate` layer. This type must provide a `classify::gate::Params` type containing a `mpsc::Sender` for the response classification channel, and a `gate::Rx` that controls the opening and closing of the `Gate` middleware. This means that the `ExtractParam` implementation can spawn a task implementing an arbitrary policy for opening and closing the gate based on classifications received over the `mpsc::Receiver` end of the classification channel, based on the target type the stack is constructed for.

This branch does not implement any failure accrual policies, nor does it integrate the `NewClassifyGate` middleware with existing proxy stacks. It simply adds these middleware. Subsequent branches will add the circuit-breaking middleware to the proxy stack, implement failure accrual policies, and configure failure accrual from individual stack targets.

This change was factored out from PR #2334 in order to merge independently.

### DIFF
--- a/linkerd/proxy/http/src/classify/channel.rs
+++ b/linkerd/proxy/http/src/classify/channel.rs
@@ -1,0 +1,335 @@
+use super::{ClassifyEos, ClassifyResponse};
+use futures::{prelude::*, ready};
+use linkerd_error::Error;
+use linkerd_stack::{layer, ExtractParam, NewService, Service};
+use pin_project::{pin_project, pinned_drop};
+use std::{
+    fmt::Debug,
+    future::Future,
+    marker::PhantomData,
+    pin::Pin,
+    task::{Context, Poll},
+};
+use tokio::sync::mpsc;
+
+#[derive(Debug)]
+pub struct NewBroadcastClassification<C, X, N> {
+    inner: N,
+    extract: X,
+    _marker: PhantomData<fn() -> C>,
+}
+
+/// A HTTP `Service` that applies a [`super::Classify`] to each request and
+/// broadcasts the result to an [`Rx`] inserted to inner requests and outer
+/// responses..
+#[derive(Debug)]
+pub struct BroadcastClassification<C: ClassifyResponse, S> {
+    inner: S,
+    tx: mpsc::Sender<C::Class>,
+    _marker: PhantomData<fn() -> C>,
+}
+
+#[derive(Clone, Debug)]
+pub struct Tx<C>(pub mpsc::Sender<C>);
+
+#[pin_project]
+pub struct ResponseFuture<C: ClassifyResponse, B, F> {
+    #[pin]
+    inner: F,
+    state: Option<State<C, C::Class>>,
+    _marker: PhantomData<fn() -> B>,
+}
+
+#[pin_project(PinnedDrop)]
+pub struct ResponseBody<C: ClassifyEos, B> {
+    #[pin]
+    inner: B,
+    state: Option<State<C, C::Class>>,
+}
+
+#[derive(Debug)]
+struct State<C, T> {
+    classify: C,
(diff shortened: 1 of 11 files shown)


linkerd/proxy/http/src/classify/channel.rs
```diff
@@ -0,0 +1,335 @@
...
+#[derive(Debug)]
+pub struct NewBroadcastClassification<C, X, N> {
+    inner: N,
+    extract: X,
+    _marker: PhantomData<fn() -> C>,
+}
+
+/// A HTTP `Service` that applies a [`super::Classify`] to each request and
+/// broadcasts the result to an [`Rx`] inserted to inner requests and outer
+/// responses..
```
olix0r, line 24: I know I wrote this but it makes absolutely no sense.

linkerd/proxy/http/src/classify/gate.rs
```diff
@@ -0,0 +1,136 @@
+// A `NewService` called `NewClassifyGate` that builds a
+// `BroadcastClassification` and a `Gate` so that the `gate::Tx` can be
+// controlled by the `BroadcastClassification`'s `Tx`.
```
olix0r, line 3: Similarly, this comment doesn't make much sense here. I don't think it's correct if moved, either.

hawkw:
@olix0r I'll try to clean up the comments — I just `git checkout`ed the changes from your branch.

olix0r (commented):
* Let's either remove incorrect comments or update them so they are clear.
* Can we instrument the breaker layers in the stack such that they do nothing meaningful until configuration is provided later? Otherwise, that should be the next step.

hawkw:
> * Let's either remove incorrect comments or update them so they are clear.

Will do! 👍

> * Can we instrument the breaker layers in the stack configured such that they do nothing meaningful until configuration is provided later? Otherwise, that should be the next step.

My plan was to do that in my next branch https://github.com/linkerd/linkerd2-proxy/pull/2354, which also adds target type params. We could just add the layer in this PR if you'd prefer it.

olix0r:
@hawkw As I said in the issue

> Since we're going to have to support the case where we do not do any circuit breaking, there's no need to merge in the consecutive failures changes in the PR that adds circuit breaking to the stack. We can review "add circuit breaking" discretely from "configure this specific circuit breaking policy."

I expected that we'd add the layers to stack in one branch, and then worry about the specifics of consecutivefailures/whatever later. That can be three branches if you want, but we should decouple _change the stack_ from _add a consecutive failures breaker_

hawkw:
Yeah, the plan for #2354 is to add only a `FailureAccrual::None` policy that configures a no-op breaker layer, and add consecutive failures (and potentially other policies) in subsequent branches. Is that what you expected? I'm happy to add a no-op breaker in this branch instead, and have #2354 just add the param type with a no-circuit-breaking variant, though, if you'd prefer?

olix0r:
@hawkw OK gotcha that plan sounds fine.

hawkw:
@olix0r I took a crack at improving the docs a bit, trying to highlight some of the stuff that was confusing to me when reviewing #2334. PLMKWYT?

olix0r (approved):

PR #2357, opened 2023-03-28
add `ConsecutiveFailures` failure accrual policy

hawkw:
Depends on https://github.com/linkerd/linkerd2-proxy/pull/2354.

This branch introduces a `ConsecutiveFailures` failure accrual policy
which marks an endpoint as unavailable if a given number of failures
occur in a row without any successes. Once the endpoint is marked as
failing, it is issued a single probe request after a delay. If the probe
succeeds, the endpoint transitions back to the available state.
Otherwise, it remains unavailable, with subsequent probe requests being
issued with an exponential backoff. The consecutive failures failure
accrual policy was initially implemented in PR #2334 by @olix0r.

A new `FailureAccrual::ConsecutiveFailures` variant is added in
`linkerd2-proxy-client-policy` for configuring the consecutive failures
failure accrual policy. The construction of circuit breakers in the
outbound stack is changed from a closure implementing `ExtractParam` to
a new `breaker::Params` type, so that the type returned can be the same
regardless of which failure accrual policy is constructed.

### DIFF
--- a/linkerd/app/outbound/src/http/breaker.rs
+++ b/linkerd/app/outbound/src/http/breaker.rs
@@ -27,9 +27,13 @@ impl<T> svc::ExtractParam<gate::Params<classify::Class>, T> for Params {
                 max_failures,
                 backoff,
             } => {
-                tracing::trace!(max_failures, backoff = ?backoff, "Using consecutive-failures failure accrual policy.");
+                tracing::trace!(
+                    max_failures,
+                    backoff = ?backoff,
+                    "Using consecutive-failures failure accrual policy.",
+                );
 
-                // 1. If the configured number of consecutive failures are encountered,'
+                // 1. If the configured number of consecutive failures are encountered,
                 //    shut the gate.
                 // 2. After an ejection timeout, open the gate so that 1 request can be processed.
                 // 3. If that request succeeds, open the gate. If it fails, increase the
--- a/linkerd/app/outbound/src/http/breaker.rs
+++ b/linkerd/app/outbound/src/http/breaker.rs
@@ -1,4 +1,5 @@
-pub mod consecutive_failures;
+mod consecutive_failures;
+use consecutive_failures::ConsecutiveFailures;
 use linkerd_app_core::{classify, proxy::http::classify::gate, svc};
 use linkerd_proxy_client_policy::FailureAccrual;
 
@@ -28,20 +29,12 @@ impl<T> svc::ExtractParam<gate::Params<classify::Class>, T> for Params {
             } => {
                 tracing::trace!(max_failures, backoff = ?backoff, "Using consecutive-failures failure accrual policy.");
 
-                // 1. If the configured number of consecutive failures are encountered,
+                // 1. If the configured number of consecutive failures are encountered,'
                 //    shut the gate.
                 // 2. After an ejection timeout, open the gate so that 1 request can be processed.
                 // 3. If that request succeeds, open the gate. If it fails, increase the
                 //    ejection timeout and repeat.
-                let breaker = consecutive_failures::ConsecutiveFailures::new(
-                    consecutive_failures::Params {
-                        max_failures,
-                        backoff,
-                        channel_capacity: self.channel_capacity,
-                    },
-                    gate,
-                    rsps,
-                );
+                let breaker = ConsecutiveFailures::new(max_failures, backoff, gate, rsps);
                 tokio::spawn(breaker.run());
 
                 prms
--- a/linkerd/app/outbound/src/http/breaker.rs
+++ b/linkerd/app/outbound/src/http/breaker.rs
@@ -2,6 +2,7 @@ mod consecutive_failures;
 use consecutive_failures::ConsecutiveFailures;
 use linkerd_app_core::{classify, proxy::http::classify::gate, svc};
 use linkerd_proxy_client_policy::FailureAccrual;
+use tracing::{trace_span, Instrument};
 
 /// Params configuring a circuit breaker stack.
 #[derive(Copy, Clone, Debug)]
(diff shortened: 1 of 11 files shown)


linkerd/app/outbound/src/http/breaker/consecutive_failures.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    /// Wait for a response to determine whether the breaker should be opened.
+    async fn probation(&mut self) -> Result<classify::Class, ()> {
+        tracing::debug!("Probation");
+        self.semaphore.add_permits(1);
+        self.gate.limit(self.semaphore.clone());
+        tokio::select! {
+            rsp = self.rsps.recv() => rsp.ok_or(()),
+            _ = self.gate.lost() => Err(()),
+        }
```
olix0r, line 102: I don't think this is a blocker (and I'm not really sure how to resolve it off the top of my head), but the current classifiers are going to behave poorly with streaming responses:

If we're in probation and the first request we allow receives a streaming response, we won't process the response summary until the end of the stream! So if that stream stays open for 10 minutes, we won't exit probation for 10 minutes. And if the streaming response ends with an error code, then we'll go back into shut after a 10 minute delay!

olix0r: I suppose we could impose a max probation time limit: if we don't exit within that period, just open the gate.

linkerd/app/outbound/src/http/breaker.rs
```diff
@@ -0,0 +1,48 @@
...
+                    "Using consecutive-failures failure accrual policy.",
+                );
+
+                // 1. If the configured number of consecutive failures are encountered,
+                //    shut the gate.
+                // 2. After an ejection timeout, open the gate so that 1 request can be processed.
+                // 3. If that request succeeds, open the gate. If it fails, increase the
+                //    ejection timeout and repeat.
+                let breaker = ConsecutiveFailures::new(max_failures, backoff, gate, rsps);
+                tokio::spawn(breaker.run());
```
olix0r, line 42: Should this include the current tracing context?

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -158,7 +145,13 @@ impl<N> Outbound<N> {
                 .push_on_service(svc::MapErr::layer_boxed())
                 .lift_new_with_target()
                 .check_new_new_service::<Balance<T>, (SocketAddr, Metadata), http::Request<http::BoxBody>>()
-                .push(http::NewClassifyGateSet::<classify::Response, _, _, _>::layer_via(mk_breaker))
+                .push(http::NewClassifyGateSet::<classify::Response, _, _, _>::layer_via(|target: &Balance<T>| {
+                    breaker::Params {
+                        accrual: target.parent.param(),
+                        // TODO(eliza): should this reuse an existing queue configuration?
+                        channel_capacity: 100,
```
olix0r, line 152: What questions do you need to answer to resolve this TODO?

linkerd/stack/src/gate.rs
```diff
@@ -72,7 +72,7 @@ impl Rx {
     }
 
     /// Waits for the gate state to change to be open.
-    async fn acquire(&mut self) -> Option<OwnedSemaphorePermit> {
+    pub async fn acquire(&mut self) -> Option<OwnedSemaphorePermit> {
```
olix0r, line 75: It's a shame this needs to be `pub` just for tests. Do we have a `test-util` feature in this crate already? It would be ideal to expose this only as a test-only utility (e.g. `pub async fn acquire_for_test` which wraps non-pub `async fun acquire`).

linkerd/app/outbound/src/http/breaker/consecutive_failures.rs
olix0r: @hawkw It seems reasonable to use the max backoff value as the time limit...

linkerd/app/outbound/src/http/breaker.rs
hawkw: yup, i meant to stick a span on it and i forgot --- good catch

linkerd/app/outbound/src/http/breaker/consecutive_failures.rs
hawkw: Yeah, that makes sense --- I'll add a probation limit. Would you prefer it in this PR or a follow-up?

linkerd/app/outbound/src/http/concrete.rs
hawkw: Well, my thought process was that IMO it's probably not something that should have its own config knob, as it's kind of an internal implementation detail...but we have to have *some* kind of number for it. We already have a configuration value for channel capacity, but I wasn't totally sure if it made sense to reuse that here. I guess it is a channel of requests (sort of --- it's a channel of classifications produced from the responses to requests, which should be <= the number of requests), so it probably does make sense to use that bound here as well. But, it felt a bit weird, so I wanted to get a second opinion.

linkerd/stack/src/gate.rs
hawkw: Can do.

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -163,7 +149,20 @@ impl<N> Outbound<N> {
...
-                    http::NewClassifyGateSet::<classify::Response, _, _, _>::layer_via(mk_breaker),
+                    http::NewClassifyGateSet::<classify::Response, _, _, _>::layer_via({
+                        // This sets the capacity of the channel used to send
+                        // response classifications to the failure accrual task.
+                        // Since the number of messages in this channel should
+                        // be roughly the same as the number of requests, size
+                        // it to the request queue capacity.
+                        // TODO(eliza): when queue capacity is configured
+                        // per-target, extract this from the target instead.
+                        let channel_capacity = config.http_request_queue.capacity;
```
olix0r, line 160: Don't we have a queue config in the balance target?

hawkw: lemme check, i thought we hadn't yet...

hawkw: took a look, the `queue::Capacity` param isn't wired up with any of the target types yet.

olix0r (approved):

PR #2364, opened 2023-03-29
configure failure accrual from the `OutboundPolicy` API

hawkw:
This branch updates the proxy to configure failure accrual policies from
the OutboundPolicy API. It updates the dependency on
`linkerd2-proxy-api` to include changes from
linkerd/linkerd2-proxy-api#223, which adds failure accrual configuration
to the `Protocol::Http1`, `Protocol::Http2`, and `Protocol::Grpc`
variants.

Currently, `ConsecutiveFailures` (added in #2357) is the only supported
failure accrual policy. If the proxy API does not configure a failure
accrual policy, failures are not accrued, and endpoints do not become
unavailable due to failure accrual.

### DIFF
--- a/linkerd/proxy/client-policy/src/lib.rs
+++ b/linkerd/proxy/client-policy/src/lib.rs
@@ -618,7 +618,6 @@ pub mod proto {
                     // TODO(eliza): if other failure accrual kinds are added
                     // that also use exponential backoffs, this could be factored out...
                     let outbound::ExponentialBackoff { min_backoff, max_backoff, jitter_ratio } =
-                        // TODO(eliza): should this be defaulted?
                         backoff.ok_or(InvalidFailureAccrual::Missing("consecutive failures backoff"))?;
                     let min = min_backoff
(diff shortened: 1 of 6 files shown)


linkerd/proxy/client-policy/src/lib.rs
```diff
@@ -590,4 +604,56 @@ pub mod proto {
...
+            let kind = accrual.kind.ok_or(InvalidFailureAccrual::Missing("kind"))?;
+            match kind {
+                failure_accrual::Kind::ConsecutiveFailures(ConsecutiveFailures {
+                    max_failures,
+                    backoff,
+                }) => {
+                    // TODO(eliza): if other failure accrual kinds are added
+                    // that also use exponential backoffs, this could be factored out...
+                    let outbound::ExponentialBackoff { min_backoff, max_backoff, jitter_ratio } =
+                        // TODO(eliza): should this be defaulted?
```
hawkw, line 621: i think it's the controller's responsibility to provide a default, if it wants us to use failure accrual...
```suggestion
```

olix0r (approved):

PR #2371, opened 2023-03-31
outbound: add failure accrual test, fix `Gate`s not becoming ready

hawkw:
This branch adds a stack test in `linkerd_app_outbound` for the
consecutive-failures failure accrual policy. This test builds the HTTP
logical and concrete stacks and tests their behavior using a mocked
`tower::Service` as the endpoint stack, and a mock stack target type and
destination resolver. Because these stack tests do not perform IO,
execute in a single thread, and use the paused Tokio timer, they should
be fully deterministic, minimizing potential flaky failures. However, we
are still testing enough of the stack to exercize the circuit-breaking
functionality integrated with other components of the proxy.

The test added in this branch exercises only the behavior of a single
concrete stack (one backend). It tests that the backend's circuit
breaker trips when the configured number of failures occur, that the
backend is probed after a probation backoff, and the behavior of failed
and successful probe requests. This test does *not* currently exercise
the behavior of distributors/load balancers when only some endpoints are
failing. Those tests can be added in a follow-up PR.

This test revealed a bug in the `Gate` middleware that prevented the
test from passing. The `Gate` middleware's `poll_ready` implementation
attempts to acquire a semaphore permit when the gate is in a limited
state, and then polls the inner service for readiness. If the inner
service is ready, the semaphore permit is stored in the gate to be
consumed in `call`. However, this implementation does not check if the
gate has *already* acquired a sempahore permit and stored it in the
gate. In that case, a subsequent call to `poll_ready` after the gate has
returned `Poll::Ready` will result in it attempting to acquire another
permit. If the semaphore only has a single permit, and the gate has
already acquired it, it will never become ready if `poll_ready` is
called a second time.

This branch fixes this bug by changing the `Gate` to only call
`self.poll_acquire()` if the gate has not acquired a permit. If it _has_
acquired a permit, the permit is temporarily moved back out of the gate
while polling the inner service, in case that service has become
unready. This way, the gate will not attempt to acquire a second permit
when it already has one, and will release the permit if the inner
service is no longer ready.

If it's preferable, I can pull the `Gate` fix out into a separate
branch, but wanted to include it with the tests so that previous commits
can be checked out to reproduce the failure.

### DIFF
--- a/linkerd/stack/src/gate.rs
+++ b/linkerd/stack/src/gate.rs
@@ -173,8 +173,15 @@ where
     type Future = S::Future;
 
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        let permit = ready!(self.poll_acquire(cx));
-        ready!(self.inner.poll_ready(cx))?;
+        // If we have previously acquired a permit, move it out of the `Gate` in
+        // order to poll the inner service. If the inner service is no longer
+        // ready, we will drop our previously acquired permit.
+        let permit = match std::mem::replace(&mut self.permit, Poll::Pending) {
+            Poll::Ready(permit) => permit,
+            Poll::Pending => ready!(self.poll_acquire(cx)),
+        };
+
+        ready!(self.inner.poll_ready(cx)?);
         self.permit = Poll::Ready(permit);
         Poll::Ready(Ok(()))
     }
--- a/linkerd/stack/src/gate.rs
+++ b/linkerd/stack/src/gate.rs
@@ -173,15 +173,11 @@ where
     type Future = S::Future;
 
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        // If we have previously acquired a permit, move it out of the `Gate` in
-        // order to poll the inner service. If the inner service is no longer
-        // ready, we will drop our previously acquired permit.
-        let permit = match std::mem::replace(&mut self.permit, Poll::Pending) {
-            Poll::Ready(permit) => permit,
-            Poll::Pending => ready!(self.poll_acquire(cx)),
-        };
-
-        ready!(self.inner.poll_ready(cx)?);
+        if self.permit.is_ready() {
+            return Poll::Ready(Ok(()));
+        }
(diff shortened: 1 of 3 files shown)


linkerd/stack/src/gate.rs
```diff
@@ -173,8 +173,15 @@ where
...
-        ready!(self.inner.poll_ready(cx))?;
+        // If we have previously acquired a permit, move it out of the `Gate` in
+        // order to poll the inner service. If the inner service is no longer
+        // ready, we will drop our previously acquired permit.
+        let permit = match std::mem::replace(&mut self.permit, Poll::Pending) {
+            Poll::Ready(permit) => permit,
+            Poll::Pending => ready!(self.poll_acquire(cx)),
+        };
+
+        ready!(self.inner.poll_ready(cx)?);
```
olix0r, line 184:
```suggestion
        if self.permit.is_ready() {
            return Poll::Ready(Ok(()));
        }
        let permit = ready!(self.poll_acquire(cx));
        ready!(self.inner.poll_ready(cx))?;
```

This seems like a simpler fix. We only ever store the permit when the inner service has polled ready. There's no need to poll it again, is there?

hawkw: That was actually what I had originally written, and it does also fix the problem. I thought that it was potentially desirable to continue polling the inner service in case its `poll_ready` implementation actually does work (e.g. if the gate is around a load balancer which needs to be polled to drive discovery). I'm not sure if this actually matters, though.

olix0r: Or: is there actually a situation where a service that has polled ready can later return pending? My understanding is that this violates the poll contract.

hawkw: You're correct that it should not return pending unless the readiness is consumed. I was just concerned about cases where `poll_ready` drives _work_ (e.g. polling discovery), but I think your simpler solution is fine here.

olix0r (approved):

linkerd/stack/src/gate.rs
olix0r: I was wrong. Balancers rely on reconnect being able to go from Ready => Pending.

PR #2372, opened 2023-04-01
outbound: Report per-route-backend request count metrics

olix0r:
When performing policy-based routing, proxies may dispatch requests through per-route backend configurations. In order to illustrate how routing rules apply and how backend distributions are being honored, this change adds two new metrics:

* `outbound_http_route_backend_requests_total`; and
* `outbound_grpc_route_backend_requests_total`

Each of these metrics includes labels that identify a routes parent (i.e. a Service), the route resource being used, and the backend resource being used.

This implementation does NOT implement any form of metrics eviction for these new metrics. This is tolerable for the short term as the cardinality of services and routes is generally much less than the cardinality of individual endpoints (where we do require timeout/eviction for metrics).

### DIFF
--- a/linkerd/app/outbound/src/http/logical/policy/route/backend/metrics.rs
+++ b/linkerd/app/outbound/src/http/logical/policy/route/backend/metrics.rs
@@ -1,0 +1,86 @@
+use ahash::AHashMap;
+use linkerd_app_core::metrics::{metrics, Counter, FmtLabels, FmtMetrics};
+use linkerd_proxy_client_policy as policy;
+use parking_lot::Mutex;
+use std::{fmt::Write, sync::Arc};
+
+metrics! {
+    outbound_http_route_backend_requests_total: Counter {
+        "The total number of outbound requests dispatched to a HTTP route backend"
+    },
+    outbound_grpc_route_backend_requests_total: Counter {
+        "The total number of outbound requests dispatched to a gRPC route backend"
+    }
+}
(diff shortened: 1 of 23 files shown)


linkerd/app/outbound/src/http/logical/policy/route/backend/metrics.rs
```diff
@@ -0,0 +1,86 @@
...
+impl RouteBackendMeta {
+    fn write_meta(
+        scope: &str,
+        meta: &policy::Meta,
+        f: &mut std::fmt::Formatter<'_>,
+    ) -> std::fmt::Result {
+        write!(f, "{scope}_group=\"{}\"", meta.group())?;
+        write!(f, ",{scope}_kind=\"{}\"", meta.kind())?;
+        write!(f, ",{scope}_namespace=\"{}\"", meta.namespace())?;
+        write!(f, ",{scope}_name=\"{}\"", meta.name())?;
```
olix0r, line 83: I don’t think these labels are sufficient.

Without port information, we can’t differentiate between ports on a service.

Should port information be reflected through `meta.section()`? Or do we need a separate field to reflect ports?

How are ports handled in the Gateway API?

olix0r: Gateway API parent references may include either/both `sectionName` and `port` (number) fields.

PR #2373, opened 2023-04-01
outbound: Emit INFO-level logs on failure accrual changes

olix0r:
There are no explicit indications of failure accrual related behavior.

This change adds INFO-level logs when a failure accrual breaker is
tripped or reeopened. These log messages are scoped within the
'endpoint' log context so that log lines include endpoint and balancer
addresses.

    INFO balance{addr=logical.test.svc.cluster.local:666}:endpoint{addr=192.0.2.41:666}:consecutive_failures: linkerd_app_outbound::http::breaker::consecutive_failures: Consecutive failure-accrual breaker closed
    INFO balance{addr=logical.test.svc.cluster.local:666}:endpoint{addr=192.0.2.41:666}:consecutive_failures: linkerd_app_outbound::http::breaker::consecutive_failures: Consecutive failure-accrual breaker reopened

PR #2374, opened 2023-04-02
outbound: Configure balancers with service metadata 

olix0r:
The OutboundPolicies API includes resource references with its
responses. These references allow us to accurately identify an arbitrary
(Kubernetes style) resource by its group, kind, namespace, and name.

To support new metrics that include these resource coordinates as
labels, we need access to the parent and backend references from the
balancer stack.

This change makes the following changes to accommodate this:

1. We introduce `ParentRef`, `RouteRef`, `BackendRef`, and
   `EndpointRef` new-types to serve as explicit markers for
   different types of metadata we may encounter.
3. The profile stack is updated to parse out metadata from service names
   in the form <name>.<ns>.svc.*.
4. We also support discovering pod metadata from the endpoint labels
   `dst_pod` and `dst_namespace`.
5. When we can't infer any metadata from a profile response, we use an
   "unknown" metadata.
6. When using policy, we use control-plane provided metadata.

In practice, pod metadata won't be surfaced by the existing code path;
but we have a relatively simple path forward to using it.

This change splits the balancer stack into its own layer so it's
distinct from the rest of the concrete stack configuration.

There are only minor functional changes in this branch:

1. The `balance{addr=...}` tracing context is replace by
   `service{ns=..,name=...}` for clarity;
2. Errors returned from the balance stack are now rendered as:
   `Service {name}.{ns}: {error}`


olix0r:
Builds on https://github.com/linkerd/linkerd2-proxy/pull/2372 https://github.com/linkerd/linkerd2-proxy/pull/2373

linkerd/app/outbound/src/http/logical/policy/router.rs
```diff
@@ -113,58 +117,72 @@ where
...
+                ),
+            ),
+            policy::BackendDispatcher::Forward(addr, ref md) => mk_concrete(
+                EndpointRef::new(md, addr.port().try_into().expect("port must not be 0")).into(),
+                concrete::Dispatch::Forward(Remote(ServerAddr(addr)), md.clone()),
             ),
-            policy::BackendDispatcher::Fail { ref message } => {
-                mk_concrete(concrete::Dispatch::Fail {
+            policy::BackendDispatcher::Fail { ref message } => mk_concrete(
+                BackendRef(policy::Meta::new_default("fail")),
```
hawkw, line 164: not a blocker for now, but it occurs to me that the `fail`ing backends _could_ potentially have non-default `BackendRef` metadata, since they're returned by the controller when a backend service doesn't exist...so they should have the metadata of the referenced (non-existent) service.

linkerd/app/outbound/src/lib.rs
```diff
@@ -244,3 +261,85 @@ pub fn trace_labels() -> HashMap<String, String> {
...
+impl std::ops::Deref for EndpointRef {
+    type Target = policy::Meta;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl EndpointRef {
+    fn new(md: &Metadata, port: NonZeroU16) -> Self {
```
hawkw, line 324: nit, take it or leave it (definitely not important): since this always returns metadata for a Pod, should this maybe be 
```suggestion
    fn new_pod(md: &Metadata, port: NonZeroU16) -> Self {
```
or something?

linkerd/app/outbound/src/lib.rs
```diff
@@ -244,3 +261,85 @@ pub fn trace_labels() -> HashMap<String, String> {
...
+        let namespace = match md.labels().get("dst_namespace") {
+            Some(ns) => ns.clone(),
+            None => return Self(UNKNOWN_META.clone()),
+        };
+        let name = match md.labels().get("dst_pod") {
+            Some(pod) => pod.clone(),
+            None => return Self(UNKNOWN_META.clone()),
```
hawkw, line 331: not that this will ever happen in practice, since we expect the controller to populate both the namespace and pod labels, but would it make sense to return a metadata that has the namespace field populated but the name field set to `<unknown>`, rather than throwing out the information that we _do_ have (the ns and the kind)?

linkerd/proxy/client-policy/src/lib.rs
```diff
@@ -469,12 +494,17 @@ pub mod proto {
                         Some(section)
                     };
 
+                    let port = u16::try_from(port)
+                        .ok()
```
hawkw, line 498: feels like this should return an `InvalidMeta` error if the port number is greater than 65535, rather than just having it be `None`?

linkerd/proxy/client-policy/src/lib.rs
```diff
@@ -469,12 +494,17 @@ pub mod proto {
...
+                        .and_then(|p| NonZeroU16::try_from(p).ok());
```
hawkw, line 499: nit: `NonZeroU16::try_from(p).ok()` is equivalent to [`NonZeroU16::new`](https://doc.rust-lang.org/stable/std/num/struct.NonZeroU16.html#method.new), which returns an `Option`:
```suggestion
                        .and_then(NonZeroU16::new)
```

hawkw (approved):
Overall this looks good to me! I left a handful of fairly minor comments.

linkerd/app/outbound/src/lib.rs
olix0r: Yes probably but way out of scope for now

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -34,9 +34,9 @@ pub struct DispatcherFailed(Arc<str>);
 
 /// Wraps errors encountered in this module.
 #[derive(Debug, thiserror::Error)]
-#[error("concrete service {addr}: {source}")]
-pub struct ConcreteError {
-    addr: NameAddr,
+#[error("{} {}.{}: {source}", backend.0.kind(), backend.0.name(), backend.0.namespace())]
```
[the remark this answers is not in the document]
olix0r: No, I think `ns/name` is an old thing we did that we should not do ever again :). Using kubectl, etc, resources are always '{kind} {name}' or '{kind}/{name}'. Using namespace in that form is ambiguous and not what we want to do in general.

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -231,6 +162,117 @@ impl<T> svc::Param<http::balance::EwmaConfig> for Balance<T> {
...
+                .push(svc::NewMapErr::layer_from_target::<BalanceError, _>())
+                .push_on_service(http::BoxResponse::layer())
+                .push_on_service(metrics.proxy.stack.layer(stack_labels("http", "balance")))
+                .instrument(|t: &Self| {
+                    let BackendRef(meta) = t.parent.param();
+                    info_span!(
+                        "service",
+                        ns = %meta.namespace(),
+                        name = %meta.name(),
+                        port = %meta.port().map(u16::from).unwrap_or(0),
```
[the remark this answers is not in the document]
olix0r: No, `port()` returns an Option

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -119,71 +115,15 @@ impl<N> Outbound<N> {
                 )
                 .instrument(|e: &Endpoint<T>| info_span!("forward", addr = %e.addr));
```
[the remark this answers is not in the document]
olix0r: We probably want this info on the endpoint stack more than the forward stack, but I'm considering that out of scope for now what we need in the short term.

PR #2377, opened 2023-04-04
outbound: Report per-route-backend request count metrics

olix0r:
When performing policy-based routing, proxies may dispatch requests through per-route backend configurations. In order to illustrate how routing rules apply and how backend distributions are being honored, this change adds two new metrics:

* `outbound_http_route_backend_requests_total`; and
* `outbound_grpc_route_backend_requests_total`

Each of these metrics includes labels that identify a routes parent (i.e. a Service), the route resource being used, and the backend resource being used.

This implementation does NOT implement any form of metrics eviction for these new metrics. This is tolerable for the short term as the cardinality of services and routes is generally much less than the cardinality of individual endpoints (where we do require timeout/eviction for metrics).

### DIFF
--- a/linkerd/app/outbound/src/http/logical/policy/route/backend.rs
+++ b/linkerd/app/outbound/src/http/logical/policy/route/backend.rs
@@ -1,11 +1,18 @@
 use super::{super::Concrete, filters};
+use crate::RouteRef;
 use linkerd_app_core::{proxy::http, svc, Error, Result};
 use linkerd_http_route as http_route;
 use linkerd_proxy_client_policy as policy;
 use std::{fmt::Debug, hash::Hash, sync::Arc};
 
+mod count_reqs;
+mod metrics;
+
+pub use self::{count_reqs::RequestCount, metrics::RouteBackendMetrics};
+
 #[derive(Debug, PartialEq, Eq, Hash)]
 pub(crate) struct Backend<T, F> {
+    pub(crate) route_ref: RouteRef,
     pub(crate) concrete: Concrete<T>,
     pub(crate) filters: Arc<[F]>,
 }
@@ -16,11 +23,17 @@ pub(crate) type Http<T> =
 pub(crate) type Grpc<T> =
     MatchedBackend<T, http_route::grpc::r#match::RouteMatch, policy::grpc::Filter>;
 
+#[derive(Clone, Debug)]
+pub struct ExtractMetrics {
+    metrics: RouteBackendMetrics,
+}
+
 // === impl Backend ===
 
 impl<T: Clone, F> Clone for Backend<T, F> {
     fn clone(&self) -> Self {
         Self {
+            route_ref: self.route_ref.clone(),
             filters: self.filters.clone(),
             concrete: self.concrete.clone(),
         }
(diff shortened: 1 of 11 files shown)


linkerd/app/outbound/src/http/logical/policy/route/backend.rs
```diff
@@ -109,3 +128,23 @@ impl<T> filters::Apply for Grpc<T> {
...
+
+impl<T> svc::ExtractParam<RequestCount, Grpc<T>> for ExtractMetrics {
+    fn extract_param(&self, params: &Grpc<T>) -> RequestCount {
+        RequestCount(self.metrics.grpc_requests_total(
+            params.params.concrete.parent_ref.clone(),
+            params.params.route_ref.clone(),
+            params.params.concrete.backend_ref.clone(),
+        ))
+    }
+}
```
hawkw, line 150: neat, this could be a nice pattern to use for other metrics stuff eventually!

linkerd/app/outbound/src/http/logical/policy/tests.rs
```diff
@@ -114,6 +147,20 @@ async fn header_based_route() {
...
+    let report = linkerd_app_core::metrics::FmtMetrics::as_display(&metrics).to_string();
+    let mut lines = report
+        .lines()
+        .filter(|l| !l.starts_with('#'))
+        .collect::<Vec<_>>();
+    lines.sort();
+    assert_eq!(
+        lines,
+        vec![
+            r#"outbound_http_route_backend_requests_total{parent_group="core",parent_kind="Service",parent_namespace="ns",parent_name="papa",parent_port="7979",parent_section_name="",route_group="policy.linkerd.io",route_kind="HTTPRoute",route_namespace="ns",route_name="default",backend_group="core",backend_kind="Service",backend_namespace="ns",backend_name="default",backend_port="8080",backend_section_name=""} 1"#,
```
hawkw, line 160: nit, not actually important: it seems like it might be nicer to use multi-line string literals here with `\` to escape whitespace? for example, this is the same string:
```suggestion
         "outbound_http_route_backend_requests_total{\
              parent_group=\"core\",\
              parent_kind=\"Service\",\
              parent_namespace=\"ns\",\
              parent_name=\"papa\",\
              parent_port=\"7979\",\
              parent_section_name=\"\",\
              route_group=\"policy.linkerd.io\",\
              route_kind=\"HTTPRoute\",
              route_namespace=\"ns\",\
              route_name=\"default\",\
              backend_group=\"core\",\
              backend_kind=\"Service\",\
              backend_namespace=\"ns\",\
              backend_name=\"default\",\
              backend_port=\"8080\",\
              backend_section_name=\"\"\
          } 1",
```
but the label values are a little easier to read...

hawkw (approved):
looks good to me, the use of `ExtractParam` for metrics is very cool and makes me want to rewrite the rest of our metrics code (eventually). no blockers from me!

linkerd/app/outbound/src/http/logical/policy/tests.rs
olix0r: yeah this test could definitely be prettier in a few ways. going to leave the mush of labels for now.

PR #2380, opened 2023-04-06
outbound: Report HTTP balancer endpoint gauges

olix0r:
It's not currently possible to know how many endpoints are in a balancer. This change adds an `outbound_http_balancer_endpoints` gauge that exposes the number of endpoints a balancer has by their current readiness status.

Note that in the concrete stack we do not currently differentiate between gRPC and HTTP backends, so all balancers are exposed under this single metric.

### DIFF
--- a/linkerd/app/outbound/src/http/concrete.rs
+++ b/linkerd/app/outbound/src/http/concrete.rs
@@ -4,7 +4,9 @@
 use super::{balance, breaker, client, handle_proxy_error_headers};
 use crate::{http, stack_labels, BackendRef, Outbound, ParentRef};
 use linkerd_app_core::{
-    classify, metrics, profiles,
+    classify,
+    metrics::{prefix_labels, EndpointLabels, OutboundEndpointLabels},
+    profiles,
     proxy::{
         api_resolve::{ConcreteAddr, Metadata, ProtocolHint},
         core::Resolve,
@@ -18,6 +20,12 @@ use linkerd_app_core::{
 use linkerd_proxy_client_policy::FailureAccrual;
 use std::{fmt::Debug, net::SocketAddr, sync::Arc};
 use tracing::info_span;
+
+mod metrics;
+#[cfg(test)]
+mod tests;
+
+pub use self::metrics::BalancerMetrics;
 
 /// Parameter configuring dispatcher behavior.
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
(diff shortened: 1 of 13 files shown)


linkerd/app/outbound/src/http/concrete/tests.rs
```diff
@@ -0,0 +1,152 @@
...
+use std::{net::SocketAddr, num::NonZeroU16, sync::Arc};
+use tokio::time;
+
+#[tokio::test(flavor = "current_thread")]
+async fn gauges_endpoints() {
+    let _trace = trace::test::trace_init();
+    let (rt, _shutdown) = runtime();
+    let outbound = Outbound::new(default_config(), rt);
+
+    let addr = format!("mysvc.myns.svc.cluster.local:80")
```
hawkw, line 18: clippy would prefer you use 
```suggestion
    let addr = "mysvc.myns.svc.cluster.local:80"
        .to_string()
```

linkerd/proxy/balance/src/gauge_endpoints.rs
```diff
@@ -0,0 +1,130 @@
...
+};
+
+/// Parameter that provides a per-balancer endpoint gauge.
+#[derive(Clone, Debug, Default)]
+pub struct EndpointsGauges {
+    pub ready: Arc<Gauge>,
+    pub pending: Arc<Gauge>,
+}
+
+/// A [`NewService`] that builds [`NewGaugeBalancerEndpoint`]s with an
```
hawkw, line 15: nit: docs check doesn't like that `NewGaugeBalancerEndpoint` isn't publicly exported
```suggestion
/// A [`NewService`] that builds `NewGaugeBalancerEndpoint`s with an
```

linkerd/proxy/balance/src/gauge_endpoints.rs
```diff
@@ -0,0 +1,130 @@
...
+/// endpoint gauge.
+#[derive(Clone, Debug, Default)]
+pub struct NewGaugeBalancerEndpoint<N> {
+    inner: N,
+    gauge: EndpointsGauges,
+}
+
+/// A [`Service`] that decrements an endpoint gauge when dropped.
+#[derive(Debug)]
+
```
hawkw, line 33: nit:
```suggestion
```

hawkw (approved):
nice! i complained about some unimportant stuff, but this looks great.

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -234,6 +244,12 @@ where
                         }
                     }),
                 )
+                .push(balance::NewGaugeEndpoints::layer_via({
+                    let metrics = metrics.http_balancer.clone();
+                    move |target: &Self| {
+                        metrics.http_endpoints(target.parent.param(), target.parent.param())
```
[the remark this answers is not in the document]
olix0r: Yeah, I see what you mean. I don't think the context is super important here so I'm okay living with this for now.

PR #2385, opened 2023-04-11
outbound: Export balancer endpoints gauges

olix0r:
d6f20a6 added a new `outbound_http_balancer_endpoints` gauge, but omitted it from metrics export. This change fixes this deficiency to ensure that this metric is rendered.


hawkw (approved):
whoops, good catch!

PR #2395, opened 2023-04-19
inbound: determine default policies using the opaque ports env var

hawkw:
The proxy injector populates an environment variable,
`LINKERD2_PROXY_INBOUND_PORTS_DISABLE_PROTOCOL_DETECTION`, with a list
of all ports marked as opaque. Currently, however, _the proxy _does not
actually use this environment variable_. Instead, opaque ports are
discovered from the policy controller. The opaque ports environment
variable was used only when running in the "fixed" inbound policy mode,
where all inbound policies are determined from environment variables,
and no policy controller address is provided. This mode is no longer
supported, and the policy controller address is now required, so the
variable is not currently used to discover inbound opaque ports.

There are two issues with the current state of things. One is that
inbound policy discovery is _non-blocking_: when an inbound proxy
receives a connection on a port that it has not previously discovered a
policy for, it uses the default policy until it has successfully
discovered a policy for that port from the policy controller. This means
that the proxy may perform protocol detection on the first connection to
an opaque port. This isn't great, as it may result in a protocol
detection timeout error on a port that the user had previously marked as
opaque. It would be preferable for the proxy to read the environment
variable, and use it to determine whether the default policy for a port
is opaque, so that ports marked as opaque disable protocol detection
even before the "actual" policy is discovered.

The other issue with the
variable is that it is currently a list of _individual port numbers_,
while the proxy injector can accept annotations that specify _ranges_ of
opaque ports. This means that when a very large number of ports are
marked as opaque, the proxy manifest must contain a list of each
individual port number in those ranges, making it potentially quite
large. See linkerd/linkerd2#9803 for details on this issue.

This branch addresses both of these problems. The proxy is changed so
that it will once again read the
variable, and use it to determine which ports should have opaque
policies by default. The parsing of the environment variable is changed
to support specifying ports as a list of ranges, rather than a list of
individual port numbers. Along with a proxy-injector change, this would
resolve the manifest size issue described in linkerd/linkerd2#9803.

This is implemented by changing the `inbound::policy::Store` type to
also include a set of port ranges that are marked as opaque. When the
`Store` handles a `get_policy` call for a port that is not already in
the cache, it starts a control plane watch for that port just as it did
previously. However, when determining the initial _default_ value for
the policy, before the control plane discovery provides one, it checks
whether the port is in a range that is marked as opaque, and, if it is,
uses an opaque default policy instead.

This approach was chosen rather than pre-populating the `Store` with
policies for all opaque ports to better handle the case where very large
ranges are marked as opaque and are used infrequently. If the `Store`
was pre-populated with default policies for all such ports, it would
essentially behave as though all ports in
`LINKERD2_PROXY_INBOUND_PORTS_DISABLE_PROTOCOL_DETECTION` were also in
`LINKERD2_PROXY_INBOUND_PORTS`, and the proxy would immediately start a
policy controller discovery watch for all opaque ports, which would be
kept open for the proxy's entire lifetime. In cases where the opaque
ports ranges include ~10,000s of ports, this causes significant
unnecessary load on the policy controller. Storing opaque port ranges
separately and using them to determine the default policy as needed
allows opaque port policies to be treated the same as non-default ports,
which are discovered as needed and can be evicted from the cache if they
are unused. If a port is in both
`LINKERD2_PROXY_INBOUND_PORTS_DISABLE_PROTOCOL_DETECTION` *and*
`LINKERD2_PROXY_INBOUND_PORTS`, the proxy will start discovery eagerly
and retain the port in the cache forever, but the default policy will be
opaque.

I've also added a test for the behavior of opaque ports where the port's
policy has not been discovered from the policy controller. That test
fails on `main`, as the proxy attempts protocol detection, but passes on
this branch.

In addition, I changed the parsing of the `LINKERD2_PROXY_INBOUND_PORTS`
environment variable to also accept ranges, because it seemed like a
nice thing to do while I was here. :)

### DIFF
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -593,8 +593,8 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
 
             // Load the the set of all known inbound ports to be discovered
             // eagerly during initialization.
-            let mut ports = match parse(strings, ENV_INBOUND_PORTS, parse_port_set)? {
-                Some(ports) => ports,
+            let mut ports = match parse(strings, ENV_INBOUND_PORTS, parse_port_range_set)? {
+                Some(ports) => ports.into_iter().flat_map(|range| range.into_iter()).collect::<HashSet<_>>(),
                 None => {
                     error!("No inbound ports specified via {}", ENV_INBOUND_PORTS,);
                     Default::default()
@@ -939,16 +939,6 @@ fn parse_addr(s: &str) -> Result<Addr, ParseError> {
         error!("Not a valid address: {}", s);
         ParseError::AddrError(e)
     })
-}
-
-fn parse_port_set(s: &str) -> Result<HashSet<u16>, ParseError> {
-    let mut set = HashSet::new();
-    if !s.is_empty() {
-        for num in s.split(',') {
-            set.insert(parse_number::<u16>(num)?);
-        }
-    }
-    Ok(set)
 }
 
 fn parse_port_range_set(s: &str) -> Result<RangeInclusiveSet<u16>, ParseError> {
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -1436,4 +1436,34 @@ mod tests {
         assert!(parse_ip_set("10.4.0.3,foobar,192.168.0.69").is_err());
         assert!(parse_ip_set("10.0.1.1/24").is_err());
     }
+
+    #[test]
+    fn ranges() {
+        fn set(ranges: impl IntoIterator<Item = std::ops::RangeInclusive<u16>>) -> Result<RangeInclusiveSet<u16>, ParseError> {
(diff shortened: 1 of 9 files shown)


linkerd/app/src/env.rs
```diff
@@ -929,11 +944,29 @@ fn parse_addr(s: &str) -> Result<Addr, ParseError> {
...
+                    e
+                })?;
+                if high < low {
+                    error!("Not a valid port range: {part}; {high} is greater than {low}");
+                    return Err(ParseError::NotAPortRange);
+                }
+                set.insert(low..=high);
+            } else {
+                set.insert(low..=low);
+            }
```
olix0r, line 969: It's probably appropriate to add some tests for this, now that it's getting a bit more complicated. There's a `mod tests` at the bottom of the file.

hawkw: Will do!

olix0r (approved):

PR #2396, opened 2023-04-19
outbound: synthesize client policies on `Unimplemented`

hawkw:
If the policy controller is from a Linkerd version earlier than 2.13.x, it will return the `Unimplemented` gRPC status code for requests to the `OutboundPolicies` API. The proxy's outbound policy client will currently retry this error code, rather than synthesizing a default policy. Since 2.13.x proxies require an `OutboundPolicy` to be discovered before handling outbound traffic, this means that 2.13.x proxies cannot handle outbound connections when the control plane is on an earlier version. Therefore, installing Linkerd 2.13 and then downgrading to 2.12 can potentially break the data plane's ability to route traffic.

In order to support downgrade scenarios, the proxy should also synthesize a default policy when receiving an `Unimplemented` gRPC status code from the policy controller. This branch changes the proxy to do that. A warning is logged which indicates that the control plane version is older than the proxy's.

### DIFF
--- a/linkerd/app/gateway/src/discover.rs
+++ b/linkerd/app/gateway/src/discover.rs
@@ -23,13 +23,6 @@ impl Gateway {
         Error = Error,
         Future = impl Send + Unpin,
     > + Clone {
-        #[inline]
-        fn is_not_found(e: &Error) -> bool {
-            errors::cause_ref::<tonic::Status>(e.as_ref())
-                .map(|s| s.code() == tonic::Code::NotFound)
-                .unwrap_or(false)
-        }
-
         use futures::future;
 
         let allowlist = self.config.allow_discovery.clone();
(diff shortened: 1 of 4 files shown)


linkerd/app/gateway/src/discover.rs
```diff
@@ -71,7 +64,14 @@ impl Gateway {
...
                     // will synthesize a default policy.
-                    Err(error) if is_not_found(&error) => tracing::debug!("Policy not found"),
+                    Err(error) if errors::has_grpc_status(&error, tonic::Code::NotFound) =>
+                        tracing::debug!("Policy not found"),
+                    // Earlier versions of the Linkerd control plane (e.g.
+                    // 2.12.x) will return `Unimplemented` for requests to the
+                    // OutboundPolicy API. Log a warning and synthesize a policy
+                    // for backwards compatibility.
+                    Err(error) if errors::has_grpc_status(&error, tonic::Code::Unimplemented) =>
+                        tracing::warn!("Policy controller returned `Unimplemented`, the control plane may be out of date."),
```
olix0r, line 74: This could generate lots o
```suggestion
                    // OutboundPolicy API.
                    Err(error) if errors::has_grpc_status(&error, tonic::Code::Unimplemented) =>
                        tracing::debug!("Policy controller returned `Unimplemented`, the control plane may be out of date."),
```

linkerd/app/outbound/src/discover.rs
```diff
@@ -108,7 +108,13 @@ impl<N> Outbound<N> {
...
                     // our lookups, since it doesn't yet serve endpoint metadata for
                     // forwarding.
-                    Err(error) if is_not_found(&error) => tracing::debug!("Policy not found"),
+                    Err(error) if errors::has_grpc_status(&error, tonic::Code::NotFound) => tracing::debug!("Policy not found"),
+                    // Earlier versions of the Linkerd control plane (e.g.
+                    // 2.12.x) will return `Unimplemented` for requests to the
+                    // OutboundPolicy API. Log a warning and synthesize a policy
+                    // for backwards compatibility.
+                    Err(error) if errors::has_grpc_status(&error, tonic::Code::Unimplemented) =>
+                        tracing::warn!("Policy controller returned `Unimplemented`, the control plane may be out of date."),
```
olix0r, line 117:
```suggestion
                    // OutboundPolicy API.
                    Err(error) if errors::has_grpc_status(&error, tonic::Code::Unimplemented) =>
                        tracing::debug!("Policy controller returned `Unimplemented`, the control plane may be out of date."),
```

olix0r (commented):
I think warnings probably aren't worth it here. During legitimate downgrades this could generate a ton of log spam for services like prometheus.

PR #2402, opened 2023-04-25
dev: Update to Rust v1.69.0

olix0r:
Update dev to v40


linkerd/app/integration/src/client.rs
```diff
@@ -119,7 +119,6 @@ impl Client {
         }
     }
 
-    #[track_caller]
```
hawkw, line 122: why was this removed? it should result in nicer panic messages if a request gets a non-200 status...

linkerd/meshtls/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![deny(rust_2018_idioms, clippy::disallowed_methods, clippy::disallowed_types)]
+#![allow(clippy::large_enum_variant)]
```
hawkw, line 2: is there a specific large enum variant in this crate, or are there multiple? would it be better to put the allow attribute on the offending enum, rather than crate-wide?

olix0r: This crate consists only of enums over the feature-flagged variants. It's complaining that the Unimplemented variants are 0B and the others are not.

linkerd/app/integration/src/client.rs
olix0r: It does nothing with async fns

linkerd/meshtls/src/lib.rs
hawkw: ah, got it!

hawkw (approved):
looks good to me!

PR #2408, opened 2023-05-03
Propagate correct span id via W3C context

whiskeysierra:
The W3C context propagation uses the wrong span id right now. That causes all spans emitted by linkerd-proxy to be siblings rather than children of their original parent.

This only applies to W3C as far as I can tell, because the B3 propagation uses the span id correctly.


hawkw (approved):
Looks good to me, thank you!

mateiidavid (approved):
Thanks @whiskeysierra!

PR #2410, opened 2023-05-18
Classify grpc requests properly in the http classifier

adleong:
The gRPC protocol always sets the HTTP response status code to 200 and instead communicates failures in a grpc-status header sent in a TRAILERS frame. Linkerd uses the HTTP response status code to determine if a response is successful, and therefore will consider all gRPC responses successful regardless of their gRPC status code. This means that functionality such as retries and circuit breaking do not function correctly with gRPC traffic.

We update the Http classifier to look for the presence of a `Content-Type: application/grpc` header and use Grpc response classification when it is set.

### DIFF
--- a/linkerd/app/core/src/classify.rs
+++ b/linkerd/app/core/src/classify.rs
@@ -71,7 +71,18 @@ impl classify::Classify for Request {
         match self {
             Self::Profile(classes) => Response::Profile(classes.clone()),
 
-            Self::ClientPolicy(ClientPolicy::Http(policy)) => Response::Http(policy.clone()),
+            Self::ClientPolicy(ClientPolicy::Http(policy)) => {
+                let is_grpc = req
+                    .headers()
+                    .get(http::header::CONTENT_TYPE)
+                    .and_then(|v| v.to_str().ok())
+                    .map(|ct| ct.starts_with("application/grpc"))
+                    .unwrap_or(false);
+                if is_grpc {
+                    return Response::Grpc(Default::default());
+                }
+                Response::Http(policy.clone())
+            }
             Self::ClientPolicy(ClientPolicy::Grpc(policy)) => Response::Grpc(policy.clone()),
 
             Self::Default => {
@@ -242,8 +253,11 @@ impl Class {
 #[cfg(test)]
 mod tests {
     use super::Class;
-    use http::{HeaderMap, Response, StatusCode};
-    use linkerd_proxy_http::classify::{ClassifyEos, ClassifyResponse};
+    use http::{HeaderMap, Request, Response, StatusCode};
+    use linkerd_proxy_http::{
+        classify::{ClassifyEos, ClassifyResponse},
+        Classify,
+    };
 
     #[test]
     fn http_response_status_ok() {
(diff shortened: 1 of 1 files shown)


linkerd/app/core/src/classify.rs
```diff
@@ -343,4 +357,21 @@ mod tests {
...
+
+    #[test]
+    fn http_handles_grpc_requests() {
+        let req = Request::builder()
+            .header(http::header::CONTENT_TYPE, "application/grpc+proto")
+            .body(())
+            .unwrap();
+        let rsp = Response::builder().status(StatusCode::OK).body(()).unwrap();
+        let mut trailers = HeaderMap::new();
+        trailers.insert("grpc-status", 4.into());
```
hawkw, line 369: should we also be testing the case where the `grpc-status` header is in the response start?

linkerd/app/core/src/classify.rs
```diff
@@ -71,7 +71,18 @@ impl classify::Classify for Request {
...
+            Self::ClientPolicy(ClientPolicy::Http(policy)) => {
+                let is_grpc = req
+                    .headers()
+                    .get(http::header::CONTENT_TYPE)
+                    .and_then(|v| v.to_str().ok())
+                    .map(|ct| ct.starts_with("application/grpc"))
+                    .unwrap_or(false);
+                if is_grpc {
+                    return Response::Grpc(Default::default());
+                }
```
hawkw, line 83: So, one difference between the behavior with `Request::Profile` and `Request::Default` is that in the `Profile` case, we will still use the configured HTTP classifier _if the response begins with a non-`200 OK` status code_ (which would indicate that it will not be terminated by a `grpc-status` (IIRC). In the `Default` case, we will always use the gRPC classifier (and just treat any 5xx HTTP status as an error), which is fine since we don't have a user-defined HTTP classifier. 

This code makes the `ClientPolicy::Http` classifier behave the same as the `Default` classifier, rather than the `Profile` classifier. I'm not sure if this is actually the most desirable behavior here --- would it be more correct to behave more like the `Profile` classifier, and use the configured HTTP classifier if the response has a 5xx HTTP status and there is no `grpc-status` trailer?

adleong: If the `Content-Type: application/grpc` header is present, I would expect the request to be classified entirely based on `grpc-status` and for the HTTP response status to be ignored, even if it's a non-200.

hawkw: I also think that this behavior is pretty reasonable; I just wanted to point out that it's different from what we do for ServiceProfiles, and make sure we were okay with that.

linkerd/app/core/src/classify.rs
```diff
@@ -71,17 +71,16 @@ impl classify::Classify for Request {
...
-                    .headers()
-                    .get(http::header::CONTENT_TYPE)
-                    .and_then(|v| v.to_str().ok())
-                    .map(|ct| ct.starts_with("application/grpc"))
-                    .unwrap_or(false);
-                if is_grpc {
+                if is_grpc(req.headers()) {
                     return Response::Grpc(Default::default());
                 }
                 Response::default()
```
hawkw, line 86: nit, take it or leave it: it occurs to me that we don't actually need to repeat this branch twice, and we could instead write
```rust
        match self {
            Self::Profile(classes) => Response::Profile(classes.clone()),
            Self::ClientPolicy(ClientPolicy::Grpc(policy)) => Response::Grpc(policy.clone()),
            // If the request's content type indicates that it is gRPC, use the
            // default gRPC classifier rather than a configured HTTP classifier.
            _ if is_grpc(req.headers()) => Response::Grpc(Default::default()),
            Self::ClientPolicy(ClientPolicy::Http(policy)) => Response::Http(policy.clone()),
            Self::Default => Response::default(),
        }
```
but it's up to you whether you think that's easier to follow or not...

hawkw (approved):
This looks good to me! I'd like to get @olix0r's approval as well before merging.

olix0r (approved):

PR #2418, opened 2023-06-02
outbound: implement `OutboundPolicies` route request timeouts

hawkw:
The latest proxy-api release, v0.10.0, adds fields to the
`OutboundPolicies` API for configuring HTTP request timeouts, based on
the proposed changes to HTTPRoute in kubernetes-sigs/gateway-api#1997.

This branch updates the proxy-api dependency to v0.10.0 and adds the new
timeout configuration fields to the proxy's internal client policy
types. In addition, this branch adds a timeout middleware to the HTTP
client policy stack, so that the timeout described by the
`Rule.request_timeout` field is now applied.

Implementing the `RouteBackend.request_timeout` field with semantics as
close as possible to those described in GEP-1742 will be somewhat more
complex, and will be added in a separate PR. The reason for this
complexity is that, as described in GEP-1742, the
`timeouts.backendRequest` timeout is intended to bound only the time it
takes for the *backend* to respond to the request, rather than bounding
both time spent in the proxy and time spent in the backend (which isthe
intention behind the `timeouts.request` timeout). Therefore, the timeout
for `backendRequest` should be started as "close" to the actual backend
client as possible, ideally in the endpoint stack after the request has
already made it through the load balancer queue. Starting this timeout
in the logical stack like the route's `request_timeout` does would mean
that time spent in the proxy's queues is included in this timeout, which
seems not to be the intention. However, if we want these timeouts to
start in the endpoint stack, we can't configure them using the target
type, because a backend may be referenced with different timeouts by
different HTTPRoutes, and we would prefer not to create separate client
stacks based on those timeout values. Instead, we will need the request
to carry the configured timeout value as an extension, and a new timeout
middleware that applies timeouts from a request extension. This will be
implemented in a follow-up PR.

### DIFF
--- a/linkerd/app/outbound/src/http/logical/policy/route.rs
+++ b/linkerd/app/outbound/src/http/logical/policy/route.rs
@@ -110,10 +110,10 @@ where
                 // consideration, so we must eagerly fail requests to prevent
                 // leaking tasks onto the runtime.
                 .push_on_service(svc::LoadShed::layer())
-                // Sets an optional request timeout.
-                .push(http::NewTimeout::layer())
                 // TODO(ver) attach the `E` typed failure policy to requests.
                 .push(filters::NewApplyFilters::<Self, _, _>::layer())
+                // Sets an optional request timeout.
+                .push(http::NewTimeout::layer())
                 .push(classify::NewClassify::layer())
                 .push(svc::ArcNewService::layer())
                 .into_inner()
--- a/linkerd/app/outbound/src/http/logical/policy/route.rs
+++ b/linkerd/app/outbound/src/http/logical/policy/route.rs
@@ -30,6 +30,7 @@ pub(crate) struct Route<T, F, E> {
     pub(super) filters: Arc<[F]>,
     pub(super) distribution: BackendDistribution<T, F>,
     pub(super) failure_policy: E,
+    pub(super) request_timeout: Option<std::time::Duration>,
 }
 
 pub(crate) type MatchedRoute<T, M, F, E> = Matched<M, Route<T, F, E>>;
@@ -109,6 +110,8 @@ where
                 // consideration, so we must eagerly fail requests to prevent
                 // leaking tasks onto the runtime.
                 .push_on_service(svc::LoadShed::layer())
+                // Sets an optional request timeout.
+                .push(http::NewTimeout::layer())
                 // TODO(ver) attach the `E` typed failure policy to requests.
                 .push(filters::NewApplyFilters::<Self, _, _>::layer())
                 .push(classify::NewClassify::layer())
@@ -121,6 +124,12 @@ where
 impl<T: Clone, M, F, E> svc::Param<BackendDistribution<T, F>> for MatchedRoute<T, M, F, E> {
     fn param(&self) -> BackendDistribution<T, F> {
         self.params.distribution.clone()
+    }
+}
+
+impl<T, M, F, E> svc::Param<http::timeout::ResponseTimeout> for MatchedRoute<T, M, F, E> {
+    fn param(&self) -> http::timeout::ResponseTimeout {
+        http::timeout::ResponseTimeout(self.params.request_timeout)
     }
 }
 
--- a/linkerd/proxy/client-policy/src/opaq.rs
+++ b/linkerd/proxy/client-policy/src/opaq.rs
@@ -127,8 +127,7 @@ pub(crate) mod proto {
             filters: NO_FILTERS.clone(),
             failure_policy: NonIoErrors::default(),
             distribution,
-            // XXX(eliza): maybe this shouldn't even be present on opaque
-            // routes...might have to turn it into a filter...
+            // Request timeouts are ignored on opaque routes.
             request_timeout: None,
         })
     }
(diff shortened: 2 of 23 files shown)


adleong (commented):
Are there any proxy metrics which are incremented when a timeout occurs? Are timeouts distinguishable from application errors?  Are request timeouts distinguishable from connect timeouts, detect timeouts, or failfast?

hawkw:
@adleong 
> Are there any proxy metrics which are incremented when a timeout occurs?

With regards to metrics, these timeouts should behave analogously to timeouts configured for ServiceProfiles. When a request times out, we return an HTTP 504 response to the application, and increment the `response_total` metric with that status code as a label. 

> Are timeouts distinguishable from application errors?

AFAIK, we don't currently have a metric label that distinguishes HTTP 504 error codes that result from internal proxy timeouts from those that are returned by an application service. It's probably a good idea to add a label or something that make proxy timeouts distinguishable from those returned by an upstream application, but this maybe deserves some additional design work --- we could probably make it possible to distinguish between any error status synthesized by the proxy and errors with the same status returned by a remote application service.

> Are request timeouts distinguishable from connect timeouts, detect timeouts, or failfast?

Protocol detection timeouts don't result in an error being returned to the application; when protocol detection times out, the proxy handles the connection as opaque. This is a different behavior than returning an error, so they are distinguishable because the proxy does something different.

Regarding failfast, the proxy will currently return an HTTP 504 for requests already dispatched to a service when it enters failfast, but returns an HTTP 503 when a new request would be dispatched to a service which is shedding load due to failfast: https://github.com/linkerd/linkerd2-proxy/blob/704ef31a286280187c1f4679170e897243f99900/linkerd/app/outbound/src/http/server.rs#L154-L163
This means both timeouts will sometimes increment the `response_total` metric with `status="504"`, so they cannot always be distinguished in metrics currently.

In all cases, we will log substantially different error messages for these timeouts, and the value of the `l5d-proxy-error` header will be different. So these timeouts can be distinguished that way.

adleong (approved):

linkerd/proxy/client-policy/src/opaq.rs
```diff
@@ -127,6 +127,9 @@ pub(crate) mod proto {
             filters: NO_FILTERS.clone(),
             failure_policy: NonIoErrors::default(),
             distribution,
+            // XXX(eliza): maybe this shouldn't even be present on opaque
+            // routes...might have to turn it into a filter...
```
mateiidavid, line 131: little bit confused about this... I guess if a service is marked as opaque, and it has a route attached to it, the route becomes opaque. Generally, the way I reasoned about things in the past was:

* If service is opaque, we don't do protocol detection
* If endpoint is opaque, we change the target port (but we may do protocol detection)

...so I guess an opaque route is a TCP route in effect? would the request timeout never be a good fit? not sure if my reasoning here is sound 🤷🏻

olix0r, line 131: What are we supposed to do with this comment? Should it be a filter? Or is this okay?

olix0r: What does a request timeout mean on an opaque route?

linkerd/app/outbound/src/http/logical/policy/route.rs
```diff
@@ -109,6 +110,8 @@ where
                 // consideration, so we must eagerly fail requests to prevent
                 // leaking tasks onto the runtime.
                 .push_on_service(svc::LoadShed::layer())
+                // Sets an optional request timeout.
+                .push(http::NewTimeout::layer())
```
mateiidavid, line 114: genuine question, not really feedback: why not apply the timeout before the filters? for example why not wrap a redirect filter in the timeout? wouldn't it fall under the work done in the proxy?

mateiidavid (approved):
Legible enough for me to review 🙃 looks good. The opaque route is confusing me a bit but think that's more due to the complexity of opaqueness, not because of this PR.

linkerd/app/outbound/src/http/logical/policy/route.rs
olix0r, line 114: I thought we supported per-backend filters. Am I wrong? Or is that missing? Or is it out of scope?

hawkw: No good reason, I wasn't thinking too hard about this. It could probably be moved to before filters.

linkerd/proxy/client-policy/src/opaq.rs
hawkw: https://github.com/linkerd/linkerd2-proxy/pull/2418/commits/26c9492af1e52ab85b9e1dcbeeed9b791e81dcbd changes this comment to simply state that the timeout is ignored on opaque routes, and adds a doc comment to `ClientPolicy` explaining the rationale behind that decision.

hawkw: Check out the comments I added in https://github.com/linkerd/linkerd2-proxy/pull/2418/commits/26c9492af1e52ab85b9e1dcbeeed9b791e81dcbd and let me know if that answers your questions?

linkerd/app/outbound/src/http/logical/policy/route.rs
hawkw: moved in https://github.com/linkerd/linkerd2-proxy/pull/2418/commits/38f3f2beda0fa557a6092c9014d62c353480e636

hawkw: You mean per-backend timeouts? Those are implemented in a separate PR, #2419, as I discussed in the PR description.

mateiidavid (approved):

PR #2419, opened 2023-06-02
outbound: implement `OutboundPolicies` backend request timeouts

hawkw:
Depends on #2418 

The latest proxy-api release, v0.10.0, adds fields to the
`OutboundPolicies` API for configuring HTTP request timeouts, based on
the proposed changes to HTTPRoute in kubernetes-sigs/gateway-api#1997.
PR #2418 updates the proxy to depend on the new proxy-api release, and
implements the `Rule.request_timeout` field added to the API. However,
that branch does *not* add a timeout for the
`RouteBackend.request_timeout` field. This branch changes the proxy to
apply the backend request timeout when configured by the policy
controller.

This branch implements `RouteBackend.request_timeout` by adding an
additional timeout layer in the `MatchedBackend` stack. This applies the
per-backend timeout once a backend is selected for a route. I've also
added stack tests for the interaction between the request and backend
request timeouts.

Note that once retries are added to client policy stacks, it may be
necessary to move the backend request timeout to ensure it occurs
"below" retries, depending on where the retry middleware ends up being
located in the proxy stack.

### DIFF
--- a/linkerd/app/outbound/src/http/endpoint.rs
+++ b/linkerd/app/outbound/src/http/endpoint.rs
@@ -119,7 +119,6 @@ impl<N> Outbound<N> {
             // is typically used (i.e. when communicating with other proxies); though
             // HTTP/1.x fallback is supported as needed.
             inner
-                .push_on_service(http::timeout::TimeoutFromRequest::layer())
                 // Drive the connection to completion regardless of whether the reconnect is being
                 // actively polled.
                 .push_on_service(svc::layer::mk(svc::SpawnReady::new))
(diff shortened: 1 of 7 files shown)


adleong:
My 2 cents is that we have some leeway in how we choose to implement this and we don't need to follow the spec to the letter.  If it is easier for us to implement and maintain to put this timeout in the logical stack after a backend is selected, I think that would be okay.  The important part is that the timeout happens after retries and after backend selection.

mateiidavid (approved):
I managed to work most of the code out and this looks good to me.

adleong (approved):

PR #2431, opened 2023-06-19
outbound: handle `NotFound` client policies in ingress mode

hawkw:
When the outbound proxy resolves an outbound policy from the policy
controller's `OutboundPolicies` API, the policy controller may return an
error with the `grpc-status` code `NotFound` in order to indicate that
the destination is not a ClusterIP service. When this occurs, the proxy
will fall back to either using a ServiceProfile, if the ServiceProfile
contains non-trivial configuration, or synthesizing a default client
policy from the ServiceProfile.

However, when the outbound proxy is configured to run in ingress mode,
the fallback behavior does not occur. Instead, the ingress mode proxy
treats any error returned by the policy controller's `OutboundPolicies`
API as fatal. This means that when an ingress controller performs its
own load-balancing and opens a connection to a pod IP directly, the
ingress mode proxy will fail any requests on that connection. This is a
bug, and is the cause of the issues described in linkerd/linkerd2#10908.

This branch fixes this by changing the ingress mode proxy to handle
`NotFound` errors returned by the policy controller. I've added similar
logic for synthesizing default policies from a discovered
ServiceProfile, or using the profile if it's non-trivial. Unfortunately,
we can't just reuse the existing `Outbound::resolver` method, as ingress
discovery may be performed for an original destination address *or* for
a DNS name, and it's necessary to construct fallback policies in either
case. Instead, I've added a new function with similar behavior that's
ingress-specific.

I've manually tested this change against the repro steps[^1] described
in linkerd/linkerd2#10908, and verified that the proxy 503s on 2.13.4,
and that it once again routes correctly after applying this change.

Fixes linkerd/linkerd2#10908.

[^1]: As described in the first comment, using Contour and podinfo.

### DIFF
--- a/linkerd/app/outbound/src/discover.rs
+++ b/linkerd/app/outbound/src/discover.rs
@@ -192,14 +192,47 @@ pub fn synthesize_forward_policy(
     addr: SocketAddr,
     metadata: policy::EndpointMetadata,
 ) -> ClientPolicy {
+    policy_for_backend(
+        meta,
+        timeout,
+        policy::Backend {
+            meta: meta.clone(),
+            queue,
+            dispatcher: policy::BackendDispatcher::Forward(addr, metadata),
+        },
+    )
+}
+
+pub fn synthesize_balance_policy(
+    meta: &Arc<policy::Meta>,
+    timeout: Duration,
+    queue: policy::Queue,
+    load: policy::Load,
+    addr: impl ToString,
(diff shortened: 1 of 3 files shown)


olix0r (approved):
This LGTM (pending linting fix).

Are we able to write an integration test for this in linkerd2?

adleong (approved):

PR #2432, opened 2023-06-19
Emit distinguishable version info

olix0r:
The proxy currently emits very little useful version information.

This change updates the proxy to support new build-time environment variables that are used to report version information:

* LINKERD2_PROXY_BUILD_TIME
* LINKERD2_PROXY_VENDOR
* LINKERD2_PROXY_VERSION

Additionally, several pre-existing Git-oriented metadata have been removed, as they were generally redundant or uninformative. The Rustc version has also been removed (since it has no real user-facing value and can be easily determined by the version/tag).

### DIFF
--- a/justfile
+++ b/justfile
@@ -15,8 +15,11 @@ toolchain := ""
 
 features := ""
 
+export LINKERD2_PROXY_VERSION := env_var_or_default("LINKERD2_PROXY_VERSION", "0.0.0-dev." + `git rev-parse --short HEAD`)
+export LINKERD2_PROXY_VENDOR := env_var_or_default("LINKERD2_PROXY_VENDOR", `whoami` + "@" + `hostname`)
+
 # The version name to use for packages.
-package_version := `git rev-parse --short HEAD`
+package_version := "v" + LINKERD2_PROXY_VERSION
 
 # Docker image name & tag.
 docker-repo := "localhost/linkerd/proxy"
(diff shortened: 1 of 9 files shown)


adleong (approved):

linkerd/app/core/src/telemetry/build_info.rs
```diff
@@ -1,14 +1,11 @@
...
-use std::fmt;
-use std::string::String;
-use std::sync::Arc;
+use std::{env, fmt};
 
-const GIT_BRANCH: &str = env!("GIT_BRANCH");
-const GIT_SHA: &str = env!("GIT_SHA");
-const GIT_VERSION: &str = env!("GIT_VERSION");
-const PROFILE: &str = env!("PROFILE");
-const RUST_VERSION: &str = env!("RUST_VERSION");
```
hawkw, line 11: any motivation for removing the rust compiler information from the metric? i know we don't really use it very often (especially because a proxy tag will include a `.rust-version.toml`), but it seems more or less free to include regardless?

justfile
```diff
@@ -15,8 +15,11 @@ toolchain := ""
 
 features := ""
 
+export LINKERD2_PROXY_VERSION := env_var_or_default("LINKERD2_PROXY_VERSION", "0.0.0-dev." + `git rev-parse --short HEAD`)
+export LINKERD2_PROXY_VENDOR := env_var_or_default("LINKERD2_PROXY_VENDOR", `whoami` + "@" + `hostname`)
```
hawkw, line 19: nice :)

hawkw (approved):
very nice!

mateiidavid (approved):

linkerd/app/core/src/telemetry/build_info.rs
olix0r: I don't think it's useful information to Linkerd operators.

PR #2433, opened 2023-06-20
Replace procinfo with procfs

pranoyk:
This PR will replace [procinfo](https://crates.io/crates/procinfo) crate which is not maintained for over 5 years with [procfs](https://crates.io/crates/procfs).

Signed-off-by: Pranoy Kumar kundu <pranoy1998k@gmail.com>

fixes - https://github.com/linkerd/linkerd2/issues/10819


linkerd/system/src/linux.rs
```diff
@@ -39,9 +42,28 @@ pub fn open_fds(pid: pid_t) -> io::Result<u64> {
...
+                }
+            }
+            Err(e) => {
+                error!("Failed to get process limits: {}", e);
+                Ok(None)
+            }
+        },
+        Err(e) => {
+            error!("Failed to get process limits: {}", e);
+            Ok(None)
```
mateiidavid, line 64: It seems that the caller of this function can handle three distinct types of values:

1. Ok(Some)
2. Ok(None) (do nothing)
3. Err, where we log the error we've encountered.

Should we perhaps return the error instead of returning Ok(None)? Feels a bit weird to log here, at a different level than the caller, when the caller already handles this path?

linkerd/system/src/linux.rs
```diff
@@ -39,9 +42,28 @@ pub fn open_fds(pid: pid_t) -> io::Result<u64> {
 }
 
 pub fn max_fds() -> io::Result<Option<u64>> {
-    let limit = pid::limits_self()?.max_open_files;
-    let max_fds = limit.soft.or(limit.hard).map(|max| max as u64);
-    Ok(max_fds)
+    match Process::myself() {
```
mateiidavid, line 45: Maybe a questionable decision, but changing the signature to return `ProcResult` as we have done above might make the code easier to grok. As far as I can tell, this function has only caller so changing the signature should be safe enough, and it's a better trade off than complicating the rest of the function and maintaining an unused code path in the caller (i.e. where we handle errors).

@hawkw might have a different opinion though.

imo it would be good to just do

```suggestion
 let max_fd = Process::myself()?.limits()?.max_open_files;
```

linkerd/system/src/linux.rs
```diff
@@ -39,9 +42,28 @@ pub fn open_fds(pid: pid_t) -> io::Result<u64> {
...
+        Ok(p) => match p.limits() {
+            Ok(limits) => {
+                let limit = limits.max_open_files;
+                match limit.soft_limit {
+                    process::LimitValue::Unlimited => match limit.hard_limit {
+                        process::LimitValue::Unlimited => Ok(None),
+                        process::LimitValue::Value(v) => Ok(Some(v)),
```
mateiidavid, line 52: I'm a little bit confused here. We're basically saying that when a soft_limit is "unlimited", then:

1. If the hard limit is also "unlimited", we return None. This means that the caller will do nothing with this value? Even though there clearly is a max? The max would probably be the system's limit here.
2. If the hard limit is set to a value, we return the hard limit. 

imo this is a bit hard to read through. Perhaps a more accurate description here is:

Soft limit has a value? Return it. Soft limit is unlimited? Then return system max. Wdyt? tbh I also wouldn't have expected to see the hard limit enum use an `Unlimited` value 🤷🏻 it's probably just a side effect of re-using it across all different limit keys. Looking at `setrlimit(2)` I see:

>  EPERM  The caller tried to increase the hard RLIMIT_NOFILE limit above the maximum defined by /proc/sys/fs/nr_open (see proc(5))

>
>        /proc/sys/fs/nr_open (since Linux 2.6.25)
>              This file imposes a ceiling on the value to which the RLIMIT_NOFILE resource limit can be raised (see > getrlimit(2)).  This ceiling is enforced for both unprivileged and privileged process.  The default value in this file is > 1048576.  (Before  Linux  2.6.25,  the
>              ceiling for RLIMIT_NOFILE was hard-coded to the same value.)

so I don't think the hard limit can truly ever be unlimited, regardless of the privilege held by the process.

I did a quick poc with `prlimit` and a web server to illustrate how it would all work. I suggest reading through `prlimit(1)`, `getrlimit(2)` and `proc(5)`

```bash
:; prlimit -p 438595
RESOURCE   DESCRIPTION                              SOFT       HARD UNITS
AS         address space limit                 unlimited  unlimited bytes
CORE       max core file size                          0  unlimited bytes
CPU        CPU time                            unlimited  unlimited seconds
DATA       max data size                       unlimited  unlimited bytes
FSIZE      max file size                       unlimited  unlimited bytes
LOCKS      max number of file locks held       unlimited  unlimited locks
MEMLOCK    max locked-in-memory address space 4174134272 4174134272 bytes
MSGQUEUE   max bytes in POSIX mqueues             819200     819200 bytes
NICE       max nice prio allowed to raise              0          0
NOFILE     max number of open files                 1024    1048576 files
NPROC      max number of processes                127239     127239 processes
RSS        max resident set size               unlimited  unlimited bytes
RTPRIO     max real-time priority                      0          0
RTTIME     timeout for real-time tasks         unlimited  unlimited microsecs
SIGPENDING max number of pending signals          127239     127239 signals
STACK      max stack size                        8388608  unlimited bytes


# Cannot set to unlimited if hard limit is not "unlimited" either
:; prlimit -p 438595 --nofile=unlimited:
prlimit: the soft limit NOFILE cannot exceed the hard limit

# Most systems don't have an option to set "unlimited" for the hard limit though
:; sudo prlimit -p 438595 --nofile=unlimited:unlimited
prlimit: failed to set the NOFILE resource limit: Operation not permitted

# Again, soft limits cannot exceed hard limits
:; sudo prlimit -p 438595 --nofile=unlimited:2048
prlimit: the soft limit NOFILE cannot exceed the hard limit

:; sudo prlimit -p 438595 --nofile=1024:2048

# Setting it down will still not be allowed if it's less than the soft limit
:; sudo prlimit -p 438595 --nofile=:1023
prlimit: the soft limit NOFILE cannot exceed the hard limit
```

mateiidavid (commented):
Left a few comments :)

linkerd/system/src/linux.rs
pranoyk: I agree with you that if we change the signature to return `ProcResult` the code becomes must cleaner and all the error handling can be eliminated.
I will be glad to make that change if you or @hawkw can confirm that.

hawkw: I'm fine with this --- I believe `io::Result` is the return type here only because the `procinfo` API returns an `io::Result`. It's fine to change this.

linkerd/system/src/linux.rs
```diff
@@ -39,9 +42,28 @@ pub fn open_fds(pid: pid_t) -> io::Result<u64> {
...
+                    },
+                    process::LimitValue::Value(v) => Ok(Some(v)),
+                }
+            }
+            Err(e) => {
+                error!("Failed to get process limits: {}", e);
```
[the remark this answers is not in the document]
hawkw: if we change the return type to `ProcResult`, we can remove this match and just use `?` --- the error will be logged by the caller, I believe.

linkerd/system/Cargo.toml
```diff
@@ -14,4 +14,4 @@ tracing = "0.1"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 libc = "0.2"
-procinfo = "0.4"
+procfs = "0.15.1"
```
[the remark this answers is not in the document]
pranoyk: I think its better to keep in all the features as is as few more features are yet to be implemented in procfs(not all proc features are yet available in this crate). This way in case we need anything more procfs features then we will have them with us

linkerd/system/src/linux.rs
pranoyk: Agreed, after going through the `prlimit(1)`, `getrlimit(2)`, `proc(5)` and `setrlimit(2)` I think it should be safe to assume that `hard_limit` always exists regardless of the privilege. I think we can safely assume that if the `soft_limit` is unlimited then we return the `hard_limit`.

pranoyk: It looks like I have to handle the `unlimited` handle for `hard_limit` as well since `hard_limit` is an enum having `unlimited` and `value(u64)`. In that case I think its better to return the sys_max value. Just to check `sys_max` value is the one found in this path right `/proc/sys/fs/file-max`? 
PS. I am new to systems and hence I might be asking dumb questions :)

linkerd/system/Cargo.toml
mateiidavid, line 17: should probably end with a newline here afaik that's what POSIX standards define as a line, no need to remove it.

mateiidavid, line 17: We're still not making use of non-default features. We're importing a bunch of additional dependencies based on default features (e.g. `chrono` and `gzip`) that we won't be using. Can we instead just use non-default here?

linkerd/system/src/linux.rs
```diff
@@ -38,10 +41,15 @@ pub fn open_fds(pid: pid_t) -> io::Result<u64> {
...
 
-pub fn max_fds() -> io::Result<Option<u64>> {
-    let limit = pid::limits_self()?.max_open_files;
-    let max_fds = limit.soft.or(limit.hard).map(|max| max as u64);
-    Ok(max_fds)
+pub fn max_fds() -> ProcResult<u64> {
+    let limits = Process::myself()?.limits()?.max_open_files;
+    match limits.soft_limit {
+        LimitValue::Unlimited => match limits.hard_limit {
+            LimitValue::Unlimited => Ok(0),
```
mateiidavid, line 48: What's 0 supposed to mean here? 🤔

linkerd/system/Cargo.toml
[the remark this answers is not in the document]
mateiidavid: Not sure I agree, we really only need this crate to get proc stats. To me, it makes more sense to have less transitive dependencies.

pranoyk: Weird, running `just fmt` removed the line. I will add it back

linkerd/system/src/linux.rs
pranoyk: In order to retrieve sys_max we need to read the value from the file itself. Instead of doing that, I thought returning 0 here would be better as `hard_limit` would always be available.

linkerd/system/src/linux.rs
pranoyk: @mateiidavid Can you confirm this? ^^

mateiidavid (commented):
Thanks @pranoyk for the work so far. This LGTM. I suspect we'll bump into some issues with `deny` since we're pulling in the same transitive dependency with two separate versions. We'll probably need to add a rule for it. There's a typo in the manifest so let's see if CI complains once that's fixed.

Also, one of the commits seems to be improperly signed. Can you [please confirm](https://github.com/linkerd/linkerd2/blob/main/CONTRIBUTING.md#option-2-public-statement) you agree to making this contribution?

Thanks!

pranoyk:
I agree to the DCO for all the commits in this PR.

mateiidavid (approved):
Nice. Seems like it's a net change.

```
+54 −54 
```

LGTM, pretty clean. Bit of a scary change to switch deps but I think it'll be fine. Thanks!

pranoyk:
@mateiidavid I was just wondering if we expect any more changes on this since it's being approved but not yet merged. 
Do let me know if we are expecting anything else on this. :)

mateiidavid:
> @mateiidavid I was just wondering if we expect any more changes on this since it's being approved but not yet merged. Do let me know if we are expecting anything else on this. :)

Hey, typically we expect more than one reviewer even if the review protection rules have been met. In this case, we're doing a change that swaps out dependencies so it might be more contentious than say a typo or dependency bump. I'm going to wait for one more review here before merging.

deny.toml
```diff
@@ -59,10 +59,12 @@ skip = [
     # Since `syn` is used by proc-macros (executed at compile time), duplicate
     # versions won't have an impact on the final binary size.
     { name = "syn" },
-]
-skip-tree = [
-    # Hasn't seen a new release since 2017. Pulls in an older version of nom.
-    { name = "procinfo" },
+    # right now we have a mix of versions of this crate in the ecosystem
+    # procfs uses 0.36.14, tempfile uses 0.37.4
+    { name = "rustix" },
```
hawkw, line 64: it might be worth opening a PR upstream to update `procfs`' dependency on `rustix`? then, we would be able to remove this.

hawkw (approved):
looks good to me!

deny.toml
pranoyk: So the latest version of rustix is 0.38.0
If we are planning to upgrade then should we upgrading the rustix version for both the dependencies ? Or shall we ignore it for now ?

hawkw: Ah, sorry, I didn't see this. It would be ideal to get both dependencies on the latest `rustix`. That isn't a blocker for merging this PR, but it could be nice to kick off.

hawkw, line 64: i think if we moved this from a `skip` section to a `skip_tree` section, we wouldn't need the separate entry for `linux-raw-sys`.

pranoyk: Yes, it would be great. But help me understand something, I just cannot see where we are using the `tempfile` crate in our code base (this is another crate that uses `rustix` and the reason why we need to add it into deny.toml). I might be making a mistake here but in case we are not using it I can go ahead and remove it.

hawkw: It looks like `tempfile` is a transitive dependency of `prost-build`, which is used for compiling protobufs. This is a build-time dependency, so it's not actually used by the proxy itself. Instead, it's used when building the proxy, to generate Rust code from protocol buffer files.

We can determine where a transitive dependency comes from by using [`cargo tree`](https://doc.rust-lang.org/cargo/commands/cargo-tree.html). If we pass the `--invert` flag to `cargo tree`, we get a tree whose root is the named crate and whose leaves are the crates that depend on that crate. For `tempfile`, it looks like this:
```console
$ cargo tree --package tempfile --invert
tempfile v3.5.0
└── prost-build v0.11.8
    └── tonic-build v0.8.4
        └── tools v0.1.0 (/home/eliza/Code/linkerd2-proxy/tools)
        [dev-dependencies]
        └── opencensus-proto v0.1.0 (/home/eliza/Code/linkerd2-proxy/opencensus-proto)
            └── linkerd-opencensus v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/opencensus)
                ├── linkerd-app v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app)
                │   ├── linkerd-app-integration v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/integration)
                │   └── linkerd2-proxy v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd2-proxy)
                └── linkerd-app-core v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/core)
                    ├── linkerd-app v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app) (*)
                    ├── linkerd-app-admin v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/admin)
                    │   └── linkerd-app v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app) (*)
                    │   [dev-dependencies]
                    │   └── linkerd-app-integration v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/integration)
                    ├── linkerd-app-gateway v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/gateway)
                    │   └── linkerd-app v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app) (*)
                    ├── linkerd-app-inbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/inbound)
                    │   ├── linkerd-app v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app) (*)
                    │   ├── linkerd-app-admin v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/admin) (*)
                    │   └── linkerd-app-gateway v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/gateway) (*)
                    │   [dev-dependencies]
                    │   └── linkerd-app-gateway v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/gateway) (*)
                    ├── linkerd-app-integration v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/integration)
                    ├── linkerd-app-outbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/outbound)
                    │   ├── linkerd-app v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app) (*)
                    │   └── linkerd-app-gateway v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/gateway) (*)
                    │   [dev-dependencies]
                    │   └── linkerd-app-gateway v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/gateway) (*)
                    └── linkerd-app-test v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/test)
                        ├── linkerd-app-inbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/inbound) (*)
                        ├── linkerd-app-integration v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/integration)
                        └── linkerd-app-outbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/outbound) (*)
                        [dev-dependencies]
                        ├── linkerd-app-gateway v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/gateway) (*)
                        ├── linkerd-app-inbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/inbound) (*)
                        └── linkerd-app-outbound v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/outbound) (*)
    [dev-dependencies]
    └── linkerd-transport-header v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/transport-header)
        └── linkerd-app-core v0.1.0 (/home/eliza/Code/linkerd2-proxy/linkerd/app/core) (*)
```

PR #2439, opened 2023-07-06
Add suport for response header filter

adleong:
Add support for the response header modifier, which was added to the proxy API in https://github.com/linkerd/linkerd2-proxy-api/pull/251

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1894,8 +1894,7 @@ dependencies = [
 [[package]]
 name = "linkerd2-proxy-api"
 version = "0.10.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "597facef5c3f12aece4d18a5e3dbba88288837b0b5d8276681d063e4c9b98a14"
+source = "git+https://github.com/linkerd/linkerd2-proxy-api?rev=f1598b46c179f5a981bfca7b660dc89903b18ab7#f1598b46c179f5a981bfca7b660dc89903b18ab7"
 dependencies = [
  "h2",
  "http",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1893,8 +1893,9 @@ dependencies = [
 
 [[package]]
 name = "linkerd2-proxy-api"
-version = "0.10.0"
-source = "git+https://github.com/linkerd/linkerd2-proxy-api?rev=f1598b46c179f5a981bfca7b660dc89903b18ab7#f1598b46c179f5a981bfca7b660dc89903b18ab7"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2348745f909668e6de2dbd175eeeac374887ffb33989a0e09766f1807b27cdfe"
 dependencies = [
  "h2",
  "http",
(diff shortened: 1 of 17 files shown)


Cargo.toml
```diff
@@ -85,3 +85,4 @@ lto = true
 webpki = { git = "https://github.com/linkerd/webpki", branch = "cert-dns-names-0.22" }
 boring = { git = "https://github.com/cloudflare/boring" }
 tokio-boring = { git = "https://github.com/cloudflare/boring" }
+linkerd2-proxy-api = { git = "https://github.com/linkerd/linkerd2-proxy-api", rev = "f1598b46c179f5a981bfca7b660dc89903b18ab7" }
```
risingspiral, line 88: Why wasn't a proxy API dependency present before?

Cargo.lock
```diff
@@ -1894,8 +1894,7 @@ dependencies = [
 [[package]]
 name = "linkerd2-proxy-api"
 version = "0.10.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "597facef5c3f12aece4d18a5e3dbba88288837b0b5d8276681d063e4c9b98a14"
+source = "git+https://github.com/linkerd/linkerd2-proxy-api?rev=f1598b46c179f5a981bfca7b660dc89903b18ab7#f1598b46c179f5a981bfca7b660dc89903b18ab7"
```
risingspiral, line 1897: Why did the source of the package change?

Cargo.toml
adleong: This is a dependency override.  Specifying this replaces the dependency on `linkerd2-proxy-api` everywhere in the project with one on a specific proxy-api commit.  Once we have the changes we need released in a proxy-api release, we will update the dependency to that version and remove this override.

linkerd/app/outbound/src/http/logical/policy/route/filters.rs
```diff
@@ -99,13 +114,32 @@ pub fn apply_http<B>(
...
+    filters: &[http::Filter],
+    rsp: &mut ::http::Response<B>,
+) -> Result<()> {
+    // TODO Do any metrics apply here?
+    for filter in filters {
+        match filter {
+            http::Filter::InjectFailure(_) => {} // InjectFailure filter does not apply to responses.
+            http::Filter::Redirect(_) => {}      // Redirect filter does not apply to responses.
+            http::Filter::RequestHeaders(_) => {} // RequestHeaders filter does not apply to responses.
+            http::Filter::InternalError(_) => {} // InternalError filter does not apply to responses.
```
hawkw, line 134: this feels...somewhat unfortunate. it would be preferable if we could iterate over all the filters once, and apply them not just to the request but to the *future*, so that the filter can act on both the request and the response as it sees fit. but, that would require a much more substantial redesign of how we represent filters, essentially turning them into arbitrary middleware, and then we run into issues with the future type being determined from the value of a dynamic list of filters...

hawkw (approved):
this does look like a workable solution to the problem. i definitely see why you weren't sure about the design of adding `apply_response` to filters, it does feel somewhat awkward. it certainly seems like it would feel nicer if the filter could just wrap the entire future, mapping either the request *or* the response, the way `Service` middleware does. but, thinking it through, I think that would make having an arbitrary list of filters that's determined dynamically at runtime much more difficult, since we would have a response future whose type varies based on what's in the list of filters. which...is hard to represent without adding boxes for type erasure for each filter in the list...

so, i think the design proposed here, though kind of awkward, is *probably* the best solution without completely redesigning how we model these filters. perhaps @olix0r has other ideas, though?

olix0r (approved):
lgtm. one cleanup suggested

PR #2452, opened 2023-08-08
dev: optimize image build

daixiang0:
- `on-create` uses much time to download and install without cache, any changes about devcontainer will make it run again. If we put those into Dockerfile, we will have cache and speed up this process.
- Add proxy settings for development behind a proxy server

Signed-off-by: Loong <loong.dai@intel.com>


olix0r (approved):
Sorry for the long delay on this. Our CI previously depended on the devcontainer image field being set. That's been fixed in this PR.

daixiang0:
@olix0r thank you very much!

PR #2454, opened 2023-08-15
build(deps): use published version of `boring`

hawkw:
The `linkerd-meshtls-boring` crate currently uses a Git dependency on `boring` and `tokio-boring`. This is because, when this crate was initially introduced, the proxy required unreleased changes to these crates. Now, however, upstream has published all the changes we depended on (this happened ages ago), and we can depend on these libraries from crates.io.

This branch removes the Git deps and updates to v3.0.0 of `boring`/`tokio-boring`. I've also changed the `cargo-deny` settings to no longer allow Git deps on these crates, as we no longer depend on them from Git.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -97,7 +97,7 @@ checksum = "13d8068b6ccb8b34db9de397c7043f91db8b4c66414952c6db944f238c4d3db3"
 dependencies = [
  "async-trait",
  "axum-core",
- "bitflags",
+ "bitflags 1.3.2",
  "bytes",
  "futures-util",
  "http",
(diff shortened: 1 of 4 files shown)


olix0r:
> This branch removes the Git deps and updates to v3.0.0 of boring/tokio-boring

I see the git removal but i do not see the dependency update.

hawkw:
> > This branch removes the Git deps and updates to v3.0.0 of boring/tokio-boring
> 
> 
> 
> I see the git removal but i do not see the dependency update.

Agh, whoops! I'll fix that.

olix0r (approved):

PR #2465, opened 2023-09-05
use `rustls-webpki` instead of `linkerd/webpki`

hawkw:
This commit changes the `linkerd-meshtls-rustls` crate to use the upstream `rustls-webpki` crate, maintained by Rustls, rather than our fork of `briansmith/webpki` from GitHub. Since `rustls-webpki` includes the change which was the initial motivation for the `linkerd/webpki` fork (rustls/webpki#42), we can now depend on upstream.

PR #2467, opened 2023-09-11
meshtls: log errors parsing client certs

hawkw:
Depends on #2465

Currently, if errors occur while parsing a client identity from a TLS certificate, the `client_identity` function in `linkerd-meshtls-rustls` will simply discard the error and return `None`. This means that we cannot easily determine *why* a connection has no client identity --- there may have been no client cert, but we may also have failed to parse a client cert that was present.

In order to make debugging these issues a little easier, I've changed this function to log any errors returned by `rustls-webpki` while parsing client certs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2465,7 +2465,7 @@ dependencies = [
 [[package]]
 name = "rustls-webpki"
 version = "0.101.5"
-source = "git+https://github.com/cpu/webpki?branch=cpu-0.101.5-prep#702d57f444e3f7d743277524e832a2363290ec4d"
+source = "git+https://github.com/cpu/webpki?rev=702d57f444e3f7d743277524e832a2363290ec4d#702d57f444e3f7d743277524e832a2363290ec4d"
 dependencies = [
  "ring",
  "untrusted",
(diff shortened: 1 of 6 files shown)


olix0r (commented):
my understanding is that the libraries already have some onerous logging behavior in these cases. Is that true? Can you test what happens when this behavior is triggered?

hawkw:
> my understanding is that the libraries already have some onerous logging behavior in these cases. Is that true? Can you test what happens when this behavior is triggered?

I believe you're thinking of `rustls` itself, which logs a bunch of stuff from the actual TLS implementation. `rustls-webpki` does not log anything (and we can confirm this by checking its dependency list for the presence of `log` or `tracing`: https://github.com/rustls/webpki/blob/64b97f32efa7ebe87d3e536144b28b04fc6550a5/Cargo.toml#L71-L74). 

These errors would occur if a TLS handshake completed successfully, but we couldn't extract a valid DNS name from the client cert in that handshake. In that case, `rustls` itself wouldn't have logged anything, because the TLS handshake completed successfully.

olix0r (approved):

PR #2484, opened 2023-10-17
balance: Log and fail stuck discovery streams.

olix0r:
In 6d2abbc, we changed how outbound proxies process discovery updates. The prior implementation used a watchdog timeout to bound the amount of time an update stream could be full. With that change, when an update channel fills, the backpressure can extend to the destination controller's gRPC response stream.

To detect and avoid this harmful (and useless) backpressure, this change modifies the balancer's discovery processing stream to exit when the balancer has 1000 unprocessed discovery updates. A sufficiently scary warning is logged.

### DIFF
--- a/linkerd/proxy/balance/src/discover/buffer.rs
+++ b/linkerd/proxy/balance/src/discover/buffer.rs
@@ -2,7 +2,7 @@ use futures_util::future::poll_fn;
 use linkerd_error::Error;
 use tokio::sync::mpsc;
 use tower::discover;
-use tracing::{debug, instrument::Instrument, trace};
+use tracing::{debug, debug_span, instrument::Instrument, trace, warn};
 
 pub type Result<K, S> = std::result::Result<discover::Change<K, S>, Error>;
 pub type Buffer<K, S> = tokio_stream::wrappers::ReceiverStream<Result<K, S>>;
@@ -16,6 +16,25 @@ where
 {
     let (tx, rx) = mpsc::channel(capacity);
 
+    let send = |tx: &mpsc::Sender<_>, change| {
+        if let Err(e) = tx.try_send(change) {
+            match e {
+                // The balancer has been dropped (and will never be used again).
+                mpsc::error::TrySendError::Closed(_) => {
+                    debug!("Discovery receiver dropped");
+                }
+
+                // The balancer is stalled and we can't continue to buffer
+                // updates for it.
+                mpsc::error::TrySendError::Full(_) => {
+                    warn!("The balancer is not processing discovery updates; aborting discovery stream");
+                }
+            }
+            return false;
+        }
+        true
+    };
+
     debug!(%capacity, "Spawning discovery buffer");
     tokio::spawn(
         async move {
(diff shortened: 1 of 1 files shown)


linkerd/proxy/balance/src/discover/buffer.rs
```diff
@@ -16,48 +16,77 @@ where
...
+
+                // The balancer is stalled and we can't continue to buffer
+                // updates for it.
+                mpsc::error::TrySendError::Full(_) => {
+                    warn!("The balancer is not processing discovery updates; aborting discovery stream");
+                }
+            }
+            return false;
+        }
+        true
```
hawkw, line 35: nit, take it or leave it: why not just 
```suggestion
        match tx.try_send(change) {
            // The balancer has been dropped (and will never be used again).
            Err(mpsc::error::TrySendError::Closed(_)) => {
                debug!("Discovery receiver dropped");
                false
            },

            // The balancer is stalled and we can't continue to buffer
            // updates for it.
            Err(mpsc::error::TrySendError::Full(_)) => {
                warn!("The balancer is not processing discovery updates; aborting discovery stream");
                false
            },

            Ok(()) => true,
        }
```
which imo is a little easier to follow

linkerd/proxy/balance/src/discover/buffer.rs
```diff
@@ -16,48 +16,77 @@ where
...
                         debug!(%error);
-                        let _ = tx.send(Err(error)).await;
+                        send(&tx, Err(error));
                         return;
                     }
                     None => {
                         debug!("Discovery stream closed");
                         return;
                     }
                 };
```
hawkw, line 85: nit, take it or leave it: it looks like all these branches _except_ for the `!send(...)` branch will `return`, why not change this to 
```rust
match res {
    Some(Ok(change)) => {
        trace!("Changed");
        if send(&tx, Ok(change)) {
            continue;
        }
    }
    Some(Err(e)) => {
        let error = e.into();
        debug!(%error);
        send(&tx, Err(error));
    }
    None => {
        debug!("Discovery stream closed");
    }
};

return;
```

linkerd/proxy/balance/src/discover/buffer.rs
```diff
@@ -16,48 +16,77 @@ where
...
+                            // balancer to get cut off from further updates,
+                            // should it ever become available again. That needs
+                            // to be fixed.
+                            //
+                            // One option would be to drop the discovery stream
+                            // and rebuild it if the balancer ever becomes
+                            // unblocked.
+                            //
+                            // Ultimately we need to track down how we're
+                            // getting into this blocked/idle state
```
hawkw, line 71: should we open a new issue to track this?

linkerd/proxy/balance/src/discover/buffer.rs
```diff
@@ -16,48 +16,77 @@ where
 {
     let (tx, rx) = mpsc::channel(capacity);
 
+    let send = |tx: &mpsc::Sender<_>, change| {
```
hawkw, line 19: nit, not a big deal: it would be nice if there was a comment here saying what the bool returned by this closure means? something like:
```suggestion
    // Attempts to send an update to the balancer, returning `true` if we should continue
    // processing discovery updates, or `false` if the balancer is dead and this task should 
    // terminate.
    let send = |tx: &mpsc::Sender<_>, change| {
```

hawkw: also, not a big deal, but it seems like this could maybe be a free function rather a closure, since AFAICT it doesn't actually close over anything in scope here?  the closure might result in more code generation than a free function...but this probably doesn't really matter.

hawkw (approved):
looks good to me! i left some very minor nits.

linkerd/proxy/balance/src/discover/buffer.rs
olix0r: In my mind all of the returning cases are exceptional. The golden case is that the send works and we loop through

linkerd/proxy/balance/src/discover/buffer.rs
olix0r: The free function is needlessly verbose. Biasing for legibility.

linkerd/proxy/balance/src/discover/buffer.rs
olix0r: We are working on existing issues that capture user-facing problems related to this functionality. They cover tracking down how we're getting into this blocked state, which is the important part of this.

linkerd/proxy/balance/src/discover/buffer.rs
hawkw: fair enough!

PR #2486, opened 2023-10-19
balance: Fail the discovery stream on queue backup

olix0r:
328826caa updated the balancer's discovery channel to prevent backing up into the discovery stream by dropping the discovery stream. This results in balancers becoming permanently stale (should they ever be used again).

This change modifies the discovery stream so that these errors are fatal for the balancer.

To fix this, we replace the `DiscoverNew` module with a `discover::NewServices` module that wraps the buffering layer. The buffer now only holds target metadata, and services are only built as the entry is dequeued from channel.

This has the (positive) side-effect that the proxy's stack_create_total metric will not be incremented before the balancer actually uses an endpoint stack. Previously, this metric would be incremented for all queued endpoint updates.

We also now log at INFO the address of all additions and removals from a balancer. This should dramatically improve diagnostics in stale endpoint situations.


linkerd/proxy/balance/src/discover/buffer.rs
```diff
@@ -60,19 +68,9 @@ where
...
-                            //
-                            // One option would be to drop the discovery stream
-                            // and rebuild it if the balancer ever becomes
-                            // unblocked.
-                            //
-                            // Ultimately we need to track down how we're
-                            // getting into this blocked/idle state
+                            // Tell the outer discovery stream that we've
+                            // stopped processing updates.
+                            overflow.store(true, std::sync::atomic::Ordering::Release);
```
hawkw, line 73: one potential concern i had while reviewing this is that the `NewServices` task isn't _woken_ when we set the overflow flag, we just set the flag. so, i wondered if it would be better to use a `Notify` or something here and have the `poll_next` for `NewServices` try to poll that `Notify`. but, i think this is actually fine, since we immediately exit here, which should result in dropping the MPSC's only sender and waking the `poll_next` call, and we do handle that case correctly in `NewServices`' `Stream` impl.

i think it might be nice if there was a comment explaining this, though, so that we don't accidentally change this code in a way that eats the potential wakeup in the future.

linkerd/proxy/balance/src/discover.rs
```diff
@@ -1,31 +1,85 @@
...
+            .load(std::sync::atomic::Ordering::Acquire)
+        {
+            return Poll::Ready(Some(Err(DiscoveryStreamOverflow(()).into())));
+        };
+
+        // Process any buffered updates.
+        let change_tgt = match ready!(this.buffer.rx.poll_recv(cx)) {
+            Some(Ok(c)) => c,
+            Some(Err(e)) => return Poll::Ready(Some(Err(e))),
+            None => return Poll::Ready(Some(Err(DiscoveryStreamOverflow(()).into()))),
```
hawkw, line 66: we drop the `mpsc::Sender` in two cases: one, when we observe an actual error, which we send before we drop the sender, and two, when the stream overflows. so, in the error case, we'll receive the error first here and return...but the subsequent `poll_next` call will return the `DiscoveryStreamOverflow` error rather than `None` to indicate that the stream terminated due to an unrecoverable error. 

i don't see any obvious issues that result from that behavior, but wanted to point it out. we could, potentially, add a boolean that tracks whether we've seen an error from the stream or not, and only return `DiscoveryStreamOverflow` on `None`s if we haven't. or, we could change the `None` arm to check the `buffer.overflow` bool and only return the overflow error if it's been set...i'm not sure if this is actually necessary, but i wanted to draw attention to it.

hawkw (approved):
this looks great to me overall! i had a couple small questions about the `Stream` implementation, but i don't think either of them are major issues.

linkerd/proxy/balance/src/discover/buffer.rs
olix0r: What task would this be waking? If the consumer is in a place where the buffer overflows, it's stuck in _ready_ so it can't be woken up. What we're doing here is to ensure that the next time it's used, it fails.

linkerd/proxy/balance/src/discover.rs
olix0r: > in the error case, we'll receive the error first here and return...but the subsequent poll_next call

If the stream returns an error there will not be a subsequent call. Errors are considered fatal.

linkerd/proxy/balance/src/discover/buffer.rs
hawkw: yeah, you're right. never mind!

PR #2487, opened 2023-10-19
chore: change `rust-toolchain` file to toml format

hawkw:
The `rust-toolchain` file containing only a Rust version number is deprecated in favor of a TOML-formatted `rust-toolchain.toml`. Using the old format seems to make Dependabot unhappy --- it complains that:

```
only rust-toolchain files formatted as TOML are supported, the non-TOML
format was deprecated by Rust
```

Therefore, this branch changes the toolchain file in this repo to the TOML format.


hawkw:
okay this works now

olix0r (approved):

PR #2491, opened 2023-10-25
gate: Detect disconnected inner services in readiness

olix0r:
If `Gate` becomes ready, it assumes the inner service remains ready indefinitely.

Load balancers rely on lazy and redudant readiness checking to avoid disconnected endpoints.

This change fixes the Gate to ensure that the inner service is always polled whenever the gate is polled.

### DIFF
--- a/linkerd/stack/src/gate.rs
+++ b/linkerd/stack/src/gate.rs
@@ -173,11 +173,12 @@ where
     type Future = S::Future;
 
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        if self.permit.is_ready() {
-            return Poll::Ready(Ok(()));
-        }
+        // If we previously polled to ready and acquired a permit, clear it so
+        // we can reestablish readiness without holding it.
+        self.permit = Poll::Pending;
         let permit = ready!(self.poll_acquire(cx));
         ready!(self.inner.poll_ready(cx))?;
+        tracing::trace!("Acquired permit");
         self.permit = Poll::Ready(permit);
         Poll::Ready(Ok(()))
     }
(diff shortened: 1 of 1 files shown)


olix0r:
It looks like this behavior was specifically added in https://github.com/linkerd/linkerd2-proxy/commit/172523fb083d80e1b508ae98c433251433344daa. Based on my reading of that commit message, it seems like this change fixes this by dropping the stored permit.

linkerd/stack/src/gate.rs
```diff
@@ -173,11 +173,12 @@ where
     type Future = S::Future;
 
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        if self.permit.is_ready() {
-            return Poll::Ready(Ok(()));
-        }
+        // If we previously polled to ready and acquired a permit, clear it so
+        // we can reestablish readiness without holding it.
+        self.permit = Poll::Pending;
```
hawkw, line 178: thanks for the comment, this makes sense to me.

linkerd/stack/src/gate.rs
```diff
@@ -262,6 +264,47 @@ mod tests {
         assert_ready!(gate.poll_ready()).expect("ok");
     }
 
+    #[tokio::test]
+    async fn gate_repolls_back_to_pending() {
```
hawkw, line 268: i think this behavior is complex enough that it would be nice to have a comment here explaining what specific properties we are testing for here?

hawkw (approved):

hawkw:
clippy failure is just a timeout: https://github.com/linkerd/linkerd2-proxy/actions/runs/6645201336/job/18055979156?pr=2491

PR #2493, opened 2023-10-26
gate: Fix readiness deadlock

olix0r:
The gate middleware controls a service's readiness so that it can exert backpressure. This is used, for instance, by the circuit breaker module so that an endpoint can go into an unavailable state after the breaker has been tripped and be marked available again as it recovers.

This change fixes a bug in that recovery scenario: when the gate is in a Limited state (i.e. when the circuit breaker puts an endpoint into Probation to test its availability), and caller (i.e. the balancer) is waiting for the endpoint to leave probation, the balancer may never be notified that the endpoint has left its probation state.

To fix this, we update the gate controller to definitively close its inner Semaphore when transitioning out of a limited state -- dropping the semaphore in the sender doesn't close it when it's being held by a receiver. This change motivates an update to the gate::Tx API so that it always produces a Semaphore when transitioning to the limited state (rather than take a shared semaphore as a caller).

This issue is somewhat masked by the balancer's polling behavior, where endpoint states are only advanced as requests are processed. It seems likely, however, that this scenario could be encountered in the wild when circuit breaking is enabled on a service.

### DIFF
--- a/linkerd/app/outbound/src/http/breaker/consecutive_failures.rs
+++ b/linkerd/app/outbound/src/http/breaker/consecutive_failures.rs
@@ -1,14 +1,12 @@
 use futures::stream::StreamExt;
 use linkerd_app_core::{classify, exp_backoff::ExponentialBackoff, proxy::http::classify::gate};
-use std::sync::Arc;
-use tokio::sync::{mpsc, Semaphore};
+use tokio::sync::mpsc;
 
 pub struct ConsecutiveFailures {
     max_failures: usize,
     backoff: ExponentialBackoff,
     gate: gate::Tx,
     rsps: mpsc::Receiver<classify::Class>,
-    semaphore: Arc<Semaphore>,
 }
 
 impl ConsecutiveFailures {
@@ -23,7 +21,6 @@ impl ConsecutiveFailures {
             backoff,
             gate,
             rsps,
-            semaphore: Arc::new(Semaphore::new(0)),
         }
     }
 
@@ -46,7 +43,7 @@ impl ConsecutiveFailures {
     /// observed.
     async fn open(&mut self) -> Result<(), ()> {
         tracing::debug!("Open");
-        self.gate.open();
+        self.gate.open().map_err(|_| ())?;
         let mut failures = 0;
         loop {
             let class = tokio::select! {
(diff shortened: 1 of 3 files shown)


hawkw:
Integration timeout looks sketchy... https://github.com/linkerd/linkerd2-proxy/actions/runs/6656142954/job/18088020951?pr=2493

linkerd/stack/src/gate.rs
```diff
@@ -115,24 +123,60 @@ impl Tx {
...
-            debug!("Gate opened");
+    pub fn open(&self) -> Result<(), Closed> {
+        if self.0.is_closed() {
+            return Err(Closed(()));
+        }
+        match self.0.send_replace(State::Open) {
+            State::Open => {}
+            State::Limited(lim) => {
+                debug!("Limited => Open");
+                lim.close();
```
hawkw, line 134: I think it might be nice if we documented at the top level that a `Limited` state's semaphore will be closed when the gate transitions back to open --- maybe add a rustdoc comment about that to the `State::Limited` variant?

linkerd/stack/src/gate.rs
```diff
@@ -115,24 +123,60 @@ impl Tx {
...
+            }
+            State::Shut => {
+                debug!("Shut => Open");
+            }
         }
+        Ok(())
     }
 
     /// Limits the gate with the provided semaphore.
```
hawkw, line 143: could we maybe add something explicitly mentioning what this method does when the gate is already in the limited state?

linkerd/app/outbound/src/http/breaker/consecutive_failures.rs
```diff
@@ -97,8 +94,8 @@ impl ConsecutiveFailures {
     /// Wait for a response to determine whether the breaker should be opened.
     async fn probation(&mut self) -> Result<classify::Class, ()> {
         tracing::debug!("Probation");
-        self.semaphore.add_permits(1);
-        self.gate.limit(self.semaphore.clone());
+        let sem = self.gate.limit().map_err(|_| ())?;
+        sem.add_permits(1);
```
hawkw, line 98: one thing i noticed here is that this change does mean that when the gate transitions to the limited state, the permits are not already present in the semaphore, so another task may be woken by the state transition, and then immediately yield again when it tries to actually wait on the semaphore. i don't think this is an actual correctness issue, since we will immediately re-wake that task by adding the permit, but it might be slightly less efficient, since we could do an unneeded round-trip through the scheduler.

we might want to have the `limit` method take a number of permits and create the new semaphore with that number of permits, so that when we actually transition the state, the permits are already present? 

not a big deal either way, just a thought.

hawkw (approved):
This looks correct to me, and the change is pretty straightforward! New tests look good, too. It might be nice to document the behavior around closing the semaphore, but no blockers from me.

PR #2496, opened 2023-11-01
inbound: Fix gRPC response classification

olix0r:
ecaaf39 changed the proxy's behavior with regard to creating [default response classifiers][default]: the defaults used to support detecting gRPC response (regardless of the request properties).

To fix this, we modify the metrics module that uses responses classifiers to *require* them without inferring defaults. This enforces the intended usage pattern so that we do not silently and implicitly fall back to the default behavior. This requirement is only strictly enforced in development builds. In release builds, the prior defaulting behavior remains in place. This minimizes the difference from the existing behavior, especially if we hit untested cases where a hard failure would be untenable (especially in metrics code).

This change also updates the `NewClassify` module that inserts the response classifier request extension so that overrides are supported. We then can install a default classifier early in request processing and override it only if specified by a route configuration.

To support this change, the http-metrics crate is updated to support querying response_total metrics without stringifying everything.

[default]: https://github.com/linkerd/linkerd2-proxy/commit/ecaaf39b4651540548e835613855cf7c39895969#diff-372e8a8a57b1fad5d94f37d2f77fdc7a45bcf708782475424b75d671f99ea1a0L97-L103

### DIFF
--- a/linkerd/app/inbound/src/http/tests.rs
+++ b/linkerd/app/inbound/src/http/tests.rs
@@ -475,7 +475,7 @@ async fn grpc_unmeshed_response_error_header() {
 async fn grpc_response_class() {
     let _trace = trace_init();
 
-    // Build a mock connector that returns
+    // Build a mock connector serves a gRPC server that returns errors.
     let connect = {
         let mut server = hyper::server::conn::Http::new();
         server.http2_only(true);
--- a/linkerd/app/inbound/src/http/tests.rs
+++ b/linkerd/app/inbound/src/http/tests.rs
@@ -6,10 +6,11 @@ use crate::{
     },
     Config, Inbound,
 };
-use hyper::{client::conn::Builder as ClientBuilder, Body, Request, Response};
+use hyper::{body::HttpBody, client::conn::Builder as ClientBuilder, Body, Request, Response};
 use linkerd_app_core::{
+    classify,
     errors::respond::L5D_PROXY_ERROR,
-    identity, io,
+    identity, io, metrics,
     proxy::http,
     svc::{self, NewService, Param},
     tls,
@@ -19,6 +20,7 @@ use linkerd_app_core::{
 use linkerd_app_test::connect::ConnectFuture;
 use linkerd_tracing::test::trace_init;
 use std::{net::SocketAddr, sync::Arc};
+use tokio::time;
 use tracing::Instrument;
 
 fn build_server<I>(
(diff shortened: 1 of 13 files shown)


linkerd/app/inbound/src/http/tests.rs
```diff
@@ -469,6 +471,84 @@ async fn grpc_unmeshed_response_error_header() {
     let _ = bg.await;
 }
 
+#[tokio::test(flavor = "current_thread")]
+async fn grpc_response_class() {
+    let _trace = trace_init();
+
+    // Build a mock connector that returns
```
hawkw, line 478: tiny nit: looks like this comment wasn't finished?

linkerd/http-metrics/src/lib.rs
```diff
@@ -25,6 +25,22 @@ where
     include_latencies: bool,
 }
 
+#[cfg(feature = "test-util")]
+impl<T: Hash + Eq, C: Hash + Eq> Report<T, requests::Metrics<C>> {
+    pub fn get_response_total(
```
hawkw, line 30: good call :+1:

linkerd/proxy/http/src/classify/insert.rs
```diff
@@ -61,8 +67,9 @@ where
 
     fn proxy(&self, svc: &mut S, mut req: http::Request<B>) -> Self::Future {
         let classify_rsp = self.classify.classify(&req);
-        let prior = req.extensions_mut().insert(classify_rsp);
-        debug_assert!(prior.is_none(), "classification extension already existed");
+        if req.extensions_mut().insert(classify_rsp).is_some() {
+            tracing::debug!("Overrode response classifier");
```
hawkw, line 71: nit, take it or leave it: maybe worth having this print what the previous classifier was?
```suggestion
        if let Some(prev) = req.extensions_mut().insert(classify_rsp) {
            tracing::debug!(?prev, "Overrode response classifier");
```

linkerd/http-metrics/src/requests/service.rs
```diff
@@ -160,13 +168,14 @@ where
     S::Error: Into<Error>,
     A: Body,
     B: Body,
-    C: ClassifyResponse + Clone + Default + Send + Sync + 'static,
+    C: ClassifyResponse + Default + Clone + Send + Sync + 'static,
```
hawkw, line 171: tiny nit: did this get de-alphabetized for a reason?

hawkw (approved):
this change looks good to me! good call on including debug assertions to help catch regressions in the tests but not take down the whole proxy in prod!

linkerd/proxy/http/src/classify/insert.rs
olix0r: that requires adding `Debug` bounds which felt a little larger than desired.

PR #2499, opened 2023-11-01
dst: Log a warning when the controller clients receive an error

olix0r:
The controller client includes a recovery/backoff module that causes resolutions to be retried when an unexpected error is encountered. These events are only logged at debugging and trace log levels.

This change updates the destination controller recovery module to log unexpected errors. This will benefit both profile and endpoint resolution.


olix0r:
Followup for https://github.com/linkerd/linkerd2/issues/11555

hawkw (approved):

PR #2503, opened 2023-11-06
split tls id and name

zaharidichev:
This change attempts to split the concept of a server name(SNI) and Tls identity(the identity of a peer as prescribed by its leaf TLS cert)


olix0r:
subsumed by #2506

PR #2507, opened 2023-11-08
Extract TLS id verification out of TLS backends

zaharidichev:
To further prepare for the introduction of Spiffe IDs in the proxy, 
this change disables TLS SAN verification in both rustls and boring
TLS implementations. The need to do that stems from the fact that
these implementations only look at DNS SANs and cannot handle
certificates that use URI SANs. 

To enable adding this functionality, TLS Id verification is not extracted
as a separate step, after the TLS handshake completes. To do that in
boring, we need to `verify_hostname`. For rustls, we need to pass a
custom `ServerCertVerifier` that skips name validation. In our case, the
verifier is nearly identical to `ustls::client::WebPkiVerifier`, except for
the step where DNS SANs are validated based on the `ServerName`.

This change is non functional, but it sets the stage to have more control
over TLS Id verification.


linkerd/meshtls/rustls/src/creds/store.rs
```diff
@@ -236,27 +242,6 @@ impl rustls::server::ResolvesServerCert for CertResolver {
...
-
-        // Verify that our certificate is valid for the given SNI name.
-        let c = self.0.cert.first()?;
-        if let Err(error) = webpki::EndEntityCert::try_from(c.as_ref())
-            .and_then(|c| c.verify_is_valid_for_subject_name(server_name))
-        {
-            debug!(%error, "Local certificate is not valid for SNI");
-            return None;
-        };
-
```
zaharidichev, line 259: This is removed, because if we assume that our certificate can holds its Id in either URI or DNS SAN, we cannot pose the restriction that the Server Name that is contained in the ClientHello matches the value of the SAN. This further begs the question of whether we should use a resolver at all. An alternative approach is to just set the certificates that will be used when we construct the tls config.

linkerd/meshtls/rustls/src/creds/verify.rs
```diff
@@ -0,0 +1,56 @@
...
+use std::{io, sync::Arc};
+use tokio_rustls::rustls::{
+    self,
+    client::{self, ServerCertVerified, ServerCertVerifier},
+    server::ParsedCertificate,
+    Certificate, RootCertStore, ServerName,
+};
+use tracing::trace;
+
+pub(crate) struct AnySANVerifier(Arc<RootCertStore>);
```
zaharidichev, line 13: Not too sure about the naming.

olix0r (commented):
I have a few low-level suggestions I'll make in a followup, but my initial high-level reaction is (1) this is great and (2) `verify_id` needs tests :)

We need to ensure that both of our verify_id implementations have the same behavior for a set of Certificate & Id pairs. Note that the linkerd-tls-test-util crate provides some readymade certificates that can be used for testing.

The key point is: by taking this verification logic out of the library--even though we're just copying an implementation in the short term--it becomes our responsibility to ensure correctness going forward.

To that end, perhaps there are rustls tests we can incorporate?

zaharidichev:
@olix0r That makes sense. I will add tests for these in my tomorrow.

hawkw:
> We need to ensure that both of our verify_id implementations have the same behavior for a set of Certificate & Id pairs. Note that the linkerd-tls-test-util crate provides some readymade certificates that can be used for testing.

If we need test certs that aren't those already checked in, we may also want to consider the [`rcgen` crate](https://docs.rs/rcgen/latest/rcgen/) to generate certificates for tests inside the test code, rather than checking in test key material to Git...

linkerd/meshtls/boring/src/client.rs
```diff
@@ -76,25 +81,36 @@ where
             let conn = connector.map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
             let config = conn
                 .configure()
-                .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
-            let io = tokio_boring::connect(config, server_name.as_str(), io)
+                .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?
+                // Switch off DNS san Verification based on the hostname
```
hawkw, line 85: tiny nit: I think SAN is also an acronym (for "Subject Alternate Name"):
```suggestion
                // Switch off DNS SAN Verification based on the hostname
```

linkerd/meshtls/boring/src/client.rs
```diff
@@ -76,25 +81,36 @@ where
...
+                        alpn = ?io.ssl().selected_alpn_protocol(),
+                        "Initiated TLS connection"
+                    );
+                    Ok(ClientIo(io))
+                },
+                         None => Err(io::Error::new(
+                    io::ErrorKind::Other,
+                    "could not extract peer cert",
+                )),
+                })
```
hawkw, line 113: some weird formatting going on here...i think it should look more like this:
```suggestion
                }).and_then(|io| match io.ssl().peer_certificate() {
                    Some(ref cert) => {
                        verify::verify_id(cert, &server_id)
                            .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
                        debug!(
                            tls = io.ssl().version_str(),
                            client.cert = ?io.ssl().certificate().and_then(super::fingerprint),
                            peer.cert = ?io.ssl().peer_certificate().as_deref().and_then(super::fingerprint),
                            alpn = ?io.ssl().selected_alpn_protocol(),
                            "Initiated TLS connection"
                        );
                        Ok(ClientIo(io))
                    },
                    None => Err(io::Error::new(
                        io::ErrorKind::Other,
                        "could not extract peer cert",
                    )),
                })
```

have we run `rustfmt` on this branch?

linkerd/meshtls/boring/src/creds/verify.rs
```diff
@@ -0,0 +1,21 @@
+use boring::x509::X509;
+use linkerd_dns_name as dns;
+use linkerd_identity as id;
+use std::io;
+
+pub(crate) fn verify_id(cert: &X509, id: &id::Id) -> io::Result<()> {
+    for san in cert.subject_alt_names().into_iter().flatten() {
+        if let Some(n) = san.dnsname() {
+            if let Ok(name) = n.parse::<dns::Name>() {
```
hawkw, line 9: if the value we get from `san.dnsname()` doesn't parse as a DNS name, it looks like we just skip it...should we maybe log a warning when we skip a subject alternate name "DNS name" that didn't parse as a valid DNS name?

linkerd/meshtls/boring/src/creds/verify.rs
```diff
@@ -0,0 +1,21 @@
...
+                if *name == id.to_str() {
+                    return Ok(());
+                }
+            }
+        }
+    }
+
+    Err(io::Error::new(
+        io::ErrorKind::Other,
+        "certificate does not match tls id",
```
hawkw, line 19:
```suggestion
        "certificate does not match TLS identity",
```

linkerd/meshtls/boring/src/creds/verify.rs
```diff
@@ -0,0 +1,21 @@
+use boring::x509::X509;
+use linkerd_dns_name as dns;
+use linkerd_identity as id;
+use std::io;
+
+pub(crate) fn verify_id(cert: &X509, id: &id::Id) -> io::Result<()> {
```
hawkw, line 6: i don't love that this returns an `io::Result`, since the only errors we return here are not, actually, IO errors. should we maybe have a new error type for `verify_id`?

linkerd/meshtls/rustls/src/client.rs
```diff
@@ -72,15 +72,23 @@ impl Connect {
...
 
         Self {
+            server_id: client_tls.server_id.into(),
             server_name,
             config,
         }
     }
 }
 
+fn extract_cert(c: &rustls::ClientConnection) -> io::Result<&rustls::Certificate> {
```
hawkw, line 82: i'm not sure if `io::Error` is really the right return type for this, since the error is not actually IO-related...should we maybe instead change the `Service::Error` associated type for `Connect`? Or, does that break some other thing?

linkerd/meshtls/rustls/src/client.rs
```diff
@@ -91,9 +99,21 @@ where
...
+                .connect(self.server_name.clone(), io)
+                .map(move |s| match s {
+                    Ok(s) => {
+                        let (_, conn) = s.get_ref();
+                        let end_cert = extract_cert(conn)?;
+                        verify::verify_id(end_cert, &server_id)?;
+                        Ok(ClientIo(s))
+                    }
+                    Err(err) => Err(err),
+                }),
```
hawkw, line 115: nit, take it or leave it: rather than using a `match`, we could just use the `?` operator here to propagate the error from `s`:
```suggestion
                .map(move |s|  {
                    let (_, conn) = s?.get_ref();
                    let end_cert = extract_cert(conn)?;
                    verify::verify_id(end_cert, &server_id)?;
                    Ok(ClientIo(s))
                }),
```

linkerd/meshtls/rustls/src/creds/verify.rs
```diff
@@ -0,0 +1,56 @@
...
+        _: &mut dyn Iterator<Item = &[u8]>,
+        ocsp_response: &[u8],
+        now: SystemTime,
+    ) -> Result<ServerCertVerified, rustls::Error> {
+        let cert = ParsedCertificate::try_from(end_entity)?;
+
+        client::verify_server_cert_signed_by_trust_anchor(&cert, &self.0, intermediates, now)?;
+
+        if !ocsp_response.is_empty() {
+            trace!("Unvalidated OCSP response: {:?}", ocsp_response.to_vec());
```
hawkw, line 39: we shouldn't need to turn the OCSP response into a `Vec` just to format it, AFAIK:
```suggestion
            trace!("Unvalidated OCSP response: {ocsp_response:?}");
```

linkerd/meshtls/rustls/src/creds/verify.rs
hawkw, line 13: nitpicky, but the general Rust convention is that acronyms in type names should only have the first letter capitalized (see https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case); so this should be
```suggestion
pub(crate) struct AnySanVerifier(Arc<RootCertStore>);
```

linkerd/meshtls/boring/src/creds/verify.rs
olix0r: i believe all of these errors need to make it to io::Error given where this occurs in the accept path.

hawkw: ahh, ugh, nevermind. this is fine, i guess!

hawkw: it just seems unfortunate for these to be turned into "IO errors" in the error metrics...

linkerd/meshtls/boring/src/client.rs
```diff
@@ -76,25 +81,36 @@ where
...
+                            alpn = ?io.ssl().selected_alpn_protocol(),
+                            "Initiated TLS connection"
+                        );
+                        Ok(ClientIo(io))
+                    },
+                    None => Err(io::Error::new(
+                        io::ErrorKind::Other,
+                        "could not extract peer cert",
+                    )),
+                })
```
hawkw, line 113: style nit, take it or leave it: rather than matching on `peer_certificate()` here, we could use [`Option::ok_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else) and bail early using `?`, allowing the code here to focus more clearly on the happy path:
```suggestion
                }).and_then(|io| {
                    let cert = io.ssl()
                        .peer_certificate()
                        .ok_or_else(|| Err(io::Error::new(
                            io::ErrorKind::Other,
                            "could not extract peer cert",
                        )))?;

                    verify::verify_id(cert, &server_id)
                        .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
                    debug!(
                        tls = io.ssl().version_str(),
                        client.cert = ?io.ssl().certificate().and_then(super::fingerprint),
                        peer.cert = ?io.ssl().peer_certificate().as_deref().and_then(super::fingerprint),
                        alpn = ?io.ssl().selected_alpn_protocol(),
                        "Initiated TLS connection"
                    );
                    Ok(ClientIo(io))
                })
```

linkerd/meshtls/boring/src/creds/verify.rs
```diff
@@ -0,0 +1,78 @@
...
+    for san in cert.subject_alt_names().into_iter().flatten() {
+        if let Some(n) = san.dnsname() {
+            if let Ok(name) = n.parse::<dns::Name>() {
+                if *name == id.to_str() {
+                    return Ok(());
+                }
+            }
+            tracing::warn!("DNS SAN name {} could not be parsed", n)
```
hawkw, line 14: if we don't `return` because the name doesn't equal the identity, we will also fall through and print "DNS SAN name could not be parsed", even though that's not actually true in that case. i think this code should probably be more like this:

```suggestion
        match n.parse::<dns::Name>() {
            Ok(name) if *name == id.to_str() => return Ok(()),
            Ok(_) => {},
            Err(error) => tracing::warn!(%error, "DNS SAN name {} could not be parsed", n),
        }
```

linkerd/meshtls/boring/src/creds/verify.rs
```diff
@@ -0,0 +1,78 @@
...
+#[cfg(test)]
+mod tests {
+    use super::verify_id;
+    use boring::x509::X509;
+    use linkerd_identity as id;
+    use rcgen::generate_simple_self_signed;
+    use std::str::FromStr;
+
+    fn generate_cert_with_name(name: Option<&str>) -> X509 {
+        let sans = name.map(|s| vec![s.into()]).unwrap_or_default();
```
hawkw, line 33: it looks like we don't have any tests for certificates with multiple SANs --- we only ever generate certs with 0-1 SANs. since we have logic for iterating over multiple matching and non-matching SANs, we should probably have tests for certs with more than one SAN...

linkerd/meshtls/rustls/src/creds/verify.rs
```diff
@@ -0,0 +1,112 @@
...
+#[cfg(test)]
+mod tests {
+    use super::verify_id;
+    use linkerd_identity as id;
+    use rcgen::generate_simple_self_signed;
+    use std::str::FromStr;
+    use tokio_rustls::rustls::Certificate;
+
+    fn generate_cert_with_name(name: Option<&str>) -> Certificate {
+        let sans = name.map(|s| vec![s.into()]).unwrap_or_default();
```
hawkw, line 67: similarly to my suggestion on the boring module, it would be good to test certs with multiple matching and non-matching SANs.

linkerd/meshtls/rustls/src/creds/verify.rs
```diff
@@ -0,0 +1,112 @@
...
+        assert!(verify_id(&cert, &id).is_err());
+    }
+
+    #[test]
+    fn cert_with_no_san_does_not_verify_for_dns_id() {
+        let dns_name = "bar.ns1.serviceaccount.identity.linkerd.cluster.local";
+        let cert = generate_cert_with_name(None);
+        let id = id::Id::from_str(dns_name).expect("should parse DNS id");
+        assert!(verify_id(&cert, &id).is_err());
+    }
```
hawkw, line 111: take it or leave it: the tests in this crate are identical to the tests in the `meshtls-boring` crate...we might want to factor the body of the tests out into a feature-flagged test support module in `linkerd-meshtls`, which both `linkerd-meshtls-rustls` and `linkerd-meshtls-boring` can depend on for testing their respective implementations?

this is a bit of extra work, so it may not really be worth it, but it does seem potentially nice to be able to ensure that both implementations are being tested with all the same test code?

linkerd/meshtls/src/verify_tests.rs
hawkw, line 1: i have a bit of a preference against using macros to generate tests, as this can sometimes result in confusing stack traces when the macro-generated tests fail.

as an alternative approach, would it be possible to have public test helper functions in this module which take a `fn` that performs the actual verification as an argument? Something like:

```rust
        fn generate_cert_with_names<Cert>(names: Vec<&str>, vec_to_cert: impl FnOnce(Vec<u8>) -> Cert) -> $C {
            let sans: Vec<String> = names.into_iter().map(|s| s.into()).collect();

            let cert_data = generate_simple_self_signed(sans)
                .expect("should generate cert")
                .serialize_der()
                .expect("should serialize");

            vec_to_cert(cert_data)
        }


        pub fn cert_with_dns_san_matches_dns_id<C>(
            verify_id: impl FnOnce(&C, &Id),
            vec_to_cert: impl FnOnce(Vec<u8> -> C),
         ) {
            let dns_name = "foo.ns1.serviceaccount.identity.linkerd.cluster.local";
            let cert = generate_cert_with_names(vec![dns_name], vec_to_cert);
            let id = id::Id::from_str(dns_name).expect("should parse DNS id");
            assert!(verify_id(&cert, &id).is_ok());
        }
```
and then have the respective TLS implementation crates call into these tests like:
```rust
// in `linkerd-meshtls-rustls`

fn vec_to_cert(data: Vec<u8>) -> Certificate {
    Certificate(data)
}

#[test]
fn cert_with_dns_san_matches_dns_id() {
    linkerd_meshtls::verify_tests::cert_with_dns_san_matches_dns_id(verify_id, vec_to_cert);
}
```

This does mean we would have to ensure that any new tests added to `linkerd_meshtls::verify_certs` are also added to both meshtls implementation crates, but I think that's okay, if it gets us panic messages that are more comprehensible...

zaharidichev: This makes sense, I did not know about the problem with the confusing stack traces. Had it implemented with a helper func initially, but ended up going with macros to avoid the need to sync test files in the end as you point out. Will revert though.

olix0r:
Note that we're going to have to allow duplicate versions of base64 in deny.toml

linkerd/meshtls/Cargo.toml
```diff
@@ -25,6 +25,7 @@ linkerd-meshtls-boring = { path = "boring", optional = true }
 linkerd-meshtls-rustls = { path = "rustls", optional = true }
 linkerd-stack = { path = "../stack" }
 linkerd-tls = { path = "../tls" }
+rcgen = "0.11.3"
```
olix0r, line 28: It looks like we're now taking on a new runtime dependency for this test code. That's undesirable. We want to make sure this dependency can be isolated to the dev-dependency graph so it isn't needed in release builds.

I think we should also avoid the circular dependency introduced here, whereby the meshtls/rustls and mesthls/boring tests now depend on the higher-level meshtls crate.

Generally, the way I'd recommend approaching this is to add a new `meshtls/test-util` crate with the helpers in its lib.rs.

Then `meshtls/boring` and `meshtls/rustls` each take dev-dependency on ../test-util.

linkerd/meshtls/boring/src/client.rs
```diff
@@ -76,25 +81,36 @@ where
...
+                        .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
+                    debug!(
+                        tls = io.ssl().version_str(),
+                        client.cert = ?io.ssl().certificate().and_then(super::fingerprint),
+                        peer.cert = ?io.ssl().peer_certificate().as_deref().and_then(super::fingerprint),
+                        alpn = ?io.ssl().selected_alpn_protocol(),
+                        "Initiated TLS connection"
+                    );
+                    Ok(ClientIo(io))
+                })
```
olix0r, line 113: I would generally advise against using chained calls (i.e. and_then) if they don't add clarity and needlessly nest logic. I also think this verification logic warrants a clearer overview here.

It's also seems appropriate to log the server name and id at trace level, at least, so the values can be inspected in a pinch.

Suggested as:

```rust
        Box::pin(async move {
            let config = connector
                .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?
                .configure()
                .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;


            // Establish a TLS connection to the server using the provided
            // `server_name` as an SNI value to the server.
            //
            // Hostname verification is DISABLED, as we do not require that the
            // peer's certificate actually matches the `server_name`. Instead,
            // the `server_id` is used to perform the appropriate form of
            // verification after the session is established.
            let io = tokio_boring::connect(config.verify_hostname(false), server_name.as_str(), io)
                .await
                .map_err(|e| match e.as_io_error() {
                    // TODO(ver) boring should let us take ownership of the error directly.
                    Some(ioe) => io::Error::new(ioe.kind(), ioe.to_string()),
                    // XXX(ver) to use the boring error directly here we have to
                    // constrain the socket on Sync + std::fmt::Debug, which is
                    // a pain.
                    None => io::Error::new(io::ErrorKind::Other, "unexpected TLS handshake error"),
                })?;

            // Servers must present a peer certificate. We extract the x509 cert
            // and verify it manually against the `server_id`.
            let cert = io.ssl().peer_certificate().ok_or_else(|| {
                io::Error::new(io::ErrorKind::Other, "could not extract peer cert")
            })?;
            verify::verify_id(&cert, &server_id)
                .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;

            debug!(
                tls = io.ssl().version_str(),
                client.cert = ?io.ssl().certificate().and_then(super::fingerprint),
                peer.cert = ?io.ssl().peer_certificate().as_deref().and_then(super::fingerprint),
                alpn = ?io.ssl().selected_alpn_protocol(),
                "Initiated TLS connection"
            );
            trace!(peer.id = %server_id, peer.name = %server_name)
            Ok(ClientIo(io))
        })
```

linkerd/meshtls/rustls/src/creds/verify.rs
```diff
@@ -0,0 +1,97 @@
...
+
+pub(crate) struct AnySanVerifier(Arc<RootCertStore>);
+
+impl AnySanVerifier {
+    pub(crate) fn new(roots: impl Into<Arc<RootCertStore>>) -> Self {
+        Self(roots.into())
+    }
+}
+
+impl ServerCertVerifier for AnySanVerifier {
```
olix0r, line 21: Presumably this logic originated somewhere else? Can we reference that source concretely and specifically? Are there any copyright disclaimers we should make?

linkerd/meshtls/rustls/src/client.rs
```diff
@@ -91,9 +99,19 @@ where
...
-        tokio_rustls::TlsConnector::from(self.config.clone())
-            .connect(self.server_name.clone(), io)
-            .map_ok(ClientIo)
+        let server_id = self.server_id.clone();
+        Box::pin(
+            tokio_rustls::TlsConnector::from(self.config.clone())
+                // XXX(eliza): it's a bummer that the server name has to be cloned here...
+                .connect(self.server_name.clone(), io)
+                .map(move |s| {
+                    let s = s?;
```
olix0r, line 108:
```suggestion
                .map_ok(move |s| {
```

linkerd/meshtls/rustls/src/client.rs
```diff
@@ -91,9 +99,19 @@ where
     }
 
     fn call(&mut self, io: I) -> Self::Future {
-        tokio_rustls::TlsConnector::from(self.config.clone())
-            .connect(self.server_name.clone(), io)
-            .map_ok(ClientIo)
+        let server_id = self.server_id.clone();
+        Box::pin(
+            tokio_rustls::TlsConnector::from(self.config.clone())
```
olix0r, line 104:
```suggestion
            // Connect to the server, sending the `server_name` SNI in the
            // client handshake. The provided config should use the
            // `AnySanVerifier` to ignore the server certificate's DNS SANs.
            // Instead, we extract the server's leaf certificate after the
            // handshake and verify that it matches the provided `server_id``.
            tokio_rustls::TlsConnector::from(self.config.clone())
```

linkerd/meshtls/rustls/src/creds/store.rs
olix0r: Is there analogous change on the boring side?

zaharidichev: No, not really. In boring we set the certificate that is going to be used, rather than relying on a resolver to pick it. There is the option to do that here as well. Not sure why the resolver has been picked as an approach. Maybe to add an extra check to ensure that the cert will have the SNI from the ClientHello.

linkerd/meshtls/rustls/src/creds/verify.rs
zaharidichev: Added a comment, not sure about disclaimers... This is licensed under Apache. Should we mention something?

linkerd/meshtls/rustls/src/creds/store.rs
olix0r: I expect it's so that the TLS library may vary the certificates it presents based on the SNI value.

linkerd/meshtls/rustls/src/creds/verify.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+}
+
+// Note that this logic is almost identical to the implementation of
+// `rustls::client::WebPkiVerifier`` that cab be found here:
+// https://github.com/rustls/rustls/blob/ccb79947a4811412ee7dcddcd0f51ea56bccf101/rustls/src/webpki/server_verifier.rs#L239
+// The only difference is that we omit the step that performs
+// DNS SAN validation. The reason for that stems from the fact that
+// SAN validation in rustls is limited to DNS SANs only while we
+// want to support alternative SAN types (e.g. URI).
```
olix0r, line 27: > This is licensed under Apache. Should we mention something?

My understanding of Apache (and MIT, and ISC, any of which rustls can be used under) is that we still have a copyright requirement for derived work. In fact, if we look at the most liberal (and shortest) license, we see that it is indeed the ONLY requirement:

> ISC License (ISC)
> Copyright (c) 2016, Joseph Birr-Pixton &lt;jpixton@gmail.com&gt;
>
> Permission to use, copy, modify, and/or distribute this software for
> any purpose with or without fee is hereby granted, **provided that the
> above copyright notice and this permission notice appear in all copies.**
> 
> THE SOFTWARE IS PROVIDED "AS IS" AND ...

And regardless of the legal requirements, It's best to bias towards crediting others work.

```suggestion
// This is derived from `rustls::client::WebPkiVerifier`.
//
//   Copyright (c) 2016, Joseph Birr-Pixton <jpixton@gmail.com>
// https://github.com/rustls/rustls/blob/ccb79947a4811412ee7dcddcd0f51ea56bccf101/rustls/src/webpki/server_verifier.rs#L239
//
// The only difference is that we omit the step that performs
// DNS SAN validation. The reason for that stems from the fact that
// SAN validation in rustls is limited to DNS SANs only while we
// want to support alternative SAN types (e.g. URI).
```

olix0r (approved):
Excellent! Thank you.

linkerd/meshtls/rustls/src/client.rs
```diff
@@ -72,15 +72,23 @@ impl Connect {
...
             config,
         }
     }
 }
 
+fn extract_cert(c: &rustls::ClientConnection) -> io::Result<&rustls::Certificate> {
+    match c.peer_certificates().and_then(|certs| certs.first()) {
+        Some(leaf_cert) => io::Result::Ok(leaf_cert),
+        None => Err(io::Error::new(io::ErrorKind::Other, "missing tls end cert")),
+    }
```
hawkw, line 86: tiny nit: isn't this match just [`Option::ok_or_else`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.ok_or_else)?

linkerd/meshtls/rustls/src/creds/store.rs
```diff
@@ -121,8 +126,9 @@ impl Store {
             NO_OCSP,
             now,
         )?;
-        debug!("Certified");
-        Ok(())
+
+        // verify the id as the cert verifier does not do that (on purpose)
```
hawkw, line 130: nit:
```suggestion
        // verify the identity, as the cert verifier does not do that (on purpose)
```
it might also be good if this comment explained _why_ we don't do that --- right now this just says it's "on purpose", but it doesn't say why...

hawkw (approved):
some last nits, but overall, this looks good to me!

olix0r:
I'm going to merge this through but feel free to address nits in followups!

PR #2509, opened 2023-11-08
Use reference-counted strings in dns::Name

olix0r:
Before b4ff7d23e, our identity::Name type used a reference-counted dns::Name. This change restores this reference-counting behavior, moving it into the lower-level `dns::Name` type.

This generally helps to minimize allocation overhead for proxied traffic.


hawkw (approved):
looks good to me!

PR #2510, opened 2023-11-10
Use heap indirection to manage type signatures

olix0r:
The proxy's release debug symbols are currently >6GB. It doesn't appear
to even be possible to build the proxy with debuginfo=2, as we overflow
u32::MAX and violate the restrictions of DWARF32.

To fix this, we more aggressively use dynamic dispatch via the new
`Stack::arc_new_box` helpers. This helps reduce our largest type
signatures from >700KB to ~50KB. We'll want to follow up with changes
that monitor the size of these signatures to prevent regressions.

This allows us to minimize many type assertions, etc.

There are no functional changes in this PR. Release builds will now
produce full debug information with file:line annotations.


linkerd/app/admin/src/stack.rs
```diff
@@ -97,13 +97,23 @@ impl Config {
...
+                    .proxy
+                    .http_endpoint
+                    .to_layer::<classify::Response, _, Permitted>(),
+            )
             .push(classify::NewClassify::layer_default())
             .push_map_target(|(permit, http)| Permitted { permit, http })
-            .push(inbound::policy::NewHttpPolicy::layer(metrics.http_authz.clone()))
+            .push(inbound::policy::NewHttpPolicy::layer(
+                metrics.http_authz.clone(),
+            ))
```
hawkw, line 111: this looks like just Rustfmt, right?

linkerd/app/inbound/src/http/router.rs
```diff
@@ -131,19 +116,9 @@ impl<C> Inbound<C> {
...
-                        .push(http_tracing::client(
-                            rt.span_sink.clone(),
-                            super::trace_labels(),
-                        ))
-                        .push(http::BoxResponse::layer())
-                        // This box is needed to reduce compile times on recent
-                        // (2021-10-17) nightlies, though this may be fixed by
-                        // https://github.com/rust-lang/rust/pull/89831. It should
-                        // be removed when possible.
-                        .push(svc::BoxService::layer())
```
hawkw, line 145: i'm assuming this `BoxService` was obviated by the boxing added in this PR?

linkerd/app/inbound/src/http/server.rs
```diff
@@ -72,39 +60,32 @@ impl<H> Inbound<H> {
...
+                //
+                // TODO(ver) This concurrency limit applies only to
+                // requests that do not yet have responses, but ignores
+                // streaming bodies. We should change this to an
+                // HTTP-specific imlementation that tracks request and
+                // response bodies.
+                .push_on_service(svc::ConcurrencyLimitLayer::new(max_in_flight_requests))
+                // Shed load by failing requests when the concurrency
+                // limit is reached.
+                .push_on_service(svc::LoadShed::layer())
```
hawkw, line 75: out of curiosity: does separating these into multiple `push_on_service` calls also have an impact on symbol size? or, was it just a drive-by refactor?

hawkw (approved):
this looks good to me!

i was a bit curious about the change from using one `push_on_service` calls with multiple layers to using a separate `push_on_service` for each layer --- does that impact symbol size, or was it just a refactor?

i suspect this won't have too big of a latency impact, since the boxing is mostly at the `NewService` level, and not at the level of individual request/response futures. i'm still curious about potential perf testing, though!

linkerd/app/admin/src/stack.rs
olix0r: yeah

linkerd/app/inbound/src/http/router.rs
olix0r: We're just doing this everywhere now, so this place isn't all that special. The boxing is now included in the arc_ helper

linkerd/app/inbound/src/http/server.rs
olix0r: The additional layer types don't help anything when we're looking at the type signatures. This keeps them a little simpler. Also the added benefit of outdenting things...

PR #2516, opened 2023-11-14
admin: Add optional /debug/pprof/profile endpoint

olix0r:
This change introduces a new, optional /debug/pprof/profile endpoint to
the proxy's admin server. This endpoint is feature-gated and, initially,
it will not be included in release builds.

When the feature is enabled, profiling requests are only permitted over
the loopback interface.

The development Dockerfile is updated to prevent stripping debug
symbols so the pprof data is hydrated with debug symbols.

### DIFF
--- a/linkerd/app/admin/src/pprof.rs
+++ b/linkerd/app/admin/src/pprof.rs
@@ -10,9 +10,7 @@ impl Pprof {
         fn query_param<'r, B>(req: &'r http::Request<B>, name: &'static str) -> Option<&'r str> {
             let params = req.uri().path_and_query()?.query()?.split('&');
             params
-                .filter_map(|p| {
-                    p.strip_prefix(name)?.strip_prefix('=')
-                })
+                .filter_map(|p| p.strip_prefix(name)?.strip_prefix('='))
                 .next()
         }
 
--- a/linkerd/app/admin/src/pprof.rs
+++ b/linkerd/app/admin/src/pprof.rs
@@ -1,0 +1,69 @@
+use linkerd_app_core::Result;
+
+#[derive(Copy, Clone, Debug)]
+pub(crate) struct Pprof;
+
+impl Pprof {
+    pub async fn profile<B>(self, req: http::Request<B>) -> Result<http::Response<hyper::Body>> {
+        use pprof::protos::Message;
+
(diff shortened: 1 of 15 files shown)


linkerd/app/admin/src/pprof.rs
```diff
@@ -0,0 +1,69 @@
...
+            std::io::Write::write_all(&mut gz, &{
+                let mut buf = Vec::new();
+                profile.encode(&mut buf)?;
+                buf
+            })?;
+            gz.finish()?
+        };
+
+        Ok(http::Response::builder()
+            .header("Content-Type", "application/octet-stream")
```
hawkw, line 65: very tiny nit: consider using [the constant from `http`](https://docs.rs/http/latest/http/header/constant.CONTENT_TYPE.html) for header values --- we can avoid the cost of validating the string literal, and IMHO it expresses intent? 
```suggestion
            .header(http::header::CONTENT_TYPE, "application/octet-stream")
```
we might also want to consider using a constant for the MIME type, for the same reason:
```rust
const MIME: HeaderValue = HeaderValue::from_static("application/octet-stream");
```

on the other hand, this probably doesn't actually matter...

linkerd/app/admin/src/server.rs
```diff
@@ -57,9 +58,22 @@ impl<M> Admin<M> {
...
         }
     }
 
+    #[cfg(feature = "pprof")]
+    pub fn with_profiling(mut self, enabled: bool) -> Self {
+        self.pprof = if enabled {
+            Some(crate::pprof::Pprof)
+        } else {
+            None
+        };
```
hawkw, line 73: nit, feel free to ignore this:
```suggestion
        self.pprof = enabled.then_some(crate::pprof::Pprof);
```

linkerd/app/admin/src/server.rs
```diff
@@ -254,6 +268,26 @@ where
                 }
             }
 
+            #[cfg(feature = "pprof")]
+            "/debug/pprof/profile.pb.gz" if self.pprof.is_some() => {
+                let pprof = self.pprof.expect("profiler must be configured");
```
hawkw, line 273: nit, take it or leave it: this `expect` could really be reworded to indicate that we _just_ entered a branch that's guarded by `self.pprof.is_some()` being `true`, so it should be unreachable...

hawkw (approved):
looks good to me! i commented on a few minor nits.

PR #2517, opened 2023-11-15
build: Use jemalloc on all platforms

olix0r:
We've been using jemalloc on amd64 for a very long time. linkerd2#11430 reports that the proxy's allocation behavior has less desirable characteristics on arm64. To fix this, we adopt use of jemallocator on all platforms.

Additionally, we change our dependencies to use tikv-jemallocator. From the docs:

> This project is the successor of jemallocator.
>
> The project is also published as jemallocator for historical
> reasons. The two crates are the same except names. For new projects,
> it's recommended to use tikv-xxx versions instead.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,6 +7,7 @@ on:
       - .github/workflows/release.yml
       - justfile
       - Cargo.toml # For release build settings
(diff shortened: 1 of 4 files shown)


olix0r:
There appear to be problems cross-compling jemalloc with clang. Discarding for now.

PR #2520, opened 2023-11-16
Add support for SPIFFE TLS Ids

zaharidichev:
This PR adds support for TLS ids that are in SPIFFE format (URI). Notable changes: 

- san verification logic has been extracted into a separate crate (`llinkerd-meshtls-util`)
- tests for ID extraction and validation logic are happening in this crate
- the `linkerd_identity::Id` type has been turned into an enum that can be either a DNS URI type
- the `Metadata` type now carries both ServerName and ServerId information


linkerd/meshtls/util/Cargo.toml
```diff
@@ -1,11 +1,19 @@
 [package]
-name = "linkerd-meshtls-test-util"
+name = "linkerd-meshtls-util"
```
hawkw, line 2: nit, take it or leave it: i might have called this `linkerd-meshtls-core`, following the naming scheme we use elsewhere in the project?

linkerd/proxy/api-resolve/src/metadata.rs
```diff
@@ -18,7 +19,7 @@ pub struct Metadata {
     tagged_transport_port: Option<u16>,
 
     /// How to verify TLS for the endpoint.
-    identity: Option<ServerId>,
+    identity: Option<(ServerId, ServerName)>,
```
olix0r, line 22: Can we perhaps change this to store a [`tls::client::ClientTls`](https://github.com/linkerd/linkerd2-proxy/blob/b786b5006e6f8b78a3c96cd4a2aaef8c799b80e7/linkerd/tls/src/client.rs#L22-L28)? This also opens the option to pass through ALPN settings (later).

linkerd/identity/src/lib.rs
```diff
@@ -4,47 +4,71 @@
...
+/// - a DNS-like identity string that matches an x509 DNS SAN
+/// - a SPIFFE ID in URI form, that matches an x509
 ///
 /// This isn't restricted to TLS or x509 uses. An authenticated Id could be
 /// provided by, e.g., a JWT.
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
-pub struct Id(pub linkerd_dns_name::Name);
+pub enum Id {
+    Dns(linkerd_dns_name::Name),
+    Uri(http::Uri),
```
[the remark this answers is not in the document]
olix0r: I would argue that this type shouldn't make any SPIFFE requirements. That seems like more of an x509-handling concern than a restriction we care about in linkerd's core types.

linkerd/meshtls/util/src/lib.rs
```diff
@@ -0,0 +1,287 @@
+use linkerd_error::Result;
+use linkerd_identity::Id;
+use std::io;
+use std::str::FromStr;
+
+fn extract_ids_from_cert(cert: &[u8]) -> Result<Vec<Id>> {
+    use x509_parser::prelude::*;
+    let (_, c) = X509Certificate::from_der(cert)?;
```
olix0r, line 8: Since this changes our DNSName validation, I think we should make this change discretely from the rest of this PR. I.e.

1. A PR that adds a new `linkerd-meshtls-verifier` crate using x509_parser, without URI handling;
2. This PR.

linkerd/proxy/api-resolve/src/metadata.rs
olix0r: This could also potentially make sense as a discrete change to help minimize the diff we need to consider for URI handling.

linkerd/identity/src/lib.rs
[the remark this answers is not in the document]
hawkw: Okay, that's kind of the question I was getting at --- I wasn't sure whether this variant represented a spiffe-specific thing or not.

zaharidichev:
@hawkw @olix0r Thanks for the thoughtful feedback, I am closing this PR now in order to split it into smaller, more focused changes, starting with https://github.com/linkerd/linkerd2-proxy/pull/2534

PR #2521, opened 2023-11-16
Update DEFAULT_OUTBOUND_TCP_QUEUE_CAPACITY to 10K

olix0r:
DEFAULT_OUTBOUND_TCP_QUEUE_CAPACITY was incorrectly set to 10 so that bursts of TCP connections could see EOF errors.

This change increases the value to match the HTTP queue capacity.

### DIFF
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -244,7 +244,7 @@ const DEFAULT_INBOUND_CONNECT_TIMEOUT: Duration = Duration::from_millis(300);
 const DEFAULT_INBOUND_CONNECT_BACKOFF: ExponentialBackoff =
     ExponentialBackoff::new_unchecked(Duration::from_millis(100), Duration::from_millis(500), 0.1);
 
-const DEFAULT_OUTBOUND_TCP_QUEUE_CAPACITY: usize = 10;
+const DEFAULT_OUTBOUND_TCP_QUEUE_CAPACITY: usize = 10_000;
 const DEFAULT_OUTBOUND_TCP_FAILFAST_TIMEOUT: Duration = Duration::from_secs(3);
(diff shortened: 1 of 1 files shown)


linkerd/app/src/env.rs
```diff
@@ -244,7 +244,7 @@ const DEFAULT_INBOUND_CONNECT_TIMEOUT: Duration = Duration::from_millis(300);
 const DEFAULT_INBOUND_CONNECT_BACKOFF: ExponentialBackoff =
     ExponentialBackoff::new_unchecked(Duration::from_millis(100), Duration::from_millis(500), 0.1);
 
-const DEFAULT_OUTBOUND_TCP_QUEUE_CAPACITY: usize = 10;
+const DEFAULT_OUTBOUND_TCP_QUEUE_CAPACITY: usize = 10_000;
 const DEFAULT_OUTBOUND_TCP_FAILFAST_TIMEOUT: Duration = Duration::from_secs(3);
 const DEFAULT_OUTBOUND_HTTP_QUEUE_CAPACITY: usize = 10_000;
```
hawkw, line 249: nit, take it or leave it: should we assign one of these defaults to the other default, so that we can't change one without also changing the other? 
```suggestion
const DEFAULT_OUTBOUND_TCP_QUEUE_CAPACITY: usize = 10_000;
const DEFAULT_OUTBOUND_TCP_FAILFAST_TIMEOUT: Duration = Duration::from_secs(3);
const DEFAULT_OUTBOUND_HTTP_QUEUE_CAPACITY: usize = DEFAULT_OUTBOUND_TCP_QUEUE_CAPACITY
```
that avoids this kind of regression in the future, and if we were to eventually decide that they should be different we could always change it back...

hawkw (approved):

PR #2523, opened 2023-11-16
stack: Add BoxCloneSyncService

olix0r:
tower::util::BoxCloneService doesn't implement Sync, which is necessary when we may want to share these services across threads (which is entirely the point of making a service cloneable, really).

This change replicates BoxCloneService as BoxCloneSyncService and updates many of Linkerd's stacks to return this new type.

This continues on the work begun in #2510 and #2514.

---

Building proxy with full debug symbols:

```
:; RUSTFLAGS='-C debuginfo=2' just profile=release build              
```

Before:

```
:; du -h target/x86_64-unknown-linux-gnu/release/linkerd2-proxy{,.dbg}
20M     target/x86_64-unknown-linux-gnu/release/linkerd2-proxy
1.5G    target/x86_64-unknown-linux-gnu/release/linkerd2-proxy.dbg
```

After:

```
:; du -h target/x86_64-unknown-linux-gnu/release/linkerd2-proxy{,.dbg}
20M     target/x86_64-unknown-linux-gnu/release/linkerd2-proxy
904M    target/x86_64-unknown-linux-gnu/release/linkerd2-proxy.dbg
```

**Result:** >500MB saved for debug symbols.

### DIFF
--- a/linkerd/app/core/src/control.rs
+++ b/linkerd/app/core/src/control.rs
@@ -71,12 +71,7 @@ impl Config {
         identity: identity::NewClient,
     ) -> svc::ArcNewService<
         (),
-        impl svc::Service<
-                http::Request<tonic::body::BoxBody>,
-                Response = http::Response<RspBody>,
-                Error = ControlError,
-                Future = impl Send,
-            > + Clone,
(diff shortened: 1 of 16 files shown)


hawkw (approved):
looks great!

PR #2529, opened 2023-11-17
test: Cleanup outbound_balancer_waits_for_ready_endpoint

olix0r:
The outbound_balancer_waits_for_ready_endpoint test doesn't reliably demonstrate the behavior it is trying to test: we setup two identical endpoints, so our tests can't be sure that the response is coming from the correct endpoint.

This change updates this test to provide backends that return varied responses so we can ensure the desired behavior.

This change also improves the test's logging. The discovery test infrastructure is updated to allow overriding the logical service address. This aids disambiguating addresses in logs.

### DIFF
--- a/linkerd/app/integration/src/tests/discovery.rs
+++ b/linkerd/app/integration/src/tests/discovery.rs
@@ -1,5 +1,6 @@
 use crate::*;
 use linkerd2_proxy_api as pb;
+use tokio::time::sleep;
 
 const HOST: &str = "disco.test.svc.cluster.local";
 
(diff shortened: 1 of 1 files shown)


linkerd/app/integration/src/tests/discovery.rs
```diff
@@ -422,66 +426,71 @@ mod http2 {
...
+        assert_eq!(
+            String::from_utf8(
+                hyper::body::to_bytes(res.into_body())
+                    .await
+                    .unwrap()
+                    .to_vec(),
+            )
+            .unwrap(),
+            "beta"
+        );
```
hawkw, line 493: tiny nit: there's [a `body_string` function in the test utils](https://github.com/linkerd/linkerd2-proxy/blob/8ce2df4faea100cc53db1a8f6f0822555fb9c685/linkerd/app/test/src/http_util.rs#L115-L127) that we could just use here: 
```suggestion
        assert_eq!(http_util::body_string(res.into_body()).await, "beta");
```

hawkw (approved):
looks good to me! i had one tiny suggestion

PR #2534, opened 2023-11-20
Consolidate tls id verification

zaharidichev:
This PR is a non-functional change that consolidates TLS Id verification so parsing of certificates, SAN extraction and verification all happen in `linkerd-meshtls-verifier` crate. In order to make this possible the [x509parser](https://github.com/rusticata/x509-parser) dependency has been introduced/


linkerd/meshtls/verifier/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+        let bar_dns_id = "bar.ns1.serviceaccount.identity.linkerd.cluster.local";
+        let spiffe_id = "spiffe://some-trust-comain/some-system/some-component";
+
+        let cert = generate_cert_with_names(vec![foo_dns_id, bar_dns_id, spiffe_id]);
+        let id = id::Id::from_str(foo_dns_id).expect("should parse DNS id");
+        assert!(verify_id(&cert, &id).is_ok());
+    }
+
+    #[test]
+    pub fn cert_with_dns_multiple_sans_none_matches_dns_id() {
```
olix0r, line 117: Is it worth adding a test with a cert that has only a spiffe_id? I.e. to document that we return an error (until the next PR).

olix0r (approved):

linkerd/meshtls/verifier/src/lib.rs
zaharidichev: If I understand you correctly, we do have such a test: https://github.com/linkerd/linkerd2-proxy/pull/2534/files#diff-f01d3fc9f7b0d20d3a9c55e350ffb752b18e784f08731b61a0afda65f9c09477R88

PR #2539, opened 2023-11-30
outbound: Move queues into the concrete stack

olix0r:
In preparation to change the load balancer's queuing behavior, this change moves the concrete stack's queue into the inner balancer and forward stacks.


zaharidichev (approved):
LGTM

hawkw (approved):

PR #2540, opened 2023-12-01
Add the PoolQueue middleware

olix0r:
Tower's [Buffer][buf] middleware is used to make a Service shareable across tasks. Within Linkerd, we've augmented the Buffer with failfast behavior, etc, as a Queue middleware. This queue plays a vital role in dispatching requests to load balancers, ensuring that load is shed when a balancer has no available endpoints.

The tower Balancer additionally holds a Stream of endpoint updates that are processed as the balancer is driven to readiness. Crucially, this means that updates to the balancer can only occur while requests are being processed. We cannot eagerly drop defunct endpoints, nor can we eagerly connect to new endpoints. We have observed situations where long-lived, mostly-idle balancers can buffer discovery updates indefinitely, bloating memory and even forcing backpressure to the control plane.

To correct this behavior, this change introduces the PoolQueue middleware. The PoolQueue is based on the tower's Buffer, but instead of composing over an inner Service, it composes over an inner Pool.

Pool is a new interface that provides an additional interface to update the pool's members and to drive all pending endpoints in the pool to be ready (decoupling the semantics of Service::poll_ready and Pool::poll_pool). Pool implementations will typically hold a ReadyCache of inner endpoints (as the tower balancer does). This change, however, does not include the concrete implementation of a Pool to be replace the balancer. A p2c pool will be introduced in a followup change.

This change has the added benefit of simplifying the endpoint discovery pipeline. We currently process Updates (including an array of endpoints) from the discovery API and convert that into a stream of discrete endpoint updates for the balancer, requiring redundant caching. The Pool interface processes Updates directly, so there is no need for the extra translation.

This basic implementation does not yet include metrics, though it includes copious diagnostic logging.

[buf]: https://github.com/tower-rs/tower/tree/bf4ea948346c59a5be03563425a7d9f04aadedf2/tower/src/buffer


linkerd/proxy/pool/src/worker.rs
```diff
@@ -0,0 +1,349 @@
...
+    T: Clone + Eq + std::fmt::Debug,
+    R: TryStream<Ok = Update<T>> + Unpin,
+    R::Error: Into<Error>,
+{
+    /// Attempts to update the pool with discovery updates.
+    ///
+    /// Additionally, this attempts to drive the pool to ready if it is
+    /// currently in failfast.
+    ///
+    /// If the discovery stream is closed, this never returns.
```
hawkw, line 154: maybe worth documenting that this only returns on errors, and that it's expected to only ever be called in a select with the request channel?

linkerd/proxy/pool/src/worker.rs
```diff
@@ -0,0 +1,349 @@
...
+            tokio::select! {
+                // Tests, especially, depend on discovery updates being
+                // processed before ready returning.
+                biased;
+
+                // If the pool updated, continue waiting for the pool to be
+                // ready.
+                res = self.discovery.discover() => {
+                    let update = res?;
+                    tracing::debug!(?update, "Discovered");
```
hawkw, line 194: take it or leave it: maybe worth swapping the `let` and the trace so that we also log on errors?
```suggestion
                update = self.discovery.discover() => {
                    tracing::debug!(?update, "Discovered");
                    let update = update?;
```

linkerd/proxy/pool/src/worker.rs
```diff
@@ -0,0 +1,349 @@
...
+                    tracing::info!(
+                        elapsed = (time::Instant::now() - since).as_secs_f64(),
+                        "Available; exited failfast"
+                    );
+                }
+                _ => unreachable!("must be in failfast"),
+            }
+            if let Err(e) = res {
+                return e.into();
+            }
```
hawkw, line 285: tiny style nit: isn't this just
```suggestion
            res?;
```

hawkw (approved):
This looks great! I had some fairly minor comments but didn't notice any significant concerns.

linkerd/proxy/pool/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+pub trait Pool<T, Req>: Service<Req> {
+    /// Updates the pool's endpoints.
+    fn update_pool(&mut self, update: Update<T>);
+
+    /// Polls pending endpoints to ready.
+    ///
+    /// This is complementary to [`Service::poll_ready`], which may also handle
+    /// driving endpoints to ready. Unlike [`Service::poll_ready`], which
+    /// returns ready when *at least one* inner endpoint is ready,
+    /// [`Pool::poll_pool`] returns ready when *all* inner endpoints are ready.
```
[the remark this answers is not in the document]
olix0r: Comment updated.

PR #2567, opened 2023-12-15
identity: add spire client

zaharidichev:
This PR adds a spire client.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -339,6 +339,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "const-oid"
+version = "0.9.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "28c122c3980598d243d63d9a704629a2d748d101f278052ff068be5a4423ab6f"
+
+[[package]]
 name = "cpp_demangle"
 version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


zaharidichev:
@olix0r This is a pretty rough implementation of a spire client. I could use some advice how to make
it more robust. I have tried following the StreamWatch pattern but without much success. Here
are two of the main problems I faced. 
- most of the implementation of StreamWatcher here depends on a service that can internally construct
a gRPC client based on a GRPC service (e.g. https://github.com/linkerd/linkerd2-proxy/blob/main/linkerd/app/inbound/src/policy/api.rs#L41). We cannot achieve this level of parametrization here if we resort to the SPIFFE crate, because
the module that contains the grpc clients is private in this crate (here: https://github.com/maxlambrecht/rust-spiffe/blob/main/spiffe/src/proto/workload.rs#L152)
- `spawn_watch` is implemented on a `StreamWatch` where we need a `Recover<tonic::Status>`. This does not work very well with the SpiffeWorkloadApi client, as exposes a `GrpcClientError` type, which might contain a `tonic::Status`. The point is that it seems that to spawn the watch itself, we need to transform the errors coming from the spiffe API to tonic::Status errors. Does that make sense? Would it make sense to just remove this hard requirement on the tonic watch and be able to use any error. If not, why? 
- seems that `spawn_watch` internally spawns a task. How should this relate to the task that is being built by the identity config, that should ultimatelly consume from this chan? 

Any pointers will be highly appreciated.

PR #2580, opened 2023-12-20
identity: add spire identity client

zaharidichev:
This PR adds SPIRE identity client that allows us to connect to the SPIRE workload API and stream new certificates through the `identity::Credentials` API. Notable changes include: 

- added `spire-proto` crate that includes the workload API protobuf definitions and codegen utilities
- added `spire-client` crate that contains all the facilities to connect to the SPIRE UDS socket and consume updates
- moved the Identity metrics into a seprate crate so it can be reused across both the control plane and spire identity clients

TODO: 
 - add more tests
 - validate and use trust roots received from SPIRE


linkerd/proxy/spire-client/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+
+    pub async fn run<C, N>(self, credentials: C, mut new_client: N)
+    where
+        C: Credentials,
+        N: Service<(), Response = watch::Receiver<SvidUpdate>>,
+        N::Error: Into<Error>,
+    {
+        match new_client.call(()).await {
+            Ok(rx) => consume_updates(&self.id, rx, credentials, &self.metrics).await,
+            Err(error) => error!("could not establish SVID stream: {}", error.into()),
```
mateiidavid, line 39: I wonder if it's better to just have the error message as a field:
```
error!(%error, "could not establish SVID stream");
```

linkerd/proxy/spire-client/src/api.rs
```diff
@@ -0,0 +1,151 @@
...
+                }
+                (leaf, intermediates)
+            }
+        };
+
+        let spiffe_id = Id::parse_uri(&proto.spiffe_id)?;
+
+        Ok(Svid {
+            spiffe_id,
+            leaf: leaf.clone(),
```
mateiidavid, line 61: Do we need the clone here? Can we just move the leaf into the Svid struct? Don't think we're going to be sharing the leaf anywhere else outside of this block.

linkerd/proxy/spire-client/src/client.rs
```diff
@@ -0,0 +1,63 @@
...
+use tokio::net::UnixStream;
+use tokio::sync::watch;
+use tonic::transport::{Endpoint, Uri};
+
+const UNIX_PREFIX: &str = "unix:";
+const TONIC_DEFAULT_URI: &str = "http://[::]:50051";
+
+#[derive(Clone, Debug)]
+pub struct Client {
+    socket: Arc<String>,
```
mateiidavid, line 14: This is a very nitpicky comment but this feels more like a socket addr rather than a socket (which I'd tend to think more of as a fd). 🤷🏻

linkerd/proxy/spire-client/src/api.rs
```diff
@@ -0,0 +1,151 @@
...
+use linkerd_proxy_http as http;
+use linkerd_tonic_watch::StreamWatch;
+use spiffe_proto::client::{
+    self as api, spiffe_workload_api_client::SpiffeWorkloadApiClient as Client,
+};
+use std::collections::HashMap;
+use tower::Service;
+use tracing::error;
+
+#[derive(Clone, Debug)]
```
mateiidavid, line 15: Should we remove this Debug impl since we're holding the private key bytes?

mateiidavid (commented):
I know this is still marked as draft but I had a look just to understand the changes. I won't be in for a while after today. Left some drive-by comments.

zaharidichev:
@olix0r 

I have: 
- moved the GRPC client into app/src as we discussed
- added more tests
- wired the Spire task into the identity without externalizing these capabilities through the config

linkerd/app/src/identity.rs
```diff
@@ -66,33 +83,52 @@ impl Config {
             &self.params.trust_anchors_pem,
         )?;
 
-        let certify = Certify::from(self.certify);
-        let metrics = certify.metrics();
-
-        let addr = self.control.addr.clone();
-
+        let metrics = IdentityMetrics::default();
```
olix0r, line 86: this seems like we're not actually wiring up metrics here.

linkerd/app/src/identity.rs
```diff
@@ -30,8 +41,14 @@ pub struct TlsParams {
     pub trust_anchors_pem: String,
 }
 
+#[derive(Clone)]
+pub enum Addr {
+    Spire(Arc<String>),
+    Linkerd(control::ControlAddr),
+}
```
olix0r, line 48: Is it possible to move this into the Provider type to avoid having redundant enums?

linkerd/proxy/identity-client-metrics/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "linkerd-proxy-identity-client-metrics"
```
olix0r, line 2: Here's an alternative approach to this (against main) https://github.com/linkerd/linkerd2-proxy/pull/2617 -- updating these metrics to reflect best practices (e.g. _total vs _count); it also means neither clients need to be explicitly aware of these metrics.

linkerd/app/src/identity.rs
zaharidichev: This is used in the `App`, [here](https://github.com/linkerd/linkerd2-proxy/pull/2580/files#diff-4689f302ab1ea17c575cc6e99ad1906a82f3fa3d387b945fe3c25567a6f5635aR351). It was returning a  `control::ControlAddr` before.

linkerd/proxy/spire-client/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+        Self { id }
+    }
+
+    pub async fn run<C, S>(self, credentials: C, mut client: S)
+    where
+        C: Credentials,
+        S: Service<(), Response = tonic::Response<watch::Receiver<SvidUpdate>>>,
+        S::Error: Into<Error> + Display,
+    {
+        match client.call(()).await {
```
olix0r, line 33: I'd like to talk a little bit more about the error handling in this task, but to satisfy the Service contract, we need to await readiness with something like `client.ready().await?.call(().await` (using the ServiceExt trait to get the ready() call).

linkerd/proxy/spire-client/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+        use x509_parser::prelude::*;
+
+        let (_, parsed_cert) = X509Certificate::from_der(&svid.leaf.0)?;
+        let exp: u64 = parsed_cert.validity().not_after.timestamp().try_into()?;
+        let exp = UNIX_EPOCH + Duration::from_secs(exp);
+
+        return credentials.set_certificate(svid.leaf, svid.intermediates, svid.private_key, exp);
+    }
+
+    Err("could not find an SVID".into())
```
olix0r, line 74: We tend not to use string errors, instead using typed errors via the thiserror crate.
```rust

#[derive(Debug, thiserror::Error)]
#[error("no matching SVID found")]
pub struct NoMatchingSVIDFound(());
```

Note that the unit field ensures the error can't be built outside of the module.

```suggestion
    Err(NoMatchingSVIDFound(()).into())
```

linkerd/proxy/spire-client/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+    loop {
+        let svid_update = updates.borrow_and_update().clone();
+        if let Err(error) = process_svid(&mut credentials, svid_update, id) {
+            tracing::error!("Error processing SVID update: {}", error);
+        }
+
+        if let Err(error) = updates.changed().await {
+            tracing::debug!("SVID watch closed; terminating {}", error);
+            return;
+        }
```
olix0r, line 56: nit:

```suggestion
        if let Err(error) = process_svid(&mut credentials, svid_update, id) {
            tracing::error!(%error, "Error processing SVID update");
        }

        if updates.changed().await.is_err() {
            tracing::debug!("SVID watch closed; terminating");
            return;
        }
```

We generally use field interpolation for errors (rather than direct string formatting).

Also, the error type from changed() is uninteresting -- it can only indicate that all receivers were dropped.

linkerd/proxy/spire-client/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+}
+
+// === impl Spire ===
+
+impl Spire {
+    pub fn new(id: Id) -> Self {
+        Self { id }
+    }
+
+    pub async fn run<C, S>(self, credentials: C, mut client: S)
```
olix0r, line 27: How should run handle errors?  Our options are basically (1) panic or (2) bubble them up. Given the choice, it's _gnerally_ preferable to bubble them up and, if necessary, panic closer to the root of the callstack.

linkerd/proxy/spire-client/src/api.rs
```diff
@@ -0,0 +1,208 @@
...
+};
+use std::collections::HashMap;
+use tower::Service;
+use tracing::error;
+
+#[derive(Clone)]
+pub struct Svid {
+    pub(super) spiffe_id: Id,
+    pub(super) leaf: DerX509,
+    pub(super) private_key: Vec<u8>,
```
olix0r, line 19: This is a bit of yakshave that we can do as a followup, but I think we really ought to have a linkerd_identity::PrivateKey type so we're not passing around sensitive byte arrays everywhere.

Why does the private key need to be crate-visible?

olix0r: Ah, I see this is for process_svid -- what if we moved process_svid into this module and made _that_ `pub(super)` -- this gives us a narrower module boundary where we can make all of the Svid stuff private

linkerd/app/src/identity.rs
```diff
@@ -55,47 +62,85 @@ impl Config {
...
+                client,
+                certify,
+                tls,
+            } => {
+                // TODO: move this validation into env.rs
+                let name = match (&tls.id, &tls.server_name) {
+                    (Id::Dns(id), sni) if id == sni => id.clone(),
+                    (_id, _sni) => {
+                        return Err(
+                            "Linkerd identity requires a TLS Id and server name to be the same"
```
olix0r, line 79: Ideally we'd use thiserror again here. Since is up at the app layer I don't care as much as I would in a lib.

linkerd/proxy/spire-client/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+    where
+        C: Credentials,
+        S: Service<(), Response = tonic::Response<watch::Receiver<SvidUpdate>>>,
+        S::Error: Into<Error> + Display + Debug,
+    {
+        let client = client.ready().await.expect("should be ready");
+        match client.call(()).await {
+            Ok(rsp) => consume_updates(&self.id, rsp.into_inner(), credentials).await,
+            Err(error) => error!(%error, "could not establish SVID stream"),
+        }
```
olix0r, line 36:
```suggestion
        let client = client.ready().await.expect("should be ready");
        let rsp = client.call(()).await.expect("spire client must gracefully handle errors");
        consume_updates(&self.id, rsp.into_inner(), credentials).await
```

Based on our earlier convo, I'm pretty sure that a failure here indicates something pathological (i.e., a bad request from the client).  If so, it seems better to panic than to log and never satisfy an identity.

olix0r (commented):
LGTM. One last (I think ;) nit

olix0r (approved):
:taco:

spiffe-proto/src/gen/spiffe.workloadapi.rs
```diff
@@ -0,0 +1,151 @@
+/// The X509SVIDRequest message conveys parameters for requesting an X.509-SVID.
+/// There are currently no request parameters.
+#[allow(clippy::derive_partial_eq_without_eq)]
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct X509svidRequest {}
+/// The X509SVIDResponse message carries X.509-SVIDs and related information,
```
olix0r, line 6:
```suggestion

/// The X509SVIDResponse message carries X.509-SVIDs and related information,
```

PR #2585, opened 2023-12-21
Read Forward endpoints from outbound policy API

mateiidavid:
Draft PR to explore what changes we'd need to make.

For the opaque stack, we change the param where we create an `opaq::Logical` target from a Sidecar target type. When a profile endpoint contains a default endpoint (i.e. just the forward address) we check whether the policy response has a Forward endpoint (we assume the protocol to be opaque).

Also started working on some integration tests. For the opaque forward test, we can see in the logs the correct stack is being built, however, the test would pass even with a forward (since we're asserting tcp read bytes which would work regardless; identity doesn't seem to do anything here).

```
[     0.097794s] DEBUG proxy{test=tests::client_policy::outbound_policy_opaque_forward:proxy}:evict{key=Opaq(Forward(Remote(ServerAddr(127.0.0.1:37291)), Metadata { labels: {}, protocol_hint: Opaque, tagged_transport_port: Some(33415), identity: Some(ClientTls { server_name: ServerName(Name("foo.ns1.serviceaccount.identity.linkerd.cluster.local")), server_id: ServerId(Dns(Name("foo.ns1.serviceaccount.identity.linkerd.cluster.local"))), alpn: None }), authority_override: None }))}: linkerd_idle_cache: Awaiting idleness

```

HTTP tests are still WIP.

### DIFF
--- a/linkerd/app/integration/src/controller.rs
+++ b/linkerd/app/integration/src/controller.rs
@@ -475,6 +475,63 @@ impl DestinationBuilder {
     }
 }
 
+impl From<DestinationBuilder> for pb::WeightedAddr {
+    fn from(
+        DestinationBuilder {
+            addr,
+            hint,
+            opaque_port,
+            addr_labels,
+            identity,
+            ..
+        }: DestinationBuilder,
+    ) -> Self {
+        let protocol_hint = match (hint, opaque_port) {
+            (Hint::Unknown, None) => None,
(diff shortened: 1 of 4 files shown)


mateiidavid:
We decided to do something a little different and wire-up the workload resp in the GetProfile path for now.

Closing for now.

PR #2617, opened 2024-01-05
identity: Update metrics to follow OpenMetrics best practices

olix0r:
In preparation to add SPIRE integration, this change extracts the identity certificate metrics from the control plane client, moving them into the linkerd-identity crate, as an implementation of Credentials trait (which is updated to include an expiry value). The app::identity module is updated to instrument the credentials store with metrics.

A `identity_cert_refresh_timestamp_seconds` metric is added to report the time when the identity certificate was last updated.

**Breaking changes**: `identity_cert_refresh_count` is renamed to `dentity_cert_refreshes_total` to follow OpenMetrics best practices, where counters must end with `_total`.

I've searched GitHub, and the only external references to this metric are in our documentation (which will have to be updated).

---

    # HELP identity_cert_expiration_timestamp_seconds Time when the this proxy's current mTLS identity certificate will expire (in seconds since the UNIX epoch).
    # TYPE identity_cert_expiration_timestamp_seconds gauge
    # UNIT identity_cert_expiration_timestamp_seconds seconds
    identity_cert_expiration_timestamp_seconds 1704500155.0
    # HELP identity_cert_refresh_timestamp_seconds Time when the this proxy's current mTLS identity certificate were last updated.
    # TYPE identity_cert_refresh_timestamp_seconds gauge
    # UNIT identity_cert_refresh_timestamp_seconds seconds
    identity_cert_refresh_timestamp_seconds 1704413735.7489427
    # HELP identity_cert_refreshes The total number of times this proxy's mTLS identity certificate has been refreshed by the Identity provider..
    # TYPE identity_cert_refreshes counter
    identity_cert_refreshes_total{result="ok"} 1
    identity_cert_refreshes_total{result="error"} 0


linkerd/identity/src/metrics.rs
```diff
@@ -0,0 +1,150 @@
...
+            prom::Unit::Seconds,
+            refresh_ts.clone(),
+        );
+
+        let refreshes = prom::Counter::default();
+        registry.register(
+            "refreshes",
+            "The total number of times this proxy's mTLS identity certificate has been refreshed by the Identity provider",
+            refreshes.clone(),
+        );
```
zaharidichev, line 44: As it is, this will get udpated only on successful refresh. I wonder whether we want to capture situations when the `Credentials::set_certificate` actually fails. 

For example we can have a label on this metrics that indicates the outcome of an attempted refresh and always set it whenever certs have been received from the provider

zaharidichev (approved):
LGTM, just a small comment. Making that a wrapper that implemented `Credentials` is quite neat

PR #2618, opened 2024-01-05
discovery: consume server_name and UriLikeIdentity from proto

zaharidichev:
This change is a follow-up from the work to split the concepts of
`ServerId` and `ServerName`. To do that we need to consume the
changes to the protobuf API introduced in: https://github.com/linkerd/linkerd2-proxy-api/pull/285


zaharidichev:
Note that this is failing because I am patching the api dependencies via a fork. Once we merge: https://github.com/linkerd/linkerd2-proxy-api/pull/285, and update the patch, this should be resolved.

olix0r (approved):
lgtm; but it's probably ~time to get the API release through so we can get this on a real release.

PR #2641, opened 2024-01-11
config:add spire client config

zaharidichev:
This change allows the proxy to be configured to use SPIRE
as it is credentials provider. There are three additional ENV
vars introduced: 

- `LINKERD2_PROXY_IDENTITY_SERVER_ID` - Configures the TLS Id of the proxy inbound server.
- `LINKERD2_PROXY_IDENTITY_SERVER_NAME` - Configures the server name of this proxy.
- `ENV_IDENTITY_SPIRE_SOCKET` - If this config is set, then the proxy will use Spire as the identity provider

For backward compatibility reasons if `ENV_IDENTITY_IDENTITY_LOCAL_NAME` is set both
tls id and server_name will be sourced from its value. If however it is not set, both `LINKERD2_PROXY_IDENTITY_SERVER_ID` and `LINKERD2_PROXY_IDENTITY_SERVER_NAME` are required. 

### DIFF
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -1,4 +1,4 @@
-use crate::{dns, gateway, identity, inbound, oc_collector, outbound, policy};
+use crate::{dns, gateway, identity, inbound, oc_collector, outbound, policy, spire};
 use linkerd_app_core::{
     addr,
     config::*,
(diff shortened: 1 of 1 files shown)


linkerd/app/src/env.rs
```diff
@@ -187,11 +187,29 @@ pub const ENV_INBOUND_IPS: &str = "LINKERD2_PROXY_INBOUND_IPS";
...
+// that clients include in the SNI extension of the ClientHello, whenever they try to
+// establish a TLS connection that shall be terminated by this proxy
+pub const ENV_IDENTITY_IDENTITY_SERVER_NAME: &str = "LINKERD2_PROXY_IDENTITY_SERVER_NAME";
+
+// If this config is set, then the proxy will be configured to use Spire as identity
+// provider
+pub const ENV_IDENTITY_SPIRE_SOCKET: &str = "LINKERD2_PROXY_IDENTITY_SPIRE_SOCKET";
+pub const IDENTITY_SPIRE_BASE: &str = "LINKERD2_PROXY_IDENTITY_SPIRE";
+const DEFAULT_SPIRE_BACKOFF: ExponentialBackoff =
+    ExponentialBackoff::new_unchecked(Duration::from_millis(1), Duration::from_millis(5), 0.1);
```
olix0r, line 211:
```suggestion
    ExponentialBackoff::new_unchecked(Duration::from_millis(100), Duration::from_secs(1), 0.1);
```

This seems a bit hot-loopy to me. If we need to backoff, let's give the OS some time to catch up and get Spire into a better state.

linkerd/app/src/env.rs
```diff
@@ -187,11 +187,29 @@ pub const ENV_INBOUND_IPS: &str = "LINKERD2_PROXY_INBOUND_IPS";
...
 pub const ENV_IDENTITY_DIR: &str = "LINKERD2_PROXY_IDENTITY_DIR";
 pub const ENV_IDENTITY_TRUST_ANCHORS: &str = "LINKERD2_PROXY_IDENTITY_TRUST_ANCHORS";
-pub const ENV_IDENTITY_IDENTITY_LOCAL_NAME: &str = "LINKERD2_PROXY_IDENTITY_LOCAL_NAME";
 pub const ENV_IDENTITY_TOKEN_FILE: &str = "LINKERD2_PROXY_IDENTITY_TOKEN_FILE";
 pub const ENV_IDENTITY_MIN_REFRESH: &str = "LINKERD2_PROXY_IDENTITY_MIN_REFRESH";
 pub const ENV_IDENTITY_MAX_REFRESH: &str = "LINKERD2_PROXY_IDENTITY_MAX_REFRESH";
 
+// This config is here for backwards compatibility. If set, both the tls id and the server
+// name will be set to the value specified in this config. The values needs to be a DNS
+// name
```
olix0r, line 196:
```suggestion
/// This config is here for backwards compatibility. If set, both the tls id and the server
/// name will be set to the value specified in this config. The values needs to be a DNS
/// name
```

Definitely a nit, but opt to use doc-comments -- especially when pub.

linkerd/app/src/env.rs
```diff
@@ -800,29 +818,44 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
+        let tls = tls?;
+
+        match strings.get(ENV_IDENTITY_SPIRE_SOCKET)? {
+            Some(socket) => identity::Config::Spire {
+                tls,
+                client: spire::Config {
+                    socket_addr: std::sync::Arc::new(socket),
+                    backoff: parse_backoff(strings, IDENTITY_SPIRE_BASE, DEFAULT_SPIRE_BACKOFF)?,
                 },
             },
```
olix0r, line 830: I think in this case we we want to validate that the local_id is a valid SPIFFE id.

zaharidichev: I have added the check for SPIFFE scheme for now. Is that enough for the time being? 

I also wonder, should we make the ID::Uri be spiffe specific and follow the requirements outlined in the SPIFFE RFC  (like we do with the dns name) or we should keep it as a URI type?

linkerd/app/src/env.rs
zaharidichev: Yes this was copy paste. I was thinking of changing it to something like 1 - 5 secs, but what you suggest also makes sense.

linkerd/app/src/env.rs
olix0r: On one hand, it would be useful to have a static assertion that the value has been validated; on the other hand, we want to use these types for values read from peers, and I don't think anything about Linkerd's _TLS_ system actually cares about SPIFFE.

So for now, I think it's sufficient to handle this where we habe both the context of (1) the configured local ID and (2) how we're going to provision credentials, i.e. "when SPIRE is enabled, validate that the configured ID is a SPIFFE ID."

olix0r (approved):

linkerd/app/src/env.rs
```diff
@@ -793,29 +812,58 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
...
-                    capacity: DEFAULT_CONTROL_QUEUE_CAPACITY,
-                    failfast_timeout,
-                },
+        let tls = tls?;
+
+        match strings.get(ENV_IDENTITY_SPIRE_SOCKET)? {
+            Some(socket) => match &tls.id {
+                // TODO: perform stricter SPIFFE ID validation following:
+                // https://github.com/spiffe/spiffe/blob/27b59b81ba8c56885ac5d4be73b35b9b3305fd7a/standards/SPIFFE-ID.md
+                identity::Id::Uri(uri) if uri.scheme() == SPIFFE_ID_URI_SCHEME => {
```
olix0r, line 821: i'm not sure about the semantics of SPIFFE but you may want to use [`eq_ignore_ascii_case`](https://doc.rust-lang.org/std/string/struct.String.html#method.eq_ignore_ascii_case)

PR #2794, opened 2024-03-14
Add IPv6 support when retrieving original destination

alpeb:
We first retrieve the socket option `SO_DOMAIN` to figure the socket's family, to then use the appropriate level and optname args when retrieving the original destination.

### DIFF
--- a/linkerd/proxy/transport/src/orig_dst.rs
+++ b/linkerd/proxy/transport/src/orig_dst.rs
@@ -117,15 +117,43 @@ mod linux {
     use tracing::warn;
 
     pub unsafe fn so_original_dst(fd: RawFd) -> io::Result<SocketAddr> {
+        let mut sockdomain: i32 = 0;
+        let mut sockdomain_len: libc::socklen_t = 32;
+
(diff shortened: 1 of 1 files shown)


linkerd/proxy/transport/src/orig_dst.rs
```diff
@@ -117,23 +117,51 @@ mod linux {
...
+            fd,
+            libc::SOL_SOCKET,
+            libc::SO_DOMAIN,
+            &mut sockdomain as *mut _ as *mut _,
+            &mut sockdomain_len as *mut _ as *mut _,
+        );
+        if ret != 0 {
+            let e = io::Error::last_os_error();
+            warn!("failed to read SO_DOMAIN: {:?}", e);
+            return Err(e);
```
olix0r, line 136: I know that you're just being consistent with what is already in this function, but if we emit warnings here, these messages are going to be double-logged (since we are also returning an error). This will end up blowing up the accept() call and it will be logged as a server error.

We should probably remove the existing warning on line 160, too

olix0r (commented):
This looks great! One small request.

PR #2803, opened 2024-03-16
Add environment variables to configure HTTP/2 keepalive timeout and interval configuration from environment variables

UsingCoding:
Add environment variables to configure HTTP/2 keepalive timeout and interval

Problem
There is no way to configure timeout and interval via environment variables.

Solution
Configure keepalive timeout and interval for inbound/outbound by `LINKERD2_PROXY_{INBOUND,OUTBOUND}_HTTP2_KEEPALIVE_{TIMEOUT,INTERVAL}`

Fixes https://github.com/linkerd/linkerd2/issues/12269


olix0r:
@UsingCoding thanks. I've been waiting to sync up with more of the team on how to proceed here. The environment-based approach is really not how we want to expose this sort of configuration, ultimately -- it's the wrong granularity of configuration. We should really be configuring this per server and per client (and not on the workload broadly). So we've treated the environment variables as a developer escape hatch for testing and never really wanted to show this configuration directly to users.

That said, adding non-environment configuration is quite a bit more work, so I'm sympathetic to simply exposing what we have today.

I'll talk to the team and come back with clearer guidance on how to proceed.

UsingCoding:
Close this PR due to [comment](https://github.com/linkerd/linkerd2/issues/12269#issuecomment-2199767041)

PR #2854, opened 2024-03-29
Bind outbound proxy to the IPv6 loopback

alpeb:
In order for the proxy to support IPv6 traffic we need to bind the outbound proxy to the IPv6 loopback (`::1`), in addition to the existing binding to the IPv4 loopback (`127.0.0.1`).

Note, this is not necessary for inbound; on the proxy side we'll set `LINKERD2_PROXY_INBOUND_LISTEN_ADDR` (and the other `*_LISTEN_ADDR`) vars to the IPv6 wildcard address (`::`) which also works for IPv4.

This change adds the `LINKERD2_PROXY_OUTBOUND_LISTEN_ADDRS` var, which is supposed to be set as `127.0.0.1:4140,[::1]:4140` (maximum two entries). If not set, we default to the value in `LINKERD2_PROXY_OUTBOUND_LISTEN_ADDR`.

~The `ServerConfig` got expanded with an optional `addr_additional` field, which should be populated with `[::1]:4140`. When set, we bind using `BindWithOrigDst`'s new method `bind_additional` which leverages `ServerConfig`'s `InsertParam` implementation for putting its `addr_additional` into `addr` for the regular binding logic to continue.~

Update:
The field `ServerConfig.addr` is now of the new type `DualListenAddr` which holds a primary `SocketAddr` and an optional one.
The outbound listener is now bound via the `DualBind` implementation of `Bind` which also binds to the second `SocketAddr` if set, merging the two streams using Tokio's `merge` stream extension method.


linkerd/proxy/transport/src/listen.rs
```diff
@@ -26,6 +26,12 @@ pub trait Bind<T> {
     type Incoming: Stream<Item = Result<(Self::Addrs, Self::Io)>> + Send + Sync + 'static;
 
     fn bind(self, params: &T) -> Result<Bound<Self::Incoming>>;
+    fn bind_additional(self, _params: &T) -> Result<Bound<Self::Incoming>>
+    where
+        Self: Sized,
+    {
+        unimplemented!();
+    }
```
olix0r, line 34: We should not leave this in a way that can lead to runtime panics. If we really do need to modify the interface, we should force all implementations to handle it. But I don't think it's necessary to modify this interface.

I recall that before--I can't find where, maybe it was verbal--I had suggested that we implement all of this _behind_ the single bind() call.  I.e. we should need minimal application changes to support this and this change can be much smaller and less invasive.

linkerd/proxy/transport/src/orig_dst.rs
```diff
@@ -100,6 +101,16 @@ where
...
+
+    fn bind_additional(self, t: &T) -> Result<Bound<Self::Incoming>> {
+        match t.param() {
+            Some(addr) => {
+                let t2 = t.insert_param(addr, ());
+                self.bind(&t2)
+            }
+            None => self.bind(t),
+        }
+    }
```
olix0r, line 113: Instead, we can provide a new type, e.g. `DualBind<B>(B, B)` that binds over a `T: Param<DualListenAddr>`, and then when we construct the app, we give it a DualBind for the outbound listener instead!

The trick to this will be joining the two resultant Incoming streams, but we should have utilities available to us to accomplish that.

alpeb:
Ok I've rebased and pushed again using the suggested approach. I introduced a new `DualBindWithOrigDst` implementation of the `Bind` trait, used by the outbound listener which forwards two bind calls to `BindWithOrigDst`. Given we need to return two addresses, I was forced to expand the `Bound` type to be a triple, which introduced some extra baggage there where we don't require binding twice. Perhaps we can modify that type a bit to make this less ugly.
I also had to introduced an `ExtractParam<ServerConfig, SocketAddr>` implementation to make that `DualBindWithOrigDst`->`BindWithOrigDst` chain work; it didn't feel great so I'm all ears on suggestions.

linkerd/proxy/transport/src/dual_bind.rs
```diff
@@ -0,0 +1,63 @@
+use crate::{
```
olix0r, line 1: nit: Would you mind moving this into `linkerd/proxy/transport/src/listen/dual.rs` so that it's coped under the listen module?

linkerd/proxy/transport/src/listen.rs
```diff
@@ -28,7 +28,7 @@ pub trait Bind<T> {
     fn bind(self, params: &T) -> Result<Bound<Self::Incoming>>;
 }
 
-pub type Bound<I> = (Local<ServerAddr>, I);
+pub type Bound<I> = (Local<ServerAddr>, Option<Local<ServerAddr>>, I);
```
olix0r, line 31: Do we need too include this when it's not used?

My expectation is that we can use this type alias when dealing with basic bind type, but that dual module should define its own
```rust
pub type Bound<I> = (Local<ServerAddr>, Option<Local<ServerAddr>>, I);
```

This where there's no ambiguity in the type system, no unwraps, etc.

alpeb:
In my last push I added the `BoundAddrs` associated type to `Bind` to  avoid having to deal with the triple everywhere. Also moved dual_bind as a submodule of listen, and other cleanups.

linkerd/proxy/transport/src/listen/dual_bind.rs
```diff
@@ -0,0 +1,79 @@
+use crate::{addrs::DualListenAddr, listen::Bind, Keepalive, ListenAddr};
+use futures::Stream;
+use linkerd_error::Result;
+use linkerd_stack::Param;
+use std::{net::SocketAddr, pin::Pin};
+use tokio::net::TcpStream;
+use tokio_stream::StreamExt;
+
+#[derive(Copy, Clone, Debug, Default)]
+pub struct DualBindWithOrigDst<B> {
```
olix0r, line 10: Nothing in this module cares about OrigDstAddr so we rename this

```suggestion
pub struct DualBind<B> {
```

linkerd/proxy/transport/src/listen/dual_bind.rs
```diff
@@ -0,0 +1,79 @@
...
+                    addr: addr2,
+                    parent: target.clone(),
+                })?;
+
+                let incoming_merged = incoming1.merge(incoming2);
+                let incoming_merged = incoming_merged.map(|res| {
+                    let (inner, tcp) = res?;
+                    Ok((inner, tcp))
+                });
+                Ok(((addr1, Some(addr2)), Box::pin(incoming_merged)))
```
olix0r, line 60: Aren't these maps redundant?
```suggestion
        let DualListenAddr(addr1, addr2) = target.param();
        let (addr1, incoming1) = self.inner.clone().bind(&Listen {
            addr: addr1,
            parent: target.clone(),
        })?;
        match addr2 {
            Some(addr2) => {
                let (addr2, incoming2) = self.inner.bind(&Listen {
                    addr: addr2,
                    parent: target.clone(),
                })?;
                Ok(((addr1, Some(addr2)), Box::pin(incoming1.merge(incoming2))))
```

olix0r (commented):
This looks great to me! Thanks. Two small suggestions.

linkerd/proxy/transport/src/listen/dual_bind.rs
```diff
@@ -0,0 +1,79 @@
...
+pub struct DualBindWithOrigDst<B> {
+    inner: B,
+}
+
+pub struct Listen<T> {
+    addr: SocketAddr,
+    parent: T,
+}
+
+// === impl DualBindTcp ===
```
olix0r, line 19:
```suggestion
// === impl DualBind ===
```

linkerd/proxy/transport/src/listen/dual_bind.rs
alpeb: Ah indeed; now I understand better what's going on.

linkerd/app/src/env.rs
```diff
@@ -463,10 +467,16 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError>
         )?
         .unwrap_or(ingress_mode);
 
-        let addr = ListenAddr(
-            outbound_listener_addr?
-                .unwrap_or_else(|| parse_socket_addr(DEFAULT_OUTBOUND_LISTEN_ADDR).unwrap()),
-        );
+        let addr = match outbound_listener_addrs {
+            Ok(Some(addrs)) if addrs.len() == 1 => DualListenAddr(addrs[0], None),
+            Ok(Some(addrs)) if addrs.len() == 2 => DualListenAddr(addrs[0], Some(addrs[1])),
```
olix0r, line 472:
```suggestion
        let addr = match outbound_listener_addrs? {
            Some(addrs) if addrs.len() == 1 => DualListenAddr(addrs[0], None),
            Some(addrs) if addrs.len() == 2 => DualListenAddr(addrs[0], Some(addrs[1])),
```

I think we ought to fail if parsing errors (otherwise what's the point of introducing the new error variant?)

olix0r (approved):

linkerd/app/src/env.rs
alpeb: Good catch :+1:

PR #2866, opened 2024-04-08
Pull identity initialization out of app building

zaharidichev:
The purpose of this change is to allow a SPIRE identity to obtained without prior knowledge of the TLS id or the roots. In addition to that, the change enables us to construct and initialize the identity before we construct the proxy App. In order to achieve that: 

- the identity config has been split into Spire and Linkerd structs
- metrics related components can be passed to the Config::build method
- the `Credentials` trait has been modified to accept roots in addition to the rest of the credentials
- when the app is constrcuted, it is assumed that identity has already been initialized.

TODO: 
- pull pkix pem parsing crate functionality into our codebase
- conage/fix proxy initialization tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -46,6 +46,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "android-tzdata"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
+
+[[package]]
+name = "android_system_properties"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "anyhow"
 version = "1.0.81"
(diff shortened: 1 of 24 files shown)


linkerd/app/Cargo.toml
```diff
@@ -34,3 +34,4 @@ tokio-stream = { version = "0.1", features = ["time", "sync"] }
 tonic = { version = "0.10", default-features = false, features = ["prost"] }
 tower = "0.4"
 tracing = "0.1"
+pkix = "0.2.3"
```
olix0r, line 37: We need to be conservative about pulling in random dependencies of various maintenance statuses... *especially* for our TLS related dependencies.

Can we use [rustls-pemfile](https://docs.rs/rustls-pemfile/latest/rustls_pemfile/), since we already depend on rustls and have a relationship with the maintiners?

Cargo.lock
```diff
@@ -307,6 +340,20 @@ version = "1.0.0"
...
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8a0d04d43504c61aa6c7531f1871dd0d418d91130162063b789da00fd7057a5e"
+dependencies = [
+ "android-tzdata",
+ "iana-time-zone",
+ "js-sys",
+ "num-traits",
+ "wasm-bindgen",
+ "windows-targets 0.52.0",
+]
```
olix0r, line 355: I would like to avoid a dependency on Chrono, unless we are actually using some advanced timezone-aware time calculation feature.

PR #2874, opened 2024-04-09
identity: change Credentials trait to allow dynamically setting roots

zaharidichev:
The SPIFFE API allows provisioning trust roots. To enable that functionality, this PR changes the `Credentials` interface and adds a `Roots` parameter. The roots can be encoded in either PEM or DER.

For now, this is a nonfunctional change as the roots that are provided via the changes API are not used.

### DIFF
--- a/linkerd/app/src/identity.rs
+++ b/linkerd/app/src/identity.rs
@@ -1,6 +1,6 @@
 use crate::spire;
 
-pub use linkerd_app_core::identity::{client, Id};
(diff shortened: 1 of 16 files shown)


olix0r:
> For now, this is a nonfunctional change as the roots that are provided via the changes API are not used.

In order to be sure that this API change is sufficient and correct, and to keep the identity system in an internally consistent state (i.e. where the TLS impls actually implement the interfaces they expose), I think we should plumb the rest of this functionality through, with tests, before merging.

PR #3068, opened 2024-07-15
feat(inbound): Log traffic with the 'audit' policy

alpeb:
Audit mode is triggered by the policy controller, which will create an authorization named "audit" allowing traffic for the given target. When the proxy processes an authorization with such name it will log it at INFO.

Also, add "audit" to the possible values for
`LINKERD2_PROXY_INBOUND_DEFAULT_POLICY`, whose effect is the same as "all-unauthenticated".

Please check linkerd/website#1805 for how this is supposed to work from the user's perspective.

### DIFF
--- a/linkerd/app/inbound/src/policy/http.rs
+++ b/linkerd/app/inbound/src/policy/http.rs
@@ -203,7 +203,7 @@ impl<T, N> HttpPolicyService<T, N> {
             .find(|a| super::is_authorized(a, self.connection.client, &self.connection.tls))
         {
             Some(authz) => {
-                if authz.meta.name() == "audit" {
+                if authz.meta.is_audit() {
                     tracing::info!(
                         server.group = %labels.server.0.group(),
                         server.kind = %labels.server.0.kind(),
--- a/linkerd/app/inbound/src/policy/http.rs
+++ b/linkerd/app/inbound/src/policy/http.rs
@@ -202,7 +202,25 @@ impl<T, N> HttpPolicyService<T, N> {
             .iter()
             .find(|a| super::is_authorized(a, self.connection.client, &self.connection.tls))
         {
-            Some(authz) => authz,
+            Some(authz) => {
+                if authz.meta.name() == "audit" {
+                    tracing::info!(
+                        server.group = %labels.server.0.group(),
+                        server.kind = %labels.server.0.kind(),
+                        server.name = %labels.server.0.name(),
+                        route.group = %labels.route.group(),
+                        route.kind = %labels.route.kind(),
+                        route.name = %labels.route.name(),
+                        client.tls = ?self.connection.tls,
+                        client.ip = %self.connection.client.ip(),
+                        authz.group = %authz.meta.group(),
+                        authz.kind = %authz.meta.kind(),
+                        authz.name = %authz.meta.name(),
+                        "Request allowed",
+                    );
+                }
+                authz
+            }
             None => {
                 tracing::info!(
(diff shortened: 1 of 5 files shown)


linkerd/app/src/env.rs
```diff
@@ -1008,7 +1008,7 @@ fn parse_default_policy(
         "all-authenticated" => {
             Ok(inbound::policy::defaults::all_authenticated(detect_timeout).into())
         }
-        "all-unauthenticated" => {
+        "all-unauthenticated" | "audit" => {
```
mateiidavid, line 1011: question: if we treat this as an `all-unauthenticated` by default, when we create the default policy will it still have the correct metadata?

What would happen if for example we have an `audit` default policy and we get an inbound connection? Will we still log with the audit labels?

mateiidavid (commented):
Left a small question on parsing default policies. I was also wondering if it's worth adding some tests.

Integration tests have a policy controller mock that implements the inbound and outbound interfaces (see https://github.com/linkerd/linkerd2-proxy/blob/main/linkerd/app/integration/src/policy.rs). There are also some unit tests in:
* Inbound TCP stack (https://github.com/linkerd/linkerd2-proxy/blob/main/linkerd/app/inbound/src/accept.rs#L154-L170)
* Inbound policy module (https://github.com/linkerd/linkerd2-proxy/blob/main/linkerd/app/inbound/src/policy.rs#L211).

I wonder if it makes sense to add a test or two to ensure this all behaves as expected.

alpeb:
I've added a new new branch in the matcher parsing LINKERD2_PROXY_INBOUND_DEFAULT_POLICY that creates a similiar ServerPolicy as the one used for all_unauthenticated, but named "audit", this name being the only thing that causes the special log entry to be produced. I     tested this manually using a proxy with a hard-coded `policy::Config::Fixed` policy with an audit default ServerPolicy to avoid overriding LINKERD2_PROXY_INBOUND_DEFAULT_POLICY's policy with something coming from the policy controller.

As for unit/integration tests, I couldn't quite find a place for this that didn't duplicate what we already have for all_unauthenticated. We could add something with some tracing subscriber that would catch the log entry but not sure if it's worth it.     Open to more suggestions.

linkerd/app/inbound/src/policy/http.rs
```diff
@@ -202,7 +202,25 @@ impl<T, N> HttpPolicyService<T, N> {
             .iter()
             .find(|a| super::is_authorized(a, self.connection.client, &self.connection.tls))
         {
-            Some(authz) => authz,
+            Some(authz) => {
+                if authz.meta.name() == "audit" {
```
olix0r, line 206: in all of these cases don't we want to check that we're dealing with a kind=default and not just an arbitrary authorization resource named audit?

alpeb: Good point, should be fixed now.

linkerd/proxy/server-policy/src/meta.rs
```diff
@@ -39,6 +39,10 @@ impl Meta {
             Self::Resource { group, .. } => group,
         }
     }
+
+    pub fn is_audit(&self) -> bool {
+        self.kind() == "default" && self.name() == "audit"
```
olix0r, line 44: We can use pattern matching to avoid string comparison here:

```suggestion
        matches!(self, Self::Default { name } if name == "audit")
```

olix0r (approved):

PR #3160, opened 2024-08-29
feat(outbound): Route TLS connections by SNI

zaharidichev:
This is a draft PR that wires together some of the recently introduced changes to the proxy in order to deliver TLS routing functionality. The changes are: 

This change adds a TLS routing stack that has the following properties: 
 - it always expects that `SNI` is present
 - uses tls routing information provided by the discovery API to perform routing based on SNI
 - does not concern itself with terminating TLS by simply proxies the encrypted stream


linkerd/io/src/sensor.rs
```diff
@@ -82,3 +82,10 @@ impl<T: PeerAddr, S> PeerAddr for SensorIo<T, S> {
         self.io.peer_addr()
     }
 }
+
+#[async_trait::async_trait]
+impl<I: Peek + Send + Sync, S: Sensor + Sync> Peek for SensorIo<I, S> {
+    async fn peek(&self, buf: &mut [u8]) -> Result<usize> {
+        self.io.peek(buf).await
+    }
+}
```
zaharidichev, line 91: Had to add that, because The TLS stack requires the IO type to be peekable. If there is an easy way to avoid that, it would be great.

linkerd/tls/route/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "linkerd-tls-route"
```
olix0r, line 2: We could probably introduce this crate in a distinct change. It should include [module-level tests like the http-route crate](https://github.com/linkerd/linkerd2-proxy/blob/main/linkerd/http/route/src/http/tests.rs).

linkerd/app/outbound/src/tls/detect.rs
```diff
@@ -0,0 +1,92 @@
...
+pub struct NoSniFoundError;
+
+#[derive(Clone, Debug, Default)]
+pub struct NewDetectSniServer<N> {
+    inner: N,
+    timeout: time::Duration,
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct DetectSniServer<T, N> {
```
olix0r, line 33: How does this differ from the middleware in linkerd/tls/src/server.rs? Can they be unified? Is there any reason that this middleware can't be defined in the linkerd-tls crate? That should be our default--it's easier to to test/change modules when they are decoupled from the app. That is, if the app isn't compiling, it can still be possible to test and make changes to this stuff.

olix0r (commented):
From a merge strategy perspective, I think we'll probably want to split out PRs for from the leaves up to the root. That is, we can easily add the tls/route crate in an independent change with its own tests. Then we can add client-policy changes that incorporate those changes, and then we can add the outbound tls stack with tests, and then finally wire it into the outbound protocol router.

We are going to have to wire this into the ingress stack as well, presumably? Or what do we expect to happen in the case where an ingress makes an upstream HTTPS call?

linkerd/app/outbound/src/tls/detect.rs
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+}
+
+impl<T, I, N, S> svc::Service<I> for DetectSniServer<T, N>
+where
+    T: Clone + Send + Sync + 'static,
+    I: io::AsyncRead + io::Peek + io::AsyncWrite + Send + Sync + Unpin + 'static,
+    N: svc::NewService<(ServerName, T), Service = S> + Clone + Send + 'static,
+    S: svc::Service<DetectIo<I>> + Send,
```
olix0r, line 63: This DetectIo is probably wrong here -- we never want to proceed if we don't detect an SNI, right? So we can be sure we're handling a PrefixedIo

zaharidichev: Not sure this is correct. The "eitherness" of `DetectIo` does not distinguish between whether an SNI was detected or not. A `DetectIo` is a PrefixedIO in the cases when we could not parse the SNI via peeking. In these cases, we read from the socket and then construct a PrefixedIo. In the case of peeking a DetectIO is the original IO... 

So I think a `DetectIo` makes sense here(i.e detection has been performed on this IO regardless of the outcome.)

Does that sound right ?

linkerd/app/outbound/src/protocol.rs
```diff
@@ -29,6 +30,7 @@ impl<N> Outbound<N> {
     pub fn push_protocol<T, I, NSvc>(
         self,
         http: svc::ArcNewCloneHttp<Http<T>>,
+        tls: svc::ArcNewCloneTcp<T, io::EitherIo<I, io::PrefixedIo<I>>>,
```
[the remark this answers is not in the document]
olix0r: Ah, this is already enforced by the protocol stack -- this either is an artifact of doing optional protocol detection (and not related to the TLS change).

linkerd/tls/src/server/required_sni.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+#[derive(Clone, Debug)]
+pub struct DetectRequiredSni<T, N> {
+    target: T,
+    inner: N,
+    timeout: time::Duration,
+}
+
+impl<N> NewDetectRequiredSni<N> {
```
cratelyn, line 50:
```suggestion
}

// === impl NewDetectRequiredSni ===

impl<N> NewDetectRequiredSni<N> {
```

linkerd/tls/src/server/required_sni.rs
```diff
@@ -0,0 +1,116 @@
...
+{
+    type Service = DetectRequiredSni<T, N>;
+
+    fn new_service(&self, target: T) -> Self::Service {
+        DetectRequiredSni::new(self.timeout, target, self.inner.clone())
+    }
+}
+
+// === impl DetectSni ===
+
```
cratelyn, line 72:
```suggestion
}

// === impl DetectRequiredSni ===

```

linkerd/tls/src/server/required_sni.rs
```diff
@@ -0,0 +1,116 @@
...
+    pin::Pin,
+    task::{Context, Poll},
+};
+use thiserror::Error;
+use tokio::time;
+use tracing::debug;
+
+#[derive(Clone, Debug, Error)]
+#[error("SNI detection timed out")]
+pub struct SniDetectionTimeoutError;
```
cratelyn, line 19: optional idea:

if this was a newtype around https://docs.rs/tokio/latest/tokio/time/error/struct.Elapsed.html, we could be extra-defensive about confirming that we've encountered a timeout. (_contd. below..._ :point_down:)

linkerd/tls/src/server/required_sni.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    fn call(&mut self, io: I) -> Self::Future {
+        let target = self.target.clone();
+        let new_accept = self.inner.clone();
+
+        // Detect the SNI from a ClientHello (or timeout).
+        let detect = time::timeout(self.timeout, detect_sni(io));
+        Box::pin(async move {
+            let (res, io) = detect.await.map_err(|_| SniDetectionTimeoutError)??;
```
cratelyn, line 108:
```suggestion
            let (res, io) = detect.await.map_err(SniDetectionTimeoutError)??;
```

(_contd. from above.._) and then this could wrap the timeout error, making sure we don't ever discard some other error we might care about.

linkerd/app/outbound/src/tls.rs
```diff
@@ -0,0 +1,134 @@
...
+
+#[derive(Clone, Debug, Default)]
+pub struct TlsMetrics {
+    balance: concrete::BalancerMetrics,
+}
+
+// === impl Outbound ===
+
+impl<C> Outbound<C> {
+    /// Builds a stack that proxies tls connections.
```
cratelyn, line 72:
```suggestion
    /// Builds a stack that proxies TLS connections.
```

linkerd/app/outbound/src/tls/concrete.rs
```diff
@@ -0,0 +1,366 @@
...
+
+/// Parameter configuring dispatcher behavior.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum Dispatch {
+    Balance(NameAddr, balance::EwmaConfig),
+    Forward(Remote<ServerAddr>, Metadata),
+    Fail { message: Arc<str> },
+}
+
+/// A backend dispatcher explicitly fails all requests.
```
cratelyn, line 32:
```suggestion
/// A backend dispatcher that explicitly fails all requests.
```

cratelyn: that suggestion fixes a typo, i believe, but is this comment affixed to the wrong structure? it seems like this is _an_ error returned by the `fail` service created below in this statement:

```rust
          let fail = svc::ArcNewService::new(|message: Arc<str>| {
                svc::mk(move |_| futures::future::ready(Err(DispatcherFailed(message.clone()))))
            });
```

cratelyn (commented):
i took another read through this work, it's looking very good! i have some nitpicks i noticed, but these are all small details. nice work @zaharidichev!

olix0r (approved):
LGTM 🎉

PR #3176, opened 2024-09-05
Add wall clock gauge to proxy metrics

sfleen:
Currently, it's not possible to accurately calculate the time remaining before the proxy certificate expires, since the proxy didn't expose its view of what the time is. This exposes the wall clock time (rounded to the nearest second) as a gauge so downstream consumers can calculate the offsets for cert expiry, as well as any other offsets for timestamps we provide.

Example metrics output:
```
$ curl http://whoami.default:4191/metrics | grep process_clock_time
...
# HELP process_clock_time_seconds Current system time for this proxy.
# TYPE process_clock_time_seconds gauge
# UNIT process_clock_time_seconds seconds
process_clock_time_seconds 1725565062.0
```

Fixes [#12413](https://github.com/linkerd/linkerd2/issues/12413)

### DIFF
--- a/linkerd/identity/src/metrics.rs
+++ b/linkerd/identity/src/metrics.rs
@@ -37,6 +37,14 @@ impl CertMetrics {
             refresh_ts.clone(),
         );
 
+        let clock_time_ts = prom::Gauge::<f64, ClockMetric>::default();
+        registry.register_with_unit(
+            "clock_time",
+            "Current system time for this proxy",
+            prom::Unit::Seconds,
+            clock_time_ts,
+        );
+
         #[derive(Clone, Debug, PartialEq, Eq, Hash, prom::encoding::EncodeLabelSet)]
         struct RefreshLabelSet {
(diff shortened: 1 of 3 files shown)


linkerd/identity/src/metrics.rs
```diff
@@ -37,6 +37,14 @@ impl CertMetrics {
             refresh_ts.clone(),
         );
 
+        let clock_time_ts = prom::Gauge::<f64, ClockMetric>::default();
```
olix0r, line 40: I think we probably aught to move this gauge into the process metrics so that it's more clearly reusable across all metrics exported by the process -- there's nothing really identity-specific about it :)

sfleen: I agree that makes more sense for a general gauge like this, updated.

olix0r (approved):
LGTM 🎉

cratelyn (approved):
:taco: :taco: :taco:

PR #3186, opened 2024-09-06
Add env var for controlling log timestamp format.

sfleen:
The current log format for the proxy shows the number of seconds since process start. This is fine for most cases, but can make logs processing and aggregation tricky, especially when using the JSON format.

This adds the `LINKERD2_PROXY_LOG_TIMESTAMP_FORMAT` env var that switches the timestamp to use the wall clock time when set to `WALL_CLOCK`. Since getting the system time is more expensive than time since process start, this flag is explicitly opt-in for users where this cost is acceptable for their use case.

Example output with the env var set:
```
2024-09-06T12:17:22.474384Z DEBUG ThreadId(01) linkerd_app::env: Only allowing connections targeting `LINKERD2_PROXY_INBOUND_IPS` allowed={10.42.0.18}
2024-09-06T12:17:22.488845Z  INFO ThreadId(01) linkerd2_proxy::rt: Using single-threaded proxy runtime
2024-09-06T12:17:22.489018Z DEBUG ThreadId(01) linkerd_app: Building app
2024-09-06T12:17:22.489062Z DEBUG ThreadId(01) linkerd_app: Building DNS client
2024-09-06T12:17:22.489397Z DEBUG ThreadId(01) linkerd_app: Building Identity client
2024-09-06T12:17:22.489728Z DEBUG ThreadId(01) linkerd_app: Building Tap server config=Disabled
2024-09-06T12:17:22.489790Z DEBUG ThreadId(01) linkerd_app: Building Destination client
```

Fixes [#12505](https://github.com/linkerd/linkerd2/issues/12505)


olix0r:
I wonder if we can do any better than adding yet another configuration option.

The general problem, as I understand it is:

* When we format logs in textual format, the existing format is basically fine, since users can use `kubectl logs --timestamp` to include the timestamp in output.
* But this is unworkable when using JSON formatted logs, since these timestamps are not encoded in the JSON.
* Furthermore, the uptime-oriented values are formatted weirdly in JSON logs (i.e. with `[ ...]`) -- this does not lend itself to reading these values from the JSON structure.
* So it seems like an ideal solution could be special cased to the JSON format, in which case we could include _both_ the uptime and timestamp values.

For context: the uptime-oriented values (inspired by the Linux kernel) are useful, so that if we get a random log snippet, we know if the proxy has just started (where errors are more likely due to initialization ordering) or has been running for ~weeks (which may imply another set of issues).

Is there a way for us to simply do the right thing for JSON-formatted logs without additional configuration? IMO, that would be to include both an uptime value and a timestamp value? And would it make sense for these values to be encoded numerically, rather than as strings? This could improve processing with tools like `jq`.

olix0r (approved):
I'm approving this, but I notice your commit message includes no body. As this is merged, lets ensure that the commit message includes a full description like that in the PR.

PR #3191, opened 2024-09-09
chore(gitignore): ignore npm files

cratelyn:
the devcontainers cli is installed via npm. this means that some npm related artifacts (a package.json, package.lock, and a node_modules directory) may be present if starting, building, or attaching to containers outside of vs code.

this commit adds these npm artifacts to the gitignore.

see <https://github.com/devcontainers/cli>.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+node_modules/**
+package-lock.json
+package.json
 target
 **/target
 **/corpus


olix0r:
> the devcontainers cli is installed via npm.

That shouldn't have anything to do with this repo though? It sounds like you ran some install commands while in this repo, which littered your working directory. But you could just have easily run those commands from your home or another directory. I'd rather not include random other things here.

cratelyn:
I'm going to close this. I would like to revisit this at some point, but for now it's a small patch I can carry as part of my personal workflow.

PR #3192, opened 2024-09-10
feat(tls-route): add TLS route matching crate

zaharidichev:
This PR includes a nonfunctional change that adds a TLS route-matching library.
The library works similarly to the http-route one but instead of relying on HTTP concepts
such as paths, headers, etc is specific to TLS ones (i.e. SNI).

At the moment the library is capable of matching routes based on SNIs.

### DIFF
--- a/linkerd/tls/route/src/lib.rs
+++ b/linkerd/tls/route/src/lib.rs
@@ -64,11 +64,14 @@ pub fn find<P>(routes: &[Route<P>], session_info: SessionInfo) -> Option<(RouteM
         let sni = if rt.snis.is_empty() {
             None
         } else {
-            trace!(%session_info.sni, "matching sni");
-            rt.snis
+            let session_sni = &session_info.sni;
+            trace!(%session_sni, "matching sni");
+            let sni_match = rt
+                .snis
                 .iter()
-                .filter_map(|a| a.summarize_match(&session_info.sni))
-                .max()
+                .filter_map(|a| a.summarize_match(session_sni))
+                .max()?;
+            Some(sni_match)
         };
 
         trace!(rules = %rt.rules.len());
--- a/linkerd/tls/route/src/lib.rs
+++ b/linkerd/tls/route/src/lib.rs
@@ -8,11 +8,6 @@ use linkerd_tls::ServerName;
 use r#match::SessionMatch;
 use tracing::trace;
 
-#[derive(Clone, Debug, Hash, PartialEq, Eq)]
-pub struct SessionInfo {
-    pub sni: ServerName,
-}
-
 pub mod r#match;
 pub mod sni;
 #[cfg(test)]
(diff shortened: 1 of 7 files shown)


linkerd/tls/route/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+#![forbid(unsafe_code)]
+
+use linkerd_tls::ServerName;
+use r#match::SessionMatch;
+use tracing::trace;
+
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub struct SessionInfo {
+    pub sni: ServerName,
+}
```
olix0r, line 14: This should be moved down below the mods with the other structs. I would probably move it to directly about the `find` impl, so that it's defined nearest where it is used.

It would also be good to include a comment a brief comment describing that this is intended to capture the routing metadata of a proxied/unterminated server-side TLS connection.

linkerd/tls/route/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub rules: Vec<Rule<P>>,
+}
+
+/// Policies for a given set of route matches.
+#[derive(Clone, Debug, Default, Hash, PartialEq, Eq)]
+pub struct Rule<P> {
+    /// A list of session matchers, *any* of which may apply.
+    ///
+    /// The "best" match is used when comparing rules.
+    pub matches: Vec<r#match::MatchSession>,
```
olix0r, line 42: So it seems that we have both SNI matching at the Route level and also the Rule level. If we're going to include MatchSni at the Route level, then we probably want MatchSession to be empty for now. Otherwise, we probably want to omit the sni matching at the Route level.

I don't have enough information to suggest one or the other. I don't really see any harm in it being defined at the Route level (since this matches the [gateway spec](https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1alpha2/tlsroute_types.go#L87)). Given that the spec does not include any rule based matches currently, I'm not sure of a good reason to include SNI matches in rules.

linkerd/tls/route/src/sni.rs
```diff
@@ -0,0 +1,94 @@
+use linkerd_dns as dns;
+use linkerd_tls::ServerName;
+
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub enum MatchSni {
```
olix0r, line 5: It's probably worth documenting SNI matching semantics with some of the Gateway API's [verbiage](https://github.com/kubernetes-sigs/gateway-api/blob/ffe4b631bf3f0333ef6c6fc902cd4755a369bda0/apis/v1alpha2/tlsroute_types.go#L51-L57).

olix0r (commented):
This is looking pretty good to me!

It would probably be good to have some tests cases for sessions that do not match any routes. For example, it would be good to explore the corner cases like a suffix match like `*.test.example.com` and a session for `test.example.com`. What other subtleties should we consider?

olix0r (approved):

olix0r:
When merging, please ensure that the commit message includes an accurate summary of the change like that in the PR description.

PR #3199, opened 2024-09-11
tls: add `DetectSni` middleware

zaharidichev:
This change adds a new `DetectSni` middleware to be used in the outbound stack
in order to extract the SNI extension from the `ClientHello` of a TLS session and apply
routing decisions based on it.

In contrast to `DetectTls` this new middleware is concerned with just extracting the SNI
as opposed to terminating the TLS session.


linkerd/tls/src/server.rs
```diff
@@ -107,27 +84,27 @@ impl<L, P, N> NewDetectTls<L, P, N> {
     }
 }
 
-impl<T, L, P, N> NewService<T> for NewDetectTls<L, P, N>
+impl<T, L, P, N> NewService<(T, Option<ServerName>)> for NewDetectTls<L, P, N>
```
olix0r, line 87: Using a tuple-target is a bit of an anti-pattern: it's fragile and doesn't really permit any intermediate layers to work. It may be appropriate here if we can bundle it up so that it's opaque to the caller, but it's preferable to use a Param or ExtractParam to get a ServerName from the target.

linkerd/tls/src/server.rs
```diff
@@ -107,27 +84,27 @@ impl<L, P, N> NewDetectTls<L, P, N> {
...
-            timeout,
+            sni,
             params: self.params.clone(),
             inner: self.inner.clone(),
         }
     }
 }
 
-impl<I, T, L, LIo, P, N, NSvc> Service<I> for DetectTls<T, L, P, N>
+impl<I, T, L, LIo, P, N, NSvc> Service<DetectIo<I>> for DetectTls<T, L, P, N>
```
olix0r, line 107: This no longer _detects_ really, right? Should this be renamed to reflect what it does?

It's also pretty awkward to me that we pass an optional SNI and the local identity here and then at process-time we check whether we should actually do TLS termination.

olix0r (commented):
I'm a little skeptical on modifying the existing behavior--trying to decouple detection and termination, because the termination ends up being highly contingent on the other layer setting it up properly--we end up exposing a bunch of the implementation details outside the module, which is complex.

It seems like we probably need to get something exposed that only reads the SNI. The TLS server module can probably reuse some internals, but it would probably be easier to focus on adding something that does exactly what we need it do and only on that so we don't incur additional type complexity.

zaharidichev:
> I'm a little skeptical on modifying the existing behavior--trying to decouple detection and termination

@olix0r  Makes sense. I have reverted these commits and pushed a simpler version that just adds a `DetectSni` middleware that makes use of the bits and pieces in the `server` module.  Also, I have incorporated your feedback about tupled targets and tried to rely on `InsertParam`. 

Not sure whether we want to just move all sni and client hellos internals into their own module that is used by both `server` and `detect_sni` or keep it as it is to minimize the change. 

One option is to have the following mods: 

- extract_sni - contains the detect_sni function + client hello details + tests
- server - contains TLS termination logic, etc
- detect_sni - contains the middleware to detect the sni.

WDYT?

linkerd/tls/src/detect_sni.rs
```diff
@@ -0,0 +1,107 @@
...
+pub struct NewDetectSni<P, N> {
+    params: P,
+    inner: N,
+}
+
+#[derive(Clone, Debug)]
+pub struct DetectSni<T, P, N> {
+    target: T,
+    inner: N,
+    timeout: Timeout,
```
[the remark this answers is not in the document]
zaharidichev: Good point. So yes, essentially we would error with a [SniDetectionTimeoutError](https://github.com/linkerd/linkerd2-proxy/pull/3199/files#diff-63c849c315c60ad6103e0c25bc2dcb4a71a83dc5df207bcc43838a8d92431346R19) if the timeout elapses. Does that make sense here?

olix0r (approved):

PR #3214, opened 2024-09-19
Vendor OpenTelemetry protocol

sfleen:
In preparation for supporting the OpenTelemetry protocol for the proxy's traces, this vendors the relevant protobufs as well as some convenience helpers here.

This works around the fact that the existing `open-telemetry` crate does not support the version of tonic that we currently use (0.10), and this vendoring can be removed once we update.

[#10111](https://github.com/linkerd/linkerd2/issues/10111)


cratelyn (approved):
this looks good to me.

nice!

PR #3221, opened 2024-09-23
Integrate OpenTelemetry into the proxy

sfleen:
OpenCensus is a deprecated protocol and is slated to be removed from upstream collectors soon.

This wires up the proxy to optionally use OpenTelmetry as the format for exported traces. Currently, this defaults to the existing OpenCensus exporter, and we can switch the default later.

[#10111](linkerd/linkerd2#10111)


Cargo.lock
```diff
@@ -3454,20 +3455,21 @@ checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
 
 [[package]]
 name = "tokio"
-version = "1.40.0"
+version = "1.38.1"
```
olix0r, line 3458: We shouldn't downgrade deps in this PR.

sfleen: Not 100% sure how this got downgraded, should be resolved now.

linkerd/app/core/src/http_tracing.rs
```diff
@@ -1,139 +1,88 @@
...
-/// it to an OpenCensus span and then sends it on the provided mpsc::Sender.
-#[derive(Clone)]
-pub struct SpanConverter {
-    kind: Kind,
-    sink: mpsc::Sender<oc::Span>,
-    labels: Labels,
+#[derive(Clone, Debug)]
+pub enum SpanSink {
+    Oc(OpenCensusSink),
+    Otel(OpenTelemetrySink),
```
olix0r, line 18: I'm a little surprised that we need to expose the type of the tracing backend to the span producer. I was expecting that we'd have a common span type from the publisher's point of view and that the converter would handle converting the span to the proper wire format. For example, is it possible to use the SpanData type in both cases?

tools/src/bin/gen-protos.rs
```diff
@@ -1,18 +1,17 @@
 fn main() {
     let opencensus_dir = {
         let manifest_dir = std::path::PathBuf::from(std::env!("CARGO_MANIFEST_DIR"));
-        manifest_dir.parent().unwrap().join("opencensus-proto")
+        manifest_dir.parent().unwrap().join("opentelemetry-proto")
```
olix0r, line 4: Does this mean that we no longer support generating opencensus protos? Don't we need to support both?

sfleen: I assumed we wouldn't have to update these again since OpenCensus is deprecated and the protocol effectively frozen, but I'm realizing if we update tonic in the meantime we'd need to regenerate these. I've re-added generating the OpenCensus protos.

linkerd/app/core/src/http_tracing.rs
sfleen: I agree it's a leaky abstraction that ideally shouldn't exist. I spent some time digging through this, turns out it's a lot easier to re-use our existing internal span representation and send that to the exporter task, letting the exporter be the one that converts the span to the protocol-specific format. This has the added benefit of offloading the (fairly small) overhead of converting the trace format onto the exporter rather than the tracing middleware.

linkerd/app/src/trace_collector.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub kind: CollectorProtocol,
+    pub span_sink: SpanSink,
+    pub task: Task,
+}
+
+impl EnabledCollector {
+    pub fn info_span(&self) -> tracing::Span {
+        match self.kind {
+            CollectorProtocol::OpenCensus => info_span!("opencensus"),
+            CollectorProtocol::OpenTelemetry => info_span!("opentelemetry"),
```
olix0r, line 46: This seems fishy to me... these span macros are going to capture the context in this file and not at the callsite. Do we really need to differentiate the span based on the collector type? Would it be sufficient to just use something more general like `info_span!("tracing")`?

linkerd/app/src/trace_collector/oc_collector.rs
```diff
@@ -0,0 +1,63 @@
...
+use linkerd_opencensus::{self as opencensus, metrics, proto};
+use std::{collections::HashMap, time::SystemTime};
+use tokio::sync::mpsc;
+use tokio_stream::wrappers::ReceiverStream;
+use tonic::body::BoxBody;
+use tonic::client::GrpcService;
+use tracing::Instrument;
+
+const SPAN_BUFFER_CAPACITY: usize = 100;
+const SERVICE_NAME: &str = "linkerd-proxy";
```
olix0r, line 15: If these configs re common across both impls, should they pulled up into the parent trace_collector module directly andreferenced from each?

olix0r (commented):
General shape of this is looking good to me. Some smaller nitty gritty feedback...

linkerd/app/core/src/http_tracing.rs
```diff
@@ -1,139 +1,73 @@
...
-    actual_size: usize,
+impl FromStr for CollectorProtocol {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "opencensus" => Ok(CollectorProtocol::OpenCensus),
+            "opentelemetry" => Ok(CollectorProtocol::OpenTelemetry),
+            _ => Err(()),
+        }
```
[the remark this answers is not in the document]
sfleen: Done.

linkerd/app/core/src/http_tracing.rs
```diff
@@ -1,139 +1,73 @@
...
-use linkerd_opencensus::proto::trace::v1 as oc;
 use linkerd_stack::layer;
-use linkerd_trace_context::{self as trace_context, TraceContext};
-use std::{collections::HashMap, sync::Arc};
-use thiserror::Error;
+use linkerd_trace_context as trace_context;
+use linkerd_trace_context::export::{ExportSpan, SpanKind, SpanLabels};
+use linkerd_trace_context::{Span, TraceContext};
+use std::str::FromStr;
+use std::sync::Arc;
```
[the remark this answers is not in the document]
sfleen: Good to know. I believe there's a way to set our rustfmt preferences in a `rustfmt.toml` file, but that's best left for a follow-up I think.

linkerd/app/src/trace_collector.rs
sfleen: In this case I think a single `tracing` span sufficient, fixed.

linkerd/app/src/trace_collector/oc_collector.rs
sfleen: Done.

linkerd/app/core/src/http_tracing.rs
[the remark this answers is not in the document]
cratelyn: https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#imports_granularity

chiming in to say that while `imports_granularity` exists as a setting available to be configured in `rustfmt.toml`, it isn't currently stable and the tracking issue appears to have been stalled for some time now.

i've used `+nightly` for this before, but it is a little unwieldy to introduce a second toolchain in a larger project like this.

alas! 😪

linkerd/app/src/trace_collector.rs
```diff
@@ -0,0 +1,114 @@
...
+            Config::Enabled(config) => match config.kind {
+                CollectorProtocol::OpenCensus => info_span!("opencensus"),
+                CollectorProtocol::OpenTelemetry => info_span!("opentelemetry"),
+            },
+        }
+    }
+
+    pub fn metrics_prefix(&self) -> &'static str {
+        match self {
+            Config::Disabled => "disabled_tracing",
```
olix0r, line 67: Does this mean we're going to start exporting a bunch of metrics with the disabled_tracing_... in the default case? That's pretty yucky.

It would be better, imo, to stub out the metrics registry when tracing is disabled so that we don't add any metrics to the global registry.

linkerd/app/src/trace_collector.rs
```diff
@@ -0,0 +1,114 @@
...
+impl Config {
+    pub fn info_span(&self) -> tracing::Span {
+        match self {
+            Config::Disabled => info_span!("disabled_tracing"),
+            Config::Enabled(config) => match config.kind {
+                CollectorProtocol::OpenCensus => info_span!("opencensus"),
+                CollectorProtocol::OpenTelemetry => info_span!("opentelemetry"),
+            },
+        }
+    }
```
olix0r, line 63: Similarly to my previous comment, this span is going to end up appearing misconfiguerd. I think it would be preferable to omit this special casing and leave the caller as `info_span!("tracing")`

linkerd/app/src/lib.rs
```diff
@@ -188,24 +188,33 @@ impl Config {
...
 
-        debug!(config = ?oc_collector, "Building client");
-        let oc_collector = {
-            let control_metrics =
-                ControlMetrics::register(registry.sub_registry_with_prefix("opencensus"));
+        debug!(config = ?trace_collector, "Building client");
+        let trace_collector = {
+            let control_metrics = ControlMetrics::register(
+                registry.sub_registry_with_prefix(trace_collector.metrics_prefix()),
+            );
```
olix0r, line 195: I think this could be something like
```suggestion
            let control_metrics = ControlMetrics::register(
                trace_collector.metrics_prefix().map(|p| registry.sub_registry_with_prefix(p)).unwrap_or_default(),
            );
```

sfleen: Done.

linkerd/app/src/trace_collector.rs
sfleen: Done.

linkerd/app/src/trace_collector.rs
sfleen: Done.

olix0r (approved):
lgtm!

linkerd/app/src/lib.rs
```diff
@@ -188,24 +188,36 @@ impl Config {
...
-        let oc_collector = {
+        debug!(config = ?trace_collector, "Building client");
+        let trace_collector = {
             let control_metrics =
-                ControlMetrics::register(registry.sub_registry_with_prefix("opencensus"));
+                ControlMetrics::register(if let Some(prefix) = trace_collector.metrics_prefix() {
+                    registry.sub_registry_with_prefix(prefix)
+                } else {
+                    &mut registry
+                });
```
olix0r, line 198: I was thinking something along the lines of

```suggestion
                if let Some(prefix) = trace_collector.metrics_prefix() {
                    ControlMetrics::register(registry.sub_registry_with_prefix(prefix))
                } else {
                    ControlMetrics::register(&mut Registry::default())
                };
```

sfleen: Gotcha, wasn't 100% sure what the correct semantics were here.

linkerd/app/src/lib.rs
```diff
@@ -188,24 +188,36 @@ impl Config {
...
-        debug!(config = ?oc_collector, "Building client");
-        let oc_collector = {
+        debug!(config = ?trace_collector, "Building client");
+        let trace_collector = {
             let control_metrics =
-                ControlMetrics::register(registry.sub_registry_with_prefix("opencensus"));
+                ControlMetrics::register(if let Some(prefix) = trace_collector.metrics_prefix() {
+                    registry.sub_registry_with_prefix(prefix)
+                } else {
+                    &mut registry
```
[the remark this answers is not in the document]
sfleen: Done

linkerd/app/src/env.rs
```diff
@@ -146,6 +147,7 @@ const ENV_OUTBOUND_DISABLE_INFORMATIONAL_HEADERS: &str =
     "LINKERD2_PROXY_OUTBOUND_DISABLE_INFORMATIONAL_HEADERS";
 
 pub const ENV_TRACE_ATTRIBUTES_PATH: &str = "LINKERD2_PROXY_TRACE_ATTRIBUTES_PATH";
+pub const ENV_TRACE_PROTOCOL: &str = "LINKERD2_PROXY_TRACE_PROTOCOL";
```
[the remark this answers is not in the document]
sfleen: Done

PR #3227, opened 2024-09-24
chore(app): Disallow warnings in policy router's backend metrics

cratelyn:
a stray `#![allow(warnings)]` directive seems to have slipped in from an initial prototyping phase.

this commit removes this directive from
`/linkerd/app/outbound/src/http/logical/policy/route/backend/metrics.rs`, and addresses the newly unearthed warnings, related to unused imports.


linkerd/app/outbound/src/http/logical/policy/route/backend/metrics.rs
```diff
@@ -1,11 +1,5 @@
-#![allow(warnings)]
```
cratelyn, line 1: :point_up: the point of this commit.

we remove this.

olix0r (approved):
🙇

PR #3228, opened 2024-09-24
docs(http/prom): Document request count middleware

cratelyn:
this is a light documentation pass outlining how the request counting middleware fits together.

this was selfishly done for my own edification, but felt worth proposing as a commit for other future readers. :memo:

### DIFF
--- a/linkerd/http/prom/src/count_reqs.rs
+++ b/linkerd/http/prom/src/count_reqs.rs
@@ -1,3 +1,5 @@
+//! A Tower middleware for counting requests processed by a service.
+
 use linkerd_stack as svc;
 use prometheus_client::{
     encoding::EncodeLabelSet,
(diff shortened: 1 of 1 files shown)


linkerd/http/prom/src/count_reqs.rs
```diff
@@ -6,18 +8,29 @@ use prometheus_client::{
...
 pub struct NewCountRequests<X, N> {
+    /// The inner [`NewService<T>`][svc::NewService].
     inner: N,
+    /// The [`ExtractParam<P, T>`][svc::ExtractParam] strategy for obtaining our parameters.
     extract: X,
 }
 
+/// Counts requests for an inner `S`-typed [`Service`][svc::Service].
+///
+/// This will increment its counter when [`call()`][svc::Service::call]ed, before calling the inner
```
cratelyn, line 32: meta: do we lint for broken doc links?

https://doc.rust-lang.org/rustdoc/lints.html#broken_intra_doc_links

we could disallow broken links via `broken_intra_doc_links`, though i know it can sometimes be a chore to fix that if it hasn't been enforced previously.

olix0r: AFAIK we don't lint docs in application code (we probably do when we publish to crates.io).

olix0r (approved):

PR #3230, opened 2024-09-24
docs(http/prom): Add metadata to Cargo.toml manifest

cratelyn:
this commit introduces a description and adds the authors.


olix0r (approved):
fwiw we tend to omit authors now that it's no longer required. The authors keys tend to be a remnant of yesteryear.

cratelyn:
Ah! That's helpful to know. I'll merge this for the sake of the `description` field, but appreciate the lore regarding the authors field. 🙂

PR #3234, opened 2024-09-25
refactor(prom): Remove `MkStreamLabel`'s associated types

cratelyn:
`MkStreamLabel` is, in short, a generic
`(&Request) -> Option<StreamLabel>` function. we use it to inspect a request, and potentially provide the caller with an object that can provide relevant labels.

the `StreamLabel` interface includes associated types for the labels used for metrics related to request/response duration, and counting status codes.

we do not however, actually need to separately define these associated types in the `MkStreamLabel` contract. instead, we can return a generic `StreamLabel` of some sort, and leave the responsibility of the (admittedly baroque) uniform function call access to our type aliases like `RecordResponseDuration` and `RecordRequestDuration`.

this is a small initial step towards simplifying code that must interact with the `MkStreamLabel` interface.

### DIFF
--- a/linkerd/app/outbound/src/http/logical/policy/route.rs
+++ b/linkerd/app/outbound/src/http/logical/policy/route.rs
@@ -173,8 +173,6 @@ impl<T> filters::Apply for Http<T> {
 }
 
 impl<T> metrics::MkStreamLabel for Http<T> {
-    type StatusLabels = metrics::labels::HttpRouteRsp;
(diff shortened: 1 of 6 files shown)


cratelyn:
there are warnings related to unused type aliases, i'll address those before marking this as ready for review.

cratelyn:
closing in favor of #3242 :soon: :construction:

PR #3242, opened 2024-09-27
refactor(http/prom): Simplify `record_response` middleware

cratelyn:
this branch makes a variety of small, non-breaking changes to the `linkerd-http-prom` crate, and the outbound proxy code related to route- and backend-level metrics.

this branch does _not_ make particularly significant breaking changes to the `StreamLabel` or `MkStreamLabel` traits. this branch is focused on minimizing the type complexity of these traits, and consequently in dependent code.

these changes are provided in distinct, atomic commits. readers are encourages to review this branch commit-by-commit. the primary changes that are worth highlighting, mentioned in comments below are:

* `StreamLabel` is now object-safe, making use of `dyn StreamLabel` possible. we now use `request::Parts` and `response::Parts` parameters in our trait methods. this means that we no longer need to be generic across request/response bodies, but retain the ability to inspect e.g. status codes.

* the `StreamLabel` interface includes associated types for the labels
used for metrics related to request/response duration, and counting
status codes. we do not however, actually need to separately define these associated
types in the `MkStreamLabel` contract. these types are removed.

* `ResponseMetrics` and `RequestMetrics` are generic over `L: StreamLabel`. these were previously generic over `DurationLabels` and `StatusLabels`, even though the same parent `L` was used in practice.

* documentation to various interfaces is added.

### DIFF
--- a/linkerd/http/prom/src/record_response.rs
+++ b/linkerd/http/prom/src/record_response.rs
@@ -31,14 +31,14 @@ pub use self::{
 ///
 /// This is specifically to support higher-cardinality status counters and
 /// lower-cardinality stream duration histograms.
-pub trait MkStreamLabel {
-    type StreamLabel: StreamLabel;
+pub trait MkStreamLabel<ReqB, RespB> {
+    type StreamLabel: StreamLabel<RespB>;
 
     /// Returns None when the request should not be recorded.
-    fn mk_stream_labeler<B>(&self, req: &http::Request<B>) -> Option<Self::StreamLabel>;
+    fn mk_stream_labeler(&self, req: &http::Request<ReqB>) -> Option<Self::StreamLabel>;
 }
 
-pub trait StreamLabel: Send + 'static {
+pub trait StreamLabel<RespB>: Send + 'static {
     type DurationLabels: EncodeLabelSet
         + Clone
         + Eq
@@ -56,7 +56,7 @@ pub trait StreamLabel: Send + 'static {
         + Sync
         + 'static;
 
-    fn init_response<B>(&mut self, rsp: &http::Response<B>);
+    fn init_response(&mut self, rsp: &http::Response<RespB>);
     fn end_response(&mut self, trailers: Result<Option<&http::HeaderMap>, &Error>);
 
     fn status_labels(&self) -> Self::StatusLabels;
@@ -64,9 +64,11 @@ pub trait StreamLabel: Send + 'static {
 }
 
 /// A set of parameters that can be used to construct a `RecordResponse` layer.
-pub struct Params<L: MkStreamLabel, M> {
+pub struct Params<L: MkStreamLabel<ReqB, RespB>, M, ReqB, RespB> {
     pub labeler: L,
     pub metric: M,
+    // Parameters are generic over their
+    _marker: std::marker::PhantomData<(ReqB, RespB)>,
 }
 
 #[derive(Clone, Debug, thiserror::Error)]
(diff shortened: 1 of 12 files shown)


linkerd/http/prom/src/record_response.rs
```diff
@@ -76,7 +58,7 @@ pub trait StreamLabel: Send + 'static {
         + Sync
         + 'static;
 
-    fn init_response<B>(&mut self, rsp: &http::Response<B>);
+    fn init_response(&mut self, rsp: &http::response::Parts);
```
cratelyn, line 61: this is a core insight i want to highlight.

we don't need or want to see the response body in our implementations of this hook. we _do_ want to e.g. record the start time, but for now, using `Parts` gives us most of the same information without having to lug around that darn `B` type parameter.

cratelyn: ditto for `MkStreamLabel`. inspecting `Parts` means that we can have object-safe versions of these traits, crucial if we want to lean on type erasure and box these up over in our application code.

cratelyn:
nb: this isn't expected to compile right now, i'm refraining from updating app code until the `linkerd-http-prom` interfaces are more situated.

linkerd/http/prom/src/record_response/response.rs
```diff
@@ -14,26 +13,20 @@ use std::{
...
+use super::{DurationFamily, MkDurationHistogram, MkStreamLabel, StreamLabel};
 
+/// Metrics type that tracks completed responses.
 #[derive(Debug)]
-pub struct ResponseMetrics<DurL, StatL> {
-    duration: DurationFamily<DurL>,
-    statuses: Family<StatL, Counter>,
+pub struct ResponseMetrics<L: StreamLabel> {
+    duration: DurationFamily<L::DurationLabels>,
+    statuses: Family<L::StatusLabels, Counter>,
```
cratelyn, line 22: here's another core idea in this branch i'd like to highlight:

in the outbound proxy code today, we define an alias tying `DurL` and `StatL` to the same `L: StreamLabel` implementation. so, the two parameters here aren't giving us any benefit or serving any need, and we can simplify the code in this submodule (_and the equivalent `request.rs` sibling_) by making this generic across `L` rather than its underlying label types.

linkerd/http/prom/src/record_response/response.rs
```diff
@@ -68,21 +60,16 @@ where
...
         (*self.statuses.get_or_create(labels)).clone()
     }
 }
 
-impl<DurL, StatL> Default for ResponseMetrics<DurL, StatL>
-where
-    StatL: EncodeLabelSet + Clone + Eq + std::fmt::Debug + std::hash::Hash + Send + Sync + 'static,
-    DurL: EncodeLabelSet + Clone + Eq + std::fmt::Debug + std::hash::Hash + Send + Sync + 'static,
-{
+impl<L: StreamLabel> Default for ResponseMetrics<L> {
```
cratelyn, line 72: :ok_hand: :woman_cook: :kiss: voila!

linkerd/http/prom/src/record_response.rs
```diff
@@ -32,30 +25,10 @@ pub use self::{
...
-        + std::hash::Hash
-        + Send
-        + Sync
-        + 'static;
-
-    type StreamLabel: StreamLabel<
-        DurationLabels = Self::DurationLabels,
-        StatusLabels = Self::StatusLabels,
-    >;
+    type StreamLabel: StreamLabel;
```
cratelyn, line 28: much simpler! :slightly_smiling_face:

linkerd/http/prom/src/record_response.rs
```diff
@@ -99,7 +72,7 @@ pub struct RequestCancelled(());
 pub struct NewRecordResponse<L, X, M, N> {
     inner: N,
     extract: X,
-    _marker: std::marker::PhantomData<fn() -> (L, M)>,
+    _marker: std::marker::PhantomData<(L, M)>,
```
sfleen, line 75: I believe `fn() -> T` allows covariance over T without triggering `dropck`. I _think_ we'd want to keep that, since this produces `(L, M)` via the extract instead of storing it.

linkerd/http/prom/src/record_response.rs
sfleen: This is awesome! Getting rid of the parameter simplified a lot by the looks of it.

sfleen (approved):
One minor nit, everything else looks great!

linkerd/http/prom/src/record_response.rs
cratelyn: great observation, thank you! 04dfa43f5395f371bd817b40c8293d47a8087b10 reverts those changes :hammer:

PR #3245, opened 2024-10-01
Allow configuring proxy trace exporter service name

sfleen:
Currently, we hard-code the trace service name to "linkerd-proxy". This allows this name to be configured through an env that can be set by the injector, annotations, etc.

linkerd/linkerd2#11157


cratelyn (approved):
LGTM!

PR #3258, opened 2024-10-04
feat(app): Add hostname label to http metrics

cratelyn:
this commit introduces an additional label to HTTP and gRPC route
metrics, containing the DNS host of requests. requests destined for an
ip address of some kind are not recorded with a hostname metric, as a
way to help minimize the impact of time series cardinality.

the core of this change is in this field addition:

```diff
// /linkerd/app/outbound/src/http/logical/policy/route/metrics/labels.rs
-pub struct Route(pub ParentRef, pub RouteRef);
+pub struct Route {
+    parent: ParentRef,
+    route: RouteRef,
+    hostname: Option<dns::Name>,
+}
```

see this part of the change to our `MkStreamLabel` implementation, used
in our metrics tracking request durtion, and counting response status
codes:

```diff
-    fn mk_stream_labeler<B>(&self, _: &::http::Request<B>) -> Option<Self::StreamLabel> {
+    fn mk_stream_labeler<B>(&self, req: &::http::Request<B>) -> Option<Self::StreamLabel> {
         let parent = self.params.parent_ref.clone();
         let route = self.params.route_ref.clone();
-        Some(metrics::LabelHttpRsp::from(metrics::labels::Route::from((
-            parent, route,
-        ))))
+        Some(metrics::LabelHttpRsp::from(metrics::labels::Route::new(
+            parent,
+            route,
+            req.uri(),
+        )))
     }
```

we now inspect the request, and use the URI to label metrics related to
this traffic by hostname.

a `http_request_hostnames()` test case is added to exercise this.

some todo comments are left, noting where we would ideally like to
simplify or generalize the machinery related to `RetryLabelExtract`, the
type that bridges the labels needed based on our `NewService<T>` target,
and the request type later accepted by the instantiated `Service<T>`.

### DIFF
--- a/linkerd/app/outbound/src/http/logical/policy/route.rs
+++ b/linkerd/app/outbound/src/http/logical/policy/route.rs
@@ -14,7 +14,6 @@ pub(crate) mod retry;
 
 pub(crate) use self::backend::{Backend, MatchedBackend};
 pub use self::filters::errors;
-use self::metrics::labels::Route as RouteLabels;
 
 pub use self::metrics::{GrpcRouteMetrics, HttpRouteMetrics};
 
@@ -117,16 +116,35 @@ where
                 .push(MatchedBackend::layer(metrics.backend.clone()))
                 .lift_new_with_target()
                 .push(NewDistribute::layer())
+                .check_new::<Self>()
+                .check_new_service::<Self, http::Request<http::BoxBody>>()
                 // The router does not take the backend's availability into
                 // consideration, so we must eagerly fail requests to prevent
                 // leaking tasks onto the runtime.
                 .push_on_service(svc::LoadShed::layer())
                 .push(filters::NewApplyFilters::<Self, _, _>::layer())
-                .push(retry::NewHttpRetry::layer(metrics.retry.clone()))
+                .push({
+                    // TODO(kate): extracting route labels like this should ideally live somewhere
+                    // else, like e.g. the `SetExtensions` middleware.
+                    let mk_extract = |rt: &Self| {
+                        let Route {
+                            parent_ref,
+                            route_ref,
+                            ..
+                        } = &rt.params;
+                        retry::RetryLabelExtract(parent_ref.clone(), route_ref.clone())
+                    };
+                    let metrics = metrics.retry.clone();
+                    retry::NewHttpRetry::layer_via_mk(mk_extract, metrics)
+                })
+                .check_new::<Self>()
+                .check_new_service::<Self, http::Request<http::BoxBody>>()
                 // Set request extensions based on the route configuration
                 // AND/OR headers
                 .push(extensions::NewSetExtensions::layer())
                 .push(metrics::layer(&metrics.requests))
+                .check_new::<Self>()
+                .check_new_service::<Self, http::Request<http::BoxBody>>()
                 // Configure a classifier to use in the endpoint stack.
                 // TODO(ver) move this into NewSetExtensions?
                 .push(classify::NewClassify::layer())
@@ -149,15 +167,6 @@ impl<T: Clone, M, F, P> svc::Param<BackendDistribution<T, F>> for MatchedRoute<T
     }
 }
 
-impl<T: Clone, M, F, P> svc::Param<RouteLabels> for MatchedRoute<T, M, F, P> {
-    fn param(&self) -> RouteLabels {
-        RouteLabels(
-            self.params.parent_ref.clone(),
-            self.params.route_ref.clone(),
-        )
-    }
-}
-
 // === impl Http ===
 
 impl<T> filters::Apply for Http<T> {
@@ -177,12 +186,14 @@ impl<T> metrics::MkStreamLabel for Http<T> {
     type DurationLabels = metrics::labels::Route;
     type StreamLabel = metrics::LabelHttpRouteRsp;
 
-    fn mk_stream_labeler<B>(&self, _: &::http::Request<B>) -> Option<Self::StreamLabel> {
+    fn mk_stream_labeler<B>(&self, req: &::http::Request<B>) -> Option<Self::StreamLabel> {
         let parent = self.params.parent_ref.clone();
         let route = self.params.route_ref.clone();
-        Some(metrics::LabelHttpRsp::from(metrics::labels::Route::from((
-            parent, route,
-        ))))
+        Some(metrics::LabelHttpRsp::from(metrics::labels::Route::new(
+            parent,
+            route,
+            req.uri(),
+        )))
     }
 }
 
@@ -231,12 +242,14 @@ impl<T> metrics::MkStreamLabel for Grpc<T> {
     type DurationLabels = metrics::labels::Route;
     type StreamLabel = metrics::LabelGrpcRouteRsp;
 
-    fn mk_stream_labeler<B>(&self, _: &::http::Request<B>) -> Option<Self::StreamLabel> {
+    fn mk_stream_labeler<B>(&self, req: &::http::Request<B>) -> Option<Self::StreamLabel> {
         let parent = self.params.parent_ref.clone();
         let route = self.params.route_ref.clone();
-        Some(metrics::LabelGrpcRsp::from(metrics::labels::Route::from((
-            parent, route,
-        ))))
+        Some(metrics::LabelGrpcRsp::from(metrics::labels::Route::new(
+            parent,
+            route,
+            req.uri(),
+        )))
     }
 }
 
--- a/linkerd/app/outbound/src/http/logical/policy/route/metrics/labels.rs
+++ b/linkerd/app/outbound/src/http/logical/policy/route/metrics/labels.rs
@@ -1,11 +1,17 @@
 //! Prometheus label types.
-use linkerd_app_core::{errors, metrics::prom::EncodeLabelSetMut, proxy::http, Error as BoxError};
+use linkerd_app_core::{
+    dns, errors, metrics::prom::EncodeLabelSetMut, proxy::http, Error as BoxError,
+};
 use prometheus_client::encoding::*;
 
 use crate::{BackendRef, ParentRef, RouteRef};
 
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
-pub struct Route(pub ParentRef, pub RouteRef);
+pub struct Route {
+    parent: ParentRef,
+    route: RouteRef,
+    hostname: Option<dns::Name>,
+}
 
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
 pub struct RouteBackend(pub ParentRef, pub RouteRef, pub BackendRef);
@@ -52,17 +58,47 @@ pub enum Error {
 
 // === impl Route ===
 
-impl From<(ParentRef, RouteRef)> for Route {
-    fn from((parent, route): (ParentRef, RouteRef)) -> Self {
-        Self(parent, route)
+impl Route {
+    pub fn new(parent: ParentRef, route: RouteRef, uri: &http::uri::Uri) -> Self {
+        let hostname = uri
+            .host()
+            .map(str::as_bytes)
+            .map(dns::Name::try_from_ascii)
+            .and_then(Result::ok);
+
+        Self {
+            parent,
+            route,
+            hostname,
+        }
+    }
+
+    #[cfg(test)]
+    pub(super) fn new_with_name(
+        parent: ParentRef,
+        route: RouteRef,
+        hostname: Option<dns::Name>,
+    ) -> Self {
+        Self {
+            parent,
+            route,
+            hostname,
+        }
     }
 }
 
 impl EncodeLabelSetMut for Route {
     fn encode_label_set(&self, enc: &mut LabelSetEncoder<'_>) -> std::fmt::Result {
-        let Self(parent, route) = self;
+        let Self {
+            parent,
+            route,
+            hostname,
+        } = self;
+
         parent.encode_label_set(enc)?;
         route.encode_label_set(enc)?;
+        ("hostname", hostname.as_deref()).encode(enc.encode_label())?;
+
         Ok(())
     }
 }
--- a/linkerd/app/outbound/src/http/logical/policy/route/metrics/tests.rs
+++ b/linkerd/app/outbound/src/http/logical/policy/route/metrics/tests.rs
@@ -4,7 +4,14 @@ use super::{
     test_util::*,
     LabelGrpcRouteRsp, LabelHttpRouteRsp, RequestMetrics,
 };
-use linkerd_app_core::svc::{self, http::BoxBody, Layer, NewService};
+use linkerd_app_core::{
+    dns,
+    svc::{
+        self,
+        http::{uri::Uri, BoxBody},
+        Layer, NewService,
+    },
+};
 use linkerd_proxy_client_policy as policy;
 
 #[tokio::test(flavor = "current_thread", start_paused = true)]
@@ -18,7 +25,7 @@ async fn http_request_statuses() {
 
     // Send one request and ensure it's counted.
     let ok = metrics.get_statuses(&labels::Rsp(
-        labels::Route(parent_ref.clone(), route_ref.clone()),
+        labels::Route::new(parent_ref.clone(), route_ref.clone(), &Uri::default()),
         labels::HttpRsp {
             status: Some(http::StatusCode::OK),
             error: None,
@@ -37,7 +44,7 @@ async fn http_request_statuses() {
     // Send another request and ensure it's counted with a different response
     // status.
     let no_content = metrics.get_statuses(&labels::Rsp(
-        labels::Route(parent_ref.clone(), route_ref.clone()),
+        labels::Route::new(parent_ref.clone(), route_ref.clone(), &Uri::default()),
         labels::HttpRsp {
             status: Some(http::StatusCode::NO_CONTENT),
             error: None,
@@ -61,7 +68,7 @@ async fn http_request_statuses() {
 
     // Emit a response with an error and ensure it's counted.
     let unknown = metrics.get_statuses(&labels::Rsp(
-        labels::Route(parent_ref.clone(), route_ref.clone()),
+        labels::Route::new(parent_ref.clone(), route_ref.clone(), &Uri::default()),
         labels::HttpRsp {
             status: None,
             error: Some(labels::Error::Unknown),
@@ -75,7 +82,7 @@ async fn http_request_statuses() {
     // Emit a successful response with a body that fails and ensure that both
     // the status and error are recorded.
     let mixed = metrics.get_statuses(&labels::Rsp(
-        labels::Route(parent_ref, route_ref),
+        labels::Route::new(parent_ref, route_ref, &Uri::default()),
         labels::HttpRsp {
             status: Some(http::StatusCode::OK),
             error: Some(labels::Error::Unknown),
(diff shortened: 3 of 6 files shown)


linkerd/app/outbound/src/http/logical/policy/route.rs
```diff
@@ -174,15 +186,15 @@ impl<T> filters::Apply for Http<T> {
 
 impl<T> metrics::MkStreamLabel for Http<T> {
     type StatusLabels = metrics::labels::HttpRouteRsp;
-    type DurationLabels = metrics::labels::Route;
+    type DurationLabels = metrics::labels::HttpRoute; // TODO(kate): do we want hostname here?
```
cratelyn, line 189: right now we'd be adding the hostname label to our durations. if cardinality concerns are going to matter anywhere, they'll be happening here.

on the other hand though, i can imagine this being tremendously useful latency info.

i'd love to hear what people think!

linkerd/app/outbound/src/http/logical/policy/route.rs
```diff
@@ -174,15 +186,15 @@ impl<T> filters::Apply for Http<T> {
...
+    fn mk_stream_labeler<B>(&self, req: &::http::Request<B>) -> Option<Self::StreamLabel> {
         let parent = self.params.parent_ref.clone();
         let route = self.params.route_ref.clone();
-        Some(metrics::LabelHttpRsp::from(metrics::labels::Route::from((
-            parent, route,
-        ))))
+        Some(metrics::LabelHttpRsp::from(
+            metrics::labels::HttpRoute::from((parent, route, req.uri().host().map(str::to_owned))),
+        ))
     }
```
cratelyn, line 198: this is the part where the actual label value gets discovered and passed into our `LabelHttpRouteRsp` label set.

cratelyn (commented):
:speech_balloon: some notes...

cratelyn:
on monday i'm going to pick this up and update the tests, which is why ci is currently failing.

linkerd/app/outbound/src/http/logical/policy/route.rs
```diff
@@ -149,15 +149,27 @@ impl<T: Clone, M, F, P> svc::Param<BackendDistribution<T, F>> for MatchedRoute<T
...
 }
 
+impl<T: Clone, M, F, P> svc::Param<HttpRouteLabels> for MatchedRoute<T, M, F, P> {
+    fn param(&self) -> HttpRouteLabels {
+        HttpRouteLabels(
+            self.params.parent_ref.clone(),
+            self.params.route_ref.clone(),
+            // we do not know the hostname at instantiation time for the stack. this is found
+            // later when we are creating our labeler.
+            None,
```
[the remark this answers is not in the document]
cratelyn: on a further read, it seems like we want to use `svc::ExtractParam`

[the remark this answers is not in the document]
cratelyn: another way, for the record, could be to store the hostname in our route match structure, but i considered that to be a little invasive.

cratelyn:
checking in an incremental snapshot, it is not expected to build right now.

linkerd/app/outbound/src/http/logical/policy/route.rs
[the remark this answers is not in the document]
cratelyn: resolved in subsequent commits. :white_check_mark:

linkerd/app/outbound/src/http/logical/policy/route.rs
cratelyn: i've refrained from adding that label here.

it's still an interesting follow-on / enhancement to consider, but i'll go ahead and mark this as resolved for now.

linkerd/app/outbound/src/http/logical/policy/route/metrics/labels.rs
```diff
@@ -50,15 +56,55 @@ pub enum Error {
...
+// === impl HttpRoute ===
+
+impl HttpRoute {
+    pub fn new(parent: ParentRef, route: RouteRef, host: Option<&str>) -> Self {
+        let host = match host.and_then(|h| url::Host::parse(h).ok()) {
+            Some(url::Host::Domain(s)) => Some(s),
+            // Refrain from emitting label values for url's whose host component are a IPv4
+            // or IPv6 address, to help prevent label cardinality from becoming too noisy.
+            None | Some(url::Host::Ipv4(_) | url::Host::Ipv6(_)) => None,
+        };
```
cratelyn, line 68: :point_up: this is where we filter out addresses from our labels

linkerd/app/outbound/src/http/logical/policy/route/metrics/labels.rs
```diff
@@ -5,17 +5,23 @@ use prometheus_client::encoding::*;
...
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
-pub struct Route(pub ParentRef, pub RouteRef);
+pub struct HttpRoute {
+    parent: ParentRef,
+    route: RouteRef,
+    host: Option<String>,
+}
+
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub struct GrpcRoute(pub ParentRef, pub RouteRef);
```
cratelyn, line 15: :point_up: this is where we add a hostname to HTTP metrics, but intentionally refrain from doing so for our gRPC metrics.

linkerd/app/outbound/src/http/logical/policy/route.rs
```diff
@@ -117,20 +116,37 @@ where
...
+                    retry::NewHttpRetry::layer_via_mk(mk_extract, metrics)
+                })
+                .check_new_service::<Self, http::Request<linkerd_app_core::proxy::http::BoxBody>>()
+                .check_new::<Self>()
                 // Set request extensions based on the route configuration
                 // AND/OR headers
                 .push(extensions::NewSetExtensions::layer())
                 .push(metrics::layer(&metrics.requests))
+                .check_new::<Self>()
+                .check_new_service::<Self, http::Request<linkerd_app_core::proxy::http::BoxBody>>()
```
cratelyn, line 145: :point_up: this is mentioned in the pr description. i am happy to walk these additions back; they made my own work easier, and it seemed like an easy gift to offer future engineers tweaking our policy stack. :gift:

linkerd/app/outbound/src/http/logical/policy/route.rs
```diff
@@ -149,15 +165,6 @@ impl<T: Clone, M, F, P> svc::Param<BackendDistribution<T, F>> for MatchedRoute<T
...
 }
 
-impl<T: Clone, M, F, P> svc::Param<RouteLabels> for MatchedRoute<T, M, F, P> {
-    fn param(&self) -> RouteLabels {
-        RouteLabels(
-            self.params.parent_ref.clone(),
-            self.params.route_ref.clone(),
-        )
-    }
-}
```
cratelyn, line 159: :point_up: here's the core bit driving the deferral of route label extraction, this is no longer some we can know in advance (_because there's a hostname for HTTP metrics_). instead, we use an `ExtractParam<P, T>` to get our labels upon being given a request to inspect.

linkerd/app/outbound/src/http/logical/policy/route/metrics/tests.rs
```diff
@@ -99,6 +99,127 @@ async fn http_request_statuses() {
...
 }
 
+#[tokio::test(flavor = "current_thread", start_paused = true)]
+async fn http_request_hostnames() {
+    const HOST_1: &str = "great.website";
+    const URI_1_1: &str = "https://great.website/path/to/index.html#fragment";
+    const URI_1_2: &str = "https://great.website/another/index.html";
+    const HOST_2: &str = "different.website";
+    const URI_2: &str = "https://different.website/index.html";
+    const URI_3: &str = "https://[3fff::]/index.html";
```
cratelyn, line 109: :point_up: tests! we show that our labels look right when sending traffic to different hosts, and to places with an opaque address.

linkerd/http/prom/src/record_response/request.rs
```diff
@@ -72,6 +72,9 @@ where
     pub fn get_statuses(&self, labels: &StatL) -> Counter {
         (*self.statuses.get_or_create(labels)).clone()
     }
+
+    // TODO(kate): it'd be nice if we could avoid creating a time series if it does not exist,
+    // so that tests can confirm that certain label sets do not exist within the family.
```
cratelyn, line 77: i've discussed this in air, but it seems like it'd be a very straightforward change to drive upstream.

linkerd/http/retry/src/lib.rs
```diff
@@ -81,31 +83,45 @@ struct Metrics {
 
 // === impl NewHttpRetry ===
 
-impl<P, L: Clone, N> NewHttpRetry<P, L, N> {
-    pub fn layer(metrics: MetricFamilies<L>) -> impl layer::Layer<N, Service = Self> + Clone {
+impl<P, X, L: Clone, F: Clone, N> NewHttpRetry<P, L, F, X, N> {
+    pub fn layer_via_mk(
+        mk_extract: F,
+        metrics: MetricFamilies<L>,
+    ) -> impl tower::layer::Layer<N, Service = Self> + Clone {
```
cratelyn, line 90: :point_up: bike-shedding welcome. other `layer_via` functions conventionally receive an `X` extractor, in our case we want a function that takes a reference to the `NewService<T>` target (_a matched route_) to construct an `X` that can later extract our `L` labels (_including a hostname_) provided the `Service<T>` target (_a request_).

in a sense this is almost like a `NewExtractParam<..>`, if we can compare it to `NewService`. see the commit message for a little bit more information.

linkerd/app/outbound/src/http/logical/policy/route/retry.rs
```diff
@@ -29,6 +31,9 @@ pub struct RetryPolicy {
     pub backoff: Option<ExponentialBackoff>,
 }
 
+#[derive(Clone, Debug)]
+pub struct RetryLabelExtract(pub ParentRef, pub RouteRef);
```
cratelyn, line 35: our new extractor, which hangs onto the matched route's parent and route metadata.

linkerd/app/outbound/src/http/logical/policy/route/retry.rs
```diff
@@ -158,3 +163,14 @@ impl RetryPolicy {
...
+// === impl RetryLabelExtract ===
+
+impl<B> ExtractParam<RouteLabels, http::Request<B>> for RetryLabelExtract {
+    fn extract_param(&self, t: &http::Request<B>) -> RouteLabels {
+        let Self(parent, route) = self;
+        let host = t.uri().host();
+
+        RouteLabels::new(parent.clone(), route.clone(), host)
+    }
+}
```
cratelyn, line 176: and here, we actually get the hostname, provided the `Service<T>` target: a request with some sort of body.

cratelyn (commented):
:mag: :bread: breadcrumbs and discussion seeds for reviewers...

linkerd/app/outbound/src/http/logical/policy/route.rs
sfleen: I'm more than fine with these, the only real concern would be compile time impact but I don't expect these to be a burden here.

linkerd/http/retry/src/lib.rs
sfleen: It took me a bit to fully grok this (the several layers of tower abstraction are always a bit of a pain to punch through), but I think this is probably the best option given the naming scheme we already have to work with.

sfleen (approved):
LGTM, I like the overall structure and nothing jumps out at me as too complex given what we're already working with.

linkerd/app/outbound/src/http/logical/policy/route.rs
```diff
@@ -117,20 +116,37 @@ where
...
                 // AND/OR headers
                 .push(extensions::NewSetExtensions::layer())
                 .push(metrics::layer(&metrics.requests))
+                .check_new::<Self>()
+                .check_new_service::<Self, http::Request<linkerd_app_core::proxy::http::BoxBody>>()
                 // Configure a classifier to use in the endpoint stack.
                 // TODO(ver) move this into NewSetExtensions?
                 .push(classify::NewClassify::layer())
-                .push(svc::NewMapErr::layer_with(|rt: &Self| {
+                .push(svc::NewMapErr::<_, _, _>::layer_with(|rt: &Self| {
```
cratelyn, line 149: ill remove this when im back at a keyboard!

linkerd/app/outbound/src/http/logical/policy/route/metrics/labels.rs
olix0r: We already have some prior art in the proxy that avoids the url dependency and is oriented the hyper URI types. Can [`Addr`](https://github.com/linkerd/linkerd2-proxy/blob/89ba5b5d8b39f54de8a8c6e1e8cf552e9c9a5527/linkerd/addr/src/lib.rs#L50) work here?

cratelyn: i took a look at the `Addr` types, but these seem subtly different from what we want here. those types are focused on working with the authority, which includes the port in addition to the hostname.

for what it is worth, we do have `url` as a transitive dependency in the workspace today, via the hickory-resolve crate, see: https://github.com/linkerd/linkerd2-proxy/pull/3258/commits/9e3ee3e6611ea96e18d74cf687a33a4c86eb04a8

my chief concern is that using the authority might exacerbate the cardinality problem, as we would have a time series counting e.g. each status code, for each port, for each host. that extra dimension could be costly, until we have some way to limit the number of time series we are emitting.

on the upside, this might help disambiguate between services running on shared infrastructure.

i'm not tremendously opinionated about hostname vs. authority though, and perhaps reusing the `Addr` crate is a nice benefit.

what do you think, @olix0r?

olix0r: Ah, I see what you mean.

I don't think that cardinality is particularly a concern, since parent refs may generally include port information; and disambiguating services by port is pretty natural.

I think the bigger concern is how this compares with values in the TLS case, where we'll have ServerNames that omit ports.

But we still could avoid the url types with, e.g. [`NameAddr::name` ](https://github.com/linkerd/linkerd2-proxy/blob/89ba5b5d8b39f54de8a8c6e1e8cf552e9c9a5527/linkerd/addr/src/lib.rs#L222).

I.e. from an Addr you can get at it via something like `addr.name_addr().and_then(NameAddr::name)`.

This also ensures some amount of validation (i.e. over &str) where we can accept invalid hostnames.

linkerd/app/outbound/src/http/logical/policy/route.rs
olix0r: We do tend to prune tehse down to avoid noise, but it's not really a problem to leave them in. If we do leave them in, though, I'd strongly prefer that we rely on the reexports to reduce noise:
```suggestion
                .check_new::<Self>()
                .check_new_service::<Self, http::Request<http::BoxBody>>()
```

linkerd/http/retry/src/lib.rs
```diff
@@ -81,31 +83,45 @@ struct Metrics {
...
     }
 }
 
-impl<T, P, L, N> NewService<T> for NewHttpRetry<P, L, N>
+impl<T, X, P, L, F, N> NewService<T> for NewHttpRetry<P, L, F, X, N>
 where
-    T: Param<L>,
     P: Policy,
     L: Clone + std::fmt::Debug + Hash + Eq + Send + Sync + prom::encoding::EncodeLabelSet + 'static,
+    F: Fn(&T) -> X + Clone,
```
olix0r, line 104: Eh, this gives me a bit of pause... we're adding an unnamable type parameter here. I think this is basically an ExtractParam?

linkerd/http/retry/src/lib.rs
```diff
@@ -179,21 +195,13 @@ where
...
-}
-
-impl<P, S> Service<http::Request<BoxBody>> for HttpRetry<P, S>
+impl<P, L, X, S> Service<http::Request<BoxBody>> for HttpRetry<P, L, X, S>
 where
     P: Policy,
     P: Param<Params>,
     P: Clone + Send + Sync + std::fmt::Debug + 'static,
+    L: Clone + std::fmt::Debug + Hash + Eq + Send + Sync + prom::encoding::EncodeLabelSet + 'static,
+    X: ExtractParam<L, http::Request<BoxBody>>,
```
olix0r, line 204: We tend to rely on ExtractParam only for Targets and less so for request parameters... Is https://github.com/linkerd/linkerd2-proxy/pull/3258/commits/69872cb5a5ed06492f023a688bf09978a2ca1411 a change that can be reviewed independently of other changes in this PR?

linkerd/http/retry/src/lib.rs
olix0r: Note https://github.com/linkerd/linkerd2-proxy/blob/89ba5b5d8b39f54de8a8c6e1e8cf552e9c9a5527/linkerd/stack/src/lib.rs#L115-L122

olix0r: To keep things ~consistentish, I'd be inclined to use `X: ExtractParam<T, ReqX>` in place of `F`--since X tends to be used for a when a NewService needs an ExtractParam, and then have a secondary `ReqX` (or something) for the second level of extraction.

cratelyn:
pushed some commits addressing some of the low-hanging fruit, there is still some feedback to address... :apple: :white_check_mark:

linkerd/http/retry/src/lib.rs
cratelyn: 750bd6e renames `F` and `X` to `X` and `ReqX`, respectively. 323dedb changes the trait bounds to use the blanket implementation linked above. good call!

linkerd/app/outbound/src/http/logical/policy/route.rs
cratelyn: ee5a353d removed this.

linkerd/app/outbound/src/http/logical/policy/route.rs
cratelyn: db00264b tweaks these to refer to the reexports. thanks for pointing that out, i hadn't realized that `http` in this scope is pointed to `linkerd_app_core::proxy::http`. :bulb:

this looks much better.

linkerd/http/retry/src/lib.rs
cratelyn: https://github.com/linkerd/linkerd2-proxy/pull/3258#discussion_r1794142955 for posterity, some of the changes in this thread helped clarify the layers of :cake: :sparkles: abstraction extraction

linkerd/app/outbound/src/http/logical/policy/route/metrics/labels.rs
cratelyn: i've pushed aed9d303. this commit does two things: (1) avoids a `url` dependency, via use of `linkerd_app_core::dns::Name`, and (2) tweaks the interfaces of `http::logical::policy::route::metrics::labels::HttpRoute` to accept `http` types rather than an `Option<&str>`.

e2447dc0 then reverts our addition of a direct `url` dependency.

this bore out really nicely, thank you for taking the time to work through this! how does this look to you now, @olix0r?

linkerd/http/retry/src/lib.rs
cratelyn: i've given it a shot, but it seems difficult to extricate that change into its own PR.

to restate my understanding, based on our conversation yesterday, one approach worth trying would be using a request extension rather than an `ExtractParam<P, T>`.

at risk of a brief detour, when we say "Target", we mean the `T` types provided as parameters to e.g. `Http<T>` or `Matched<T>`, at the `NewService<T>` layer. is that right? `Request<BodyBody>` is a concrete target for _this_ `Service<T>`, which had me confused, so i wanted to restate my understanding of this after reading more where we use `ExtractParam`.

i took a swing at this, here: https://github.com/linkerd/linkerd2-proxy/tree/kate/add-hostname-label-to-http-metrics.failed-extension-attempt

i found myself running into a lot of trouble with this, unfortunately. there's still something slightly incongruent stemming from the fact that we have labels that straddle the boundary between a construction time (_route and parent references_) and call-time (_request `Parts`_).

linkerd/app/outbound/src/http/logical/policy/route.rs
```diff
@@ -117,16 +116,33 @@ where
...
+                .check_new::<Self>()
+                .check_new_service::<Self, http::Request<http::BoxBody>>()
                 // The router does not take the backend's availability into
                 // consideration, so we must eagerly fail requests to prevent
                 // leaking tasks onto the runtime.
                 .push_on_service(svc::LoadShed::layer())
                 .push(filters::NewApplyFilters::<Self, _, _>::layer())
-                .push(retry::NewHttpRetry::layer(metrics.retry.clone()))
+                .push({
+                    let mk_extract = |rt: &Self| {
```
olix0r, line 127: Let's add a TODO here to note that this extractions of labels probably belongs somewhere else (like SetExtensions).

linkerd/app/outbound/src/http/logical/policy/route/metrics/labels.rs
```diff
@@ -1,21 +1,29 @@
...
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
-pub struct Route(pub ParentRef, pub RouteRef);
+pub struct HttpRoute {
+    parent: ParentRef,
+    route: RouteRef,
+    name: Option<dns::Name>,
+}
+
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub struct GrpcRoute(pub ParentRef, pub RouteRef);
```
olix0r, line 17: I think we should leave these as unified so that gRPC routes also include a hostname.

linkerd/app/outbound/src/http/logical/policy/route/metrics/labels.rs
```diff
@@ -1,21 +1,29 @@
...
 use prometheus_client::encoding::*;
 
 use crate::{BackendRef, ParentRef, RouteRef};
 
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
-pub struct Route(pub ParentRef, pub RouteRef);
+pub struct HttpRoute {
+    parent: ParentRef,
+    route: RouteRef,
+    name: Option<dns::Name>,
```
olix0r, line 13: 'name' can be a bit ambiguous in the context of our code... especially with regard to identity vs hosts vs ...

```suggestion
    hostname: Option<dns::Name>,
```

linkerd/http/retry/src/lib.rs
```diff
@@ -223,6 +231,8 @@ where
         };
 
         let params = policy.param();
+        let labels = self.extract.extract_param(&req);
```
olix0r, line 234: Perhaps worth a TODO that we'd like to avoid knowing about labels here in the future.

cratelyn: added in 1c03f91 :hammer:

linkerd/http/retry/src/lib.rs
cratelyn: we resolved this in-air, and confirmed the "target" terminology. :heavy_check_mark:

linkerd/app/outbound/src/http/logical/policy/route/metrics/labels.rs
cratelyn: marking this as resolved, based on in-air conversation. :slightly_smiling_face:

cratelyn:
**note:** i have squashed this branch into a single commit, due to the fact that some changes were walked back in the process of review. https://github.com/linkerd/linkerd2-proxy/tree/kate/add-hostname-label-to-http-metrics.pre-squash-24-10-14 this branch has been pushed for posterity, which contains these changes and incremental review commits left intact.

after squashing, i've also rebased this branch on top of the latest state of `main`.

cratelyn:
:curly_loop: tying up a final loop, updated the PR description above, and the commit message to be up-to-date with what we're landing now that we've gone through review. this should now be ready to ship :shipit:

olix0r (approved):
🌮

PR #3262, opened 2024-10-07
Extract zone locality into separate TCP metric

sfleen:
Currently, the destination controller populates a zone locality field in the metric labels for an endpoint. This works fine, but can be expensive to query against in prometheus due to the high cardinality of the metrics it is attached to.

This introduces a separate `outbound_tcp_bytes_total` metric with only the zone locality as a label, which is the most common use of this label.

This includes some extra infrastructure to wire the new metric through as we currently don't use prometheus-client for transport metrics. Eventually, the existing metrics could and/or should be migrated, but that is saved for a future change.

### DIFF
--- a/linkerd/app/core/src/metrics.rs
+++ b/linkerd/app/core/src/metrics.rs
@@ -9,14 +9,13 @@
 pub use crate::transport::labels::{TargetAddr, TlsAccept};
 use crate::{
     classify::Class,
-    control, http_metrics, opencensus, opentelemetry, profiles, stack_metrics,
-    svc::Param,
-    tls,
+    control, http_metrics, opencensus, opentelemetry, profiles, proxy, stack_metrics, svc, tls,
     transport::{self, labels::TlsConnect},
 };
 use linkerd_addr::Addr;
 pub use linkerd_metrics::*;
 use linkerd_proxy_server_policy as policy;
+use prometheus_client::encoding::EncodeLabelValue;
 use std::{
     fmt::{self, Write},
     net::SocketAddr,
@@ -102,7 +101,30 @@ pub struct OutboundEndpointLabels {
     pub server_id: tls::ConditionalClientTls,
     pub authority: Option<http::uri::Authority>,
     pub labels: Option<String>,
+    pub zone_locality: OutboundZoneLocality,
     pub target_addr: SocketAddr,
+}
+
+#[derive(Debug, Copy, Clone, Default, Hash, Eq, PartialEq, EncodeLabelValue)]
+pub enum OutboundZoneLocality {
+    #[default]
+    Unknown,
+    Local,
+    Remote,
+}
+
+impl OutboundZoneLocality {
+    pub fn new(metadata: &proxy::api_resolve::Metadata) -> Self {
+        if let Some(is_zone_local) = metadata.is_zone_local() {
+            if is_zone_local {
+                OutboundZoneLocality::Local
+            } else {
+                OutboundZoneLocality::Remote
+            }
+        } else {
+            OutboundZoneLocality::Unknown
+        }
+    }
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
@@ -217,7 +239,7 @@ impl Metrics {
 
 // === impl CtlLabels ===
 
-impl Param<ControlLabels> for control::ControlAddr {
+impl svc::Param<ControlLabels> for control::ControlAddr {
     fn param(&self) -> ControlLabels {
         ControlLabels {
             addr: self.addr.clone(),
@@ -358,6 +380,12 @@ impl FmtLabels for RouteAuthzLabels {
             self.authz.kind(),
             self.authz.name(),
         )
+    }
+}
+
+impl svc::Param<OutboundZoneLocality> for OutboundEndpointLabels {
+    fn param(&self) -> OutboundZoneLocality {
+        self.zone_locality
     }
 }
 
--- a/linkerd/app/outbound/src/http/concrete.rs
+++ b/linkerd/app/outbound/src/http/concrete.rs
@@ -2,10 +2,14 @@
 //! and distributes HTTP requests among them.
 
 use super::{balance::EwmaConfig, client, handle_proxy_error_headers};
-use crate::{http, stack_labels, BackendRef, Outbound, ParentRef};
+use crate::{
+    http, stack_labels,
+    zone::{tcp_zone_labels, TcpZoneLabels},
+    BackendRef, Outbound, ParentRef,
+};
 use linkerd_app_core::{
     config::{ConnectConfig, QueueConfig},
-    metrics::{prefix_labels, EndpointLabels, OutboundEndpointLabels},
+    metrics::{prefix_labels, EndpointLabels, OutboundEndpointLabels, OutboundZoneLocality},
     profiles,
     proxy::{
         api_resolve::{ConcreteAddr, Metadata, ProtocolHint},
@@ -212,9 +216,22 @@ where
         OutboundEndpointLabels {
             authority: self.parent.param(),
             labels: prefix_labels("dst", self.metadata.labels().iter()),
+            zone_locality: self.param(),
             server_id: self.param(),
             target_addr: self.addr.into(),
         }
+    }
+}
+
+impl<T> svc::Param<OutboundZoneLocality> for Endpoint<T> {
+    fn param(&self) -> OutboundZoneLocality {
+        OutboundZoneLocality::new(&self.metadata)
+    }
+}
+
+impl<T> svc::Param<TcpZoneLabels> for Endpoint<T> {
+    fn param(&self) -> TcpZoneLabels {
+        tcp_zone_labels(self.param())
     }
 }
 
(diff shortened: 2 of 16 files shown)


linkerd/app/core/src/metrics.rs
```diff
@@ -102,9 +102,18 @@ pub struct OutboundEndpointLabels {
     pub server_id: tls::ConditionalClientTls,
     pub authority: Option<http::uri::Authority>,
     pub labels: Option<String>,
+    pub zone_locality: OutboundZoneLocality,
     pub target_addr: SocketAddr,
 }
 
+#[derive(Debug, Copy, Clone, Default, Hash, Eq, PartialEq, EncodeLabelValue)]
+pub enum OutboundZoneLocality {
+    #[default]
```
cratelyn, line 111: TIL! this syntax stabilized in https://doc.rust-lang.org/std/default/trait.Default.html#enums in rust 1.62.0. :heart_eyes:

linkerd/app/outbound/src/http/endpoint.rs
```diff
@@ -5,6 +5,7 @@ use super::{
     NewRequireIdentity,
 };
 use crate::{tcp::tagged_transport, Outbound};
+use linkerd_app_core::metrics::OutboundZoneLocality;
 use linkerd_app_core::{
```
cratelyn, line 9: nit: this should be placed with the other `linkerd_app_core` imports

linkerd/app/outbound/src/opaq/concrete.rs
```diff
@@ -323,12 +324,27 @@ where
...
+            if is_zone_local {
+                OutboundZoneLocality::Local
+            } else {
+                OutboundZoneLocality::Remote
+            }
+        } else {
+            OutboundZoneLocality::Unknown
+        }
+    }
+}
```
cratelyn, line 346: if feels a little unfortunate that this is repeated with `linkerd/app/outbound/src/http/concrete.rs`. i wonder if we could share some code somehow. if we added a constructor like...

```rust
impl OutboundZoneLocality {
    fn new(metadata: &Metadata) -> Self {
        // if let Some ... 
    }
}
```

we might be able to dry these up, and have each of these `svc::Param<OutboundZoneLocality>` implementations both delegate to that shared logic.

linkerd/proxy/api-resolve/src/metadata.rs
```diff
@@ -75,6 +79,7 @@ impl Metadata {
             authority_override,
             weight,
             http2,
+            is_zone_local: zone_locality,
```
cratelyn, line 82:
```suggestion
            is_zone_local,
```

nit: we could rename this parameter and rely on punning. it seems subjectively clearer to put an `is_..` prefix on a boolean too?

linkerd/app/core/src/metrics.rs
sfleen: It's super handy, it only works on field-less enum variants for now but I think there's plans to change that in the near future.

cratelyn (commented):
i see that this _is_ still marked as a draft, but you mentioned wanting some feedback earlier, so i took a look through this.

this looks good! i left one small comment wondering if it's possible to share the logic that converts a boolean to a label value.

linkerd/app/outbound/src/http/endpoint.rs
sfleen: Done

linkerd/app/outbound/src/opaq/concrete.rs
sfleen: Agree, this is one thing in here I wanted to fix before converting to a regular PR. Good call on the constructor, that's made this considerably cleaner.

linkerd/proxy/api-resolve/src/metadata.rs
sfleen: Done, looks like this was a variable missed in an automated rename.

linkerd/app/outbound/src/zone.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+impl TcpZoneMetricsParams {
+    pub fn register(registry: &mut prom::Registry) -> Self {
+        let bytes = prom::Family::default();
+        registry.register(
+            "bytes",
+            "The total amount of read and written bytes to this proxy",
+            bytes.clone(),
+        );
```
olix0r, line 19: We probably want to use [register_with_unit](https://docs.rs/prometheus-client/latest/prometheus_client/registry/struct.Registry.html#method.register_with_unit) to enforce best practices (including unit comment metadata). This will also force to choose a more narrowly descriptive/disambiguating name...

Off the top of my head, maybe something like `transfer_cost`, which will give us something like `outbound_tcp_transfer_cost_bytes_total`.

To that end, do we need to separate the receive and send sides of this to accurately reflect cost? I imagine that upload and download are not necessarily priced identically. We could reflect that with op=Recv and op=Send labels, or we could use distinct metrics -- the former being more amenable to aggregation when we do not have distinct upload/download costs.

sfleen: Good call on the unit and naming, the name was definitely something that needed refinement.

As far as separating out the metrics, I think one metric with labels is better for aggregation as you mention. I'd imagine symmetric costs is the more common scenario, while still leaving the option for separating them out.

linkerd/transport-metrics/src/zone/sensor.rs
```diff
@@ -0,0 +1,24 @@
+use crate::zone::TcpZoneMetrics;
+
+#[derive(Clone, Debug)]
+pub struct ZoneMetricsSensor {
+    pub metrics: TcpZoneMetrics,
```
olix0r, line 5: this scope can be narrowed, right?

```suggestion
    pub(super) metrics: TcpZoneMetrics,
```

linkerd/app/core/src/metrics.rs
```diff
@@ -9,14 +9,15 @@
...
-    svc::Param,
-    tls,
+    control, http_metrics, opencensus, opentelemetry, profiles, stack_metrics, tls,
     transport::{self, labels::TlsConnect},
 };
 use linkerd_addr::Addr;
 pub use linkerd_metrics::*;
+use linkerd_proxy_api_resolve::Metadata;
 use linkerd_proxy_server_policy as policy;
+use linkerd_stack::Param;
```
olix0r, line 19: nit: We tend to use the local reexports like svc::Param (as before) and proxy::api_resolve::Metadata

olix0r (commented):
Only minor comments.

linkerd/transport-metrics/src/zone/sensor.rs
sfleen: This can't be narrowed because it's in the associated type for the client (which itself has to be public).

olix0r: I recognize that the _type_ cannot be narrowed, but I was suggesting that the _field_ scope be narrowed, since it's only referehced in the client module (right?).

linkerd/app/outbound/src/http/concrete.rs
```diff
@@ -5,6 +5,7 @@ use super::{balance::EwmaConfig, client, handle_proxy_error_headers};
 use crate::{http, stack_labels, BackendRef, Outbound, ParentRef};
 use linkerd_app_core::{
     config::{ConnectConfig, QueueConfig},
+    metrics::OutboundZoneLocality,
     metrics::{prefix_labels, EndpointLabels, OutboundEndpointLabels},
```
olix0r, line 9: nit, consistency.

linkerd/app/outbound/src/zone.rs
```diff
@@ -0,0 +1,79 @@
...
+pub struct TcpZoneMetricsParams {
+    transfer_cost: prom::Family<TcpZoneLabels, prom::Counter>,
+}
+
+impl TcpZoneMetricsParams {
+    pub fn register(registry: &mut prom::Registry) -> Self {
+        let transfer_cost = prom::Family::default();
+        registry.register_with_unit(
+            "transfer_cost",
+            "The total cost of data transferred in and out of this proxy in bytes",
```
olix0r, line 20: More yakshaving, not a blocker:

```suggestion
            "The total amount data transferred in and out of this proxy, by cost zone",
```

linkerd/app/outbound/src/zone.rs
```diff
@@ -0,0 +1,79 @@
...
+    registry::Unit,
+};
+
+#[derive(Clone, Debug, Default)]
+pub struct TcpZoneMetricsParams {
+    transfer_cost: prom::Family<TcpZoneLabels, prom::Counter>,
+}
+
+impl TcpZoneMetricsParams {
+    pub fn register(registry: &mut prom::Registry) -> Self {
```
olix0r, line 16: It would be most idiomatic to move this into linkerd/transport-metrics/src/zone.rs and have it be generic over the label type -- then the metrics details can all be private to the module and the labeling context remains in the outbound app.

olix0r (approved):
This looks good generally! I'd recommend moving the MetricsFamily into the leaf crate for idiomatic/organizational reasons, but I don't think that substantially impacts the shape of the solution.

linkerd/app/outbound/src/zone.rs
sfleen: Took a bit to unwind some orphan rule errors I was getting, but I agree this makes more sense.

PR #3272, opened 2024-10-10
Ver/zd/tls stack

olix0r:
commit f78aa64be9afa2ab96c03f384b9fe78d71fd4942
Author: Oliver Gould <ver@buoyant.io>
Date:   Thu Oct 10 14:18:31 2024 +0000

    tls: Avoid InsertParam parameter.
    
    We don't actually use InsertParam all that much--only in the TLS server (which
    is obviously why it was included here). This change removes the InsertParam in
    favor of using a tuple, generally reducing boilerplate.
    
    It turns out that the TLS stack already has a map_target to handle turning the
    tuple-target into a Tls type, so it shouldn't be needed.

commit 67c0e849f8368473f722c6cdb0bffaecca5c8635
Author: Oliver Gould <ver@buoyant.io>
Date:   Thu Oct 10 14:49:32 2024 +0000

    tls: Remove ExtractParam from detect_sni
    
    Similarly, we don't actually care about extracting a timeout from the target.
    Using an ExtractParam causes needless boilerplate.
    
    This change updates the stack module to simply take a timeout parameter at
    construction time.

commit f6202d661163d7fdf5692b3a30f2a93e1e3496b8
Author: Oliver Gould <ver@buoyant.io>
Date:   Thu Oct 10 15:03:42 2024 +0000

    tls: Make the detect_tls module private
    
    We now only need to export the NewDetectSni type. The module reexport is not
    necessary.

commit ae88d2eecb7570b12161e4983c7890f56a7cdb88 (HEAD -> ver/zd/tls-stack, origin/ver/zd/tls-stack)
Author: Oliver Gould <ver@buoyant.io>
Date:   Thu Oct 10 15:20:17 2024 +0000

    tls: Reorganize NewDetectRequiredSni under the server module
    
    Because the DetectTls and DetectSni types are so similar -- and implemented in
    the context of a server inspecting a provided connection (and not a client
    establishing a TLS connection), this change reorganizes the module:
    
    * The DetectSni types are renamed to DetectRequiredSni to better reflect their
      purpose and difference from the DetectTls type.
    * The detect_sni module is renamed and moved to server::required_sni. This
      module is private and the relevant types are reexported from the server

### DIFF
--- a/linkerd/tls/src/server/required_sni.rs
+++ b/linkerd/tls/src/server/required_sni.rs
@@ -22,20 +22,32 @@ pub struct SniDetectionTimeoutError;
 #[error("Could not find SNI")]
 pub struct NoSniFoundError;
 
+/// A NewService that instruments an inner stack with knowledge of the
+/// connection's TLS ServerName (i.e. from an SNI header).
+///
+/// This differs from the parent module's NewDetectTls in a a few ways:
+///
+/// - It requires that all connections have an SNI.
+/// - It assumes that these connections may not be terminated locally, so there
+///   is no concept of a local server name.
+/// - There are no special affordances for mutually authenticated TLS, so we
+///   make no attempt to detect the client's identity.
+/// - The detection timeout is fixed and cannot vary per target (for
+///   convenience, to reduce needless boilerplate).
 #[derive(Clone, Debug)]
-pub struct NewDetectSni<N> {
+pub struct NewDetectRequiredSni<N> {
     inner: N,
     timeout: time::Duration,
 }
 
 #[derive(Clone, Debug)]
-pub struct DetectSni<T, N> {
+pub struct DetectRequiredSni<T, N> {
     target: T,
     inner: N,
     timeout: time::Duration,
 }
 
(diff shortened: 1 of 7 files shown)


linkerd/tls/src/server/required_sni.rs
```diff
@@ -0,0 +1,116 @@
...
+/// A NewService that instruments an inner stack with knowledge of the
+/// connection's TLS ServerName (i.e. from an SNI header).
+///
+/// This differs from the parent module's NewDetectTls in a a few ways:
+///
+/// - It requires that all connections have an SNI.
+/// - It assumes that these connections may not be terminated locally, so there
+///   is no concept of a local server name.
+/// - There are no special affordances for mutually authenticated TLS, so we
+///   make no attempt to detect the client's identity.
```
zaharidichev, line 34: What are the special affordances in the other case? From my understand, in the `DetectTLS`, we determine whether we terminate based on whether the SNI matches the local identitty. Is that what you are refering to here?

PR #3294, opened 2024-10-23
Rate Limiting PoC

alpeb:
Some caveats:

- This doesn't abide yet to the agreed API as declared in linkerd/linkerd2-proxy-api#388 . Instead, this is based on an earlier API proposal using "specifiers" to define the buckets keys.
- The limiting logic has been added directly into the inbound http policy middleware. It relies on [governor](https://docs.rs/governor/latest/governor/). If we're not allowing to configure bursting in the first implementation, we might reconsider implementing something simpler directly ourselves.
- There is actually an additional middleware (`RateLimitPolicyService`) that is currently commented out (used in the initial demo), that implemented a simpler approach, inspired by Tower's own rate-limiting middleware.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -380,6 +380,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-utils"
+version = "0.8.20"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


alpeb:
Closing in favor of #3305 and #3307

PR #3304, opened 2024-10-29
policy: use client-policy responses in opaq stack

zaharidichev:
This work (based on the early [sketch seen here](https://github.com/linkerd/linkerd2-proxy/compare/ver/opaq-policy)) changes the `opaq` routing stack to use not only profiles routing but to consider client policy responses as well. The way this is done is that we use the `opaq:Routes` type to express the routing information our stack contains. This type is geared towards the client-policy API. Whenever we need to use the profiles response, we simply translate its contents to a client-policy one.

Notable changes in this PR are: 

- we introduce an `OpaqSidecar` type that holds `opaq::Routes` objects which are routes that have been constructed from either `profiles` or `policy` response
- during the construction of the `OpaqSidecar` we consider profiles responses only if there are extra targets in the response (in order to support `TrafficSplit` functionality). In any other case we prefer the policy response.
- we depend on the new, not yet released proxy API in order to support the notion of invalid routes and backends and enable explicit failure for TCP and TLS routes.
- we eagerly fail routes that have been marked as forbidden and record that in tcp/tls metrics.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1338,6 +1338,7 @@ dependencies = [
  "linkerd-io",
  "linkerd-meshtls",
  "linkerd-meshtls-rustls",
+ "linkerd-opaq-route",
  "linkerd-proxy-client-policy",
  "linkerd-retry",
  "linkerd-stack",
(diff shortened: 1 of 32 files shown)


olix0r (commented):
This draft looks good, but for the sake of keeping our commits intentional, it would be good if we can introduce the "refactor" changes to the opaque stack independently of new features and API changes...

Are the API changes necessary to update the opaq stack? Can we introduce the _new_ features separately?

zaharidichev:
closed in favor of #3306

PR #3305, opened 2024-10-29
Add rate-limiters to ServerPolicy

alpeb:
This adds the local_rate_limit module to the server-policy crate, that
`ServerPolicy` uses for its new `local_rate_limit` field, containing
three optional rate-limiters: total, identity, overrides (this one is
really a vector of limiters, one per configured override).

I tried putting that under `Protocol` instead, but the `PartialEq`
requirement made it very hard to follow. `Server` OTOH doesn't really
require that trait, so I was able to remove it and accommodate the
limiters.

I made sure to avoid pulling the dashmap dependency in `governor`; ~I
haven't checked yet the necessity of the "jitter" and "quanta" features.~

This temporarily overrides linkerd2-proxy-api dependency to pick changes
from linkerd/linkerd2-proxy-api#388

## Update

The `HttpPolicyService` middleware has been expanded to call the `local_rate_limit` module above to perform the rate-limit check.

### DIFF
--- a/linkerd/app/inbound/src/http/server.rs
+++ b/linkerd/app/inbound/src/http/server.rs
@@ -203,9 +203,7 @@ impl errors::HttpRescue<Error> for ServerRescue {
             ));
         }
 
-        if errors::is_caused_by::<linkerd_proxy_server_policy::local_rate_limit::RateLimitError>(
-            &*error,
-        ) {
+        if errors::is_caused_by::<linkerd_proxy_server_policy::RateLimitError>(&*error) {
             return Ok(errors::SyntheticHttpResponse::rate_limited(error));
         }
 
--- a/linkerd/app/inbound/src/http/server.rs
+++ b/linkerd/app/inbound/src/http/server.rs
@@ -203,6 +203,12 @@ impl errors::HttpRescue<Error> for ServerRescue {
             ));
         }
 
+        if errors::is_caused_by::<linkerd_proxy_server_policy::local_rate_limit::RateLimitError>(
+            &*error,
+        ) {
+            return Ok(errors::SyntheticHttpResponse::rate_limited(error));
+        }
+
         if errors::is_caused_by::<crate::GatewayDomainInvalid>(&*error) {
             return Ok(errors::SyntheticHttpResponse::not_found(error));
         }
--- a/linkerd/app/inbound/src/policy/http.rs
+++ b/linkerd/app/inbound/src/policy/http.rs
@@ -9,7 +9,7 @@ use linkerd_app_core::{
     svc::{self, ServiceExt},
     tls,
     transport::{ClientAddr, OrigDstAddr, Remote},
-    Error, Result,
+    Conditional, Error, Result,
 };
 use linkerd_proxy_server_policy::{grpc, http, route::RouteMatch};
 use std::{sync::Arc, task};
@@ -171,6 +171,8 @@ where
             }
         };
 
+        try_fut!(self.check_rate_limit());
+
         future::Either::Left(
             self.inner
                 .new_service((permit, self.target.clone()))
@@ -286,6 +288,21 @@ impl<T, N> HttpPolicyService<T, N> {
         self.metrics
             .route_not_found(labels, self.connection.dst, self.connection.tls.clone());
         HttpRouteNotFound(()).into()
+    }
+
+    fn check_rate_limit(&self) -> Result<()> {
+        let id = match self.connection.tls {
+            Conditional::Some(tls::ServerTls::Established {
+                client_id: Some(tls::ClientId(ref id)),
+                ..
+            }) => Some(id),
+            _ => None,
+        };
+        self.policy
+            .borrow()
+            .local_rate_limit
+            .check(id)
+            .map_err(Into::into)
     }
 }
 
--- a/linkerd/app/inbound/src/policy/http/tests.rs
+++ b/linkerd/app/inbound/src/policy/http/tests.rs
@@ -382,7 +382,7 @@ async fn rate_limit_allow() {
     let rmeta = Meta::new_default("default");
 
     // Rate-limit with plenty of room for two consecutive requests
-    let rl = LocalRateLimit::new_no_overrides(Some(10), Some(5));
+    let rl = LocalRateLimit::new_no_overrides_for_test(Some(10), Some(5));
 
     let authorizations = Arc::new([Authorization {
         meta: rmeta.clone(),
@@ -433,7 +433,7 @@ async fn rate_limit_deny() {
     let rmeta = Meta::new_default("default");
 
     // Rate-limit with room for only one request per second
-    let rl = LocalRateLimit::new_no_overrides(Some(10), Some(1));
+    let rl = LocalRateLimit::new_no_overrides_for_test(Some(10), Some(1));
 
     let authorizations = Arc::new([Authorization {
         meta: rmeta.clone(),
--- a/linkerd/app/inbound/src/policy/http/tests.rs
+++ b/linkerd/app/inbound/src/policy/http/tests.rs
@@ -1,7 +1,7 @@
 use super::*;
 use crate::policy::{Authentication, Authorization, Meta, Protocol, ServerPolicy};
 use linkerd_app_core::{svc::Service, Infallible};
-use linkerd_proxy_server_policy::LocalRateLimit;
+use linkerd_proxy_server_policy::{LocalRateLimit, RateLimitError};
 
 macro_rules! conn {
     ($client:expr, $dst:expr) => {{
@@ -376,11 +376,14 @@ async fn http_filter_inject_failure() {
 }
 
 #[tokio::test(flavor = "current_thread")]
-async fn rate_limit() {
+async fn rate_limit_allow() {
     use linkerd_app_core::{Ipv4Net, Ipv6Net};
 
     let rmeta = Meta::new_default("default");
+
+    // Rate-limit with plenty of room for two consecutive requests
     let rl = LocalRateLimit::new_no_overrides(Some(10), Some(5));
+
     let authorizations = Arc::new([Authorization {
         meta: rmeta.clone(),
         networks: vec![Ipv4Net::default().into(), Ipv6Net::default().into()],
@@ -392,6 +395,7 @@ async fn rate_limit() {
         rl
     );
 
+    // First request should be allowed
     let rsp = svc
         .call(
             ::http::Request::builder()
@@ -405,6 +409,74 @@ async fn rate_limit() {
         .get::<HttpRoutePermit>()
         .expect("permitted");
     assert_eq!(permit.labels.route.route, rmeta);
+
+    // Second request should be allowed as well
+    let rsp = svc
+        .call(
+            ::http::Request::builder()
+                .body(hyper::Body::default())
+                .unwrap(),
+        )
+        .await
+        .expect("serves");
+    let permit = rsp
+        .extensions()
+        .get::<HttpRoutePermit>()
+        .expect("permitted");
+    assert_eq!(permit.labels.route.route, rmeta);
+}
+
+#[tokio::test(flavor = "current_thread")]
+async fn rate_limit_deny() {
+    use linkerd_app_core::{Ipv4Net, Ipv6Net};
+
+    let rmeta = Meta::new_default("default");
+
+    // Rate-limit with room for only one request per second
+    let rl = LocalRateLimit::new_no_overrides(Some(10), Some(1));
+
+    let authorizations = Arc::new([Authorization {
+        meta: rmeta.clone(),
+        networks: vec![Ipv4Net::default().into(), Ipv6Net::default().into()],
+        authentication: Authentication::Unauthenticated,
+    }]);
+
+    let (mut svc, _tx) = new_svc!(
+        Protocol::Http1(Arc::new([http::default(authorizations.clone())])),
+        rl
+    );
+
+    // First request should be allowed
+    let rsp = svc
+        .call(
+            ::http::Request::builder()
+                .body(hyper::Body::default())
+                .unwrap(),
+        )
+        .await
+        .expect("serves");
+    let permit = rsp
+        .extensions()
+        .get::<HttpRoutePermit>()
+        .expect("permitted");
+    assert_eq!(permit.labels.route.route, rmeta);
+
+    // Second request should be denied
+    let rsp = svc
+        .call(
+            ::http::Request::builder()
+                .body(hyper::Body::default())
+                .unwrap(),
+        )
+        .await
+        .expect_err("should deny");
+    let err = rsp
+        .downcast_ref::<RateLimitError>()
+        .expect("rate limit error");
+    match err {
+        RateLimitError::PerIdentity(rps) => assert_eq!(rps, &std::num::NonZeroU32::new(1).unwrap()),
+        _ => panic!("unexpected error"),
+    };
 }
 
 #[tokio::test(flavor = "current_thread")]
--- a/linkerd/app/inbound/src/policy/http/tests.rs
+++ b/linkerd/app/inbound/src/policy/http/tests.rs
@@ -404,11 +404,7 @@ async fn rate_limit_allow() {
         )
         .await
         .expect("serves");
-    let permit = rsp
-        .extensions()
-        .get::<HttpRoutePermit>()
-        .expect("permitted");
-    assert_eq!(permit.labels.route.route, rmeta);
+    assert_eq!(rsp.status(), ::http::StatusCode::OK);
 
     // Second request should be allowed as well
     let rsp = svc
@@ -419,11 +415,7 @@ async fn rate_limit_allow() {
         )
         .await
         .expect("serves");
-    let permit = rsp
-        .extensions()
-        .get::<HttpRoutePermit>()
-        .expect("permitted");
-    assert_eq!(permit.labels.route.route, rmeta);
+    assert_eq!(rsp.status(), ::http::StatusCode::OK);
 }
 
 #[tokio::test(flavor = "current_thread")]
@@ -455,11 +447,7 @@ async fn rate_limit_deny() {
         )
         .await
         .expect("serves");
-    let permit = rsp
-        .extensions()
-        .get::<HttpRoutePermit>()
-        .expect("permitted");
-    assert_eq!(permit.labels.route.route, rmeta);
+    assert_eq!(rsp.status(), ::http::StatusCode::OK);
 
     // Second request should be denied
     let rsp = svc
--- a/linkerd/app/inbound/src/policy/http/tests.rs
+++ b/linkerd/app/inbound/src/policy/http/tests.rs
@@ -29,6 +29,7 @@ macro_rules! new_svc {
                     kind: "Server".into(),
                     name: "testsrv".into(),
                 }),
+                local_rate_limit: Arc::new(Default::default()),
             },
         );
         let svc = HttpPolicyService {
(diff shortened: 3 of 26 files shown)


linkerd/proxy/server-policy/Cargo.toml
```diff
@@ -10,6 +10,7 @@ publish = false
 proto = ["linkerd-http-route/proto", "linkerd2-proxy-api", "prost-types"]
 
 [dependencies]
+governor = { version = "0.6", default-features = false, features = ["std", "jitter", "quanta"] }
```
olix0r, line 13:
```suggestion
governor = { version = "0.6", default-features = false, features = ["std", "jitter"] }
```

We're not using the quanta clock so it's fine to omit it.

linkerd/proxy/server-policy/src/local_rate_limit.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub total: Option<RateLimiter<NotKeyed, InMemoryState, DefaultClock>>,
+    pub identity: Option<RateLimiter<String, DefaultKeyedStateStore<String>, DefaultClock>>,
+    pub overrides: Vec<HttpLocalRateLimitOverride>,
+}
+
+#[derive(Debug)]
+pub struct HttpLocalRateLimitOverride {
+    pub ids: Vec<String>,
+    pub rate_limit: RateLimiter<Vec<String>, DefaultKeyedStateStore<Vec<String>>, DefaultClock>,
+}
```
olix0r, line 19: * I think we can maintain a narrower API--no need to expose all of the governor types, we
* It would be preferable to use the Identity types directly here rather than require String round-tripping (allocations).
* I think the override rate limiter is misconfigured in your version--it should not be keyed on Vec<String>.
* Furthermore, it seems preferable to store the overrides for efficient lookup (i.e. in a HashMap).

```suggestion
#[derive(Debug, Default)]
pub struct LocalRateLimit {
    total: Option<RateLimit<InMemoryState>>,
    per_identity: Option<RateLimit<DefaultKeyedStateStore<Option<Id>>>>,
    overrides: HashMap<Id, Arc<RateLimit<InMemoryState>>>,
}

#[derive(Debug)]
struct RateLimit<S: StateStore> {
    rps: NonZeroU32,
    limiter: RateLimiter<S::Key, S, DefaultClock>,
}

#[derive(Debug, thiserror::Error, PartialEq, Eq)]
pub enum RateLimitError {
    #[error("total rate limit exceeded: {0}rps")]
    Total(NonZeroU32),
    #[error("per-identity rate limit exceeded: {0}rps")]
    PerIdentity(NonZeroU32),
    #[error("override rate limit exceeded: {0}rps")]
    Override(NonZeroU32),
}
```

```rust
impl LocalRateLimit {
    pub fn check(&self, id: Option<&Id>) -> Result<(), RateLimitError> {
        if let Some(lib) = &self.total {
            if lim.limiter.check().is_err() {
                return Err(RateLimitError::Total(lim.rps));
            }
        }

        if let Some(id) = id {
            if let Some(lim) = self.overrides.get(id) {
                if lim.limiter.check().is_err() {
                    return Err(RateLimitError::Override(lim.rps));
                }
                return Ok(());
            }
        }

        if let Some(lim) = &self.per_identity {
            if limiter.check_key(&lim.id.cloned()).is_err() {
                return Err(RateLimitError::PerIdentity(lim.rps));
            }
        }

        Ok(())
    }
}
```

The server-policy crate can hide all of the specifics about rate limiting and let callers have a very simple interface.

linkerd/proxy/server-policy/src/local_rate_limit.rs
```diff
@@ -0,0 +1,72 @@
...
+                            .into_iter()
+                            .flat_map(|cl| cl.identities.into_iter().map(|id| id.name))
+                            .collect();
+                        let quota =
+                            Quota::per_second(NonZeroU32::new(lim.requests_per_second).unwrap());
+                        let rate_limit = RateLimiter::keyed(quota);
+                        HttpLocalRateLimitOverride { ids, rate_limit }
+                    })
+                })
+                .collect();
```
olix0r, line 63: Note that we cannot call `unwrap` (or expect) when handling input from the control plane. We need to make sure the None-ness propagates to cause the item to be ignored:
```suggestion
         let total = proto.total.and_then(|lim| {
                let rps = NonZeroU32::new(lim.requests_per_second)?;
                let limiter = RateLimiter::direct(Quota::per_second(rps));
                Some(RateLimit { rps, limiter })
            });

            let per_identity = proto.identity.and_then(|lim| {
                let rps = NonZeroU32::new(lim.requests_per_second)?;
                let limiter = RateLimiter::hashmap(Quota::per_second(rps));
                Some(RateLimit { rps, limiter })
            });

            let overrides = proto
                .overrides
                .into_iter()
                .flat_map(|ovr| {
                    let Some(lim) = ovr.limit else {
                        return vec![];
                    };

                    let Some(rps) = NonZeroU32::new(lim.requests_per_second) else {
                        return vec![];
                    };
                    let limiter = RateLimiter::direct(Quota::per_second(rps));
                    let limit = Arc::new(RateLimit { rps, limiter });

                    ovr.clients
                        .into_iter()
                        .flat_map(|cl| {
                            cl.identities
                                .into_iter()
                                .filter_map(|id| id.name.parse::<Id>().ok())
                        })
                        .map(move |id| (id, limit.clone()))
                        .collect::<Vec<(Id, Arc<RateLimit<InMemoryState>>)>>()
                })
                .collect::<HashMap<Id, Arc<RateLimit<InMemoryState>>>>();

```

linkerd/proxy/server-policy/src/lib.rs
```diff
@@ -14,10 +16,11 @@ pub use self::{
 };
 pub use linkerd_http_route as route;
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug)]
 pub struct ServerPolicy {
     pub protocol: Protocol,
     pub meta: Arc<Meta>,
+    pub local_rate_limit: Arc<HttpLocalRateLimit>,
```
olix0r, line 23: This isn't a big problem, but it seems like it would be most consistent (with both the proxy prior art and API) for the rate limit to be encoded in the `Protocol`, since it's not relevant for certain protocol variants.

```rust
    Detect {
        http: Arc<[http::Route]>,
        http_local_rate_limit: Arc<LocalRateLimit>
        timeout: time::Duration,
        tcp_authorizations: Arc<[Authorization]>,
    },
    Http1(Arc<[http::Route]>, Arc<LocalRateLimit>),
    Http2(Arc<[http::Route]>, Arc<LocalRateLimit>),
```

linkerd/proxy/server-policy/src/local_rate_limit.rs
alpeb: > I think the override rate limiter is misconfigured in your version--it should not be keyed on Vec.

As per linkerd/linkerd2#13231, an `HTTPLocalRateLimitPolicy` would look something like this:
```yaml
apiVersion: policy.linkerd.io/v1alpha1
kind: HTTPLocalRateLimitPolicy
metadata:
  namespace: emojivoto
  name: web-rl
spec:
  targetRef:
    group: policy.linkerd.io
    kind: Server
    name: web-http
  total:
    requestsPerSecond: 100
  identity:
    requestsPerSecond: 20
  overrides:
  - requestsPerSecond: 10
    clientRefs:
    - kind: ServiceAccount
      namespace: emojivoto
      name: default
```
Under overrides, I implemented the `clientRefs` as an array, which explains the types I was hinting at (I may have made that an array _after_ our conversations, sorry about that). It's definitely simpler to have it be just one object though like in your suggestion, right?

linkerd/proxy/server-policy/src/lib.rs
alpeb: The problem is that governor's `RateLimiter` doesn't implement `PartialEq`, which `Protocol` requires. That requirement can be lifted from the types depending on `Protocol`, except for `ServerPermit`, which uses it for test assertions. So we could either manually implement `PartialEq` for `ServerPermit` just so that those tests aren't affected, or implement it for our `LocalRateLimit` wrapper you suggested above but it'd only check for equality on rps and ids. Any thoughts?

linkerd/proxy/server-policy/src/local_rate_limit.rs
olix0r: The override representation in the YAML doesn't dictate how we store them at runtime.

Also note that I interpret the overrides to support "pooled" limits. For example:

```yaml
  - requestsPerSecond: 10
    clientRefs:
    - kind: ServiceAccount
      namespace: emojivoto
      name: web
    clientRefs:
    - kind: ServiceAccount
      namespace: emojivoto
      name: default
```

In this situation I would expect these two accounts to _share_ a rate limit of 10RPS. This is distinct from the configuration in which they have distinct but identical limits:

```yaml
  - requestsPerSecond: 10
    clientRefs:
    - kind: ServiceAccount
      namespace: emojivoto
      name: web
  - requestsPerSecond: 10
    clientRefs:
    - kind: ServiceAccount
      namespace: emojivoto
      name: default
```

I imagine that we could build a map of overrides from the existing spec:

```rust

            let overrides = proto
                .overrides
                .into_iter()
                .flat_map(|ovr| {
                    let Some(lim) = ovr.limit else {
                        return vec![];
                    };

                    let Some(rps) = NonZeroU32::new(lim.requests_per_second) else {
                        return vec![];
                    };
                    let limiter = RateLimiter::direct(Quota::per_second(rps));
                    let limit = Arc::new(RateLimit { rps, limiter });

                    ovr.clients
                        .into_iter()
                        .flat_map(|cl| {
                            cl.identities
                                .into_iter()
                                .filter_map(|id| id.name.parse::<Id>().ok())
                        })
                        .map(move |id| (id, limit.clone()))
                        .collect::<Vec<(Id, Arc<RateLimit<InMemoryState>>)>>()
                })
                .collect::<HashMap<Id, Arc<RateLimit<InMemoryState>>>>();
```

This supports shared limits and efficient map-based lookups.

linkerd/proxy/server-policy/src/lib.rs
olix0r: After looking at your follow-on PR, this change probably isn't worth it -- it's more complicated to consume the rate limit without much benefit.

linkerd/proxy/server-policy/src/lib.rs
```diff
@@ -131,6 +135,28 @@ pub mod proto {
...
+                        local_rate_limit,
+                        ..
+                    })
+                    | api::proxy_protocol::Kind::Http2(api::proxy_protocol::Http2 {
+                        local_rate_limit,
+                        ..
+                    }) => local_rate_limit.unwrap_or_default().into(),
+                    _ => Default::default(),
+                }
+            };
```
olix0r, line 158: nit: We can make this slightly more compact by removing redundant braces. The other blocks that do this define temporary variables.

```suggestion
            let local_rate_limit = match protocol
                .clone()
                .and_then(|api::ProxyProtocol { kind }| kind)
                .ok_or(InvalidServer::MissingProxyProtocol)?
            {
                api::proxy_protocol::Kind::Detect(api::proxy_protocol::Detect {
                    http_local_rate_limit,
                    ..
                }) => http_local_rate_limit.unwrap_or_default().into(),
                api::proxy_protocol::Kind::Http1(api::proxy_protocol::Http1 {
                    local_rate_limit,
                    ..
                })
                | api::proxy_protocol::Kind::Http2(api::proxy_protocol::Http2 {
                    local_rate_limit,
                    ..
                }) => local_rate_limit.unwrap_or_default().into(),
                _ => Default::default(),
            };
```

linkerd/app/inbound/src/policy/http.rs
```diff
@@ -171,6 +171,8 @@ where
             }
         };
 
+        try_fut!(self.check_rate_limit());
```
olix0r, line 174: It would be good to exercise this in a [test](https://github.com/linkerd/linkerd2-proxy/blob/b580e657e565d77e4138c34397fea7182fd1797b/linkerd/app/inbound/src/policy/http/tests.rs).

linkerd/proxy/server-policy/src/local_rate_limit.rs
```diff
@@ -0,0 +1,105 @@
...
+    #[error("total rate limit exceeded: {0}rps")]
+    Total(NonZeroU32),
+    #[error("per-identity rate limit exceeded: {0}rps")]
+    PerIdentity(NonZeroU32),
+    #[error("override rate limit exceeded: {0}rps")]
+    Override(NonZeroU32),
+}
+
+impl LocalRateLimit {
+    pub fn check(&self, id: Option<&Id>) -> Result<(), RateLimitError> {
```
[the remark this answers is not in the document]
alpeb: Sounds good :+1:  I will be testing using governor's [FakeRelativeClock](https://docs.rs/governor/latest/governor/clock/struct.FakeRelativeClock.html) for this, given they don't rely on tokio.

linkerd/proxy/server-policy/src/local_rate_limit.rs
```diff
@@ -0,0 +1,173 @@
...
+    #[error("per-identity rate limit exceeded: {0}rps")]
+    PerIdentity(NonZeroU32),
+    #[error("override rate limit exceeded: {0}rps")]
+    Override(NonZeroU32),
+}
+
+// === impl LocalRateLimit ===
+
+impl LocalRateLimit {
+    pub fn new_no_overrides(
```
olix0r, line 48: We would typically [feature-gate this](https://github.com/linkerd/linkerd2-proxy/pull/3328) to assert that it's only used for tests.

linkerd/app/inbound/src/policy/http/tests.rs
```diff
@@ -363,6 +375,38 @@ async fn http_filter_inject_failure() {
...
+                .body(hyper::Body::default())
+                .unwrap(),
+        )
+        .await
+        .expect("serves");
+    let permit = rsp
+        .extensions()
+        .get::<HttpRoutePermit>()
+        .expect("permitted");
+    assert_eq!(permit.labels.route.route, rmeta);
```
olix0r, line 407: Is there some way that we could actually test actually fails when a rate limit is exceeded? E.g. Set of rate limit of 1 and send 2 requests.

linkerd/app/inbound/src/policy/http/tests.rs
```diff
@@ -1,6 +1,7 @@
 use super::*;
 use crate::policy::{Authentication, Authorization, Meta, Protocol, ServerPolicy};
 use linkerd_app_core::{svc::Service, Infallible};
+use linkerd_proxy_server_policy::local_rate_limit::LocalRateLimit;
```
olix0r, line 4: With a reexport we can simplify this:
```suggestion
use linkerd_proxy_server_policy::LocalRateLimit;
```

linkerd/app/inbound/src/http/server.rs
```diff
@@ -203,6 +203,12 @@ impl errors::HttpRescue<Error> for ServerRescue {
             ));
         }
 
+        if errors::is_caused_by::<linkerd_proxy_server_policy::local_rate_limit::RateLimitError>(
```
olix0r, line 206:
```suggestion
        if errors::is_caused_by::<linkerd_proxy_server_policy::RateLimitError>(
```

linkerd/app/inbound/src/policy/http.rs
```diff
@@ -287,6 +289,21 @@ impl<T, N> HttpPolicyService<T, N> {
             .route_not_found(labels, self.connection.dst, self.connection.tls.clone());
         HttpRouteNotFound(()).into()
     }
+
+    fn check_rate_limit(&self) -> Result<()> {
+        let id = match self.connection.tls {
+            Conditional::Some(tls::ServerTls::Established {
```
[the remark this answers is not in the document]
alpeb: That's actually not reexported in the tls module. Should I add it?

linkerd/app/inbound/src/policy/http/tests.rs
alpeb: Yep, just added in my latest commit :+1:

linkerd/app/inbound/src/policy/http.rs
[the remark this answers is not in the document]
olix0r: Ah I was thinking of:
```suggestion
            tls::ConditionalServerTls::Some(tls::ServerTls::Established {
```

I think it's fine to leave as-is with the Conditional import.

linkerd/app/inbound/src/policy/http/tests.rs
```diff
@@ -363,6 +375,110 @@ async fn http_filter_inject_failure() {
...
+                .body(hyper::Body::default())
+                .unwrap(),
+        )
+        .await
+        .expect("serves");
+    let permit = rsp
+        .extensions()
+        .get::<HttpRoutePermit>()
+        .expect("permitted");
+    assert_eq!(permit.labels.route.route, rmeta);
```
olix0r, line 411: We can omit the checks that are covered by other tests--they're not necessary to test the rate limiting behavior.

```suggestion
    assert_eq!(rsp.status(), ::http::StatusCode::OK);
```

linkerd/app/inbound/src/policy/http/tests.rs
```diff
@@ -363,6 +375,110 @@ async fn http_filter_inject_failure() {
...
```
olix0r, line 426:
```suggestion
    assert_eq!(rsp.status(), ::http::StatusCode::OK);
```

linkerd/app/inbound/src/policy/http/tests.rs
```diff
@@ -363,6 +375,110 @@ async fn http_filter_inject_failure() {
...
```
olix0r, line 462:
```suggestion
    assert_eq!(rsp.status(), ::http::StatusCode::OK);
```

linkerd/app/inbound/src/policy/http/tests.rs
```diff
@@ -363,6 +375,110 @@ async fn http_filter_inject_failure() {
...
+        )
+        .await
+        .expect_err("should deny");
+    let err = rsp
+        .downcast_ref::<RateLimitError>()
+        .expect("rate limit error");
+    match err {
+        RateLimitError::PerIdentity(rps) => assert_eq!(rps, &std::num::NonZeroU32::new(1).unwrap()),
+        _ => panic!("unexpected error"),
+    };
```
olix0r, line 479: nit: slightly more readable:

```suggestion
    match err {
        RateLimitError::PerIdentity(rps) => assert_eq!(rps.into(), 1),
        _ => panic!("unexpected error"),
    }
```

olix0r (approved):
Take note that we'll want to update the commit message to reflect the current state of this branch.

linkerd/app/inbound/src/policy/http/tests.rs
alpeb: Clippy doesn't like that :disappointed: 
```
error[E0283]: type annotations needed
   --> linkerd/app/inbound/src/policy/http/tests.rs:465:60
    |
465 |         RateLimitError::PerIdentity(rps) => assert_eq!(rps.into(), 1),
    |                                                            ^^^^
    |
    = note: cannot satisfy `_: std::convert::From<&std::num::NonZeroU32>`
    = note: required for `&std::num::NonZeroU32` to implement `std::convert::Into<_>`
help: try using a fully qualified path to specify the expected types
    |
465 |         RateLimitError::PerIdentity(rps) => assert_eq!(<&std::num::NonZeroU32 as std::convert::Into<T>>::into(rps), 1),
    |                                                        +++++++++++++++++++++++++++++++++++++++++++++++++++++++   ~
```

PR #3306, opened 2024-10-29
feat(outbound): Use the policy API for opaq routing

zaharidichev:
This work (based on the early [sketch seen here](https://github.com/linkerd/linkerd2-proxy/compare/ver/opaq-policy)) changes the `opaq` routing stack to use not only profiles routing but to consider client policy responses as well. The way this is done is that we use the `opaq:Routes` type to express the routing information our stack contains. This type is geared towards the client-policy API. Whenever we need to use the profiles response, we simply translate its contents to a client-policy one.

Notable changes in this PR are: 

- we introduce an `OpaqSidecar` type that holds `opaq::Routes` objects which are routes that have been constructed from either `profiles` or `policy` response
- during the construction of the `OpaqSidecar` we consider profiles responses only if there are extra targets in the response (in order to support `TrafficSplit` functionality). In any other case we prefer the policy response.


linkerd/app/outbound/Cargo.toml
```diff
@@ -45,6 +45,7 @@ linkerd-proxy-client-policy = { path = "../../proxy/client-policy", features = [
 ] }
 linkerd-retry = { path = "../../retry" }
 linkerd-tls-route = { path = "../../tls/route" }
+linkerd-opaq-route = { path = "../../opaq-route" }
```
olix0r, line 48: nit: move above linkerd-proxy-client-policy (sorting)

linkerd/opaq-route/src/lib.rs
```diff
@@ -0,0 +1,49 @@
+//! An TLS route matching library for Linkerd to support the TLSRoute
```
olix0r, line 1:
```suggestion
//! A TCP route matching library for Linkerd to support the TCPRoute
```

linkerd/opaq-route/src/match.rs
```diff
@@ -0,0 +1,21 @@
+use std::cmp::Ordering;
+
+/// Matches TCP sessions. For now, this is a placeholder
+#[derive(Clone, Debug, Default, Hash, PartialEq, Eq)]
+pub struct MatchSession(());
```
olix0r, line 5: I suspect this is coming from the TLSRoute type--do we expect TCP routes to use "session" terminology?

linkerd/app/outbound/src/opaq/concrete.rs
```diff
@@ -23,18 +23,26 @@ use linkerd_app_core::{
...
+    Addr, Error, Infallible, NameAddr,
 };
 use std::{fmt::Debug, net::SocketAddr, sync::Arc};
 use tracing::info_span;
 
 /// Parameter configuring dispatcher behavior.
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub enum Dispatch {
-    Balance(NameAddr, NameAddr, balance::EwmaConfig),
+    Balance(Addr, NameAddr, balance::EwmaConfig),
```
olix0r, line 34: I'm curious about this change and its relation to LogicalAddr. Generally speaking, we've used the `profiles::LogicalAddr` type to indicate that a profile has given us back a named address for the service we're talking to; and we furthermore have some guarantees that balancers are always using named addresses.

This change uses a new LogicalAddr to associate with _all_ traffic. The only real use for this I see is that we may use this as part of the endpoint metrics labels, which ignores the non-named variant. Given all of this, I'm curious to what extent we actually want a new LogicalAddr type--it seems like having a `Param<Option<profiles::LogicalAddr>>` is actually what we want -- we only want to set this label value with the legacy logical address returned from profile updates. For any new metrics, we'll want ParentRef and BackendRef labels.

linkerd/opaq-route/src/match.rs
zaharidichev: Should we think of a better name. `MatchTCPConnection` ? Or try to get rid of this whole machinery here. And yes, this is coming from the TLSRoute type.

linkerd/app/outbound/src/opaq/concrete.rs
zaharidichev: @olix0r So based on this comment, I went and looked over all his again. I have put a PR with some proposed changes along what you have suggested: https://github.com/linkerd/linkerd2-proxy/pull/3309

Namely: 

- removed `opaq::LogicalAddr`, we only use the `Addr` type now in routes, etc
- added the ability for the target to give you an `Option<profiles::LogicalAddr> `that is Some only if we are using profiles
- this should allow us to preserve the semantics of the balancer metrics with respect to the concrete/logical labels that are different if we are using traffic splits.

I even think that the addr types on the `opaq::Routes` type should not be an `Addr` but some variation of a `SocketAddr`. After all, could it ever be a `NamedAddr` in there?

linkerd/proxy/client-policy/src/lib.rs
```diff
@@ -206,7 +206,7 @@ impl ClientPolicy {
                 opaque: opaq::Opaque {
                     // TODO(eliza): eventually, can we configure the opaque
                     // policy to fail conns?
-                    policy: None,
+                    routes: std::sync::Arc::new([]),
```
olix0r, line 209: We can remove this TODO now, right? The empty policy will fail connections, right?

We probably ought to use a static NO_OPAQ_BACKENDS to be consistent (prevents needless heap allocations).

linkerd/proxy/client-policy/src/lib.rs
```diff
@@ -175,7 +175,7 @@ impl ClientPolicy {
...
                 },
```
olix0r, line 179: Similarly, this TODO is addressed, and we can use a static routes. This does suggest that we may want to follow up with a filter that fails connections with an error, as we do for HTTP. This may also serve as a mechanism for propagating denials.

linkerd/app/outbound/src/opaq/logical.rs
```diff
@@ -87,229 +62,48 @@ impl<N> Outbound<N> {
...
-                    },
-                    concrete.into_inner(),
-                )
+                .push_on_service(svc::layer::mk(move |concrete: N| {
+                    svc::stack(concrete.clone())
+                        .push(router::Router::layer())
+                        .push(svc::NewMapErr::layer_from_target::<LogicalError, _>())
+                        .arc_new_clone_tcp()
+                        .into_inner()
+                }))
```
olix0r, line 74: Can simplify this:

```suggestion
                .push_on_service(router::Router::layer())
                .push_on_service(svc::NewMapErr::layer_from_target::<LogicalError, _>())
                // Rebuild the inner router stack every time the watch changes.
                .push(svc::NewSpawnWatch::<Routes, _>::layer_into::<
                    router::Router<T>,
                >())
                .arc_new_clone_tcp()
```

linkerd/app/outbound/src/opaq/concrete.rs
```diff
@@ -87,11 +94,27 @@ impl prom::encoding::EncodeLabelSet for ConcreteLabels {
...
+            Some(profiles::LogicalAddr(addr)) => addr.to_string(),
+            None => bal.addr.to_string(),
+        };
+
         self.metrics(&ConcreteLabels {
-            logical: bal.logical.to_string().into(),
-            concrete: bal.concrete.to_string().into(),
+            logical: logical.into(),
+            concrete: bal.addr.to_string().into(),
         })
```
olix0r, line 118: I started a branch but haven't got it sensible yet, but we also need to consider the case where a balancer is provided by policy -- in that case we want to include parent and backend references (as we do in HTTP). The logical and concrete values cannot be removed due to backwards compatibility, but we should provide reference-oriented labels in the policy case to be consistent.

linkerd/opaq-route/src/match.rs
zaharidichev: This makes sense indeed. Will merge and update control plane to return only one route always.

linkerd/app/outbound/src/opaq/concrete.rs
zaharidichev: @ver This makes total sense. I have put together a branch that does that. Will submit a PR agaisnt this one. So in my proposal, for the two cases we balance we get the following results. I do not see a point to special casing profiles. We can still populate all this data based on the logical addrs that we get.


#### The case when we use the **policy response** when we target a Service
```
outbound_tcp_balancer_queue_requests_total{
    parent_group="core"
    parent_kind="Service"
    parent_namespace="mixed-env"
    parent_name="legacy-app-cluster"
    parent_port="80"
    parent_section_name=""
    backend_group="core"
    backend_kind="Service"
    backend_namespace="mixed-env"
    backend_name="legacy-app-cluster"
    backend_port="80"
    backend_section_name=""
    logical="legacy-app-cluster.mixed-env.svc.cluster.local:80"
    concrete="legacy-app-cluster.mixed-env.svc.cluster.local:80"
} 1
```

#### The case when we use the **profiles** response because a `TrafficSplit` is configured
```
outbound_tcp_balancer_queue_requests_total{
    parent_group="core"
    parent_kind="Service"
    parent_namespace="trafficsplit-sample"
    parent_name="backend-svc"
    parent_port="8080"
    parent_section_name=""
    backend_group="core"
    backend_kind="Service"
    backend_namespace="trafficsplit-sample"
    backend_name="failing-svc"
    backend_port="8080"
    backend_section_name=""
    logical="backend-svc.trafficsplit-sample.svc.cluster.local:8080"
    concrete="failing-svc.trafficsplit-sample.svc.cluster.local:8080"
} 2
```

linkerd/app/gateway/src/opaq.rs
```diff
@@ -3,8 +3,15 @@ use inbound::{GatewayAddr, GatewayDomainInvalid};
 use linkerd_app_core::{io, profiles, svc, tls, transport::addrs::*, Error};
 use linkerd_app_inbound as inbound;
 use linkerd_app_outbound as outbound;
+use tokio::sync::watch;
 
-pub type Target = outbound::opaq::Logical;
+#[derive(Clone, Debug)]
+pub struct Target {
+    orig_dst: OrigDstAddr,
```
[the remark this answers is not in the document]
zaharidichev: Thanks a lot. I made it compile and removed the profiles_logical from the gateway opaq target. Turned it into in a PR: https://github.com/linkerd/linkerd2-proxy/pull/3338

linkerd/app/outbound/src/opaq.rs
```diff
@@ -57,26 +58,20 @@ impl<C> Outbound<C> {
                 stk.push_new_idle_cached(config.discovery_idle_timeout)
                     // Use a dedicated target type to configure parameters for
                     // the opaque stack. It also helps narrow the cache key.
-                    .push_map_target(|t: T| Opaq(t.param()))
+                    .push_map_target(|parent: T| Opaq(parent))
```
olix0r, line 61:
```suggestion
                    .push_map_target(Opaq)
```

olix0r (approved):

PR #3308, opened 2024-10-30
feat(app): Backend frame count metrics

cratelyn:
### :cloud: overview

this introduces a new tower middleware for Prometheus metrics, used for instrumenting HTTP and gRPC
response bodies, and observing (a) the number of frames yielded by a body, and (b) the number of bytes
included in body frames.

this middleware allows operators to reason about how large or small the packets being served in a backend's
response bodies are.

a route-level middleware that instruments request bodies will be added in a follow-on PR.

 ### 📝 changes

an overview of changes made here:

* the `linkerd-http-prom` has a new `body_data` submodule. it exposes `request` and `response` halves, to
  be explicit about which body is being instrumented on a `tower::Service`.

* the `linkerd-http-prom` crate now has a collection of new dependencies: `bytes` is added as a dependency
  in order to inspect the data chunk when the inner body yields a new frame. `futures-util` and `http-body`
  are added as dev-dependencies for the accompanying test coverage.

* body metrics are affixed to the `RouteBackendMetrics<L>` structure, and registered at startup.

### :link: related

* #3334
* #3337
* #3318

### DIFF
--- a/linkerd/app/outbound/src/http/logical/policy/route/backend/metrics.rs
+++ b/linkerd/app/outbound/src/http/logical/policy/route/backend/metrics.rs
@@ -110,7 +110,7 @@ impl<L: StreamLabel> RouteBackendMetrics<L> {
         &self,
         l: &labels::RouteBackend,
     ) -> linkerd_http_prom::body_data::response::BodyDataMetrics {
-        self.body_metrics.get(l)
+        self.body_metrics.metrics(l)
     }
 }
 
@@ -156,6 +156,6 @@ where
         let Self(families) = self;
         let labels = labels::RouteBackend(t.param(), t.param(), t.param());
 
-        families.get(&labels)
+        families.metrics(&labels)
     }
 }
--- a/linkerd/app/outbound/src/http/logical/policy/route/backend/metrics.rs
+++ b/linkerd/app/outbound/src/http/logical/policy/route/backend/metrics.rs
@@ -1,6 +1,7 @@
 use crate::{BackendRef, ParentRef, RouteRef};
 use linkerd_app_core::{metrics::prom, svc};
 use linkerd_http_prom::{
+    body_data::response::{BodyDataMetrics, NewRecordBodyData, ResponseBodyFamilies},
     record_response::{self, NewResponseDuration, StreamLabel},
     NewCountRequests, RequestCount, RequestCountFamilies,
 };
@@ -15,6 +16,7 @@ mod tests;
 pub struct RouteBackendMetrics<L: StreamLabel> {
     requests: RequestCountFamilies<labels::RouteBackend>,
     responses: ResponseMetrics<L>,
+    body_metrics: ResponseBodyFamilies<labels::RouteBackend>,
 }
 
 type ResponseMetrics<L> = record_response::ResponseMetrics<
@@ -26,14 +28,24 @@ pub fn layer<T, N>(
     metrics: &RouteBackendMetrics<T::StreamLabel>,
 ) -> impl svc::Layer<
     N,
-    Service = NewCountRequests<
-        ExtractRequestCount,
-        NewResponseDuration<T, ExtractRecordDurationParams<ResponseMetrics<T::StreamLabel>>, N>,
+    Service = NewRecordBodyData<
+        ExtractRecordBodyDataParams,
+        NewCountRequests<
+            ExtractRequestCount,
+            NewResponseDuration<T, ExtractRecordDurationParams<ResponseMetrics<T::StreamLabel>>, N>,
+        >,
     >,
 > + Clone
 where
     T: MkStreamLabel,
     N: svc::NewService<T>,
+    NewRecordBodyData<
+        ExtractRecordBodyDataParams,
+        NewCountRequests<
+            ExtractRequestCount,
+            NewResponseDuration<T, ExtractRecordDurationParams<ResponseMetrics<T::StreamLabel>>, N>,
+        >,
+    >: svc::NewService<T>,
     NewCountRequests<
         ExtractRequestCount,
         NewResponseDuration<T, ExtractRecordDurationParams<ResponseMetrics<T::StreamLabel>>, N>,
@@ -44,12 +56,16 @@ where
     let RouteBackendMetrics {
         requests,
         responses,
+        body_metrics,
     } = metrics.clone();
+
     svc::layer::mk(move |inner| {
         use svc::Layer;
-        NewCountRequests::layer_via(ExtractRequestCount(requests.clone())).layer(
-            NewRecordDuration::layer_via(ExtractRecordDurationParams(responses.clone()))
-                .layer(inner),
+        NewRecordBodyData::layer_via(ExtractRecordBodyDataParams(body_metrics.clone())).layer(
+            NewCountRequests::layer_via(ExtractRequestCount(requests.clone())).layer(
+                NewRecordDuration::layer_via(ExtractRecordDurationParams(responses.clone()))
+                    .layer(inner),
+            ),
         )
     })
 }
@@ -57,15 +73,20 @@ where
 #[derive(Clone, Debug)]
 pub struct ExtractRequestCount(RequestCountFamilies<labels::RouteBackend>);
 
+#[derive(Clone, Debug)]
+pub struct ExtractRecordBodyDataParams(ResponseBodyFamilies<labels::RouteBackend>);
+
 // === impl RouteBackendMetrics ===
 
 impl<L: StreamLabel> RouteBackendMetrics<L> {
     pub fn register(reg: &mut prom::Registry, histo: impl IntoIterator<Item = f64>) -> Self {
         let requests = RequestCountFamilies::register(reg);
         let responses = record_response::ResponseMetrics::register(reg, histo);
+        let body_metrics = ResponseBodyFamilies::register(reg);
         Self {
             requests,
             responses,
+            body_metrics,
         }
     }
 
@@ -83,6 +104,14 @@ impl<L: StreamLabel> RouteBackendMetrics<L> {
     pub(crate) fn get_statuses(&self, l: &L::StatusLabels) -> prom::Counter {
         self.responses.get_statuses(l)
     }
+
+    #[cfg(test)]
+    pub(crate) fn get_response_body_metrics(
+        &self,
+        l: &labels::RouteBackend,
+    ) -> linkerd_http_prom::body_data::response::BodyDataMetrics {
+        self.body_metrics.get(l)
+    }
 }
 
 impl<L: StreamLabel> Default for RouteBackendMetrics<L> {
@@ -90,6 +119,7 @@ impl<L: StreamLabel> Default for RouteBackendMetrics<L> {
         Self {
             requests: Default::default(),
             responses: Default::default(),
+            body_metrics: Default::default(),
         }
     }
 }
@@ -99,6 +129,7 @@ impl<L: StreamLabel> Clone for RouteBackendMetrics<L> {
         Self {
             requests: self.requests.clone(),
             responses: self.responses.clone(),
+            body_metrics: self.body_metrics.clone(),
         }
     }
 }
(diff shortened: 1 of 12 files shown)


linkerd/http/prom/src/body_data/metrics.rs
```diff
@@ -0,0 +1,94 @@
...
+        + Sync
+        + 'static,
+{
+    const RESP_BODY_FRAMES_TOTAL_NAME: &'static str = "resp_body_frames_total";
+    const RESP_BODY_FRAMES_TOTAL_HELP: &'static str =
+        "Counts the number of frames in response bodies.";
+
+    const RESP_BODY_FRAMES_BYTES_NAME: &'static str = "resp_body_frames_bytes";
+    const RESP_BODY_FRAMES_BYTES_HELP: &'static str =
+        "Counts the total number of bytes in response bodies.";
```
cratelyn, line 54: the name and help strings for the new metrics added here.

linkerd/app/outbound/src/http/logical/policy/route/backend/metrics.rs
```diff
@@ -114,3 +145,17 @@ where
...
+where
+    T: svc::Param<ParentRef> + svc::Param<RouteRef> + svc::Param<BackendRef>,
+{
+    fn extract_param(&self, t: &T) -> BodyDataMetrics {
+        let Self(families) = self;
+        let labels = labels::RouteBackend(t.param(), t.param(), t.param());
+
+        families.get(&labels)
+    }
+}
```
cratelyn, line 161: here's how we extract the backend's `BodyDataMetrics` for a target type; we acquire the parent, route, and backend refs and use that label set to look up the corresponding time series from each family.

linkerd/app/outbound/src/http/logical/policy/route/backend/metrics.rs
```diff
@@ -26,14 +28,24 @@ pub fn layer<T, N>(
...
     N,
-    Service = NewCountRequests<
-        ExtractRequestCount,
-        NewResponseDuration<T, ExtractRecordDurationParams<ResponseMetrics<T::StreamLabel>>, N>,
+    Service = NewRecordBodyData<
+        ExtractRecordBodyDataParams,
+        NewCountRequests<
+            ExtractRequestCount,
+            NewResponseDuration<T, ExtractRecordDurationParams<ResponseMetrics<T::StreamLabel>>, N>,
+        >,
```
cratelyn, line 36: the additions to this function are the part of this change that spark the least joy.

this strikes me as one of the primary sharp edges we can consider sanding down in future refactoring work. my hunch is that we might be able to restructure this in a way where we can use `impl Service` syntax at function boundaries, and spare ourselves the need to name all of our successive layers (_and their accompanying parameter extractors and `NewService<T>` bounds_).

cratelyn (commented):
:bread: breadcrumbs for review...

linkerd/http/prom/src/body_data/response.rs
```diff
@@ -0,0 +1,105 @@
...
+    RespB: Body + Send + 'static,
+    RespB::Data: Send + 'static,
+    RespB::Error: Into<Error>,
+{
+    type Response = Response<BoxBody>;
+    type Error = S::Error;
+    type Future = futures::future::MapOk<
+        S::Future,
+        Box<dyn FnOnce(Response<RespB>) -> Self::Response + Send + 'static>,
+    >;
```
olix0r, line 75: nit: It's probably better to Box the whole future than the Map fn... it's the same number of allocations, but minimizes the type so that the `S` and `Self` need not be repeated.

linkerd/http/prom/src/body_data/response.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    fn poll_ready(
```
olix0r, line 77: Not sure if this is strictly necessary but we tend to mark passthroughs like this as inline.

```suggestion
    #[inline]
    fn poll_ready(
```

cratelyn: good catch! i'll file that away as a general practice to follow as well. b4e2f82 adds this `#[inline]` attribute.

in this spirit i went a little further and performed the same change for the `Body` middleware. doing this for `poll_trailers(..)` felt moderately spicy so i checked the generated code for pin projection [here](https://github.com/taiki-e/pin-project/blob/f1aa3b97295659ab82f83fcc54d975f79f10e886/examples/struct-default-expanded.rs#L56-L66).

linkerd/http/prom/src/body_data/response.rs
cratelyn: :package: cd2d1b9 boxes this future up!

cratelyn:
this pr has been rebased onto `main` as of release `v2.263.1`.

linkerd/http/prom/src/body_data/metrics.rs
```diff
@@ -0,0 +1,94 @@
...
+    resp_body_frames_bytes: Family<L, Counter>,
+}
+
+/// Counters to instrument a request or response body.
+#[derive(Clone, Debug, Default)]
+pub struct BodyDataMetrics {
+    /// Counts the number of request body frames.
+    pub frames_total: Counter,
+    /// Counts the total number of bytes in request body frames.
+    pub frames_bytes: Counter,
```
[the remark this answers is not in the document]
cratelyn: unfortunately these need to be public, because our test code lives in `linkerd-app-outbound`. applying this change, the tests are no longer able to compile.

```
/workspaces/linkerd2-proxy$ cargo check --target-dir=target-devcontainer/ --message-format=short --workspace --tests
linkerd/app/outbound/src/http/logical/policy/route/backend/metrics/tests.rs:158:9: error[E0451]: field `frames_total` of struct `BodyDataMetrics` is private
linkerd/app/outbound/src/http/logical/policy/route/backend/metrics/tests.rs:159:9: error[E0451]: field `frames_bytes` of struct `BodyDataMetrics` is private
```

the offending snippet is...

```rust
    // Acquire the counters for this backend.
    tracing::info!("acquiring response body metrics");
    let labels = labels::RouteBackend(parent_ref.clone(), route_ref.clone(), backend_ref.clone());
    let BodyDataMetrics {
        frames_total,
        frames_bytes,
    } = metrics.get_response_body_metrics(&labels);
```

i started on a patch to this via some `#[cfg(test)]` accessors, but realized it's nullified if we're going to be moving to a `Histogram` (_which doesn't provide a way to read its buckets_).

i'll keep this open until i've finished that, but noting why these are `pub` in the meantime.

linkerd/http/prom/src/body_data/response.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn instrument_response<B>(resp: Response<B>, metrics: BodyDataMetrics) -> Response<BoxBody>
+    where
+        B: Body + Send + 'static,
+        B::Data: Send + 'static,
+        B::Error: Into<Error>,
+    {
+        resp.map(|b| super::body::Body::new(b, metrics))
+            .map(BoxBody::new)
+    }
+}
```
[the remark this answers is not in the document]
cratelyn: 76f1f6dac148d08cec24859371f95714aed9750f

linkerd/http/prom/src/body_data/metrics.rs
```diff
@@ -0,0 +1,94 @@
...
+        );
+
+        Self {
+            resp_body_frames_total,
+            resp_body_frames_bytes,
+        }
+    }
+
+    /// Returns the [`BodyDataMetrics`] for the given label set.
+    pub fn get(&self, labels: &L) -> BodyDataMetrics {
```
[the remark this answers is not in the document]
cratelyn: 7731f966cad5198817edd8f7bd458eca06e9951d

linkerd/http/prom/src/body_data/metrics.rs
```diff
@@ -0,0 +1,94 @@
...
+
+use linkerd_metrics::prom::{self, Counter, Family, Registry};
+
+/// Counters for response body frames.
+#[derive(Clone, Debug)]
+pub struct ResponseBodyFamilies<L> {
+    /// Counts the number of response body frames.
+    resp_body_frames_total: Family<L, Counter>,
+    /// Counts the total number of bytes in response body frames.
+    resp_body_frames_bytes: Family<L, Counter>,
```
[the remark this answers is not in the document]
cratelyn: i've opened up an upstream PR against the `prometheus-client` library, prometheus/client_rust#242.

the goal of this is to provide `count()` and `sum()` methods that we can use to preserve the test coverage included in [linkerd/app/outbound/src/http/logical/policy/route/backend/metrics/tests.rs](https://github.com/linkerd/linkerd2-proxy/pull/3308/files#diff-5bbe228dadb053930239bddb79e24ca61324b28ae42fd6eed0bee9c6a22285f5R179-R238), and confirm that response body frames are being recorded as expected.

linkerd/http/prom/src/body_data/metrics.rs
cratelyn: i'm lightly in favor of them, because they explicitly declare to readers what the name and help strings are for the metrics emitted by this middleware. while they are not _mechanically_ required, i find this pattern to be friendly and helpful for readers trying to reason about the metrics provided by a piece of middleware, particularly if they are not Rust experts themselves.

while i have a preference for `const`s like this however, i agree with the desire for a convergent style if that outweighs the benefit, and am happy to remove them. what do you think?

linkerd/http/prom/src/body_data/metrics.rs
```diff
@@ -0,0 +1,94 @@
...
+    L: prom::encoding::EncodeLabelSet
+        + std::fmt::Debug
+        + std::hash::Hash
+        + Eq
+        + Clone
+        + Send
+        + Sync
+        + 'static,
+{
+    const RESP_BODY_FRAMES_TOTAL_NAME: &'static str = "resp_body_frames_total";
```
olix0r, line 48: We shouldn't introduce a new metric naming convetion (`resp_`). Existing metrics use the full `response_`.

olix0r (commented):
I through together a quick example of what a histogram implementation here looks like: https://github.com/linkerd/linkerd2-proxy/compare/kate/record-body-data...ver/kate/record-body-data (It compiles, though I didn't check the test status).

This replaces the counter/total metrics pair: the frame_size counters give us the total number of frames processed, and the sum gives us the total number of bytes. Getting in the proper metrics in a first pass saves us a deprecation cycle.

I chose a fairly constrained set of buckets to give us orders of magnitude: 128, 1024, 10240 (and we'll get an +Inf).

I also tried to handle the wrapping case a little more gracefully.

cratelyn:
:memo: **NB:** c874231 was lacking a DCO signoff, so i've added one on your behalf to satiate CI, @olix0r. that's now 225674b.

see; https://github.com/linkerd/linkerd2-proxy/runs/33282345859

linkerd/http/prom/src/body_data/metrics.rs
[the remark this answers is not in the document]
cratelyn: addressed in 225674b.

linkerd/http/prom/src/body_data/metrics.rs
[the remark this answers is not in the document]
cratelyn: on further inspection, `#[cfg(test)]` accessors are not a viable route, because that flag does not propagate to dependencies when running unit tests. unfortunate, because that's a pattern i quite like!

~~these fields~~ this field (_now a single histogram, via 225674b_) will need to remain public to account for test code.

cratelyn:
this pr has been rebased onto `main` as of release `v2.666.0`.

cratelyn:
> I through together a quick example of what a histogram implementation here looks like: [kate/record-body-data...ver/kate/record-body-data](https://github.com/linkerd/linkerd2-proxy/compare/kate/record-body-data...ver/kate/record-body-data) (It compiles, though I didn't check the test status).
> 
> This replaces the counter/total metrics pair: the frame_size counters give us the total number of frames processed, and the sum gives us the total number of bytes. Getting in the proper metrics in a first pass saves us a deprecation cycle.
> 
> I chose a fairly constrained set of buckets to give us orders of magnitude: 128, 1024, 10240 (and we'll get an +Inf).
> 
> I also tried to handle the wrapping case a little more gracefully.

changes from this branch have been incorporated here. i've added a DCO signoff, [here](https://github.com/linkerd/linkerd2-proxy/pull/3308#issuecomment-2489378294).

olix0r (approved):
This looks great!

linkerd/app/outbound/src/http/logical/policy/route/backend/metrics.rs
cratelyn: weirdly, ci failed after i applied those changes here. not going to investigate that further rn, reverted.

PR #3309, opened 2024-10-31
remove opaq logical addr

zaharidichev:
This change has the following effects: 
- we remove opaq::LogicalAddr
- we try to preserve the behavior of the endpoint metrics by using the profiles logical addr if we have used profiles for resolution.


olix0r (approved):
Yep, this looks good to me.

PR #3318, opened 2024-11-04
refactor(app): Simplify backend metrics tower layer

cratelyn:
**nb:** this is based on #3308. this is a refactor branch aimed at finding a way to simplify the somewhat burdensome types in the metrics middleware's `layer(..)` function.

#### refactor(app): Define layer type aliases
    
this isn't a "simplification" so much as an outlining of the chunky types in this function signature.

that said... it's subjectively nice to outline the definition of our layers into a `MetricsLayer<T, N>`
that we can use as shorthand in the function signature of `layer(..)`.
    
#### refactor(app): Simplify backend metrics layer bounds
    
this commit changes the backend-level metrics layer function, so that it invokes
`linkerd_app_core::svc::layers()` rather than `linkerd_app_core::svc::layer::mk()`.

this has a nice effect in that it simplifies the bounds of the `layer` function.

### DIFF
--- a/linkerd/app/outbound/src/http/logical/policy/route/backend/metrics.rs
+++ b/linkerd/app/outbound/src/http/logical/policy/route/backend/metrics.rs
@@ -38,20 +38,6 @@ pub fn layer<T, N>(
 > + Clone
 where
     T: MkStreamLabel,
-    N: svc::NewService<T>,
-    NewRecordBodyData<
-        ExtractRecordBodyDataParams,
-        NewCountRequests<
-            ExtractRequestCount,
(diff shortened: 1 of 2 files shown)


cratelyn:
this pr has been rebased onto `main` as of release `v2.263.1`.

olix0r (approved):

cratelyn:
for the record, this change seemed to cause ci failures in #3308 and #3334. i'm not fully sure why, and am refraining from investigating further. just noting here for visibility and posterity, that `mk()` and `layers()` don't seem to be completely equivalent. i'm unsure why that failure was observed in those pr's, but not this one. :raised_eyebrow: strange! i'm avoiding falling into the nerdsnipe :hole: :rabbit2:

PR #3319, opened 2024-11-04
Allow duplicate weighted backend keys

sfleen:
Currently, if the proxy receives two backends with the same metadata, one of the backends will get dropped because the backend metadata is used as the key in a hash map. Attempting to then randomly distribute requests to the backends can panic when selecting the now non-existent backend.

This is fixed by no longer using backend metadata as a hash map key, instead generating separate IDs that are stored in a vec to retain the declared order of backends while also being used to look up the backend and associated weight independetly.

Validated with new unit tests excercising duplicate backend keys, as well as a few around the invariants use to store the backends.

### DIFF
--- a/linkerd/distribute/src/keys.rs
+++ b/linkerd/distribute/src/keys.rs
@@ -1,21 +1,16 @@
+use ahash::{HashMap, HashMapExt};
 use rand::{
     distributions::{WeightedError, WeightedIndex},
     prelude::Distribution as _,
     Rng,
 };
-use std::{collections::HashMap, hash::Hash};
+use std::hash::Hash;
 
 /// Uniquely identifies a key/backend pair for a distribution. This allows
 /// backends to have the same key and still participate in request distribution.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct KeyId {
+pub(crate) struct KeyId {
     idx: usize,
-}
-
-impl KeyId {
-    pub(crate) fn new(idx: usize) -> Self {
-        Self { idx }
-    }
 }
 
 #[derive(Debug)]
@@ -35,6 +30,14 @@ pub struct WeightedKey<K> {
 pub(crate) struct WeightedKeySelector<'a, K> {
     keys: &'a WeightedServiceKeys<K>,
     index: WeightedIndex<u32>,
+}
+
+// === impl KeyId ===
+
+impl KeyId {
+    pub(crate) fn new(idx: usize) -> Self {
+        Self { idx }
+    }
 }
 
 // === impl UnweightedKeys ===
--- a/linkerd/distribute/src/keys.rs
+++ b/linkerd/distribute/src/keys.rs
@@ -1,0 +1,160 @@
+use rand::{
+    distributions::{WeightedError, WeightedIndex},
+    prelude::Distribution as _,
+    Rng,
+};
+use std::{collections::HashMap, hash::Hash};
+
+/// Uniquely identifies a key/backend pair for a distribution. This allows
+/// backends to have the same key and still participate in request distribution.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct KeyId {
+    idx: usize,
+}
+
+impl KeyId {
+    pub(crate) fn new(idx: usize) -> Self {
+        Self { idx }
+    }
+}
+
+#[derive(Debug)]
+pub struct ServiceKeys<K> {
+    ids: Vec<KeyId>,
+    keys: HashMap<KeyId, K>,
+}
+
+pub type WeightedServiceKeys<K> = ServiceKeys<WeightedKey<K>>;
+
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub struct WeightedKey<K> {
+    pub key: K,
+    pub weight: u32,
+}
+
+pub(crate) struct WeightedKeySelector<'a, K> {
+    keys: &'a WeightedServiceKeys<K>,
+    index: WeightedIndex<u32>,
+}
+
+// === impl UnweightedKeys ===
+
+// PartialEq, Eq, and Hash are all valid to implement for UnweightedKeys since
+// there is a defined iteration order for the keys, but it cannot be automatically
+// derived for HashMap fields.
+impl<K: PartialEq> PartialEq for ServiceKeys<K> {
+    fn eq(&self, other: &Self) -> bool {
+        if self.ids != other.ids {
+            return false;
+        }
+
+        for id in &self.ids {
+            if self.keys.get(id) != other.keys.get(id) {
+                return false;
+            }
+        }
+
+        true
+    }
+}
+
+impl<K: Eq> Eq for ServiceKeys<K> {}
+
+impl<K: Hash> Hash for ServiceKeys<K> {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.ids.hash(state);
+        // Normally we would also hash the length, but self.ids and
+        // self.keys have the same length
+        for id in &self.ids {
+            self.keys.get(id).hash(state);
+        }
+    }
+}
+
+impl<K> ServiceKeys<K> {
+    pub(crate) fn new(iter: impl Iterator<Item = K>) -> Self {
+        let mut ids = Vec::new();
+        let mut keys = HashMap::new();
+        for (idx, key) in iter.enumerate() {
+            let id = KeyId::new(idx);
+            ids.push(id);
+            keys.insert(id, key);
+        }
+
+        Self { ids, keys }
+    }
+
+    pub(crate) fn is_empty(&self) -> bool {
+        self.ids.is_empty()
+    }
+
+    pub(crate) fn len(&self) -> usize {
+        self.ids.len()
+    }
+
+    /// Returns the key `K` associated with the given [`KeyId`].
+    ///
+    /// The output of using a [`KeyId`] not produced by the same instance of
+    /// [`ServiceKeys`] is unspecified, and it is likely to panic.
+    ///
+    /// # Panics
+    ///
+    /// This will panic if no entry is associated with the given lookup key.
+    pub(crate) fn get(&self, id: KeyId) -> &K {
+        self.keys
+            .get(&id)
+            .expect("distribution lookup keys must be valid")
+    }
+
+    fn try_get_id(&self, idx: usize) -> Option<KeyId> {
+        self.ids.get(idx).copied()
+    }
+
+    pub(crate) fn iter(&self) -> impl Iterator<Item = &KeyId> {
+        self.ids.iter()
+    }
+}
+
+// === impl WeightedKeys ===
+
+impl<K> WeightedServiceKeys<K> {
+    pub(crate) fn into_unweighted(self) -> ServiceKeys<K> {
+        ServiceKeys {
+            ids: self.ids,
+            keys: self
+                .keys
+                .into_iter()
+                .map(|(id, key)| (id, key.key))
+                .collect(),
+        }
+    }
+
+    pub(crate) fn weighted_index(&self) -> Result<WeightedIndex<u32>, WeightedError> {
+        WeightedIndex::new(self.ids.iter().map(|&id| self.get(id).weight))
+    }
+
+    pub(crate) fn validate_weights(&self) -> Result<(), WeightedError> {
+        self.weighted_index()?;
+        Ok(())
+    }
+
+    pub(crate) fn selector(&self) -> WeightedKeySelector<'_, K> {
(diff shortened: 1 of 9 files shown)


linkerd/distribute/src/params.rs
```diff
@@ -16,17 +20,38 @@ pub enum Distribution<K> {
...
+
+impl KeyId {
+    pub(crate) fn new(idx: usize) -> Self {
+        Self { idx }
+    }
+}
+
+#[derive(Debug)]
+pub struct UnweightedKeys<K> {
+    ids: Vec<KeyId>,
```
sfleen: This is the key part of this change. It decouples the backend keys from the iteration order of the backends with a separate key type. I considered using something like [SlotMap](https://docs.rs/slotmap/latest/slotmap/), but since we don't modify or remove anything from these maps after they're created we don't need the guarantee of stable keys that it provides and the slight performance cost that comes with it.

linkerd/distribute/src/keys.rs
```diff
@@ -0,0 +1,152 @@
...
+
+impl KeyId {
+    pub(crate) fn new(idx: usize) -> Self {
+        Self { idx }
+    }
+}
+
+#[derive(Debug)]
+pub struct UnweightedKeys<K> {
+    ids: Vec<KeyId>,
```
sfleen, line 23: This is the key part of this change. It decouples the backend keys from the iteration order of the backends with a separate key type. I considered using something like [SlotMap](https://docs.rs/slotmap/latest/slotmap/), but since we don't modify or remove anything from these maps after they're created we don't need the guarantee of stable keys that it provides and the slight performance cost that comes with it.

linkerd/distribute/src/keys.rs
```diff
@@ -0,0 +1,152 @@
...
+        // self.keys have the same length
+        for id in &self.ids {
+            self.keys.get(id).hash(state);
+        }
+    }
+}
+
+impl<K> UnweightedKeys<K> {
+    pub(crate) fn new(iter: impl Iterator<Item = K>) -> Self {
+        let mut ids = Vec::new();
```
sfleen, line 76: This is how the invariant of the IDs and keys being a 1:1 relation is guaranteed by only allowing construction via enumeration of the iterator over keys.

linkerd/distribute/src/params.rs
```diff
@@ -64,46 +67,46 @@ impl<K> Default for Distribution<K> {
...
-            keys,
-            weights,
-        })))
+
+        weighted_keys.validate_weights()?;
+        Ok(Self::RandomAvailable(Arc::new(weighted_keys)))
     }
 
-    pub(crate) fn keys(&self) -> &[K] {
+    pub(crate) fn make_svc_from_keys<T>(
```
sfleen, line 96: This was another key change. Instead of exposing a slice of keys to create the backend services, this inverts the control to something similar to a combinator to decouple how they keys are iterated over and mapped into services.

linkerd/distribute/src/stack.rs
```diff
@@ -55,16 +55,11 @@ where
...
         // Build the backends needed for this distribution, in the required
         // order (so that weighted indices align).
         let newk = self.inner.new_service(target);
-        let backends = dist
-            .keys()
-            .iter()
-            .map(|k| (k.clone(), newk.new_service(k.clone())))
-            .collect();
-        Distribute::new(backends, dist)
+        Distribute::new(dist, |k| newk.new_service(k.clone()))
```
sfleen, line 63: And now this is much simpler and is decoupled from how the service keys are stored.

cratelyn (commented):
here's a group of comments, _most_ of which are nits, and one [possible bug](https://github.com/linkerd/linkerd2-proxy/pull/3319#discussion_r1828506021)?

zooming out, i'm curious to tug at whether we can dodge the need to maintain a separate "keyspace" in order to account for duplicate keys.

i'll offer some more thoughts on that in a follow-up comment, but wanted to offer these comments in the meantime!

linkerd/distribute/src/service.rs
```diff
@@ -88,25 +92,22 @@ where
...
                     if svc.poll_ready(cx)?.is_ready() {
-                        self.ready_idx = Some(idx);
+                        self.ready_idx = Some(id);
                         return Poll::Ready(Ok(()));
                     }
 
                     // Zero out the weight of the backend we just tried so that
                     // it's not selected again.
-                    match index.update_weights(&[(idx, &0)]) {
+                    match selector.disable_backend(id) {
```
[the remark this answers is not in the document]
cratelyn: figured this out, it's by design that this only disables the backend for the duration of this `loop {}`.

i'm going to mark this as resolved!

linkerd/distribute/src/service.rs
```diff
@@ -196,11 +197,31 @@ mod tests {
...
+        assert!(svcs.is_empty());
+        dist
+    }
+
     #[test]
     fn empty_pending() {
         let mut dist_svc = mock::Spawn::new(Distribute::<&'static str, mock::Mock<(), ()>>::new(
             Default::default(),
-            Default::default(),
+            |_| unimplemented!(),
```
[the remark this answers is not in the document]
cratelyn: having spent some more time reading these tests, i like this change actually. it's a nice extra assurance that the test will panic if this ends up being called. that seems like a nice extra layer of assurance for a test about empty distributions never becoming available.

linkerd/distribute/src/keys.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub struct UnweightedKeys<K> {
+    ids: Vec<KeyId>,
+    keys: HashMap<KeyId, K>,
+}
+
+pub type WeightedKeys<K> = UnweightedKeys<WeightedKey<K>>;
```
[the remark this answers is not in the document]
cratelyn: i've wrapped my head around this more this morning. i'm going to mark this as resolved!

linkerd/distribute/src/params.rs
sfleen: Done.

linkerd/distribute/src/stack.rs
```diff
@@ -55,16 +55,11 @@ where
     /// Referencing other keys causes a panic.
     fn new_service(&self, target: T) -> Self::Service {
         let dist = self.extract.extract_param(&target);
-        tracing::debug!(backends = ?dist.keys(), "New distribution");
+        tracing::debug!(backends = ?dist, "New distribution");
 
         // Build the backends needed for this distribution, in the required
         // order (so that weighted indices align).
```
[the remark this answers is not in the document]
sfleen: Done

linkerd/distribute/src/params.rs
```diff
@@ -64,46 +63,46 @@ impl<K> Default for Distribution<K> {
...
+        &self,
+        mut make_svc: impl FnMut(&K) -> T,
```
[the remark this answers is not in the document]
sfleen: It took a bit of fiddling to get the tests to agree but I like this solution better. Done.

cratelyn (approved):

sfleen:
I've added an extra refactoring that creates separate structs for each distribution variant, which I think makes it a lot clearer and easier to understand.

linkerd/distribute/src/keys.rs
```diff
@@ -0,0 +1,160 @@
...
+    distributions::{WeightedError, WeightedIndex},
+    prelude::Distribution as _,
+    Rng,
+};
+use std::{collections::HashMap, hash::Hash};
+
+/// Uniquely identifies a key/backend pair for a distribution. This allows
+/// backends to have the same key and still participate in request distribution.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct KeyId {
```
olix0r, line 11: It looks like this type scope can be limited--we can keep this an internal implementation detail:
```suggestion
pub(crate) struct KeyId {
```


With this change, we would typically move the declaration down below the real pub types.

linkerd/distribute/src/keys.rs
```diff
@@ -0,0 +1,160 @@
...
+    idx: usize,
+}
+
+impl KeyId {
+    pub(crate) fn new(idx: usize) -> Self {
+        Self { idx }
+    }
+}
```
olix0r, line 19: nit: We tend to avoid mixing impls and type defs -- should generally move down below the other struct definitions.

linkerd/distribute/src/service/empty.rs
```diff
@@ -0,0 +1,38 @@
...
+    fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        tracing::debug!("empty distribution will never become ready");
+        Poll::Pending
+    }
+
+    #[inline]
+    fn call(&mut self, _: Req) -> Self::Future {
+        unreachable!("EmptySelection is never ready")
+    }
+}
```
olix0r, line 38: This type doesn't really carry any weight -- we can easily collapse this case into the parent service

linkerd/distribute/src/service/empty.rs
```diff
@@ -0,0 +1,38 @@
+use linkerd_stack::Service;
+use std::{
+    marker::PhantomData,
+    task::{Context, Poll},
+};
+
+#[derive(Debug, Copy, Clone)]
+pub struct EmptySelection<S> {
```
olix0r, line 8:
```suggestion
pub(crate) struct EmptySelection<S> {
```

linkerd/distribute/src/service/first.rs
```diff
@@ -0,0 +1,80 @@
+use crate::keys::ServiceKeys;
+use linkerd_stack::{NewService, Service};
+use std::task::{Context, Poll};
+
+#[derive(Debug)]
+pub struct FirstAvailableSelection<S> {
```
olix0r, line 6: Not actually public:

```suggestion
pub(crate) struct FirstAvailableSelection<S> {
```

linkerd/distribute/src/service/random.rs
```diff
@@ -0,0 +1,120 @@
...
+use rand::{distributions::WeightedError, rngs::SmallRng, SeedableRng};
+use std::{
+    collections::HashMap,
+    hash::Hash,
+    sync::Arc,
+    task::{Context, Poll},
+};
+
+#[derive(Debug)]
+pub struct RandomAvailableSelection<K, S> {
```
olix0r, line 12:
```suggestion
pub(crate) struct RandomAvailableSelection<K, S> {
```

linkerd/distribute/src/keys.rs
```diff
@@ -0,0 +1,160 @@
...
+
+#[derive(Debug)]
+pub struct ServiceKeys<K> {
+    ids: Vec<KeyId>,
+    keys: HashMap<KeyId, K>,
```
olix0r, line 24: Note that this crate already uses an AHashMap (faster hash algo), so it would be suitable to use that here as well
```suggestion
    keys: ahash::AHashMap<KeyId, K>,
```

linkerd/distribute/src/service/random.rs
```diff
@@ -0,0 +1,120 @@
...
+    keys: Arc<WeightedServiceKeys<K>>,
+    backends: HashMap<KeyId, S>,
```
olix0r, line 14: And here as well:
```suggestion
    backends: ashash::AHashMap<KeyId, S>,
```

olix0r (approved):
Overall approach looks good!

* We should `pub(crate)` structs that are not part of the public interface.
* We should use AHashMap to be consistent with the backend cache in this crate.
* tioli: The EmptySelection is trivial and can be eliminated.

linkerd/distribute/src/keys.rs
sfleen: Done

linkerd/distribute/src/keys.rs
sfleen: Done

linkerd/distribute/src/keys.rs
sfleen: Done

linkerd/distribute/src/service/empty.rs
sfleen: Done

linkerd/distribute/src/service/empty.rs
sfleen: Done

linkerd/distribute/src/service/first.rs
sfleen: Done

linkerd/distribute/src/service/random.rs
sfleen: Done

linkerd/distribute/src/service/random.rs
sfleen: Done

PR #3326, opened 2024-11-05
metrics: Add TCP and TLS level route metrics

zaharidichev:
This PR adds the following two metrics to both the TLS and TCP stack: 

- `outbound_(tcp/tls)_route_open_total` -  provides parent and route reference information.
- `outbound_(tcp/tls)_route_close_total` - provides parent and route information as well as an error label that is populated in case the connection has been closed due to an error.

In addition to that, we add the `forbidden` field on both the tls and tcp Route types. If this is set to true, the router stacks will through an Errno with a value of `ECONNREFUSED`. This will get propagated to the error label of the tcp close metric in order to indicate that this connection has been refused.

An example of these metrics for egress look like: 

Allowed connection to httpbin.org
```
outbound_tcp_route_open_total{
    parent_group="policy.linkerd.io"
    parent_kind="EgressNetwork"
    parent_namespace="linkerd-egress"
    parent_name="all-egress-traffic"
    parent_port="80"
    parent_section_name=""
    route_group=""
    route_kind="default"
    route_namespace=""
    route_name="tcp-egress-allow"
    target_ip="54.173.64.189"
    target_port="80"
} 1
```

Denied connection to httpbin.org
```
outbound_tcp_route_close_total{
    parent_group="policy.linkerd.io"
    parent_kind="EgressNetwork"
    parent_namespace="linkerd-egress"
    parent_name="all-egress-traffic"
    parent_port="80"
    parent_section_name=""
    route_group=""
    route_kind="default"
    route_namespace=""
    route_name="tcp-egress-deny"
    target_ip="54.173.64.189"
    target_port="80"
    error="ECONNREFUSED"
} 1
```


linkerd/app/outbound/src/opaq/logical/route.rs
```diff
@@ -78,13 +97,21 @@ where
                 // consideration, so we must eagerly fail requests to prevent
                 // leaking tasks onto the runtime.
                 .push_on_service(svc::LoadShed::layer())
+                .push_filter(|rt: Self| -> Result<_, Error> {
+                    if rt.params.forbidden {
+                        Err(Errno::from(Code::ECONNREFUSED).into())
```
zaharidichev, line 102: Not sure this will actually work for Windows. We seem to not have this enum there...

linkerd/transport-metrics/src/tcp.rs
```diff
@@ -0,0 +1,104 @@
+use linkerd_errno::Errno;
```
zaharidichev, line 1: Any suggestions around naming, placement module organization, etc are all welcome

linkerd/app/outbound/src/opaq/logical/route.rs
```diff
@@ -77,13 +96,21 @@ where
...
+                    } else {
+                        Ok(rt)
+                    }
+                })
```
olix0r, line 105: I don't think we want to use Errno -- the existing IO errors won't be represented like this.

In a first version, we should probably opt for bespoke (thiserror) types indicating Forbidden (and Internal?) errors. We may need to adjust that in light of metrics but for now it seems like a starting point that will at least show up in the logs.

I like using push_filter for this, though 👍 .

olix0r (commented):
Let's get a version of this PR that focuses on wiring the forbidden (and invalid backend) variants into the stack as very simple "filters" -- this gets our types set up the way we want without getting into changing the proxy-api. That is, we can read the routeerror and invalidbackend types to build filters for now; and in the future we have the option to add real filter variants in the API.

I'll have some additional suggestions for the metrics portion of this, but I think it makes sense to finish wiring up the API in a distinct change.

zaharidichev:
closing in favor of: https://github.com/linkerd/linkerd2-proxy/pull/3354

PR #3335, opened 2024-11-06
feat(inbound): Add metrics for rate-limiting

alpeb:
Added the counter `inbound_http_local_ratelimit_total`:

```
# HELP inbound_http_local_ratelimit_total The total number of inbound HTTP requests that were rate-limited
# TYPE inbound_http_local_ratelimit_total counter
inbound_http_local_ratelimit_total{target_addr="10.42.0.22:8080",target_ip="10.42.0.22",target_port="8080",srv_group="policy.linkerd.io",srv_kind="server",srv_name="web-http",ratelimit_group="policy.linkerd.io",ratelimit_kind="HTTPLocalRateLimitPolicy",ratelimit_name="web-http",ratelimit_error="per-identity(20)",tls="true",client_id="default.emojivoto.serviceaccount.identity.linkerd.cluster.local"} 19
```

This implied adding another field `local_rate_limit_meta` to ServerPolicy to track the metadata.

### DIFF
--- a/linkerd/app/core/src/metrics.rs
+++ b/linkerd/app/core/src/metrics.rs
@@ -404,7 +404,7 @@ impl FmtLabels for HTTPLocalRateLimitLabels {
                 self.error,
             )
         } else {
-            Ok(())
+            write!(f, ",ratelimit_error=\"{}\"", self.error)
         }
     }
 }
--- a/linkerd/app/core/src/metrics.rs
+++ b/linkerd/app/core/src/metrics.rs
@@ -96,11 +96,11 @@ pub struct RouteAuthzLabels {
     pub authz: Arc<policy::Meta>,
 }
 
-// Labels referencing an inbound HTTP local rate limit.
+/// Labels referencing an inbound HTTP local rate limit.
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
 pub struct HTTPLocalRateLimitLabels {
     pub server: ServerLabel,
-    pub rate_limit: Arc<policy::Meta>,
+    pub rate_limit: Option<Arc<policy::Meta>>,
     pub error: String,
 }
 
@@ -394,14 +394,18 @@ impl FmtLabels for RouteAuthzLabels {
 impl FmtLabels for HTTPLocalRateLimitLabels {
     fn fmt_labels(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         self.server.fmt_labels(f)?;
-        write!(
-            f,
-            ",ratelimit_group=\"{}\",ratelimit_kind=\"{}\",ratelimit_name=\"{}\",ratelimit_error=\"{}\"",
-            self.rate_limit.group(),
-            self.rate_limit.kind(),
-            self.rate_limit.name(),
-            self.error,
-        )
+        if let Some(rl) = &self.rate_limit {
+            write!(
+                f,
+                ",ratelimit_group=\"{}\",ratelimit_kind=\"{}\",ratelimit_name=\"{}\",ratelimit_error=\"{}\"",
+                rl.group(),
+                rl.kind(),
+                rl.name(),
+                self.error,
+            )
+        } else {
+            Ok(())
+        }
     }
 }
 
--- a/linkerd/app/core/src/metrics.rs
+++ b/linkerd/app/core/src/metrics.rs
@@ -96,6 +96,14 @@ pub struct RouteAuthzLabels {
     pub authz: Arc<policy::Meta>,
 }
 
+// Labels referencing an inbound HTTP local rate limit.
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct HTTPLocalRateLimitLabels {
+    pub server: ServerLabel,
+    pub rate_limit: Arc<policy::Meta>,
+    pub error: String,
+}
+
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct OutboundEndpointLabels {
     pub server_id: tls::ConditionalClientTls,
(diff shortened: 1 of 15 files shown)


linkerd/proxy/server-policy/src/lib.rs
```diff
@@ -144,18 +146,25 @@ pub mod proto {
...
+                }) => local_rate_limit.unwrap_or_default(),
                 _ => Default::default(),
             };
 
+            let local_rate_limit_meta = local_rate_limit
+                .clone()
+                .metadata
+                .map(TryInto::try_into)
+                .transpose()?
+                .unwrap_or(Meta::Default { name: "".into() });
```
cratelyn, line 166: this empty name gives me pause. if we already have a `Meta` object in `local_rate_limit`, why not just use that? `Option<T>` values can be encoded as labels, (_see below_), so maybe we can work with optional metadata without creating a default, unnamed `Meta` when none exists.

what do you think?

https://docs.rs/prometheus-client/latest/prometheus_client/encoding/trait.EncodeLabelValue.html#impl-EncodeLabelValue-for-Option%3CT%3E

cratelyn (commented):
i have a question about our model, specifically about whether `Option<Meta>` may be the right way to handle rate limit policies that might not have any metadata.

similarly, i wonder if it might be more natural to add this to the `LocalRateLimit` structure itself, rather than attaching a sibling field in our `ServerPolicy`.

i'm not _too_ strongly attached to these ideas, but things like `Default { name: "" }` feel like we're cutting against the grain of the wood, so to speak. :carpentry_saw: :thought_balloon:

alpeb:
Thanks @cratelyn, that makes a lot of sense! In 2eaf27f6 I've moved meta under LocalRateLimit and could get rid of all those awkward empty defaults. I'm still not leveraging EncodeLabelValue, but gonna take a look at that to see what I can do. That shouldn't block this PR though, IMO.

linkerd/app/core/src/metrics.rs
```diff
@@ -394,14 +394,18 @@ impl FmtLabels for RouteAuthzLabels {
...
+                f,
+                ",ratelimit_group=\"{}\",ratelimit_kind=\"{}\",ratelimit_name=\"{}\",ratelimit_error=\"{}\"",
+                rl.group(),
+                rl.kind(),
+                rl.name(),
+                self.error,
+            )
+        } else {
+            Ok(())
+        }
```
cratelyn, line 408:
```suggestion
        } else {
            write!(f, ",ratelimit_error=\"{}\"", self.error)
        }
```

one last tweak; in the case where no metadata exists, we can still format the error label

linkerd/app/core/src/metrics.rs
```diff
@@ -96,11 +96,11 @@ pub struct RouteAuthzLabels {
     pub authz: Arc<policy::Meta>,
 }
 
-// Labels referencing an inbound HTTP local rate limit.
+/// Labels referencing an inbound HTTP local rate limit.
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
 pub struct HTTPLocalRateLimitLabels {
     pub server: ServerLabel,
```
cratelyn, line 102: i noticed that we do not use this field in the `FmtLabels` implementation. should we? is this extraneous?

cratelyn (commented):
thanks for addressing my comments! one little nit, regarding how we format labels when the metadata does not exist, and one small question about a field that i don't see being used.

linkerd/app/core/src/metrics.rs
alpeb: It is used [here](https://github.com/linkerd/linkerd2-proxy/pull/3335/files#diff-cbb69683b3eaa8be3d3ab859c0b39bb83708d56eead603980bff21137de4591eR396) :-)

cratelyn (approved):

linkerd/app/core/src/metrics.rs
cratelyn: oh, so it is!! thanks :slightly_smiling_face:

linkerd/app/core/src/metrics.rs
```diff
@@ -96,6 +96,14 @@ pub struct RouteAuthzLabels {
     pub authz: Arc<policy::Meta>,
 }
 
+/// Labels referencing an inbound HTTP local rate limit.
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct HTTPLocalRateLimitLabels {
```
olix0r, line 101: Let's keep this in the app-inbound crate -- we should generally avoid adding things to the core crate if we can avoid it.

linkerd/app/inbound/src/policy.rs
```diff
@@ -133,6 +134,19 @@ impl AllowPolicy {
         ServerLabel(self.server.borrow().meta.clone())
     }
 
+    pub fn ratelimit_label(&self, error: &RateLimitError) -> HTTPLocalRateLimitLabels {
+        let error = match error {
+            RateLimitError::Total(rps) => format!("total({})", rps),
+            RateLimitError::PerIdentity(rps) => format!("per-identity({})", rps),
+            RateLimitError::Override(rps) => format!("override({})", rps),
```
olix0r, line 141: We should not use label values to encode the RPS like this. Let's keep label values simple, encoding only one piece of information.

What do you think about making this a simple `scope="total|identity"` -- the "per-identity" limit is really the default when there is no override for a client.

alpeb: Sounds good to me. Changed in 6fc6d756.

linkerd/app/inbound/src/metrics/authz.rs
```diff
@@ -52,6 +57,13 @@ struct TcpInner {
     terminate: Mutex<HashMap<ServerKey, Counter>>,
 }
 
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct HTTPLocalRateLimitLabels {
+    pub server: ServerLabel,
+    pub rate_limit: Option<Arc<Meta>>,
+    pub scope: String,
```
olix0r, line 64: If we update this to reflect that scopes are known statically, we can avoid needless string allocation

```suggestion
    pub scope: &'static str,
```

PR #3338, opened 2024-11-07
suggestions on correct address types

zaharidichev:


linkerd/app/gateway/src/opaq.rs
```diff
@@ -49,24 +49,22 @@ impl Gateway {
...
-        let addr: GatewayAddr = opaq.param();
-        let discovery: &outbound::Discovery<T> = &opaq;
-
-        let Some(profile) = svc::Param::<Option<profiles::Receiver>>::param(discovery) else {
+        let addr = opaq.param();
+        let (routes, paddr) = outbound::opaq::routes_from_discovery(
+            addr.0.clone().into(),
+            svc::Param::param(&*opaq),
+            svc::Param::param(&*opaq),
+        );
```
zaharidichev, line 62: I could not figure out why my call to `::param `was not compiling without the type annotations

olix0r: Ah yeah the compiler is getting tripped up by our svc::Param type constraint, but since we don't have the Param trait in scope, we can't just use `(*opaq).param()`. By moving this into the function call position, we can use the argument's type to eliminate the explicit type hint.

In https://github.com/linkerd/linkerd2-proxy/pull/3338/commits/0b7bef9f006c113b0d81846f32f585c089442f52 I've imported svc::Param so we can eliminate more of the boilerplate.

PR #3355, opened 2024-11-12
feat(outbound): TCP/TLS route metrics

olix0r:
This PR adds transport level route metrics for TCP and TLS. The metrics added are: 
- outbound_(tcp/tls)_route_open_total - provides parent and route reference information.
- outbound_(tcp/tls)_route_close_total - provides parent and route information as well as an error label that is populated in case the connection has been closed due to an error.


linkerd/transport-metrics/src/tcp/client.rs
```diff
@@ -0,0 +1,112 @@
...
+    fn call(&mut self, io: I) -> Self::Future {
+        const ERRNO_UNKNOWN: i32 = 0;
+        let target = self.target.clone();
+        let new_accept = self.inner.clone();
+        let labels: L = target.param();
+        let params = self.params.clone();
+
+        Box::pin(async move {
+            let metrics_open = params.metrics_open(labels.clone());
+            let svc = new_accept.new_service(target);
```
olix0r, line 92: This should be structured such that we don't "break" the stacking--the NewService should eagerly do as much work as possible and this future should only handle registering completion metrics.

olix0r (commented):
My general sense is:

* We don't want the errno changes, or anything to do with errno in the short term.
* We should use a simple 'Error' enum with expected errors ('Forbidden', 'I/O', 'Invalid', 'Unexpected', '')--we can expand this with additional variants matching against IO error codes, but we can use a bounded set rather than one that takes a numeric code.
* If we have a constrained set of errors, we should eagerly register them rather than lazily.
  * I.e. we have a TcpRouteMetricsFamily that is registered against route labels in a NewService, which produces a TcpRouteMetrics with various error counters.
* My sense is that we don't gain much from the current factoring/reusability of InstrumentConnection--we may be better off having all of the metrics and labels defined in the outbound crate. I'm not sure that the middleware portion of this is particularly resusable. Given that errors are defined in the outbound crate, it probably makes sense for all of the middleware to be bound in the outbound crate. If we do want a reusable middleware, it probabably need not be prom::Counter-oriented (i.e. it could hold a trait that calls a function on `Result<(), &Error>`).

This is all to try to avoid being generic in places that are not really advantageous.

linkerd/app/outbound/src/metrics/transport.rs
```diff
@@ -0,0 +1,271 @@
...
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, io: I) -> Self::Future {
+        let target = self.target.clone();
+        let new_accept = self.inner.clone();
+        let metrics = self.metrics.clone();
+
+        Box::pin(async move {
+            let svc = new_accept.new_service(target);
```
olix0r, line 231: Can't this happen in the NewService impl so that this middleware can properly proxy poll_ready()?

zaharidichev: Will give it a try

zaharidichev: Addressed in https://github.com/linkerd/linkerd2-proxy/pull/3355/commits/2549251dd15c71a8e64cb0bf9673e2515669ea16

linkerd/app/outbound/src/metrics/transport.rs
```diff
@@ -0,0 +1,262 @@
...
+        self.inner.poll_ready(cx).map_err(Into::into)
+    }
+
+    fn call(&mut self, io: I) -> Self::Future {
+        let metrics = self.metrics.clone();
+        let inner = self.inner.clone();
+
+        Box::pin(async move {
+            metrics.inc_open();
+            match inner.oneshot(io).await.map_err(Into::into) {
```
olix0r, line 223:
```suggestion

        metrics.inc_open();
        let call = self.inner.call(io);

        Box::pin(async move {
            match call.await.map_err(Into::into) {
```

The clone/oneshot invalidates the prior poll_ready -- poll_ready "reserves" us the ability to send a request. We want to use that reservation. When we clone, we invalidate that reservation. Oneshot combines waiting for a reservation and dispatching the request.

olix0r:
LGTM once we fix the service middleware.

linkerd/app/outbound/src/metrics/transport.rs
zaharidichev: @olix0r  Done in https://github.com/linkerd/linkerd2-proxy/pull/3355/commits/5ac135212f1a7f51d083d99bf7e7d7231d5d972d

zaharidichev (approved):

PR #3377, opened 2024-11-21
policy: use TLS/TCP filters from proxy API response

zaharidichev:
This PR updates the proxy API dependency to a new version that uses filters to express errors for TLS and TCP routes. 

This PR depends on https://github.com/linkerd/linkerd2-proxy-api/pull/405

### DIFF
--- a/linkerd/app/outbound/src/tls/logical/route/filters.rs
+++ b/linkerd/app/outbound/src/tls/logical/route/filters.rs
@@ -10,11 +10,11 @@ where
     let filters: &[tls::Filter] = &t.param();
     if let Some(filter) = filters.iter().next() {
         match filter {
-            tls::Filter::ForbiddenRoute => {
+            tls::Filter::Forbidden => {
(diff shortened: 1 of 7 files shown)


linkerd/app/outbound/src/tls/logical/route/filters.rs
```diff
@@ -10,11 +10,11 @@ where
...
     if let Some(filter) = filters.iter().next() {
         match filter {
-            tls::Filter::ForbiddenRoute => {
+            tls::Filter::Forbidden => {
                 return Err(errors::TLSForbiddenRoute.into());
             }
 
-            tls::Filter::InvalidBackend(message) => {
+            tls::Filter::Invalid(message) => {
                 return Err(errors::TLSInvalidBackend(message.clone()).into());
```
zaharidichev, line 18: We can change the name of these error types to be more generic if we want to. We currently return Invalid only if backends are invalid and Forbidden only if a route is forbidden, so this seems fine as it is for now 🤷

PR #3378, opened 2024-11-21
metrics: correctly reflect closed connections in route metrics

zaharidichev:
This PR fixes a problem that was introduced in the following commit: https://github.com/linkerd/linkerd2-proxy/pull/3355/commits/5ac135212f1a7f51d083d99bf7e7d7231d5d972d

The problem is that if a particular route is configured to fail a connection, the `poll_ready` call on the inner service will return the error and the service will never get called. Therefore in order to capture that in metrics, we need to increment the metrics counter if we encounter errors while calling` poll_ready` on the inner service.

### DIFF
--- a/linkerd/app/outbound/src/metrics/transport.rs
+++ b/linkerd/app/outbound/src/metrics/transport.rs
@@ -211,11 +211,7 @@ where
 
     #[inline]
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        self.inner.poll_ready(cx).map_err(|error| {
-            let error = error.into();
-            self.metrics.inc_closed(Some(&*error));
-            error
-        })
+        self.inner.poll_ready(cx).map_err(Into::into)
     }
 
     fn call(&mut self, io: I) -> Self::Future {
--- a/linkerd/app/outbound/src/metrics/transport.rs
+++ b/linkerd/app/outbound/src/metrics/transport.rs
@@ -211,7 +211,11 @@ where
 
     #[inline]
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        self.inner.poll_ready(cx).map_err(Into::into)
+        self.inner.poll_ready(cx).map_err(|error| {
+            let error = error.into();
+            self.metrics.inc_closed(Some(&*error));
+            error
+        })
     }
 
     fn call(&mut self, io: I) -> Self::Future {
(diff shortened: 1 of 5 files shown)


linkerd/app/outbound/src/metrics/transport.rs
```diff
@@ -211,7 +211,11 @@ where
 
     #[inline]
     fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        self.inner.poll_ready(cx).map_err(Into::into)
+        self.inner.poll_ready(cx).map_err(|error| {
+            let error = error.into();
+            self.metrics.inc_closed(Some(&*error));
```
zaharidichev, line 216: I wonder whether we should only increment the closed counter or we should increment both the open and the close ones.. 🤔

cratelyn:
this isn't a comment on this particular change so much as the architecture of this middleware, which predates this PR, but...

a "closed" counter is usually, in my experience, a good place to lean on RAII guards that increment a closed counter when dropped. that could be stored as an `Option<Deferred>` value that will increment the "unexpected" counter by default when dropped, to work around tricky edge cases like this.

that'd be a more significant change though, so i'll just note the idea here for posterity.

olix0r:
> The problem is that if a particular route is configured to fail a connection, the poll_ready call on the inner service will return the error and the service will never get called.

The poll contract isn't really tied to the connection contract: calling poll_ready doesn't indicate that connection is being attempted, strictly speaking.

I think a preferable solution is to modify the stack to replace the push_filter with a service that runs the filter in the request-response path...

zaharidichev:
@olix0r Addressed feedback in https://github.com/linkerd/linkerd2-proxy/pull/3378/commits/9c3dc5afa9e38dc327c8e4eedcb9a0ffebcacdac

linkerd/app/outbound/src/opaq/logical/route/filters.rs
```diff
@@ -1,30 +1,98 @@
...
-    Ok(t)
+    fn call(&mut self, io: I) -> Self::Future {
+        let call = self.inner.call(io);
+        let apply = self.apply_filters();
+
+        Box::pin(async move {
+            apply?;
+            call.await.map_err(Into::into)
+        })
+    }
```
olix0r, line 95: Note that this can potentially dispatch the connection even when the filters error--if the call future is spawned on another task, it may progress despite the fact that we don't poll it in the returned future.

Instead, we can do something like:

```rust
    type Future = future::Either<
        future::ErrInto<S::Future, Error>,
        future::Ready<Result<S::Response, Error>>,
    >; 

        if let Err(e) = self.apply_filters() {
            return future::Either::Right(future::err(e));
        }
        future::Either::Left(self.inner.call(io).err_into())
    }
```

This also saves us an allocation (Box).

olix0r (commented):
I can push a commit that addresses this.

olix0r (approved):

PR #3382, opened 2024-11-21
refactor(proxy/http): extricate `Body` middleware types

cratelyn:
this branch pulls a number of components providing `http_body::Body` middleware out of `linkerd-proxy-http` and into new libraries: `linkerd-http-classify`, `linkerd-http-stream-timeouts`, `linkerd-http-retain`, and `linkerd-http-override-authority`.

these are each performed in distinct commits, to facilitate review.

PR #3386, opened 2024-11-22
refactor(proxy/http): create `linkerd-http-upgrade` crate

cratelyn:
this moves the inter-related `upgrade` and `glue` submodules out of the
`linkerd-proxy-http` library and into a new standalone crate.

this pulls one last `Body` middleware out of the core `linkerd-proxy-http` library.

NB: this is based on #3382.


cratelyn:
i've cherry-picked these commits into #3382, i'm going to close this.

PR #3418, opened 2024-12-04
ci: increase linkerd-install readiness timeout

olix0r:
CI runs frequently timeout waiting for pod readiness. There is a hardcoded 1m timeout.

This change makes this timeout configurable, specifying a 2m timeout in the CI configuration.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -127,6 +127,8 @@ jobs:
     if: needs.meta.outputs.cargo_changed == 'true' || needs.meta.outputs.rust_changed == 'true'
     timeout-minutes: 20
     runs-on: ubuntu-latest
+    env:
+      WAIT_TIMOUT: 2m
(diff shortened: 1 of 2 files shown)


.github/workflows/pr.yml
```diff
@@ -127,6 +127,8 @@ jobs:
     if: needs.meta.outputs.cargo_changed == 'true' || needs.meta.outputs.rust_changed == 'true'
     timeout-minutes: 20
     runs-on: ubuntu-latest
+    env:
+      WAIT_TIMOUT: 2m
```
alpeb, line 131:
```suggestion
      WAIT_TIMEOUT: 2m
```

cratelyn (approved):

.github/workflows/pr.yml
cratelyn: very good catch, thank you

PR #3419, opened 2024-12-04
chore(deps): Upgrade tokio-rustls to 0.26

sfleen:
This bumps rustls itself from 0.21 to 0.23, which comes with a few breaking API changes. Most of these are limited to types being moved or renamed, or how the certificate verifiers are constructed.


linkerd/meshtls/rustls/Cargo.toml
```diff
@@ -12,11 +12,11 @@ test-util = ["linkerd-tls-test-util"]
...
 futures = { version = "0.3", default-features = false }
 ring = { version = "0.17", features = ["std"] }
-rustls-pemfile = "1.0"
-rustls-webpki = { version = "0.101.5", features = ["std"] }
+rustls-pemfile = "2.2"
+rustls-webpki = { version = "0.102.8", features = ["std"] }
 thiserror = "1"
 tokio = { version = "1", features = ["macros", "rt", "sync"] }
-tokio-rustls = { version = "0.24", features = ["dangerous_configuration"] }
+tokio-rustls = { version = "0.26", default_features = false, features = ["ring", "logging"] }
```
sfleen, line 19: By default, `tokio-rustls` uses the `aws-lc-rs` backend. I'd like to keep any backend changes out of a version bump like this, so I've kept this as our current `ring` backend and we can change this later if we so desire.

linkerd/meshtls/rustls/src/creds/store.rs
```diff
@@ -57,12 +59,17 @@ pub(super) fn server_config(
     // controlling the set of trusted signature algorithms), but they provide good enough
     // defaults for now.
     // TODO: lock down the verification further.
-    let client_cert_verifier = Arc::new(
-        rustls::server::AllowAnyAnonymousOrAuthenticatedClient::new(roots),
-    );
-    rustls::ServerConfig::builder()
-        .with_cipher_suites(TLS_SUPPORTED_CIPHERSUITES)
-        .with_safe_default_kx_groups()
+    let mut provider = rustls::crypto::ring::default_provider();
```
sfleen, line 62: This is the most invasive part of this upgrade. In order to set the supported cipher suites to just the ones we want, we have create a crypto provider separately from the provider.

Also, the `AllowAnyAnonymousOrAuthenticatedClient` has been removed and replaced with client/server verifier builder methods.

linkerd/app/outbound/Cargo.toml
```diff
@@ -54,7 +54,7 @@ futures-util = "0.3"
 http-body = "0.4"
 hyper = { version = "0.14", features = ["deprecated", "http1", "http2"] }
 tokio = { version = "1", features = ["macros", "sync", "time"] }
-tokio-rustls = "0.24"
+tokio-rustls = "0.26"
```
olix0r, line 57: Needs default-features=false, etc

olix0r (commented):
Consider pulling the rustls dependencies up to the workload (/Cargo.toml) so we can pin the featureset in one place

linkerd/meshtls/rustls/src/creds/verify.rs
```diff
@@ -1,23 +1,33 @@
...
+    roots: Arc<RootCertStore>,
+    supported: WebPkiSupportedAlgorithms,
+}
 
 impl AnySanVerifier {
     pub(crate) fn new(roots: impl Into<Arc<RootCertStore>>) -> Self {
-        Self(roots.into())
+        Self {
+            roots: roots.into(),
+            supported: rustls::crypto::ring::default_provider().signature_verification_algorithms,
```
sfleen, line 25: This change I'm not entirely confident in. The cert verification methods below now take a list of supported algorithms instead of implicitly using the default, and this was the only reasonable way I was able to get the default list that the `WebPkiServerVerifier` that the implementation we copied was using.

linkerd/app/outbound/src/tls/logical/tests.rs
```diff
@@ -171,28 +172,37 @@ fn generate_client_hello(sni: &str) -> Vec<u8> {
...
     let hs_payload = HandshakeMessagePayload {
         typ: HandshakeType::ClientHello,
         payload: HandshakePayload::ClientHello(ClientHelloPayload {
             client_version: ProtocolVersion::TLSv1_2,
             random: Random::from([0; 32]),
-            session_id: SessionId::empty(),
+            session_id: SessionId::read(&mut Reader::init(&[0])).unwrap(),
             cipher_suites: vec![CipherSuite::TLS_NULL_WITH_NULL_NULL],
             compression_methods: vec![Compression::Null],
-            extensions: vec![ClientExtension::make_sni(sni.borrow())],
```
sfleen, line 195: The `ClientExtension::make_sni` was made private (these aren't part of the `rustls` public API), which is the motivator behind the changes in this file.

The purpose of this function is to generate a `ClientHello` message for tests, and in my playing around I wasn't able to find a more ergonomic way of generating one outside of this. I'm definitely open to ideas so we don't have to rely on internal `rustls` types for this.

linkerd/app/outbound/Cargo.toml
sfleen: Done.

linkerd/meshtls/rustls/src/creds.rs
```diff
@@ -104,7 +107,7 @@ pub fn default_for_test() -> (Store, Receiver) {
 }
 
 mod params {
-    use tokio_rustls::rustls;
+    use tokio_rustls::rustls::{self};
```
cratelyn, line 110: nit: can't this remain as `use tokio_rustls::rustls;`? i'm suspecting this is residue from development.

cratelyn (approved):
this looks good to me! great work.

as a small aside, i hadn't ever thought to try `dependency.workspace = true` syntax before. that's nice. :slightly_smiling_face:

linkerd/meshtls/rustls/src/creds.rs
sfleen: Good catch, definitely development residue. Fixed.

Cargo.lock
```diff
@@ -4,19 +4,13 @@ version = 3
...
 dependencies = [
  "gimli",
 ]
 
-[[package]]
-name = "adler"
-version = "1.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
-
```
olix0r, line 19: Are these dependency changes required as part of your update, or are these creeping in another way? I would expect a much [narrower diff here](https://gist.github.com/olix0r/40b68edfdc6b335ed1537d4908d8abdf).

To get back to a cogent state: `cargo co main -- Cargo.lock && cargo check`

linkerd/meshtls/rustls/Cargo.toml
```diff
@@ -12,11 +12,11 @@ test-util = ["linkerd-tls-test-util"]
...
 futures = { version = "0.3", default-features = false }
 ring = { version = "0.17", features = ["std"] }
-rustls-pemfile = "1.0"
-rustls-webpki = { version = "0.101.5", features = ["std"] }
+rustls-pemfile = "2.2"
+rustls-webpki = { version = "0.102.8", features = ["std"] }
 thiserror = "1"
 tokio = { version = "1", features = ["macros", "rt", "sync"] }
-tokio-rustls = { version = "0.24", features = ["dangerous_configuration"] }
+tokio-rustls.workspace = true
```
olix0r, line 19: this is cool! but for consistency let's use
```suggestion
tokio-rustls = { workspace = true }
```

linkerd/meshtls/rustls/src/creds/store.rs
```diff
@@ -1,37 +1,38 @@
...
 use linkerd_error::Result;
 use linkerd_identity as id;
 use linkerd_meshtls_verifier as verifier;
 use ring::{rand, signature::EcdsaKeyPair};
 use std::{convert::TryFrom, sync::Arc};
 use tokio::sync::watch;
-use tokio_rustls::rustls;
+use tokio_rustls::rustls::{self, server::WebPkiClientVerifier};
 use tracing::debug;
+use webpki::types::UnixTime;
```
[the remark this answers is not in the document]
sfleen: I wasn't able to puzzle this one out, but turns out we don't need the `rustls_webpki` dep at all and can just use the types exported from `tokio_rustls` directly instead.

linkerd/meshtls/rustls/Cargo.toml
sfleen: Done.

Cargo.lock
sfleen: I think a `cargo update` bumped a bunch of the transitive deps in here, reverting them seems to have sorted it out.

linkerd/meshtls/rustls/src/creds.rs
```diff
@@ -27,19 +27,22 @@ pub fn watch(
...
+                warn!(%error, "invalid trust anchors file");
+                return Err(error.into());
+            }
+            Ok(cert) => certs.push(cert),
+        };
+    }
+    if certs.is_empty() {
+        warn!("no valid certs in trust anchors file");
+        return Err("no trust roots in PEM file".into());
+    }
```
olix0r, line 43: In the interest of minimizing the diff, we can keep this more consistent with the pre-existing code:
```suggestion
    let certs = match rustls_pemfile::certs(&mut std::io::Cursor::new(roots_pem))
        .collect::<Result<Vec<_>, _>>()
    {
        Err(error) => {
            warn!(%error, "invalid trust anchors file");
            return Err(error.into());
        }
        Ok(certs) if certs.is_empty() => {
            warn!("no valid certs in trust anchors file");
            return Err("no trust roots in PEM file".into());
        }
        Ok(certs) => certs,
    };
```

linkerd/meshtls/rustls/src/creds/store.rs
```diff
@@ -1,37 +1,37 @@
...
 pub(super) fn client_config_builder(
-    cert_verifier: Arc<dyn rustls::client::ServerCertVerifier>,
+    cert_verifier: Arc<dyn rustls::client::danger::ServerCertVerifier>,
 ) -> rustls::ConfigBuilder<rustls::ClientConfig, rustls::client::WantsClientCert> {
-    rustls::ClientConfig::builder()
-        .with_cipher_suites(TLS_SUPPORTED_CIPHERSUITES)
-        .with_safe_default_kx_groups()
+    let mut provider = rustls::crypto::ring::default_provider();
+    provider.cipher_suites = TLS_SUPPORTED_CIPHERSUITES.to_vec();
+
```
olix0r, line 33: It is probably preferable to provide a single `provider()` function that returns a provider with our customizations applied, decoupling most of our code from ring, ideally.

linkerd/meshtls/rustls/src/creds/verify.rs
olix0r: Can we instead use the restricted set that we know about?

https://github.com/linkerd/linkerd2-proxy/pull/3419/files#diff-7c54eeeb80146328f2d2a680f933677d0645157007e161590456652105f809adL114

linkerd/meshtls/rustls/src/creds/verify.rs
```diff
@@ -1,23 +1,33 @@
...
+        }
     }
 }
 
-// This is derived from `rustls::client::WebPkiVerifier`.
+// This is derived from `rustls::client::WebPkiServerVerifier`.
```
olix0r, line 30: We should probably update the link/copyright below this to refer to the appropriate code.

linkerd/meshtls/rustls/src/creds.rs
sfleen: Done

linkerd/meshtls/rustls/src/creds/store.rs
sfleen: Done

linkerd/meshtls/rustls/src/creds/verify.rs
sfleen: Done

olix0r (commented):
Thanks for making the requested changes. Something is still amiss with the Cargo.lock, though. To fix this, I would probably run something akin to `git co origin/main -- Cargo.lock && cargo check`. There's probably something funky in your dev setup that is causing gratuitous lock modification?

sfleen:
I've fixed the Cargo.lock, best as I can figure RA in vscode was clobbering lockfile changes for some reason.

olix0r (approved):

olix0r:
Good work, @sfleen!

cratelyn:
nice job, @sfleen!

PR #3428, opened 2024-12-05
refactor(app/test): minor tweaks to `linkerd-app-test`

cratelyn:
while handling the upgrade to hyper 1.0, i noticed some small changes that'd be nice to make.

most importantly, with respect to https://github.com/linkerd/linkerd2/issues/8733, this commit outlines `http_util::http_request()`. hyper 1.0 provides version specific `SendRequest` types for HTTP/1 and HTTP/2, so rather than try to be polymorphic across both senders, we send the request at the call site.

two other commits remove `pub` attributes for functions that aren't called externally. we'll address `run_proxy()` and `connect_client()` in a subsequent PR, because the dependent tests use both HTTP/1 and HTTP/2.


linkerd/app/inbound/src/http.rs
```diff
@@ -74,7 +74,14 @@ pub mod fuzz {
                             .header(header_name, header_value)
                             .body(Body::default())
                         {
-                            let rsp = http_util::http_request(&mut client, req).await;
+                            let rsp = client
+                                .ready()
+                                .await
+                                .expect("HTTP client poll_ready failed")
+                                .call(req)
+                                .await
```
olix0r, line 82: tioli: you can use [ServiceExt::oneshot](https://docs.rs/tower/latest/tower/trait.ServiceExt.html#method.oneshot) to make all of these calls `client.clone().oneshot(req).await.expect("...")`

olix0r (approved):

linkerd/app/inbound/src/http.rs
cratelyn: this didn't seem to compile unfortunately, because the `SendRequest` type isn't cloneable.

that said, none of these tests _need_ it to be, because they only send a single request. so, https://github.com/linkerd/linkerd2-proxy/pull/3428/commits/d16ccc0f6e50e7ae09673fc6036fba3cd33be2d0 calls `oneshot()`, sans cloning.

this looks much better, thank you for suggesting this :slightly_smiling_face:

PR #3456, opened 2024-12-13
chore(hyper): define hyper as a workspace dependency

cratelyn:
this commit alters various crates' manifests, pointing to a common workspace-level hyper dependency.

note that the lockfile is not altered; this commit does *not* affect the version of hyper
used, or have any other affect on the dependency graph. this will make future maintenance
of our hyper dependency marginally easier.

see linkerd/linkerd2#8733 for more information on upgrading to hyper 1.0.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,6 +97,7 @@ debug = 1
 lto = true
 
 [workspace.dependencies]
+hyper = { version = "0.14", default-features = false }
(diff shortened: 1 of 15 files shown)


olix0r (approved):
It's probably also worth including http and http-body...

cratelyn:
> It's probably also worth including http and http-body...

very astute point! i will do that in a subsequent follow-up this morning :city_sunrise:

PR #3466, opened 2024-12-16
chore(cargo.toml): define http workspace dependencies

cratelyn:
this is a follow-up to https://github.com/linkerd/linkerd2-proxy/pull/3456, addressing a suggestion that was made
during review:

- <https://github.com/linkerd/linkerd2-proxy/pull/3456#pullrequestreview-2502446232>
- <https://github.com/linkerd/linkerd2/issues/8733>

this branch modifies the workspace manifest, defining `http` and `http-body`
as common workspace dependencies.

no changes to the lockfile are made because these commits do not affect
the dependency graph of the project.

PR #3467, opened 2024-12-16
refactor: move away from legacy `hyper::body::HttpBody`

cratelyn:
this is an incremental step away from hyper 0.14's request and response
body interfaces, and towards the 1.0 body types. see
<https://github.com/linkerd/linkerd2/issues/8733> for more information
about upgrading to hyper 1.0.

hyper 0.14 provides a `hyper::body::Body` that is removed in the 1.0
interface. `hyper-util` now provides a workable default body type. hyper
0.14 reëxports `http_body::Body` as `HttpBody`. hyper 1.0 reëxports this
trait as `hyper::body::Body` without any renaming.

this commit moves application code away from hyper's legacy `Body` type
and the `HttpBody` trait alias. this commit moves assorted interfaces
towards the boxed `BoxBody` type instead. when possible, code is tweaked
such that it refers to the reëxport in `linkerd-proxy-http`, rather than
directly through `hyper`.

NB: this commit is based upon https://github.com/linkerd/linkerd2-proxy/pull/3466.

### DIFF
--- a/linkerd/app/admin/src/server.rs
+++ b/linkerd/app/admin/src/server.rs
@@ -12,13 +12,9 @@
 
 use futures::future::{self, TryFutureExt};
 use http::StatusCode;
-use hyper::{
-    body::{Body, HttpBody},
-    Request, Response,
-};
 use linkerd_app_core::{
     metrics::{self as metrics, FmtMetrics},
-    proxy::http::ClientHandle,
+    proxy::http::{Body, BoxBody, ClientHandle, Request, Response},
     trace, Error, Result,
 };
 use std::{
@@ -45,7 +41,7 @@ pub struct Admin<M> {
     pprof: Option<crate::pprof::Pprof>,
 }
 
-pub type ResponseFuture = Pin<Box<dyn Future<Output = Result<Response<Body>>> + Send + 'static>>;
+pub type ResponseFuture = Pin<Box<dyn Future<Output = Result<Response<BoxBody>>> + Send + 'static>>;
 
 impl<M> Admin<M> {
     pub fn new(
@@ -73,30 +69,30 @@ impl<M> Admin<M> {
         self
     }
 
-    fn ready_rsp(&self) -> Response<Body> {
+    fn ready_rsp(&self) -> Response<BoxBody> {
         if self.ready.is_ready() {
             Response::builder()
                 .status(StatusCode::OK)
                 .header(http::header::CONTENT_TYPE, "text/plain")
-                .body("ready\n".into())
+                .body(BoxBody::new::<String>("ready\n".into()))
                 .expect("builder with known status code must not fail")
         } else {
             Response::builder()
                 .status(StatusCode::SERVICE_UNAVAILABLE)
-                .body("not ready\n".into())
+                .body(BoxBody::new::<String>("not ready\n".into()))
                 .expect("builder with known status code must not fail")
         }
     }
 
-    fn live_rsp() -> Response<Body> {
+    fn live_rsp() -> Response<BoxBody> {
         Response::builder()
             .status(StatusCode::OK)
             .header(http::header::CONTENT_TYPE, "text/plain")
-            .body("live\n".into())
+            .body(BoxBody::new::<String>("live\n".into()))
             .expect("builder with known status code must not fail")
     }
 
-    fn env_rsp<B>(req: Request<B>) -> Response<Body> {
+    fn env_rsp<B>(req: Request<B>) -> Response<BoxBody> {
         use std::{collections::HashMap, env, ffi::OsString};
 
         if req.method() != http::Method::GET {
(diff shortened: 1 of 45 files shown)


linkerd/app/core/src/errors/respond.rs
```diff
@@ -388,7 +388,7 @@ impl<R> Respond<R> {
 
 impl<B, R> respond::Respond<http::Response<B>, Error> for Respond<R>
 where
-    B: Default + hyper::body::HttpBody,
+    B: Default + linkerd_proxy_http::Body,
```
cratelyn, line 391: :point_up: an example of replacing `hyper::body::HttpBody` with a stable `linkerd_proxy_http::Body`.

linkerd/metrics/src/serve.rs
```diff
@@ -33,22 +33,22 @@ impl<M> Serve<M> {
 }
 
 impl<M: FmtMetrics> Serve<M> {
-    pub fn serve<B>(&self, req: http::Request<B>) -> std::io::Result<http::Response<Body>> {
+    pub fn serve<B>(&self, req: http::Request<B>) -> std::io::Result<http::Response<BoxBody>> {
```
cratelyn, line 36: :point_up: see places like this as well, where we move interfaces towards `BoxBody`.

linkerd/app/admin/src/server/json.rs
```diff
@@ -41,26 +44,28 @@ pub(crate) fn accepts_json<B>(req: &http::Request<B>) -> Result<(), http::Respon
...
 }
 
-fn mk_rsp(status: StatusCode, val: &impl serde::Serialize) -> http::Response<Body> {
+fn mk_rsp(status: StatusCode, val: &impl serde::Serialize) -> http::Response<BoxBody> {
     match serde_json::to_vec(val) {
         Ok(json) => http::Response::builder()
             .status(status)
             .header(header::CONTENT_TYPE, JSON_HEADER_VAL.clone())
-            .body(json.into())
+            .body(BoxBody::new(http_body::Full::new(bytes::Bytes::from(json))))
```
sfleen, line 60: `BoxBody::new(http_body::Full::new(bytes::Bytes::from(` is even more cumbersome, is there a way to shorten this with another `From` impl?

linkerd/app/admin/src/server.rs
```diff
@@ -73,30 +69,30 @@ impl<M> Admin<M> {
...
     }
 
-    fn ready_rsp(&self) -> Response<Body> {
+    fn ready_rsp(&self) -> Response<BoxBody> {
         if self.ready.is_ready() {
             Response::builder()
                 .status(StatusCode::OK)
                 .header(http::header::CONTENT_TYPE, "text/plain")
-                .body("ready\n".into())
+                .body(BoxBody::new::<String>("ready\n".into()))
```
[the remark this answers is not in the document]
cratelyn: how does bf08e2c look to you?

i added a `from_static(..)` constructor, which is approximately equivalent but follows in the grain of other `http` types like [`HeaderName`](https://docs.rs/http/latest/http/header/struct.HeaderName.html#method.from_static), [`HeaderValue`](https://docs.rs/http/latest/http/header/struct.HeaderValue.html#method.from_static), [`Authority`](https://docs.rs/http/latest/http/uri/struct.Authority.html#method.from_static), [`PathAndQuery`](https://docs.rs/http/latest/http/uri/struct.PathAndQuery.html#method.from_static), and [`Uri`](https://docs.rs/http/latest/http/uri/struct.Uri.html#method.from_static).

thank you for suggesting this! the turbofish felt a little icky as i was writing this, so i'm happy to see some of the call-sites be cleaned up by this change. :slightly_smiling_face:

linkerd/app/admin/src/server/json.rs
cratelyn: i wrestled with this one a bit, and some of the trouble feels like a fundamental reflection of how the hyper interface will change in 1.0. there used to be a https://docs.rs/hyper/0.14.31/hyper/body/struct.Body.html type included that let us use a simple `.into()` before.

as we prepare for upgrading, we'll need to move towards providing our own `B: Body` type. that'll be a https://docs.rs/http-body-util/latest/http_body_util/struct.Full.html in the 1.0 interface, which expects a `D: Buf` [buffer.](https://docs.rs/bytes/1.6.0/bytes/buf/trait.Buf.html)

so, some of this _is_ just a bit more cumbersome, but 4557952 tries to mitigate that a bit by hoisting this out of the `match` arm and leaning on `Result::map` to apply these constructors.

does that look a little better to you, @sfleen?

cratelyn:
NB: i just enabled auto-merge :slightly_smiling_face:

linkerd/app/admin/src/server.rs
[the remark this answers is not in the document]
sfleen: Looks good!

linkerd/app/admin/src/server/json.rs
sfleen: Pulling that out helps a fair amount I think. It's unfortunate that the hyper API has boxed us in like that, but I don't see a way around it that isn't fraught with footguns.

PR #3469, opened 2024-12-17
Change default trace protocol to OpenTelemetry

sfleen:
Now that OpenTelemetry has been available for a few releases, we should enable it as the default to begin to phase out OpenCensus.

This sets the default protocol for the proxy, we'll need to make the corresponding changes to the control plane as well.


olix0r:
Doesn't the control plane always have to tell the proxy to use the opentelemetry protocol?

This change, aiui, would cause proxies to default to opentelemetry when running with an older control plane that does not know to set this default.

I think we do *not* want to change the proxy's defaulting behavior with this in mind: the decision should be driven/controlled by the injector.

PR #3479, opened 2024-12-19
chore(http/upgrade): replace `hyper::Body` with `BoxBody`

cratelyn:
`hyper::Body` is removed in the 1.0 version.

this commit removes it from our upgrade facilities, using a generic body parameter that defaults to BoxBody.

see <https://github.com/linkerd/linkerd2/issues/8733>.

### DIFF
--- a/linkerd/http/upgrade/src/glue.rs
+++ b/linkerd/http/upgrade/src/glue.rs
@@ -1,9 +1,9 @@
 use crate::upgrade::Http11Upgrade;
-use bytes::Bytes;
-use futures::TryFuture;
+use futures::{ready, TryFuture};
 use http_body::Body;
 use hyper::client::connect as hyper_connect;
 use linkerd_error::{Error, Result};
+use linkerd_http_box::BoxBody;
 use linkerd_io::{self as io, AsyncRead, AsyncWrite};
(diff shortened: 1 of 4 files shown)


linkerd/http/upgrade/src/glue.rs
```diff
@@ -50,9 +51,13 @@ pub struct HyperConnectFuture<F> {
 
 // === impl UpgradeBody ===
 
-impl Body for UpgradeBody {
-    type Data = Bytes;
-    type Error = hyper::Error;
+impl<B> Body for UpgradeBody<B>
+where
+    B: Body + Unpin,
```
sfleen, line 56: Correct me if I'm wrong here, but I don't think the `Unpin` bound is necessary here since we mark `B` as `#[pin]` in the type

sfleen (approved):

linkerd/http/upgrade/src/glue.rs
cratelyn: you are right, and i have to commend and thank you for the watchful eye.

in `main` this isn't currently pinned, so this was needed to add the `B` parameter originally in development, but tweaking how we poll the body (_see lines 70-80, below_) means this bound is indeed frivolous now.

039d90bd removes this.

PR #3483, opened 2024-12-20
Include pod UID and container name as optional trace labels

sfleen:
It's useful to include these fields in the trace span attributes to help correlate traffic within a cluster.

This propagates these labels through the trace initialization, similar to how we handle the service name (and other labels that last for the entire pod lifetime).

See https://github.com/linkerd/linkerd2/pull/13501 for the corresponding control plane change.

### DIFF
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -149,6 +149,7 @@ const ENV_OUTBOUND_DISABLE_INFORMATIONAL_HEADERS: &str =
 const ENV_TRACE_ATTRIBUTES_PATH: &str = "LINKERD2_PROXY_TRACE_ATTRIBUTES_PATH";
 const ENV_TRACE_PROTOCOL: &str = "LINKERD2_PROXY_TRACE_PROTOCOL";
 const ENV_TRACE_SERVICE_NAME: &str = "LINKERD2_PROXY_TRACE_SERVICE_NAME";
+const ENV_TRACE_EXTRA_ATTRIBUTES: &str = "LINKERD2_PROXY_TRACE_EXTRA_ATTRIBUTES";
 
 /// Constrains which destination names may be used for profile/route discovery.
 ///
@@ -432,6 +433,7 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
     let hostname = strings.get(ENV_HOSTNAME);
 
     let trace_attributes_file_path = strings.get(ENV_TRACE_ATTRIBUTES_PATH);
+    let trace_extra_attributes = strings.get(ENV_TRACE_EXTRA_ATTRIBUTES);
     let trace_protocol = strings.get(ENV_TRACE_PROTOCOL);
     let trace_service_name = strings.get(ENV_TRACE_SERVICE_NAME);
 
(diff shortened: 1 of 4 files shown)


cratelyn:
you've been clippy'ed :paperclip: :scream_cat: 

```
Error: error: this function has too many arguments (8/7)
  --> linkerd/app/src/trace_collector/otel_collector.rs:21:1
   |
21 | / pub(super) fn create_collector<S>(
22 | |     addr: ControlAddr,
23 | |     hostname: Option<String>,
24 | |     pod_uid: Option<String>,
...  |
29 | |     legacy_metrics: metrics::Registry,
30 | | ) -> EnabledCollector
   | |_____________________^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#too_many_arguments
   = note: `-D clippy::too-many-arguments` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::too_many_arguments)]`
```

linkerd/app/src/env.rs
```diff
@@ -266,6 +266,9 @@ const ENV_OUTBOUND_HTTP1_CONNECTION_POOL_IDLE_TIMEOUT: &str =
 
 const ENV_SHUTDOWN_GRACE_PERIOD: &str = "LINKERD2_PROXY_SHUTDOWN_GRACE_PERIOD";
 
+const ENV_POD_UID: &str = "_pod_uid";
+const ENV_POD_CONTAINER_NAME: &str = "_pod_containerName";
```
cratelyn, line 270: the case here is mildly surprising to me, but that's an implementation detail that i'll consider out of scope for a pr in the proxy.

cratelyn (approved):
besides the note about clippy lints above, this looks like a good PR to me. this should be a nice improvement to our traces!

linkerd/app/src/env.rs
sfleen: It's a little odd to me too, but this is the convention that the control plane uses for populating k8s metadata and I didn't see a strong reason to deviate from that.

olix0r, line 270: We've generally been pretty adamant (and successful) about not leaking kubernetes concepts (pods, containers, uids, etc) into the proxy's codebase.

You may notice that we load a bunch of attributes from a file at starttime:

https://github.com/linkerd/linkerd2-proxy/pull/3483/files#diff-0ee25f0088036de1851c9f218f1abb2fd0c4abecefe846112c34c6533b2267ccR840-R845

is it possible to encode these attributes into this file so the proxy doesn't need special casing for them?

olix0r (changes requested):

linkerd/app/src/env.rs
sfleen: AFAICT, the downward API doesn't let you encode more than one field ref in a single file, nor does it give you particularly good control over the keys for the values (unless you're using labels or annotations, and I'd rather not dump all of those for a pod into each trace).

That being said, I think we can still encode these through environment variables as a list of arbitrary extra fields to add. Thoughts on https://github.com/linkerd/linkerd2/pull/13544?

sfleen: Cool, I've updated this PR to reflect those changes to the attributes, PTAL

olix0r (approved):

PR #3484, opened 2024-12-20
Use correct semantic conventions for trace labels

sfleen:
The OpenTelemetry spec defines the semantic conventions that HTTP services should use for the labels included in traces: https://opentelemetry.io/docs/specs/semconv/http/http-spans/

Previously, we were using an outdated version of this spec for the OpenCensus traces. This updates the labels to match the current spec.

The notable changes updates to the path for HTTP method and status code, the fields that include the URL parts, and more rigorously following the standard for propagating the Host header.

### DIFF
--- a/linkerd/trace-context/src/service.rs
+++ b/linkerd/trace-context/src/service.rs
@@ -1,5 +1,6 @@
 use crate::{propagation, Span, SpanSink};
 use futures::{future::Either, prelude::*};
+use http::Uri;
 use linkerd_stack::layer;
 use std::{
     collections::HashMap,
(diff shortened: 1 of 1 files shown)


linkerd/trace-context/src/service.rs
```diff
@@ -35,20 +36,35 @@ impl<K: Clone, S> TraceContext<K, S> {
...
-        labels.insert("http.method", format!("{}", req.method()));
-        let path = req
-            .uri()
-            .path_and_query()
-            .map(|pq| pq.as_str().to_owned())
-            .unwrap_or_default();
-        labels.insert("http.path", path);
-        if let Some(authority) = req.uri().authority() {
-            labels.insert("http.authority", authority.as_str().to_string());
+        let mut labels = HashMap::with_capacity(6);
```
cratelyn, line 39: nit: it looks like there are 7 entries:
`url.scheme`, `url.path`, `url.query`, `server.address`, `server.port`, `http.request.method`, and `http.response.status_code`.

i'll call this a nit only because the capacity of `5` that was there before this PR has the same issue.

linkerd/trace-context/src/service.rs
```diff
@@ -35,20 +36,35 @@ impl<K: Clone, S> TraceContext<K, S> {
     }
 
     fn request_labels<B>(req: &http::Request<B>) -> HashMap<&'static str, String> {
```
cratelyn, line 38:
```suggestion
    /// Returns labels for the provided request.
    ///
    /// The OpenTelemetry spec defines the semantic conventions that HTTP services should use for the labels included in traces: https://opentelemetry.io/docs/specs/semconv/http/http-spans/
    fn request_labels<B>(req: &http::Request<B>) -> HashMap<&'static str, String> {
```

it might be nice to add a note documenting where the spec we are following lives.

**nb:** this suggestion will need to be formatted, sorry :see_no_evil:

cratelyn (approved):
this looks good! i have two really minor nits to offer, but the change itself seems great :slightly_smiling_face:

linkerd/trace-context/src/service.rs
sfleen: Good call, fixed

PR #3540, opened 2025-01-17
refactor(http/upgrade): `Http11Upgrade` is `Clone`

cratelyn:
refactor(http/upgrade): `Http11Upgrade` is `Clone`

this commit makes `Http11Upgrade` a cloneable type.

see <https://github.com/linkerd/linkerd2/issues/8733>.

in the 1.0 interface of the `http` crate, request and response
extensions must now satisfy a `Clone` bound.

`Http11Upgrade` was written before this was the case, and is very
intentionally designed around the idea that it *not* be cloneable.

`insert_half()` in particular could cause the proxy to panic if it were
to clone a request or response's extensions. it might call
`insert_half()` a second time, and discover that the `TryLock<T>` had
already been set.

moreover, holding on to a copy of the extensions would prevent the
`Drop` method for `Inner` from being called. This would cause
connections that negotiate an HTTP/1.1 upgrade to deadlock due to the
`OnUpgrade` futures never being polled, and failing to create a `Duplex`
that acts as the connection's I/O transport.

this commit makes use of the alterations to `Http11Upgrade` made in
previous commits, and adds a *safe* implementation of `Clone`. by
only shallowly copying the extension, we tie the upgrade glue to a
*specific* request/response.

the extension can be cloned, but any generated copies will be inert.

### DIFF
--- a/linkerd/http/upgrade/src/upgrade.rs
+++ b/linkerd/http/upgrade/src/upgrade.rs
@@ -90,19 +90,23 @@ impl Http11Upgrade {
     }
 
     pub fn insert_half(self, upgrade: OnUpgrade) {
-        match self.half {
-            Half::Server => {
-                let mut lock = self
-                    .inner
+        match self {
+            Self {
+                inner,
+                half: Half::Server,
+            } => {
+                let mut lock = inner
                     .server
                     .try_lock()
                     .expect("only Half::Server touches server TryLock");
                 debug_assert!(lock.is_none());
                 *lock = Some(upgrade);
             }
-            Half::Client => {
-                let mut lock = self
-                    .inner
+            Self {
+                inner,
+                half: Half::Client,
+            } => {
+                let mut lock = inner
                     .client
                     .try_lock()
                     .expect("only Half::Client touches client TryLock");
--- a/linkerd/http/upgrade/src/upgrade.rs
+++ b/linkerd/http/upgrade/src/upgrade.rs
@@ -26,7 +26,7 @@ use try_lock::TryLock;
 // implement `Clone` for this type.
 pub struct Http11Upgrade {
     half: Half,
-    inner: Arc<Inner>,
+    inner: Option<Arc<Inner>>,
 }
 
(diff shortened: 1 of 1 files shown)


linkerd/http/upgrade/src/upgrade.rs
cratelyn, line 1: disclaiming that i've seen the rustdoc warnings further down in this file. i'll pick those up on monday.

linkerd/http/upgrade/src/upgrade.rs
```diff
@@ -80,35 +78,40 @@ impl Http11Upgrade {
...
+                half: Half::Client,
+            } => {
+                let mut lock = inner
                     .client
                     .try_lock()
                     .expect("only Half::Client touches client TryLock");
                 debug_assert!(lock.is_none());
                 *lock = Some(upgrade);
             }
+            Self { inner: None, .. } => {}
```
olix0r, line 114: I don't think we should let this case occur gracefully -- if inset_half is called on a defunct clone of an upgrade request, it means the upgrade won't succeed, and we shouldn't continue as if it would. So insert_half should probably be converted to return a Result so that the [HTTP client](https://github.com/linkerd/linkerd2-proxy/blob/8b64841fc12c70f19a3276f6921b5aeaf3d046c0/linkerd/proxy/http/src/h1.rs#L139-L169) can bubble the error up to the caller.

olix0r (commented):
and doc lints, as you noted...

linkerd/http/upgrade/src/upgrade.rs
cratelyn: that's a great idea! 10f6c99 and 59d31d5 perform this transformation, propagating this error up in the HTTP/1 client.

linkerd/http/upgrade/src/upgrade.rs
cratelyn: resolved, ba9b6c4

olix0r (approved):

PR #3556, opened 2025-01-22
feat(http/retry): add a unit test suite to `PeekTrailersBody<B>`

cratelyn:
https://github.com/linkerd/linkerd2/issues/8733 tracks the upgrade to hyper 1.0.

the `PeekTrailersBody<B>` body middleware, provided by the `linkerd-http-retry` crate, is uniquely positioned to be adversely affected by the changes to the `Body` trait. this type, in order to attempt to peek an inner `B`-typed body's HTTP/2 trailers, has some subtle and nuanced edge cases.

the intent of this type is to only yield the asynchronous task responsible for reading the body _once_. depending on what the inner body yields, and when, this can result in combinations of: no data frames and no trailers, no data frames with trailers, one data frame and no trailers, one data frame with trailers, or two data frames. moreover, depending on which of these are yielded, the body will call `.await` some scenarios, and only poll functions _once_ in others.

migrating this to the `Frame<T>` and `poll_frame()` style of the 1.0 `Body` interface, away from the 0.4 interface that provides distinct `poll_data()` and `poll_trailers()` methods, is fundamentally tricky.

this branch is most importantly focused on introducing a unit test suite to `PeekTrailersBody<B>`, to help provide assurance that our body middleware behaves as expected across this migration boundary.

along the way, some documentation is added, and some minor changes are made to the control flow of the polling logic to carve a path for later changes related to polling.

PR #3559, opened 2025-01-22
feat(http/retry): model `PeekTrailersBody<B>` with `Frame<T>` (work-in-progress)

cratelyn:
this is work in progress.

this is based on #3556 and #3558.

see https://github.com/linkerd/linkerd2/issues/8733 for more information.


cratelyn:
ci failures at the moment are to be expected.

linkerd/http/retry/src/peek_trailers.rs
```diff
@@ -17,31 +18,34 @@ use std::{
...
-    /// The inner body's trailers, if it was terminated by a `TRAILERS` frame
-    /// after 0-1 DATA frames, or an error if polling for trailers failed.
+    /// This variant indicates that the inner body's trailers could not be observed.
+    Unknown {
+        first: Option<Result<B::Data, B::Error>>,
+        second: Option<Result<B::Data, B::Error>>,
+        /// The inner [`Body`].
+        #[pin]
+        inner: B,
+    },
```
cratelyn, line 40: bikeshedding for this variant is welcome. `Stream` seemed like another alternative, but it feels like a particularly overloaded term. i chose `Unknown` because it doesn't _guarantee_ that two DATA frames exist: the second frame may have not been immediately available.

linkerd/http/retry/src/compat.rs
```diff
@@ -0,0 +1,115 @@
...
+    pub(crate) fn frame(&mut self) -> combinators::Frame<'_, B> {
+        combinators::Frame(self)
+    }
+}
+
+/// Future that resolves to the next frame from a `Body`.
+///
+/// NB: This is a vendored stand-in for [`Frame<'a, T>`][frame], and and can be replaced once
+/// we upgrade from http-body 0.4 to 1.0. This file was vendored, and subsequently adapted to this
+/// project, at commit 86fdf00.
```
cratelyn, line 49: "vendored" here is doing some heavy lifting, as the `poll()` implementation must do a bit more work to determine whether we should delegate down to `poll_data()` or `poll_data()`, but the "shape" of this in terms of its `Future::Output` etc., are the same.

linkerd/http/retry/src/compat.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    pub(crate) fn into_inner(self) -> B {
+        self.inner
+    }
+
+    /// Returns a future that resolves to the next frame.
+    pub(crate) fn frame(&mut self) -> combinators::Frame<'_, B> {
+        combinators::Frame(self)
+    }
```
cratelyn, line 42: this gives us a method that lines up with https://docs.rs/http-body-util/0.1.2/src/http_body_util/lib.rs.html#34-39

cratelyn: ...so we can use this to poll for frames in a forward-compatible way, and then call `into_inner()` to get the `B` back out when we're finished peeking.

linkerd/http/retry/src/peek_trailers.rs
```diff
@@ -94,55 +108,103 @@ impl<B: Body> PeekTrailersBody<B> {
...
+    /// frame, and `Err(trailers)` if it is given a TRAILERS frame.
+    ///
+    /// This is an internal helper to facilitate pattern matching in `read_body(..)`, above.
+    fn split_frame(frame: crate::compat::Frame<B::Data>) -> Result<B::Data, HeaderMap> {
+        frame
+            .into_data()
+            .map_err(|frame| match frame.into_trailers() {
+                Ok(trls) => trls,
+                // Safety: this is not reachable, we called `into_data()` above.
+                Err(_) => unreachable!("into_data() and `into_trailers()` both returned `Err(_)`"),
```
cratelyn, line 206: i'm unsure how much to worry about the commented out `//Unknown(Box<dyn Frameish>),` variant in the vendored frame code. in other words, might this someday be reachable?

`Result<T, E>` gives us something we can pattern match on, which `Frame<T>` does not for the sake of semantic versioning. if we fear hitting this in the future, we could define a local enum like so...

```rust
enum ParsedFrame<T> {
    Data(T),
    Trailers(HeaderMap),
    Other(Frame<T>),
}
```

...and return that here. what do you think?

linkerd/http/retry/src/peek_trailers.rs
```diff
@@ -17,31 +18,34 @@ use std::{
 /// If the first frame of the body stream was *not* a `TRAILERS` frame, this
 /// behaves identically to a normal body.
 #[pin_project(project = Projection)]
-pub struct PeekTrailersBody<B: Body = BoxBody> {
-    /// The inner [`Body`].
+pub enum PeekTrailersBody<B: Body = BoxBody> {
```
cratelyn, line 21: a note for review: this part of the diff is best viewed in a side-by-side view, or by reading the file itself in tree. we've largely replaced how we model `PeekTrailersBody<B>`, so this diff does not present itself particularly well.

as noted elsewhere, we cannot pattern match on a `Frame<T>`. between that, and the multiple edge-cases we are bound to encounter depending on the kinds of frames yielded, and the readiness of the second frame, storing fields like `first_data: Option<Result<Frame<B::Data>, B::Error>>` ended up causing significant headaches in previous experiments.

cratelyn (commented):
:speech_balloon: breadcrumbs for review...

linkerd/http/retry/src/peek_trailers.rs
olix0r: This isn't a hard blocker, but it would generally be preferable to do something like

```
pub struct PeekTrailersBody<B: Body = BoxBody>(Inner<B>);

enum Inner<B> { ... }
```

so that the enum variants/fields are private.

linkerd/http/retry/src/peek_trailers.rs
olix0r: Perhaps `Buffered`? In this state we've buffered data and can't peek further.

linkerd/http/retry/src/peek_trailers.rs
```diff
@@ -94,55 +108,103 @@ impl<B: Body> PeekTrailersBody<B> {
...
-        Self {
-            inner: body,
-            first_data,
-            trailers,
-        }
+        body
     }
 
-    /// Returns a response with an inert [`PeekTrailersBody<B>`].
+    /// Splits a `Frame<T>` into a `Result<T, E>`.
```
olix0r, line 193: It might be slightly clearer as an `Either`? We can borrow a an Either type from `futures::future::Either`.

linkerd/http/retry/src/peek_trailers.rs
olix0r: Or this could be `Option<Either<B::Data, HeaderMap>>`, and then the caller can be adapted so this is handled as an Empty body... Since the API specifically doesn't give us exhaustive matching it's probably best to handle it cleanly.

olix0r (approved):
This is really cool! A few suggestions

linkerd/http/retry/src/peek_trailers.rs
cratelyn: great idea. this is done in a3096ef4

linkerd/http/retry/src/peek_trailers.rs
cratelyn: agreed. this is addressed in 501b3cf0

linkerd/http/retry/src/peek_trailers.rs
cratelyn: i like this name much better, thank you! renamed in 2557a2f3

cratelyn:
@olix0r, review comments have been addressed! i'd love if you could take a quick look before i hit the merge button :slightly_smiling_face:

olix0r (approved):
looks great!

PR #3561, opened 2025-01-24
metrics: Make hostname label on TLS metrics configurable

zaharidichev:
This is a draft PR that illustrates a proposed approach for piping configuration for blanking the hostname label for outbound route metrics.

### DIFF
--- a/linkerd/app/outbound/src/sidecar.rs
+++ b/linkerd/app/outbound/src/sidecar.rs
@@ -17,6 +17,8 @@ use std::fmt::Debug;
 use tokio::sync::watch;
 use tracing::info_span;
 
+pub(crate) type AllowHostnameLabels = bool;
+
 /// A target type holding discovery information for a sidecar proxy.
 #[derive(Clone, Debug)]
 struct Sidecar {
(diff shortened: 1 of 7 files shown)


linkerd/app/outbound/src/sidecar.rs
```diff
@@ -376,6 +384,12 @@ impl svc::Param<watch::Receiver<tls::Routes>> for TlsSidecar {
     }
 }
 
+impl svc::Param<AllowHostnameLabels> for TlsSidecar {
```
zaharidichev, line 387: In my mind sticking that to the target allows for easier way to eventually rely on discovery information for configuring that options. Correct me if I am wrong here.

olix0r:
I spent some more time on this... I found it helpful to look at the HTTP side of this change in tandem so we could find a consistent approach: going back on what I said verbally earlier today, I think we want to attach this policy at the route level... the control plane should be able to enable this functionality on a per-client level, and so the route response is a natural place to control it.

A fairly targeted version of this here:
* https://github.com/linkerd/linkerd2-proxy/pull/3606

Based on the setup PRs:
* https://github.com/linkerd/linkerd2-proxy/pull/3604
* https://github.com/linkerd/linkerd2-proxy/pull/3605

PR #3606, opened 2025-02-10
feat(outbound)!: disable 'hostname' metrics label

olix0r:
The 'hostname' label value is derived from the request URI and can be influenced
by applications. If a client uses a high cardinality of hostname values, this
leads to the proxy exporting many metrics and ultimately exhausting resources.

This change disables the 'hostname' label by default. A new field is outed to
the HTTP and TLS route parameters, `export_hostname_labels`, which can be used
to enable the 'hostname' label; however, this type is not yet configurable.
Changes are required to the proxy API and, ultimately, the policy controller, to
make this field configurable.

When the hostname is not exported, an empty label value is used.

BREAKING: The 'hostname' label is now disabled by default.

### DIFF
--- a/linkerd/proxy/client-policy/src/http.rs
+++ b/linkerd/proxy/client-policy/src/http.rs
@@ -14,6 +14,7 @@ pub struct RouteParams {
     pub timeouts: Timeouts,
     pub retry: Option<Retry>,
     pub allow_l5d_request_headers: bool,
+    pub export_hostname_labels: bool,
 }
 
 // TODO: keepalive settings, etc.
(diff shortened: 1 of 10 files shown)


cratelyn (approved):
:heavy_check_mark: nice!

linkerd/proxy/client-policy/src/http.rs
```diff
@@ -299,7 +300,13 @@ pub mod proto {
             .ok_or(InvalidHttpRoute::Missing("distribution"))?
             .try_into()?;
 
-        let mut params = RouteParams::try_from_proto(timeouts, retry, allow_l5d_request_headers)?;
+        let export_hostname_labels = false;
```
zaharidichev, line 303: maybe a comment that it should come from discovery

zaharidichev (approved):
looks great

PR #3609, opened 2025-02-11
Fix missing attributes in OpenTelemetry traces

sfleen:
A previous change enabling OpenTelmetry forgot to include attributes from the span itself, in addition to the request level spans as we do in the old OpenCensus exporter. This adds those attributes back in.


olix0r (approved):
When merging this, please use a conventional commit summary. For example:
```
fix(opentelemetry): preserve span labels in traces
```

PR #3614, opened 2025-02-12
refactor(http/retry): outline `ForwardCompatibleBody<B>`

cratelyn:
in https://github.com/linkerd/linkerd2-proxy/pull/3559 (https://github.com/linkerd/linkerd2-proxy/commit/4b53081a84080decd0a52216ef3582f8e0066ab4), we introduced a backported `Frame<T>` type, and a
`ForwardCompatibleBody<B>` type that allows us to interact with a
`http_body::Body` circa 0.4.6 in terms of frame-based interfaces that
match those of the 1.0 interface.

see https://github.com/linkerd/linkerd2/issues/8733 for more information on upgrading hyper.

in https://github.com/linkerd/linkerd2-proxy/pull/3559, we narrowly added this as an internal submodule of the
`linkerd-http-retry` library. these facilities however, would have
utility in other places such as `linkerd-app-core`.

this commit pulls these compatibility shims out into a
`linkerd-http-body-compat` library so that they can be imported and
reused elsewhere.

PR #3616, opened 2025-02-13
refactor(app/core): prepare rescue body for http-body upgrade

cratelyn:
refactor(app/core): prepare rescue body for http-body upgrade

this commit makes some minor alterations to our error recovery body
middleware. see https://github.com/linkerd/linkerd2/issues/8733 for more
information.

this commit removes an `assert!` statement from the implementation of
`<Response<R, B> as Body>::poll_data()`. see the documentation of
`Body::poll_frame()`:

> Once the end of the stream is reached, implementations should
> continue to return [`Poll::Ready(None)`].

https://github.com/hyperium/http-body/commit/1090bfff26884779e026f4cab37a2a3ee165a69d#diff-33aabe8c2aaa7614022addf244245e09bbff576a67a9ae3c6938c8a868201d36R60-R61

to do this, this commit introduces a distinct terminal state
`Inner::Rescued` to represent when the underlying `B`-typed body has
yielded an error and been rescued. once in this state the body will
yield no more data frames, instead yielding a collection of trailers
describing the mid-stream error that was encountered by the underlying
body.

the call to `R::rescue` is also moved down into the helper function fka
`grpc_trailers()`. this helps the function follow the grain of our
"state machine" a little more directly.

see #3615, #3614, and #3611 for pretext to this change.

### DIFF
--- a/linkerd/app/core/src/errors/body.rs
+++ b/linkerd/app/core/src/errors/body.rs
@@ -2,7 +2,7 @@ use super::{
     header::{GRPC_MESSAGE, GRPC_STATUS},
     respond::{HttpRescue, SyntheticHttpResponse},
 };
-use http::header::HeaderValue;
+use http::{header::HeaderValue, HeaderMap};
 use linkerd_error::{Error, Result};
 use pin_project::pin_project;
 use std::{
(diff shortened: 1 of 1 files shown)


olix0r (approved):
nice!

PR #3627, opened 2025-02-20
chore(app/spire): make spire compilation possible for non-linux targets

zaharidichev:
At the moment the fact that SPIRE uses Unix domain sockets prevents us from building
the proxy for non-unix targets. This PR adds special casing that allows us to build for
these targets but makes the spire client fail at runtime. Non-unix support for Spire shall
be introduced at a later stage.

### DIFF
--- a/linkerd/app/src/spire.rs
+++ b/linkerd/app/src/spire.rs
@@ -1,12 +1,12 @@
 use linkerd_app_core::{exp_backoff::ExponentialBackoff, Error};
 use std::sync::Arc;
-use tokio::net::UnixStream;
 use tokio::sync::watch;
-use tonic::transport::{Endpoint, Uri};
 
 pub use linkerd_app_core::identity::client::spire as client;
 
+#[cfg(target_os = "linux")]
 const UNIX_PREFIX: &str = "unix:";
+#[cfg(target_os = "linux")]
 const TONIC_DEFAULT_URI: &str = "http://[::]:50051";
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


linkerd/app/src/spire.rs
```diff
@@ -17,17 +17,27 @@ pub struct Config {
 
 // Connects to SPIRE workload API via Unix Domain Socket
 pub struct Client {
+    #[allow(dead_code)]
```
cratelyn, line 20: can we use `cfg_attr` to only mark this type as `dead_code` for `not(target_os = "linux")`?

linkerd/app/src/spire.rs
```diff
@@ -66,3 +79,21 @@ impl tower::Service<()> for Client {
...
+        &mut self,
+        _cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        unimplemented!("Spire is supported on Linux only")
+    }
+
+    fn call(&mut self, _req: ()) -> Self::Future {
+        unimplemented!("Spire is supported on Linux only")
+    }
+}
```
cratelyn, line 99: this makes me a little nervous. why not avoid a service implementation entirely for non-linux builds, and handle this platform specific behavior at the call site where we use this client?

similarly, for the `From<Config>` implementation above, i'm worried this introduces the risk of runtime panics.

olix0r: AIUI, this is a first step that will setup the named pipe PR. Just a temporary landing spot.

cratelyn: ah, good point. i'll mark this as resolved. :white_check_mark:

zaharidichev: Yes, my desire here is to isolate changes as much as possible to this module. And it is a temporary placeholder change to get us to compile for a windows target. I will follow with a named pipe support shortly after.

linkerd/app/src/spire.rs
zaharidichev: makes sense, will address

cratelyn (approved):
thanks :slightly_smiling_face:

PR #3628, opened 2025-02-20
chore(transport): Ignore TCP_USER_TIMEOUT on non-Linux systems and warn

zaharidichev:
Right now the proxy can be configured so that `TCP_USER_TIMEOUT` is set on connections.
This functionality is not present for non-linux targets. This introduces a warning in case a user
tries to set `TCP_USER_TIMEOUT` on a proxy that has been compiled for a target other than Linux.

### DIFF
--- a/linkerd/proxy/transport/src/lib.rs
+++ b/linkerd/proxy/transport/src/lib.rs
@@ -71,6 +71,7 @@ fn set_keepalive_or_warn(
     tokio::net::TcpStream::from_std(stream)
 }
 
+#[cfg(target_os = "linux")]
 fn set_user_timeout_or_warn(
     tcp: TcpStream,
     user_timeout: Option<Duration>,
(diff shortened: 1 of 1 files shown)


linkerd/proxy/transport/src/lib.rs
```diff
@@ -85,3 +86,14 @@ fn set_user_timeout_or_warn(
...
     tokio::net::TcpStream::from_std(stream)
 }
+
+#[cfg(not(target_os = "linux"))]
+fn set_user_timeout_or_warn(
+    tcp: TcpStream,
+    user_timeout: Option<Duration>,
+) -> io::Result<TcpStream> {
+    if user_timeout.is_some() {
+        tracing::warn!("TCP_USER_TIMEOUT is supported on Linux only.");
```
olix0r, line 96: Eh, won't this print a warning on every single connection? Undesirable.
```suggestion
        tracing::debug!("TCP_USER_TIMEOUT is supported on Linux only.");
```

olix0r (changes requested):

olix0r (approved):

PR #3683, opened 2025-03-03
chore(just): add windows os target

olix0r:
To support cross-compilation to windows, this change adds an 'os' param to the justfile, used in the release to cross-build to x86_64-pc-windows-gnu.

This will produce a binary named 'linkerd2-proxy-v2.999.9-x86_64.exe'.

The proxy does not yet compile on windows, so this is a placeholder for now.

### DIFF
--- a/justfile
+++ b/justfile
@@ -28,26 +28,30 @@ docker-image := docker-repo + ":" + docker-tag
 
 # The architecture name to use for packages. Either 'amd64', 'arm64', or 'arm'.
 arch := "amd64"
+# The OS name to use for packages. Either 'linux'.
+os := "linux"
 
 libc := 'gnu'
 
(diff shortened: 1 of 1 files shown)


sfleen (approved):

zaharidichev (approved):

justfile
```diff
@@ -28,26 +28,30 @@ docker-image := docker-repo + ":" + docker-tag
 
 # The architecture name to use for packages. Either 'amd64', 'arm64', or 'arm'.
 arch := "amd64"
+# The OS name to use for packages. Either 'linux'.
```
zaharidichev, line 31: either linux or windows ?

olix0r, line 31:
```suggestion
# The OS name to use for packages. Either 'linux' or 'windows'.
```

PR #3687, opened 2025-03-03
refactor(app/integration): remove inert `NewService<T>`

cratelyn:
elsewhere in our codebase, we follow a pattern that can be called a "new service". this is a `Service<T>` whose response `S` is itself a `Service<U>`.

new services are often useful for dealing with particular connection semantics, and provide us a way to model a connection that services many requests.

our test server code makes use of a `Svc`, which wraps a reference to a map of uri's and routes. there is an associated `NewSvc` type that does not provide any material benefit. this `NewSvc` type is a `Service<()>` that never exerts backpressure, nor performs any action besides `Arc::clone`ing the map of routes.

this commit golfs down `linkerd_app_integration::server::Server`, by directly cloning the routes into a `Svc(_)`, without the need for polling a future or handling an (impossible) error.

### DIFF
--- a/linkerd/app/integration/src/server.rs
+++ b/linkerd/app/integration/src/server.rs
@@ -193,7 +193,7 @@ impl Server {
             drain.clone(),
             async move {
                 tracing::info!("support server running");
-                let mut new_svc = NewSvc(Arc::new(self.routes));
+                let svc = Svc(Arc::new(self.routes));
                 if let Some(delay) = delay {
                     let _ = listening_tx.take().unwrap().send(());
                     delay.await;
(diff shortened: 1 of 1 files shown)


olix0r (approved):
The more you know: when a `Service<T>::Response: Service<Req>`, we call that a _MakeService<T, Req>_ (not sure if that trait alias still exists somewhere).

Not relevant here. Your change is fine. Just some more terminology context.

cratelyn:
> The more you know: when a `Service<T>::Response: Service<Req>`, we call that a _MakeService<T, Req>_ (not sure if that trait alias still exists somewhere).
> 
> Not relevant here. Your change is fine. Just some more terminology context.

this is helpful background context. i'd seen `linkerd_stack::NewService<T>`, below, but it is useful to see how this connects to [`tower::make::MakeService<T, U>`](https://docs.rs/tower/latest/tower/make/trait.MakeService.html).

https://github.com/linkerd/linkerd2-proxy/blob/main/linkerd/stack/src/new_service.rs#L6-L11

thank you!

PR #3693, opened 2025-03-03
refactor(app/integration): clean up `Server` constructors

cratelyn:
this commit removes some misdirection from the various constructors for our test server.

currently, we expose a family of constructor functions `server::new()`, `server::http1()`, ..., and so forth.

each of these invoke a private `server::Server::http1()`, `server::Server::http2()`, `server::Server::http2_tls()`, ..., counterpart, which then delegates down once more to another private constructor `server::Server::new()`.

this is all a bit roundabout, particularly because these private constructors are not used by any other internal code in the `server` submodule.

this commit removes these inherent `Server` constructors, since they are private and not used by any test code. each free-standing constructor function is altered to instead directly construct a `Server`.

**NB:** based upon #3692.


olix0r (approved):

cratelyn:
rebased upon `main`, now that #3690 and #3692 have landed.

PR #3697, opened 2025-03-04
refactor(app/integration): remove `pb` reëxport

cratelyn:
`linkerd_app_integration::controller` provides a reëxport of the protobuffer types provided by `linkerd2_proxy_api::destination`.

this reëxport is only used in three places externally. for simplicity, this commit demotes this alias to an internally visible `use`, and updates this handful of references to use
`linkerd2_proxy_api::destination` directly.

### DIFF
--- a/linkerd/app/integration/src/controller.rs
+++ b/linkerd/app/integration/src/controller.rs
@@ -1,6 +1,6 @@
 use super::*;
 
-pub use linkerd2_proxy_api::destination as pb;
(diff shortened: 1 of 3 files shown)


cratelyn:
i see this is a bit of a convention in other submodules, so i am going to opt to close this. if we want to step away from this pattern, for which i do not see a strong reason at the moment, we should apply it consistently.

PR #3718, opened 2025-03-07
chore(deps): `tower` is a workspace dependency

cratelyn:
pr #3715 missed a small handful of cargo dependencies. this commit marks these so that they use the workspace-level tower version.


olix0r:
i'd really like to go a different way with this, ultimately: linkerd-stack should be the only crate that depends on tower internally and everything we use in this repo should consume linkerd-stack. Not a blocker, since that's a larger chagne.

PR #3731, opened 2025-03-10
feat(runtime): support CORES_MIN, CORES_MAX, and CORES_MAX_RATIO

olix0r:
When the proxy boots up, it needs to select a number of I/O worker threads to allocate to the runtime. This change adds a new environment variable that allows this value to scale based on the number of CPUs available on on the host.

A CORES_MAX_RATIO value of 1.0 will allocate one worker thread per CPU core. A lesser value will allocate fewer worker threads.

The CORES_MIN value sets a lower bound on the number of worker threads to use. The CORES_MAX value sets an upper bound.

### DIFF
--- a/linkerd2-proxy/src/rt.rs
+++ b/linkerd2-proxy/src/rt.rs
@@ -48,7 +48,7 @@ pub(crate) fn build() -> Runtime {
         .and_then(|v| {
             let opt = v.parse::<f64>().ok().filter(|n| *n > 0.0 && *n <= 1.0);
             if opt.is_none() {
-                warn!(LINKERD2_PROXY_CORES = %v, "Ignoring invalid configuration");
+                warn!(LINKERD2_PROXY_CORES_MAX_RATIO = %v, "Ignoring invalid configuration");
             }
             opt
         });
--- a/linkerd2-proxy/src/rt.rs
+++ b/linkerd2-proxy/src/rt.rs
@@ -1,5 +1,8 @@
+use std::num::NonZeroUsize;
+
+use linkerd_app::Workers;
 use tokio::runtime::{Builder, Runtime};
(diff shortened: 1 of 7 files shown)


linkerd2-proxy/src/rt.rs
```diff
@@ -9,45 +12,74 @@ pub(crate) fn build() -> Runtime {
     //
     // The basic scheduler is used when the threaded scheduler would provide no
     // benefit.
-    let mut cores = std::env::var("LINKERD2_PROXY_CORES")
+
+    let min_cores = std::env::var("LINKERD2_PROXY_CORES_MIN")
```
cratelyn, line 16: now that this environment variable is no longer respected, and we use these `_MIN`/`_MAX`/`_RATIO` variables instead, should this branch be marked as a breaking change?

linkerd2-proxy/src/rt.rs
```diff
@@ -9,45 +12,74 @@ pub(crate) fn build() -> Runtime {
...
+            }
+            opt
+        });
+
+    let cores_ratio = std::env::var("LINKERD2_PROXY_CORES_MAX_RATIO")
+        .ok()
+        .and_then(|v| {
+            let opt = v.parse::<f64>().ok().filter(|n| *n > 0.0 && *n <= 1.0);
+            if opt.is_none() {
+                warn!(LINKERD2_PROXY_CORES = %v, "Ignoring invalid configuration");
```
cratelyn, line 51:
```suggestion
                warn!(LINKERD2_PROXY_CORES_MAX_RATIO = %v, "Ignoring invalid configuration");
```

cratelyn (approved):
one nit about an event field, but this otherwise looks great. nice work!

PR #3761, opened 2025-03-14
chore: use workspace-level package metadata

cratelyn:
this commit adds a `[workspace.package]` table at the root of the cargo workspace. constituent manifests are updated to use the workspace-level metadata.

this is generally a superficial chore, but has a pleasant future upside: when new rust editions are released (e.g. 2024), we will only need to update the edition specified at the root of the workspace.

### DIFF
--- a/linkerd/http/access-log/Cargo.toml
+++ b/linkerd/http/access-log/Cargo.toml
@@ -1,10 +1,10 @@
 [package]
 name = "linkerd-http-access-log"
-version = "0.1.0"
-authors = ["Linkerd Developers <cncf-linkerd-dev@lists.cncf.io>"]
-license = "Apache-2.0"
+version = { workspace = true }
+authors = { workspace = true }
+license = { workspace = true }
 edition = "2018"
-publish = false
+publish = { workspace = true }
 
 [dependencies]
 futures-core = "0.3"
--- a/linkerd/http/detect/Cargo.toml
+++ b/linkerd/http/detect/Cargo.toml
@@ -1,8 +1,10 @@
 [package]
 name = "linkerd-http-detect"
-version = "0.1.0"
-edition = "2021"
-publish = false
+version = { workspace = true }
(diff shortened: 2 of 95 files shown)


linkerd/http/access-log/Cargo.toml
```diff
@@ -1,10 +1,10 @@
...
 name = "linkerd-http-access-log"
-version = "0.1.0"
-authors = ["Linkerd Developers <cncf-linkerd-dev@lists.cncf.io>"]
-license = "Apache-2.0"
+version = { workspace = true }
+authors = { workspace = true }
+license = { workspace = true }
 edition = "2018"
-publish = false
+publish = { workspace = true }
```
cratelyn, line 7: i've been intentional about this explicitly _not_ bumping any crates edition. that'd be an easy follow-on, but i want this change to be strictly a syntactic change.

linkerd/http/detect/Cargo.toml
```diff
@@ -1,8 +1,10 @@
 [package]
 name = "linkerd-http-detect"
-version = "0.1.0"
-edition = "2021"
-publish = false
+version = { workspace = true }
+authors = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
+publish = { workspace = true }
```
cratelyn, line 7: and as we can see, some packages had slightly inconsistent metadata. this helps us apply uniform metadata.

opencensus-proto/Cargo.toml
```diff
@@ -1,10 +1,10 @@
 [package]
 name = "opencensus-proto"
-version = "0.1.0"
+version = { workspace = true }
 authors = ["The OpenCensus Authors"]
```
cratelyn, line 4: another nice thing about this is that it highlights what crates are exceptions to the norm: this for example is vendored.

olix0r (approved):

PR #3769, opened 2025-03-17
chore(just): add .exe extension to windows binaries

zaharidichev:
#3683 introduced the `os` param into our just file. This parameter was used to set the `_package_name` variable but not the `_target_bin` one. As a result of that the `_strip:` task was failing for windows builds as it was incorrectly looking for a binary that was missing (one without an .exe extension).

This PR adds the following changes: 
 - correctly adds the `.exe` extension to `_target_bin` when `os` is windows
 - adds the `os` to the name of the produces artifact

The latter means that release artifacts shall be names as: 

```
linkerd2-proxy-v2.286.0-windows-amd64.tar.gz 
linkerd2-proxy-v2.286.0-linux-amd64.tar.gz 
```

### DIFF
--- a/justfile
+++ b/justfile
@@ -50,8 +50,8 @@ _target := if os + '-' + arch == "linux-amd64" {
 _cargo := 'just-cargo profile=' + profile + ' target=' + _target + ' toolchain=' + toolchain
 
 _target_dir := "target" / _target / profile
-_target_bin := _target_dir / "linkerd2-proxy"
-_package_name := "linkerd2-proxy-" + package_version + "-" + arch + (if libc == 'musl' { '-static' } else { '' }) + (if os == 'windows' { '.exe' } else { '' })
+_target_bin := _target_dir / "linkerd2-proxy" + if os == 'windows' { '.exe' } else { '' }
+_package_name := "linkerd2-proxy-" + package_version + "-" + os + "-" + arch + if libc == 'musl' { '-static' } else { '' }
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -50,8 +50,8 @@ _target := if os + '-' + arch == "linux-amd64" {
 _cargo := 'just-cargo profile=' + profile + ' target=' + _target + ' toolchain=' + toolchain
 
 _target_dir := "target" / _target / profile
-_target_bin := _target_dir / "linkerd2-proxy"
-_package_name := "linkerd2-proxy-" + package_version + "-" + arch + (if libc == 'musl' { '-static' } else { '' }) + (if os == 'windows' { '.exe' } else { '' })
+_target_bin := _target_dir / "linkerd2-proxy" + if os == 'windows' { '.exe' } else { '' }
+_package_name := "linkerd2-proxy-" + package_version + "-" + os + "-" + arch + if libc == 'musl' { '-static' } else { '' }
```
zaharidichev, line 54: I am not too sure about this one. The reason being is that it will change the format of the artifacts produced on release. This will also necessitate changes in any scripts that operate on these artifacts (i.e. `fetch-proxy`). Possible options: 
- go an update the scripts in other repos
- add the os param only for windows artifacts

olix0r: this sounds right to me

olix0r (approved):

PR #3770, opened 2025-03-17
feat(policy): Allow outbound hostname metrics

sfleen:
Outbound hostname metrics were recently disabled. This conditionally re-enables those through a `LINKERD2_PROXY_OUTBOUND_METRICS_HOSTNAME_LABELS` env var, wired through the policy/routing config with the option of individual policies and routes to set this separately from the global config.


linkerd/app/outbound/src/lib.rs
```diff
@@ -151,12 +152,18 @@ impl Outbound<()> {
...
-            })
+        policy::Api::new(
+            workload,
+            limits,
+            Duration::from_secs(10),
+            export_hostname_labels,
+            client,
+        )
+        .into_watch(backoff)
+        .map_result(|response| match response {
```
olix0r, line 163: nit: we tend to use 'res' when handling a one off result (i.e. as compared to rsp for response).

```suggestion
        .map_result(|res| match res {
```

olix0r (approved):
looks great!

PR #3793, opened 2025-03-20
fix(transport): repair IPv6 support

alpeb:
In #3626, we refactored the origin_dst determination logic to utilize socket2 calls. However, this change inadvertently disrupted IPv6 and dual-stack support, causing the server to fail to start when deployed on such network configurations:

```
WARN ThreadId(01) inbound: linkerd_app_core::serve: Server failed to accept connection error=No such file or directory (os error 2)
```

This change reintroduces detection of the current network family, calling socket2's `original_dst()` or `original_dst_ipv6()` depending on the case.

Tested fine in both IPv6 and dual-stack Kind clusters.

### DIFF
--- a/linkerd/proxy/transport/src/orig_dst.rs
+++ b/linkerd/proxy/transport/src/orig_dst.rs
@@ -6,7 +6,7 @@ use futures::prelude::*;
 use linkerd_error::Result;
 use linkerd_io as io;
 use linkerd_stack::Param;
-use std::pin::Pin;
+use std::{net::SocketAddr, pin::Pin};
 use tokio::net::TcpStream;
 
 #[derive(Copy, Clone, Debug, Default)]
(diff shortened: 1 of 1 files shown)


alpeb:
@zaharidichev, tagging you for this—can you test firing it up on Windows?

linkerd/proxy/transport/src/orig_dst.rs
```diff
@@ -83,27 +83,32 @@ where
...
             let addrs = Addrs { inner, orig_dst };
-            Ok((addrs, tcp))
+            Ok((addrs, stream))
         });
 
         Ok((addr, Box::pin(incoming)))
     }
 }
-
-fn orig_dst(sock: TcpStream) -> io::Result<(OrigDstAddr, TcpStream)> {
```
olix0r, line 95: can we preserve the orig_dst helper function rather than inlining the whole thing?

olix0r (approved):

PR #3810, opened 2025-03-26
chore(deps): update to tonic and tonic-build 0.13

cratelyn:

PR #3821, opened 2025-03-27
chore(deps): group grpc dependabot groups

cratelyn:
* https://github.com/linkerd/linkerd2/blob/main/.github/dependabot.yml#L81-L86
* #3809
* https://github.com/linkerd/linkerd2-proxy-api/pull/455#discussion_r2017528014

tonic and prost are interconnected. we should bump them in lockstep.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,6 +25,10 @@ updates:
       futures:
         patterns:
           - "futures*"
+      grpc:
+        patterns:
+          - "prost*"
+          - "tonic*"
       hickory:
         patterns:
           - "hickory*"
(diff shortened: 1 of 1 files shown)


olix0r (approved):

olix0r:
Actually.. doesn't this need to be grouped with linkerd2-proxy-api?

cratelyn:
> Actually.. doesn't this need to be grouped with linkerd2-proxy-api?

yeah.. in practice we'll only want patch updates from these until the proxy api bumps to a new major/minor version. if it's alright with you, i might punt on that here, since protobufs and how we generate/check bindings are a broader can of worms we'll need to think about in #3810. what do you think?

PR #3822, opened 2025-03-27
feat(app,dns): add prometheus metrics to `Dns`

cratelyn:
this commit introduces a new metric family tracking the rate and outcome
of dns lookups made by the linkerd2 proxy. this metric family has two
labels, tracking (a) the control plane client responsible for a
particular lookup, (b) the outcome of the lookup.

this metric is named `control_dns_resolutions_total`.

this metric includes labels to count DNS resolutions for each distinct
control plane client, by record type (A/AAAA or SRV), and by outcome
(success or failure).

this commit generally does this via the addition of some new interfaces
to `linkerd-dns`'s `Resolver` structure. the `resolve_addrs()` method is
extended to increment particular counters if they have been installed.

the `linkerd-app` crate's `Dns` type now encapsulates its resolver, and
callers acquire a new resolver by providing a client name to its
`resolver()` method. this uses the client name to construct label sets
and create the corresponding time series for each client.

once proxies with this patch are running, and the viz extension has been
installed, one can query this metric like so:

**nb:** this screenshot shows an early prototype, this metric has since
been renamed.

linkerd-dns-prometheus-metrics

this promQL query...

```
sum(rate(control_dns_resolutions_total[1m])) by (app,client,result) > 0
```

...will show the per-minute rate of dns lookups/failures across each
application workload, for each control-plane client, for each possible
outcome.

### DIFF
--- a/linkerd/app/core/src/dns.rs
+++ b/linkerd/app/core/src/dns.rs
@@ -24,14 +24,20 @@ pub struct Dns {
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
 struct Labels {
     client: &'static str,
-    outcome: Outcome,
+    record_type: RecordType,
+    result: Outcome,
+}
+
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+enum RecordType {
+    A,
+    Srv,
 }
 
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
 enum Outcome {
-    A,
-    Srv,
-    Failure,
+    Ok,
+    NotFound,
 }
 
 // === impl Dns ===
@@ -39,23 +45,45 @@ enum Outcome {
 impl Dns {
     /// Returns a new [`Resolver`].
     pub fn resolver(&self, client: &'static str) -> Resolver {
-        let Self {
-            resolver,
-            dns_resolutions_total,
-        } = self;
+        let metrics = self.metrics(client);
 
-        let get_counter = |outcome| {
-            dns_resolutions_total
-                .get_or_create(&Labels { client, outcome })
-                .clone()
-        };
-        let metrics = Metrics {
-            a_records_resolved: get_counter(Outcome::A),
-            srv_records_resolved: get_counter(Outcome::Srv),
-            lookups_failed: get_counter(Outcome::Failure),
-        };
+        self.resolver.clone().with_metrics(metrics)
+    }
 
-        resolver.clone().with_metrics(metrics)
+    fn metrics(&self, client: &'static str) -> Metrics {
+        let family = &self.dns_resolutions_total;
+
+        let a_records_resolved = (*family.get_or_create(&Labels {
+            client,
+            record_type: RecordType::A,
+            result: Outcome::Ok,
+        }))
+        .clone();
+        let a_records_not_found = (*family.get_or_create(&Labels {
+            client,
+            record_type: RecordType::A,
+            result: Outcome::NotFound,
+        }))
+        .clone();
+        let srv_records_resolved = (*family.get_or_create(&Labels {
+            client,
+            record_type: RecordType::Srv,
+            result: Outcome::Ok,
+        }))
+        .clone();
+        let srv_records_not_found = (*family.get_or_create(&Labels {
+            client,
+            record_type: RecordType::Srv,
+            result: Outcome::NotFound,
+        }))
+        .clone();
+
+        Metrics {
+            a_records_resolved,
+            a_records_not_found,
+            srv_records_resolved,
+            srv_records_not_found,
+        }
     }
 }
 
@@ -94,10 +122,15 @@ impl ConfigureResolver for Config {
 
 impl EncodeLabelSet for Labels {
     fn encode(&self, mut encoder: LabelSetEncoder<'_>) -> Result<(), std::fmt::Error> {
-        let Self { client, outcome } = self;
+        let Self {
+            client,
+            record_type,
+            result,
+        } = self;
 
         ("client", *client).encode(encoder.encode_label())?;
-        ("outcome", outcome).encode(encoder.encode_label())?;
+        ("record_type", record_type).encode(encoder.encode_label())?;
+        ("result", result).encode(encoder.encode_label())?;
 
         Ok(())
     }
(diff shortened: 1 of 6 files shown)


linkerd/app/core/src/dns.rs
```diff
@@ -31,3 +91,34 @@ impl ConfigureResolver for Config {
...
 }
+
+// === impl Labels ===
+
+impl EncodeLabelSet for Labels {
+    fn encode(&self, mut encoder: LabelSetEncoder<'_>) -> Result<(), std::fmt::Error> {
+        let Self { client, outcome } = self;
+
+        ("client", *client).encode(encoder.encode_label())?;
+        ("outcome", outcome).encode(encoder.encode_label())?;
```
cratelyn, line 102: :bicyclist: :house: open to naming suggestions for these labels!

linkerd/dns/src/lib.rs
```diff
@@ -86,16 +122,30 @@ impl Resolver {
...
                 debug!(srv.error = %srv_error, "Falling back to A/AAAA record lookup");
                 let (ips, delay) = match self.resolve_a_or_aaaa(name).await {
                     Ok(res) => res,
-                    Err(a_error) => return Err(ResolveError { a_error, srv_error }),
+                    Err(a_error) => {
+                        lookups_failed.as_ref().map(Counter::inc);
+                        return Err(ResolveError { a_error, srv_error });
+                    }
                 };
+                a_records_resolved.as_ref().map(Counter::inc);
```
olix0r, line 148: This is a little lossy: it doesn't explicitly count SRV lookups, so it takes some reasoning to understand how many lookups are actually being performed.

We could clarify this by splitting the record type and the result.

```
dns_resolutions_total{client="policy",record_type="SRV",result="ok"} 1
dns_resolutions_total{client="policy",record_type="SRV",result="not_found"} 0
dns_resolutions_total{client="policy",record_type="A/AAAA",result="ok"} 0
dns_resolutions_total{client="policy",record_type="A/AAAA",result="not_found"} 0
dns_resolutions_total{client="destination",record_type="SRV",result="ok"} 0
dns_resolutions_total{client="destination",record_type="SRV",result="not_found"} 1
dns_resolutions_total{client="destination",record_type="A/AAAA",result="ok"} 1
dns_resolutions_total{client="destination",record_type="A/AAAA",result="not_found"} 0
```

This gives us an easy query to measure total request load generated from a client while maintaining visibility into the record types and results.

I've minorly bikeshedded the naming here:

* We are explicitly counting the resolutions, not the results now.
* No benefit for control_ prefix in names (control_ is a metrics scope, the client names are the next component -- it's a bug if the trace_collector client doesn't behave that way).
* now we are simply reporting an ok|not_found result

linkerd/dns/src/lib.rs
```diff
@@ -76,7 +83,36 @@ impl Resolver {
...
+            ..self
+        }
+    }
+
+    /// Installs a counter tracking the number of lookups that failed.
+    pub fn with_lookups_failed_counter(self, counter: Counter) -> Self {
+        Self {
+            lookups_failed: Some(counter),
+            ..self
+        }
```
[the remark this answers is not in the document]
cratelyn: sure thing. 2f4d29ee addresses this point.

linkerd/dns/src/lib.rs
cratelyn: can do! a6632f9c, ff73bca5, 2f4d29ee make changes to address this feedback: the metric is renamed, the labels are changed to match the outline above, and the `control_` prefix is now removed.

linkerd/app/core/src/dns.rs
```diff
@@ -8,16 +17,93 @@ pub struct Config {
...
 
 impl Config {
-    pub fn build(self) -> Dns {
+    pub fn build(self, registry: &mut Registry) -> Dns {
+        let dns_resolutions_total = Family::default();
+        registry.register(
+            "dns_resolutions_total",
+            "Counts the number of DNS records that have been resolved.",
+            dns_resolutions_total.clone(),
+        );
```
olix0r, line 99: I'm fairly sure this will result in a metric called `dns_resolutions_total_total`: counters implicitly have _total appended (but the metric name in the HELP does not include the _total suffix).

Furthermore, as we do elsewhere, we should omit common prefixes like dns_ and set them from the caller.

```suggestion
        registry.register(
            "resolutions",
            "Counts the number of DNS records that have been resolved.",
            dns_resolutions_total.clone(),
        );
```

I'd also probably opt to name the field `resolutions` but that's a pure nit.

linkerd/app/src/lib.rs
```diff
@@ -134,7 +134,7 @@ impl Config {
         let (metrics, report) = Metrics::new(admin.metrics_retain_idle);
 
         debug!("Building DNS client");
-        let dns = dns.build();
+        let dns = dns.build(&mut registry);
```
olix0r, line 137: and then we can use:

`
```suggestion
        let dns = dns.build(registry.sub_registry_with_prefix("control_dns"));
```

This helps to scope the dns client like the other controller related clients.

linkerd/app/core/src/dns.rs
```diff
@@ -1,5 +1,14 @@
+use linkerd_metrics::prom::{
+    encoding::{EncodeLabelSet, EncodeLabelValue, LabelSetEncoder, LabelValueEncoder},
+    Counter, Family, Registry,
+};
+use prometheus_client::encoding::EncodeLabel;
```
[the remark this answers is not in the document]
cratelyn: good catch, thanks! this wasn't intentional, rust-analyzer has trouble choosing the right place to get these symbols, for me at least. c0bc2e9 fixes this.

linkerd/app/core/src/dns.rs
cratelyn: 1437755 addresses this; removing the `_total` suffix and omitting the `dns_` prefix locally in favor of a sub-registry.

linkerd/app/src/lib.rs
cratelyn: see [above](https://github.com/linkerd/linkerd2-proxy/pull/3822#discussion_r2023111280), this is addressed in 1437755.

cratelyn:
> Looking great! Just some last polish requests, and then let's get the description updated before merge :) Thanks!

the description has been updated to reflect the name and label-set for this metric, post-review. review comments above have been addressed, and ci is now passing. :tada:

olix0r (approved):

PR #3833, opened 2025-04-01
chore(proxy/http): use `hyper_util::rt::tokio::TokioExecutor`

cratelyn:
this commit removes the `linkerd-http-executor` crate, and replaces all usage of its `TracingExecutor` type with the `TokioExecutor` type provided by `hyper-util`.

this work is based upon hyperium/hyper-util#166. that change, included in the 0.1.11 release, altered the `TokioExecutor` type so that it propagates tracing context when the `tracing` feature is enabled.

with that change made, our `TracingExecutor` type is now redundant.

* https://github.com/hyperium/hyper-util/pull/166
* https://github.com/hyperium/hyper-util/blob/master/CHANGELOG.md#0111-2025-03-31


Cargo.toml
```diff
@@ -131,5 +129,10 @@ version = "0.1.3"
 default-features = false
 features = ["channel"]
 
+[workspace.dependencies.hyper-util]
+version = "0.1"
+default-features = false
+features = ["tokio", "tracing"]
```
cratelyn, line 135: NB: https://github.com/hyperium/hyper-util/pull/166/files#diff-bdb64647cd31ba75babe48b3287899e6116a0a8c8267624ba227f86df5937467

to prevent any footguns wherein we might neglect to set this feature flag, and inadvertently fail to propagate trace context, i've set this at the workspace level.

linkerd/http/executor/src/lib.rs
cratelyn, line 1: :wave: see ya!

linkerd/proxy/http/src/lib.rs
```diff
@@ -34,12 +34,12 @@ pub use http::{
     uri, Method, Request, Response, StatusCode,
 };
 pub use http_body::Body;
+pub use hyper_util::rt::tokio::TokioExecutor;
 pub use linkerd_http_box::{BoxBody, BoxRequest, BoxResponse, EraseResponse};
 pub use linkerd_http_classify as classify;
 pub use linkerd_http_detect::{
     DetectMetrics, DetectMetricsFamilies, DetectParams, Detection, NewDetect,
 };
-pub use linkerd_http_executor::TracingExecutor;
```
cratelyn, line 42: i've reexported this from `linkerd-proxy-http`, as we did for the previous executor.

i took the time to update callers to use the new name, rather than doing this under an alias. vim macros made this a straight-forward bit of work.

olix0r (approved):

PR #3834, opened 2025-04-01
chore(deps): define `tracing` workspace dependency

cratelyn:
this commit hoists `tracing`, used liberally throughout our project, such that it is managed as a single workspace dependency.

this will be helpful someday when a 0.2 release happens.


olix0r:
I think we also want to update https://github.com/linkerd/linkerd2-proxy/blob/686934cec691744affb10d3e246211be6ca0d88d/linkerd2-proxy/Cargo.toml#L29

cratelyn:
> I think we also want to update
> 
> https://github.com/linkerd/linkerd2-proxy/blob/686934cec691744affb10d3e246211be6ca0d88d/linkerd2-proxy/Cargo.toml#L29

indeed! thanks for spotting that. :pray:

cratelyn:
fixed the note above, rebased on `main` to address conflicts with #3833.

olix0r (approved):

cratelyn:
conflicts with #3887 addressed. :recycle:

PR #3883, opened 2025-04-24
Add aws-lc-rs as optional rustls backend

sfleen:
This has a few benefits. Primarily this gives us a reasonable path to creating FIPS-enabled builds on architectures other than x86-64, as well as a path away from using BoringSSL as a backend.

Additionally, rustls has been using the aws-lc-rs library as the default backend for a little while now, so this gives us the opportunity to stay in line with the most widely used option in the ecosystem.

### DIFF
--- a/linkerd/meshtls/Cargo.toml
+++ b/linkerd/meshtls/Cargo.toml
@@ -8,6 +8,9 @@ publish = { workspace = true }
 
 [features]
 rustls = ["linkerd-meshtls-rustls", "__has_any_tls_impls"]
+rustls-aws-lc = ["rustls", "linkerd-meshtls-rustls/aws-lc"]
+rustls-aws-lc-fips = ["rustls-aws-lc", "linkerd-meshtls-rustls/aws-lc-fips"]
+rustls-ring = ["rustls", "linkerd-meshtls-rustls/ring"]
 boring = ["linkerd-meshtls-boring", "__has_any_tls_impls"]
 boring-fips = ["boring", "linkerd-meshtls-boring/fips"]
 # Enabled if *any* TLS impl is enabled.
--- a/linkerd/meshtls/rustls/src/backend.rs
+++ b/linkerd/meshtls/rustls/src/backend.rs
@@ -1,0 +1,11 @@
+#[cfg(feature = "aws-lc")]
+mod aws_lc;
+#[cfg(feature = "ring")]
+mod ring;
+
+#[cfg(feature = "aws-lc")]
+pub use aws_lc::{default_provider, SUPPORTED_SIG_ALGS, TLS_SUPPORTED_CIPHERSUITES};
+#[cfg(all(not(feature = "aws-lc"), feature = "ring"))]
+pub use ring::{default_provider, SUPPORTED_SIG_ALGS, TLS_SUPPORTED_CIPHERSUITES};
+#[cfg(all(not(feature = "aws-lc"), not(feature = "ring")))]
+compile_error!("No rustls backend enabled. Enabled one of the \"ring\" or \"aws-lc\" features");
--- a/linkerd/meshtls/rustls/src/creds.rs
+++ b/linkerd/meshtls/rustls/src/creds.rs
@@ -2,6 +2,8 @@ mod receiver;
 mod store;
 pub(crate) mod verify;
 
+use crate::backend;
+
 pub use self::{receiver::Receiver, store::Store};
 use linkerd_dns_name as dns;
 use linkerd_error::Result;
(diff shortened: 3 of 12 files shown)


cratelyn:
it looks like ci is not passing on this branch, due to `cargo-deny`.

linkerd/meshtls/Cargo.toml
```diff
@@ -8,6 +8,9 @@ publish = { workspace = true }
 
 [features]
 rustls = ["linkerd-meshtls-rustls", "__has_any_tls_impls"]
+rustls-ring = ["rustls", "linkerd-meshtls-rustls/ring"]
+rustls-aws-lc = ["rustls", "linkerd-meshtls-rustls/aws-lc"]
+rustls-aws-lc-fips = ["rustls-aws-lc", "linkerd-meshtls-rustls/aws-lc-fips"]
```
cratelyn, line 13:
```suggestion
rustls-aws-lc = ["rustls", "linkerd-meshtls-rustls/aws-lc"]
rustls-aws-lc-fips = ["rustls-aws-lc", "linkerd-meshtls-rustls/aws-lc-fips"]
rustls-ring = ["rustls", "linkerd-meshtls-rustls/ring"]
```

nit, can we alphabetize these? same for the equivalent list in `linkerd/meshtls/rustls/Cargo.toml`.

linkerd/meshtls/rustls/src/backend.rs
```diff
@@ -0,0 +1,11 @@
...
+mod aws_lc;
+#[cfg(feature = "ring")]
+mod ring;
+
+#[cfg(feature = "aws-lc")]
+pub use aws_lc::{default_provider, SUPPORTED_SIG_ALGS, TLS_SUPPORTED_CIPHERSUITES};
+#[cfg(all(not(feature = "aws-lc"), feature = "ring"))]
+pub use ring::{default_provider, SUPPORTED_SIG_ALGS, TLS_SUPPORTED_CIPHERSUITES};
+#[cfg(all(not(feature = "aws-lc"), not(feature = "ring")))]
+compile_error!("No rustls backend enabled. Enabled one of the \"ring\" or \"aws-lc\" features");
```
cratelyn, line 11: is there a situation where we would want both backends enabled? does that make any sense, semantically?

if not, perhaps it would make sense to additionally emit a `compile_error!` when both flags are enabled:

```rust
#[cfg(all(feature = "aws-lc", feature = "ring"))]
compile_error!("Multiple rustls backends enabled. Enabled only one of the \"ring\" or \"aws-lc\" features");
```

otherwise, i can imagine what would be a very unfortunate failure mode: someone sets `--features ring` when building this crate, but inadvertently uses the `aws-lc` provider due to the semantics above, where `aws-lc` takes precedent when both flags are set.

linkerd/meshtls/rustls/src/creds.rs
```diff
@@ -121,51 +124,8 @@ mod params {
         rustls::SignatureScheme::ECDSA_NISTP256_SHA256;
     pub const SIGNATURE_ALG_RUSTLS_ALGORITHM: rustls::SignatureAlgorithm =
         rustls::SignatureAlgorithm::ECDSA;
-    // A subset of the algorithms supported by rustls+ring, imported from
-    // https://github.com/rustls/rustls/blob/v/0.23.21/rustls/src/crypto/ring/mod.rs#L107
```
cratelyn, line 125: can we preserve this comment? or ideally, document what these mappings represent, now that we have multiple.

linkerd/meshtls/Cargo.toml
cratelyn: ...and for `linkerd2-proxy/Cargo.toml`!

cratelyn (commented):
:dog: :speech_balloon:

linkerd/meshtls/rustls/src/backend.rs
sfleen: I had originally considered that, but I ended up ultimately deciding against it because of the slightly annoying way that cargo features should be purely additive. This also mirrors how the boringssl/rustls are currently handled, where it defaults to one or the other if both are enabled.

linkerd/meshtls/rustls/src/creds.rs
sfleen: Re-added this to the rustls backend.

linkerd/meshtls/Cargo.toml
sfleen: Done

linkerd/meshtls/rustls/src/backend.rs
cratelyn: for posterity, see: https://github.com/linkerd/linkerd2-proxy/blob/b4fb4277d46ac72ffc08d57e2d806fa90de7e270/linkerd/meshtls/src/lib.rs#L61-L81

sounds good!

cratelyn (approved):

cratelyn (commented):
this looks good, besides the need to satiate cargo deny.

sfleen:
cargo deny should be sorted out now, wanted to make sure there were no more structural changes i had to make first.

cratelyn (approved):
this looks great!

PR #3903, opened 2025-05-06
feat(proxy/http): recover abandoned http requests

cratelyn:
:construction: :construction: :construction:


cratelyn:
leaving a note here:

i'm driving some work upstream to help us adopt `try_send_request`. see https://github.com/hyperium/hyper/pull/3883#issuecomment-2860212062, https://github.com/hyperium/hyper/pull/3922, and https://github.com/hyperium/hyper/pull/3892.

there are still some open design questions about what this should look like, and on our end it will require some thought to figure out how we should dispatch recovered requests within our `tower`-oriented architecture.

PR #3988, opened 2025-07-01
refactor: `FmtLabels` impls use exhaustive bindings

cratelyn:
this is based on #3987.

### DIFF
--- a/linkerd/app/core/src/metrics.rs
+++ b/linkerd/app/core/src/metrics.rs
@@ -525,7 +525,13 @@ impl StackLabels {
 
 impl FmtLabels for StackLabels {
     fn fmt_labels(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.direction.fmt_labels(f)?;
-        write!(f, ",protocol=\"{}\",name=\"{}\"", self.protocol, self.name)
+        let Self {
+            direction,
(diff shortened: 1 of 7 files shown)


linkerd/app/core/src/metrics.rs
```diff
@@ -409,16 +427,25 @@ impl svc::Param<OutboundZoneLocality> for OutboundEndpointLabels {
 
 impl FmtLabels for OutboundEndpointLabels {
     fn fmt_labels(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        if let Some(a) = self.authority.as_ref() {
+        let Self {
+            server_id,
+            authority,
+            labels,
+            // TODO(kate): this label is not currently emitted.
+            zone_locality: _,
```
cratelyn, line 435: in the process of writing this pull request, i found this:

this label, `zone_locality`, is not included in our outbound endpoint labels. this was added in #3262. i've refrained from adding it here, to avoid changing any behavior. should we file a ticket for this, and is this elision intentional?

olix0r: There's at least no _harm_ in omitting it, since it not at a higher cardinality than the target_addr

olix0r (approved):

PR #4008, opened 2025-07-17
chore(ci): enable overriding the runner in workflows

olix0r:
We use the ubuntu-24.04 runner by default, but in forks this may not be appropriate. This change updates the runners to support overriding via the LINKERD2_PROXY_RUNNER variable.


cratelyn (commented):
:heavy_check_mark: belatedly, approving this as well. a hearty thank you for doing this.

PR #4020, opened 2025-07-25
updating metrics descriptions

joedrf:
- Removed unnecessary `the` from each description
- Swapped `were` to `was`


cratelyn (approved):
nice! thank you for noticing this, and for taking the time to fix this typo.

we really appreciate it, @joedrf. :bouquet:

PR #4025, opened 2025-07-28
fix(tracing): Remove opentelemetry vendoring

sfleen:
We used to vendor the opentelemetry protobufs because we relied on a version of tonic that `opentelemetry-sdk` didn't support. Now that we've upgraded tonic, we can remove the vendoring and use the official APIs.

Note that we had to downgrade opentelemetry from 0.30 to 0.29, once we update to tonic 0.13 we can upgrade back.


olix0r:
@sfleen does this put us in a position where we are unable to take updates? would it be better to drive the tonic update first?

olix0r (approved):

PR #4029, opened 2025-07-29
feat(tls): Remove ring as crypto backend

sfleen:
The broader ecosystem has mostly moved to aws-lc-rs as the primary rustls backend, and we should follow suit. This will also simplify the maintenance of the proxy's TLS implementation in the long term.

There will need to be some refactoring to clean up the rustls provider interfaces, but that will come in follow-ups.


.github/workflows/release.yml
```diff
@@ -145,11 +145,22 @@ jobs:
     env:
       LINKERD2_PROXY_VENDOR: ${{ github.repository_owner }}
       LINKERD2_PROXY_VERSION: ${{ needs.meta.outputs.version }}
+      # TODO: add to dev image
+      AWS_LC_SYS_CFLAGS_aarch64_unknown_linux_gnu: "-fuse-ld=/usr/aarch64-linux-gnu/bin/ld"
+      AWS_LC_SYS_CFLAGS_aarch64_unknown_linux_musl: "-fuse-ld=/usr/aarch64-linux-gnu/bin/ld"
```
cratelyn, line 150: my understanding is this is done in https://github.com/linkerd/dev/pull/67?

should we wait for that to land and be released in v48?

sfleen: I think having these in the dev containers is a nice-to-have but not a necessity, especially since we can unblock the work here relatively simply.

sfleen: Good idea, added

deny.toml
```diff
@@ -55,6 +42,8 @@ deny = [
     { name = "rustls", wrappers = ["tokio-rustls"] },
     # rustls-webpki should be used instead.
     { name = "webpki" },
+    # aws-lc-rs should be used instead.
+    { name = "ring" }
```
cratelyn, line 46: :+1: nice! appreciate you doing this.

cratelyn (commented):
:heavy_check_mark: it seems that builds aren't working right now, likely related to the conditional compilation changes made to our meshtls crates in this diff.

i'm content with this change, broadly speaking, though! :heavy_check_mark:

linkerd2-proxy/Cargo.toml
```diff
@@ -8,12 +8,11 @@ publish = { workspace = true }
 description = "The main proxy executable"
 
 [features]
-default = ["meshtls-rustls-aws-lc"]
+default = ["meshtls-rustls"]
 meshtls-boring = ["linkerd-meshtls/boring"]
 meshtls-boring-fips = ["linkerd-meshtls/boring-fips"]
-meshtls-rustls-aws-lc = ["linkerd-meshtls/rustls-aws-lc"]
-meshtls-rustls-aws-lc-fips = ["linkerd-meshtls/rustls-aws-lc-fips"]
-meshtls-rustls-ring = ["linkerd-meshtls/rustls-ring"]
```
olix0r, line 16: I'd like to avoid changing any of the existing aws-lc feature names in this PR and keep this PR strictly to removing ring.

sfleen: Done

cratelyn:
:tada: so excited to see this land!

PR #4032, opened 2025-07-29
chore(deps): update from tokio 1.45 to 1.47

cratelyn:
based on #4031.

supercedes #4027.

bumps [tokio](https://github.com/tokio-rs/tokio) from 1.45.0 to 1.47.0.
- [release notes](https://github.com/tokio-rs/tokio/releases)
- [commits](https://github.com/tokio-rs/tokio/compare/tokio-1.45.0...tokio-1.47.0)

### DIFF
--- a/linkerd/app/integration/src/tests/telemetry/tcp_errors.rs
+++ b/linkerd/app/integration/src/tests/telemetry/tcp_errors.rs
@@ -169,44 +169,6 @@ async fn inbound_success() {
     metric.assert_in(&metrics).await;
 }
 
-/// Tests both of the above cases together.
-#[tokio::test]
-async fn inbound_multi() {
-    let _trace = trace_init();
-
-    let (proxy, metrics) = Test::default().run().await;
-    let client = crate::tcp::client(proxy.inbound);
-
-    let metric = metric(&proxy);
-    let timeout_metric = metric.clone().label("error", "tls detection timeout");
-    let io_metric = metric.label("error", "i/o");
-
-    let tcp_client = client.connect().await;
-
-    tokio::time::sleep(TIMEOUT + Duration::from_millis(15)) // just in case
-        .await;
-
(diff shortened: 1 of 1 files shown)


deny.toml
```diff
@@ -76,6 +76,8 @@ skip-tree = [
     { name = "pprof" },
     # aws-lc-rs uses a slightly outdated version of bindgen
     { name = "bindgen", version = "0.69.5" },
+    # socket v0.6 is still propagating through the ecosystem
+    { name = "socket2", version = "0.5" },
```
cratelyn, line 80: pointing this out :point_up_2: tokio 1.47 depends upon socket2@v0.6, so we must note that duplicate socket2@v0.5 dependencies should be ignored by `cargo deny` while this propagates throughout the rest of our dependency tree.

linkerd/app/integration/src/tests/telemetry/tcp_errors.rs
```diff
@@ -124,26 +124,6 @@ async fn inbound_timeout() {
...
-
-    tcp_client.write(TcpFixture::HELLO_MSG).await;
-    drop(tcp_client);
-
-    metric(&proxy)
-        .label("error", "i/o")
-        .value(1u64)
-        .assert_in(&metrics)
-        .await;
-}
```
cratelyn, line 145: this no longer returns an error. we've removed it, and the `inbound_multi` test below which is redundant with this test.

cratelyn (commented):
:bread: breadcrumbs...

linkerd/app/integration/src/tests/telemetry/tcp_errors.rs
sfleen: Should we still have a test that increments on an IO error in general? It looks like this mechanism doesn't work anymore but it still seems reasonable to have some test for IO error metrics.

PR #4033, opened 2025-07-29
feat(tls): Include rustls crypto provider metrics

sfleen:
This includes a small set of metrics about the currently installed rustls crypto provider and the algorithms it is configured to use.

We don't have 100% assurance that a default crypto provider has been installed before registering the metric, but in local testing it never appeared to be a problem. When we refactor the rustls initialization we can add an extra guarantee that we've initialized it by this point.

Example metric:
```
# HELP rustls_info Proxy TLS info.
# TYPE rustls_info gauge
rustls_info{tls_suites="TLS13_AES_128_GCM_SHA256,TLS13_AES_256_GCM_SHA384,TLS13_CHACHA20_POLY1305_SHA256,",tls_kx_groups="X25519,secp256r1,secp384r1,X25519MLKEM768,",tls_rand="AwsLcRs",tls_key_provider="AwsLcRs",tls_fips="false"} 1
```

### DIFF
--- a/linkerd/app/core/src/tls_info.rs
+++ b/linkerd/app/core/src/tls_info.rs
@@ -1,0 +1,76 @@
+use linkerd_metrics::prom;
+use prometheus_client::encoding::{EncodeLabelSet, EncodeLabelValue, LabelValueEncoder};
+use std::{
+    fmt::{Error, Write},
+    sync::{Arc, OnceLock},
+};
+use tracing::error;
+
+static TLS_INFO: OnceLock<Arc<TlsInfo>> = OnceLock::new();
+
+#[derive(Clone, Debug, Default, Hash, PartialEq, Eq, EncodeLabelSet)]
+pub struct TlsInfo {
+    tls_suites: MetricValueList,
+    tls_kx_groups: MetricValueList,
+    tls_rand: String,
+    tls_key_provider: String,
+    tls_fips: bool,
+}
+
+#[derive(Clone, Debug, Default, Hash, PartialEq, Eq)]
+struct MetricValueList {
+    values: Vec<&'static str>,
+}
+
+impl FromIterator<&'static str> for MetricValueList {
+    fn from_iter<T: IntoIterator<Item = &'static str>>(iter: T) -> Self {
+        MetricValueList {
+            values: iter.into_iter().collect(),
+        }
+    }
+}
+
+impl EncodeLabelValue for MetricValueList {
+    fn encode(&self, encoder: &mut LabelValueEncoder<'_>) -> Result<(), Error> {
+        for value in &self.values {
+            value.encode(encoder)?;
+            encoder.write_char(',')?;
+        }
+        Ok(())
+    }
+}
+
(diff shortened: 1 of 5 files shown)


linkerd/app/core/src/tls_info.rs
```diff
@@ -0,0 +1,83 @@
...
+    if let Some(tls_info) = TLS_INFO.get() {
+        let _ = fam.get_or_create(tls_info);
+        return fam;
+    }
+
+    let Some(provider) = tokio_rustls::rustls::crypto::CryptoProvider::get_default() else {
+        // If the crypto provider hasn't been initialized, we return the metrics family with an
+        // empty set of metrics.
+        return fam;
+    };
```
cratelyn, line 54: i'm of two minds on this: returning here seems like a conservative choice, but looking at the call-site gives me some concern.

if the crypto provider hasn't been initialized, is that a bug? we don't seem to call this repeatedly, so if the provider is set up after we call this function, we wouldn't report on it correctly.

linkerd/app/core/src/tls_info.rs
```diff
@@ -0,0 +1,83 @@
...
+        .flat_map(|cipher_suite| cipher_suite.suite().as_str())
+        .collect::<MetricValueList>();
+    let kx_groups = provider
+        .kx_groups
+        .iter()
+        .flat_map(|suite| suite.name().as_str())
+        .collect::<MetricValueList>();
+    let info = Arc::new(TlsInfo {
+        tls_suites: suites,
+        tls_kx_groups: kx_groups,
```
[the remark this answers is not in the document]
sfleen: Done.

linkerd/app/core/src/tls_info.rs
```diff
@@ -0,0 +1,83 @@
...
+        tls_rand: format!("{:?}", provider.secure_random),
+        tls_key_provider: format!("{:?}", provider.key_provider),
+        tls_fips: provider.fips(),
+    });
+
+    info!("rustls info: {info:?}");
+
+    let info = match TLS_INFO.set(info.clone()) {
+        Ok(_) => info,
+        Err(info) => info,
```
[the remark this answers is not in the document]
sfleen: Looking at this I realized that I'm actually doing a worse implementation of `OnceLock::get_or_init()`, so I've replaced it with that. I think that obviates the need for some of the logging here.

linkerd/app/core/src/tls_info.rs
cratelyn: ah! i'll register my disposition that i have a somewhat strong preference for returning an error here, but i shan't block on it. an error log gets us an indicator of some sort, which is my primary concern :bowing_woman:

cratelyn (approved):
:sparkles:

linkerd/app/core/src/tls_info.rs
cratelyn: (soft request to file an issue tracking this follow-on should we proceed to merge as-is!)

PR #4038, opened 2025-08-01
feat(tls): Remove boring as a TLS implementation

sfleen:
BoringSSL, as we use it today, doesn't integrate well with the broader rustls ecosystem, so this removes it. This will also simplify the maintenance of the proxy's TLS implementation in the long term.

There will need to be some refactoring to clean up the rustls provider interfaces, but that will come in follow-ups.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1433,13 +1433,10 @@ dependencies = [
  "http-body",
  "http-body-util",
  "hyper",
- "hyper-util",
  "ipnet",
  "linkerd-addr",
  "linkerd-conditional",
  "linkerd-dns",
- "linkerd-duplex",
- "linkerd-errno",
  "linkerd-error",
  "linkerd-error-respond",
  "linkerd-exp-backoff",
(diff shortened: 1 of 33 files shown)


linkerd2-proxy/Cargo.toml
```diff
@@ -8,12 +8,8 @@ publish = { workspace = true }
 description = "The main proxy executable"
 
 [features]
-default = ["meshtls-rustls-aws-lc"]
-meshtls-boring = ["linkerd-meshtls/boring"]
-meshtls-boring-fips = ["linkerd-meshtls/boring-fips"]
-meshtls-rustls-aws-lc = ["linkerd-meshtls/rustls-aws-lc"]
-meshtls-rustls-aws-lc-fips = ["linkerd-meshtls/rustls-aws-lc-fips"]
-meshtls-rustls-ring = ["linkerd-meshtls/rustls-ring"]
```
olix0r, line 16: it looks like this PR should be based on the branch that removes ring... i'd make the same recommendation that this leaves the existing aws-featuers as they are.

PR #4064, opened 2025-08-13
fix(outbound): fix clientpolicy not work after remove service profile

wgjak47:
This is a PR for fixing https://github.com/linkerd/linkerd2/issues/14274. In linkerd/idle-cache/src/lib.rs get_or_insert_with function, seems the sidercar with service profile and sidercar with client policy has the same key, which cause the cache didn't expire after we delete serviceprofile and create the new httproute for the target sevice for pods that keep requesting the target service. this pr will create different keys for service profile httpsidecar and client policy httpsidecar to fix this problem.


olix0r:
@wgjak47 thanks for submitting this. i should have some bandwidth to review next week.

PR #4065, opened 2025-08-13
chore(deps): Manually remove unused dependencies

sfleen:
This was created by running `$ just clippy -- -D unused_crate_dependencies`, removing any errors that appeared, and checking that the proxy still compiled with that dependency removed.

Ideally, this would be a lint we can leave on and not have to do this kind of cleanup in the future. However, none of the tools in the ecosystem are satisfactory. The `unused_crate_dependencies` lint has a lot of false positives for crates with integration tests, and the most common tool, [cargo-udeps](https://github.com/est31/cargo-udeps), requires a nightly compiler.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1514,13 +1514,10 @@ dependencies = [
  "http-body",
  "http-body-util",
  "hyper",
- "hyper-util",
  "ipnet",
  "linkerd-addr",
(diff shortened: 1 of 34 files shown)


cratelyn (approved):
this is fantastic. thank you for doing this!

PR #4066, opened 2025-08-13
chore(deps): Update tonic deps to 0.13

sfleen:


Cargo.toml
```diff
@@ -135,4 +135,5 @@ default-features = false
 features = ["tokio", "tracing"]
 
 [workspace.dependencies.linkerd2-proxy-api]
-version = "0.16.0"
+git = "https://github.com/linkerd/linkerd2-proxy-api"
+branch = "sfleen/tonic"
```
sfleen, line 139: This will need to be changed to the latest release of the proxy API once we cut a new one

PR #4127, opened 2025-09-02
feat(app/inbound): count inbound requests

cratelyn:
this branch introduces a telemetry layer to the inbound proxy's connection-level stack.

when running emojivoto, and port-forwarding to the voting service, we
can see a labeled metric shaped like so:
    
```text
 curl localhost:4191/metrics --silent | rg 'inbound_requests'
 # HELP inbound_requests The total number of requests dispatched.
 # TYPE inbound_requests counter
 inbound_requests_total{parent_group="",parent_kind="default",parent_name="all-unauthenticated",parent_port="8080"} 148
```

### DIFF
--- a/linkerd/app/inbound/src/http/router.rs
+++ b/linkerd/app/inbound/src/http/router.rs
@@ -1,7 +1,4 @@
-use crate::{
-    policy::{self, Permitted},
-    stack_labels, Inbound,
-};
+use crate::{policy, stack_labels, Inbound};
 use linkerd_app_core::{
     classify, errors, http_tracing, profiles,
     proxy::{http, tap},
@@ -268,10 +265,8 @@ where
     T: Param<tls::ConditionalServerTls>,
 {
     fn from(permitted: &'a policy::Permitted<T>) -> Self {
-        let (permit, target) = match permitted {
-            Permitted::Grpc { permit, target } => (permit, target),
-            Permitted::Http { permit, target } => (permit, target),
-        };
+        let target = permitted.target_ref();
+        let permit = permitted.permit_ref();
 
         let labels = [
             ("srv", &permit.labels.route.server.0),


linkerd/app/inbound/src/http/router/metrics.rs
```diff
@@ -0,0 +1,115 @@
...
+        family.metrics(&RequestCountLabels {
+            protocol,
+            server,
+            server_addr: target.param(),
+            tls: target.param(),
+        })
+    }
+}
+
+// === impl Labels ===
```
cratelyn, line 62: todo: fix this comment

cratelyn: done.

linkerd/app/inbound/src/http/router/metrics.rs
```diff
@@ -0,0 +1,115 @@
...
+            tls,
+            protocol,
+        } = self;
+
+        ("parent_group", meta.group()).encode(enc.encode_label())?;
+        ("parent_kind", meta.kind()).encode(enc.encode_label())?;
+        ("parent_name", meta.name()).encode(enc.encode_label())?;
+        ("parent_port", *port).encode(enc.encode_label())?;
+
+        ("server_addr", server_addr.ip().to_string()).encode(enc.encode_label())?;
```
cratelyn, line 80: should this possibly be called `parent_addr`?

linkerd/app/inbound/src/http/router/metrics.rs
```diff
@@ -0,0 +1,115 @@
...
+
+        (
+            "protocol",
+            match protocol {
```
olix0r, line 84: I don't think this is appropriate here...

What would it mean to have a RequestCount with TLS? Or Detect?

If we're going to document a protocol here, it should reflect the application protocol specifically. It's probably better to omit this.

linkerd/app/inbound/src/http/router/metrics.rs
```diff
@@ -0,0 +1,98 @@
...
+            Conditional::Some(ServerTlsLabels::Passthru { sni: _ }) => {
+                ("tls", "passthru").encode(enc.encode_label())?;
+            }
+            Conditional::Some(ServerTlsLabels::Established { client_id: _ }) => {
+                ("tls", "established").encode(enc.encode_label())?;
+            }
+            Conditional::None(why) => {
+                ("tls", why.to_string()).encode(enc.encode_label())?;
+            }
+        }
```
olix0r, line 88: We should probably be careful about how this overlaps with the existing TLS labels (I don't think the pre-existing label values are particularly valuable here.


Brainstorming... Should this be a:

`meshtls_client_id="..."` with a value that is optionally set with the client id?

linkerd/app/inbound/src/http/router/metrics.rs
```diff
@@ -0,0 +1,115 @@
...
+        )
+            .encode(enc.encode_label())?;
+
+        match tls.as_ref().map::<_, NoServerTls, _>(ServerTls::labels) {
+            Conditional::Some(ServerTlsLabels::Passthru { sni: _ }) => {
+                ("tls", "passthru").encode(enc.encode_label())?;
+            }
+            Conditional::Some(ServerTlsLabels::Established { client_id: _ }) => {
+                ("tls", "established").encode(enc.encode_label())?;
+            }
```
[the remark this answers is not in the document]
cratelyn: see discussion below. resolved :white_check_mark:

linkerd/app/inbound/src/http/router/metrics.rs
cratelyn: this label is removed in bedef13.

cratelyn: marking this conversation as resolved, now that these labels are removed.

cratelyn:
re-requested a review before i merge this :mage_woman: :mag_right:

PR #4165, opened 2025-09-16
feat(app/inbound): response body frame size metrics

cratelyn:
https://github.com/linkerd/linkerd2-proxy/pull/3308 introduced tower
middleware to the outbound proxy that records the size and number of
response body's data frames.

this commit introduces an equivalent layer to the inbound proxy's
prometheus metrics.

this is based on https://github.com/linkerd/linkerd2-proxy/pull/4166.


cratelyn:
force-pushed atop `main` now that #4166 has merged.

PR #4166, opened 2025-09-16
refactor(app/inbound): metrics layer accepts `InboundMetrics`

cratelyn:
this commit makes a small refactor to the inbound proxy's `metrics::layer()` middleware in the interest of future-proofing it before additional metrics layers are introduced.

this will mean that the call site in `http/router.rs` will not need to be updated repeatedly as we introduce request/response body frame size metrics, duration histograms, and so forth.

### DIFF
--- a/linkerd/app/inbound/src/http/router.rs
+++ b/linkerd/app/inbound/src/http/router.rs
@@ -245,7 +245,7 @@ impl<C> Inbound<C> {
                 .push(svc::NewOneshotRoute::layer_via(|t: &policy::Permitted<T>| {
                     LogicalPerRequest::from(t)
                 }))
-                .push(self::metrics::layer(rt.metrics.request_count.clone()))
+                .push(self::metrics::layer(&rt.metrics))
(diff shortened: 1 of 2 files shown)


linkerd/app/inbound/src/http/router/metrics.rs
cratelyn, line 1: a unified diff for this change looks terrible, recommend viewing this side-by-side.

linkerd/app/inbound/src/http/router.rs
```diff
@@ -245,7 +245,7 @@ impl<C> Inbound<C> {
                 .push(svc::NewOneshotRoute::layer_via(|t: &policy::Permitted<T>| {
                     LogicalPerRequest::from(t)
                 }))
-                .push(self::metrics::layer(rt.metrics.request_count.clone()))
+                .push(self::metrics::layer(&rt.metrics))
```
cratelyn, line 248: this is the point: pass in `rt.metrics` so that we don't have to keep adding a new parameter each time that we add a new layer in the future.

PR #4168, opened 2025-09-17
chore(deps): Clean up cargo deny exemptions

sfleen:
The `linkerd-trace-context` package unecessarily used a few outdated dependencies that were keeping some other dependencies locked to old versions.

This updates the following dependencies in `linkerd-trace-context` that were already updated in other packages (or unused in general):
- `rand` 0.8 -> 0.9
- `thiserror` 1 -> 2
- `base64` 0.13 -> 0.22

The associated `deny.toml` entries have also been removed outright or updated to a narrower exemption.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -259,12 +259,6 @@ dependencies = [
 
 [[package]]
 name = "base64"
-version = "0.13.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


sfleen:
Deprecation warnings are sorted out, I forgot that `base64` had changed their API into something pretty rough to use, so I may have a follow-up that swaps it out for something like `data_encoding` which has a more ergonomic API.

cratelyn (approved):

PR #4174, opened 2025-09-19
refactor(http/prom): `request::NewRecordBodyData` is metrics agnostic

cratelyn:
this commit makes a mechanical refactor in the interest of paving ground towards using the request body metrics middleware in the inbound proxy.

because of minor differences in how we represent protocols like gRPC in the inbound and outbound proxies, we must account for:

(a): distinct `NewService<T>`'s accepting `Grpc<T>` and `Http<T>` targets, respectively, each holding a single metrics `Family<L>`.

(b): a single `NewService<T>` that matches on the `Permitted<T>` enum to select one of two metrics `Family<L>`s.

see linkerd/linkerd2-proxy#4119 for more background on this `Permitted<T>` enum.

so, we should make the `NewRecordBodyData` agnostic to its metrics, rather than holding a concrete `RequestBodyFamilies<L>`.

this commit hoists the metrics out of the middleware, and into the `X` and `ReqX` extractors used in the `NewService<T>` and `Service<T>` layers, respectively.

bear particular attention to this part of the diff:

```diff
-    ReqX: ExtractParam<L, Request<ReqB>>,
-    L: linkerd_metrics::prom::encoding::EncodeLabelSet
-        + std::fmt::Debug
-        + std::hash::Hash
-        + Eq
-        + Clone
-        + Send
-        + Sync
-        + 'static,
+    ReqX: ExtractParam<BodyDataMetrics, Request<ReqB>>,
```

in other words, rather than using `X` and subsequently `ReqX` to extract our labels `L`, we expect these extractors to yield a `BodyDataMetrics`: one time series that should be incremented accordingly by the instrumented request.

in even simpler words: `ReqX` now calls `familes.metrics(&labels)` too.

### DIFF
--- a/linkerd/app/outbound/src/http/logical/policy/route/metrics.rs
+++ b/linkerd/app/outbound/src/http/logical/policy/route/metrics.rs
@@ -5,7 +5,7 @@ use linkerd_app_core::{
     svc,
 };
 use linkerd_http_prom::{
-    body_data::request::{NewRecordBodyData, RequestBodyFamilies},
+    body_data::request::{BodyDataMetrics, NewRecordBodyData, RequestBodyFamilies},
     record_response::{self, StreamLabel},
 };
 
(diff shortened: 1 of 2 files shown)


cratelyn:
i have rebased this atop main as of 0060a4c3.

cratelyn:
this branch has been rebased atop main, now that #4181 has landed.

olix0r (approved):

PR #4175, opened 2025-09-20
fix(http/route): fix prefix path validation

adleong:
Fixes https://github.com/linkerd/linkerd2/issues/14110

When using a `HTTPRoute` `RequestRedirect` filter with a `ReplacePrefixMatch`, the proxy incorrectly returns an error: `redirect paths must be absolute`.  This is because the validation logic is backwards and it returns this error when the redirect path IS absolute.

We correct the validation logic by inverting it.


cratelyn (approved):
thank you very much for fixing this :relaxed:

PR #4190, opened 2025-09-24
refactor(app/outbound): remove frivolous `NewService<T>` bounds

cratelyn:
these bounds are extensive, but don't serve a mechanical purpose at the
moment.

we can simplify our route metrics layer in the outbound proxy by
removing them.


olix0r (commented):
So, these aren't load-bearing, but they are a helpful assertion if changes are made that break the stack's compilation: these assertions will give a clearer indication of why the N type isn't used.

At a minimum, I'd suggest preserving the `N: svc::NewService<T>` assertion which is idiomatic and lightweight

olix0r (approved):

PR #4199, opened 2025-09-26
refactor(http/prom): rename `NewRecordBodyData::layer_via()`

cratelyn:
in linkerd/linkerd2-proxy#4174, we refactored this middleware so that it
is now agnostic to the particular metrics structure used by the
extractor, meaning that the constructor no longer accepts a
`RequestBodyFamilies<L>` parameter.

in the wake of this change, our request body frame size metrics
middleware has a constructor with a type signature like `X -> impl
Layer`.

elsewhere throughout the project, constructors that accept an extractor
and return a `svc::Layer` are typically called `layer_via()`.

this commit renames `NewRecordBodyData::new()` to
`NewRecordBodyData::layer_via()`.

### DIFF
--- a/linkerd/app/inbound/src/http/router/metrics.rs
+++ b/linkerd/app/inbound/src/http/router/metrics.rs
@@ -37,7 +37,7 @@ pub(super) fn layer<N>(
 
     let request = {
         let extract = ExtractRequestBodyDataParams(request_body_data.clone());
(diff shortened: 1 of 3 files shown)


sfleen (approved):

cratelyn:
rebased on main to address conflicts with #4203.

PR #4223, opened 2025-10-06
feat(tracing): Include more tracing semantic convention labels

sfleen:
https://opentelemetry.io/docs/specs/semconv/http/http-spans/ defines the semantic conventions for traces of HTTP requests and responses. We currently include some of these labels, but not all that are recommended.

This adds in more of the labels, especially around the URI, transport, and header values.


olix0r (approved):
lgtm

PR #4279, opened 2025-11-12
chore(tracing): Use upstream tracing batch exporter

sfleen:
The proxy used its own custom trace exporter for some time. It was built before the upstream OpenTelemetry libraries were available, and they have outlived their usefulness.

This replaces the custom exporter with a batch exporter configured to export to the same endpoint with the same service configuration.

In the future, this exporter can be installed as a global default trace processor, which would decouple it from the service layer where the proxy generates spans for requests.

### DIFF
--- a/linkerd/app/src/trace_collector/otel_collector.rs
+++ b/linkerd/app/src/trace_collector/otel_collector.rs
@@ -1,13 +1,6 @@
 use super::EnabledCollector;
 use linkerd_app_core::{control::ControlAddr, proxy::http::Body, Error};
-use linkerd_opentelemetry::{
-    self as opentelemetry, metrics,
-    proto::{
-        tonic::common::v1::{any_value, AnyValue, KeyValue},
-        transform::common::tonic::ResourceAttributesWithSchema,
-    },
-    semconv,
-};
+use linkerd_opentelemetry::{self as opentelemetry, metrics, otel::KeyValue, sdk, semconv};
 use std::{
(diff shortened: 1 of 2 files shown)


linkerd/app/src/trace_collector/otel_collector.rs
```diff
@@ -38,37 +31,33 @@ where
...
-    resources.attributes.0.extend(
-        attributes
-            .extra
-            .into_iter()
-            .map(|(key, value)| value.with_key(&key)),
-    );
+    let resource = sdk::Resource::builder()
+        .with_attribute(KeyValue::new(
+            semconv::attribute::PROCESS_PID,
+            std::process::id() as i64,
```
unleashed: This may be a good time to future-proof potential pid space growth?

```suggestion
            i64::from(std::process::id()),
```

linkerd/opentelemetry/src/lib.rs
```diff
@@ -15,229 +15,132 @@ use opentelemetry::{
...
+
+    let processor = BatchSpanProcessor::builder(SpanExporter {
+        client: TraceServiceClient::new(client),
+        resource,
+        metrics,
+    })
+    .with_batch_config(
+        BatchConfigBuilder::default()
+            .with_max_queue_size(1000)
+            .with_scheduled_delay(Duration::from_secs(5))
```
unleashed: Have these hardcoded numbers defined as constants?

PR #4531, opened 2026-05-11
chore(deps): define `rand` as a workspace dependency

cratelyn:
this commit defines our dependencies on rand throughout the cargo
workspace in terms of a single workspace-level dependency. no changes to
Cargo.lock are included here because this is a noöp like-for-like
change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,6 +110,7 @@ prometheus-client = { version = "0.23" }
 prost = { version = "0.14" }
 prost-build = { version = "0.14", default-features = false }
 prost-types = { version = "0.14" }
+rand = { version = "0.9", default-features = false }
 rustls-pki-types = { version = "1.14" }
 rustls-webpki = { version = "0.103.13", default-features = false }
 tokio-rustls = { version = "0.26", default-features = false, features = [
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -110,6 +110,7 @@ prometheus-client = { version = "0.23" }
 prost = { version = "0.14" }
 prost-build = { version = "0.14", default-features = false }
 prost-types = { version = "0.14" }
+rand = { version = "0.9", default-features = false }
```
cratelyn: https://github.com/rust-random/rand/compare/0.9.4...0.10.1#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542L31-R32

the list of default features changed from 0.9 to 0.10. to follow what's generally a best practice for avoiding breaking changes due to default features being removed, and to prepare for an upgrade to 0.10, this commit uses `default-features = false` in our workspace dependency.

individual crate-level dependencies have added `thread_rng` where needed.

cratelyn: `"std", "std_rng", "os_rng"` are also included in the default feature set, with the latter renamed to `sys_rng` in 0.10.

to reviewers: should we include these too?

adleong (approved):

Cargo.toml
cratelyn: i pushed d5f94a6 to address this. i'm going to mark this as resolved. :white_check_mark:

PR #4537, opened 2026-05-21
Add EWMA and load biasing crates for failure-aware P2C balancing

unleashed:
Today the proxy's P2C load balancer uses Tower's PeakEwma, which tracks
only round-trip time. An endpoint returning fast 503s or 429s looks
"fast" to PeakEwma, so P2C keeps routing traffic to it. This is exactly the
opposite of what operators want.

This PR adds the building blocks to make P2C failure-aware, but does not
wire anything in the proxy stack yet to keep the reviews' scope manageable.
Follow-up PR's will make use of these building blocks to activate this code
and implement related features in the circuit breaker.

Here are the main components:

- **linkerd-ewma**. A standalone EWMA crate that supports non-mutating
  time-projected reads and records externally supplied peak values.
  Tower's internal RttEstimate is private, mutates on read, and samples
  only elapsed request time, so it cannot provide the read-only load
  projection or the injected failure penalties that failure-aware load
  balancing needs.

- **retry_after module in linkerd-http-classify**. Parsers for HTTP
  Retry-After (delay-seconds and HTTP-date per RFC 7231) and gRPC
  grpc-retry-pushback-ms (per gRPC A6 spec), so the load biaser and the
  upcoming circuit breaker can honor server backoff hints.

- **linkerd-load-biaser**. A Tower Service wrapper implementing
  `tower::load::Load` that tracks per-endpoint RTT via EWMA and injects
  temporary load penalties on failure responses (HTTP 429/503/5xx, gRPC
  RESOURCE_EXHAUSTED/UNAVAILABLE). When a Retry-After hint is present the
  penalty is amplified to remain meaningful through the server-requested
  backoff window. The load metric is `rtt * (pending + 1)`, giving P2C
  the ability to steer traffic away from unhealthy endpoints while preserving
  the same behavior as PeakEwma when all of them are healthy.

### DIFF
--- a/linkerd/ewma/Cargo.toml
+++ b/linkerd/ewma/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "linkerd-ewma"
+version = { workspace = true }
+authors = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
+publish = { workspace = true }
+
+[dependencies]
+tokio = { version = "1", features = ["time"] }
+
+[dev-dependencies]
+tokio = { version = "1", features = ["macros", "rt", "test-util"] }
--- a/linkerd/ewma/Cargo.toml
+++ b/linkerd/ewma/Cargo.toml
@@ -7,7 +7,7 @@ edition = { workspace = true }
 publish = { workspace = true }
 
 [dependencies]
-tokio = { version = "1", features = ["time"] }
+tokio = { version = "1", default-features = false, features = ["time"] }
 
 [dev-dependencies]
 tokio = { version = "1", features = ["macros", "rt", "test-util"] }
--- a/linkerd/ewma/src/lib.rs
+++ b/linkerd/ewma/src/lib.rs
@@ -1,3 +1,10 @@
+//! Exponentially weighted moving average (EWMA) with time-based decay.
+//!
+//! A standalone EWMA implementation that supports non-mutating time-projected
+//! reads via [`Ewma::get_at`] and dual-metric tracking (RTT + penalty) under a
+//! single lock. Tower's internal `RttEstimate` is private, mutates on read, and
+//! cannot support the penalty dimension that failure-aware load balancing needs.
+
 #![deny(rust_2018_idioms, clippy::disallowed_methods, clippy::disallowed_types)]
 #![forbid(unsafe_code)]
 
--- a/linkerd/ewma/src/lib.rs
+++ b/linkerd/ewma/src/lib.rs
@@ -485,4 +485,63 @@ mod tests {
         // Should trigger a debug_assert
         let _ = ewma.get_at(now);
     }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_last_update_tracks_construction() {
+        let now = Instant::now();
+        let ewma = Ewma::new(Duration::from_secs(10), now);
+        assert_eq!(ewma.last_update(), now);
+
+        let ewma = Ewma::new_with_value(Duration::from_secs(10), now, 1.0);
+        assert_eq!(ewma.last_update(), now);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_last_update_advances_on_add() {
+        let now = Instant::now();
+        let mut ewma = Ewma::new(Duration::from_secs(10), now);
+
+        let t1 = now + Duration::from_secs(1);
+        ewma.add(1.0, t1);
+        assert_eq!(ewma.last_update(), t1);
+
+        let t2 = now + Duration::from_secs(5);
+        ewma.add(0.5, t2);
+        assert_eq!(ewma.last_update(), t2);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_last_update_advances_on_reset() {
+        let now = Instant::now();
+        let mut ewma = Ewma::new(Duration::from_secs(10), now);
+
+        ewma.add(1.0, now + Duration::from_secs(1));
+        let reset_at = now + Duration::from_secs(5);
+        ewma.reset(1.0, reset_at);
+        assert_eq!(ewma.last_update(), reset_at);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_last_update_unchanged_by_read() {
+        let now = Instant::now();
+        let mut ewma = Ewma::new(Duration::from_secs(10), now);
+
+        let t1 = now + Duration::from_secs(1);
+        ewma.add(1.0, t1);
+        let _ = ewma.get_at(now + Duration::from_secs(10));
+        assert_eq!(ewma.last_update(), t1);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_last_update_unchanged_by_stale_add() {
+        let now = Instant::now();
+        let mut ewma = Ewma::new(Duration::from_secs(10), now);
+
+        let t1 = now + Duration::from_secs(5);
+        ewma.add(1.0, t1);
+
+        // Stale timestamp (before last update) is silently dropped.
+        ewma.add(0.0, now + Duration::from_secs(2));
+        assert_eq!(ewma.last_update(), t1);
+    }
 }
--- a/linkerd/ewma/src/lib.rs
+++ b/linkerd/ewma/src/lib.rs
@@ -218,6 +218,27 @@ mod tests {
     }
 
     #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_add_peak_retains_higher_value() {
+        let now = Instant::now();
+        let mut ewma = Ewma::new(Duration::from_secs(10), now);
+
+        ewma.add_peak(10.0, now + Duration::from_secs(1));
+        assert_eq!(ewma.get(), 10.0);
+
+        // A lower sample arrives shortly after. The decayed projection
+        // would be close to 9.9, well above 1.0. Now add_peak() should add
+        // the measurement, but it should only slightly move the EWMA,
+        // leaving the moving average still close to 10.
+        ewma.add_peak(1.0, now + Duration::from_millis(1100));
+        let v = ewma.get();
+
+        assert!(
+            v > 9.0 && v < 10.0,
+            "expected EWMA value between 9 and 10, got {v}"
+        );
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
     async fn test_zero_decay_clamped_in_new() {
         let now = Instant::now();
         let zero = Ewma::new(Duration::ZERO, now);
--- a/linkerd/ewma/src/lib.rs
+++ b/linkerd/ewma/src/lib.rs
@@ -70,6 +70,11 @@ impl Ewma {
         self.value
     }
 
+    /// Returns the timestamp of the most recent update.
+    pub fn last_update(&self) -> time::Instant {
+        self.timestamp
+    }
+
     /// Returns the decayed value projected to the given time, without modifying stored state.
     ///
     /// Instead of returning the raw stored value, this applies exponential decay based
--- a/linkerd/ewma/src/lib.rs
+++ b/linkerd/ewma/src/lib.rs
@@ -3,6 +3,11 @@
 
 use tokio::time;
 
+/// Minimum decay duration to prevent division-by-zero in EWMA computations.
+/// Chosen as the smallest Duration that is strictly positive without overriding
+/// validated configs from the control plane (CP should reject decay=0).
+pub const MIN_DECAY: time::Duration = time::Duration::from_millis(1);
+
 /// An exponentially-weighted moving average.
 #[derive(Debug)]
 pub struct Ewma {
@@ -17,15 +22,60 @@ impl Ewma {
     #[must_use]
     pub fn new(decay: time::Duration, timestamp: time::Instant) -> Self {
         Self {
-            decay: decay.as_secs_f64(),
+            decay: decay.max(MIN_DECAY).as_secs_f64(),
             timestamp,
             value: f64::INFINITY,
         }
     }
 
+    /// Creates an EWMA with a specific initial value.
+    ///
+    /// This constructor allows setting an initial value, useful for
+    /// success rate tracking where you want to start at 100% (1.0)
+    /// success rate.
+    #[must_use]
+    pub fn new_with_value(decay: time::Duration, timestamp: time::Instant, initial: f64) -> Self {
+        debug_assert!(!initial.is_nan(), "EWMA initial value must not be NaN");
+        Self {
+            decay: decay.max(MIN_DECAY).as_secs_f64(),
+            timestamp,
+            value: initial,
+        }
+    }
+
+    /// Resets the EWMA to a new value and timestamp.
+    ///
+    /// This overwrites the current value and timestamp, useful for
+    /// resetting success rate tracking after recovery from a tripped state.
+    ///
+    /// Precondition: `ts` should be <= the timestamps passed to subsequent
+    /// `add()` calls. If it is not, those `add()` calls are silently
+    /// dropped because `add()` discards samples whose timestamp is at or
+    /// before the stored one.
+    pub fn reset(&mut self, value: f64, ts: time::Instant) {
+        debug_assert!(!value.is_nan(), "EWMA reset value must not be NaN");
+        self.value = value;
+        self.timestamp = ts;
+    }
+
     /// Returns the current value of the average.
     pub fn get(&self) -> f64 {
         self.value
+    }
+
+    /// Returns the decayed value projected to the given time, without modifying stored state.
+    ///
+    /// Instead of returning the raw stored value, this applies exponential decay based
+    /// on elapsed time since the last update. This is required for load balancing where
+    /// stale measurements should lose influence over time.
+    pub fn get_at(&self, now: time::Instant) -> f64 {
+        debug_assert!(!self.value.is_nan(), "EWMA value must not be NaN");
+
+        if self.value.is_infinite() || now <= self.timestamp {
+            return self.value;
+        }
+        let elapsed = now.saturating_duration_since(self.timestamp);
+        self.value * (-elapsed.as_secs_f64() / self.decay).exp()
     }
 
     /// Updates the weighted moving average with a new value and timestamp.
@@ -52,8 +102,16 @@ impl Ewma {
         self.timestamp = ts;
     }
 
+    /// Updates the EWMA with a peak value, replacing the current value if the
+    /// new value exceeds the decayed projection.
+    ///
+    /// When replacement occurs, the stored timestamp is set to `ts`, which
+    /// may be earlier than the previously stored timestamp. This resets the
+    /// decay reference point, so subsequent projections via `get_at()` measure
+    /// elapsed time from `ts`.
     pub fn add_peak(&mut self, value: f64, ts: time::Instant) {
-        if self.value < value {
+        debug_assert!(!value.is_nan(), "EWMA peak value must not be NaN");
+        if self.value.is_infinite() || self.get_at(ts) < value {
             self.value = value;
             self.timestamp = ts;
             return;
@@ -77,6 +135,9 @@ impl Ewma {
 mod tests {
     use super::*;
     use tokio::time::{Duration, Instant};
+
+    // Literal value for exp(-1.0), since it's not const
+    const EXP_NEG1: f64 = 0.36787944117144233;
 
     #[tokio::test(flavor = "current_thread", start_paused = true)]
     async fn test_new() {
@@ -120,6 +181,296 @@ mod tests {
         ewma.add(1.0, now + Duration::from_secs(1));
         assert_eq!(ewma.get(), 1.0);
         ewma.add(0.0, now + Duration::from_secs(11));
-        assert_eq!(ewma.get(), 0.36787944117144233);
+        assert_eq!(ewma.get(), EXP_NEG1);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_new_with_value() {
+        let now = Instant::now();
+        let ewma = Ewma::new_with_value(Duration::from_secs(10), now, 1.0);
+
+        assert_eq!(ewma.get(), 1.0);
+
+        // Verify this behaves like a normal EWMA after initialization
+        let mut ewma = Ewma::new_with_value(Duration::from_secs(10), now, 1.0);
+
+        ewma.add(0.0, now + Duration::from_secs(10));
+        // After one decay period value decays towards zero.
+        assert_eq!(ewma.get(), EXP_NEG1);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_add_peak_from_infinity_same_timestamp() {
+        let now = Instant::now();
+        let mut ewma = Ewma::new(Duration::from_secs(10), now);
+
+        assert_eq!(ewma.get(), f64::INFINITY);
+
+        // Same timestamp as construction. The first real value should always
+        // take effect regardless of timestamp.
+        ewma.add_peak(0.5, now);
+        assert_eq!(ewma.get(), 0.5);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_add_peak_replaces_decayed_value() {
+        let now = Instant::now();
+        let mut ewma = Ewma::new(Duration::from_secs(10), now);
+
+        // Set initial peak of 10.0 at t=1s
+        ewma.add_peak(10.0, now + Duration::from_secs(1));
+        assert_eq!(ewma.get(), 10.0);
+
+        // After 25s of decay (t=26s), the decayed projection is:
+        // 10.0 * exp(-25/10) = 10.0 * exp(-2.5) = 0.8208...
+        // Since 0.8208 < 1.0, the new value should replace the stale peak.
+        ewma.add_peak(1.0, now + Duration::from_secs(26));
+        assert_eq!(ewma.get(), 1.0);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_zero_decay_clamped_in_new() {
+        let now = Instant::now();
+        let zero = Ewma::new(Duration::ZERO, now);
+        let min = Ewma::new(Duration::from_millis(1), now);
+
+        // Both should produce identical behavior since ZERO gets clamped to 1ms
+        assert_eq!(zero.get(), min.get());
+
+        // After adding a value, get_at should produce finite, identical results
+        let mut zero = Ewma::new(Duration::ZERO, now);
+        let mut min = Ewma::new(Duration::from_millis(1), now);
+        zero.add(5.0, now + Duration::from_secs(1));
+        min.add(5.0, now + Duration::from_secs(1));
+
+        let projected_zero = zero.get_at(now + Duration::from_secs(2));
+        let projected_min = min.get_at(now + Duration::from_secs(2));
+
+        assert!(
+            projected_zero.is_finite(),
+            "get_at must be finite with clamped decay"
+        );
+        assert!(
+            !projected_zero.is_nan(),
+            "get_at must not be NaN with clamped decay"
+        );
+        assert_eq!(projected_zero, projected_min);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_zero_decay_clamped_in_new_with_value() {
+        let now = Instant::now();
+        let zero = Ewma::new_with_value(Duration::ZERO, now, 1.0);
+        let min = Ewma::new_with_value(Duration::from_millis(1), now, 1.0);
+
+        let projected_zero = zero.get_at(now + Duration::from_secs(1));
+        let projected_min = min.get_at(now + Duration::from_secs(1));
+        assert!(
+            projected_zero.is_finite(),
+            "get_at must be finite with clamped decay"
+        );
+        assert!(
+            !projected_zero.is_nan(),
+            "get_at must not be NaN with clamped decay"
+        );
+        assert_eq!(projected_zero, projected_min);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_zero_decay_add_is_safe() {
+        let now = Instant::now();
+        let mut ewma = Ewma::new(Duration::ZERO, now);
+
+        ewma.add(5.0, now + Duration::from_secs(1));
+        ewma.add(0.5, now + Duration::from_secs(2));
+
+        let val = ewma.get();
+        assert!(
+            val.is_finite(),
+            "add() result must be finite with clamped decay"
+        );
+        assert!(
+            !val.is_nan(),
+            "add() result must not be NaN with clamped decay"
+        );
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_reset() {
+        let now = Instant::now();
+        let mut ewma = Ewma::new(Duration::from_secs(10), now);
+
+        // Set state
+        ewma.add(0.5, now + Duration::from_secs(1));
+        assert_eq!(ewma.get(), 0.5);
+
+        // Reset to a new value
+        ewma.reset(1.0, now + Duration::from_secs(2));
+        assert_eq!(ewma.get(), 1.0);
+
+        // Verify EWMA continues working after reset.
+        ewma.add(0.0, now + Duration::from_secs(12));
+        assert_eq!(ewma.get(), EXP_NEG1);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_get_at_same_timestamp() {
+        let now = Instant::now();
+        let decay = Duration::from_secs(10);
+        let add_at = now + Duration::from_secs(1);
+        let read_at = add_at;
+
+        let mut ewma = Ewma::new(decay, now);
+        ewma.add(0.5, add_at);
+
+        assert_eq!(ewma.get_at(read_at), 0.5);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_get_at_past_timestamp() {
+        let now = Instant::now();
+        let decay = Duration::from_secs(10);
+        let add_at = now + Duration::from_secs(1);
+        let read_at = now + Duration::from_millis(500);
+
+        let mut ewma = Ewma::new(decay, now);
+        ewma.add(0.5, add_at);
+
+        assert_eq!(ewma.get_at(read_at), 0.5);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_get_at_infinity() {
+        let now = Instant::now();
+        let decay = Duration::from_secs(10);
+        let probe_same = now;
+        let probe_near = now + Duration::from_secs(1);
+        let probe_far = now + Duration::from_secs(100);
+
+        // A new Ewma without addding values should project INFINITY
+        // at every timestamp.
+        let ewma = Ewma::new(decay, now);
+        assert_eq!(ewma.get_at(probe_same), f64::INFINITY);
+        assert_eq!(ewma.get_at(probe_near), f64::INFINITY);
+        assert_eq!(ewma.get_at(probe_far), f64::INFINITY);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_get_at_decay() {
+        let now = Instant::now();
+        let decay = Duration::from_secs(10);
+        let add_at = now + Duration::from_secs(1);
+        let read_at = now + Duration::from_secs(11);
+
+        let mut ewma = Ewma::new(decay, now);
+        ewma.add(1.0, add_at);
+
+        // Verify that get_at applies value * exp(-elapsed/decay) correctly
+        // without changing internal state.
+        assert_eq!(ewma.get_at(read_at), EXP_NEG1);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_get_at_large_elapsed() {
+        let now = Instant::now();
+        let decay = Duration::from_secs(10);
+        let add_at = now + Duration::from_secs(1);
+        let read_at = now + Duration::from_secs(3600);
+
+        let mut ewma = Ewma::new(decay, now);
+        ewma.add(1.0, add_at);
+
+        let result = ewma.get_at(read_at);
+        assert!(
+            result.is_finite(),
+            "get_at at large elapsed must be finite, got {result}"
+        );
+        assert!(
+            result < 1e-10,
+            "get_at at large elapsed must be near zero, got {result}"
+        );
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_get_at_non_mutation() {
+        let now = Instant::now();
+        let decay = Duration::from_secs(10);
+        let first_add_at = now + Duration::from_secs(1);
+        let read_at = now + Duration::from_secs(6);
+
+        let mut ewma = Ewma::new(decay, now);
+        ewma.add(1.0, first_add_at);
+
+        // Take internal state before the read
+        let value_before = ewma.value;
+        let timestamp_before = ewma.timestamp;
+
+        // Read must not mutate
+        let _ = ewma.get_at(read_at);
+
+        assert_eq!(ewma.value, value_before);
+        assert_eq!(ewma.timestamp, timestamp_before);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_get_at_after_reset() {
+        const EWMA_VAL: f64 = 0.5;
+        const DECAY_AT_15S_VAL: f64 = EWMA_VAL * EXP_NEG1;
+
+        let now = Instant::now();
+        let decay = Duration::from_secs(10);
+        let first_add_at = now + Duration::from_secs(1);
+        let reset_at = now + Duration::from_secs(5);
+        let immediate_read_at = reset_at;
+        let decay_read_at = now + Duration::from_secs(15);
+
+        let mut ewma = Ewma::new(decay, now);
+        // Use a large value here to make sure we detect any issues with
+        // reset not working properly, since we'd likely see a wildly
+        // different value in the assert below.
+        ewma.add(100.0, first_add_at);
+
+        // Reset replaces both value and timestamp
+        ewma.reset(EWMA_VAL, reset_at);
+
+        // Must return the freshly-reset value
+        assert_eq!(ewma.get_at(immediate_read_at), EWMA_VAL);
+
+        // Read one decay period (10s) after the reset (so 15s decay).
+        assert_eq!(ewma.get_at(decay_read_at), DECAY_AT_15S_VAL);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_get_at_between_adds() {
+        const EXP_NEG0_5: f64 = 0.6065306597126334;
+
+        let now = Instant::now();
+        let decay = Duration::from_secs(10);
+        let first_add_at = now + Duration::from_secs(1);
+        let read_at = now + Duration::from_secs(6);
+        let second_add_at = now + Duration::from_secs(11);
+
+        let mut ewma = Ewma::new(decay, now);
+        ewma.add(1.0, first_add_at);
+
+        assert_eq!(ewma.get_at(read_at), EXP_NEG0_5);
+
+        ewma.add(0.0, second_add_at);
+
+        // Final state after the second add must be exp(-1.0),
+        // ensuring get_at() doesn't change the internal state.
+        assert_eq!(ewma.get(), EXP_NEG1);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    #[cfg(debug_assertions)]
+    #[should_panic(expected = "NaN")]
+    async fn test_get_at_debug_asserts_nan() {
+        let now = Instant::now();
+        // Inject NaN via new_with_value
+        let ewma = Ewma::new_with_value(Duration::from_secs(10), now, f64::NAN);
+
+        // Should trigger a debug_assert
+        let _ = ewma.get_at(now);
     }
 }
--- a/linkerd/ewma/src/lib.rs
+++ b/linkerd/ewma/src/lib.rs
@@ -97,7 +97,7 @@ impl Ewma {
         if ts <= self.timestamp {
             return;
         }
-        if self.value == f64::INFINITY {
+        if self.value.is_infinite() {
             self.value = value;
             self.timestamp = ts;
             return;
--- a/linkerd/ewma/src/lib.rs
+++ b/linkerd/ewma/src/lib.rs
@@ -154,6 +154,25 @@ mod tests {
     }
 
     #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_add_ignores_stale_timestamps() {
+        let now = Instant::now();
+        let t1 = now + Duration::from_secs(2);
+        let mut ewma = Ewma::new(Duration::from_secs(10), now);
+
+        // Store an initial value at t1, replacing the inf sentinel.
+        ewma.add(1.0, t1);
+        assert_eq!(ewma.get(), 1.0);
+
+        // A measurement at the stored timestamp is discarded.
+        ewma.add(99.0, t1);
+        assert_eq!(ewma.get(), 1.0);
+
+        // Also discarded if the timestamp is before the stored one.
+        ewma.add(99.0, t1 - Duration::from_secs(1));
+        assert_eq!(ewma.get(), 1.0);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
     async fn test_add_peak() {
         let now = Instant::now();
         let mut ewma = Ewma::new(Duration::from_secs(10), now);
--- a/linkerd/ewma/src/lib.rs
+++ b/linkerd/ewma/src/lib.rs
@@ -1,0 +1,125 @@
+#![deny(rust_2018_idioms, clippy::disallowed_methods, clippy::disallowed_types)]
+#![forbid(unsafe_code)]
+
+use tokio::time;
+
+/// An exponentially-weighted moving average.
+#[derive(Debug)]
+pub struct Ewma {
+    value: f64,
+    decay: f64,
+    timestamp: time::Instant,
+}
+
+// === impl Ewma ===
+
+impl Ewma {
+    #[must_use]
+    pub fn new(decay: time::Duration, timestamp: time::Instant) -> Self {
+        Self {
+            decay: decay.as_secs_f64(),
+            timestamp,
+            value: f64::INFINITY,
+        }
+    }
+
+    /// Returns the current value of the average.
+    pub fn get(&self) -> f64 {
+        self.value
+    }
+
+    /// Updates the weighted moving average with a new value and timestamp.
+    ///
+    /// Precondition: `value` must not be NaN. Passing NaN poisons the EWMA
+    /// irreversibly (all subsequent reads return NaN).
+    pub fn add(&mut self, value: f64, ts: time::Instant) {
+        debug_assert!(!value.is_nan(), "EWMA input value must not be NaN");
+        if ts <= self.timestamp {
+            return;
+        }
+        if self.value == f64::INFINITY {
+            self.value = value;
+            self.timestamp = ts;
+            return;
+        }
+
+        self.value = {
+            let elapsed = ts.saturating_duration_since(self.timestamp);
+            let alpha = 1.0 - (-elapsed.as_secs_f64() / self.decay).exp();
+            self.value * (1.0 - alpha) + value * alpha
+        };
+
+        self.timestamp = ts;
+    }
+
+    pub fn add_peak(&mut self, value: f64, ts: time::Instant) {
+        if self.value < value {
+            self.value = value;
+            self.timestamp = ts;
+            return;
+        }
+        self.add(value, ts)
+    }
+
+    /// Computes 1/elapsed since the last update and feeds it through `add()`.
+    pub fn add_rate(&mut self, ts: time::Instant) {
+        if ts <= self.timestamp {
+            return;
+        }
+        let elapsed = ts.saturating_duration_since(self.timestamp);
+        if !elapsed.is_zero() {
+            self.add(1.0 / elapsed.as_secs_f64(), ts);
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use tokio::time::{Duration, Instant};
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_new() {
+        let now = Instant::now();
+        let ewma = Ewma::new(Duration::from_secs(10), now);
+        assert_eq!(ewma.get(), f64::INFINITY);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_add() {
+        let now = Instant::now();
+        let mut ewma = Ewma::new(Duration::from_secs(10), now);
+        ewma.add(1.0, now + Duration::from_secs(1));
+        assert_eq!(ewma.get(), 1.0);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_add_rate() {
+        let now = Instant::now();
+        let mut ewma = Ewma::new(Duration::from_secs(10), now);
+        ewma.add_rate(now + Duration::from_secs(1));
+        assert_eq!(ewma.get(), 1.0);
+        ewma.add_rate(now + Duration::from_secs(3));
+        assert_eq!(ewma.get(), 0.9093653765389909);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_add_peak() {
+        let now = Instant::now();
+        let mut ewma = Ewma::new(Duration::from_secs(10), now);
+        ewma.add_peak(1.0, now + Duration::from_secs(1));
+        assert_eq!(ewma.get(), 1.0);
+        ewma.add_peak(2.0, now + Duration::from_secs_f64(1.5));
+        assert_eq!(ewma.get(), 2.0);
+    }
+
+    #[tokio::test(flavor = "current_thread", start_paused = true)]
+    async fn test_decay() {
+        let now = Instant::now();
+        let mut ewma = Ewma::new(Duration::from_secs(10), now);
+        ewma.add(1.0, now + Duration::from_secs(1));
+        assert_eq!(ewma.get(), 1.0);
+        ewma.add(0.0, now + Duration::from_secs(11));
+        assert_eq!(ewma.get(), 0.36787944117144233);
+    }
+}
--- a/linkerd/ewma/src/lib.rs
+++ b/linkerd/ewma/src/lib.rs
@@ -1,9 +1,11 @@
 //! Exponentially weighted moving average (EWMA) with time-based decay.
 //!
-//! A standalone EWMA implementation that supports non-mutating time-projected
-//! reads via [`Ewma::get_at`] and dual-metric tracking (RTT + penalty) under a
-//! single lock. Tower's internal `RttEstimate` is private, mutates on read, and
-//! cannot support the penalty dimension that failure-aware load balancing needs.
+//! A standalone EWMA that supports non-mutating, time-projected reads via
+//! [`Ewma::get_at`] and records externally supplied peak values via
+//! [`Ewma::add_peak`]. Tower's internal `RttEstimate` is private, mutates on
+//! read, and samples only elapsed request time, so it cannot provide the
+//! read-only load projection or the injected failure penalties that
+//! failure-aware load balancing needs.
 
 #![deny(rust_2018_idioms, clippy::disallowed_methods, clippy::disallowed_types)]
 #![forbid(unsafe_code)]
--- a/linkerd/ewma/src/lib.rs
+++ b/linkerd/ewma/src/lib.rs
@@ -354,7 +354,7 @@ mod tests {
         let probe_near = now + Duration::from_secs(1);
         let probe_far = now + Duration::from_secs(100);
 
-        // A new Ewma without addding values should project INFINITY
+        // A new Ewma without adding values should project INFINITY
         // at every timestamp.
         let ewma = Ewma::new(decay, now);
         assert_eq!(ewma.get_at(probe_same), f64::INFINITY);
--- a/linkerd/ewma/src/lib.rs
+++ b/linkerd/ewma/src/lib.rs
@@ -10,9 +10,10 @@
 
 use tokio::time;
 
-/// Minimum decay duration to prevent division-by-zero in EWMA computations.
-/// Chosen as the smallest Duration that is strictly positive without overriding
-/// validated configs from the control plane (CP should reject decay=0).
+/// Smallest decay duration the estimator will use.
+///
+/// The constructors clamp any supplied decay up to this value, which avoids
+/// a division by zero when computing `elapsed / decay`.
 pub const MIN_DECAY: time::Duration = time::Duration::from_millis(1);
 
 /// An exponentially-weighted moving average.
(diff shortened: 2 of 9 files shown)


linkerd/http/classify/Cargo.toml
```diff
@@ -10,6 +10,7 @@ publish = { workspace = true }
 futures = { version = "0.3", default-features = false }
 http = { workspace = true }
 http-body = { workspace = true }
+httpdate.workspace = true
```
cratelyn: nit: just to follow convention with the rest of the file

```suggestion
httpdate = { workspace = true }
```

raykroeker (commented):
@unleashed Thanks for the documentation.  It really helps understand the intent.
+100

linkerd/ewma/src/lib.rs
cratelyn, line 1: > A standalone EWMA crate that supports non-mutating
time-projected reads and dual-metric tracking (RTT + penalty) under a
single lock. Tower's internal RttEstimate is private, mutates on read,
and cannot support the penalty dimension.

i'm having trouble connecting the dot from this note in the pull request description to this code.

i don't see an `rtt_estimate` value, like the [tower code in question](https://docs.rs/tower/latest/src/tower/load/peak_ewma.rs.html#91-110). is that just `value` here?

cratelyn: it would also be helpful if we could record some of this rationale in a crate-level `//! doc comment`, so that we retain some record about why we're not using tower's equivalent EWMA.

copy-pasting the excerpt quoted above is fine by me :)

linkerd/ewma/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "linkerd-ewma"
+version = { workspace = true }
+authors = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
+publish = { workspace = true }
+
+[dependencies]
+tokio = { version = "1", features = ["time"] }
```
cratelyn: maybe `default-features = false` would be nice here too, since we're not setting up a runtime or anything else in this crate. `time` interfaces seem like all we're using here!

linkerd/http/classify/src/retry_after.rs
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    // === gRPC pushback tests ===
+
+    const MAX_GRPC: Duration = Duration::from_millis(300_000);
+
+    #[test]
+    fn parse_grpc_pushback_positive() {
+        let mut headers = HeaderMap::new();
+        headers.insert("grpc-retry-pushback-ms", HeaderValue::from_static("5000"));
```
cratelyn: should this and the tests below use the `GRPC_RETRY_PUSHBACK_MS`? tests above use `http::header::RETRY_AFTER`, so that'd be consistent.

linkerd/load-biaser/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+http = { workspace = true }
+linkerd-http-classify = { path = "../http/classify" }
+linkerd-stack = { path = "../stack" }
+parking_lot = "0.12"
+pin-project = "1"
+tokio = { version = "1", features = ["io-util", "net", "time"] }
+tokio-test = { version = "0.4", optional = true }
+tower = { workspace = true, features = ["load"] }
+tower-service = { workspace = true }
+tracing = { workspace = true }
```
cratelyn:
```suggestion
futures = { version = "0.3", default-features = false }
http = { workspace = true }
parking_lot = "0.12"
pin-project = "1"
tokio = { version = "1", features = ["io-util", "net", "time"] }
tokio-test = { version = "0.4", optional = true }
tower = { workspace = true, features = ["load"] }
tower-service = { workspace = true }
tracing = { workspace = true }

linkerd-ewma = { path = "../ewma" }
linkerd-http-classify = { path = "../http/classify" }
linkerd-stack = { path = "../stack" }
```

nit, alphabetize and pulling path-based dependencies into a separate block

linkerd/load-biaser/src/lib.rs
cratelyn, line 1: flagging this as the file in this pull request that i have not yet reviewed. i'll take a look through the rest of this after the weekend...

cratelyn (dismissed):
thanks for breaking these additions out into a standalone pull request, separate from the changes we'll be making in our proxy stack(s). that really helped expedite review of this.

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1582 @@
...
+        // Store the uncapped value. Each consumer applies their own cap via
+        // rate_limit_hint(max).
+        if let Some(d) = linkerd_http_classify::retry_after::parse_retry_after(
+            self.status(),
+            self.headers(),
+            Duration::MAX,
+        ) {
+            self.extensions_mut().insert(CachedRateLimitHint(d));
+            return;
+        }
```
cratelyn: i'm a little confused about why this `_max` isn't used. do we have any implementations of this trait where we _do_ use the `max` parameter in this method? i don't see it ever used.

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1582 @@
...
+    InternalError,
+}
+
+/// Cached Retry-After hint stored in HTTP response extensions.
+///
+/// Stores the **uncapped** parsed value. Each consumer applies its own cap
+/// via `rate_limit_hint(max)`, so different callers (e.g. load biaser vs
+/// circuit breaker) can use different maximums from the same cached value.
+#[derive(Clone, Copy, Debug)]
+pub struct CachedRateLimitHint(pub Duration);
```
cratelyn: does this need to be `pub`?

the uncapped value, along with the fact that this is intended for use with `rate_limit_hint`, makes me wonder if a constructor `pub fn new` could work for creating these, while preventing accidents with an uncapped duration in the future.

linkerd/load-biaser/src/lib.rs
unleashed: No we don't use it anywhere, it is there purely for symmetry with the other trait method in case we'd come up with an impl using it (but I don't really see any future need for it).

linkerd/load-biaser/src/lib.rs
unleashed: Good suggestion, I'll adapt it to use a constructor.

linkerd/load-biaser/src/lib.rs
cratelyn: in that case, what do you think about removing the parameter? we can always reintroduce it if a future need arises!

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,597 @@
...
+                        let penalty_val = match hint {
+                            FailureHint::RateLimited | FailureHint::ServiceUnavailable => {
+                                match resp.rate_limit_hint(shared.max_duration) {
+                                    Some(ra) if ra.as_secs_f64() > 0.0 => {
+                                        let decay_secs =
+                                            shared.penalty_decay.max(MIN_DECAY).as_secs_f64();
+                                        let amplified = base_penalty
+                                            * RETRY_AFTER_PENALTY_FACTOR
+                                            * (ra.as_secs_f64() / decay_secs).exp();
+                                        amplified.min(1e12)
```
cratelyn: what's this `1e12` constant? my understanding is that this is clamping the value of this to stay beneath an upper bound. since we have a `MIN_DECAY` constant, wny not define this as a `MAX_PENALTY` constant?

less cosmetically: this seems like a _excessively_ high constant. one trillion seconds would be 31,709.8 years.

linkerd/http/classify/src/retry_after.rs
```diff
@@ -0,0 +1,320 @@
...
+/// Per gRPC A6 spec:
+/// - Positive i64: retry after this many milliseconds
+/// - Negative i64: do not retry (returns `None`)
+///
+/// This function does **not** check grpc-status; that is a classification
+/// concern left to the caller.
+///
+/// Returns `None` for:
+/// - Missing header/trailer
+/// - Negative values (interpreted as "do not retry")
```
adleong: The semantics of `grpc-retry-pushback-ms` are defined in the context of retries, where a response of "do not retry" makes sense.  In this case, this isn't a response we can or should honor (we wouldn't want to leave the endpoint in probation forever) but we should be explicit about how we treat this.  The options would be either:
* ignore the negative value and use the original backoff schedule.  this is accomplished by returning `None` as you do here, but we should update the comment to clarify that we are not respecting the "do not retry" semantics
* treat the "do not retry" message as a signal that the endpoint is maximally overloaded and use the maximum allowed probation backoff (`max`).

linkerd/http/classify/src/retry_after.rs
```diff
@@ -0,0 +1,320 @@
...
+    tracing::debug!(%s, "Failed to parse Retry-After header");
+    None
+}
+
+/// The grpc-retry-pushback-ms header/trailer name.
+const GRPC_RETRY_PUSHBACK_MS: &str = "grpc-retry-pushback-ms";
+
+/// Parse grpc-retry-pushback-ms from headers or trailers.
+///
+/// Per gRPC A6 spec:
```
adleong: Adding a link to https://github.com/grpc/proposal/blob/master/A6-client-retries.md#pushback would be helpful.

linkerd/ewma/src/lib.rs
```diff
@@ -0,0 +1,547 @@
...
+
+#![deny(rust_2018_idioms, clippy::disallowed_methods, clippy::disallowed_types)]
+#![forbid(unsafe_code)]
+
+use tokio::time;
+
+/// Minimum decay duration to prevent division-by-zero in EWMA computations.
+/// Chosen as the smallest Duration that is strictly positive without overriding
+/// validated configs from the control plane (CP should reject decay=0).
+pub const MIN_DECAY: time::Duration = time::Duration::from_millis(1);
```
cratelyn: > Chosen as the smallest Duration that is strictly positive without overriding validated configs from the control plane (CP should reject decay=0).

i was having trouble parsing this sentence upon my first read. to repeat my understanding of this, the control plane won't allow any annotations that configure a minimum under 1 millisecond (e.g. 500ns).

is that correct? just to connect the dots, where does this validation happen?

linkerd/http/classify/src/retry_after.rs
```diff
@@ -0,0 +1,320 @@
...
+/// - Positive i64: retry after this many milliseconds
+/// - Negative i64: do not retry (returns `None`)
```
cratelyn: i was having trouble finding the spec in question. @adleong pointed me to https://github.com/grpc/proposal/blob/master/A6-client-retries.md. it would be helpful to put this link here.


```suggestion
/// Per the [gRPC A6 spec]:
///
/// > The value is to be an ASCII encoded signed 32-bit integer with no unnecessary leading zeros
/// > that represents how many milliseconds to wait before sending a retry. If the value for
/// > pushback is negative or unparseble, then it will be seen as the server asking the client not
/// > to retry at all.
///
/// [gRPC A6 spec]: https://github.com/grpc/proposal/blob/master/A6-client-retries.md#pushback
```

cratelyn (commented):
questions

linkerd/ewma/src/lib.rs
```diff
@@ -0,0 +1,547 @@
...
+        self.value
+    }
+
+    /// Returns the timestamp of the most recent update.
+    pub fn last_update(&self) -> time::Instant {
+        self.timestamp
+    }
+
+    /// Returns the decayed value projected to the given time, without modifying stored state.
+    ///
```
adleong, line 77: I believe this is mathematically equivalent to updating the stored state and updating the stored timestamp, which is what Tower's PeakEwma does: https://tower-rs.github.io/tower/src/tower/load/peak_ewma.rs.html#228

While reviewing, I got tripped up by this change and it took me quite a while to convince myself that they are actually equivalent.  Is there a motivation for doing this differently than how Tower PeakEwma does this (and consequently how Linkerd did it prior to this change)?

linkerd/ewma/src/lib.rs
```diff
@@ -0,0 +1,547 @@
...
+        if self.value.is_infinite() || self.get_at(ts) < value {
+            self.value = value;
+            self.timestamp = ts;
+            return;
+        }
+        self.add(value, ts)
+    }
+
+    /// Computes 1/elapsed since the last update and feeds it through `add()`.
+    pub fn add_rate(&mut self, ts: time::Instant) {
```
adleong: I couldn't find where this is used.  Is it unused?

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
+//! Load tracking with response failure awareness.
+//!
+//! This module provides a `LoadBiaser` wrapper that tracks request latency (RTT)
+//! and detects failure responses (HTTP 429, 503, 5xx), applying artificial
+//! penalties. Unlike Tower's `PeakEwma`, this implementation uses
+//! `linkerd_ewma::Ewma` and returns `f64` metrics directly, enabling
+//! integration with P2C load balancing.
+//!
+//! This can wrap any service (`Load` trait not required) and it tracks RTT via
```
adleong: Why do we specify here that the inner service doesn't need the Load trait?  It's this Service that impls Load, not the inner one, just like in Tower Ewma: https://tower-rs.github.io/tower/src/tower/load/peak_ewma.rs.html#114

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
...
+#[cfg(feature = "tokio-test")]
+impl ResponseFailureHint for tokio_test::io::Mock {}
+
+/// Amplification factor for Retry-After penalties.
+///
+/// When a 429 or 503 response includes a Retry-After header, the load biaser injects
+/// an amplified penalty so it remains meaningful through the server's requested
+/// avoidance window. The injected value is:
+///
+///   `penalty_secs * RETRY_AFTER_PENALTY_FACTOR * exp(retry_after / penalty_decay)`
```
adleong: I don't understand how we came up with this formula.  Running through an example, if we have `penalty=5s`, `PENALTY_FACTOR=0.5`, `penalty_decay=10s` and we receive a response with a `5s` retry-after hint, then the injected penalty value would be ~4.12s.  That penalty value would decay to 2.5s after the 5s `retry-after` duration and then continue to decay towards 0 after that.

Compare that to the base penalty of `5s` for failures without retry-after hints.  This means that traffic would be weighted towards endpoints whose responses returned failures with retry-after hints over endpoints whose responses did not provide hints.  This isn't what I would expect.

More generally, what is the relationship between the retry-after hint and the load function desired to be?  This seems like it's engineered to have a certain property, but it's not clear what that property is.

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
...
+///
+/// which decays via EWMA to `penalty_secs * RETRY_AFTER_PENALTY_FACTOR` at
+/// exactly `t = retry_after`. The factor controls how aggressively the endpoint
+/// is avoided near the Retry-After deadline:
+///
+/// At 1.0 the endpoint is fully avoided (~0% traffic) through the window and 46s
```
adleong: In order for an endpoint to be fully avoided, it needs to have a higher load estimate than any other endpoint so that the P2C algorithm will never choose it.  With a `PENALTY_FACTOR=1.0`, I don't see how that would be true.

How was the 46s duration computed? What does this mean?

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
...
+/// beyond -- too aggressive for early-recovery discovery. At 0.1, traffic leaks
+/// well before the deadline. We use 0.5: the penalty at deadline is penalty_secs/2
+/// (roughly 50x healthy load in a 5-endpoint pool), which lets occasional probes
```
adleong: I don't understand this comment about the penalty being roughly 50x health load in a 5-endpoint pool.  What does that mean and how was it computed?

adleong: The ability for requests to go to the penalized endpoint will depend on how that endpoint's load estimate compared to the load estimate of the other endpoints in the balancer.  I don't think we can easily say anything about how frequently requests will do to this endpoint without knowing the global state of the balancer.

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
...
+/// through in the second half while still exceeding a plain failure penalty for
+/// RA >= ~7s. For pools with 100+ endpoints the factor barely matters because P2C
```
adleong: Where does the 7s retry-after value come from here?

adleong: The comment about endpoints in large pools being rarely drawn is confusing to me.  All endpoints in the pool have an equal chance to be drawn and all endpoints will have a load function that is computed when they are drawn.  I don't see why this would barely matter.

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
...
+    /// The penalty value to inject on failure responses (429, 503, 5xx) in seconds
+    pub penalty_secs: f64,
+
+    /// Decay duration for the penalty EWMA.
+    /// Controls how quickly the penalty decays after a failure response
+    pub penalty_decay: Duration,
+
+    /// Whether load biasing penalties are enabled. When false, only RTT tracking
+    /// is active (PeakEwma equivalent).
+    pub enabled: bool,
```
adleong: I suspect this boolean can be removed once the load-biaser is updated to be an alternative to tower's PeakEwma instead of being a drop-in replacement with additional functionality.

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
...
+/// Combines the EWMA trackers (under a mutex for atomic RTT+penalty reads),
+/// the in-flight request counter, and the immutable config fields that every
+/// response future needs. One allocation per endpoint instead of two, and
+/// futures carry a single `Arc` clone instead of copying config fields.
+#[derive(Debug)]
+struct SharedState {
+    /// RTT and penalty EWMAs; read-locked in load(), write-locked in poll()
+    metrics: RwLock<LoadMetrics>,
+    /// Count of in-flight requests (up on call, down on response)
+    pending: AtomicU32,
```
adleong: Instead of maintaining a dedicated atomic counter, tower's PeakEwma uses the strong reference count to know how many pending requests there are: https://tower-rs.github.io/tower/src/tower/load/peak_ewma.rs.html#138

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
...
+        let penalty = if penalty_val.is_infinite() {
+            0.0
+        } else {
+            penalty_val
+        };
+
+        // Load = RTT * (pending + 1), minimum will be `penalty`
+        // The +1 ensures idle endpoints have some load based on RTT
+        let base = rtt * f64::from(pending.saturating_add(1));
+        let load = f64::max(base, penalty);
```
adleong: Comparing `base` and `penalty` here seems really strange.  Base is the load estimate, computed by multiplying the current rtt estimate by the number of pending requests.  On the other hand, the penalty is a fixed penaty value, scaled by a constant `PENALTY_FACTOR` and an exponential factor depending on the retry-after hint.

For example, if the rtt estimate is `100ms` and there are 50 pending requests, then `base=5s` and (with the example from earlier) the penalty value could be `4.12s`.  This demonstrates that when `rtt` or concurrency is high, the penalty is ignored entirely.

Thinking in terms of dimensional analysis, `base` has units of `Duration * Concurrency` whereas the penalty just has units of `Duration`.  I wouldn't expect to be able to compare these values.

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
...
+                            FailureHint::InternalError => base_penalty,
+                        };
+
+                        tracing::debug!(
+                            penalty_secs = penalty_val,
+                            rtt_secs = elapsed,
+                            ?hint,
+                            "Detected failure response - injecting load penalty"
+                        );
+                        metrics.penalty.add_peak(penalty_val, now);
```
adleong: Whenever we encounter failures without hints, we will `add_peak` with the base penalty value (e.g. `5s`).  This means that, in the absence of retry-after hints, the penalty estimate will always be equal to the base penalty.

On the other hand, when we do encounter retry-after hints, we add a measurement of a scaled penalty value which may be higher or lower than the previously measured value depending on the retry-after hint and the `PENALTY_FACTOR`.  I'm not sure what properties this has or why we want it to work this way.

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
...
+        Poll::Ready(result)
+    }
+}
+
+#[pin_project::pinned_drop]
+impl<F, Rsp> PinnedDrop for LoadBiaserFuture<F, Rsp> {
+    fn drop(self: Pin<&mut Self>) {
+        let this = self.project();
+        // Only decrement if we haven't already done so in poll().
+        // Avoids leaking the pending count upon cancellation.
```
adleong: I think this can be avoided by using the Arc strong reference count, as suggested above, instead of a dedicated atomic counter.

adleong (changes requested):
I spent a lot of time trying to understand the math here.  I left specific questions inline, but at a higher level I think this is a lot more complex than it needs to be.

When we get a failure response, can't we just record an rtt observation of `retry-after` (or the base penalty if no retry-after hint is present) instead of the actual rtt?  This means that, from the load balancer's perspective, when we get a response that returned in 1ms with a retry-after of 5s, it's treated as a response which returned in 5s.

This is vastly simpler and gives us the desired property that overloaded endpoints which return quickly don't get artificially higher load due to their quick responses.

This implementation would then be nearly identical to Tower's PeakEwma.  The only difference being logic in the Handle drop to record a penalty value instead of the actual RTT: https://tower-rs.github.io/tower/src/tower/load/peak_ewma.rs.html#292

linkerd/ewma/src/lib.rs
adleong: Upon reviewing this more closely, I don't think they're actually equivalent.

Tower's version decays on each read *and updates the timestamp*.  This version projects the decay without updating the value in place *and does not update the timestamp*.  This matters when determining the elapsed time used for weighting during `add`.  Tower's version will use elapsed since last `load()` whereas this version uses elapsed since last `add`.

Arguably, this version is better since `load()` is more of a read operation than a write one.  But this is a subtle behavior difference between Tower and this code that could affect load balancing.

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
...
+        shared.pending.fetch_sub(1, Ordering::Release);
+        *this.completed = true;
+
+        Poll::Ready(result)
+    }
+}
+
+#[pin_project::pinned_drop]
+impl<F, Rsp> PinnedDrop for LoadBiaserFuture<F, Rsp> {
+    fn drop(self: Pin<&mut Self>) {
```
adleong: Also worth noting that Tower's PeakEwma records an RTT observation on cancelation: https://tower-rs.github.io/tower/src/tower/load/peak_ewma.rs.html#292 and this code does not.

Not sure which behavior is better, but worth noting this subtle behavioral difference.

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
...
+impl<F, Rsp, E> Future for LoadBiaserFuture<F, Rsp>
+where
+    F: Future<Output = Result<Rsp, E>>,
+    Rsp: ResponseFailureHint,
+{
+    type Output = F::Output;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        let mut result = ready!(this.inner.poll(cx));
```
adleong: Note that Linkerd currently measures RTT by time to first data body:

https://github.com/BuoyantIO/buoyant-proxy/blob/buoyant/main/linkerd/proxy/http/src/balance.rs#L4

That's different from what's here where we measure RTT by time to response future complete.

linkerd/ewma/src/lib.rs
unleashed: Close to correct, IIRC the last changes I worked on the CP actually check only for negative decay values and that comment is now stale,  since the CP won't enforce MIN_DECAY there (it'd be adding an extra detail we don't really need there). We enforce it here in the Ewma constructor instead.

linkerd/ewma/src/lib.rs
unleashed: Intended for future use and eventually discarded. Will remove.

linkerd/ewma/src/lib.rs
unleashed: Yes, Tower's implementation both decays and updates the timestamp on read, whereas here we project at the provided instant without modifying the value or timestamp. The only difference when using Tower's implementation versus linkerd-ewma (replaced later on) is that on each read Tower's decays and updates the timestamp, so it weights the next request completion by time elapsed since the last load() call, and linkerd-ewma weights by time elapsed since the last add() call. In normal RTT conditions and when moving trending upwards these should be equivalent (due to hitting the peak branch), and with varying RTTs the Tower implementation looks a bit more reactive to newer samples due to calling load() and updating the Tower's Ewma in dependency with the amount of P2C endpoints and request rate (both impact the amount of load() calls, and having the Ewma depend on that looks extremely wrong to me), but I would need to test both head to head to make sure on how they vary across different scenarios.

The read-only path also avoids P2C Ewma checks being serialized due to writes.

linkerd/http/classify/src/retry_after.rs
unleashed: Added @cratelyn's suggestion.

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
...
+
+/// Connection tuples (Connection, Metadata) used by TCP/TLS paths never indicate failures.
+/// This is meant for MakeConnection services that return `(I, M)` tuples.
+impl<C, M> ResponseFailureHint for (C, M) {}
+
+/// TCP streams never indicate failures (no HTTP status codes).
+impl ResponseFailureHint for tokio::net::TcpStream {}
+
+/// Duplex streams (used in testing) never indicate failures.
+impl ResponseFailureHint for tokio::io::DuplexStream {}
```
cratelyn: i am concerned about these implementations. as i understand it, `NewLoadBiaser<N, Req>` is a `NewService<T>` layer that instruments the emitted `Service<T>`s with a `LoadBiaser<N, Req>` wrapper.

i see why we want this trait to be implemented for `http::Response<B>`, because a response has a status code and headers that we would wish to inspect in order to compute our rate limit hint.

what i'm having trouble understanding is why we would ever want to install a service upon a service that yields `TcpStream` responses. if we add this layer to such a service, what purpose would it serve?

that concerns me because, unit tests aside, it could lead to situations in which it is easy to install this layer into a stack and not realize that it is unintentionally inert. removing these seems like it would ensure that the load biaser is correctly placed such that it inspects HTTP/gRPC traffic.

linkerd/ewma/src/lib.rs
```diff
@@ -0,0 +1,547 @@
...
+        }
+
+        self.value = {
+            let elapsed = ts.saturating_duration_since(self.timestamp);
+            let alpha = 1.0 - (-elapsed.as_secs_f64() / self.decay).exp();
+            self.value * (1.0 - alpha) + value * alpha
+        };
+
+        self.timestamp = ts;
+    }
```
cratelyn, line 113: i had a lot of difficulty understanding the difference between `add()`, `add_peak()`, and `add_rate()`. i don't see a clear reason why we would ever want to update the weighted moving average but ignore a peak value.

upon inspection, i noticed that this method only seems to be called from within the test suite, with all external calls in the load biaser using `add_peak()`, which underlines that intuition. only the test suite appears to use `add()`, but nothing there seems intent on ignoring peaks either.

to help keep you moving forward on this work, i've taken the time to craft a pull request that performs this change, in https://github.com/linkerd/linkerd2-proxy/pull/4548.

linkerd/ewma/src/lib.rs
```diff
@@ -0,0 +1,547 @@
...
+
+    /// Returns the current value of the average.
+    pub fn get(&self) -> f64 {
+        self.value
+    }
+
+    /// Returns the timestamp of the most recent update.
+    pub fn last_update(&self) -> time::Instant {
+        self.timestamp
+    }
```
cratelyn: i found myself a little confused about what the purpose of this interface is, and determined that it isn't used externally in the load biaser.

https://github.com/linkerd/linkerd2-proxy/pull/4549 moves this behind a `#[cfg(test)]` gate, to signal that this is an accessor to facilitate testing, like we do in the `load-biaser` crate.

linkerd/ewma/src/lib.rs
cratelyn: okay, can we please update this comment so that it is not stale?

linkerd/load-biaser/src/lib.rs
unleashed: Sorry, this should be a constant, although I was not sure of the actual value I'd assign. So this is not a duration, it's a camp on the injected penalty value. Since the penalty decay is now configurable, using a small value for it would make `exp(retry_after / decay_secs)` overflow to infinity, and we treat an infinity penalty as unset, meaning no penalty. This way we ensure we don't accidentally remove the penalty to something that should be heavily penalized. With that said, I am thinking on reworking this part based on comments by @adleong I still need to review.

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
...
+        }
+        // Try gRPC retry-pushback-ms (for trailers-only responses)
+        if self.status() == http::StatusCode::OK && is_grpc(self.headers()) {
+            if let Some(d) = linkerd_http_classify::retry_after::parse_grpc_retry_pushback(
+                self.headers(),
+                Duration::MAX,
+            ) {
+                self.extensions_mut().insert(CachedRateLimitHint::new(d));
+            }
+        }
```
cratelyn, line 187: i'm having trouble understanding how this would handle trailers.

https://docs.rs/http/latest/src/http/response.rs.html#342-344

i checked the `http` source very cursorily, wondering if responses with empty bodies might fold trailers into the headers returned by `rsp.headers()`, but i don't see anything to indicate that.

am i missing something, or wouldn't we need to either make use of a `Body` middleware to signal when we've inspected the trailers? that seems like it would introduce more complexity to this pull request, so alternatively we could consider documenting that this solution does not yet inspect gRPC trailers like `retry-pushback-ms` and treat that as a follow-up addition.

linkerd/load-biaser/src/lib.rs
cratelyn: i presume the `tokio_test::io::Mock` impl below could also be removed, should we go forward with this?

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
...
+
+    /// Returns the rate limit hint if available.
+    ///
+    /// Checks cached value first (from a previous `attach_parsed_rate_limit_hint` call).
+    /// If no cached value, attempts to parse the header directly (capping at `max`).
+    /// Returns `None` only if the header is absent or unparseable.
+    fn rate_limit_hint(&self, _max: Duration) -> Option<Duration> {
+        None
+    }
+}
```
cratelyn, line 113: [below](https://github.com/linkerd/linkerd2-proxy/pull/4537/changes#r3342274092), i note that it seems like we're implementing this trait for an overly broad collection of "response" types. if we're dealing with a response type that does not have a coherent notion of failure, my understanding is that this middleware would not do anything meaningful.

presuming we can remove those transport-level `ResponseFailureHint` impls, we may also be able to remove these default `None` implementations. this would further guard against the load biaser being installed on services whose responses we cannot extract failure hints.

linkerd/load-biaser/src/lib.rs
cratelyn: i added a related comment [above](https://github.com/linkerd/linkerd2-proxy/pull/4537/changes#r3342923901) about this trait.

linkerd/load-biaser/src/lib.rs
adleong: My understanding of HTTP/2 is that when there are no DATA frames in a message, all headers and trailers are typically sent in the same HEADERS frame.

linkerd/http/classify/src/retry_after.rs
unleashed: I'd leave the "maximally overloaded" part out (letting the circuit breaker logic handle the backoff after a None here) and just update instead the comment to clarify we won't and can't honor a "do-not-retry" request (for the reason you mention).

linkerd/load-biaser/src/lib.rs
unleashed: Ok, as written this may be confusing. It's just referencing the fact that the type `S` has no bounds at the language level, with us implementing the `Load` trait ourselves, so if you had a service you could skip implementing `Load` on `S` (or remove the impl altogether) if you wished. This is probably not worth even specifying here, I think I'll just remove it.

linkerd/load-biaser/src/lib.rs
unleashed: What the function should do is make a failing or rate-limited endpoint look more loaded (less attractive to P2C) than healthy ones, with a longer server-requested backoff producing stronger/longer avoidance, while behaving like PeakEwma when everything is healthy. The amplified penalty formula doesn't achieve that cleanly, since below about 7s of Retry-After (at the default factor/decay) the injected penalty is smaller than the plain failure penalty, so it inverts. Retuning the formula to accomplish this, given the configurable parameters while keeping some defaults, seems difficult enough given the potential for widely different retry-after policies. Rathat than doing that I'll explore changing the mechanism as you suggest elsewhere, using the Retry-After (or the base penalty) as an effective RTT through the normal EWMA, so load is monotonic in the RA value by construction and there's no separate penalty term to reason about.

linkerd/load-biaser/src/lib.rs
unleashed: The comment should be "its load far exceeds a healthy peer's load", which should be the case for a lone failing endpoint among healthy ones but isn't a guarantee.

The 46s is the result of illustrating, with default values and assuming a healthy endpoint takes ~50ms to respond, why factor=1.0 is too aggressive: at factor 1.0 the penalty after the retry-after window (at tRA) is `penalty_secs` (5s default * 1.0), decaying as `5 * exp(-t / penalty_decay)` (with penalty_decay 10s default) down to a healthy baseline (`rtt * (pending + 1)`, which if we take a healthy rtt at 50ms and an empty queue with pending = 0, is ~0.05), so that `t = 10 * ln(5/0.05)`, ie. setting the decayed penalty to the baseline, `5*e^(-t/10)`, t is ~46.05 seconds. That is 46s of extra avoidance past the retry-after window assuming the endpoint would be considered healthy at those 50ms of latency. `penalty_secs` and `penalty_decay` are configurable, so that number moves with the configuration parameters, and it was computed at the defaults.

This is an imperfect solution since some parameters are not configurable and some others are, so any choice will have trade-offs (in fact on second thought I'd probably go with a lower factor value). Opening up all parameters for configuration seems overkill.

So when there's a failure, from t0 to tRA, the intention is that the penalty is large enough and decays to `penalty_secs * FACTOR` at tRA, so the endpoint is avoided. After tRA, the penalty keeps decaying but is still above what a healthy endpoint's load looks like (ie. 0.05 in our case), so the effect is that the endpoint spends more time without being selected by P2C provided its peers are all healthy. With the default values above, we'd need 46s after tRA to look like a healthy endpoint (but note that this changes the moment you have other penalized endpoints). So the comment basically says that the 1.0 factor is too aggressive.

With that said, the lack of more detail and assumptions on default values, plus lacking an explanation like the text above should be fixed. However, doing that inline in these comments, with the additional considerations on the factor, seems a bit excessive, so not sure what to do about it.

linkerd/load-biaser/src/lib.rs
unleashed: "~50x healthy load" is the tRA penalty relative to a healthy endpoint's P2C load taking default values (see comment above). At the 0.5 factor the tRA penalty is `penalty_secs * 0.5 = 2.5`, the healthy baseline is `rtt * (pending+1)` for a fast, idle endpoint, taken as ~0.05 (50ms RTT, empty queue). So 2.5/0.05 is 50x. Like the above 1.0 factor comment, this is illustrating the ratio at the default but configurable `penalty_secs` and that baseline, not a fixed property.

unleashed: Agreed. The 5 endpoint pool is mostly decoration, except for just statistics doing their thing in comparison to a large pool: with a 100+ endpoint pool, without knowing anything else about it, the odds of having a specific endpoint selected are smaller. Additionally, the expectation is that it's going to be drawn less since it's more likely that it will be compared to healthier endpoints, ie. easier to be compared to other unhealthy endpoint and drawn for a generally healthy pool (ie. 3 healthy vs 2 unhealthy, where pairing the 2 unhealthy is likely, compared to 98 healthy vs 2 unhealthy, where it's relatively unlikely).

linkerd/load-biaser/src/lib.rs
unleashed: The 7s is the Retry-After above which the amplified penalty finally exceeds a plain failure penalty: `penalty_secs * 0.5 * exp(RA/10) > penalty_secs`, so that solving for RA you get `RA > 10*ln 2`, ~6.9s (at the default factor 0.5 and decay 10s). So it's the boundary of the inversion you identified on the formula above. Below it, a retry-after failure is penalized less than a plain failure.

This problem (and the associated complexity) makes me consider switching to adopting your suggestion and using the hint as effective RTT so it behaves monotonically, without a threshold.

unleashed: I think it's easy to read the comment as attributing importance to the factor regarding pool size, as if an endpoint in a pool would be drawn less because of it. That's not what I mean, pool size doesn't change how much the factor matters.

So the factor sets the penalized endpoint's load, and that load decides whether it loses its P2C comparison (with P2C routing to the lower load of the two it draws, a high enough load means it gets skipped). That comparison works the same way regardless of pool size. What pool size does change is how often that endpoint is drawn. P2C samples 2 of N per request, so in a large pool any single endpoint is in fewer comparisons and affects fewer routing decisions overall, so it's just a matter of probability.

linkerd/load-biaser/src/lib.rs
unleashed: Yes.

linkerd/load-biaser/src/lib.rs
unleashed: Good catch, this is a bug.

The penalty was initially meant as a floor so that endpoints would be penalized by a constant, fixed value, so right after a failure the endpoint's load was at least the penalty, avoiding the problem of 503s or 429s treated as low load, fast endpoints. However, as evidenced by your dimensionality analysis, it's not scaling the penalty accordingly.

Something closer to `max(rtt, penalty) * (pending + 1)` would be conceptually better, though I would need to take a closer look. We are basically looking for an effective RTT that should scale with concurrency, adding in a scaled penalty for failures.

linkerd/load-biaser/src/lib.rs
unleashed: The intent of `add_peak` here is that a failure should push the endpoint's penalty up and let it decay, rather than be averaged away by the next fast response. So without hints the estimate sits at an arbitrary base penalty, and a hint pushes it higher (or, as you found out, sometimes lower, which is problematic). I think I'm going to take your suggestion elsewhere and have no separate penalty estimate at all so that the failure observation (hint or base penalty) goes into the RTT Ewma via `add_peak`, so it's would just be "a failure looks like a slow response that then decays" which should be easier to reason about.

linkerd/load-biaser/src/lib.rs
unleashed: That's interesting. I'll note this to go over it when revisiting the effects on the drop implementation and cancellation (and maybe on Clone impls).

linkerd/load-biaser/src/lib.rs
unleashed: Probably so. I'll need to think about the consequences of such a change at some point later.

linkerd/load-biaser/src/lib.rs
unleashed: This is important, thanks for noting it.

I thought about this initially and went over the case of "shutting down" and cancelling. However this could also be a cancellation due to a timeout, which I suspect is the edge case Tower's implementation is catching, in which case they would just give up and take the time since it was sent as RTT, which would help identify an unhealthy endpoint.

I'd say we want to take the same approach.

linkerd/load-biaser/src/lib.rs
unleashed: Yes, this was chosen intentionally because it simplifies the implementation and allows to update RTT and penalty at once under a single write lock. The real difference here would be caused by responses that would return headers quickly (which is always <= than the time to first data frame) but then have slowly computed bodies, responses with large bodies or streaming responses, in which there may be a difference between the time to first message (or useful byte) and the time to head. So that is essentially the case in which headers are received quickly, but then there's a lapse of time until data arrives.

This was a trade-off that changes the measure under the specific scenario above (basically measuring faster RTTs in those cases), with the assumption that those are not common/important enough, and given the code already has what it needs to work at that point, I went with it.

However, this can be improved and made to measure RTT by time to first data body, only with more complexity and having to deal with stuff like `load()` seeing an inconsistent state, so likely requiring more expensive sync.

I'll note this to revisit when considering other changes to the biaser.

linkerd/load-biaser/src/lib.rs
unleashed: Yes, it may sound counter-intuitive, but a "trailers-only" gRPC response sends all the information in HEADERS.

In cases where there are actual trailers in DATA frames the biaser won't see those. That would require a Body middleware, which is scope creep hard to justify given how this feature is mainly motivated by HTTP response headers. The comment is basically implying this, as a best-effort way to support one of the cases where it's cheap for us to determine there's a pushback hint.

N.B: OTOH the circuit breaker has access to the body and will look at the pushback hint.

linkerd/ewma/src/lib.rs
unleashed: `add` itself needs to stay public, since it's later on used in #4546, in the circuit breaker.

Regarding the differences, `add` always moves the estimate towards the new value, and the amount it moves it depends on the elapsed time since the last update (moving average), so that a higher sample moves it up, a lower one moves it down. That's what we use in the breaker.

On the other hand `add_peak` checks if the new value is above the current estimate, and if so it won't move the estimate towards that value, but jump straight to it (which helps with penalties in the load-biaser, as we don't care that an endpoint was quick in the past if it just started returning 503s or 429s), and if the new value isn't above the estimate, it just behaves like `add`, moving the estimate down slowly.

Not sure if this is the correct level of detail for updating the comments. Perhaps we can repurpose #4548 to improve them?

linkerd/ewma/src/lib.rs
unleashed: This was expected to be used later on in breaker code, but so far I've not had to use it (the breaker reads the decayed value via `get_at`, and updates through `add`, so it never needs the raw timestamp). I think #4549 is a good idea (couldn't review yet).

linkerd/load-biaser/src/lib.rs
unleashed: Agreed, these are speculative impls at this point, I ended up not needing them. The Mock impl can be removed too.

linkerd/load-biaser/src/lib.rs
unleashed: Yes, we can do that after removing the impls, since we'll just have the `http::Response<B>` impl.

linkerd/ewma/src/lib.rs
unleashed: Noted, will do when I get back at reworking this.

linkerd/ewma/src/lib.rs
unleashed: Additionally we can remove `add_rate` and any related tests.

linkerd/load-biaser/src/lib.rs
cratelyn: makes sense. if the convention for these responses is that they contain no DATA frames, this should be fine then!

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
...
+            inner,
+            shared,
+            config,
+        }
+    }
+
+    /// Returns a reference to the inner service.
+    pub fn get_ref(&self) -> &S {
+        &self.inner
+    }
```
cratelyn: is this used anywhere? i couldn't find anywhere it's used, perhaps this can be removed.

linkerd/load-biaser/src/lib.rs
cratelyn: i am working on applying this suggestion now. in doing so, i believe that i have found a likely bug with how `pending` is tracked. i'll have a pull request up with more info later today.

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1701 @@
...
+        biaser.shared.pending.fetch_add(1, Ordering::AcqRel);
+
+        // RTT * (1 + 1) = 0.05 * 2 = 0.1
+        let load_one_pending = biaser.load();
+
+        // Increment again
+        biaser.shared.pending.fetch_add(1, Ordering::AcqRel);
+
+        // RTT * (2 + 1) = 0.05 * 3 = 0.15
+        let load_two_pending = biaser.load();
```
cratelyn: see https://github.com/linkerd/linkerd2-proxy/pull/4551.

i added some `assert_eq!` statements here to confirm that this math was correct.

these comments do not accurately report the load values here.

linkerd/load-biaser/src/lib.rs
cratelyn: https://github.com/linkerd/linkerd2-proxy/pull/4556 i investigated this, there's a pull request here. i did not track down all of the test changes that would be needed to make CI green, but am providing this here to save you some trouble.

cratelyn: > i am working on applying this suggestion now. in doing so, i believe that i have found a likely bug with how `pending` is tracked. i'll have a pull request up with more info later today.

i'm also a bit worried that we don't share our `pending` count correctly. it looks like we only store this state at the request level, increment it from 0 to 1, and back down to 0 after a response is yielded.

i'm not sure that, as written, this would properly track `pending` across all of the in-flight requests for a given endpoint.

linkerd/ewma/src/lib.rs
cratelyn: > on each read Tower's decays and updates the timestamp, so it weights the next request completion by time elapsed since the last load() call, and linkerd-ewma weights by time elapsed since the last add() call.

i'm not sure that this is correct. i believe i see some confusion arising from this characterization that tower mutates this estimate "on read", so i spent some time looking into this. on further inspection, i believe that @adleong is correct that these are _not_ behaviorally equivalent, and i'm thus having trouble seeing why we need to reimplement a different EWMA middleware.

in tower's implementation the estimate is updated here, in `PeakEwma::update_estimate()`:

https://github.com/cratelyn/tower/blob/251296dc54a044383dffd16d2179b443e2615672/tower/src/load/peak_ewma.rs#L156-L159

this then takes the decay period, obtains the current timestamp, and uses this to update the weighted moving average:

https://github.com/cratelyn/tower/blob/251296dc54a044383dffd16d2179b443e2615672/tower/src/load/peak_ewma.rs#L227-L282

as has been pointed out, this `update_estimate()` function is called when `Load::load()` is called, usually by a `tower::balance` load-balancer:

https://github.com/cratelyn/tower/blob/251296dc54a044383dffd16d2179b443e2615672/tower/src/load/peak_ewma.rs#L140-L142

however, this estimate is also updated when a `Handle` is dropped.

https://github.com/cratelyn/tower/blob/251296dc54a044383dffd16d2179b443e2615672/tower/src/load/peak_ewma.rs#L287-L295

this happens whenever a `PeakEwma` service is `call()`ed, and the resulting `TrackCompletionFuture` yields a `S::Response`.

https://github.com/cratelyn/tower/blob/251296dc54a044383dffd16d2179b443e2615672/tower/src/load/completion.rs#L80

in other words, this would update the estimate when (a) the service yields a response, _or_ (b) when `Load::load` is called. i want to be very precise here and differentiate the notion of a "read" from `load()`. note that the `tower::load::peak_ewma::Cost` metric doesn't have a meaningful interface besides its derived `PartialOrd` implementation, we don't call that ourselves, this happens inside of a middleware layer like `tower::balance::p2c::Balance`:

https://github.com/cratelyn/tower/blob/251296dc54a044383dffd16d2179b443e2615672/tower/src/balance/p2c/service.rs#L168-L171

my understanding is that we are working around the inability to access the RTT estimate for our own purposes, so that we can compute a load function that takes into account both this RTT estimate _and_ additional penalties for particular status codes like 429's or 5XX's.

i've put together https://github.com/tower-rs/tower/pull/871.

i'd like some help understanding why this wouldn't be suitable for our purposes. this would let us access the RTT estimate. if we want to strictly preserve the existing behavior, we can call `load()` and let the estimate be updated. if we want to avoid load calculations updating the estimate and its timestamp, we can opt not to call `load()`. my personal preference would be to preserve behavior as much as possible.

unleashed:
Addressed quite a few comments with a large refactoring. The ewma should now be way closer to Tower's, and the load biaser is now using a single RTT ewma, retry-after/grpc-pushback hints now behave monotonically, the measurement is taken by default when the first data frame arrives, and we use a u32 instead of a float in the config allowing us to derive Hash/Eq.

I think this may close a few of the open PR's against this one, @cratelyn. Probably a good moment to take another look?

linkerd/ewma/src/lib.rs
unleashed: > my understanding is that we are working around the inability to access the RTT estimate for our own purposes, so that we can compute a load function that takes into account both this RTT estimate and additional penalties for particular status codes like 429's or 5XX's.

Yes, this is the main motivation.

> i've put together tower-rs/tower#871.
>
> i'd like some help understanding why this wouldn't be suitable for our purposes. this would let us access the RTT estimate. if we want to strictly preserve the existing behavior, we can call load() and let the estimate be updated. if we want to avoid load calculations updating the estimate and its timestamp, we can opt not to call load(). my personal preference would be to preserve behavior as much as possible.

After the refactor I posted to converge on a single estimate for both RTT and penalties, the reason that we still can't drop the crate is the need to update the estimate with a penalty. The load biaser records a failure as either a hint-based or a penalized RTT by adding a peak. So a fast failure raises that endpoint's effective RTT and then decays like any real sample, flowing through the same `rtt * (pending + 1)` cost. PeakEwma takes its sample only from elapsed time (`recv_at - sent_at`) inside a private `update()` function, so there's no facility to inject a value, and a custom `TrackCompletion` can only delay the `Handle` drop, not change the magnitude. tower-rs/tower#871 gives us the read side we want, but not the write side.

With just tower's 871 in, the best we could do is read tower's PeakEwma RTT and keep the penalty in a separate decaying accumulator that would need its own separate locking, recombined at load() time. It would walk back part of the refactor I just finished to have the RTT and the penalty combined into a single RTT estimate scaling with the amount of pending requests.

unleashed: Besides needing to combine the penalty, which is the main obstacle, there's the update-on-read issue (and as a part of that, the locking). This is ultimately a correctness issue, not a requirement that prevents us from using PeakEwma.

Let me clarify what I mean: when I say that Tower's PeakEwma updates on read, what I mean is that when the balancer calls `load()` on it, which on paper sounds like a read-only operation, it computes but also updates its state with an artificial observation, which is conceptually wrong (there's nothing to measure) but importantly/happily doesn't corrupt the estimate value because it sees 0ms elapsed between `send_at` and `recv_at`.

Our implementation doesn't update the state at that point, we do so when we receive a response (and now also in cancellation/errors), which provides a real measure. Tower's implementation also writes separately when it receives a real response. The difference is Tower performs an additional write per `load()` with caveats (see below), which scales with requests, and we don't.

What we do in our implementation is a read-only operation: projection (via `get_at`), which is to say we compute the same as Tower except we avoid storing the estimate and updating the `updated_at` timestamp, so we can assess the load and avoid writing. We don't really need to write anything when assessing the load because we only write when we see real measurements, taking a read lock. It's only when we have an actual measurement that we take the write lock and update the internal state.

There are a couple of consequences to the way Tower is doing this that I think are not desirable. First, it is potentially having more contention, but that depends on how we use it. I _think_ we don't have concurrent P2C load calls, and any load-read vs measure-write contention also exists with RwLocks (contention characteristics may be a bit different though).

Second, Tower's version creates an unnecessary 0ms measurement, but while this has the (accidental?) property that being time-based it does not impact the accuracy of the estimate, it still has an undesired effect in the `updated_at` timestamp, which is reset to whatever time we call `load()`. That timestamp is guaranteed to be always greater than or equal than the equivalent in our implementation, which translates to giving any real measurement (unless a peak) the same or less weight than it should.

As an example of a scenario where this makes a difference: when having a period of silence on a given endpoint that's been slower than all the others in a pool, that is, it's received no requests, a new request being suddenly routed to it (say, if another endpoint just becomes slower and is compared to it by P2C), and with the endpoint responding successfully, a new real measurement will take place. That measurement should have a relatively significant weight in the estimate, since the endpoint has seen nothing but this request/response for a while. At that point this implementation assigns the weight correctly: the endpoint's load estimate should move sharply, according to the weight. However PeakEwma's barely moves the estimate because the endpoint will presumably have been routinely checked by P2C during that period as a result of routing other requests, causing the time since last `updated_at` (ie. last load() call) to the response to be small instead of large, thus having a much smaller weight than it should, basically being interpreted as "just another sample in the midst of many" in a short amount of time.

I'd expect in common, normal operation cases we'd see very small differences, after all the timestamp in our implementation is also updated as a function of the number of requests (or rather, responses). However, I would expect this difference to be larger in edge cases like the scenario above (after all it depends on multiple factors like the number of requests, how they are distributed in time, the pool size).

unleashed:
I think most open threads are either stale or addressed with the latest changes.

linkerd/load-biaser/src/lib.rs
unleashed: Removed with latest changes.

linkerd/load-biaser/src/lib.rs
unleashed: Should've been fixed via the `pending` refactor.

linkerd/load-biaser/src/lib.rs
unleashed: Removed via refactor to converge on a single Ewma.

linkerd/http/classify/src/retry_after.rs
unleashed: Fixed.

linkerd/ewma/src/lib.rs
unleashed: Removed.

linkerd/load-biaser/src/lib.rs
unleashed: Dropped/reworded.

linkerd/load-biaser/src/lib.rs
unleashed: Changed to suggested mechanism.

linkerd/load-biaser/src/lib.rs
unleashed: Removed.

linkerd/load-biaser/src/lib.rs
unleashed: Refactored to use `Arc::strong_count`.

linkerd/load-biaser/src/lib.rs
unleashed: Fixed.

linkerd/load-biaser/src/lib.rs
unleashed: Changed as suggested.

linkerd/load-biaser/src/lib.rs
unleashed: Please check the refactor to use `Arc::strong_count` introduced recently.

linkerd/load-biaser/src/lib.rs
unleashed: Now we also record a measurement on cancellation (including transport errors), which equates to being a very slow response measurement.

linkerd/load-biaser/src/lib.rs
unleashed: Made this first data frame measurement behavior the default (it adds a dependency on the tracker type, but I think it's well justified and convenient -- alternatively we could remove the convenience and expect users to provide the type themselves).

linkerd/load-biaser/src/lib.rs
unleashed: Removed.

linkerd/ewma/src/lib.rs
unleashed: Removed.

linkerd/load-biaser/src/lib.rs
unleashed: Removed transport + None impls.

cratelyn:
> Addressed quite a few comments with a large refactoring. The ewma should now be way closer to Tower's, and the load biaser is now using a single RTT ewma, retry-after/grpc-pushback hints now behave monotonically, the measurement is taken by default when the first data frame arrives, and we use a u32 instead of a float in the config allowing us to derive Hash/Eq.
> 
> I think this may close a few of the open PR's against this one, @cratelyn. Probably a good moment to take another look?

i've closed the following:

* https://github.com/linkerd/linkerd2-proxy/pull/4548
* https://github.com/linkerd/linkerd2-proxy/pull/4549
* https://github.com/linkerd/linkerd2-proxy/pull/4550
* https://github.com/linkerd/linkerd2-proxy/pull/4556

linkerd/load-biaser/src/lib.rs
cratelyn: d16388a00a9d6d111c870726db141c4bf59b1bee

linkerd/load-biaser/src/lib.rs
cratelyn: d16388a00a9d6d111c870726db141c4bf59b1bee

linkerd/ewma/src/lib.rs
cratelyn: thank you for taking the time to respond about this. i'm still having mixed feelings about using a different EWMA middleware here, because it might introduce behavior changes in our load balancing _besides_ the addition of load-biasing penalties.

however, at this point i'm inclined to disagree about this and move forward. this is expected to be an "alpha" feature. if we discover that some of these behavioral changes cause an adverse impact, we can walk this part back and use the same EWMA middleware that the existing mode uses. 

i'd like @adleong to read through this before i mark this thread as resolved, but otherwise i'm willing to experiment with a different EWMA for now.

linkerd/load-biaser/src/lib.rs
cratelyn: i've rebased #4551, it seems like the load values are still off by a factor of 2. i'm not sure what might be causing that, but it is one item i'd like to get sorted out before we land this.

linkerd/ewma/src/lib.rs
cratelyn: @adleong, @unleashed, and i discussed this thread of discussion in a synchronous meeting today. we agreed that so long as this new EWMA implementation is an opt-in change, we all feel comfortable with experimenting here.

with that consensus reached, i am going to mark this thread as resolved.

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1517 @@
...
+                // The completion tracker will drop the handle now or at first body
+                // frame, and that drop will record the measured RTT.
+                let output = this.completion.track_completion(handle, resp);
+
+                Poll::Ready(Ok(output))
+            }
+            Err(e) => {
+                // A transport error (connection refused, reset) or cancellation with
+                // no response to classify. Record its elapsed time.
+                drop(handle);
```
adleong, line 547: I notice that this seems to be a deliberate difference from the tower implementation.  But is there a functional difference?  In the tower case, if the inner Future fails, then the TrackCompletionFuture will also fail and presumably be dropped, dropping the held Handle along with it.  So what's the intention here as it differs from Tower?

linkerd/load-biaser/src/lib.rs
```diff
@@ -0,0 +1,1517 @@
...
+    shared: Arc<SharedState>,
+    sent_at: Instant,
+    enabled: bool,
+}
+
+impl Handle {
+    /// Records the measured elapsed time, then prevents the drop from recording
+    /// again.
+    fn record_elapsed(&mut self, now: Instant) {
+        if self.enabled {
```
adleong: I think we could clean up the need for protecting against double-recording here and simplify by having the Handle struct hold a `penalty: Option<Duration>`.  This way, when the completion Future receives a failure Response, it can set the penalty into the Handle and then drop it.

When the Handle is dropped, it can record the stored penalty, if it is `Some` and the elapsed time otherwise.

linkerd/ewma/src/lib.rs
adleong: I'll just flag that diverging from Tower's EWMA algorithm here may give us support/debugging headaches.  As @cratelyn points out above, the Load Biaser is like Tower's PeakEwma except that: it adds penalties for failures AND it updates its RTT more aggressively by not resetting the timestamp each time the load function is calculated.

I don't know what differences in behavior will manifest as a result of that change, if any.  But it may make it difficult to know if behaviors in user environments are due to the RTT algorithm change or due to the penalty injection.

It would be my preference to keep this work scoped to penalty injection and not pull in changes to the EWMA algorithm, even if we believe these changes to likely be improvements.  But since this is all isolated within the opt-in load biaser, this is a non-blocking preference.

adleong (approved):

cratelyn:
my only remaining piece of blocking feedback is https://github.com/linkerd/linkerd2-proxy/pull/4537#discussion_r3349722319.

linkerd/load-biaser/src/lib.rs
adleong: Based on my reading of this, I think the test is correct but the comments are wrong.

In the first example, we initialize the ewma with an initial value of 100ms.  Then, 1ms later, we record a new value of 50ms.  Since very little time has elapsed (1ms), the established value is weighted MUCH higher than the new value: 

```
alpha = 1.0 - (-elapsed.as_secs_f64() / self.decay).exp() 
          = 1.0 - (-0.001 / 10).exp()
          = 1.0 - 0.9999
          = 0.0001
```

so the new RTT estimate should be almost identical to the previous estimate, not the new observation.

unleashed: The problem is that the `inject_rtt()` test helper was using `add_peak()` with the default RTT of 0.1, and we were trying to set an initial value of 0.05, resulting in a no-op. This also explains why the load result was always 2x.

Fixed in 7ae6284436f43d40b4cc68b03000eb72528e13a2.

linkerd/load-biaser/src/lib.rs
unleashed: No functional difference. This is made explicit here because conceptually it's where we want the measurement to take place, but will be dropped anyway right before we return if we don't run drop at this point.

There is a subtle timing difference, where arguably we do the right thing by first measuring then constructing the retval, with the implicit case doing the reverse, but this is negligible for most intents and purposes.

cratelyn (approved):

linkerd/load-biaser/src/lib.rs
unleashed: That's way cleaner, but behavior is different. It would penalize endpoints when the completion tracker finishes, not when we know that the endpoint should be penalized.

This is not an issue in the normal case, but it would allow a struggling endpoint returning a 429/5XX status and a very slow body to defer the penalty while the balancer keeps considering the endpoint healthy, meanwhile routing traffic to it. In such a scenario it's possible or even likely the endpoint will keep struggling (perhaps even more so) with more traffic, while still not having returned the initial data frame. During that time window until the first penalty lands, which could be several seconds, the endpoint receives traffic that should have been directed elsewhere, because the growth of the load function is minimal compared to a penalty.

If we conclude that it's best to just defer this signal to the time of completion then I agree we can clean this up nicely, but I think the current behavior is useful and a good trade-off at the cost of this boolean and a bit more code.

adleong: If we use `Option<Handle>` as the H type, we can drop the `Handle` itself so that the value is recorded immediately and pass `None` into the completion tracker.

unleashed: Ok, done in fe559b5365e57942335472d997a1513bc0ecb71b. This still has a behavioral change that I think makes no real difference: since we drop the handle when penalizing, the pending count will be decreased before completion, so the load will not be exactly the same as before this change, but since the endpoint is penalized, the contribution from the pending count until it completes is relatively very small, so I think it probably does not matter.

unleashed:
Added an important fix. I think this should be ready to merge now.

cratelyn (approved):

PR #4548, opened 2026-06-02
refactor(ewma): `Ewma::add()` can be private

cratelyn:
this commit makes a refactoring proposal into
linkerd/linkerd2-proxy#4537. as such, this pull request targets the
`amr/load-biaser` branch, not `main`.

`Ewma` currently provides two public "add" methods that each update the
weighted moving average, but only one updates the peak value
bookkeeping. it's not intuitive as a consumer of this interface which of
these should be used in a given situation.

upon investigation, `add_peak()` is the only method called externally,
with `add()` only being used in the test suite. the test suite doesn't
depend on any behavior specific to `add()`, and still passes when using
`add_peak()`.

to help present a cleaner interface, this commit makes `add()` a private
method, so that callers are presented with a single interface for
updating the moving average. tangentially, see `add_rate()` being
removed in this discussion:
https://github.com/linkerd/linkerd2-proxy/pull/4537#discussion_r3337514719)

### DIFF
--- a/linkerd/ewma/src/lib.rs
+++ b/linkerd/ewma/src/lib.rs
@@ -90,10 +90,6 @@ impl Ewma {
         self.value * (-elapsed.as_secs_f64() / self.decay).exp()
     }
 
-    /// Updates the weighted moving average with a new value and timestamp.
-    ///
-    /// Precondition: `value` must not be NaN. Passing NaN poisons the EWMA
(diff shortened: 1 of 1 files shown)


cratelyn:
see https://github.com/linkerd/linkerd2-proxy/pull/4537#discussion_r3348175606.

PR #4549, opened 2026-06-02
refactor(ewma): `last_update()` is a test interface

cratelyn:
this commit makes a refactoring proposal into
linkerd/linkerd2-proxy#4537. as such, this pull request targets the
`amr/load-biaser` branch, not `main`.

`last_update()` is a method that only exists to facilitate unit tests.
accordingly, it does not need to be a public method.

to help present an intuitive interface about how `Ewma` should be used,
and to signal intent for the purpose of this accessor, we move this into
the associated test suite, behind a `#[cfg(test)]` gate.

### DIFF
--- a/linkerd/ewma/src/lib.rs
+++ b/linkerd/ewma/src/lib.rs
@@ -70,11 +70,6 @@ impl Ewma {
         self.value
     }
 
-    /// Returns the timestamp of the most recent update.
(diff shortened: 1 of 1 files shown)


cratelyn:
see linkerd/linkerd2-proxy@d7c03b51b98641337ee1042925e0c2051b66a7d4.

PR #4556, opened 2026-06-04
refactor(load-biaser): replace pending with arc strong count

cratelyn:
see
https://github.com/linkerd/linkerd2-proxy/pull/4537/#discussion_r3337713210.

this commit explores using `Arc::strong_count` for tracking the number
of pending requests.

some tests fail as a result of this change, which i haven't fully
tracked down.


cratelyn:
see linkerd/linkerd2-proxy@d16388a00a9d6d111c870726db141c4bf59b1bee.

PR #4561, opened 2026-06-08
feat(outbound): refactored unified circuit breaker with success-rate and wired up load-biaser

unleashed:
This supersedes #4546.

Here we adapt to linkerd2-proxy-api 0.20.0 (but see below for notes), with a unified breaker (opt-in) that uses both consecutive-failures and success-rate policies, supporting pushback hints. This PR also wires the load-biaser (opt-in).

All features are opt-in, with the explicit goal of leaving default behavior as-is.

[additional details on the PR TBD here -- for the time being the note below is the other most important change]

#### Note

*A relevant note on a design direction question*: the success-rate policy in this branch has been **switched from using an internal EWMA (similar to Finagle) to a time-based fixed bucket sliding window** (similar to Envoy, gRPC and others). This has been done because the EWMA has some edge cases, particularly on low traffic volume, that are cleanly solved by the sliding window due to being rate-independent:

[Quoted](https://github.com/twitter/finagle/blob/babfbf95ad63d0a380856a4851c661075fe64707/finagle-core/src/main/scala/com/twitter/finagle/liveness/FailureAccrualPolicy.scala#L204-L208) from Finagle:

> Returns a policy based on an exponentially-weighted moving average success rate over a window of requests. A moving average is used so the success rate calculation is biased towards more recent requests; for an endpoint with low traffic, the window will span a longer time period and early successes and failures may not accurately reflect the current health.

The result is that at low traffic with the exponentially moving average a new sample can skew the rate very quickly and trip due to having a large weight. The rationale for a switch is that we are probably going to hit the low traffic case in many cases, potentially confusing users, while the high traffic cases remain covered -- and additionally, instinctively "45 out of 100 failed in the last 10s" seems easier to understand.

This, however, means that the messages in linkerd2-proxy-api, while workable, are (ab)used to signal different things (ie. decay would now be a window length), and while we don't need technically to change the API, it'd be good to do so (we can tune the annotation surface and still use the same 0.20.0 shape though, but that is confusing).

**Rolling back to use an EWMA is still an option**, though, so if you'd prefer that make it clear in the comments.


linkerd/app/outbound/src/http/breaker/success_rate.rs
```diff
@@ -0,0 +1,480 @@
...
+/// One millisecond matches the moving average's own decay floor.
+const MIN_BUCKET_WIDTH: Duration = Duration::from_millis(1);
+
+/// Floor for the whole window, the smallest span that still gives one floored
+/// bucket per slice. The window is `bucket_width * BUCKETS`, so flooring `decay`
+/// here rather than flooring each bucket keeps the realized window equal to the
+/// configured `decay` for any value at or above this floor. Flooring the bucket
+/// width alone would instead round a decay between this floor and one bucket up
+/// to this span without the caller asking for it.
+const MIN_DECAY: Duration =
```
adleong: Throughout this PR we use the name "decay" for what I think is the "window".  I understand we have "decay" in the proxy API because we had planned to use EWMA here originally.  But I think we should use the more accurate name "window" throughout the proxy code and just include a note where we translate the "decay" parameter in the proxy API and use it for the "window".  

If time permits, we can go back and rename this field in the proxy API.

cratelyn (approved):

linkerd/app/outbound/src/http/breaker/unified.rs
```diff
@@ -0,0 +1,2094 @@
...
+    }
+}
+
+/// Configuration for [`UnifiedBreaker::new`].
+///
+/// The engine takes plain primitive parameters, and the config layer reduces its
+/// policy types to these before it builds the breaker, so the state machine never
+/// depends on how the configuration is represented.
+#[derive(Debug)]
+pub(crate) struct UnifiedBreakerConfig {
```
adleong: Isn't this struct identical to UnifiedBreaker, besides having public fields?

linkerd/app/outbound/src/http/breaker/unified.rs
```diff
@@ -0,0 +1,2094 @@
...
+    }
+
+    /// Probation state that admits one probe request, bounded by `deadline`.
+    ///
+    /// `gate.limit(1)` admits one request. `Some(class)` holds the probe verdict,
+    /// a `None` verdict means the probe produced no class within `deadline`, and an
+    /// `Err` means the gate was lost. The probe may give no class when a
+    /// `ResponseFuture` is cancelled before its head resolves, or when a
+    /// classification send drops on a full channel. The deadline then keeps the
+    /// loop moving. A probe that is never routed cannot block the breaker or hold
```
adleong: I don't understand this need for a deadline or protection against an unrouted probe.  The existing consecutive failures breaker has no such protections.  Are we sure this is necessary?

linkerd/app/outbound/src/http/breaker/retry_after.rs
```diff
@@ -0,0 +1,1167 @@
+//! Retry-After and grpc-retry-pushback-ms parsing for rate limiting responses.
```
adleong: These RetryAfterStore and GrpcRetryPushbackStore introduce a lot of complexity that's challenging to reason about.  There's a fundamental mismatch here where circuit breaking is a property of the endpoint but the retry-after hints are a property of individual responses.  imagine a scenario where each responsive gave us a different retry-after value and many different responses contributed towards tripping the breaker.

what you have here is some kind of aggregate measure where we maintain a retry-after value for the endpoint which is influenced by each retry-after response.  this introduces complexity and coupling.
I'd recommend either:

* taking retry-after out of project scope here.  using retry-after to inform probation is potentially a good hint at how overloaded a server is, but using it in this way isn't what the spec intended and there's a fundamental mismatch between endpoint and response.  plus, our jittered exponential backoff is probably going to give us behavior that's good enough, even if the server happens to be signaling a longer retry-after.
* or, we can make a good faith attempt to incorporate the retry-after hint by only considering the current response (i.e. the one that tripped the breaker, or the probe response).  In this case, we can just put the hint on the Class and remove the stores.  It's not perfect, but it doesn't need to be: we're just treating retry-after as an advisory hint.

linkerd/app/outbound/src/http/breaker/wrap_classify.rs
```diff
@@ -0,0 +1,510 @@
+//! Per-endpoint gate wiring for the circuit breaker.
+//!
+//! Each balancer endpoint sits behind a [`svc::Gate`] whose readiness a breaker
+//! task controls, and the gate is paired with a response classifier that
+//! broadcasts each classification over the breaker's channel. The stores and the
+//! breaker task are built here once per endpoint, so a hint seen on one endpoint
+//! extends only that endpoint's backoff and one endpoint's failures never gate
+//! another.
+//!
+//! When failure accrual is set up for the target and the policy respects server
```
adleong: I think with my above suggestion about retry-after, this file would no longer be needed.

linkerd/app/outbound/src/http/concrete/balance.rs
```diff
@@ -43,7 +42,36 @@ pub type BalancerMetrics = BalancerMetricsParams<ConcreteLabels>;
...
+    }
+}
+
+impl<T> svc::Param<Load> for Balance<T> {
+    fn param(&self) -> Load {
+        self.load
+    }
+}
+
+impl<T> svc::Param<linkerd_proxy_client_policy::PenaltyPeakEwma> for Balance<T> {
```
adleong, line 60: I'm not as familiar with how stacks are put together and I don't think we should get into it here so this is not a blocker for this PR.  But I do want to raise it as a question for my own education and potentially a note for future cleanup:

Could Balance be parameterized on the specific Load type (`Balance<T, L>`) so that we can impl only the correct Param types for each Load measure?

linkerd/app/outbound/src/http/breaker/unified.rs
```diff
@@ -0,0 +1,2094 @@
...
+            return false;
+        }
+        match class {
+            // A 429 is already handled above, so the only remaining bar is a 5xx
+            // a client policy let through as a non-failure.
+            classify::Class::Http(Ok(status)) => !status.is_server_error(),
+            classify::Class::Http(Err(_)) => false, // 5xx failure.
+            // RESOURCE_EXHAUSTED is handled above, so any other accepted Ok code
+            // is a recovered probe.
+            classify::Class::Grpc(Ok(_)) => true,
```
adleong: I think only `Grpc(Ok(0))` should be treated as success here.  All non-zero gRPC status codes should be treated as failures.

linkerd/app/outbound/src/http/breaker/retry_after.rs
unleashed: I don't think retry hints are a property of individual responses, and the spec for Retry-After makes is quite generic and brief, referencing a follow-up request and scoped to the service. It's common however to see these headers used for a variety of different meanings, with some rate limiters using them to signal different things (like gating resource consumption rates, instead of scoping to the service). This is why there's a draft for rate limiting headers I worked on to provide a scope (and other details).

I think removing the support for hints detracts from the value proposition. If we see the headers in the outbound stack chances are high that they'll refer to the endpoint, and keeping the last-seen hint would be useful in cases where the last response does not provide it, which in an ideal world would not happen, but it can happen and you can safely assume the endpoint is probably as overloaded or more as it was when we last read the hint. Ultimately the motivation is to help avoid the extra load on the endpoint caused by quick probes sent to it that will be unlikely to succeed (the early backoff ones), using the most recent information we got with the deadline we got (so a 30s R-A doesn't necessarily delay probes for 30s, but until the time the R-A asked for if we tripped later), capping the value so that we still keep probing in a reasonable timeframe.

I accept the complexity argument, and "good enough" being a thing. I think we can do better, though, and some others do so (admittedly, their breaker architecture is different and they have specific separate retry layers, but we have to start somewhere).

With that said, I don't think we need a non-core feature now, and can rethink or refactor this, or one of your proposals, later on.

linkerd/app/outbound/src/http/breaker/success_rate.rs
unleashed: Agreed.

linkerd/app/outbound/src/http/breaker/unified.rs
unleashed: Yes, this needs cleaning up.

linkerd/app/outbound/src/http/breaker/wrap_classify.rs
unleashed: Yes, I think that's the case.

linkerd/app/outbound/src/http/concrete/balance.rs
unleashed: I think so, with a bit of refactoring.

linkerd/app/outbound/src/http/breaker/unified.rs
unleashed: You can have `Grpc(Ok(NOT_FOUND))`, and others, that shouldn't fail a probe, and note that `class.is_success()` returns true already for `Grpc(Ok(_))`.

I think I can remove a few of these entries and delegate to `class.is_success()` directly, though.

linkerd/app/outbound/src/http/breaker/unified.rs
unleashed: Yes. If a probe is sent we have a few scenarios in which the recv call is left pending forever, with the task parked and the gate left shut.

For example, a response to the probe may be received, but the classification is dropped because the channel is full, or no response is ever received, or the future is cancelled before the response head is resolved. In those cases we unpark the task via the deadline and try again with the next backoff.

adleong: The current implementation does not have this protection: https://github.com/linkerd/linkerd2-proxy/blob/main/linkerd/app/outbound/src/http/breaker/consecutive_failures.rs#L98-L105

Are you saying that there's a bug in that code? As far as I know this has never manifested.

unleashed: Yes, there's a bug in that code, but it's not like it's permanent or growing in severity over time. If for whatever reason in the readiness of the endpoint changes, or when the destination controller pushes an updated set of addresses, it resets.

adleong: We discussed keeping this consistent with the behavior of consecutive failures and fixing this bug in both breakers as a separate change.

linkerd/app/outbound/src/http/breaker/retry_after.rs
adleong: We discussed keeping this out of scope for this initial work.

linkerd/app/outbound/src/http/breaker/unified.rs
unleashed: Removed UnifiedBreakerConfig in b3adddb76ef171c5cc9cbd30501e08234d480481, and the breaker now takes its parameters directly like the consecutive-failures breaker.

linkerd/app/outbound/src/http/breaker/unified.rs
unleashed: Cleaned this up in eefc2bb0aa7b5821922b3e07cc7536f1e542aa6a -- there's arguably a bug I won't fix here (deferred to a separate PR) to keep parity with CF: gRPC's PERMISSION_DENIED is classified as an error and the breaker will count it as such, but it should be exempted from contributing towards a trip.

linkerd/app/outbound/src/http/breaker/unified.rs
unleashed: Partially removed via #4565 and then via aec5108ec3c2311bccfe4c0ee9d34c4f8f007295, 1f659ac04e5e345cdf1f49c8dcef45f34012d5ac, a6e552b97144d2864c77627ddf86a5689b872905, f4847af0996651fde66ccbb01ec9dcb72b2a5a6c, 43dd9ba3e7c9b6c3b173a56474d3210ce116cbfa, 57a13af6be4134aa3065f4cb482c96a674f59ef8, and 1eddfb03d7738afd0653d538e2463c2e85900a9a.

There's yet another bug reintroduced when merging #4565, which is the removal of the channel draining when probing: `while self.rsps.try_recv().is_ok() {}`. I assume this is made to keep parity with the existing breaker, so will defer to a separate fix.

linkerd/app/outbound/src/http/breaker/success_rate.rs
unleashed: I think this this has been fixed via the merging of #4565 and the note's been added in fbd850aed080842ce59b6feccbaeeb4281daa3d8.

linkerd/app/outbound/src/http/breaker/retry_after.rs
unleashed: Removed via #4565.

linkerd/app/outbound/src/http/breaker/wrap_classify.rs
unleashed: Removed via #4565.

linkerd/app/outbound/src/http/concrete/balance.rs
unleashed: Deferred.

linkerd/load-biaser/src/lib.rs
```diff
@@ -506,37 +517,29 @@ impl<F, Rsp, E, C> Future for LoadBiaserFuture<F, Rsp, C>
...
-                    drop(handle);
-                    None
-                } else {
-                    Some(handle)
-                };
-
-                // Successful responses keep the handle until completion.
-                // Failed responses pass `None` since their penalty was recorded
-                // above.
+                    let value = handle.shared.effective_rtt(hint, &resp);
```
adleong: Here we cache the hint and then immediately read it.  There isn't anywhere else we read this cached hint.  So I think we don't need this caching at all and can simply parse the hint and use it immediately.

adleong (approved):

linkerd/load-biaser/src/lib.rs
unleashed: Deferred to another PR.

unleashed:
Rebased, solved conflicts and removed TODO markers.

PR #4566, opened 2026-06-09
refactored unified breaker with retry after

adleong:
This PR implements my feedback from https://github.com/linkerd/linkerd2-proxy/pull/4561, simplifying retry-after hinting to only look at the most recent failure.  In order to accomplish this, we store the retry-after hint on the classification.

Tests have not been updated.


cratelyn:
my understanding is that this can be closed, because we opted to take #4565. is that correct @unleashed?

cratelyn:
> my understanding is that this can be closed, because we opted to take #4565. is that correct @unleashed?

i am going to close this since this wasn't acknowledged. please indicate if there is work here that needs to be continued.

PR #4571, opened 2026-06-11
refactor(load-biaser): Remove unused retry-after hint caching

adleong:
The load biaser uses `CachedRateLimitHint` to cache retry-after hints from a response in the response extension metadata.  However, the only place this is ever read is immediately after caching it.  Therefore, we can simplify by skipping the caching altogether and just using the value directly after it's parsed.

This also allows us to get rid of `CachedRateLimitHint` entirely.


cratelyn (approved):
this looks good to me, but i believe that @unleashed should approve this as the author of the underlying code in question.

PR #4586, opened 2026-07-07
refactor(tap): simplify client id types

cratelyn:
this commit performs a small refactor to the tap system.

we represent the client id's that are permitted as a hash set of
TLS identities. however, there is never more than a single identity. see
this line from `linkerd_app::env`:

```rust
vec![id]
    .into_iter()
    .map(tls::ClientId)
    .collect()
```

we take _one_ identity, provided by an environment variable, and then
put it into a HashSet for no discernible reason. this commit changes the
plumbing carrying this identifier so that we faithfully represent it as
a single identifier.

### DIFF
--- a/linkerd/app/src/env.rs
+++ b/linkerd/app/src/env.rs
@@ -879,8 +879,8 @@ pub fn parse_config<S: Strings>(strings: &S) -> Result<super::Config, EnvError> 
     };
 
     let tap = tap?
-        .map(|(addr, ids)| super::tap::Config::Enabled {
-            permitted_client_ids: ids,
+        .map(|(addr, permitted_client_id)| super::tap::Config::Enabled {
+            permitted_client_id,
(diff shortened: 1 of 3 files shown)


raykroeker (approved):
Looks good.  It took a minute to dig into the equality for the client id, and the environment parsing functions.

Is it worth adding test coverage for the change?

cratelyn:
> Looks good. It took a minute to dig into the equality for the client id, and the environment parsing functions.

makes sense! i figured this might present a nice opportunity to get acquainted with how things flow down from the environment and into the proxy.

> Is it worth adding test coverage for the change?

in this case i don't believe so. this is a change to how we're modeling our structures, without any logical changes that might introduce behavior changes. if we have a set of 1 identifier, and now represent it as a single identifier, the code still compiling is a good guarantee this is a well-formed change.

PR #4604, opened 2026-08-17
fix(route-http-match): adjust route matching to more closely align to…

anthoturc:
… gateway api httproute spec

also address edge case around ties between exact and prefix match

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2098,6 +2098,7 @@ dependencies = [
  "linkerd2-proxy-api",
  "rand 0.10.2",
  "regex",
+ "rstest",
  "thiserror",
  "tracing",
  "url",
@@ -3369,6 +3370,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "proc-macro-crate"
+version = "3.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e67ba7e9b2b56446f1d419b1d807906278ffa1a658a8a5d8a39dcb1f5a78614f"
+dependencies = [
+ "toml_edit",
+]
+
+[[package]]
 name = "proc-macro2"
 version = "1.0.106"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3685,6 +3695,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d6f6ff9a378485b298a5286656da665ba74413d36db0979633275d2e708145d4"
 
 [[package]]
+name = "relative-path"
+version = "1.9.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ba39f3699c378cd8970968dcbff9c43159ea4cfbd88d43c00b22f2ef10a435d2"
+
+[[package]]
 name = "resolv-conf"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


adleong:
Looking at how route precedence ordering happens, it seems like this logic is actually spread across a few different places and the result may not be what we want.

In the policy controller, routes are sorted by timestamp then by name: https://github.com/linkerd/linkerd2/blob/main/policy-controller/grpc/src/outbound.rs#L677 with older timestamps getting sorted to the front of the list.  This is intended to give HTTPRoutes with older timestamps higher priority.

Then in the proxy, we define a cmp which compares based on match length (and match type, with your change).

Finally, the proxy uses that cmp to pick the maximum match (https://github.com/linkerd/linkerd2-proxy/blob/main/linkerd/http/route/src/lib.rs#L93).  But if there are multiple maximal matches then `max` picks the last one.  According to the sorting from the policy controller, this would be the one with the latest timestamp rather than oldest.

So our priority matching may be wrong in multiple ways.  I'd recommend the following actions:
1. verify this priority matching behavior manually or with an integration test
2. as a short-term solution, we can update the match picking logic in the proxy to prefer earlier matches in the list when there's a tie for maximum instead of preferring the last one
3. as a long-term solution, it's confusing that the priority logic is split across two different locations (timestamp/name in one place, match length in another).  it would be better to do all the priority sorting in one place rather than splitting it up.  the full ordering rules are defined in https://gateway-api.sigs.k8s.io/reference/api-spec/main/spec/#httprouterule

anthoturc:
I agree that it's a little confusing for the matching to be split this way. I'm interested in the longer term fix here. I'll take some time to refactor and see where I land. I did add some unit tests for the precedence ordering but I can add an inter test for more complete verification.

I think in my ideal world, the config and spec are completely handled in the data plane.

edit: I'll create a separate task for doing the long term work. For now, I will go with the short-term approach and the integration test for verification.

adleong:
Something to keep in mind is that I don't think we send the resource timestamps to the proxy so the data plane may not have all the necessary information to determine the full priority order.  In my opinion, the control plane is likely a better place to do this sorting.

I'm not sure why we originally decided to do match length sorting in the data plane but it might have been because we were calculating the match length based on the length of the matched segment in the URI, which is only known at request time by the proxy.  But it looks like the spec https://gateway-api.sigs.k8s.io/reference/api-spec/main/spec/#httprouterule doesn't say we have to do this and leaves the priority order of Regex matches up to the implementation.

anthoturc:
the latest commit adds verification of the gateway htttproute spec via integration tests. also I did introduce a `find_max` that will propagate the "first" best match (as opposed to the last) in case of a tie

cratelyn:
i see that i have been requested as a reviewer for this.

> Something to keep in mind is that I don't think we send the resource timestamps to the proxy so the data plane may not have all the necessary information to determine the full priority order. In my opinion, the control plane is likely a better place to do this sorting.

i'd like to defer to @adleong, who has context for this already. if her advice is that the control plane would be a better place to do this sorting, particularly if the proxy does not have a complete picture of the information to determine priority order, perhaps we should investigate a fix there instead?

> 2. as a short-term solution, we can update the match picking logic in the proxy to prefer earlier matches in the list when there's a tie for maximum instead of preferring the last one

my understanding is that this branch currently performs option 2 suggested [above](https://github.com/linkerd/linkerd2-proxy/pull/4604#issuecomment-5319652485). if so, i would not block on this, but would want to see a ticket filed tracking the long-term solution described in step 3.

anthoturc:
@cratelyn, added you as a reviewer in case you had some cycles

the PR does implement option 2. i created this ticket for the long-term work in 3 and will start working on it as a fast follow to this

anthoturc:
fwiw, here is the follow up: https://github.com/linkerd/linkerd2/issues/15614

adleong (approved):

PR #4605, opened 2026-08-17
fix(http-host-matching): make http host matching case insensitive

anthoturc:
this will ensure that the host name matching is RFC compliant w.r.t DNS names (i.e. case-insensitive)

### DIFF
--- a/linkerd/http/route/src/http/match/host.rs
+++ b/linkerd/http/route/src/http/match/host.rs
@@ -56,7 +56,7 @@ impl MatchHost {
                 if !h.ends_with('.') {
                     host = host.strip_suffix('.').unwrap_or(host);
                 }
-                if h == host {
+                if h.eq_ignore_ascii_case(host) {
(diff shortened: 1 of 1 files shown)


adleong (commented):
I notice that in some places we're normalizing to lowercase and in other places we're doing a case insensitive compare.  Do we need both?

anthoturc:
I think we need both.

I had added the normalization for the case where the HTTPRoute had a host defined in an unexpected way "Example.Com" vs "example.com" (which are the same).

The case-insensitive check is there for the runtime `Host` header/authority value (which can be anything). During my testing, I saw that the `http` crate's `host()` function (on the authority) would return the original Host header without normalizing it.

anthoturc:
after some offline discussion, Alex and I agreed that we don't strictly need the normalization logic since the case-insensitive match check covers us even when the control plane sends "weird" looking host names.

adleong (approved):

PR #4616, opened 2026-08-26
fix(http): only emit exact size of unary bodies without trailers

### LINKED ISSUE #15414: http/retry: PeekTrailersBody reports exact DATA size when trailers are buffered, causing  intermittent gRPC-Web missing trailers
### What is the issue?

# http/retry: PeekTrailersBody reports exact DATA size when trailers are buffered, causing intermittent gRPC-Web missing trailers

## Summary

We are seeing intermittent gRPC-Web responses where the client receives HTTP 200 and the DATA frame, but the gRPC-Web trailer frame is missing.

The failing response looks like:

```text
http_status=200
http_content_length=-1
http_transfer_encoding=chunked
body_bytes=5
data_frames=1
has_trailer=false
trailer_bytes=0
grpc_web_code=unknown
```

The 5-byte body is the empty gRPC-Web DATA frame:

```text
00 00 00 00 00
```

The expected successful response is 26 bytes:

```text
5-byte DATA frame
+
gRPC-Web trailer frame containing grpc-status: 0
```

So the response is being terminated after DATA, before the trailer frame is delivered.

## Impact

gRPC-Web clients fail with errors equivalent to:

```text
missing trailer
```

or receive a 200 response that cannot be interpreted as a valid gRPC-Web response because `grpc-status` is absent.

## Observed environment

Observed with Linkerd proxy:

```text
edge-26.5.1
```

Topology:

```text
HTTP/1.1 gRPC-Web client
  -> meshed Traefik
  -> meshed API service
  -> meshed downstream gRPC service
```

The downstream service returns OK. The API service also logs successful egress.

## Regression finding

We added a deterministic unit test for `PeekTrailersBody`:

```text
DATA("hello") + immediate TRAILERS
-> trailers are peeked
-> size_hint.lower() == 5
-> size_hint.exact() == None
```

This checks the invariant:

> A response body that still has trailers to emit must not report an exact DATA-only size.

Results:

```text
edge-25.1.1 / proxy v2.271.0: passes
edge-25.5.1 / proxy v2.295.0: fails
current main before fix: fails
```

In failing versions, `PeekTrailersBody` reports:

```text
size_hint.exact() == Some(5)
```

even though a trailers frame is buffered and still needs to be emitted.

## Suspected regression point

Likely introduced by:

```text
dd4fbcdb feat(http/retry): model `PeekTrailersBody<B>` with `Frame<T>` (#3559)
```

The old implementation in proxy `v2.271.0` preserved the inner body size hint and only increased the lower bound for buffered DATA. Therefore, for DATA + trailers it did not advertise an exact size.

The newer `Frame<T>`-based implementation models DATA + immediate trailers as `Inner::Unary` and returns:

```rust
SizeHint::with_exact(data.remaining() as u64)
```

That reports the DATA length as the exact response body size, despite trailers being buffered.

## Why this is unsafe

For trailers-bearing responses, the DATA byte length is not enough to describe the full response stream. There is still trailing metadata to emit.

If `PeekTrailersBody` advertises:

```text
exact size = DATA bytes
```

downstream HTTP layers may treat the response as complete after those DATA bytes. In gRPC-Web, that can surface as:

```text
HTTP 200
body_bytes=5
missing grpc-status trailer
```

## Expected behavior

When `PeekTrailersBody` has buffered both DATA and trailers:

```text
size_hint.lower() = DATA length
size_hint.exact() = None
```

It can say "at least N DATA bytes are available", but must not say "the whole response is exactly N bytes".

## Actual behavior in affected versions

When `PeekTrailersBody` has buffered both DATA and trailers:

```text
size_hint.lower() = DATA length
size_hint.exact() = Some(DATA length)
```

## Proposed fix

Change `PeekTrailersBody::size_hint()` so that `Inner::Unary` only reports an exact size when no trailers are buffered.

Pseudo-code:

```rust
Inner::Unary {
    data: Some(Ok(data)),
    trailers,
} => {
    let size = data.remaining() as u64;

    if trailers.is_none() {
        return SizeHint::with_exact(size);
    }

    let mut hint = SizeHint::new();
    hint.set_lower(size);
    hint
}
```

## Proposed regression test

Add a test in:

```text
linkerd/http/retry/src/peek_trailers.rs
```

Test behavior:

```rust
#[tokio::test]
async fn peeking_data_with_trailers_does_not_report_exact_size() {
    let body = MockBody::default()
        .then_yield_data(Poll::Ready(data()))
        .then_yield_trailer(Poll::Ready(trailers()));

    let peek = PeekTrailersBody::read_body(body).await;

    assert!(peek.peek_trailers().is_some());
    assert_eq!(peek.size_hint().lower(), 5);
    assert_eq!(peek.size_hint().exact(), None);

    let (data, trailers) = collect(peek).await;
    assert_eq!(data, "hello");
    assert_eq!(trailers.unwrap().get("trailer").unwrap(), "shiny");
}
```

### How can it be reproduced?

https://github.com/kostyay/linkerd-peek-trailers-size-hint-repro/blob/main/tests/peek_trailers_size_hint.rs

### Logs, error output, etc

From our repro grpc-repro grpc-web-client, the client itself does not throw a library
 exception. It parses the HTTP response and reports the missing gRPC-Web trailer as fields:

 ```text
   http_status=200
   body_bytes=5
   data_frames=1
   has_trailer=false
   trailer_bytes=0
   grpc_web_code=unknown
   grpc_web_message=
 ```

 Meaning:

 - HTTP layer succeeded: 200
 - body contained only the gRPC-Web DATA frame: body_bytes=5
 - no gRPC-Web trailer frame was present: has_trailer=false
 - therefore no grpc-status could be parsed: grpc_web_code=unknown

 Equivalent real gRPC-Web/browser client error would usually be something like:

 ```text
   missing trailer
 ```

 or:

 ```text
   missing grpc-status
 ```

 Depending on client library, it may surface as:

 ```text
   ConnectError: [unknown] missing trailer
 ```

 or a generic Unknown/Internal gRPC-Web error because the HTTP response is 200 but lacks the
 required grpc-status trailer.

### output of `linkerd check -o short`

``` linkerd-version
  ---------------
  ‼ cli is up-to-date
      is running version 25.10.7 but the latest edge version is 26.6.3
      see https://linkerd.io/2/checks/#l5d-version-cli for hints

  control-plane-version
  ---------------------
  ‼ control plane is up-to-date
      is running version 26.6.1 but the latest edge version is 26.6.3
      see https://linkerd.io/2/checks/#l5d-version-control for hints
  ‼ control plane and cli versions match
      control plane running edge-26.6.1 but cli running edge-25.10.7
      see https://linkerd.io/2/checks/#l5d-version-control for hints

  linkerd-control-plane-proxy
  ---------------------------
  ‼ control plane proxies are up-to-date
      some proxies are not running the current version:
        * linkerd-destination-5fb4ccdcd5-9bn87 (edge-26.6.1)
        * linkerd-destination-5fb4ccdcd5-dsv4k (edge-26.6.1)
        * linkerd-destination-5fb4ccdcd5-q7bhg (edge-26.6.1)
        * linkerd-destination-5fb4ccdcd5-rznlp (edge-26.6.1)
        * linkerd-destination-5fb4ccdcd5-tbbsh (edge-26.6.1)
        * linkerd-identity-cb95d9dfb-4ffl5 (edge-26.6.1)
        * linkerd-identity-cb95d9dfb-7sdfk (edge-26.6.1)
        * linkerd-identity-cb95d9dfb-hxgv4 (edge-26.6.1)
        * linkerd-identity-cb95d9dfb-jjrbn (edge-26.6.1)
        * linkerd-identity-cb95d9dfb-tgscd (edge-26.6.1)
        * linkerd-proxy-injector-656d99fbf7-5bcbt (edge-26.6.1)
        * linkerd-proxy-injector-656d99fbf7-cq4tb (edge-26.6.1)
        * linkerd-proxy-injector-656d99fbf7-nqvsv (edge-26.6.1)
        * linkerd-proxy-injector-656d99fbf7-q4zfn (edge-26.6.1)
        * linkerd-proxy-injector-656d99fbf7-tf7wm (edge-26.6.1)
      see https://linkerd.io/2/checks/#l5d-cp-proxy-version for hints
  ‼ control plane proxies and cli versions match
      linkerd-destination-5fb4ccdcd5-9bn87 running edge-26.6.1 but cli running edge-25.10.7
      see https://linkerd.io/2/checks/#l5d-cp-proxy-cli-version for hints

  linkerd-multicluster
  --------------------
  ‼ Link and CLI versions match
              * cgc-platform-us-central1: unable to determine version
          * platform: CLI version is edge-25.10.7 but Link version is edge-25.1.1
      see https://linkerd.io/2/checks/#l5d-multicluster-links-version for hints
  ‼ extension is managing controllers
              * using legacy service mirror controller for Link: cgc-platform-us-central1
          * using legacy service mirror controller for Link: platform
      see https://linkerd.io/2/checks/#l5d-multicluster-managed-controllers for hints
  × probe services able to communicate with all gateway mirrors
          Failed to get gateway status for cgc-platform-us-central1: no -admin port found for
  container linkerd-service-mirror-cgc-platform-us-central1-5d7d48585-th42x/service-mirror

      Failed to get gateway status for platform: no -admin port found for container
  linkerd-service-mirror-platform-59cfcc6d7d-k56mg/service-mirror

      see https://linkerd.io/2/checks/#l5d-multicluster-gateways-endpoints for hints
  ‼ multicluster extension proxies are up-to-date
      some proxies are not running the current version:
        * linkerd-gateway-d49994c49-8psx2 (edge-26.6.1)
        * linkerd-gateway-d49994c49-flrqw (edge-26.6.1)
        * linkerd-gateway-d49994c49-g5g58 (edge-26.6.1)
        * linkerd-local-service-mirror-68699668c6-kt2f2 (edge-26.6.1)
        * linkerd-service-mirror-cgc-platform-us-central1-5d7d48585-th42x (edge-26.6.1)
        * linkerd-service-mirror-platform-59cfcc6d7d-k56mg (edge-26.6.1)
        * linkerd-gateway-d49994c49-8psx2 (edge-26.6.1)
        * linkerd-gateway-d49994c49-flrqw (edge-26.6.1)
        * linkerd-gateway-d49994c49-g5g58 (edge-26.6.1)
        * linkerd-local-service-mirror-68699668c6-kt2f2 (edge-26.6.1)
        * linkerd-service-mirror-cgc-platform-us-central1-5d7d48585-th42x (edge-26.6.1)
        * linkerd-service-mirror-platform-59cfcc6d7d-k56mg (edge-26.6.1)
      see https://linkerd.io/2/checks/#l5d-multicluster-proxy-cp-version for hints
  ‼ multicluster extension proxies and cli versions match
      linkerd-gateway-d49994c49-8psx2 running edge-26.6.1 but cli running edge-25.10.7
      see https://linkerd.io/2/checks/#l5d-multicluster-proxy-cli-version for hints

  Identical to the earlier run — one × (multicluster gateway probe), the rest ‼ version-skew
  warnings.```

### Environment

- Cluster: GKE

### Possible solution

_No response_

### Additional context

_No response_

### Would you like to work on fixing this bug?

None

### PR DESCRIPTION
**Pull request by cratelyn**
## background
  
  fixes linkerd/linkerd2#15414.
  
  we received reports where grpc-web clients would intermittently receive
  responses where the trailers were missing.
  
  while RFC 9113 § 8.1.1 states:
  
  > A request or response that includes message content can include a
  content-length header field. A request or response is also malformed if
  the value of a content-length header field does not equal the sum of the
  DATA frame payload lengths that form the content, unless the message is
  defined as having no content.
  
  grpc-web is a separate protocol that can be sent over any HTTP/*
  version, and has some differences from HTTP/2 that are outlined in more
  detail here:
  
  <https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md>
  
  importantly, there is "no dependency on HTTP/2 specific framing", and
  "response statuses [are] encoded as part of the response body".
  
  moreover, many HTTP/1.1 implementations, including intermediary "middle
  boxes", might not support trailers in practice. RFC 9110 § 6.5.1
  "_Limitations on Use of Trailers_" says:
  
  > Trailer fields can be difficult to process by intermediaries that
  forward messages from one protocol version to another. If the entire
  message can be buffered in transit, some intermediaries could merge
  trailer fields into the header section (as appropriate) before it is
  forwarded. However, in most cases, the trailers are simply discarded.
  
  the latter part of this directly affects the `PeekTrailersBody<B>`
  middleware, which does attempt to buffer an entire message in transit.
  
   ## changes
  
  this commit adds an additional arm to the `match` block in
  `<PeekTrailersBody<B> as Body>::size_hint()`.
  
  this has an important effect: we no longer emit an *exact* size hint about
  the size of a peeked body when trailers are present. this preserves a
  lower-bound, so that the proxy can still performantly reserve a
  reasonable amount of memory for the inner body in advance.
  
  test coverage introduced in this branch is updated to show that unary
  bodies affected by this change no longer hint about an upper-bound if
  trailers are present. furthermore, the regression test provided now also
  passes.

### DIFF
--- a/linkerd/http/retry/src/peek_trailers.rs
+++ b/linkerd/http/retry/src/peek_trailers.rs
@@ -373,6 +373,8 @@ mod tests {
         let peek = PeekTrailersBody::read_body(empty).await;
         assert!(peek.peek_trailers().is_none());
         assert!(peek.is_end_stream());
+        assert_eq!(peek.size_hint().lower(), 0);
+        assert_eq!(peek.size_hint().upper(), None);
 
         let (data, trailers) = collect(peek).await;
         assert_eq!(data, "");
@@ -386,6 +388,8 @@ mod tests {
         let peek = PeekTrailersBody::read_body(only_trailers).await;
         assert!(peek.peek_trailers().is_some());
         assert!(peek.is_end_stream().not());
+        assert_eq!(peek.size_hint().lower(), 0);
+        assert_eq!(peek.size_hint().upper(), None);
 
         let (data, trailers) = collect(peek).await;
         assert_eq!(data, "");
@@ -401,6 +405,8 @@ mod tests {
         let peek = PeekTrailersBody::read_body(body).await;
         assert!(peek.peek_trailers().is_some());
         assert!(peek.is_end_stream().not());
+        assert_eq!(peek.size_hint().lower(), 5);
+        assert_eq!(peek.size_hint().upper(), Some(5));
 
         let (data, trailers) = collect(peek).await;
         assert_eq!(data, "hello");
@@ -417,6 +423,8 @@ mod tests {
         let peek = PeekTrailersBody::read_body(body).await;
         assert!(peek.peek_trailers().is_none());
         assert!(peek.is_end_stream().not());
+        assert_eq!(peek.size_hint().lower(), 5);
+        assert_eq!(peek.size_hint().upper(), None);
 
         let (data, trailers) = collect(peek).await;
         assert_eq!(data, "hello");
@@ -433,6 +441,8 @@ mod tests {
         let peek = PeekTrailersBody::read_body(body).await;
         assert!(peek.peek_trailers().is_some());
         assert!(peek.is_end_stream().not());
+        assert_eq!(peek.size_hint().lower(), 5);
+        assert_eq!(peek.size_hint().upper(), Some(5));
 
         let (data, trailers) = collect(peek).await;
         assert_eq!(data, "hello");
@@ -449,6 +459,8 @@ mod tests {
         let peek = PeekTrailersBody::read_body(body).await;
         assert!(peek.peek_trailers().is_none());
         assert!(peek.is_end_stream().not());
+        assert_eq!(peek.size_hint().lower(), 10);
+        assert_eq!(peek.size_hint().upper(), None);
 
         let (data, trailers) = collect(peek).await;
         assert_eq!(data, "hellohello");
@@ -467,6 +479,8 @@ mod tests {
         let peek = PeekTrailersBody::read_body(body).await;
         assert!(peek.peek_trailers().is_none());
         assert!(peek.is_end_stream().not());
+        assert_eq!(peek.size_hint().lower(), 5);
+        assert_eq!(peek.size_hint().upper(), None);
 
         let (data, trailers) = collect(peek).await;
         assert_eq!(data, "hellohello");
--- a/linkerd/http/retry/src/peek_trailers.rs
+++ b/linkerd/http/retry/src/peek_trailers.rs
@@ -486,4 +486,22 @@ mod tests {
         assert_eq!(data, "hellohello");
         assert_eq!(trailers.unwrap().get("trailer").unwrap(), "shiny");
     }
+
+    /// Regression test for linkerd/linkerd2-proxy#15414.
+    #[tokio::test]
+    async fn peeking_data_with_trailers_does_not_report_exact_size() {
+        let body = MockBody::default()
+            .then_yield_data(Poll::Ready(data()))
+            .then_yield_trailer(Poll::Ready(trailers()));
+
+        let peek = PeekTrailersBody::read_body(body).await;
+
+        assert!(peek.peek_trailers().is_some());
+        assert_eq!(peek.size_hint().lower(), 5);
+        assert_eq!(peek.size_hint().exact(), None);
+
+        let (data, trailers) = collect(peek).await;
+        assert_eq!(data, "hello");
+        assert_eq!(trailers.unwrap().get("trailer").unwrap(), "shiny");
+    }
 }
--- a/linkerd/http/retry/src/peek_trailers.rs
+++ b/linkerd/http/retry/src/peek_trailers.rs
@@ -299,10 +299,26 @@ where
             Inner::Passthru { inner } => inner.size_hint(),
             Inner::Unary {
                 data: Some(Ok(data)),
-                ..
+                trailers: None,
             } => {
                 let size = data.remaining() as u64;
                 http_body::SizeHint::with_exact(size)
+            }
+            Inner::Unary {
+                data: Some(Ok(data)),
+                trailers: Some(_),
+            } => {
+                // NB: An exact size hint is not provided if we have any trailers to emit.
+                // Protocols like grpc-web encode the response status as part of the response body,
+                // so providing an exact hint would might cause clients to see errors related to
+                // missing trailers.
+                //
+                // For more information, see:
+                // https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2.
+                let size = data.remaining() as u64;
+                let mut hint = http_body::SizeHint::new();
+                hint.set_lower(size);
+                hint
             }
             Inner::Unary {
                 data: None | Some(Err(_)),
@@ -406,7 +422,7 @@ mod tests {
         assert!(peek.peek_trailers().is_some());
         assert!(peek.is_end_stream().not());
         assert_eq!(peek.size_hint().lower(), 5);
-        assert_eq!(peek.size_hint().upper(), Some(5));
+        assert_eq!(peek.size_hint().upper(), None);
 
         let (data, trailers) = collect(peek).await;
         assert_eq!(data, "hello");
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: linkerd/http/retry/src/peek_trailers.rs [hunk 1]
```diff
@@ -467,9 +495,29 @@ mod tests {
...
         assert!(peek.is_end_stream().not());
+        assert_eq!(peek.size_hint().lower(), 5);
+        assert_eq!(peek.size_hint().upper(), None);
 
         let (data, trailers) = collect(peek).await;
         assert_eq!(data, "hellohello");
         assert_eq!(trailers.unwrap().get("trailer").unwrap(), "shiny");
     }
+
+    /// Regression test for linkerd/linkerd2-proxy#15414.
```
anthoturc: super nit:

```suggestion
    /// Regression test for linkerd/linkerd2#15414.
```

anthoturc (approved):

anthoturc:
thank you for the detailed description and links to relevant RFCs!

linkerd/http/retry/src/peek_trailers.rs
cratelyn: super nit, super appreciated! :bowing_woman: fixed in 85cafd6.

PR #4617, opened 2026-08-28
feat: add outbound mTLS-required default policy

### LINKED ISSUE #15615: Add an option (policy) to enforce mTLS for outbound traffic
### What problem are you trying to solve?

A meshed client will send plaintext traffic when a Kubernetes Service points to a non-meshed pod.
For example, a Service initially points to a meshed server. If its selector or endpoints are changed to point to a non-meshed pod, the client’s Linkerd proxy receives no destination identity and connects without mTLS.
Inbound authorization does not protect against this because the traffic no longer reaches the original server.
I need a way for clients to require mTLS and fail closed when the destination is not meshed.

### How should the problem be solved?

Add an outbound policy that requires mTLS for selected Services or workloads.
When enabled, the outbound proxy should reject the connection if:
- The destination has no Linkerd identity.
- mTLS cannot be established.
Linkerd should never fall back to plaintext when this policy applies.

### Any alternatives you've considered?

Inbound authorization can require authenticated traffic at the server, but it does not protect the client if a Service is redirected to another backend.
Kubernetes NetworkPolicy can restrict network access, but it cannot verify a Linkerd destination identity or guarantee that traffic uses Linkerd mTLS.

### How would users interact with this feature?

A simple namespace or workload annotation could require mTLS:
```
metadata:
  annotations:
    config.linkerd.io/default-outbound-policy: all-authenticated
```

Or A set of CRDs that specify fine-grained policy for specific ports, routes, workloads, etc.

### Would you like to work on this feature?

maybe

gnarlex:
### Problem

The outbound proxy always falls back to a cleartext connection when a target endpoint has no mesh identity (i.e. is not meshed). Operators who want to restrict a workload to only communicate with meshed targets (or targets outside the cluster) cannot enforce this: unmeshed, in-cluster destinations are silently reached in the clear.

### Solution

Add `LINKERD2_PROXY_OUTBOUND_DEFAULT_POLICY`, the outbound counterpart to `LINKERD2_PROXY_INBOUND_DEFAULT_POLICY`. Whether a target is meshed is only known once an endpoint is resolved (a meshed endpoint carries a mesh identity in its discovery metadata), so the policy is enforced at connect time by a new `RequireMeshIdentity` guard in the shared TCP connect stack. The guard refuses a cleartext connection when discovery provided no mesh identity for the endpoint
(`ConditionalClientTls::None(NotProvidedByServiceDiscovery)`).

Supported values:
- `all-unauthenticated` (default): unchanged; cleartext permitted.
- `all-authenticated`: require mTLS for every outbound endpoint.
- `cluster-authenticated`: require mTLS only for endpoints within `LINKERD2_PROXY_POLICY_CLUSTER_NETWORKS`; egress outside the cluster is permitted in the clear.

Other cleartext reasons (loopback, identity disabled, ingress without discovery) are never refused, and the default (`all-unauthenticated`) is byte-for-byte unchanged.

### Validation

Added unit tests for env parsing (accepted values, the cluster-networks requirement, rejected values, unset) and behavioral tests for the connect-time guard (in-cluster unmeshed refused; out-of-cluster unmeshed permitted under cluster-authenticated; `all-authenticated` refuses any unmeshed target; non-discovery cleartext permitted; default never refuses).

---

Fixes linkerd/linkerd2#15615

### DIFF
--- a/linkerd/app/outbound/src/tcp/connect.rs
+++ b/linkerd/app/outbound/src/tcp/connect.rs
@@ -1,8 +1,9 @@
-use crate::Outbound;
+use crate::{policy, Outbound};
 use futures::future;
 use linkerd_app_core::{
     io, svc, tls,
     transport::{addrs::*, ConnectTcp},
+    IpMatch,
 };
 use std::task::{Context, Poll};
 
@@ -17,14 +18,47 @@ pub struct Connect {
 #[derive(Clone, Debug)]
 pub struct PreventLoopback<S>(S);
 
+/// Enforces the outbound default policy's mTLS requirement: refuses a cleartext
+/// connection to a target that discovery did not associate with a mesh
+/// identity.
+///
+/// Whether a target is meshed is expressed by its
+/// [`tls::ConditionalClientTls`]: a meshed endpoint resolves to
+/// `Some(..)` (mTLS), while an unmeshed one resolves to
+/// `None(NotProvidedByServiceDiscovery)`. Only the latter is subject to
+/// enforcement — other cleartext reasons (loopback, identity disabled, ingress
+/// without discovery) are always permitted.
+#[derive(Clone, Debug)]
+pub struct RequireMeshIdentity<S> {
+    inner: S,
+    requirement: Requirement,
+}
+
+#[derive(Clone, Debug)]
+enum Requirement {
+    /// No enforcement; cleartext to unmeshed targets is permitted
+    /// (`all-unauthenticated`).
+    Disabled,
+    /// Every unmeshed target is refused (`all-authenticated`).
+    All,
+    /// Only unmeshed targets within these networks are refused; targets outside
+    /// the cluster are permitted (`cluster-authenticated`).
+    WithinCluster(IpMatch),
+}
+
 // === impl Outbound ===
 
 impl Outbound<()> {
-    pub fn to_tcp_connect(&self) -> Outbound<PreventLoopback<ConnectTcp>> {
-        let connect = PreventLoopback(ConnectTcp::new(
+    pub fn to_tcp_connect(&self) -> Outbound<PreventLoopback<RequireMeshIdentity<ConnectTcp>>> {
+        let connect = ConnectTcp::new(
             self.config.proxy.connect.keepalive,
             self.config.proxy.connect.user_timeout,
-        ));
+        );
+        let requirement = Requirement::new(
+            self.config.default_policy,
+            self.config.cluster_networks.clone(),
+        );
+        let connect = PreventLoopback(RequireMeshIdentity::new(connect, requirement));
         self.clone().with_stack(connect)
     }
 }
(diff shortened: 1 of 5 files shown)


linkerd/app/outbound/src/tcp/connect.rs
```diff
@@ -17,14 +18,47 @@ pub struct Connect {
...
+        let connect = ConnectTcp::new(
             self.config.proxy.connect.keepalive,
             self.config.proxy.connect.user_timeout,
-        ));
+        );
+        let requirement = Requirement::new(
+            self.config.default_policy,
+            self.config.cluster_networks.clone(),
+        );
+        let connect = PreventLoopback(RequireMeshIdentity::new(connect, requirement));
```
thesw4rm, line 61: Brand new to this project, with some Rust experience. Can we fix the variable names?

```suggestion
        let connect = ConnectTcp::new(
            self.config.proxy.connect.keepalive,
            self.config.proxy.connect.user_timeout,
        );
        let tlsAuthRequirement = Requirement::new(
            self.config.default_policy,
            self.config.cluster_networks.clone(),
        );
        self.clone().with_stack(
            PreventLoopback(RequireMeshIdentity::new(connect, requirement))
        )
```

linkerd/app/outbound/src/tcp/connect.rs
```diff
@@ -17,14 +18,47 @@ pub struct Connect {
...
+/// enforcement — other cleartext reasons (loopback, identity disabled, ingress
+/// without discovery) are always permitted.
+#[derive(Clone, Debug)]
+pub struct RequireMeshIdentity<S> {
+    inner: S,
+    requirement: Requirement,
+}
+
+#[derive(Clone, Debug)]
+enum Requirement {
```
thesw4rm, line 38: Requirement on its own might not be descriptive enough. Maybe

```suggestion
enum TLSAuthRequirement
```

linkerd/app/outbound/src/tcp/connect.rs
```diff
@@ -17,14 +18,47 @@ pub struct Connect {
...
+/// Whether a target is meshed is expressed by its
+/// [`tls::ConditionalClientTls`]: a meshed endpoint resolves to
+/// `Some(..)` (mTLS), while an unmeshed one resolves to
+/// `None(NotProvidedByServiceDiscovery)`. Only the latter is subject to
+/// enforcement — other cleartext reasons (loopback, identity disabled, ingress
+/// without discovery) are always permitted.
+#[derive(Clone, Debug)]
+pub struct RequireMeshIdentity<S> {
+    inner: S,
+    requirement: Requirement,
```
thesw4rm, line 34:
```suggestion
    tlsAuthRequirement: TLSAuthRequirement,

thesw4rm (commented):
I'm not currently associated to this project. I was hoping to make my first contribution here. Please feel free to ignore my review!

PR #4542, opened 2026-05-27
fix(proxy): add max_header_list_size config for HTTP/2 client

wahajahmed010:
After the Hyper 0.14→1.x migration, the HTTP/2 client defaults to `DEFAULT_MAX_HEADER_LIST_SIZE = 16 KiB`, which is too restrictive for gRPC services with large trailers.

This change adds client-side configurability of the max header list size via the `LINKERD2_PROXY_HTTP2_SETTINGS_MAX_HEADER_LIST_SIZE` proxy environment variable.

Note that the new config is not set through the API. However, the setting can be configured through the `config.linkerd.io/proxy-additional-env` annotation

fixes linkerd/linkerd2#15199

PR #4618, opened 2026-08-31
feat(http): add max connection age for HTTP/2 server connections

ergdevops:
Addresses linkerd/linkerd2#15623.

## Problem

Pooled proxy-to-proxy HTTP/2 connections currently live forever: TCP and HTTP/2 keepalives keep them healthy indefinitely, `MAX_IDLE_CONNS_PER_ENDPOINT` (default 10,000) only caps the count, the HTTP/1 pool idle timeout doesn't apply to them, and the outbound discovery idle timeout only fires when a service receives no traffic at all. As far as we can tell there is no setting that closes a pooled connection for being idle or bounds its lifetime (checked `linkerd/app/src/env.rs` on `main` and back through `v2.311.0`).

This matters for memory: each held server-side connection retains its receive-buffer high-water mark — bounded by the connection window (1MB default) — for its entire lifetime. In our mesh (arm64 EKS), a large caller fleet holds ~300 pooled connections into each server pod; after ~300 rps load bursts we measured ~700KB retained per connection, i.e. ~200MB of proxy RSS per pod that was only ever released when peer pods restarted. Verified via `tcp_open_total`/`tcp_close_total` accounting on the admin endpoint — the pooling itself works as designed; the cost is the unbounded buffer lifetime.

## Change

Adds a `max_connection_age` setting to the shared HTTP/2 `ServerParams`, configurable per server prefix like every other `SERVER_HTTP2` setting:

- `LINKERD2_PROXY_INBOUND_SERVER_HTTP2_MAX_CONNECTION_AGE` (duration) — the inbound proxy-to-proxy listener, where pooled peer connections accumulate;
- `LINKERD2_PROXY_OUTBOUND_SERVER_HTTP2_MAX_CONNECTION_AGE` (duration) — the outbound (app-facing) listener, for symmetry with the other paired settings.

**Unset (the default) preserves current behavior exactly.** The admin server builds its params with `Default::default()` and is unaffected regardless of environment.

When set, the HTTP/2 serve loop arms a timer as an additional `select!` branch alongside the existing drain/teardown triggers, reusing the same `graceful_shutdown()` (GOAWAY) path: no new streams are accepted, in-flight streams run to completion, then the connection closes and the client reconnects transparently — the identical sequence every pod drain already exercises. A deterministic per-connection jitter of up to +10% (derived from the client's ephemeral port; no new dependency) prevents synchronized shutdown storms. Each shutdown is logged at DEBUG with the client address.

Includes an integration test covering: requests served normally before the age elapses, an in-flight stream completing across the age boundary, and the connection closing once drained — plus an env test asserting the setting parses independently under both server prefixes.

## Results from our environment (15m age, production-like mesh)

- Proxy RSS became a bounded ~15-minute sawtooth instead of a monotonic plateau — buffers released every cycle with no pod restarts. (Observed on a build with jemalloc enabled on aarch64 — see #4613 — so freed memory was actually returned to the OS.)
- Zero failed requests across many recycle cycles; p50/p99 unchanged through recycle waves.
- Cost: one TLS handshake per connection per age period, spread over ~90s by the jitter — below measurement noise in-cluster.

Similar in spirit to gRPC's `MAX_CONNECTION_AGE`; follows the same params/env plumbing pattern as #4542.

### DIFF
--- a/linkerd/app/src/env/http2.rs
+++ b/linkerd/app/src/env/http2.rs
@@ -9,6 +9,11 @@ pub(super) fn parse_server<S: Strings>(
     Ok(ServerParams {
         flow_control: Some(parse_flow_control(strings, base)?),
         keep_alive: parse_keep_alive(strings, &format!("{base}_KEEP_ALIVE"))?,
+        max_connection_age: parse(
+            strings,
+            &format!("{base}_MAX_CONNECTION_AGE"),
+            parse_duration,
+        )?,
         max_concurrent_streams: parse(
             strings,
             &format!("{base}_MAX_CONCURRENT_STREAMS"),
@@ -107,6 +112,7 @@ mod tests {
         env.insert("TEST_KEEP_ALIVE_INTERVAL", "2s");
         env.insert("TEST_INITIAL_STREAM_WINDOW_SIZE", "1");
         env.insert("TEST_INITIAL_CONNECTION_WINDOW_SIZE", "2");
+        env.insert("TEST_MAX_CONNECTION_AGE", "900s");
         let expected = h2::ServerParams {
             flow_control: Some(h2::FlowControl::Fixed {
                 initial_stream_window_size: 1,
@@ -116,6 +122,7 @@ mod tests {
                 interval: Duration::from_secs(2),
                 timeout: Duration::from_secs(1),
             }),
+            max_connection_age: Some(Duration::from_secs(900)),
             max_concurrent_streams: Some(3),
             max_frame_size: Some(4),
             max_header_list_size: Some(5),
--- a/linkerd/app/src/env/http2.rs
+++ b/linkerd/app/src/env/http2.rs
@@ -154,4 +154,31 @@ mod tests {
             }
         );
     }
+
+    /// Like every other SERVER_HTTP2 setting, max connection age is parsed
+    /// under both the inbound and outbound server prefixes.
+    #[test]
+    fn max_connection_age_parses_for_both_servers() {
+        let mut env = HashMap::default();
+        env.insert(
+            "LINKERD2_PROXY_INBOUND_SERVER_HTTP2_MAX_CONNECTION_AGE",
+            "15m",
+        );
+        env.insert(
+            "LINKERD2_PROXY_OUTBOUND_SERVER_HTTP2_MAX_CONNECTION_AGE",
+            "1h",
+        );
+
+        let inbound = parse_server(&env, "LINKERD2_PROXY_INBOUND_SERVER_HTTP2").unwrap();
+        assert_eq!(
+            inbound.max_connection_age,
+            Some(Duration::from_secs(15 * 60))
+        );
+
+        let outbound = parse_server(&env, "LINKERD2_PROXY_OUTBOUND_SERVER_HTTP2").unwrap();
+        assert_eq!(
+            outbound.max_connection_age,
+            Some(Duration::from_secs(60 * 60))
+        );
+    }
 }
--- a/linkerd/http/h2/src/lib.rs
+++ b/linkerd/http/h2/src/lib.rs
@@ -5,6 +5,11 @@ pub struct ServerParams {
     pub flow_control: Option<FlowControl>,
     pub keep_alive: Option<KeepAlive>,
     pub max_concurrent_streams: Option<u32>,
+    /// If set, server connections are gracefully shut down (GOAWAY) after this
+    /// age, bounding how long a pooled peer connection (and the buffer
+    /// high-water memory it retains) can live. A per-connection jitter of up
+    /// to +10% is applied to avoid synchronized shutdowns.
+    pub max_connection_age: Option<Duration>,
 
     // Internals
     pub max_frame_size: Option<u32>,
--- a/linkerd/http/h2/src/lib.rs
+++ b/linkerd/http/h2/src/lib.rs
@@ -5,10 +5,6 @@ pub struct ServerParams {
     pub flow_control: Option<FlowControl>,
     pub keep_alive: Option<KeepAlive>,
     pub max_concurrent_streams: Option<u32>,
-    /// If set, server connections are gracefully shut down (GOAWAY) after this
-    /// age, bounding how long a pooled peer connection (and the buffer
-    /// high-water memory it retains) can live. A per-connection jitter of up
-    /// to +10% is applied to avoid synchronized shutdowns.
     pub max_connection_age: Option<Duration>,
 
     // Internals
--- a/linkerd/proxy/http/src/server.rs
+++ b/linkerd/proxy/http/src/server.rs
@@ -8,9 +8,10 @@ use std::{
     future::Future,
     pin::Pin,
     task::{Context, Poll},
+    time::Duration,
 };
 use tower::Service;
-use tracing::{debug, Instrument};
+use tracing::{debug, info, Instrument};
 
 #[cfg(test)]
 mod tests;
@@ -36,6 +37,7 @@ pub struct ServeHttp<N> {
     version: Variant,
     http1: hyper::server::conn::http1::Builder,
     http2: hyper::server::conn::http2::Builder<TokioExecutor>,
+    max_connection_age: Option<Duration>,
     inner: N,
     drain: drain::Watch,
 }
@@ -70,6 +72,7 @@ where
             keep_alive,
             flow_control,
             max_concurrent_streams,
+            max_connection_age,
             max_frame_size,
             max_header_list_size,
             max_send_buf_size,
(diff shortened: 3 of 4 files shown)


linkerd/http/h2/src/lib.rs
```diff
@@ -5,6 +5,11 @@ pub struct ServerParams {
     pub flow_control: Option<FlowControl>,
     pub keep_alive: Option<KeepAlive>,
     pub max_concurrent_streams: Option<u32>,
+    /// If set, server connections are gracefully shut down (GOAWAY) after this
+    /// age, bounding how long a pooled peer connection (and the buffer
+    /// high-water memory it retains) can live. A per-connection jitter of up
+    /// to +10% is applied to avoid synchronized shutdowns.
```
thesw4rm: This might be too much context for a struct parameter. Also, none of the other variables are documented within the struct.

linkerd/proxy/http/src/server.rs
```diff
@@ -154,17 +158,18 @@ where
...
         let http1 = self.http1.clone();
         let http2 = self.http2.clone();
+        let max_connection_age = self.max_connection_age;
 
         let res = io.peer_addr().map(|pa| {
             let (handle, closed) = ClientHandle::new(pa);
             let svc = self.inner.new_service(handle.clone());
             let svc = SetClientHandle::new(handle, svc);
-            (svc, closed)
+            (svc, closed, pa)
```
thesw4rm: nit: writing out the full `peer_addr` might be clearer

linkerd/proxy/http/src/server.rs
```diff
@@ -154,17 +158,18 @@ where
...
         });
 
         Box::pin(
             async move {
-                let (svc, closed) = res?;
+                let (svc, closed, peer) = res?;
```
thesw4rm: nit: `peer` is a different variable name than `pa` in `let res = ...`. I suggest changing this also to `peer_addr`

linkerd/proxy/http/src/server.rs
```diff
@@ -201,6 +206,24 @@ where
...
+                        let max_age = async move {
+                            match max_connection_age {
+                                Some(age) => {
+                                    let jitter =
+                                        age.mul_f64(f64::from(peer.port() % 128) / 1280.0);
+                                    tokio::time::sleep(age + jitter).await;
+                                }
+                                None => std::future::pending().await,
+                            }
+                        };
```
thesw4rm: Using jitter to shut down expired TCP connections can lead to very racy behavior. Using the source port is clever to add better randomness, but connections won't necessarily be started at the same time and synchronized shutdowns are still very possible.

At a larger scale, the sheer number of expired HTTP/2 connections can still result in races. Can we instead maintain an in-memory structure to store all the connections, and have an async job to clean them up? Similar to how a `TCP_CLOSE` state is handled in the Linux kernel.

linkerd/proxy/http/src/server.rs
```diff
@@ -216,6 +239,11 @@ where
                                 Pin::new(&mut conn).graceful_shutdown();
                                 conn.await?;
                             }
+                            () = &mut max_age => {
+                                debug!(client.addr = %peer, "Max connection age reached; gracefully shutting down the connection");
+                                Pin::new(&mut conn).graceful_shutdown();
+                                conn.await?;
+                            }
```
thesw4rm, line 269: Very minor nit: This code can be combined with the previous case to prevent duplication. Probably would end up being a larger refactor though. Left an example just in case.
```rust
enum Teardown {
    /// Client closed the connection; nothing left to do.
    ClientClosed,
    /// Process draining: GOAWAY, then hold the release guard until close.
    Drain(drain::ReleaseShutdown),
    /// Stack teardown or max age: GOAWAY, then close.
    Graceful,
}

let teardown = tokio::select! {
    res = &mut conn => {
        debug!(?res, "The client is shutting down the connection");
        res?;
        Teardown::ClientClosed
    }
    shutdown = drain.signaled() => {
        debug!("The process is shutting down the connection");
        Teardown::Drain(shutdown)
    }
    () = closed => {
        debug!("The stack is tearing down the connection");
        Teardown::Graceful
    }
    () = &mut max_age => {
        debug!(client.addr = %peer, "Max connection age reached; gracefully shutting down the connection");
        Teardown::Graceful
    }
};

match teardown {
    Teardown::ClientClosed => {}
    Teardown::Drain(shutdown) => {
        Pin::new(&mut conn).graceful_shutdown();
        shutdown.release_after(conn).await?;
    }
    Teardown::Graceful => {
        Pin::new(&mut conn).graceful_shutdown();
        conn.await?;
    }
}
```

ergdevops: thank you @thesw4rm for the inputs will look into all the comments and do test out things on test env and one the commit is done will update you.

linkerd/http/h2/src/lib.rs
ergdevops: Done in 2bac29a — dropped the doc comment to match the struct's style; the context now lives in the commit message.

linkerd/proxy/http/src/server.rs
ergdevops: Done in 2bac29a — renamed to `peer_addr`.

linkerd/proxy/http/src/server.rs
ergdevops: Done in 2bac29a — both are `peer_addr` now, consistent through the closure and the serve loop.

linkerd/proxy/http/src/server.rs
ergdevops: Adopted in 2bac29a — thanks for sketching it out! Took your `Teardown` enum essentially as-is (ClientClosed / Drain(guard) / Graceful), so the select arms only classify and a single match performs the shutdown. Validated on a test cluster: max-age recycles fire with jitter as expected and pod drain remains clean.

linkerd/app/src/env/http2.rs
```diff
@@ -9,6 +9,11 @@ pub(super) fn parse_server<S: Strings>(
     Ok(ServerParams {
         flow_control: Some(parse_flow_control(strings, base)?),
         keep_alive: parse_keep_alive(strings, &format!("{base}_KEEP_ALIVE"))?,
+        max_connection_age: parse(
+            strings,
+            &format!("{base}_MAX_CONNECTION_AGE"),
```
thesw4rm, line 14: The PR title says it targets only inbound HTTP/2 connections. However, this field will also target a `LINKERD2_PROXY_OUTBOUND_SERVER_HTTP2_MAX_CONNECTION_AGE`. Is that intentional?

ergdevops: @thesw4rm you are right , i see every other SERVER_HTTP2 setting already comes as an inbound/outbound pair, I should also make this test follow the same standard. 
Once done i will update the PR and issue metadata accordingly.

ergdevops: The feature code already supported both — the setting sits on the shared `ServerParams` and parses through the common `SERVER_HTTP2` plumbing, so nothing was inbound-specific. I've added a test covering the inbound and outbound prefixes explicitly, and updated the title/description to document both env vars.

ergdevops:
@thesw4rm With the symmetry addressed and tests covering both prefixes, I believe the feature is complete and the PR title now matches the actual scope of the change — but happy to take any further feedback.

Separately, since you've contributed here before: could you help me understand the cadence the linkerd maintainers typically follow for reviewing and merging feature PRs like this? I have this PR and one other (#4613) waiting on maintainer review, and I'd like to set my expectations right — and do whatever I can from my side to make them easy to review.

thesw4rm:
@ergdevops I have never contributed here before haha. I'm just going around reviewing PRs

linkerd/app/src/env/http2.rs
thesw4rm: The issue still mentions `inbound` [specifically](https://github.com/linkerd/linkerd2/issues/15623) 

> A configurable max connection age (and/or idle timeout) for inbound HTTP/2 server connections

I think the problem isn't clear whether it should apply only to inbound, or both inbound/outbound. I don't see a problem with a max age for both, but I'm pretty new to this project.

ergdevops:
> @thesw4rm With the symmetry addressed and tests covering both prefixes, I believe the feature is complete and the PR title now matches the actual scope of the change — but happy to take any further feedback.
> 
> Separately, since you've contributed here before: could you help me understand the cadence the linkerd maintainers typically follow for reviewing and merging feature PRs like this? I have this PR and one other (#4613) waiting on maintainer review, and I'd like to set my expectations right — and do whatever I can from my side to make them easy to review.

@adleong apologies for the tag but could you please help me with this request, thank you.

